PR #5, opened 2016-05-07
Added a missing param in a signature

alex:


damienmg:
Hello, thanks you for this contribution. Unfortunately we are removing this ruleset (I will delete it today) from the bazel repository. In the same way than any other rules, this rule should be found at https://github.com/bazelbuild/rules_rust.

Sorry for the inconvenience.

damienmg:
oops gosh I though this was on the bazel repository.

LGTM

PR #6, opened 2016-05-10
Fix labels for Rust binaries to use those under @io_bazel_rules_rust

davidzchen:
Since the Rust rules have been deleted from @bazel_tools, we should use the
labels for the Rust binaries under this workspace instead.


damienmg:
LGTM

PR #19, opened 2016-07-25
Add a dependency on the cc_wrapper, so it's present when rust_binary …

kchodorow:
…is executed

Otherwise, the local_config_cc isn't written to the execroot on OS X because it
isn't a dependency of the build rule.

Fixes https://github.com/bazelbuild/bazel/issues/1556.


davidzchen:
LGTM. Thanks!

PR #23, opened 2016-09-02
Add crosstool files input to more action

### LINKED ISSUE #20: Missing dependency causes breakage on Mac
Breakage: http://ci.bazel.io/job/rules_rust/BAZEL_VERSION=HEAD,PLATFORM_NAME=darwin-x86_64/259/console

Reason: we now have sandbox on Mac, and the missing dependency cc_wrapper is not in the sandbox so we can't find it.

Suggested fix: add dependency //tools/defaults:crosstool to every rule that need gcc.

damienmg:
Add it to all the necessary action for `bazel test //...`
to pass all tests on Darwin.

Fixes #20.

### DIFF
--- a/rust/rust.bzl
+++ b/rust/rust.bzl
@@ -353,7 +353,8 @@ def _rust_binary_impl(ctx):
       depinfo.transitive_libs +
       [ctx.file._rustc] +
       ctx.files._rustc_lib +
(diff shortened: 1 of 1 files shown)


davidzchen:
LGTM. Thanks!

PR #25, opened 2016-11-28
Update example subdirectory to be a local repository

katre:
This makes the examples slightly clearer by using a syntax that is closer to how end-users will use the rust rules, and helps to ensure that example rules can't be accidentally used from the main rust rules.


katre:
The latest changes to the test/ package won't work properly until https://bazel-review.googlesource.com/c/7571/ is merged and released.

katre:
I've added a fork of test_rules.bzl so that tests pass, we can remove it when released Bazel includes https://github.com/bazelbuild/bazel/commit/b89f30e129a4ef6eb1e81fb865117e4845892d74

davidzchen:
Sounds good. Thanks, @katre. Is this ready for review now?

katre:
Yes, with the fork of test_rules.bzl this is ready for review.

When the fixed test_rules.bzl is released (next month, I guess?) I'll send another PR to remove it.

davidzchen:
LGTM. Thanks!

PR #33, opened 2017-02-17
Add example of FFI with interop between Rust and C.

davidzchen:


davidzchen:
Depends on bazelbuild/rules_rust#34

davidzchen:
+cc @acmcarther @mattmoor

examples/matrix/src/matrix.c
```diff
@@ -0,0 +1,124 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
```
damienmg: Is this original code? can we just vendor existing code for those kind of examples?

davidzchen: Yes, this is original code. I wanted a small example that tested an interesting set of FFI features, such as wrapping a C struct rather than just an opaque pointer, making a safe Rust wrapper over the C API, etc. I'm planning to reuse this example elsewhere, such as for rules_d, etc.

PR #34, opened 2017-02-17
Allow rust_test and rust_binary to depend on cc_library targets.

davidzchen:


davidzchen:
Test added in bazelbuild/rules_rust#33

damienmg:
I don't really understand what that PR is doing, maybe also add @mhlopko for C++

davidzchen:
Thanks, Damien! :)

rust/rust.bzl
```diff
@@ -366,7 +367,7 @@ def _rust_binary_impl(ctx):
   depinfo = _setup_deps(ctx.attr.deps,
                         ctx.label.name,
                         output_dir,
-                        is_library=False)
+                        allow_cc_deps=False)
```
asayers: Looks like you forgot to actually allow rust_binary rules to depend on cc_library

Issue #36, opened 2017-02-21
rust_library should produce non-rlib crates with correct file extensions


davidzchen:
Currently, the `rust_library` rule only outputs rlibs regardless of the value of `crate_type`. In order to correctly support all the different [crate types](https://doc.rust-lang.org/reference.html#linkage), `rust_library should produce the crates with the correct file extensions.

### Proposal

* Make the `rust_library` **rule** internal by renaming it to `_rust_library`
* Add a new `outs` attribute to the `_rust_library` rule to support [custom outputs](https://bazel.build/versions/master/docs/skylark/cookbook.html#outputs-custom)
* Add a `rust_library` macro that wraps the `_rust_library` rule. Based on the value of `crate_type`, it adds a `_rust_library` rule with `outs` set with the name of the output library with the correct file extensions according to the [Rust documentation](https://doc.rust-lang.org/reference.html#linkage)

davidzchen:
+cc @dslomov @laurentlb @acmcarther @mattmoor @damienmg

damienmg:
SGTM

acmcarther:
This was fixed in:
https://github.com/bazelbuild/rules_rust/pull/38

... which produces outputs with the right file extension for the toolchain and crate_type.

PR #38, opened 2017-02-22
Support crate_type file extensions more fully

acmcarther:
Removes the hard coded rust_lib output, and declares the output file within the rust_library_impl. This lets us vary the file extension based on the toolchain -- OSX produces `.dylib` where linux would produce `.so`, for example.

Original description follows:
---
Tentatively addresses #36.

WIP: Need tests and understand platform implications

@davidzchen: Do we need the additional complexity of a hidden build rule? This change seems to work on my platform to produce dylibs that can be loaded correctly.

I've no idea how to handle different platforms. I added it as an additional attr, but I think a `config_setting` probably makes more sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ The master branch should always be current with the latest bazel, as such you ca
 ## rust_library
 
 ```python
-rust_library(name, srcs, crate_root, deps, data, crate_features, rustc_flags)
+rust_library(name, srcs, crate_root, crate_type, deps, data, crate_features, rustc_flags)
 ```
 
 <table class="table table-condensed table-bordered table-params">
@@ -112,7 +112,24 @@ rust_library(name, srcs, crate_root, deps, data, crate_features, rustc_flags)
           if <code>srcs</code> contains only one file.
         </p>
       </td>
-    </td>
+    </tr>
+    <tr>
+      <td><code>crate_type</code></td>
+      <td>
+        <code>String, optional</code>
+        <p>
+          The type of crate to be produced during library compilation. This
+          list closely matches Cargo's own notion of crate-type, and the
+          available options are "lib", "rlib", "dylib", "cdylib", "staticlib",
+          and "proc-macro".
+        </p>
+        <p>
+          The exact output depends on the selected toolchain but generally will
+          match what Cargo would do. If binary compilation is desired, use
+          <code>rust_binary</code> instead of the "bin" crate type.
+        </p>
+      </td>
+    </tr>
     <tr>
       <td><code>deps</code></td>
       <td>
--- a/rust/rust.bzl
+++ b/rust/rust.bzl
@@ -208,8 +208,8 @@ def _determine_lib_name(name, crate_type, toolchain):
     # All platforms produce 'rlib' here
     extension = ".rlib"
   elif crate_type == "bin":
-    fail("crate_type of 'bin' was detected. Please compile this crate as a " +
-         "rust_binary.")
+    fail("crate_type of 'bin' was detected in a rust_library. Please compile "
+         + "this crate as a rust_binary instead.")
 
   if not extension:
     fail(("Unknown crate_type: %s. If this is a cargo-supported crate type, "
@@ -570,6 +570,10 @@ Args:
 
     If `crate_root` is not set, then this rule will look for a `lib.rs` file or
     the single file in `srcs` if `srcs` contains only one file.
+  crate_type: The type of linkage to use for building this library. Options
+    include "lib", "rlib", "dylib", "cdylib", "staticlib", and "proc-macro"
+
+    The exact output file will depend on the toolchain used.
   deps: List of other libraries to be linked to this library target.
 
     These can be either other `rust_library` targets or `cc_library` targets if
--- a/rust/rust.bzl
+++ b/rust/rust.bzl
@@ -204,7 +204,7 @@ def _determine_lib_name(name, crate_type, toolchain):
     extension = toolchain.dylib_ext
   elif crate_type == "staticlib":
     extension = toolchain.staticlib_ext
-  elif crate_type in ("rlib", "lib"):
+  elif crate_type in ("lib", "rlib"):
     # All platforms produce 'rlib' here
     extension = ".rlib"
   elif crate_type == "bin":
--- a/rust/rust.bzl
+++ b/rust/rust.bzl
@@ -55,13 +55,6 @@ RUST_FILETYPE = FileType([".rs"])
 
 A_FILETYPE = FileType([".a"])
 
-LIBRARY_CRATE_TYPES = [
-    "lib",
-    "rlib",
-    "dylib",
-    "staticlib",
-]
-
 # Used by rust_doc
 HTML_MD_FILETYPE = FileType([
     ".html",
@@ -205,6 +198,27 @@ def _find_crate_root_src(srcs, file_names=["lib.rs"]):
       return src
   fail("No %s source file found." % " or ".join(file_names), "srcs")
 
+def _determine_lib_name(name, crate_type, toolchain):
+  extension = None
+  if crate_type in ("dylib", "cdylib"):
+    extension = toolchain.dylib_ext
+  elif crate_type == "staticlib":
+    extension = toolchain.staticlib_ext
+  elif crate_type in ("rlib", "lib", "proc-macro"):
+    # All platforms produce 'rlib' here
+    extension = ".rlib"
+  elif crate_type == "bin":
+    fail("crate_type of 'bin' was detected. Please compile this crate as a " +
+         "rust_binary.")
+
+  if not extension:
+    fail(("Unknown crate_type: %s. If this is a cargo-supported crate type, "
+         + "please file an issue!") % crate_type)
+
+
+  return "lib{name}{extension}".format(name=name,
+                                       extension=extension)
+
 def _crate_root_src(ctx, file_names=["lib.rs"]):
   if ctx.file.crate_root == None:
     return _find_crate_root_src(ctx.files.srcs, file_names)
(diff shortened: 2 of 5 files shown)


rust/rust.bzl
```diff
@@ -167,6 +188,19 @@ def _get_features_flags(features):
     features_flags += ["--cfg feature=\\\"%s\\\"" % feature]
   return features_flags
 
+def _get_crate_type_and_target_outputs(name, crate_type, platform):
+  extension = LIBRARY_AND_PLATFORM_TO_EXTENSION[crate_type or "rlib", platform or "linux"]
```
davidzchen: Break up this line. See: https://bazel.build/versions/master/docs/skylark/bzl-style.html

rust/rust.bzl
```diff
@@ -629,15 +663,14 @@ _rust_toolchain_attrs = {
...
 }
 
 rust_library = rule(
     _rust_library_impl,
     attrs = _rust_library_attrs + _rust_toolchain_attrs,
     fragments = ["cpp"],
-    outputs = {
-        "rust_lib": "lib%{name}.rlib",
-    },
+    outputs = _get_crate_type_and_target_outputs,
```
davidzchen: This does not seem idiomatic Skylark. Also, it is unclear to me how this would actually work where we pass in a function for the `outputs` parameter of `rule`. Is this an obscure Bazel/Skylark feature? I do not see it documented as a way to specify outputs in both the [Skylark documentation](https://bazel.build/versions/master/docs/skylark/rules.html#output-files) and the [Skylark Cookbook](https://bazel.build/versions/master/docs/skylark/cookbook.html#outputs-default).

@laurentlb @dslomov Can you comment on this?

rust/rust.bzl
```diff
@@ -25,6 +25,27 @@ LIBRARY_CRATE_TYPES = [
     "staticlib",
 ]
```
davidzchen: Also, please rebase and resolve merge conflicts.

rust/rust.bzl
damienmg: This is a problem on our side, that should be documented. I believe @fweikert is the one who implemented the callback for outputs.

fweikert: I didn't implement the feature (it has been in Bazel since April 2015), but I have a pending change that improves the documentation for implicit outputs in general (including callbacks).

Currently, the documentation of the "outputs" parameter mentions callbacks:
https://bazel.build/versions/master/docs/skylark/lib/globals.html#rule

damienmg:
Removing myself as David can perfectly manage this review

rust/rust.bzl
davidzchen: Ah! Good to know. Thanks for the clarification. Looking forward to the documentation.

davidzchen:
@acmcarther This change LGTM and will be good go after rebasing and the minor style comment above. Thanks for working on this!

rust/rust.bzl
acmcarther: I was curious where I even got this from.... I made this change a while ago for my own use but couldn't remember how I'd found this feature, let alone find it again in the documentation.

Thanks all!

acmcarther:
@davidzchen Do you have any advice on resolving the platform issue? I'm not confident "platform" should be part of the build rules, but I'm also not sure I want to even solve the problem in this PR.

We could kick the can down the road by assuming everyone is linux... That's *still* an improvement over the current state, which just expected an rlib output...

davidzchen:
@acmcarther The only way I am aware of to get the host platform as a string would be to use [`ctx.fragments.cpp.cpu`](https://bazel.build/versions/master/docs/skylark/lib/fragments.html), but that requires `ctx`, which is only available in the rule implementation function. In this case, we want that information in the `outputs` callback. I also tried setting the `default` for the `platform` attribute to a `select` based on the `//rust:darwin` and `//rust:k8` `config_setting`s:

```python
_rust_library_attrs = {
    ...
    "platform": attr.string(
        default = select({
            "//rust:darwin": "darwin",
            "//rust:k8": "linux",
        })),
}
```

However, this does not work because `default` only accepts a `string`.

@damienmg, @fweikert - Is it currently possible to get this information from a `outputs` callback? If not, what is the most idiomatic way to implement this behavior?

acmcarther:
- Rebased
- Addressed comments
- Added dylib building rule

@davidzchen Rebase killed the commit your comment was on. Can you ack the review?

I'm pretty unsatisfied with the current platform situation, but I think we can improve the situation later.

acmcarther:
@davidzchen This is ready, but I can't address your review comment because the commit is gone. PTAL

rust/rust.bzl
acmcarther: This is done (and was done a while ago)

Hoping posting this comment will complete the review.

davidzchen:
Sorry for the late reply. Let's go ahead and get this committed. Can you add a TODO comment to address the platforms issue in the previous comments and rebase?

acmcarther:
Rebased. I still cant clear up the "requested changes" snippet. Looks to be stuck...

acmcarther:
OK:

This is broken in a non-trivial way. The (new?) build bot is also building for darwin. This PR adds a rust_library of type "dylib" that is producing a `.dylib` file on darwin. This PR was attempting to kick platform detection down the road by including a `platform` attr. That's not going to work in this situation.

@davidzchen: Do you have any suggestions on proceeding? I can make the required change, but it might take a while. I don't think theres a "short cut" here -- rustc will *always* produce platform dependent `outs` for dylibs, and thats a sore point for Bazel.

mfarrugi:
This looks like it's also a blocker for dynamically linking cc_libs, as https://github.com/bazelbuild/rules_rust/blob/master/rust/rust.bzl#L169 also takes the shortcut of assuming only static linking.

Is this stalled on anything in particular, @davidzchen?

acmcarther:
This needs a second look to either make it smarter (w.r.t. platform extensions), or a trick to get the tests to pass.

I'll take a look at updating this PR this weekend.

mfarrugi:
@acmcarther I started adding support for linking to native dylibs, but `bazel run` and `bazel test` don't appear to pass LD_LIBRARY_PATH to the binary. 

Did this change actually support running or testing crates downstream of dylibs? It would be straightforward to add a wrapper similar to the bench rule, but I don't want to conflict with this change or redo work.

mfarrugi:
@acmcarther What were the test failures or the fix for this? CI doesn't have logs for it.

acmcarther:
mfarrugi@ Expected output file not produced under OSX. rustc is producing xyz.dylib, where xyz.so is expected.

Need to figure out a trick, perhaps using the toolchain, to dictate the expected output file.

acmcarther:
There are three moving parts as I see it here:

1) The selected toolchain
2) The desired output target
3) The desired crate type

(1) dictates the defaults for all crate types. (2) selects and overrides the values decided by (1), while (3) narrows the output down to a single possible value. (2) and (3) are already knowable within the output filepath determining callback, but I can't find any mechanism to determine (1). I [dropped a line in bazel-discuss](https://groups.google.com/forum/#!topic/bazel-discuss/fQuhTifRG1g) to see what might be available.
___

A totally different approach could be evaluated as well. `Rustc` does support an `-o OUTPUT_NAME` directive, which should let us force the file to have a consistent extension. This isn't easy to use for two reasons though
1) Using `-o` requires us to specify the full file path, not just as directory as we're currently doing. 
2) _I believe_ rustc relies on the output file extension to know how to handle the dependency. 

If (2) is true, we'd need to retain metadata with the rust_library to know how to turn it back into something rustc knows how to link.

acmcarther:
Apparently (from the bazel-discuss post) you do not need to specify the outputs directly -- you can specify them in the impl with a call to "declare_outputs". This lets us avoid the callback shenanegans entirely and defer this bit of planning until the impl. By the time we enter the impl, we have all the attrs and the toolchain.

Will update with this fix some time today.

acmcarther:
Tests now pass! @mfarrugi: PTAL at your earliest convenience. Hopefully this works with https://github.com/bazelbuild/rules_rust/pull/61 with minimal changes.

README.md
```diff
@@ -112,7 +112,24 @@ rust_library(name, srcs, crate_root, deps, data, crate_features, rustc_flags)
...
+    </tr>
+    <tr>
+      <td><code>crate_type</code></td>
+      <td>
+        <code>String, optional</code>
+        <p>
+          The type of crate to be produced during library compilation. This
+          list closely matches Cargo's own notion of crate-type, and the
+          available options are "lib", "rlib", "dylib", "cdylib", "staticlib",
+          and "proc-macro".
```
acmcarther: This isn't evaluated as the required crate dependencies are impractical to import without `cargo-raze`.

For the record, the required crates are
```
extern crate proc_macro;
extern crate syn;
extern crate quote;
```

rust/rust.bzl
```diff
@@ -205,6 +198,27 @@ def _find_crate_root_src(srcs, file_names=["lib.rs"]):
       return src
   fail("No %s source file found." % " or ".join(file_names), "srcs")
 
+def _determine_lib_name(name, crate_type, toolchain):
+  extension = None
+  if crate_type in ("dylib", "cdylib"):
+    extension = toolchain.dylib_ext
+  elif crate_type == "staticlib":
+    extension = toolchain.staticlib_ext
+  elif crate_type in ("rlib", "lib", "proc-macro"):
```
mfarrugi: proc-macro should be a dylib (though I don't know if rustc actually cares)

acmcarther: Ah this is important. We're relying on rustc to "produce the right thing" without us telling it what to make. Indeed, it is producing a dylib in this case, so we need to expect a dylib. Not sure how I became under the impression that this was an rlib -- my local code pre-osx fix used dylib here.

Fixed.

acmcarther:
ok to test

acmcarther:
@buchgr No problem, thanks for helping resolve the issue. 

I suspected it was some form of webui after leafing through https://github.com/bazelbuild/continuous-integration and not seeing a secret code phrase.

buchgr:
@acmcarther all future pull requests from you should trigger CI without a verification step. I have added an ACL passed on github collaborators. external contributors will require a project collaborator to approve it.  please file an issue on our continous-integration repo if that's not the case

PR #45, opened 2017-04-07
Move rust repository setup to a new Skylark file.

katre:


davidzchen:
Can you please take a look at the build failure?

The change itself looks reasonable. For my understanding, what is the background for this?

katre:
Fixed the rustlib vs rust_lib issue.

katre:
The background is the work I am doing in Bazel on platforms and toolchains. This is a pre-requsite to adding a rust_toolchain rule and having that be used with Bazel.

The actual toolchain part will have to wait for several releases until the feature is released, but I am using this as a test case (see https://github.com/katre/rules_rust/tree/02-rust-toolchain-rule for the rust_toolchain rule itself).

davidzchen:
Understood. Thanks!

mattmoor:
This broke a whole bunch of documentation, because this no longer works:
```python
load("@io_bazel_rules_rust//rust:rust.bzl", "rust_repositories")
```

I would recommend simply loading `rust_repositories` into `rust.bzl` for backwards-compatibility.

katre:
This has been merged since April. I'll send a PR to update the documentation.

PR #48, opened 2017-07-05
Add out_dir_tar common argument for all rust_* rules

acmcarther:
**TL;DR: This facilitates running build scripts, packing up the genn'd files, and then "sideloading" the sources to be available for the include! macro. It doesn't actually *do* that, but it lets other build rules do that.**

This CL is testing the waters for one potential approach to implement build.rs support in rules_rust. It isn't intended to be used generally without good reason -- in most cases rule generating tooling should specify this argument. The way it works (in my own experimentation) is as follows:

A preplanned cargo dependency tree is constructed. For each crate, the targets are identified. For each "lib" target, a `rust_library` target is generated. For each "build-script" target, a  `rust_binary` and `genrule` are generated. The rust binary contains the build script sources, as well as all dependencies and build_dependencies. It is executed by the genrule -- which packages the output into a tar. This tar is linked to the rust_library.

"cargo-raze": https://github.com/google/cargo-raze

An example plan, from that repo and `bazel query --output=build //examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:all`

https://gist.github.com/acmcarther/b69c1a7eb6c713f69c67175336b1e6cf

The most important bits
```python
# /home/alex/projects/cargo-raze/examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2/BUILD:13:1
rust_library(
  name = "kernel32",
  srcs = ["//examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:src/lib.rs"],
  out_dir_tar = "//examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:kernel32_sys_build_script_executor",
)

# /home/alex/projects/cargo-raze/examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2/BUILD:13:1
genrule(
  name = "kernel32_sys_build_script_executor",
  tools = ["//examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:kernel32_sys_build_script"],
  outs = ["//examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:kernel32_sys_out_dir_outputs.tar.gz"],
)

# /home/alex/projects/cargo-raze/examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2/BUILD:13:1
rust_binary(
  name = "kernel32_sys_build_script",
  srcs = ["//examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:src/lib.rs"],
  crate_root = "//examples/complicated_cargo_library/vendor/kernel32-sys-0.2.2:build.rs",
)
```


acmcarther:
Minor "keep_alive" comment.

I was going to let this "fall on the floor", but it turns out that some crates _really_ want to run their build scripts. In experimentation, I found it impractical to try to work around the build script for `x11-dl`, `wayland-protocol`, and `wayland-client`, as they discover a wealth of information about the local machine and generate platform specific code in a manner that is very tedious to replicate by hand.

I'll look into cleaning this up (tests, passing build) so that this is mergeable soon. A more "modern" example here: https://github.com/acmcarther/cargo-raze-examples/blob/a1cd2bf6abf41eeb73972cde6cb5b0435cf1cfc9/bazel/complicated_cargo_library/vendor/x11-2.17.0/BUILD#L65

acmcarther:
I'm highly motivated to get this merged soonish (after careful review), as https://github.com/google/cargo-raze depends on this change to handle dependencies with build scripts.

I'm going to try to get the test passing and ready to go very soon.

examples/hello_out_dir/BUILD
```diff
@@ -0,0 +1,26 @@
...
+    "@io_bazel_rules_rust//rust:rust.bzl",
+    "rust_binary",
+    "rust_library",
+    "rust_test",
+)
+
+rust_library(
+    name = "hello_out_dir",
+    srcs = ["src/lib.rs"],
+    out_dir_tar = ":repacked_srcs.tar.gz",
```
johnynek: can you really only ever have one of these for a target? Are there no cases where you might have more than one?

acmcarther: It isn't possible to have more than one with Cargo -- and i didn't *expect* anyone to use this outside that context. I probably lack imagination on this point.

I don't think it would be much trouble to facilitate having more than one here but I'm not sure what the best way to handle files within the tars with the same name.

johnynek: we could fail the build if they collide, or just untar in order taking the rightmost. I would probably prefer failing I guess.... but if you think no one needs, just a comment.

acmcarther: Added a note to the readme about this. I believe this sort of use case could be supported on the user side by unzipping several archives and merging them, which lets a user choose the best way to solve this problem.

(also it's a cop out on my end :stuck_out_tongue:)

Definitely willing to revisit this if an issue is opened or I missed a common use case!

PR #52, opened 2017-08-25
Update rules_rust to use Bazel's toolchains feature.

### LINKED ISSUE #14: Make it easier to use Rust rules with different versions of Rust
Currently, the version of Rust used by the rules are not easily configurable. We should make it easier to use the Rust rules with different Rust versions, such as via a `rust_toolchain` rule (similar to `cc_toolchain`) or look at the approach taken by tools such as multirust.

It would also be good to make it easy to switch between having Bazel download a set of Rust binaries and using the ones installed on the local system.

katre:
This allows rust targets to only depend on (and thus download) the
toolchains that are used, dynamically chosen based on the current target
platform.

This requires use of Bazel 0.5.4, and will not work with earlier releases of Bazel.

Fixes #14.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -1,0 +1,201 @@
+"""
+Toolchain rules used by Rust.
+"""
+
+ZIP_PATH = "/usr/bin/zip"
+
+# Utility methods that use the toolchain provider.
+def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
+                         depinfo, rust_flags=[]):
+  """
+  Constructs the rustc command used to build the current target.
+  """
+
+  # Paths to cc (for linker) and ar
+  cpp_fragment = ctx.host_fragments.cpp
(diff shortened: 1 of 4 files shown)


katre:
@davidzchen Do you have time to review this, and do you have an opinion of the fact that it requires the latest Bazel? Should we cut a release of rules_rust before merging this, for people who haven't upgraded to use?

davidzchen:
Sorry for the delay. This looks great. Thanks, John!

rust/toolchain.bzl
```diff
@@ -0,0 +1,201 @@
...
+def _get_path_str(dirs):
+  return ":".join(dirs)
+
+def _get_first_file(input):
+  if hasattr(input, "files"):
+    for f in input.files:
+      return f
+  return input
+
+def _get_files(input):
```
ittaiz: this and _get_first_file seem like good candidates for skylib cc @laurentlb

PR #66, opened 2018-02-02
Fix runfiles test: Macos runfiles are quite different

acmcarther:
Apparently runfiles under macos have an important difference from Unix (and I presume, windows): I believe they live directly in the working directory, not "$target_name.runfiles"

This is a sloppy fix that changes the mechanics of the test according to the target os.

Notice: @mfarrugi

EDIT: Fix silly mistake regarding runfiles explanation: "data" is the data file itself.

### DIFF
--- a/examples/hello_runfiles/src/lib.rs
+++ b/examples/hello_runfiles/src/lib.rs
@@ -5,11 +5,15 @@ use std::path::PathBuf;
 pub fn get_runfiles_dir() -> io::Result<PathBuf> {
     let mut path = std::env::current_exe()?;
 
-    let mut name = path.file_name().unwrap().to_owned();
-    name.push(".runfiles");
-
-    path.pop();
-    path.push(name);
+    if cfg!(target_os = "macos") {
+      path.pop();
+      path.push("data");
+    } else {
+      let mut name = path.file_name().unwrap().to_owned();
+      name.push(".runfiles");
+      path.pop();
+      path.push(name);
+    }
 
     Ok(path)
 }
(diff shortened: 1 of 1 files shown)


acmcarther:
Adding mhlopko@ to see if there might be more context behind this discrepancy in the behavior of runfiles.

mfarrugi:
`data` is our name for the dir, not a bazel path afaik. It looks like on osx the working dir is the runfiles root maybe? strange that this isn't the case on linux.

In any case, maybe it's better to keep the rule modifications without the example, if the actual location of runfiles is ill-defined.

acmcarther:
Whoops, you're right. Updated description and fixed accordingly.

I don't have a strong opinion either way regarding the test but I do want to know why runfiles isn't consistent.

acmcarther:
@mhlopko @davidzchen

mhlopko:
Sorry for radio silence. Had hard time to find free evenings lately. In general I don't think you should hardcode runfiles location in your source code. Paths change depending on the platform (I'm not sure why there is a difference between mac and linux, I can find out if anybody cares :) but e.g. on windows we don't use symlinks at all and instead we provide runfiles.manifest that binary should read and use to find absolute paths.

@laszlocsomor did a ton of work on runfiles lately and provided libraries that your code can use to compute absolute runfile path in a reliable way. I'll try to find the documentation.

acmcarther:
FTR: Source here:
https://github.com/bazelbuild/bazel/tree/master/src/tools/runfiles

We would want to add a zero-dependency rust library there that does something analogous to cpp.

laszlocsomor:
FYI, state of runfiles libraries:
- Java is released with Bazel 0.11.0
- Python is ready but not yet released, need to move it to //tools/py/runfiles but otherwise it'll be ready
- C++ and Bash are partially implemented
- all runfiles libraries have to move to `@bazel_tools//tools/<language>/runfiles` (@lberki knows more why), the `@bazel_tools//tools/runfiles` location was the wrong idea

PR #75, opened 2018-02-27
python calling rust example

mfarrugi:
This is not well supported by python rules, but sorta works. https://github.com/bazelbuild/bazel/issues/1475 outlines the hacky setup.

Might also want to create a pyo3 or similar example outside of this repo.

### DIFF
--- a/examples/ffi/python_calling_rust/py_library_from_rust.bzl
+++ b/examples/ffi/python_calling_rust/py_library_from_rust.bzl
@@ -1,0 +1,61 @@
+# TODO: Native extensions are what we really want, cdll.LoadLibrary is not particularly great.
+def py_library_from_rust(name, rust_lib):
+    # 1. Put the library into _solib_ because of the ODR.
+    #
+    # nb. Without this, the .so doesn't make it to runfiles at all
+    native.cc_library(
+        name = "c_wrapper",
+        srcs = [rust_lib],
+    )
+
+    # 2. Elaborate dance to figure out the path to the shared object library in _solib_...
+    crate_name = rust_lib.split(":")[-1]
+    out = name + ".py"
+    native.genrule(
+        name = "__generate" + out,
+        srcs = [rust_lib],
+        outs = [out],
+        # TODO: Was there a template rule somewhere?
+        cmd = """
+        cat << EOM > $(OUTS)
+
+from __future__ import print_function
+
+import os
+from ctypes import cdll
+
+# TODO: This is a hack to compensate for rootpath not giving us the _solib_* path...
+def _find_runfiles_solib(file_name):
+    # nb. The runfiles root is already on python path, so we don't need a Runfiles object.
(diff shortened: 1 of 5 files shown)


acmcarther:
I can't unblock this for CI. Could I bother you to push a commit (perhaps removing "debug stuff", unless you think you'll need it to debug OSX)?

I think the ci pipeline forgot about this verification request and the link is now 404ing.

mfarrugi:
I don't think it's right referring to the mangled solib path, but don't have a good idea of what The Right Way is yet. You just need any commit to re-trigger CI?

acmcarther:
Oops, didn't submit my comment!

Yeah I just need a commit so that the CI sees it.

examples/ffi/python_calling_rust/py_library_from_rust.bzl
```diff
@@ -0,0 +1,61 @@
...
+        # TODO: Was there a template rule somewhere?
+        cmd = """
+        cat << EOM > $(OUTS)
+
+from __future__ import print_function
+
+import os
+from ctypes import cdll
+
+# TODO: This is a hack to compensate for rootpath not giving us the _solib_* path...
```
mfarrugi: I cannot quite put my finger on how hacky this is.. let me know what ya'll think.

mfarrugi: This ultimately searches all of runfiles, so unless the design of collecting shared libraries in one location changes dramatically, it should be safe. That said, searching should really not be necessary and I'm not sure if it's a deficiency in python/c++ rules or my lack of knowledge here.

mfarrugi:
I did this mostly to see how difficult / complicated setting up the integration was, I think it's reasonably straightforward (and would be dead simple if the $(rootpath)  gave the path we wanted).

I think the best case scenario for this sort of integration is for python to have the capacity to take a header and a shared library and generate all of the boilerplate, and for rust to be able to produce the shared library and header (with cbindgen). IIUC this is roughly how [milksnake](https://github.com/getsentry/milksnake) works, as opposed to directly creating a python native extension. The API would look roughly the same as the current example, however would work for functions that aren't `i32` in and `i32` out.

In general I don't have an immediate use for this, but I think it's nice to say "look, it's easy to use rust with xxx". It is okay with me for this to continue to sit in a pull request (it'll be a year old soon!), but I don't mind providing a similar level of guidance / support as with the rest of the rules.

aiuto:
I would rather not have it sit in a PR forever. if it works and we can put it in CI, then you should dust it off to commit it. If it doesn't really work, how about turning it into examples/ffi/python_calling_rust/readme.md. There you can describe the approach and how a more disciplined solution might work. A PR is essentially undiscoverable.

mfarrugi:
Will close this for now, should be just as easy to find.

Would make sense to revisit when bazel-python has support for native extensions.

Issue #64, opened 2017-12-11
Document configuring custom toolchains


mfarrugi:
This [WORKSPACE](https://github.com/mfarrugi/cargo-raze-example-stdx/blob/master/WORKSPACE#L7) and [custom_rust_toolchain.bzl](https://github.com/mfarrugi/cargo-raze-example-stdx/blob/master/bazel/custom_rust_toolchain.bzl) is an example of setting up toolchains outside of the default one that is part of these rules.

Is the toolchain feature stable enough to update the documentation to recommend this flavor of custom toolchain configuration? Is it possible to configure targets to require different toolchains (eg. targets that are nightly only)?

mfarrugi:
This is what I'm using right now, which grabs the latest versions asap and exposes rustfmt for easy use.

```python
load(
    "@io_bazel_rules_rust//rust:repositories.bzl",
    "BUILD_for_compiler",
    "BUILD_for_stdlib",
    "BUILD_for_rust_toolchain",
    "BUILD_for_toolchain",
)

def register_rust_toolchains(rustup_path):
    # We only care about compiling for linux, from linux.
    triple = "x86_64-unknown-linux-gnu"

    stable_workspace = "rust_stable"
    stable_toolchain = "stable"
    native.new_local_repository(
        name = stable_workspace,
        path = rustup_path + "/stable-x86_64-unknown-linux-gnu",
        build_file_content = "\n".join([
            BUILD_for_compiler(triple),
            BUILD_for_stdlib(triple),
            BUILD_for_rust_toolchain(
                name = stable_toolchain,
                workspace_name = stable_workspace,
                exec_triple = triple,
                target_triple = triple
            ),
        ])
    )

    nightly_workspace = "rust_nightly"
    nightly_toolchain = "nightly"
    native.new_local_repository(
        name = nightly_workspace,
        path = rustup_path + "/nightly-x86_64-unknown-linux-gnu",
        build_file_content = "\n".join([
            BUILD_for_compiler(triple),
            BUILD_for_stdlib(triple),
            BUILD_for_rust_toolchain(
                name = nightly_toolchain,
                workspace_name = nightly_workspace,
                exec_triple = triple,
                target_triple = triple
            )
        ])
    )

    native.new_local_repository(
        name = "rust_toolchains",
        path = rust_configurations["rust_toolchains"],
        build_file_content = "\n".join([
            BUILD_for_toolchain(stable_toolchain, stable_workspace, triple, triple),
[...]

PR #77, opened 2018-03-20
Update Rust to 1.24.1

negator:
Upgrades Rust to take advantage of newer features, namely incremental compilation:
https://github.com/rust-lang/rust/blob/master/RELEASES.md

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -124,7 +124,7 @@ rust_toolchain(
 def rust_repositories():
   native.new_http_archive(
       name = "rust_linux_x86_64",
-      url = "https://static.rust-lang.org/dist/rust-1.20.0-x86_64-unknown-linux-gnu.tar.gz",
+      url = "https://static.rust-lang.org/dist/rust-1.24.1-x86_64-unknown-linux-gnu.tar.gz",
       strip_prefix = "rust-1.24.1-x86_64-unknown-linux-gnu",
       sha256 = "4567e7f6e5e0be96e9a5a7f5149b5452828ab6a386099caca7931544f45d5327",
       build_file_content = RUST_LINUX_BUILD_FILE,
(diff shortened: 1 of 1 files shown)


acmcarther:
@buchgr: Can you take a look at this CI failure? It looks like it wasn't able to run the CI authorization process as the list of repository collaborators is privileged. If you don't have cycles, I can take a look later today

---

I'm not sure that incremental compilation will work "out of the box" in Bazel. In particular, if it does, that mean we're leaking state somewhere (this is probable), as we'll need to explicitly capture the cached state somewhere known to Bazel.

From what I can tell, this is something that needs upstream thought leadership. Swift folks seem to be having similar issues: https://github.com/bazelbuild/rules_apple/issues/21

acmcarther:
More notes:

Scala folks have an incremental compiler called "Zinc". This issue (https://github.com/bazelbuild/rules_scala/issues/328) is a starting point for their integration.

An older discussion here has a lot of details on why it doesn't work out of the box: https://groups.google.com/d/msg/bazel-discuss/3iUy5jxS3S0/1Fd_IsIEBgAJ

buchgr:
@acmcarther thanks for notifying. There was a permission error. It should be fixed now.

rust/repositories.bzl
```diff
@@ -125,16 +125,16 @@ def rust_repositories():
   native.new_http_archive(
       name = "rust_linux_x86_64",
       url = "https://static.rust-lang.org/dist/rust-1.20.0-x86_64-unknown-linux-gnu.tar.gz",
```
acmcarther: Forgot to change url (here and below)?

acmcarther:
Thanks for the speedy fix buchgr@!

rust/repositories.bzl
negator: Oof sorry about that!

acmcarther:
@negator: Belatedly, thanks for your contribution! I guess we missed three whole minor versions up until now....

I don't see any reason to delay submitting this, so I'll go ahead

negator:
Excellent! Thanks for extremely quick responses!

mfarrugi:
@negator it is also straight forward to manage versions outside of the rules, see #64 

@acmcarther We may want to make it easier to parameterize toolchain configuration, although not immediately clear how to remove the boilerplate in [this](https://github.com/mfarrugi/cargo-raze-example-stdx/blob/master/bazel/custom_rust_toolchain.bzl)

negator:
@mfarrugi Yes thank you I did see that, and that was the approach I was initially considering. However, upgrading the libs seemed very straightforward. 

Additionally I was hoping to get incremental compilation for free, however as @acmcarther points out, this is not in fact for free. A persistent worker that invokes `rustc -C...` with a `tmp` dir appears to be the correct way forward. This also appears straightforward, but will require a fair amount of boilerplate. Rules closure could perhaps be used as a reference: https://github.com/bazelbuild/rules_closure/blob/master/java/io/bazel/rules/closure/ClosureWorker.java

lilianmoraru:
Rust 1.25 is out.

PR #81, opened 2018-03-28
Update readme for 0.0.6

acmcarther:


lilianmoraru:
The docs might also need an update(these mention `0.0.5`): https://bazelbuild.github.io/rules_rust/rust/rust.html#overview

PR #82, opened 2018-04-04
Export BAZEL_GENFILES_DIR and BAZEL_BIN_DIR at compilation.

### LINKED ISSUE #79: Including a genfile at compile time is not possible

SirVer:
Fixes #79.


acmcarther:
This is really appealing as a solution, but I'm afraid of unilaterally LGTMing it because it would be very hard to undo later.

Let me see if I can get @mhlopko for a one off analysis (this should be bitesized enough that he can look at it soonish).

SirVer:
Closing this as we agree that it is not the right approach.

PR #83, opened 2018-04-12
Allow `rust_binary` to depend on `cc_library`

neachdainn:
Both the documentation and the error message say that it is possible for `rust_binary` targets to depend on `cc_library` targets. Currently, trying to do so results in an error.

I believe this one small change is all that is necessary. I have tested this with my own project and it appears to work but I am unfamiliar enough with Bazel and Skylark that I am not 100% certain that it is correct.

### DIFF
--- a/rust/rust.bzl
+++ b/rust/rust.bzl
@@ -323,7 +323,7 @@ def _rust_binary_impl(ctx):
   depinfo = _setup_deps(ctx.attr.deps,
                         ctx.label.name,
                         output_dir,
-                        allow_cc_deps=False)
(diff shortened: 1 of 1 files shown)


mfarrugi:
Looks like it was an oversight in #34, lgtm.

@acmcarther ping :)

acmcarther:
Whoops, I guess this was missed in the original review. Lucky that the error immediately indicated that this was a bug!

@neachdainn Thanks for the quick fix!
@mfarrugi Thanks for the ping!

LGTM

PR #85, opened 2018-05-18
Add FreeBSD toolchain.

dforsyth:
This is a pretty simple (mostly copy/paste) change to add support for FreeBSD.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -64,6 +64,39 @@ filegroup(
 )
 """
 
+RUST_FREEBSD_BUILD_FILE = """
+filegroup(
+    name = "rustc",
+    srcs = ["rustc/bin/rustc"],
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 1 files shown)


dforsyth:
I signed it.

mfarrugi:
Lgtm, but I'm not sure we want the default rules to include everything by default, though it probably costs nothing.

@acmcarther any thoughts?

acmcarther:
Thanks for sending out this PR dforsyth@!

---

I don't see any reason to be opposed to this, but I _would_ like start generating these definitions in the near future if possible. The list of platforms is quite extensive:
https://www.rust-lang.org/en-US/other-installers.html#standalone

As it stands this toolchain won't be part of the test suite and I'm not sure if it makes sense to try to add it to the suite, so I'll submit now. If you have any thoughts on that @dforsyth + @mfarrugi I'd love to hear them.

PR #88, opened 2018-05-27
Fix the example for test in the README

damienmg:
The method is called "greet" not "greeting"

EDIT: Added a method "greeting" and made the ""greet" method call it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,9 +604,13 @@ impl Greeter {
     pub fn new(greeting: &str) -> Greeter {
         Greeter { greeting: greeting.to_string(), }
     }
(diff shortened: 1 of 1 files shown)


acmcarther:
LGTM:

How was this caught / how to avoid in the future? Build looks green.

damienmg:
PTAL @acmcarther actually the fix was incorrect. This time the example passed `bazel test`.

Ideally testing the example in the README would be great but I don't think there is any framework for that in Bazel. IIRC @laurentlb was talking about doing so for Skylark, maybe it was done since then.

Issue #70, opened 2018-02-14
Support cargo environment variables


mfarrugi:
Cargo passes a variety of environment variables to crates at compile time. We should pass (some of?) those variables to more easily build crates.io packages.

 https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

acmcarther:
Its possible to sidestep this issue in some cases by specifying the environment variables in a bazel.rc

*//tools/bazel.rc*
```
build --action_env=CARGO_PKG_NAME=mock --action_env=CARGO_PKG_VERSION_PATCH=mock --action_env=CARGO_PKG_VERSION_MAJOR=mock --action_env=CARGO_PKG_VERSION_MINOR=mock --action_env=CARGO_PKG_VERSION=mock2 --action_env=CARGO_PKG_AUTHORS=mock2
run --action_env=CARGO_PKG_NAME=mock --action_env=CARGO_PKG_VERSION_PATCH=mock --action_env=CARGO_PKG_VERSION_MAJOR=mock --action_env=CARGO_PKG_VERSION_MINOR=mock --action_env=CARGO_PKG_VERSION=mock2 --action_env=CARGO_PKG_AUTHORS=mock2
test --action_env=CARGO_PKG_NAME=mock --action_env=CARGO_PKG_VERSION_PATCH=mock --action_env=CARGO_PKG_VERSION_MAJOR=mock --action_env=CARGO_PKG_VERSION_MINOR=mock --action_env=CARGO_PKG_VERSION=mock2 --action_env=CARGO_PKG_AUTHORS=mock2
```
If we want to actually pass useful information, we'll need to make a more substantial change to the common_attrs. Perhaps the CARGO_PKG fields could be bundled into an optional struct?

I'm really uneasy about the MANIFEST_DIR  (and of course, the CARGO) vars though.

PR #89, opened 2018-05-29
Add default cargo environment variable to the execution

damienmg:
Some existing rust code depends on those environment variables being
defined at compile-time using the env!() macro.

This inject the environment variables listed at https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates.
and make "version" an attribute of all rust_* rules.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -14,8 +14,6 @@ def _get_rustc_env(ctx):
   else:
     pre = ""
   return [
-    "CARGO=bazel",  # Bazel is actually performing the build, not cargo
-    "CARGO_MANIFEST_DIR=\"$(basename \"$PWD/%s\")\"" % ctx.build_file_path,
     "CARGO_PKG_VERSION=" + version,
     "CARGO_PKG_VERSION_MAJOR=" + v1,
     "CARGO_PKG_VERSION_MINOR=" + v2,
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -6,186 +6,225 @@ load(":utils.bzl", "relative_path")
 
 ZIP_PATH = "/usr/bin/zip"
 
+def _get_rustc_env(ctx):
+    version = ctx.attr.version if hasattr(ctx.attr, "version") else "0.0.0"
+    v1, v2, v3 = version.split(".")
+    if "-" in v3:
+        v3, pre = v3.split("-")
+    else:
+        pre = ""
+    return [
+        "CARGO=bazel",  # Bazel is actually performing the build, not cargo
+        "CARGO_MANIFEST_DIR=\"$(basename \"$PWD/%s\")\"" % ctx.build_file_path,
+        "CARGO_PKG_VERSION=" + version,
+        "CARGO_PKG_VERSION_MAJOR=" + v1,
+        "CARGO_PKG_VERSION_MINOR=" + v2,
+        "CARGO_PKG_VERSION_PATCH=" + v3,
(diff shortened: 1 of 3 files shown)


acmcarther:
Two comments

---

First, can you run your formatter in another PR (no need to change this one)? I'll LGTM that one quick and we'll be able to see the real diff here once you rebase.

---
Second, on the substance of the change:

Historically, my approach to this problem has been one of two things:
1) Send PRs to crates to turn this `env!` into an `option_env!`. Works in the small, but not in the large.
2) Instruct users (of rules_rust) to add an action_env to all rust_library invocations for this env var. Workable but clunky.

This is now a third option:
3) Expose cargo_version as a knob.

How do you imagine it being used? From my perspective, this isn't a value that is easy to sensibly tune. The effect that a given cargo_version has is not well defined, and highly specific to whatever hack a build script does, so I prefer an outcome that hides this from the user.

One way of accomplishing that would be to extract the version of cargo through the toolchain (and ultimately through the downloaded tarball). Unfortunately, I didn't find details on a cursory pass of the `1.24.0` tar. That leaves me dissatisfied, and I'm not sure how I feel about this change absent a sensible default.

Thoughts?

acmcarther:
https://github.com/bazelbuild/rules_rust/issues/70 related

acmcarther:
Oh. I misunderstood. The version setting here is the crate's own version. This doesn't require any kind of special tuning and makes sense to expose (at least, for parity with Cargo).

damienmg:
Oops sorry for the formatting, a problem in my IDE settings, the 4 space indent is incorrect and the correct formatting is the original one. Reverted.

Yes this is a first approach to #70, I need to import some package that use correct env!() to get their package version.

rust/toolchain.bzl
```diff
@@ -6,6 +6,27 @@ load(":utils.bzl", "relative_path")
...
 
+def _get_rustc_env(ctx):
+  version = ctx.attr.version if hasattr(ctx.attr, "version") else "0.0.0"
+  v1, v2, v3 = version.split(".")
+  if "-" in v3:
+    v3, pre = v3.split("-")
+  else:
+    pre = ""
+  return [
+    "CARGO=bazel",  # Bazel is actually performing the build, not cargo
```
acmcarther: I'm not sure if this is appropriate. This envvar is used occasionally inside of build scripts to invoke cargo itself. Those folks would be in for a rude awakening when trying to shell out to `cargo` and getting `bazel` instead..

rust/toolchain.bzl
```diff
@@ -6,6 +6,27 @@ load(":utils.bzl", "relative_path")
...
+    "CARGO_MANIFEST_DIR=\"$(basename \"$PWD/%s\")\"" % ctx.build_file_path,
```
acmcarther: I'm lukewarm on including this. This arg has weird semantics, and anybody trying to use it probably is going to have a bad/weird time in a bazel context. Consider that a BUILD file may not be in the same place that the corresponding Cargo.toml would be.

Are you agreeable to omitting this for now and addressing the lack of this arg when the problem presents itself?

rust/toolchain.bzl
damienmg: Since it is for build step not running the build.rs script it shouldn't matter, but since I don't need it I removed it.

rust/toolchain.bzl
damienmg: Sounds good, I just need the version for now.

damienmg:
@acmcarther: PTAL

acmcarther:
LGTM

PR #94, opened 2018-06-06
Add --color=always to rust output

damienmg:
So that colors are transmitted to Bazel output and the end-users get them
too.


mfarrugi:
If Ci is happy, I can't think of any other concerns. lgtm.

PR #90, opened 2018-05-31
Update to Rust Compiler 1.26.1

damienmg:

PR #96, opened 2018-06-11
Fix remote build

damienmg:
Tested with local docker build

This was broken by #90. @acmcarther @nlopezgi


mfarrugi:
@damienmg @buchgr is is possible to check this in the CI now or will it be in the near future?

nlopezgi:
thanks for sending this fix, I will test it with remote execution once its in and let you know if I run into issues.

nlopezgi:
Tested successfully with remote execution. afaik, it should be now possible to setup CI to test your rules with remote execution. I will sync up with @buchgr to provide more info.

xingao267:
I added the config to test rules_rust on RBE in https://github.com/bazelbuild/rules_rust/pull/111, but it seems there are a lot of tests failing on RBE due to some commits between this one and 1944c8a9c104fb73e3058a86204373df63b79eca.

Issue #86, opened 2018-05-21
Thoughts on how to add `--rustopts`


GregBowyer:
Hi all,

rightly? probably wrongly our current build configuration lets developers opt into using the sanitizers via `bazelrc` configurations.

One of these configurations abuses `--copt` to add `-fsanitize=address` to the cflags we compile under, as a result we get asan binaries.

I think I can make this happen with platforms / toolchains in the rust_rules, but it might be good to still think about how and if there is a way to specify flags for rustc in a more adhoc fashion.

Not urgent more a discussion point.

mfarrugi:
I don't quite follow, do you mean that we should support something different than (or in addition to) the `rustc_flags` parameter supported by the rules?

mhlopko:
This doc (once finished) describes our plans with the Skylark rules and bazel options: https://docs.google.com/document/d/1vc8v-kXjvgZOdQdnxPTaV0rrLxtP2XwnD2tAZlYJOqw/edit#heading=h.pn8nbhno42p5. Up until then there's not much we can do here (except hacks using --define or similar).

UebelAndre:
@hlopko do you know if there was a recent PR that solved for this? I vaguely remember something that seemed relevant. Will do some more digging later

hlopko:
I think you meant https://github.com/bazelbuild/rules_rust/pull/566?

UebelAndre:
Looks like this was resolved by https://github.com/bazelbuild/rules_rust/pull/566 closing out.

PR #97, opened 2018-06-11
Use Bazel's COMPILATION_MODE var to determine opt-level and debuginfo…

vitalyd:
… codegen options

This change looks at Bazel's `COMPILATION_MODE` var, surfaced on the CLI as `-c [dbg|fastbuild|opt]`, and adjusts the `opt-level` and `debuginfo` codegen options passed to `rustc`.  I tried to mirror what `cargo` does in its `dev` and `release` build profiles.

If there's a better way (I'm fairly new to `bazel` and `bazel rust`), I'm all ears.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -52,6 +52,10 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
   extra_filename = ""
   if output_hash:
     extra_filename = "-%s" % output_hash
+    
+  # TODO consider handling 'fastbuild'
+  # opt_level and debug_info combination roughly matches Cargo's "dev" and "release" profiles
+  (opt_level, debug_info) = (0, 2) if ctx.var["COMPILATION_MODE"] == "dbg" else (3, 0)
 
   return " ".join(
       ["set -e;"] +
@@ -68,7 +72,8 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
           src.path,
           "--crate-name %s" % crate_name,
           "--crate-type %s" % crate_type,
-          "--codegen opt-level=3",  # @TODO Might not want to do -o3 on tests
+          "--codegen opt-level=%s" % opt_level,  # @TODO Might not want to do -o3 on tests
+          "--codegen debuginfo=%s" % debug_info,
           # Disambiguate this crate from similarly named ones
           "--codegen metadata=%s" % extra_filename,
           "--codegen extra-filename='%s'" % extra_filename,
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -25,6 +25,27 @@ def _get_rustc_env(ctx):
     "CARGO_PKG_HOMEPAGE=",
   ]
 
+def _get_comp_mode_codegen_opts(ctx):
+  comp_mode = ctx.var["COMPILATION_MODE"]
+  opt_level = None
+  debug_info = None
+  if comp_mode == "dbg":
+    opt_level = 0
+    debug_info = 2
+  elif comp_mode == "opt":
+    opt_level = 3
+    debug_info = 0
+  elif comp_mode == "fastbuild":
+    opt_level = 0
+    debug_info = 0
+  else:
+    fail("Unrecognized COMPILATION_MODE: %s" % comp_mode)
+
+  return struct(
+    opt_level = opt_level,
+    debug_info = debug_info
+  )
+
 # Utility methods that use the toolchain provider.
 def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
                          depinfo, output_hash=None, rust_flags=[]):
@@ -53,9 +74,7 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
   if output_hash:
     extra_filename = "-%s" % output_hash
     
-  # TODO consider handling 'fastbuild'
-  # opt_level and debug_info combination roughly matches Cargo's "dev" and "release" profiles
-  (opt_level, debug_info) = (0, 2) if ctx.var["COMPILATION_MODE"] == "dbg" else (3, 0)
+  codegen_opts = _get_comp_mode_codegen_opts(ctx)
 
   return " ".join(
       ["set -e;"] +
@@ -72,8 +91,8 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
           src.path,
           "--crate-name %s" % crate_name,
           "--crate-type %s" % crate_type,
-          "--codegen opt-level=%s" % opt_level,  # @TODO Might not want to do -o3 on tests
-          "--codegen debuginfo=%s" % debug_info,
+          "--codegen opt-level=%s" % codegen_opts.opt_level,  # @TODO Might not want to do -o3 on tests
+          "--codegen debuginfo=%s" % codegen_opts.debug_info,
           # Disambiguate this crate from similarly named ones
           "--codegen metadata=%s" % extra_filename,
           "--codegen extra-filename='%s'" % extra_filename,
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -25,26 +25,12 @@ def _get_rustc_env(ctx):
     "CARGO_PKG_HOMEPAGE=",
   ]
 
-def _get_comp_mode_codegen_opts(ctx):
+def _get_comp_mode_codegen_opts(ctx, toolchain):
   comp_mode = ctx.var["COMPILATION_MODE"]
-  opt_level = None
-  debug_info = None
-  if comp_mode == "dbg":
-    opt_level = 0
-    debug_info = 2
-  elif comp_mode == "opt":
-    opt_level = 3
-    debug_info = 0
-  elif comp_mode == "fastbuild":
-    opt_level = 0
-    debug_info = 0
-  else:
-    fail("Unrecognized COMPILATION_MODE: %s" % comp_mode)
-
-  return struct(
-    opt_level = opt_level,
-    debug_info = debug_info
-  )
+  if not comp_mode in toolchain.compilation_mode_opts:
+    fail("Unrecognized compilation mode %s for toolchain." % comp_mode)
+  
+  return toolchain.compilation_mode_opts[comp_mode]
 
 # Utility methods that use the toolchain provider.
 def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
@@ -74,7 +60,7 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
   if output_hash:
     extra_filename = "-%s" % output_hash
     
-  codegen_opts = _get_comp_mode_codegen_opts(ctx)
+  codegen_opts = _get_comp_mode_codegen_opts(ctx, toolchain)
 
   return " ".join(
       ["set -e;"] +
@@ -91,7 +77,7 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
           src.path,
           "--crate-name %s" % crate_name,
           "--crate-type %s" % crate_type,
-          "--codegen opt-level=%s" % codegen_opts.opt_level,  # @TODO Might not want to do -o3 on tests
+          "--codegen opt-level=%s" % codegen_opts.opt_level,
           "--codegen debuginfo=%s" % codegen_opts.debug_info,
           # Disambiguate this crate from similarly named ones
           "--codegen metadata=%s" % extra_filename,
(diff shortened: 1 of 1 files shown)


mfarrugi:
lgtm (sans cla), any thoughts @acmcarther?

rust/toolchain.bzl
```diff
@@ -52,6 +52,10 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
   extra_filename = ""
   if output_hash:
     extra_filename = "-%s" % output_hash
+    
+  # TODO consider handling 'fastbuild'
+  # opt_level and debug_info combination roughly matches Cargo's "dev" and "release" profiles
+  (opt_level, debug_info) = (0, 2) if ctx.var["COMPILATION_MODE"] == "dbg" else (3, 0)
```
acmcarther: Do you have an aversion to adding support for `fastbuild` in this PR? Best way to do that is probably to pull this into a separate function, with the function returning a struct containing `opt_level` and `debug_info`. (0, 0) is probably most appropriate for `fastbuild`.

With that change, it would probably be appropriate to drop the "@TODO" from the opt-level line as well.

vitalyd: @acmcarther, the problem with `(0,0)` for `fastbuild` is that `fastbuild` is the default mode.  This means we'll get a rather useless binary/lib: no debug info and no optimizations.  This is essentially why I punted on "truly" supporting `fastbuild` - I think it requires some thought, unless you think it being the default isn't really an issue and users should explicitly opt in for `opt` builds when they want performance or `dbg` if they want full debug info.

vitalyd: For convenience, [here](https://docs.bazel.build/versions/master/user-manual.html#flag--compilation_mode) is the description of the 3 modes in bazel's user manual.  `debuginfo=0, opt-level=0` *is* in the spirit of `fastbuild`, but my main concern is a change of behavior for existing users that aren't explicitly specifying this option, which admittedly, seems quite C/C++ oriented (as its description even mentions).  That's also the reason I wasn't sure using `-c` was appropriate for non-C/C++ builds, even if they can mirror the debug info/opt level settings.

For my uses, I'm fine with explicitly specifying `-c opt` or `-c dbg`.  In `cargo` land, `fastbuild` is also somewhat close to `cargo check`, which is essentially "run everything up to LLMV codegen".  It doesn't produce an artifact, however, but *is* super quick :).  Although we likely can't use that for `fastbuild`.

acmcarther: > my main concern is a change of behavior for existing users that aren't explicitly specifying this option, which admittedly, seems quite C/C++ oriented (as its description even mentions).

To be honest, I'm personally not overly concerned about changing the default optimization level. There have been some discussions about correctly supporting -c (opt|dbg|fastbuild) elsewhere, and I think the expectation is that they should work as they do in C++. If a user wants to have a different optimization level per compilation target, they can do so via rustc_flags. By default, I do specify an optimization level in `cargo-raze`, so I expect that this will be a no-op for many build targets.

> In cargo land, fastbuild is also somewhat close to cargo check, which is essentially "run everything up to LLMV codegen". It doesn't produce an artifact, however, but is super quick :). Although we likely can't use that for fastbuild.

I don't particularly care for `cargo check` as a basis to work forward from as far as rust_library itself is concerned but if people are interested in this functionality, I think that [Bazel Aspects](https://docs.bazel.build/versions/master/skylark/aspects.html) are a good fit. They can traverse the build tree and convert rust_library invocations into those no-output-y check-like invocations

---

All that said, take those perspectives with a grain of salt. I'm the resident Cargo/"existing rust practice" antagonist though, so /cc @mfarrugi and @damienmg for more well developed opinions about changing the default optimization level.

mfarrugi:
Our build would need #86 to revert a change to the default behavior,
because we control this in a bazelrc that is building with fastbuild and
--copts.

From that perspective, I don't mind the half measure.

(No thoughts on cargo check yet)

vitalyd:
I've signed the CLA in the meantime.

rust/toolchain.bzl
damienmg: I tend to default to dbg mode in my setup. I think fastbuild is not correct but is kept because in Google, binary tend to be very big so fastbuild is a better compromise.

damienmg: Also to be consistent I would just do the fastbuild do a fast build and expect people to have -c dbg on their workstation. Also probably having fastbuild on a CI is a good thing

vitalyd:
@acmcarther, I've made the suggested changes.

vitalyd:
@acmcarther / @mfarrugi, do you think this is good to go?

mfarrugi:
Ci is unhappy and I don't think I can fix either check. @acmcarther?

acmcarther:
I just unblocked the CI and it passed, sans CLA.

@vitalyd: I'm not sure what the CLA bot uses to trigger a CLA re-evaluation, can you say `I signed it!` verbatim here?

Also check https://cla.developers.google.com/clas to double check that status.

vitalyd:
I signed it!

vitalyd:
@acmcarther, I looked at that cla link and it all seems good.  I tried a verbatim reply above - let's see if that nudges it.  The instructions above said:
[quote]
Once you've signed (or fixed any issues), please reply here (e.g. I signed it!) and we'll verify it.
[/quote]
It wasn't clear whether it had to be that text verbatim or that was just an example (and hence the "e.g.").  But I guess we'll find out soon enough :)

damienmg:
@acmcarther you can check people's CLA on signcla/
@vitalyd You do not appear neither with your github account nor with your username in the database. You might have registered with another email than the one appearing on the commit? Also don't forget to put your github account on the form.

damienmg:
Also about the content:
In C++ the flag set depending on the compilation mode is something set in the crosstool which the legacy version of the toolchain rule, this would probably make sense to have the debug level and the opt level depending on the compilation mode be a dictionary that you define through the toolchain. You can give default value as the current one so it will be in practice the same but the user would have possibility to override by specifying its own toolchain.

vitalyd:
>@vitalyd You do not appear neither with your github account nor with your username in the database. You might have registered with another email than the one appearing on the commit? Also don't forget to put your github account on the form

@damienmg, thanks for the pointer! I'm (embarrassingly) late to the github game, and I indeed only signed the agreement with my corp account.  I've now signed it with my personal account.

vitalyd:
>In C++ the flag set depending on the compilation mode is something set in the crosstool which the legacy version of the toolchain rule, this would probably make sense to have the debug level and the opt level depending on the compilation mode be a dictionary that you define through the toolchain. You can give default value as the current one so it will be in practice the same but the user would have possibility to override by specifying its own toolchain.

@damienmg, so I'm quite new to Bazel.  Are you suggesting that C++'s CROSSTOOL configuration is the model to follow or that it's "legacy" and should _not_ be followed?

I like your suggestion of making it a toolchain configuration, though - is there an example I can look at if it's not the C++ toolchain? Want to see how this is defined and then wired up at bazel build time.  Thanks

rust/toolchain.bzl
```diff
@@ -68,7 +91,8 @@ def build_rustc_command(ctx, toolchain, crate_name, crate_type, src, output_dir,
           src.path,
           "--crate-name %s" % crate_name,
           "--crate-type %s" % crate_type,
-          "--codegen opt-level=3",  # @TODO Might not want to do -o3 on tests
+          "--codegen opt-level=%s" % codegen_opts.opt_level,  # @TODO Might not want to do -o3 on tests
```
damienmg: Nit: remove todo

damienmg:
The CROSSTOOL file is the legacy version of the toolchain rules and basically toolchain where develop with the C++ CROSSTOOL example in mind. Which means that if C++ where to use the toolchain rules, they would store the configuration that depends on the compilation mode a toolchain attribute :)

I don't think such example exists, rules_go hard-code the flags in their code. Anyway that could looks like (with my PR to add the target triplet in it) :
```python
def _get_comp_mode_codegen_opts(ctx, toolchain):
  mode =  ctx.var["COMPILATION_MODE"]
  if not mode in toolchain.compilation_mode_opts:
    fail("Unrecognized compilation mode %s for toolchain for triplet %s", available modes are %s % (mode, toolchain.triplet, compilation_mode_opts.keys()))
  return toolchains.compilation_mode_opts[mode]
...
  codegen_opts = _get_comp_mode_codegen_opts(ctx, toolchain)
...
def _rust_toolchain_impl(ctx):
   compilation_mode_opts = {}
   for k, v in ctx.attr.opt_level.items():
     if not k in ctx.attr.debug_info:
       fail("Compilation mode %s is not defined in debug_info but is defined opt_level")
     compilation_mode_opts[k] = struct(opt_level = v, debug_info = ctx.attr.debug_info[k])
   for k, v in ctx.attr.debug_info.items():
     if not k in ctx.attr.opt_level:
       fail("Compilation mode %s is not defined in opt_level but is defined debug_info")
   toolchain = platform_common.ToolchainInfo(
       ...
       compilation_mode_opts = compilation_mode_opts,
       ...
       crosstool_files = ctx.files._crosstool)
  return [toolchain]

rust_toolchain = rule(
     _rust_toolchain_impl,
     attrs = {
     ...
     opt_level = attr.string_dict(default = {"opt": "3", "dbg": "0", "fastbuild": "0"}),
     debug_info =  attr.string_dict(default = {"opt": "0", "dbg": "2", "fastbuild": "0"}),
     ...
```

rust/toolchain.bzl
```diff
@@ -25,6 +25,13 @@ def _get_rustc_env(ctx):
     "CARGO_PKG_HOMEPAGE=",
   ]
 
+def _get_comp_mode_codegen_opts(ctx, toolchain):
+  comp_mode = ctx.var["COMPILATION_MODE"]
+  if not comp_mode in toolchain.compilation_mode_opts:
+    fail("Unrecognized compilation mode %s for toolchain." % comp_mode)
```
vitalyd: @damienmg, I didn't see your target triple changes in so leaving that out of the message (for now).

vitalyd:
@damienmg, I applied your suggestion of making the two settings toolchain options.

Tangentially, do you know how to query bazel to see what it would set `opt_level` and `debug_info` to without doing an actual build? I know I can run a build with `/s` to get verbose output, but I'd like to know a way to query toolchain options selected with, say, a particular `COMPILATION_MODE` without building a target.

damienmg:
I don't know any way to do that because you want the action manifest. cquery can tell you which toolchain is selected `bazel cquery -c opt 'kind(rust_toolchain, deps(@examples//fibonacci:fibonacci_test))'` but that won't tell you how the action is generate from there.

vitalyd:
>I don't know any way to do that because you want the action manifest. cquery can tell you which toolchain is selected bazel cquery -c opt 'kind(rust_toolchain, deps(@examples//fibonacci:fibonacci_test))' but that won't tell you how the action is generate from there.

That's too bad, but ok.

So, are we good on this change or is there anything else I should do? cc @acmcarther

rust/toolchain.bzl
```diff
@@ -221,6 +241,8 @@ rust_toolchain = rule(
         "_crosstool": attr.label(
             default = Label("//tools/defaults:crosstool"),
         ),
+        "opt_level": attr.string_dict(default = {"opt": "3", "dbg": "0", "fastbuild": "0"}),
```
acmcarther: Optional Nit:

I find the interchanging use of rust's opt and dbg and bazel's "opt" and "db" to be very confusing here.

Are you open to making this more verbose: "compilation_mode_to_opt_level" and "compilation_mode_to_debug_info_level"?

rust/toolchain.bzl
```diff
@@ -197,6 +207,15 @@ def _out_dir_setup_cmd(out_dir_tar):
 # The rust_toolchain rule definition and implementation.
 
 def _rust_toolchain_impl(ctx):
+  compilation_mode_opts = {}
+  for k, v in ctx.attr.opt_level.items():
```
acmcarther: Optional Nit:

I think I personally prefer just using the known variants of COMPILATION_MODE, rather than letting this be freeform. That is to say, just iterate over ["opt", "dbg", "fastbuild"], and disregard other values.

That is to say, I think that the toolchain parameters probably have too much flexibility.

rust/toolchain.bzl
vitalyd: How about naming the attributes `rust_opt_level` and `rust_debug_info`, respectively? I'm not opposed to the verbose names you suggest, but I'm also not bothered by the current scheme either :).

acmcarther: I come from a Java context, so I don't feel sufficiently spoon fed when `rust_opt_level` contains something that isn't a `string` or `num`. The map implies some kind of relationship or nuance that is hidden from the attr name.

It seems _marginally_ better to my eyes though, so if status quo is the way to go, then `rust_` SGTM.

vitalyd: IMHO, unless someone presents a slam dunk option, I’m inclined to leave this as-is and save the bikeshedding for a future iteration, perhaps when we expand the list of codegen options we support.

Otherwise, we’re iterating on names none of which seem significantly better than the others.

vitalyd:
@acmcarther / @damienmg, as mentioned, I'm inclined to push the latest set of changes in this PR.  Unless you guys have something that you insist on changing, I don't think we're making any forward progress with the name changes.  Let me know what you think.

mfarrugi:
Looks like we've been at a mergeable spot for a while, can address any remaining bikeshedding separately.

It's worth noting that these variable names are part of the public API though, and at some point they may be painful to change.

PR #98, opened 2018-06-12
Update repository rules to supported versions

### LINKED ISSUE #93: Please move from native to Skylark rules for external git/http repositories
Bazel is planning to move away from native rules for external repositories to Skylark rules, mainly because bundling every version control system in bazel does not scale. As part of this effort, we plan to submit https://bazel-review.googlesource.com/c/bazel/+/55932 soon, which will cause bazel by default to insist that the skylark versions the git and http rules for external repositories be used. This can be disabled by a flag for at least the next 6 months, but we still would appreciate if you could adopt your repository to use the Skylark versions of those rules soon.

A test run of downstream projects with that flag flip in bazel enabled can be found at https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/286

For more background, see
- the original announcement https://groups.google.com/d/topic/bazel-discuss/dO2MHQLwJF0/discussion
- and the documentation of the incompatible changes https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#remove-native-git-repository

Thanks.

dslomov:
Fixes #93.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,4 +1,6 @@
 workspace(name = "io_bazel_rules_rust")
+load("@bazel_tools//tools/build_defs/repo:git.bzl",
+     "git_repository", "new_git_repository")
 
 local_repository(
     name = "examples",
@@ -18,7 +20,7 @@ new_git_repository(
     name = "libc",
     remote = "https://github.com/rust-lang/libc",
     tag = "0.2.20",
-    build_file = "libc.BUILD",
+    build_file = "@//:libc.BUILD",
 )
 
 # Used for documenting Rust rules.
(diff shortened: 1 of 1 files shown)


mfarrugi:
Flaky build? Link gives me a 404. 

Are we still prevented from moving this into the examples workspace?

dslomov:
Build fixed, sorry about that.

> Are we still prevented from moving this into the examples workspace?

I am not quite sure what you refer to, sorry - can you explain?

mfarrugi:
Looks like it's still unhappy. I was talking about the Todo on https://github.com/bazelbuild/rules_rust/blob/master/WORKSPACE#L13 but it doesn't need to be addressed here regardless.

dslomov:
Ok 3rd time's the charm (I hope). I didn't realize the root does not have a BUILD file.

PR #99, opened 2018-06-14
Set the Rust target platform triplet using the toolchain description

damienmg:
This change add a 'triplet' information to the toolchain that is
then used to specify the target platform to Rust.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -131,6 +131,7 @@ rust_toolchain(
     staticlib_ext = ".a",
     dylib_ext = ".so",
     os = "linux",
+    triplet = "x86_64-unknown-linux-gnu",
     visibility = ["//visibility:public"],
 )
 
@@ -157,6 +158,7 @@ rust_toolchain(
     staticlib_ext = ".a",
     dylib_ext = ".dylib",
     os = "mac os x",
+    triplet = "x86_64-apple-darwin",
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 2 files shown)


acmcarther:
This LGTM with a note:

At some later stage, we're going to also want to convert that triple into a fully specified set of flags and vars to make them available for final dependency resolution. This is a stage possibly separate from locking, as described in https://github.com/bazelbuild/rules_rust/pull/100

An example of these details:
https://github.com/acmcarther/cargo-raze-examples/blob/8dba8be7d0ef1e71a4258a5d43c6808f884a7c7f/bazel/complicated_cargo_library/Cargo.bzl#L10-L71

This list was generated via shelling out to `rustc --print=cfg --target=$TARGET` and could be invoked in any manner. It shouldn't change based on the environment it is run under:
https://github.com/acmcarther/cargo-raze/blob/1a2759645ba1e69a230695d1f5d0527a9453789d/src/util.rs#L41-L50

Those details are what are required to fully resolve optional dependencies and flags.

damienmg:
@acmcarther sorry I do not follow, do you mean you want to download the dependency based on the target? The list of config in rustc is set by rustc directly and derived from the target triplet.

damienmg:
Was there any follow-up to be taken here?

acmcarther:
Sorry for the delay.

I don't have any follow up, except to ask that this PR not be submitted. I have a more specific PR that has to break this down by `exec` and `target` which should be ready in a couple of days:

https://github.com/bazelbuild/rules_rust/compare/acm-proper-platform-support#diff-90a5d7c3813ba42b7fcb679e5bb86085R293

PR #100, opened 2018-06-14
Work In Progress: Quick prototype of a cargo_crate repository rule

### LINKED ISSUE #2: Add cargo_crate repository rule
Add a `cargo_crate` [repository rule](https://docs.google.com/document/d/1jKbNXOVp2T1zJD_iRnVr8k5D0xZKgO8blMVDlXOksJg/preview) for fetching Rust crates from [Cargo.io](https://crates.io/).

damienmg:
This rule would allow cargo raze to just output a bzl file in remote mode
and then rely on the rule to generate the correct BUILD files.

An exemple of rule that cargo raze could generate would looks like:

```python
cargo_crate(
  name = "ansi_term",  # The crate name
  version = "0.11.0",  # The crate version
  # List of resolved dependency version.
  # If a dependency is needed by the library and absent from this list
  # the cargo_crate rule will emit a warning and not add the dependency in
  # the build rule, making it potentially unable to compile.
  locked_deps = {"winapi": "0.3.5"},
  # Other optional attributes:
  #  - flags: contains a list of rustc flags to add to the rust_{library,binary} rules.
  #  - data: contains the data attribute to propagate to the rust_{library,binary} rules.
  #  - sha256: SHA-256 checksum of the crate, recommended since it enhance Bazel caching.
)
```

To-do:

- [ ] Stabilize API
- [ ] Document
- [ ] Add automated tests

Fixes #2

### DIFF
--- a/rust/cargo_manifest_to_build.py
+++ b/rust/cargo_manifest_to_build.py
@@ -1,0 +1,561 @@
+#!/usr/bin/env python
+import sys
+import json
+import subprocess
+import argparse
+import StringIO
+
+parser = argparse.ArgumentParser()
+parser.add_argument('dep', nargs='*',
+                    help='List of dep_name=dep_version resolved versions')
+parser.add_argument('--dep_format', default='@io_crates_{name}__{version}//:{name}',
+                    help='How to format a dependency')
+parser.add_argument('--data', required=False,
+                    help='Content of the data file')
+parser.add_argument('--flag', action='append', default=[],
+                    help='Extra arguments for compilation')
+parser.add_argument('--output',
+                    type=argparse.FileType('w'),
+                    default=sys.stdout,
+                    help='output file')
+
+LICENSE_TYPES = {
+    "AFL-2.1": "notice",
+    "Apache-1.0": "notice",
+    "Apache-1.1": "notice",
+    "Apache-2.0": "notice",
+    "Artistic-1.0": "notice",
+    "Artistic-2.0": "notice",
+    "BSD-1-Clause": "notice",
+    "BSD-3-Clause": "notice",
+    "libtiff": "notice",
+    "BSL-1.0": "notice",
+    "CC-BY-3.0": "notice",
+    "CC-BY-4.0": "notice",
+    "ISC": "notice",
+    "LPL-1.02": "notice",
+    "Libpng": "notice",
+    "MIT": "notice",
+    "MS-PL": "notice",
+    "NCSA": "notice",
+    "OpenSSL": "notice",
+    "PHP-3.0": "notice",
+    "PHP-3.01": "notice",
+    "Python-2.0": "notice",
+    "TCP-wrappers": "notice",
+    "Unicode-DFS-2015": "notice",
+    "Unicode-DFS-2016": "notice",
+    "W3C": "notice",
+    "W3C-19980720": "notice",
+    "W3C-20150513": "notice",
+    "X11": "notice",
+    "Xnet": "notice",
+    "ZPL-2.0": "notice",
+    "ZPL-2.1": "notice",
+    "Zend-2.0": "notice",
+    "Zlib": "notice",
+    "CC0-1.0": "unencumbered",
+    "Unlicense": "unencumbered",
+    "AGPL-1.0": "restricted",
+    "AGPL-3.0": "restricted",
+    "AGPL-3.0-only": "restricted",
+    "AGPL-3.0-or-later": "restricted",
+    "WTFPL": "restricted",   
+    "Beerware": "restricted",
+    "EUPL-1.0": "restricted",
+    "EUPL-1.1": "restricted",
+    "EUPL-1.2": "restricted",
+    "SISSL": "restricted",
+    "SISSL-1.2": "restricted",
+    "CC-BY-NC-1.0": "restricted",
+    "CC-BY-NC-2.0": "restricted",
+    "CC-BY-NC-2.5": "restricted",
+    "CC-BY-NC-3.0": "restricted",
+    "CC-BY-NC-4.0": "restricted",
+    "CC-BY-NC-ND-1.0": "restricted",
+    "CC-BY-NC-ND-2.0": "restricted",
+    "CC-BY-NC-ND-2.5": "restricted",
+    "CC-BY-NC-ND-3.0": "restricted",
+    "CC-BY-NC-ND-4.0": "restricted",
+    "CC-BY-NC-SA-1.0": "restricted",
+    "CC-BY-NC-SA-2.0": "restricted",
+    "CC-BY-NC-SA-2.5": "restricted",
+    "CC-BY-NC-SA-3.0": "restricted",
+    "CC-BY-NC-SA-4.0": "restricted",
+    "OFL-1.0": "restricted",
+    "OFL-1.1": "restricted",
+    "CPL-1.0": "reciprocal",
+    "APSL-2.0": "reciprocal",
+    "CDDL-1.0": "reciprocal",
+    "CDDL-1.1": "reciprocal",
+    "EPL-1.0": "reciprocal",
+    "IPL-1.0": "reciprocal",
+    "MPL-1.0": "reciprocal",
+    "MPL-1.1": "reciprocal",
+    "MPL-2.0": "reciprocal",
+    "Ruby": "reciprocal",
+    "0BSD": "restricted",    
+    "AAL": "restricted",     
+    "ADSL": "restricted",    
+    "AFL-1.1": "restricted", 
+    "AFL-1.2": "restricted", 
+    "AFL-2.0": "restricted", 
+    "AFL-3.0": "restricted", 
+    "AMDPLPA": "restricted", 
+    "AML": "restricted",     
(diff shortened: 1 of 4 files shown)


rust/crate_repository.bzl
```diff
@@ -0,0 +1,97 @@
...
+    ]
+    if rctx.attr.data:
+        args.append("--data=" + rctx.attr.data)
+    result = rctx.execute(args)
+    if result.stderr:
+        print(result.stderr)
+
+crate_repository = repository_rule(
+    _impl,
+    attrs = {
```
acmcarther: Nit: Consider including the toolchain or target triple as well to be asserted against

`flags` and `locked_deps` are target-specific.

rust/crate_repository.bzl
```diff
@@ -0,0 +1,97 @@
...
+        "--flag=" + f
+        for f in rctx.attr.flags
+    ]
+    if rctx.attr.data:
+        args.append("--data=" + rctx.attr.data)
+    result = rctx.execute(args)
+    if result.stderr:
+        print(result.stderr)
+
+crate_repository = repository_rule(
```
acmcarther: I'm not totally clear on this -- Does this need to retain some mechanism to override a dep (e.g. instead of using some generated build target dep, accept a target directly)?

rust/cargo_manifest_to_build.py
```diff
@@ -0,0 +1,561 @@
...
+        "resolved_deps": {v.split("=", 1)[0]: v.split("=", 1)[1] for v in args.dep},
+        "output": args.output,
+        "flags": args.flag or [],
+    }
+    return ctxt
+
+if __name__ == "__main__":
+    json = json.load(StringIO.StringIO(subprocess.check_output([
+        "cargo",
+        "metadata",
```
acmcarther: This invocation strikes me as "fishy". You're largely using locally generated crate metadata, and expecting it to match what cargo-raze used to generate the exact set of dependencies.

In isolation, you cannot trust the metadata emitted by `cargo metadata` -- you need to run this for the entire workspace and then select the metadata for a single crate.

acmcarther:
Do you want to have a higher bandwidth discussion about this (GVC/Something else)?

I suspect you're stepping into a landmine here by assuming that metadata will be present that won't be, or that you can run these cargo commands independently on a per crate basis.

rust/cargo_manifest_to_build.py
```diff
@@ -0,0 +1,561 @@
...
+def extend_context(ctxt, json):
+    out_dir_tar = ""
+    if any([t["kind"] == ["custom-build"]] for t in json["targets"]):
+        out_dir_tar = "\n    out_dir_tar = \":%s_build_script_executor\"," % json["name"].replace("-", "_")
+    data = ""
+    if ctxt["data"]:
+        data = "\n    data = r\"\"\"%s\"\"\"," % ctxt["data"]
+    for d in json["dependencies"]:
+        if d["name"] not in ctxt["resolved_deps"]:
+            sys.stderr.write("WARNING: Cannot resolve dependency on crate '%s'\n" % d["name"])
```
acmcarther: This isn't  necessarily erroneous.

If cargo-raze determines that the given combination of `features` and `target` indicate that a feature should be omitted, then it will not be present in the set of resolved deps.

rust/cargo_manifest_to_build.py
```diff
@@ -0,0 +1,561 @@
...
+    return {
+        "name": json["name"],
+        "workspace_root": ctxt["workspace_root"],
+        "flags": ctxt["flags"],
+        "version": json["version"],
+        "other": "%s%s" % (out_dir_tar, data),
+        "deps": resolve_deps([d for d in json["dependencies"] if not d["kind"]], ctxt),
+        "build_deps": resolve_deps([d for d in json["dependencies"] if d["kind"] == "build"], ctxt),
+        "features": json["features"].keys(),
```
acmcarther: Experimentally, I found that crate["features"] yielded from `cargo metadata` as described below is never going to have the required features, because they cannot be identified in isolation. I think they need to be supplied by the caller (of the repository_rule) on a per-target basis with the wider crate context in mind

rust/cargo_manifest_to_build.py
damienmg: Right maybe we should add a "n/a" value that explicitely says we should skip that deps. That way people that right themselves the target will get the warning if they omit a dep?

rust/cargo_manifest_to_build.py
damienmg: This behavior was copied from cargo raze. We can definitely provide a way to override it from the attributes but it would diverge from the way cargo raze works.

rust/cargo_manifest_to_build.py
damienmg: Sorry I fail to follow here. Each crate declare its direct deps and we rely on external source to provide us the locked version, why would this invocation be wrong? I meant it barely exposes the information in the Cargo.toml file in a JSON file, it does not do any resolution.

rust/crate_repository.bzl
damienmg: No, that something I need to add to that PR.

rust/crate_repository.bzl
damienmg: Do you mean to have flags & locked_deps being selectable by platform? I don't really see where the target triple can be injected here. That's why I set the target triple in the toolchain rather than trying to inject the --target flag in the BUILD file.

Note that AFAIK there is no way to get the toolchain at this stage since it is not resolved yet when we are doing repository loading.

damienmg:
@acmcarther: getting on VC might be a bit difficult as I am in Munich, but feel free to ping me on hangout (my ldap is dmarting)

rust/crate_repository.bzl
acmcarther: I initially assumed that these would be target specific repository rules since the attrs are target-specific. Presumably multiple resolved trees of crates could avoid bringing in the same crate twice using the native.existing_rule mechanism you added in another PR. It isn't clear to me what happens with the BUILD files though...

It's possible that instead the target switching could be baked into the attrs somehow though. This provides a more intuitive interface at the cost of some additional complexity.

rust/cargo_manifest_to_build.py
acmcarther: There is a difference between what `cargo raze` does and what you're doing here. Cargo raze runs the metadata analysis once for the whole tree with all of the workspace context present. If I understand what you're doing here completely, you're running a similar process once per crate brought in.

rust/cargo_manifest_to_build.py
acmcarther: The locked crate version isn't the only piece of non-local state that needs to be brought in from external sources. You also need to bring in the set of features wanted by all upstream entities.

Consider a scenario where you have the following dependency chain:
A -(wants)-> B -> C

They use the default feature set. Now, consider a new crate D such that
A -> D -> C
... Where D adds a feature directive to it's C where ["with_serde"] is a wanted C feature.

This is what I called a "non-local" effect -- this new feature flag must also be provided to the C repository rule declaration... with the addition of the new locked serde deps.

As a final wrench to throw in -- this can happen per target if C is a ` [target.*.dependencies]` target-specific dependency of D.

acmcarther:
Discussion in https://github.com/rust-lang/cargo/pull/5122 is topical regarding some of the nuances of `depenencies` and `features` as reported by `cargo metadata`.

damienmg:
So some more though after yesterday VC:

- Features propagation is really close in cargo to the feature_flag in Bazel as it can propagate upward but it doesn't seem you can declare a feature_flag attribute in Skylark at the moment. /cc @serynth: is there a plan to expose more feature flags to Skylark?
- According to comment rust-lang/cargo#5122, they would love to move to a dependency resolution first then let the build phase handle feature related parts which would match Bazel model with the build is responsible for configuration transition and dependency resolution / fetching is a step that happens before that (except that Bazel download only a dependency if you actually needs it during the build, that's just an optimisation).

My view is that there should be minimal generated stuff to check in. At the moment with cargo-raze, there is a bunch of BUILD file whose content is, IIUC, entirely determined by the crate's Cargo.toml file. The only part that depends on reverse dependency is which feature to build (which is exactly the feature_flag) and for which we could generate X target from the cargo_crate rule.

I also find it nicer on the UX side to have most of the stuff baked into Bazel so the file cargo raze generate is still editable by hand afterward.

serynth:
Re: feature_flags:
We'd like to expose more of the feature flag functionality to Skylark, but there's no specific plan at the moment. The current goal is to get feature flags a little more stable and reduce the amount of waste they produce before opening them up further.

That said, the ability to transmit information from target to dependencies (opposite the normal target to dependers direction) is more generally provided by configuration, of which feature flags are a limited subset. AFAIK, Skylark configuration is on its way, though I have no idea offhand what the plan is for that - @juliexxia might know better. Feature flags being more exposed to Skylark might happen along a similar timeline, as configuration is even more powerful than feature flags.

damienmg:
I am going to close this one until feature flag are there, I don't think we gain much vs the current cargo raze until then.

UebelAndre:
@damienmg any updates here?

Issue #101, opened 2018-06-29
Support alternative --target values on the same toolchain


acmcarther:
Currently the stdlib and platform exts are baked into the toolchain as attrs:

https://github.com/bazelbuild/rules_rust/blob/1a05fccbd9b18430cd5ea2b0af1c79d14773870c/rust/toolchain.bzl#L222

... but this doesn't actually match how the Rust compiler works. Any given rust compiler can compile for any target platform provided that the right --target flag, stdlib, and linker are provided[1]. Keeping the current attrs as _defaults_ seems fine, but we should also support retrieving the other stdlibs if an alternative --target is specified.

https://github.com/google/cargo-raze/pull/54 has a little bit of discussion about the high level details of what a given target-triple should mean in terms of config_settings among other things.

[1] https://github.com/japaric/rust-cross#cross-compiling-with-rustc

acmcarther:
I now have a workable approach:

It's possible to use the pkg_tar rule to create a synthetic rustup-style toolchain directory as follows:

**1) Pull down all stdlibs of interest and repackage with build files similar to the following**
```python
load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar", "pkg_deb")

package(
    default_visibility = ["//visibility:public"],
)
# Sourced from
# https://static.rust-lang.org/dist/rust-std-1.26.2-armv7-unknown-linux-gnueabihf.tar.gz
#
# Results in a filetree matching the following
#
# - lib/rustlib/armv7-unknown-linux-gnueabihf/...
#
# ... matching the $TARGET_TRIPLE, of course
pkg_tar(
    name = "rust-std",
    # Generated from $VERSION-$TARGET_TRIPLE
    srcs = glob([
        "./rust-std-1.26.2-armv7-unknown-linux-gnueabihf/**"
    ]),
    # Generated from $VERSION-$TARGET_TRIPLE
    strip_prefix = "rust-std-1.26.2-armv7-unknown-linux-gnueabihf/rust-std-1.26.2-armv7-unknown-linux-gnueabihf",
)
```
**2) Pull down the compiler for the exec/host target, in a matching manner**
```python
load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar", "pkg_deb")

package(
    default_visibility = ["//visibility:public"],
)
# Sourced from
# https://static.rust-lang.org/dist/rustc-1.26.2-x86_64-unknown-linux-gnu.tar.gz
#
[...]

PR #104, opened 2018-07-12
Add utility to generate BZL containing tool shas

acmcarther:
This adds a util that can generate a BZL file that contains all known shas by downloading the `sha256` file from static.rust-lang.org.

This isn't particularly useful _yet_, but it will be necessary if we support fetching stdlibs separately from the compiler (e.g. rust-0.0.0-$TARGET versus rustc-(...) and rust-std-(...)). This also opens the door to bringing in other tools such as rustfmt and rls.

ctx: https://github.com/bazelbuild/rules_rust/issues/101#issuecomment-404624169

@mfarrugi: Are you up to reviewing this?


util/fetch_shas.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+# Enumerates the list of expected downloadable files, loads the SHAs for each file, then
+# dumps the result to stdout
+#
+# SHould be run from workspace root.
```
mfarrugi: Should :)

util/fetch_shas_VERSIONS.txt
```diff
@@ -0,0 +1,4 @@
+1.26.0
+1.26.1
+1.26.2
+1.27.0
```
mfarrugi: 1.27.1 just released

acmcarther: Added.

util/fetch_shas.sh
acmcarther: Updated

PR #105, opened 2018-07-14
Add Rust target_triple utilities, convert repositories.bzl to use, add platform(..)

acmcarther:
This PR adds some target triple manipulating utilities that will be needed for cross compilation support per https://github.com/bazelbuild/rules_rust/issues/101. In order to avoid this new code being dead code, I converted the existing repositories.bzl to use these utilities internally.

I also added platforms derived from the target triples accessible via @io_bazel_rules_rust//platforms (e.g. @io_bazel_rules_rust//platforms:x86_64-unknown-linux-gnu) which can be specified as a constraint on the command line using something like
```
bazel build @examples//hello_world:hello_world --platforms="@io_bazel_rules_rust//rust/platforms:s390x-unknown-linux-gnu"
```
... but since none of the current toolchains are cross compilation compatible, this is pointless as exec_triple will always equal target_triple...

```
INFO: Build options have changed, discarding analysis cache.
ERROR: While resolving toolchains for target @examples//hello_world:hello_world: no matching toolchains found for types @io_bazel_rules_rust//rust:toolchain
ERROR: Analysis of target '@examples//hello_world:hello_world' failed; build aborted: no matching toolchains found for types @io_bazel_rules_rust//rust:toolchain
INFO: Elapsed time: 0.271s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

No rush on review
R: @mfarrugi @damienmg

### DIFF
--- a/rust/platforms/BUILD
+++ b/rust/platforms/BUILD
@@ -1,0 +1,25 @@
+# See https://forge.rust-lang.org/platform-support.html
+# for the full set of triples
+#
+# The list of supported triples consists of triples that can be unambiguously identified using
+# built in --cpu and --platform flags.
+SUPPORTED_TRIPLES = [
+    "aarch64-unknown-linux-gnu",
+    "i686-apple-darwin",
+    "i686-pc-windows-msvc",
+    "i686-unknown-freebsd",
+    "i686-unknown-linux-gnu",
+    "powerpc-unknown-linux-gnu",
+    "s390x-unknown-linux-gnu",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-msvc",
+    "x86_64-unknown-freebsd",
+    "x86_64-unknown-linux-gnu",
+]
+
+load("@io_bazel_rules_rust//:triple_mappings.bzl", "triple_to_constraint_set")
+
+[native.platform(
+    name = target_triple,
+    constraint_values = triple_to_constraint_set(target_triple),
+) for target_triple in SUPPORTED_TRIPLES]
--- a/rust/platforms/BUILD
+++ b/rust/platforms/BUILD
@@ -17,7 +17,7 @@ SUPPORTED_TRIPLES = [
     "x86_64-unknown-linux-gnu",
 ]
 
-load("@io_bazel_rules_rust//:triple_mappings.bzl", "triple_to_constraint_set")
+load("@io_bazel_rules_rust//rust:triple_mappings.bzl", "triple_to_constraint_set")
 
 [native.platform(
     name = target_triple,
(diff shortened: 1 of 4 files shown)


mfarrugi:
Would be good for the build file template functions to be public, was doing something like https://github.com/mfarrugi/cargo-raze-example-stdx/blob/master/bazel/custom_rust_toolchain.bzl beforehand.

Can the rust_toolchain definitions move into the _generic_build_file part, or is that separation still needed?

acmcarther:
That change SGTM but it will be a short term one as the cross compilation compatible toolchains will end up separated anyway as theres going to be a toolchain declaration per combination of compiler and stdlib.

acmcarther:
Actually, correction on that comment:

So, the current separation serves to break the toolchain declaration apart from the downloading of the toolchains, but if this comment is to be believed...

https://github.com/bazelbuild/rules_rust/blob/566f08621099f200ce4ebf20045f0a106b820336/rust/repositories.bzl#L108-L111

Then it specifically serves to avoid downloading the remote repositories if the toolchain is unused. That is to say, it's possible to declare the toolchain without actually pulling it down. That won't be the case if the toolchain declaration lives inside the BUILD file for the downloaded toolchain because we won't be able to register the toolchain without downloading the whole tar.

TL;DR: I want to keep it as-is.

I'm ok with exposing the template functions though.

rust/platforms/BUILD
```diff
@@ -0,0 +1,25 @@
...
+    "s390x-unknown-linux-gnu",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-msvc",
+    "x86_64-unknown-freebsd",
+    "x86_64-unknown-linux-gnu",
+]
+
+load("@io_bazel_rules_rust//rust:triple_mappings.bzl", "triple_to_constraint_set")
+
+[native.platform(
```
damienmg: nit: remove `native.`

rust/platforms/BUILD
```diff
@@ -0,0 +1,25 @@
+# See https://forge.rust-lang.org/platform-support.html
```
damienmg: This BUILD file doesn't seemed used anywhere, remove?

acmcarther: Removed for now. The platform decls are intended to parallel @bazel_tools//platforms:* and be used as a convenience for declaring intent to cross compile, but they're not useful now.

rust/platforms/BUILD
acmcarther: Not applicable due to removal, but I'll remember this when it's added back.

acmcarther:
Making a note to:
- Combine toolchain and toolchain file decl into the new_http_repository BUILD decl per reviewers
- Add back the `platform/BUILD` once it's useful.

Issue #102, opened 2018-07-06
add a bindgen rule


mfarrugi:
It would be nice to make bindgen easily available, asking users to provide an llvm label.

@acmcarther did you mentioning having this implemented somewhere already?

acmcarther:
Surfacing bindgen would require us to do one of three things (as I see it)
1) Couple rules_rust to cargo-raze by exposing a cargo-raze workspace that contains the bindgen binary
2) Check in a generated set of BUILD files for bindgen and its deps
3) Include the bindgen genrule, but do not provide a default binding for the bindgen binary itself

Here is my genrule for it: https://github.com/acmcarther/void/blob/master/tools/rust/bindgen.bzl . This requires adding the `extra_aliased_targets = [ "cargo_bin_bindgen" ]` item to the raze section

Usage:
https://github.com/acmcarther/void/blob/37e3c7c569c963d4b6d6cb9c39f032650a14eb15/third_party/reliable_io.BUILD

What about an LLVM label?

mfarrugi:
I was thinking #3. At a glance it looks like your setup is loading the
system clang/llvm, though I haven't tested through the build.rs scripts yet.

acmcarther:
Oh, you're referring to Bindgen's clang-sys dependency? My current configuration is loading libclang at runtime I think (via libloading), but you could directly link it.

mfarrugi:
Made some progress here, chasing down all the deps to get it working sandboxed was mighty tedious!

Invocation looks like this:

```python
rust_bindgen(
    name = "bindings",
    header = ":wrapper.h",
    c_lib = "//path/to/c/package:target",
    # Boilerplate
    bindgen = "//ext/cargo:cargo_bin_bindgen",
    clang = "//ext/clang:clang",
    libllvm = "//ext/clang:lib",
    libstdcxx = "//ext/gcc:lib",
)
```
Any thoughts on trimming the boilerplate? I haven't seen any way to configure rules in the WORKSPACE for usage elsewhere, though I imagine it must exist.

I need to tidy it before sharing.

Minor tangent: overriding raze targets is fairly awkward; it might be good to support a 'handwritten_build = path/to/BUILD' for the likes of libloading, libbacktrace and whatever other degenerate crates are compiling their own c portion.

acmcarther:
[...]

PR #107, opened 2018-07-18
Add custom repository_rule for fetching toolchains

acmcarther:
This PR replaces the standard "new_http_repository" invocation that previously downloaded the "rust-*-*" package with a custom repository_rule that downloads a "rustc-*-*", then bundles it with the corresponding "rust-std-*-*" and any others that are requested.

It integrates the suggestion from https://github.com/bazelbuild/rules_rust/pull/105 to put the toolchain declaration in the remote repository but I think it isn't working correctly as I'm seeing all toolchains get downloaded. I'll review.

This PR is not sufficient alone for true cross compilation. That will require
- Support for custom linkers (propagated up to repository rule decl) as most targets need custom linkers
- Support for rust-mingw for windows targets

Please review at your leisure:
@damienmg 
@mfarrugi


rust/toolchain.bzl
```diff
@@ -247,6 +247,8 @@ rust_toolchain = rule(
         "staticlib_ext": attr.string(mandatory = True),
         "dylib_ext": attr.string(mandatory = True),
         "os": attr.string(mandatory = True),
+        "exec_triple": attr.string(mandatory = True),
+        "target_triple": attr.string(mandatory = True),
```
mfarrugi: Making these mandatory is a breaking change for anyone with a custom toolchain, may want to hold off

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+def rust_repository_set(name, version, exec_triple, extra_stdlib_triples, iso_date = None):
+    """Assembles a remote repository for the given params and yielding the 
+    names of the generated toolchains.
+
+    Args:
+      name: The name of the generated repository
+      version: The version of the tool among "nightly", "beta', or an exact version.
+      iso_date: The date of the tool (or None, if the version is a specific version).
+      exec_triple: The rust-style target that this compiler runs on
+      extra_stdlib_triples: Additional rust-style targets that this set of toolchains
```
mfarrugi: extra_target_triples maybe a little clearer, would also default this to None

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+
+    rust_toolchain_repositories(
+        name = name,
+        exec_triple = exec_triple,
+        extra_stdlib_triples = extra_stdlib_triples,
+        iso_date = iso_date,
+        version = version,
+    )
+
+    toolchain_name_tmpl = "@{name}//:toolchain_for_{triple}"
```
mfarrugi: took me a second to get tmpl == template, spell it out :)?

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+    _check_version_valid(version, iso_date, param_prefix)
+
+    tool_suburl = produce_tool_suburl(tool_name, target_triple, version, iso_date)
+    url = "https://static.rust-lang.org/dist/{}.tar.gz".format(tool_suburl)
+
+    tool_path = produce_tool_path(tool_name, target_triple, version)
+    ctx.download_and_extract(
+        url,
+        output = "",
+        sha256 = FILE_KEY_TO_SHA.get(tool_suburl) or "",
```
mfarrugi: This basically means users can't use versions rules_rust doesn't know about without patching it themselves (unless providing "" skips the check?)

Also this doesn't work well with vendored toolchains that would've been setup w/ new_local_repository.. not sure what to do here.

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
-    constraint_sets_serialized = "[{}]".format(", ".join(constraint_set_strs))
+    target_constraint_set = triple_to_constraint_set(target_triple)
+    target_constraint_set_strs = []
+    for constraint in target_constraint_set:
+        target_constraint_set_strs.append("\"{}\"".format(constraint))
+    target_constraint_sets_serialized = "[{}]".format(", ".join(target_constraint_set_strs))
 
-    return """toolchain(
+    return """
+toolchain(
```
mfarrugi: toolchain needs to be separate from rust_toolchain to avoid the eager download

rust/repositories.bzl
```diff
@@ -1,37 +1,50 @@
...
+    """Cleans a tool name for use as a bazel workspace name"""
+
+    return some_string.replace("-", "_").replace(".", "p")
+
+def BUILD_for_compiler(target_triple):
+    """Emits a BUILD file the compiler .tar.gz."""
 
     system = triple_to_system(target_triple)
-    return """filegroup(
+    return """load("@io_bazel_rules_rust//rust:toolchain.bzl", "rust_toolchain")
```
mfarrugi: nit; would start these templates w/ a blank line

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+                         $TOOL_NAME-$VERSION-$TARGET_TRIPLE.
+      version: The version of the tool among "nightly", "beta', or an exact version.
+      iso_date: The date of the tool (or None, if the version is a specific version).
+      target_triple: The rust-style target triple of the tool
+    """
+
+    _check_version_valid(version, iso_date, param_prefix)
+
+    tool_suburl = produce_tool_suburl(tool_name, target_triple, version, iso_date)
+    url = "https://static.rust-lang.org/dist/{}.tar.gz".format(tool_suburl)
```
mfarrugi: some breadcrumbs to https://static.rust-lang.org/dist/index.html may be useful

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+    """
+
+    target_triple = ctx.attr.exec_triple
+    load_arbitrary_tool(
+        ctx,
+        iso_date = ctx.attr.iso_date,
+        param_prefix = "rustc_",
+        target_triple = target_triple,
+        tool_name = "rustc",
+        tool_subdirectory = "rustc",
```
mfarrugi: So grabbing cargo and/or rustfmt from this tarball requires another function + subdirectory, but should cache to the same tarball?

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
-def _default_toolchains():
-    all_toolchains = [
-        ("rust-linux-x86_64", "x86_64-unknown-linux-gnu"),
-        ("rust-darwin-x86_64", "x86_64-apple-darwin"),
-        ("rust-freebsd-x86_64", "x86_64-unknown-freebsd"),
+def _check_version_valid(version, iso_date, param_prefix = ""):
+    """Verifies that the provided rust version and iso_date make sense."""
+
+    if not version and iso_date:
+        fail("{param_prefix}iso_date must be paired with a {param_prefix}version".format(param_prefix = param_prefix))
```
mfarrugi: skylark needs f-strings :(

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+    Args:
+      name: The name of the generated repository
+      version: The version of the tool among "nightly", "beta', or an exact version.
+      iso_date: The date of the tool (or None, if the version is a specific version).
+      exec_triple: The rust-style target that this compiler runs on
+      extra_stdlib_triples: Additional rust-style targets that this set of toolchains
+                            should support.
+    """
+
+    rust_toolchain_repositories(
```
mfarrugi: can this return the inner labels, or can we make a constant for "toolchain_for_{triple}"?

util/fetch_shas_TOOLS.txt
```diff
@@ -1 +1,3 @@
 rust
```
mfarrugi: When would 'rust' be used with this setup?

mfarrugi:
Bit of a tangent, I exposed rusftmt (for #102) and cargo (for cbindgen, though it didn't pan out) and will wait on this to assemble a PR.

rust/repositories.bzl
```diff
@@ -1,37 +1,50 @@
...
+    return some_string.replace("-", "_").replace(".", "p")
+
+def BUILD_for_compiler(target_triple):
+    """Emits a BUILD file the compiler .tar.gz."""
 
     system = triple_to_system(target_triple)
-    return """filegroup(
+    return """load("@io_bazel_rules_rust//rust:toolchain.bzl", "rust_toolchain")
+
+filegroup(
```
damienmg: opt: you could replace all those filegroup with one rust_toolchain that refer directly to the file.

rust/repositories.bzl
```diff
@@ -42,98 +55,272 @@ filegroup(
...
+
+    tool_suburl = produce_tool_suburl(tool_name, target_triple, version, iso_date)
+    url = "https://static.rust-lang.org/dist/{}.tar.gz".format(tool_suburl)
+
+    tool_path = produce_tool_path(tool_name, target_triple, version)
+    ctx.download_and_extract(
+        url,
+        output = "",
+        sha256 = FILE_KEY_TO_SHA.get(tool_suburl) or "",
+        type = "tar.gz",
```
damienmg: nit: remove this function get's the type by the extension if it matches.

rust/repositories.bzl
acmcarther: Done

rust/repositories.bzl
acmcarther: I'll hold off for now in that case... This is a change we can make in a pretty insulated way in a future PR.

rust/repositories.bzl
acmcarther: Done in docstring + above the url decl.

rust/repositories.bzl
acmcarther: Empty string does indeed skip the sha check. 

I do want to have a good story for reusing these pieces with a vendored toolchain. I'm thinking we have that already (with some assembly required), but I'm ready for more feedback. Here's what I have in mind:

Assuming a toolchain is prepared to match a rustup-style toolchain (e.g. you can use rustup dir itself)

//:toolchain.bzl
```python
def rustup_derived_toolchain(name):
  new_local_reposistory(
    name = name,
    path = "/home/acmcarther/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/"
    build_file_content = "\n".join([
      BUILD_for_compiler("x86_64-unknown-linux-gnu"),
      BUILD_for_stdlib("x86_64-unknown-linux-gnu"),
      BUILD_for_toolchain(
        workspace_name = name, 
        name = "toolchain_for_x86_64-unknown-linux-gnu", 
        exec_triple = "x86_64-unknown-linux-gnu",
        target_triple = "x86_64-unknown-linux-gnu"
      )]
    )
```

WORKSPACE
```python
load(":toolchain.bzl", "rustup_derived_toolchain")

rustup_derived_toolchain("rustup_toolchain")
```

rust/repositories.bzl
acmcarther: Done

rust/repositories.bzl
acmcarther: I see a couple of questions here:

---

"another function + subdirectory"
Yes, the followup CL will have one more of these functions per tool, but I'm planning to put those options into a separate "tool bundle" repository rule provided that an argument isn't made for including them into the compiler repository rule.

"cache to the same (...)"
This actually serves the opposite purpose, sort of. The subdirectory param names the second level dir that the contents of the "tool" lives in. (FWIW, the top level dir is just the thing i called the "tool_suburl").

The param is used for extracting the contents from the second level dir so that a composite directory similar to a `rustup` toolchain can be produced where all of the binaries live in the same bin/ dir, and all of the libraries live under lib/rustlib/ further under $TARGET_TRIPLE/ (**not** $HOST_TRIPLE, BTW). It's a minor detail, but the files here aren't ever repacked after this point -- the individual tars are all just dumped into this new remote workspace.

There is a particular motivation for doing this, by the way. My initial impl pulled the stdlibs down separately, but rustc uses paths local to the rustc executable to find the stdlib, so the final dir structure needs to match the structure produced by rustup.

Some examples of  the "second level dir":
rustfmt-(...).tar.gz contains `/rustfmt-nightly-(...)/rustfmt-preview/bin/rustfmt`

In this case, the tool_subdirectory is rustfmt-preview.

rust-std-nightly-($TARGET_TRIPLE).tar.gz contains `/rust-std-nightly-($TARGET_TRIPLE)/rust-std-($TARGET_TRIPLE)/lib/rustlib/($TARGET_TRIPLE)/*.rlib`

In this case, the tool_subdirectory is "rust-std-(...)".

rust/repositories.bzl
acmcarther: Done

rust/repositories.bzl
acmcarther: This is a repository_rule, so I don't think it can return a conventional value. We just have to have a contract between the implementation of the repository rule and the macro that wraps it.

How does an (optional) "toolchain_name_prefix" arg sound to make the connection between the distinct usages of "toolchain_for" clearer? I made the change already, LMK what you think.

rust/repositories.bzl
acmcarther: Done

rust/toolchain.bzl
acmcarther: I think these will need to be mandatory very soon (as soon as when target_triple is passed to rustc), but I can make it optional now. Done.

util/fetch_shas_TOOLS.txt
acmcarther: I don't expect it to be used at all, actually, but I hesitated to remove it from the sha generation list. I think it might be a good idea to remove it sometime after this change is done.

rust/repositories.bzl
acmcarther: This is now done (via a sibling repository rule)

I think this adds a pretty large amount of cognitive overhead for the usage, but it serves a purpose.

rust/repositories.bzl
```diff
@@ -42,98 +78,333 @@ filegroup(
...
+    for extra_target_triple in extra_target_triples:
+        all_toolchain_names.append(toolchain_name_template.format(
+            name = name,
+            triple = exec_triple,
+        ))
+
+    # Register toolchains
+    native.register_toolchains(*all_toolchain_names)
 
 # Eventually with better toolchain hosting options we could load only one of these, not both.
```
mfarrugi: Not sure what this comment was referring to before, but looks rotted now for sure

rust/repositories.bzl
```diff
@@ -42,98 +78,333 @@ filegroup(
...
+            ),
+            exec_triple = ctx.attr.exec_triple,
+            parent_workspace_name = ctx.attr.parent_workspace_name,
+            target_triple = ctx.attr.exec_triple,
+        ),
+    ]
 
-{all_toolchain_BUILDs}
-""".format(all_toolchain_BUILDs = "\n".join(all_toolchain_BUILDs))
+    for target_triple in ctx.attr.extra_target_triples:
```
mfarrugi: can dedupe

for target_triple in [ctx.attr.exec_triple] + ctx.attr.extra_target_triples

rust/repositories.bzl
```diff
@@ -42,98 +78,333 @@ filegroup(
...
+    rust_toolchain_repository_proxy(
+        name = name + "_toolchains",
+        exec_triple = exec_triple,
+        extra_target_triples = extra_target_triples,
+        parent_workspace_name = name,
+        toolchain_name_prefix = "toolchain_for",
+    )
+
+    toolchain_name_template = "@{name}_toolchains//:toolchain_for_{triple}"
+    all_toolchain_names = [toolchain_name_template.format(
```
mfarrugi: similar dedupe

rust/repositories.bzl
mfarrugi: iirc the rustup layout is different than the tar, but that bit of finagling ought to be ok

rust/repositories.bzl
```diff
@@ -42,98 +78,333 @@ filegroup(
...
+      extra_target_triples: Additional rust-style targets that this set of toolchains
+                            should support.
+    """
+
+    rust_toolchain_repository(
+        name = name,
+        exec_triple = exec_triple,
+        extra_target_triples = extra_target_triples,
+        iso_date = iso_date,
+        toolchain_name_prefix = "toolchain_for",
```
mfarrugi: these two ought to be DEFAULT_TOOLCHAIN_NAME_PREFIX

mfarrugi:
more nits, but lgtm

rust/repositories.bzl
acmcarther: I don't want to belabor the point too much, but all of the filesystem shenanigans here are built around making our usage of the raw tars match rustup _exactly_, because the compiler has expectations baked into it that files live in certain places relative to each other. This wasn't a problem for us before because using the "rust" download rather than "rustc plus rust-std" incidentally puts rustc in the right place relative to the stdlib.

For clarity, I'll add the below to the PR description as well.

---
Example comparison between Rustup and this repository_rule:
- For Bazel: Added x86_64-unknown-freebsd to `extra_target_triples`
- For Rustup: Ran `rustup target add x86_64-unknown-freebsd`

Dir structure comparison
[$tree -I "share" external/rust_linux_x86_64](https://gist.github.com/acmcarther/6e3f8b439aacb6760bea975311de2209)
[$tree -I "share" ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu"](https://gist.github.com/acmcarther/b20348cc6f93cfaf1d640045109dec0e)

PR #108, opened 2018-07-22
add rule for bindgen (#102)

mfarrugi:
#102 

This is what I got working for me, but I haven't tested it against anything else.

Main issues I can see,
1. This is not really testable within this repo, given the implicit raze dependency
2. the cc toolchain finagling seems rough at best, not sure it works w/ vanilla toolchain

I imagine I can setup a small raze environment based on the examples in this repo, but I would like to get comments on the cc_toolchain usage (and anything else) before continuing on this.


rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def _rust_bindgen(ctx):
+    rust_toolchain = ctx.toolchains["@io_bazel_rules_rust//rust:toolchain"]
+
+    bindgen = ctx.executable.bindgen
+    rustfmt = rust_toolchain.rustfmt if hasattr(rust_toolchain, 'rustfmt') else None
```
acmcarther: I don't know if rustfmt should be expected to be present in the toolchain.

I think this should be initially set up to expect the rustfmt binary to be passed as an attr. If we want to bake this into the toolchain, we should take the toolchain as an attr.

rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+
+rust_bindgen = rule(
+    _rust_bindgen,
+    attrs = {
+        "header": attr.label(allow_single_file = True),
+        "cc_lib": attr.label(),
+        "bindgen": attr.label(
+            executable = True,
+            cfg = "host",
+            default = Label("@cargo//:cargo_bin_bindgen"),
```
acmcarther: I don't think this is a good default. It isn't clear where a new user would get this binary, and "@cargo//" isn't an established convention.

rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+            command="{} --write-mode=plain {} > {}".format(rustfmt.path, unformatted.path, output.path),
+            tools=[rustfmt],
+        )
+
+rust_bindgen = rule(
+    _rust_bindgen,
+    attrs = {
+        "header": attr.label(allow_single_file = True),
+        "cc_lib": attr.label(),
+        "bindgen": attr.label(
```
acmcarther: These binaries and libraries probably should be prefixed with "_" to clearly separate them from standard parameters.

rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+        ctx.actions.run_shell(
+            inputs=depset([rustfmt, unformatted]),
+            outputs=[output],
+            command="{} --write-mode=plain {} > {}".format(rustfmt.path, unformatted.path, output.path),
+            tools=[rustfmt],
+        )
+
+rust_bindgen = rule(
+    _rust_bindgen,
+    attrs = {
```
acmcarther: Can you add a blacklist_types parameter? I've needed this parameter in every case that I used bindgen.

rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+            # Bindgen loads libclang at runtime, so we setup LD_LIBRARY_PATH
+            "LD_LIBRARY_PATH": ":".join([f.dirname for f in clang_libs]),
+            "CLANG_PATH": clang.path,
+        },
+        executable=bindgen,
+        arguments=[args],
+        tools=[bindgen, clang],
+    )
+
+    if rustfmt:
```
acmcarther: I think it's probably a good idea to separate the presence of the rustfmt binary from the decision to use rustfmt. Does it make sense to add a "should_fmt" bool?

rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+    if header not in cc_lib.cc.transitive_headers:
+        fail("Header {} is not in {}'s transitive closure of headers.".format(ctx.attr.header, cc_lib))
+
+    # rustfmt is not in the usual place, so bindgen would fail to find it
+    bindgen_args = ["--no-rustfmt-bindings"]
+    clang_args = []
+
+    output = ctx.outputs.out
+
+    clang_libs = depset(libclang.cc.libs + libstdcxx.cc.libs)
```
acmcarther: What kind of providers are these? I think the attrs need more documentation...

rust/bindgen.bzl
```diff
@@ -0,0 +1,156 @@
...
+
+    bindgen = ctx.executable.bindgen
+    rustfmt = rust_toolchain.rustfmt if hasattr(rust_toolchain, 'rustfmt') else None
+    clang = ctx.executable.clang
+    libclang = ctx.attr.libclang
+    libstdcxx = ctx.attr.libstdcxx
+    cc_toolchain = ctx.attr.cc_toolchain
+    cc_fragment = ctx.fragments.cpp
+
+    # nb. We can't grab the cc_library`s direct headers, so a header must be provided.
```
acmcarther: This usage seems a bit complicated to me. Bindgen really just generates a header file... I feel like we shouldn't require cc_lib to use bindgen.

What motivated this change (relative to https://github.com/acmcarther/void/blob/master/tools/rust/bindgen.bzl)?

rust/rust.bzl
```diff
@@ -49,6 +49,17 @@ rust_repositories()
...
 
+# Reexporting some functions as part of our public api
+
+load(":bindgen.bzl", _rust_bindgen="rust_bindgen")
+
+rust_bindgen = _rust_bindgen
+""" See :bindgen.bzl for a complete description. """
+
+
+# @TODO Move the rest of the stuff in this file elsewhere..
```
acmcarther: ????

rust/bindgen.bzl
mfarrugi: This may have made more sense when the toolchain was derived from the 'rust' component.

Providing the toolchain as an attr seems awkward since this needs the exec_platform toolchain, which afaiu should be looked up like this; I think it makes sense to still have rust's binary tooling associated with rust_toolchain, though I am not attached to it.

rust/bindgen.bzl
mfarrugi: We should establish a convention, @cargo seems like the simplest. 

Should we omit the default entirely?

rust/bindgen.bzl
mfarrugi: If including rustfmt is an optional part of the toolchain configuration, I don't see why one would go through the trouble of including it and not want it to apply to the generated code. But adding the option is fine

rust/bindgen.bzl
mfarrugi: .cc implies cc_* providers, will add and check (though again I am not sold on this being the best way to traverse a clang installation)

rust/bindgen.bzl
mfarrugi: The transitive includes can be quite sprawling when dealing with an existing codebase, and this dramatically simplifies the end-user invocation.

I agree that is is inordinately complicated, but I don't know how to better extract what we need from the cc stuff

rust/rust.bzl
mfarrugi: Copying setup of https://github.com/bazelbuild/rules_go/blob/master/go/def.bzl , we probably want to move implementations out of this file and separately move some other stuff around (eg. toolchain.bzl is not at all self explanatory now that 'toolchain' is a bazel term, and should be split into toolchain.bzl and rustc.bzl or something similar)

(regardless i will remove this todo here)

mfarrugi:
I am happy to expand on the names. My thought process was something like `bindgen = ctx.executable.bindgen` was clear, and the rule was relatively short. 

Do you think it is better to reference ctx.executable.bindgen every time or go with `bindgen_bin` etc?

mfarrugi:
On the skeptical bit, I think the cc provider and toolchain usage is weird, and even the clang usage feels like it should maybe come from a clang cc_toolchain.

rust/bindgen.bzl
damienmg: /cc @mhlopko is the owner of C++ on Bazel side :)

rust/bindgen.bzl
mfarrugi: (Just noticed this diff is missing the addition of the attribute to toolchain.bzl ...)

rust/bindgen.bzl
acmcarther: Several points:

- I'd pursue a workflow where this command was exposed as an alias with the binaries pre-populated. That means anyone working in my workspace has the tool ready to use. The current implementation can't support this approach as providing the binary would prevent callers from disabling formatting
- All would-be formatters need to know the path to the rustfmt binary. In a large repo this doesn't seem like a reasonable expectation. This applies to some of the other args here (clang/libclang) hence point 1.

rust/bindgen.bzl
acmcarther: Ignore this comment. "_" makes the argument _private_, which is decidedly not what we want to do.

I do think these args should stand out more, but I'm not sure how to accomplish that. There is a real difference between the tools that this rule needs and the inputs to the bindgen invocation in my mind.

rust/bindgen.bzl
acmcarther: I support omitting the default. 

I'm receptive to declaring a default workflow around external deps though.

mfarrugi: Doesn't _really_ work with libraries, might work with binaries as @$binary//:$binary, @cargo//:$binary, or @rust//:$binary..  Happy to punt on this, hasn't come up in other scenarios and only affects one parameter here.

rust/bindgen.bzl
acmcarther: I'm not sure if by "alias" you mean the hypothetical simplified local version of the rule or the binary itself. 

**Former:**

I'd follow my ongoing conventions and put this under //tools/rust:rust_bindgen

**Latter:**

If rustfmt comes from a rule parallel to the toolchain per https://github.com/bazelbuild/rules_rust/issues/101, i'd expect it to be brought in via a repository_rule like
```python
rust_extended_tools(
  name = "any_name",
  rustfmt_version = "0.6.1"
)
```
and would thus live under "@any_name//:rustfmt" or "@any_name//:bin/rustfmt".

If not that, then I'd expect this to be brought in by cargo. I prefer following Google's internal policies for internal deps and bring them into the working tree, so I'd expect it to be called something like "//third_party/cargo:cargo_bin_rustfmt".

mfarrugi: By 'alias' I meant the curried version of the rule, yes.

iirc the reason to avoid //third_party is because it doesn't work with recursive workspaces as well -- definitely punting on canonized naming for now.

mfarrugi:
I haven't come back to this yet because upstreaming this while requiring libstdcxx as a parameter is extremely suspect to me, though I haven't had any issues with the current incarnation.

mfarrugi:
Shipping this requires either writing a `clang_toolchain` rule, or figuring out how to use the current cc toolchain api with a clang installation appropriately.

@damienmg can you comment on whether the latter is actually a good idea?

damienmg:
I think the latter is wiser, especially since the c++ toolchain can now be written in Starlark. I'll try to talk a bit with @mhlopko a bit latter about it to see if he can help.

mfarrugi:
Updated this to use a toolchain, which makes it a little cleaner. This "Works On My Machine"..

Things to do before pushing:
- [ ] Re-address comments from before.
- [ ] Add default toolchain + working example, similar to protobuf rules
- [ ] Add/update docs

The example should involve a tarball of clang and a raze imported bindgen, which is hopefully straight forward.

rust/private/rustc.bzl
```diff
@@ -297,7 +300,7 @@ def _create_out_dir_action(ctx):
     out_dir = ctx.actions.declare_directory(ctx.label.name + ".out_dir")
     ctx.actions.run_shell(
         # TODO: Remove system tar usage
-        command = "rm -fr {dir} && mkdir {dir} && tar -xzf {tar} -C {dir}".format(tar = tar_file.path, dir = out_dir.path),
+        command = "rm -fr {dir} && mkdir {dir} && env PATH=$PATH:/bin tar -xzf {tar} -C {dir}".format(tar = tar_file.path, dir = out_dir.path),
```
mfarrugi: Looks like my version of bazel changed the default path, which is suspect. Also needed to make this patch in one of the raze genrules..

rust/bindgen.bzl
mfarrugi: This is preempted by using a toolchain without defaults instead.

rust/bindgen.bzl
mfarrugi: There are a lot of flags, are we okay with just providing `bindgen flags`?

rust/bindgen.bzl
mfarrugi: I think this is also resolved by using a `toolchain` instead.

rust/bindgen.bzl
mfarrugi: Added the `providers = ["cc"]` requirement.

mfarrugi:
The Ubuntu 16.04 (local and rbe) CI failures are due to the builders having an older version of libstdc++ than libclang wants... 

```
thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at bazel-out/host/bin/_solib_k8/_U@clang_S_S_Clibclang.so___Uexternal_Sclang_Slib/libclang.so could not be opened: bazel-out/host/bin/_solib_k8/_U@local_Ulinux_S_S_Clibstdc++___Uexternal_Slocal_Ulinux/libstdc++.so.6: version `GLIBCXX_3.4.22\' not found (required by bazel-out/host/bin/_solib_k8/_U@clang_S_S_Clibclang.so___Uexternal_Sclang_Slib/libclang.so)"', libcore/result.rs:1009:5
```

Ubuntu 14.04 failure is probably caused by using the wrong clang tarball
```
Inconsistency detected by ld.so: dl-open.c: 689: _dl_open: Assertion `_dl_debug_initialize (0, args.nsid)->r_state == RT_CONSISTENT' failed!
``` 

Will look to fix for 16.04...

mfarrugi:
Alright this looks like it works, ptal @damienmg @acmcarther 

There is something to be said about how much boilerplate is required to support a ~150 line rule.

rust/bindgen.bzl
mfarrugi: These got pulled into the toolchain.

.bazelci/presubmit.yml
```diff
@@ -1,24 +1,34 @@
...
+  ubuntu1804:
+    build_targets: *default_targets
+    test_targets:
+      - "..."
+      - "@docs//..."
+      - "@examples//..."
   macos:
     osx_targets: &osx_targets
-    - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
+    - "--"
```
damienmg: Why remove the comment?

bindgen/bindgen.bzl
```diff
@@ -0,0 +1,167 @@
+# Copyright 2015 The Bazel Authors. All rights reserved.
```
damienmg: Added file should have the year of the addition:
```suggestion
# Copyright 2019 The Bazel Authors. All rights reserved.
```

bindgen/bindgen.bzl
```diff
@@ -0,0 +1,167 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@io_bazel_rules_rust//rust:rust.bzl", "rust_library")
+
+def rust_bindgen_library(name, header, cc_lib, **kwargs):
```
damienmg: Since this library forward to a rust_library and this file control the behavior of rust_bindgen, why not explicitely list arguments for rust_bindgen and use the kwargs to forward to rust_library?

bindgen/bindgen.bzl
```diff
@@ -0,0 +1,167 @@
...
+    if header not in cc_lib.cc.transitive_headers:
+        fail("Header {} is not in {}'s transitive headers.".format(ctx.attr.header, cc_lib), "header")
+
+    toolchain = ctx.toolchains["@io_bazel_rules_rust//bindgen:bindgen_toolchain"]
+    bindgen_bin = toolchain.bindgen
+    rustfmt_bin = toolchain.rustfmt
+    clang_bin = toolchain.clang
+    libclang = toolchain.libclang
+
+    # TODO: This should not need to be explicitly provided, see below TODO.
```
damienmg: To make this more readable, I would suggest to put the whole explanation here rather in the below TODO.

bindgen/bindgen.bzl
```diff
@@ -0,0 +1,167 @@
...
+            #       would have the correct rpaths and not require this nor would this rule
+            #       have a direct dependency on libstdc++.so
+            #       This is unnecessary if the system libstdc++ suffices, which may not always be the case.
+            "LD_LIBRARY_PATH": ":".join([f.dirname for f in libstdcxx.cc.libs]),
+        },
+        arguments = [args],
+        tools = [clang_bin],
+    )
+
+    if rustfmt_bin:
```
damienmg: Is running rustfmt needed on generated sources? I meant, they are not made for being read anyway.

bindgen/bindgen.bzl
```diff
@@ -0,0 +1,167 @@
...
+        ctx.actions.run_shell(
+            inputs = depset([rustfmt_bin, unformatted_output]),
+            outputs = [output],
+            command = "{} --emit stdout --quiet {} > {}".format(rustfmt_bin.path, unformatted_output.path, output.path),
```
damienmg: If rustfmt_bin is not defined, then output.path will never be written, leading to a no output library.

bindgen/bindgen.bzl
```diff
@@ -0,0 +1,167 @@
...
+            executable = True,
+            cfg = "host",
+            mandatory = False,
+        ),
+        "clang": attr.label(
+            doc = "The label of a `clang` executable.",
+            executable = True,
+            cfg = "host",
+        ),
+        "libclang": attr.label(
```
damienmg: I am surprised you did not got hit by https://github.com/bazelbuild/bazel/issues/6889. This will get shipped in the host configuration :(

bindgen/repositories.bzl
```diff
@@ -0,0 +1,49 @@
+# Copyright 2018 The Bazel Authors. All rights reserved.
```
damienmg:
```suggestion
# Copyright 2019 The Bazel Authors. All rights reserved.
```

rust/private/rustc.bzl
damienmg: This is not correct (tar might not be in /bin, e.g. on macOS), is it because of the strict environment change we discussed during the summit?

I don't have issue running rules_rust on my Darwin and Ubuntu machines with bazel 0.21, maybe a problem with nix?

bindgen/bindgen.bzl
acmcarther: I've personally found it useful (in my own hacky version of this) when inspecting generated files. In my experience, bindgen can be a bit finnicky and require special params to configure generation.

rust/bindgen.bzl
acmcarther: bindgen_flags works for me.

rust/bindgen.bzl
acmcarther: Yeah this SGTM

WORKSPACE
```diff
@@ -40,6 +40,10 @@ load("//proto:repositories.bzl", "rust_proto_repositories")
 
 rust_proto_repositories()
 
+load("//bindgen:repositories.bzl", "rust_bindgen_repositories")
+
+rust_bindgen_repositories()
```
acmcarther: @damienmg  already mentioned this concern in his high-level comment, but I'll reiterate it here:

---

Whats the suggestion for users who want fine grained control of their cargo deps?

For reference, our suggestion for the similar proto/grpc case is [available here](https://github.com/bazelbuild/rules_rust/tree/master/proto#customizing-dependencies).

bindgen/bindgen.bzl
acmcarther: +1 on this. If we're exposing this as a "rust_library", it probably should have the same attrs, plus the bindgen ones.

bindgen/repositories.bzl
```diff
@@ -0,0 +1,49 @@
...
+
+    raze_fetch_remote_crates()
+
+    native.register_toolchains("@io_bazel_rules_rust//bindgen:example-bindgen-toolchain")
+
+def _linux_rust_bindgen_repositories():
+    # Releases @ http://releases.llvm.org/download.html
+    maybe(
+        http_archive,
+        name = "clang",
```
acmcarther: Nit: Is conditionally loading "@clang" preferable to loading our own "@bindgen-clang"?

I feel a bit weird about us trying to fetch "@clang", even if we're doing it nicely by only loading it if it hasn't yet been loaded. I'm curious what the argument is in support of occupying the name "@clang".

bindgen/bindgen.bzl
acmcarther: Not sure if @mfarrugi fully follows the issue, but I'm a bit confused.

Isn't host configuration the configuration we want to run bindgen in?

bindgen/raze/crates.bzl
```diff
@@ -0,0 +1,415 @@
...
+    )
+
+    _new_http_archive(
+        name = "raze__libloading__0_5_0",
+        url = "https://crates-io.s3-us-west-1.amazonaws.com/crates/libloading/libloading-0.5.0.crate",
+        type = "tar.gz",
+        sha256 = "9c3ad660d7cb8c5822cd83d10897b0f1f1526792737a179e73896152f85b88c2",
+        strip_prefix = "libloading-0.5.0",
+        # TODO: This is a manual patch, need https://github.com/google/cargo-raze/issues/58
+        build_file = Label("//bindgen/raze:libloading.BUILD")
```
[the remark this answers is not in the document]
acmcarther: Maybe this patch should live outside of this file, just so that it doesn't seem hidden?

That can be accomplished by declaring "raze__libloading__0_5_0" ahead of time, thanks to usage of "_new_http_archive". Best place is probably in "rust_bindgen_repositories" just before calling this.

rust/bindgen.bzl
mfarrugi: I think this is necessary to traverse the transitive headers, I don't see a way around this.

bindgen/raze/crates.bzl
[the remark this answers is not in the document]
mfarrugi: Sure I can do that. 

Small tangent.. do you have an idea of how complicated https://github.com/google/cargo-raze/issues/58 is? Would like to handle this in a way that works more closely to the `raze = "Vendor"` setup.

.bazelci/presubmit.yml
mfarrugi: I moved the comment up to the first occurence

bindgen/bindgen.bzl
mfarrugi: Can do. 

Is "The Bazel Authors" also correct for this project?

bindgen/bindgen.bzl
mfarrugi: Beyond being useful, this is what bindgen does by default. I think it is best to keep parity with other tools whenever possible.

bindgen/bindgen.bzl
mfarrugi: This is handled by Line 63. Do you think that's too tricky?

bindgen/bindgen.bzl
mfarrugi: We want to run this at build-time on the host, so I don't follow what "shipped" means. (I also don't understand the issue :))

WORKSPACE
mfarrugi: Documentation for this toolchain is basically identical, I will paste it and make the relevant edits.

bindgen/repositories.bzl
mfarrugi: I have no opinion on this, as I haven't seen other usages of clang (via tarball or otherwise). 
I can scope the name to us, but we should be compatible with any clang executable (would anyone use `@clang` for something else?

Given that the deps management should be through the toolchain instead of providing correctly named external repository, I think shadowing it to `@bindgen_clang` makes sense.

mfarrugi:
I'll look to finish this up soon. 

Any ideas on skipping the `local_linux` setup on osx? Looks like it requires a repository rule.. but maybe one of you knows of something lighter?

rust/private/rustc.bzl
mfarrugi: Grabbed latest bazel with apt and it worked.. so definitely an issue with nix-borne bazel.

mfarrugi:
Now I just need to skip loading local_linux on osx...

mfarrugi:
This probably is not far from working on osx, but I don't have a mac to test with and I don't want bounce diffs off of CI.

ptal!

bindgen/README.md
```diff
@@ -0,0 +1,82 @@
...
+</div>
+
+## Overview
+
+These rules are for using [Bindgen][bindgen] to generate [Rust][rust] bindings to C (and some C++) ilbraries.
+
+[rust]: http://www.rust-lang.org/
+[bindgen]: https://github.com/rust-lang/rust-bindgen
+
+See the [bindgen example](../examples/ffi/rust_calling_c/simple/BUILD:9) for a more complete example of use.
```
damienmg:
```suggestion
See the [bindgen example](../examples/ffi/rust_calling_c/simple/BUILD#L9) for a more complete example of use.
```

bindgen/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+To use the Rust bindgen rules, add the following to your `WORKSPACE` file to add the
+external repositories for the Rust bindgen toolchain (in addition to the [rust rules setup](..)):
+
+```python
+load("@io_bazel_rules_rust//bindgen:repositories.bzl", "rust_bindgen_repositories")
+
+rust_bindgen_repositories()
+```
+This makes the default toolchain defined in @io_bazel_rules_rust available, it is defined [here](./BUILD).
```
damienmg:
```suggestion
This makes the default toolchain defined in [`@io_bazel_rules_rust`](./BUILD) available.
```

bindgen/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+[raze]: https://github.com/google/cargo-raze
+
+It will load crate dependencies of bindgen that are generated using
+[cargo raze][raze] inside the rules_rust
+repository. However, using those dependencies might conflict with other uses
+of [cargo raze][raze]. If you need to change
+those dependencies, please see the [dedicated section below](#custom-deps).
+
+For additional information about Bazel toolchains, see [here](https://docs.bazel.build/versions/master/toolchains.html).
```
damienmg:
```suggestion
For additional information, see the [Bazel toolchains documentation](https://docs.bazel.build/versions/master/toolchains.html).
```

bindgen/bindgen.bzl
damienmg: Sounds good, can you just add a comment about it?

bindgen/bindgen.bzl
damienmg: No that's fine, I missed it, that's all :) (although I did look at it several time, I was probably tired :p).

bindgen/bindgen.bzl
damienmg: If we want to run this in host configuration please add cfg = "host". The default value of cfg is "target" but due to  bazelbuild/bazel#6889 it somehow becomes the host configuration. This is a bug in Bazel.

mfarrugi:
Looks like there are some unrelated RBE failures.. 
```
ERROR: While resolving toolchains for target @examples//fibonacci:fibonacci_test: invalid registered toolchain '@bazel_toolchains//configs/ubuntu16_04_clang/1.1/bazel_0.22.0/cpp:cc-toolchain-clang-x86_64-default': no such package '@bazel_toolchains//configs/ubuntu16_04_clang/1.1/bazel_0.22.0/cpp': BUILD file not found on package path
```

other than that I think this looks good. 

@acmcarther any other comments?

bindgen/README.md
```diff
@@ -0,0 +1,82 @@
...
+  <ul>
+    <li><a href="docs/index.md#rust_bindgen_toolchain">rust_bindgen_toolchain</a></li>
+    <li>rust_bindgen</li>
+    <li>rust_bindgen_library</li>
+  </ul>
+</div>
+
+## Overview
+
+These rules are for using [Bindgen][bindgen] to generate [Rust][rust] bindings to C (and some C++) ilbraries.
```
acmcarther: s/ilbraries/libraries

mfarrugi:
@nlopezgi where's the right place to complain about rbe ci issues? (https://buildkite.com/bazel/rules-rust-rustlang?branch=master failing after a trivial change convinces me this is not our fault)

nlopezgi:
Hi @mfarrugi 
Sorry for the super late reply on this. 
The error you were getting related to `@bazel_toolchains//configs/ubuntu16_04_clang/1.1/bazel_0.22.0/cpp:cc-toolchain-clang-x86_64-default` not being found is likely because the WORKSPACE file pointer to bazel-toolchains needs to be updated regularly (https://releases.bazel.build/bazel-toolchains.html). This is the case because with each Bazel version there are changes to the toolchain configurations that need to be used for RBE builds. Apologies for the issue. We are working towards a better mechanism (https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo.bzl) that will make it easier to keep things working with the fast pace of Bazel releases.

PR #115, opened 2018-08-09
Fix rust_doc rule with transitive native dependencies.

mfarrugi:
Not the cleanest way to do it, but I wanted to avoid a larger refactor.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -125,7 +125,7 @@ def build_rustdoc_command(ctx, toolchain, rust_doc_zip, depinfo, lib_rs, target,
         ] + doc_flags +
         depinfo.search_flags +
         # rustdoc can't do anything with native link flags, and blows up on them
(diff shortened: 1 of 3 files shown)


rust/toolchain.bzl
```diff
@@ -124,7 +124,9 @@ def build_rustdoc_command(ctx, toolchain, rust_doc_zip, depinfo, lib_rs, target,
             "-o %s" % docs_dir,
         ] + doc_flags +
         depinfo.search_flags +
-        depinfo.link_flags + [
+        # rustdoc can't do anything with native link flags, and blows up on them
```
acmcarther: Set of all flags is just link flags + externs? Or do we maybe need to preserve others as well?

acmcarther:
Hacky fix SGTM.

rust/toolchain.bzl
mfarrugi: yes, it's just those two

PR #124, opened 2018-08-28
Fix warnings in examples.

mfarrugi:
Removes a little noise from `bazel test @examples//... ...`

(Cherrpicked out of #108)


mfarrugi:
https://buildkite.com/bazel/rules-rust-rustlang/builds/171 says failing.. but I can't see it :man_shrugging:

PR #133, opened 2018-10-19
Migrate rules_rust to the new Starlark C++ toolchain API

### LINKED ISSUE #131: Migrate rust rules towards new C++ toolchain API
Rules rust use currently deprecated C++ toolchain api on `ctx.fragments.cpp`. This API will be removed in Bazel 0.20 ([incompatible change removing API from `ctx.fragments.cpp` tracking issue](https://github.com/bazelbuild/bazel/issues/6380), [incompatible change removing API from `CcToolchainInfo` tracking issue](https://github.com/bazelbuild/bazel/issues/6434)). See linked issues for migration notes.

I will temporarily disable rules_rust from https://buildkite.com/bazel/bazel-with-downstream-projects-bazel, and will reenable once this issue is fixed.

mhlopko:
This is needed for rules_rust to be forward compatible with Bazel 0.20.
Tracking issues for migration:

* https://github.com/bazelbuild/bazel/issues/6380
* https://github.com/bazelbuild/bazel/issues/6434

Fixes #131.


mhlopko:
Tests pass with Bazel 0.18, for some reason Bazel ci still tests with bazel 0.17 on windows and ubuntu.

mhlopko:
Ah, I just discovered https://github.com/bazelbuild/bazel-skylib/blob/master/lib/versions.bzl. Let me rewrite the version checking logic with that

mhlopko:
And 10 commits later, we're green. Please take a look now. Thank you!

rust/private/rustc.bzl
```diff
@@ -178,10 +189,43 @@ def rustc_compile_action(
...
+        user_link_flags = ctx.fragments.cpp.linkopts
+    else:
+        user_link_flags = depset(ctx.fragments.cpp.linkopts)
+
     # Paths to cc (for linker) and ar
     cpp_fragment = ctx.host_fragments.cpp
-    cc = cpp_fragment.compiler_executable
-    ar = cpp_fragment.ar_executable
+    cc_toolchain = find_cpp_toolchain(ctx)
+    feature_configuration = cc_common.configure_features(
```
acmcarther: Where is `cc_common` defined and how does it tie into cc_toolchain?

rust/private/rustc.bzl
```diff
@@ -178,10 +189,43 @@ def rustc_compile_action(
...
+        cc_toolchain = cc_toolchain,
+        requested_features = ctx.features,
+        unsupported_features = ctx.disabled_features,
+    )
+    ld = cc_common.get_tool_for_action(
```
acmcarther: Tangentially related to this PR:

Its not clear to me how the linker gets selected. Is the target platform the same as the one specified (indirectly) in `find_cpp_toolchain`?

rust/private/rustc.bzl
```diff
@@ -191,8 +235,6 @@ def rustc_compile_action(
     if ar_str.find("libtool", 0) != -1:
         ar = "/usr/bin/ar"
```
acmcarther: Is this old `ar` finding logic obsolete with your changes?

mfarrugi: Looks like this can be removed

```
>> rustc --version
rustc 1.26.1 (827013a31 2018-05-25)

>> rustc -C help
...
    -C                ar=val -- this option is deprecated and does nothing
```

rust/private/rustc.bzl
mhlopko: cc_common is currently builtin, the list of all of the builtins is here: https://docs.bazel.build/versions/master/skylark/lib/skylark-overview.html.

rust/private/rustc.bzl
mhlopko: C++ rules don't currently use platforms, they have their own thing controlled by `--crosstool_top`, `--cpu`, and `--compiler` (and their corresponding `host` variants, such as `--host_crosstool_top`. Here in the rule we are getting target C++ toolchain, when we're in the target configuration, and host toolchain when in host configuration (so if somebody just tells Bazel to `build //foo`, we build `foo` in target configuration. If somebody writes a genrule that has `foo` as a tool, Bazel will build it in host configuration). Did I answer your question?

rust/private/rustc.bzl
```diff
@@ -178,10 +189,43 @@ def rustc_compile_action(
...
+        action_name = CPP_LINK_EXECUTABLE_ACTION_NAME,
+    )
+    ar = cc_common.get_tool_for_action(
+        feature_configuration = feature_configuration,
+        action_name = CPP_LINK_STATIC_LIBRARY_ACTION_NAME,
+    )
+    link_variables = cc_common.create_link_variables(
+        feature_configuration = feature_configuration,
+        cc_toolchain = cc_toolchain,
+        is_linking_dynamic_library = False,
```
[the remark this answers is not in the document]
mhlopko: Ideally we would ask C++ toolchain for options twice, once for executable, once for dynamic library. In practice, they usually only differ in dynamic library having `--shared`, which rustc might be initiative enough to add on its own. But I don't think tbh.

rust/private/rustc.bzl
```diff
@@ -266,18 +307,18 @@ def _compute_rpaths(toolchain, output_dir, depinfo):
...
             toolchain.os,
             depinfo.transitive_dylibs,
         ))
 
     # Multiple dylibs can be present in the same directory, so deduplicate them.
-    return [
-        "$ORIGIN/" + relative_path(output_dir, lib_dir)
+    return depset([
+        relative_path(output_dir, lib_dir)
         for lib_dir in _get_dir_names(depinfo.transitive_dylibs)
```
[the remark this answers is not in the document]
mhlopko: Nope, it doesn't, it just takes the strings passed and puts them into templates that are configured in the CROSSTOOL file (https://docs.bazel.build/versions/master/crosstool-reference.html for docs). The C++ toolchain API is really low-level, it pretty much only deals with creating the correct command line.

mhlopko:
I'll submit now, feel free to continue discussing, I'll gladly answer or upload more fixed in a separate PR. I want to get this one in quickly so we can enable rules_rust on Bazel CI. Thanks for understanding.

rust/private/rustc.bzl
```diff
@@ -178,10 +189,43 @@ def rustc_compile_action(
     if ctx.file.out_dir_tar:
         compile_inputs.append(ctx.file.out_dir_tar)
 
+    rpaths = _compute_rpaths(toolchain, output_dir, depinfo)
+
+    if not hasattr(ctx.fragments.cpp, "linkopts"):
+        fail("Bazel is too old, please use at least Bazel 0.18.0")
+    elif versions.is_at_least("0.18.0", BAZEL_VERSION):
+        user_link_flags = ctx.fragments.cpp.linkopts
+    else:
```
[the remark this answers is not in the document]
mfarrugi: Having master support a smaller range sounds fine.

rust/private/rustc.bzl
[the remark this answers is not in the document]
mfarrugi: If we have ```rust_library(..., crate_type = "cdylib")``` should this be set to True? 
When would we be asking the toolchain for options twice?

laurentlb:
I think the commit means that all users of `rules_rust` need to add this in their WORKSPACE file:

```python
    http_archive(
        name = "bazel_skylib",
        url = "https://github.com/bazelbuild/bazel-skylib/archive/0.5.0.tar.gz",
        sha256 = "b5f6abe419da897b7901f90cbab08af958b97a8f3575b0d3dd062ac7ce78541f",
        strip_prefix = "bazel-skylib-0.5.0",
    )

    load("@io_bazel_rules_rust//:workspace.bzl", "bazel_version")
    bazel_version(name = "bazel_version")
```

PR #136, opened 2018-10-19
Set the target triplet when invoking rustc

damienmg:
The target triplet is coming from the toolchain and should be passed
along to rustc.


acmcarther:
LGTM, though for future readers: This does *not* fix cross compilation. For cross compilation to work correctly, the linker needs to support cross linking.

Issue #114, opened 2018-08-09
Support $crate_name.rs as a default crate_root?


mfarrugi:
I find myself doing this fairly often:
```
rust_binary(
    name = "some_thingy",
    crate_root = "some_thingy.rs",
    ...
)
```

Are we open to supporting this as a default value to look for? 
How do others use `crate_root`? 

This would be a straightforward change around https://github.com/bazelbuild/rules_rust/blob/master/rust/rust.bzl#L188

acmcarther:
This suggestion matches my own usage. My binaries almost always use $name.rs, unless $name.rs is already taken in which case I go with $name_bin.rs.

vitalyd:
Most code in the wild (cargo-based, albeit) that I've seen uses `main.rs` as the crate root for a binary.

acmcarther:
That's true, but that usually comes from the fact that cargo doesn't permit having multiple binaries in a project unless you use the bin/ dir (IIRC).

My personal view is that the "lib.rs" and "main.rs" conventions don't make a lot of sense in a Bazel-driven world where you can have arbitrary numbers of libraries and binaries in the same dir. This does contend somewhat with the fact that rustc does weird file inference based on declared modules though...

acmcarther:
@mfarrugi: I'm not sure if this is discussed elsewhere, but I wanted to open a line of dialogue and this seems nearest to being related.

In your own usage, have you thought about or discussed internally what the right organizational strategy should be for Rust crates in a multi-language monorepo? I'm becoming increasingly convinced that Rust code should live in its own area in the file tree primarily for the reason that Rust treats the filesystem in specific and uncontrollable ways for code structure. A person viewing a mixed Rust-C++ filetree could not tell at a glance if a directory exists for general purpose, or just because Rust needs them to define extensible modules (i.e. modules which can have submodules in separate files).

I briefly entertained the idea of embedding rust code in tar archives to avoid the mixed usage of directories, but I found that untenable due to issues having my editor treat -tar files as conceptual directories.

mfarrugi:
This wasn't given much thought. Our current structure is roughly 1 git
submodule = 1 "project", and each submodule has a directory per language
used {java, python, rust, etc ...}.

The only alternative I would be interested in is having language as one of
the top level directories, like google3, but rust doesn't benefit from this
in the way python and java do. I have done this for personal projects, but
they don't get particularly big or polyglot.

PR #137, opened 2018-10-20
Look for ${crate_name}.rs as a crate_root by default.

mfarrugi:
#114


rust/private/rust.bzl
```diff
@@ -15,15 +15,6 @@
 load(":private/rustc.bzl", "CrateInfo", "rustc_compile_action")
 load(":private/utils.bzl", "find_toolchain", "relative_path")
 
-def _find_crate_root_src(srcs, file_names = ["lib.rs"]):
-    """Finds the source file for the crate root."""
-    if len(srcs) == 1:
-        return srcs[0]
-    for src in srcs:
-        if src.basename in file_names:
```
mfarrugi: Needed to check for lib.rs/main.rs before $crate.rs, so had to refactor a little bit.

mfarrugi:
1. Yes, they're covered and I made some local changes to run through the error case.
2. I'll update the readme. Side note, there's 3 places we have documenation (inline, readme, docs/ html), and I imagine the latter 2 are mostly out of date.. what are we supposed to be doing here?

mfarrugi:
The other thing to consider is that docs/html is set up to render at https://bazelbuild.github.io/rules_rust/, the bazelbuild github pages repo appears to be private, so I can't tell how amenable that is to rendering markdown.

PR #141, opened 2018-10-21
STATIC_RUST_URL sets the static.rust-lang.org URL

damienmg:
This change add the STATIC_RUST_URL environment variable to overwrite
the URL to https://static.rust-lang.org when downloading the rust
toolchain. This let the user specify using --action_env a mirror of
static.rust-lang.org instead.

This is particularly useful in China where https://static.rust-lang.org
is really slow. For chinese user, the best way to proceed after this
change is to add the following line in your ~/.bazelrc:

```
build --action_env STATIC_RUST_URL=http://mirrors.ustc.edu.cn/rust-static
```

Note that we do not make the rule recompile on the change of the environment variable since it should not matter which way we fetch the binary, as long as we get it and the checksum match in the end.


damienmg:
Intentionally, it has some test failures I need to retest first. I'll reopen it when it will be ready for review

PR #142, opened 2018-10-22
Enable dylib support on osx

mfarrugi:
follow up from #133, want to see green on ci


mfarrugi:
And it wasn't that easy! Looks like the native deps don't make it to the linker invocation at all.

PR #144, opened 2018-10-23
Add builtin rust platform decls with examples

acmcarther:
This PR adds config_setting items under `@io_bazel_rules_rust//rust/platform` with names matching rust platform triples. This roughly matches [rules_go](https://github.com/bazelbuild/rules_go/blob/00adf001c27a5ef82c90e740e1d63509ead69ec4/go/platform/BUILD.bazel).

An open question prior to merging this is how we should handle the stdlib component of triples. Bazel doesn't seem to have much support for switching between glibc and others. Normally this wouldn't be a major issue, but without this support we can't distinguish between many of the triples (e.g. x86_64-unknown-linux-gnu && x86_64-unknown-linux-musl).

@damienmg Can you confirm that this (partially -- rules_rust changes only) matches what you were requesting in https://github.com/google/cargo-raze/pull/54#issuecomment-432297752


acmcarther:
For some reason the UI won't let me request review from damienmg@. Perhaps because he's not a collaborator? Regardless, please feel free to review as well.

rust/platform/BUILD
```diff
@@ -0,0 +1,7 @@
+package(default_visibility = ["//visibility:public"])
+
+exports_files(["triple_mappings.bzl"])
```
mfarrugi: fyi these are for skydoc which is currently broken

examples/per_platform_printer/BUILD
```diff
@@ -0,0 +1,36 @@
...
+    ],
+)
+
+rust_library(
+    name = "printer",
+    srcs = [
+        "lib.rs",
+        "print_generic.rs",
+    ] + select({
+        "@io_bazel_rules_rust//rust/platform:x86_64-unknown-linux-gnu": [
```
mfarrugi: It feels a little weird to select on the triple when we only care about the os.

Are the individual constraints exposed at all? I had raised https://github.com/bazelbuild/bazel/issues/4167 at some point..

rust/platform/platform.bzl
```diff
@@ -0,0 +1,42 @@
+load(":triple_mappings.bzl", "triple_to_constraint_set")
+
+# All T1 Platforms should be supported
+#
+# TODO: Review: How to handle stdlib variants? There doesn't appear to be a way to distinguish between them based on the constraint values available in @bazel_tools//platforms
```
mfarrugi: I can't offer anything here, perhaps there is something in the CROSSTOOL that would help us figure out if musl is setup? I would be surprised if there is a simple answer to this.

mfarrugi:
Seems fine to me, I imagine this will end up punting wrt musl.

examples/per_platform_printer/BUILD
acmcarther: Sorry, I wrote this with the context of the cargo-raze change in my head (which shall use full triples for ease of implementation). You're right, and golang also did the same thing where they generated config_settings at each level of granuarity.

https://github.com/bazelbuild/rules_go/blob/00adf001c27a5ef82c90e740e1d63509ead69ec4/go/platform/list.bzl#L86-L95

I've added them here as well now.

rust/platform/BUILD
acmcarther: I think this decl is also necessary in order for the bzl file to have public visibility. Something about how "not mentioning" a bzl file makes it not exist to the rest of the file tree. I can't recall.

mfarrugi: iirc it suffices to have an empty BUILD file

damienmg:
Indeed I am not added as a collaborator in this repo.

Anyway I think this PR is good. As for the Todo, the platform selection should select automatically the good library as declared in the crosstool file but the default crosstool does not include libmusl afaict (k believe @mhlopko have seen something around musl at some point in the past).

I also wonder if maybe for some architecture it would be more efficient to simply declare the corresponding c++ toolchain in the rules rust repo in the future, although that might be a tad too much work for now.

rust/platform/BUILD
acmcarther: Ack: Removing

rust/platform/platform.bzl
acmcarther: I've changed this from a "review" todo to a simple statement, so, i'm punting on it.

acmcarther:
@damienmg: Are you interested in being a collaborator? Your comment wasn't clear either way on the issue, but I don't think anyone would have a problem with adding you.

PR #151, opened 2018-11-01
Replace usages of `ctx.action` with `ctx.actions.{run, run_shell}`

mfarrugi:
This got fairly yak-shavy, but addresses most of the remainder of #117 

Relevant commit messages:

- Add `DepInfo` provider and update accessor syntax. 
    * Required making all imports absolute to work around https://github.com/bazelbuild/bazel/issues/3493
- split rustdoc_test into its own file
- rename setup_deps to collect_deps

PR #153, opened 2018-11-09
Test rustc flag and feature flags

damienmg:
PR #151 broke some usage of those flags that make `cargo raze` build file not work any more.

This PR adds those 2 flag to a library so we notice those breakage. It allow the change for "rustc_flags" and fix the "features" argument.

As a consequence passing `["--arg value"]` to rustc_flags no longer works but that's fine (it's a reasonable breaking change). One should either do `["--arg", "value"]` or `["--arg=value"]`.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -208,7 +208,7 @@ def rustc_compile_action(
     args.add("--emit=dep-info,link")
     args.add("--color", "always")
     args.add("--target", toolchain.target_triple)
-    args.add_all(ctx.attr.crate_features, before_each = "--cfg", format_each = "feature=%s")
+    args.add_all(ctx.attr.crate_features, before_each = "--cfg", format_each = "feature=\"%s\"")
     args.add_all(rust_flags)
     args.add_all(ctx.attr.rustc_flags)
 
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -208,7 +208,7 @@ def rustc_compile_action(
     args.add("--emit=dep-info,link")
     args.add("--color", "always")
     args.add("--target", toolchain.target_triple)
-    args.add_all(ctx.attr.crate_features, before_each = "--cfg", format_each = "feature=\"%s\"")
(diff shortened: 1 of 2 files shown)


mfarrugi:
Ah, didn't think of it as a breaking change. LGTM.

What's WIP about this?

damienmg:
For now I fixed the args for one and sent a PR on cargo-raze to fix it (it's a breaking change but one that make sense).

The crate_features is still broken, I am trying to figure out why.

damienmg:
Ok the fix is there now :)

rust/private/rustc.bzl
```diff
@@ -208,7 +208,7 @@ def rustc_compile_action(
     args.add("--emit=dep-info,link")
     args.add("--color", "always")
     args.add("--target", toolchain.target_triple)
-    args.add_all(ctx.attr.crate_features, before_each = "--cfg", format_each = "feature=%s")
+    args.add_all(ctx.attr.crate_features, before_each = "--cfg", format_each = "feature=\"%s\"")
```
mfarrugi: nit; a little easier to read as 'feature="%s"'

This quoting is necessary because features can have spaces in them?

damienmg:
Side-note: this change does not introduce the incompatible change, it was introduced by #151. It just make it explicit that there is a breaking change.

The error I had was okish but could be better: `error: Unrecognized option: 'cap-lints allow'`

The only difficulty was that it was coming from cargo-raze generated BUILD files.

acmcarther:
Yeah, I understand that the original issue came from https://github.com/bazelbuild/rules_rust/pull/151. I was mostly just wondering if this was also a good time to make the user-facing error easier to understand. It doesn't seem to me that there will be another obvious opportunity unless someone files a bug.

Ultimate decision is up to you since you're the one that had to debug the original issue, so you're most familiar with how hard that is to do with the current output.

damienmg:
If I were setting the flag myself in the BUILD file the error would have been easy to understand, I had to figure things out because the error came from generated BUILD file. 

I don't think it is worth the added complexity to output another error message (that would be probably 2-3 lines of code)

acmcarther:
Done!

damienmg:
Thanks!

PR #154, opened 2018-11-09
fix: corner case in crate dependencies

damienmg:
If a crate mod2 depends on a crate mod1, that a third crate mod3 depends
on mod2 and mod1 (an declare the extern crate in that order) then mod1
is not yet loaded when loading mod2 for compiling mod3 and since neither
are in the list of dependency directory, the compilation fails horribly.

### DIFF
--- a/test/chained_direct_deps/BUILD
+++ b/test/chained_direct_deps/BUILD
@@ -19,7 +19,7 @@ rust_library(
     name = "mod3",
     srcs = ["mod3.rs"],
     deps = [
-        "mod2",
+        ":mod2",
         ":mod1",
     ],
 )


mfarrugi:
Lgtm.

To rephrase: when an indirect dependency depends on a direct dependency, rustc can't find it.

Sorry for the trouble.

rust/private/rustc.bzl
```diff
@@ -299,7 +299,11 @@ def _get_dir_names(files):
 def add_crate_link_flags(args, dep_info):
     # nb. Crates are linked via --extern regardless of their crate_type
     args.add_all(dep_info.direct_crates, map_each = _crate_to_link_flag)
-    args.add_all(dep_info.indirect_crates, map_each = _get_crate_dirname, uniquify = True, format_each = "-Ldependency=%s")
+    args.add_all(
+        depset(transitive=[dep_info.indirect_crates, dep_info.direct_crates]),
```
mfarrugi: This can be dep_info.transitive_crates, and then indirect_crates can be removed

test/chained_direct_deps/BUILD
```diff
@@ -0,0 +1,42 @@
...
+    name = "mod2",
+    srcs = ["mod2.rs"],
+    deps = [":mod1"],
+)
+
+rust_library(
+    name = "mod3",
+    srcs = ["mod3.rs"],
+    deps = [
+        "mod2",
```
mfarrugi: nit; :mod2

test/chained_direct_deps/BUILD
```diff
@@ -0,0 +1,42 @@
+package(default_visibility = ["//visibility:public"])
```
mfarrugi: if these don't need to be public, can remove this

test/chained_direct_deps/mod3.rs
```diff
@@ -0,0 +1,26 @@
+// This crate depends on 2 crates with one of them depending on the other one.
+// If the crates are not set correctly in the dependency chain, this crate won't
+// compile. The order of the `extern crate` is important to trigger that bug.
```
mfarrugi: Do have any idea why the order matters? This is surprising

rust/private/rustc.bzl
mfarrugi:
```suggestion
        dep_info.transitive_crates,
```
(fancy github features :))

damienmg: :) I actually suggested a month ago to build a chrome extension to do a "Suggested Fix" feature and then Github released it :) (we have a similar feature inside Google)

damienmg:
And of course the CLA bot is dumb with that new feature, get the author mixed up, now CLA bot is in the limbo.

test/chained_direct_deps/BUILD
damienmg: Done. I set it public because that was also the case for the other test build file but there is really no reason for it.

test/chained_direct_deps/mod3.rs
damienmg: Yes as said in the PR explanation: putting mod1 before mod2 would load mod1 and make it visible as a dependency. Inverting the order make sure mod2 is loaded while mod1 is not. So the crate resolver looks for the list of deps path,  and fail to find mod1.

PR #122, opened 2018-08-21
Deduplicate rule impls, fix rust_doc_test (#117)

mfarrugi:
This is reasonable as is, but I will work through the doc rules as well.

Open to any and all nits, comments on naming, etc, as this is pretty subjective refactoring stuff.

Functional Changes:
- runfiles is now populated for libraries, which should be meaningless
- progress messages are a little different
- rust_doc_test is no longer a noop :)

PR #155, opened 2018-11-11
Fix rust_doc_test with transitive dependencies and add tests for it

damienmg:
Add tests that fails with:
error[E0463]: can't find crate for
 --> test/chained_direct_deps/mod3.rs:4:1
  |
4 | extern crate mod1;
  | ^^^^^^^^^^^^^^^^^^ can't find crate

And fix it.

### DIFF
--- a/rust/private/rustdoc_test.bzl
+++ b/rust/private/rustdoc_test.bzl
@@ -24,10 +24,7 @@ def _rust_doc_test_impl(ctx):
 
     toolchain = find_toolchain(ctx)
 
-    dep_info = collect_deps(
-        [ctx.attr.dep],
-        toolchain,
-    )
+    dep_info = ctx.attr.dep[DepInfo]
 
     # Construct rustdoc test command, which will be written to a shell script
     # to be executed to run the test.
@@ -67,8 +64,9 @@ def _build_rustdoc_test_script(toolchain, dep_info, crate):
     link_flags = []
     link_search_flags = []
 
+    link_flags += ["--extern " + crate.name + "=" + crate.output.short_path]
     link_flags += ["--extern " + c.name + "=" + c.output.short_path for c in d.direct_crates]
-    link_search_flags += ["-Ldependency={}".format(dirname(c.output.short_path)) for c in d.indirect_crates]
+    link_search_flags += ["-Ldependency={}".format(dirname(c.output.short_path)) for c in d.transitive_crates]
 
     link_flags += ["-ldylib=" + get_lib_name(lib) for lib in d.transitive_dylibs.to_list()]
     link_search_flags += ["-Lnative={}".format(dirname(lib.short_path)) for lib in d.transitive_dylibs]
@@ -93,7 +91,7 @@ set -e;
 rust_doc_test = rule(
     _rust_doc_test_impl,
     attrs = {
-        "dep": attr.label(mandatory = True),
+        "dep": attr.label(mandatory = True, providers=[CrateInfo]),
     },
     executable = True,
     test = True,
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -31,6 +31,7 @@ platforms:
     - "@examples//..."
     - "-//test/conflicting_deps:conflicting_deps_test"
     # rust_doc_test is likely not fully sandboxed
+    - "-//test/chained_direct_deps:mod3_doc_test"
     - "-@examples//fibonacci:fibonacci_doc_test"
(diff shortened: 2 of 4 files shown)


damienmg:
```bash
$ external/rust_linux_x86_64/bin/rustdoc --test test/chained_direct_deps/mod3.rs --crate-name mod3 -Ldependency=test/chained_direct_deps -Ldependency=test/chained_direct_deps --extern mod3=test/chained_direct_deps/libmod3-1975258125.rlib --extern mod1=test/chained_direct_deps/libmod1-1975198543.rlib  --extern mod2=test/chained_direct_deps/libmod2-1975228334.rlib

running 1 test
test test/chained_direct_deps/mod3.rs - am_i_the_world (line 19) ... FAILED

failures:

---- test/chained_direct_deps/mod3.rs - am_i_the_world (line 19) stdout ----
        error: expected one of `.`, `;`, `?`, `}`, or an operator, found `assert`
 --> test/chained_direct_deps/mod3.rs:23:1
  |
5 | #)
  |   - expected one of `.`, `;`, `?`, `}`, or an operator here
6 | assert!(
  | ^^^^^^ unexpected token

error[E0425]: cannot find function `am_i_the_world` in this scope
 --> test/chained_direct_deps/mod3.rs:21:3
  |
4 |   am_i_the_world("world") == true
  |   ^^^^^^^^^^^^^^ not found in this scope

thread 'test/chained_direct_deps/mod3.rs - am_i_the_world (line 19)' panicked at 'couldn't compile the test', librustdoc/test.rs:321:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    test/chained_direct_deps/mod3.rs - am_i_the_world (line 19)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filteredout
```

Adding the deps as extern pass this error and result in a different error

damienmg:
Those last errors could be legit, meaning that before #122 it was marking pass a test that was failing

damienmg:
Ok should be all good for merging now.

rust/private/rustdoc_test.bzl
```diff
@@ -24,10 +24,7 @@ def _rust_doc_test_impl(ctx):
 
     toolchain = find_toolchain(ctx)
 
-    dep_info = collect_deps(
```
mfarrugi: This is building a crate that has 1 direct dependency, so it's slightly different. 
Would prefer to revert this and the link_flags change, assuming that d.transitive_crates change was the actual fix.

mfarrugi:
Right, before #122 rust_doc_test was a noop.

rust/private/rustdoc_test.bzl
damienmg: IIUC this is not true, rust doc test build integration test inside the crate. Anyway, just having transitive_crates is not enough, you actually need to define --extern for all direct deps of the dep of rust_doc_test.

mfarrugi: I think that logically each doctest is built in a separate crate, hence examples containing externs
https://github.com/bazelbuild/rules_rust/blob/78fc36d22e69a5113af63d59210a28c7ea55fbb8/tools/runfiles/runfiles.rs#L16-L20
Though that doesn't mean flags et al would be the same, as you've found..

(mostly for me:)
Cargo _appears_ to do the same thing, https://github.com/rust-lang/cargo/blob/18c8c5fb9934eefb554f915bccceada54c9c29c9/src/cargo/ops/cargo_test.rs#L181
where doctest.deps comes from https://github.com/rust-lang/cargo/blob/88839c97a544b9b87f9b48ad7e96233c5f35b99d/src/cargo/core/compiler/context/mod.rs#L230 but without L67

acmcarther:
This seems to me to be the second damienmg@ CL referred to in https://github.com/bazelbuild/rules_rust/pull/92, but I can't tell if the comment on doc test dependencies is addressed.

Can you confirm damienmg@?

damienmg:
Yes this the second PR I was talking about. From what I understood, @mfarrugi was basically approving what I said that there is no better way to do it without a lot of boilerplate and that it was similar to what cargo was doing.

acmcarther:
Ah, I see. OK, thanks for the clarification.

mfarrugi:
I had approved this thinking you had the ability to merge it (and adjust
the squashed commit as you wanted).

damienmg:
@mfarrugi: no worries, we need to wait for people from Bazel team to change the permission of the repo. I pinged them on another thread.

PR #159, opened 2018-11-16
Pass -p to mkdir to make sure we do not fails if bazel already created the output directory

damienmg:
Should fix bazelbuild/continuous-integration#379


mfarrugi:
lgtm. You have merge rights now, right?

damienmg:
Yes I was waiting for the CI results with the new change :)

PR #160, opened 2018-11-17
Support passing link_flags to rust_*

damienmg:
If one needs to add extra link dependency to a library, e.g. on macOS to
link against a framework, then it needs to be included in the
`--codegen link-args=` argument. The way rustc_flag behave (passed
before the link-args) prevent the flag to be consumed by rustc. For more
those flags needs to be passed to all link action that link the crate,
hence a flag that propage to downstream dependency.

This was tested with using an example of gRPC and TLS on macOS, which
needs to link against the CoreFramework. libcore-framework-sys should be
the one carring that flag but `cargo raze` does not handle those at the
moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,16 @@ rust_library(name, srcs, crate_root, crate_type, deps, data, crate_features, rus
       </td>
     </tr>
     <tr>
+      <td><code>link_flags</code></td>
+      <td>
+        <code>List of strings, optional</code>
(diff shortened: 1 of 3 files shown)


damienmg:
I don't think this is needed anymore. I found out that `-l` rustc flag is passed down correctly, so `-lframework=CoreFoundation` is the correct rustc flag to set.

Issue #161, opened 2018-11-22
Fix Bazel incompatible changes


laurentlb:
I'd like to be able to build rules_rust with `--all_incompatible_changes`:
`cd rules_rust && bazel build --all_incompatible_changes //...`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

mhlopko:
Blocked by https://github.com/bazelbuild/rules_sass/issues/62

PR #162, opened 2018-11-22
Cleanups for compatibility with future Bazel releases

laurentlb:
Many of the changes were made using Buildifier (which is why the files
are reformatted).

Also, update some of the dependencies

Progress towards #161

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain", "relative_path")
+load("@io_bazel_rules_rust//rust:private/utils.bzl", "relative_path")
 load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
     "CPP_LINK_EXECUTABLE_ACTION_NAME",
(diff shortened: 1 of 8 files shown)


proto/raze/crates.bzl
```diff
@@ -3,26 +3,26 @@ cargo-raze crate workspace functions
 
 DO NOT EDIT! Replaced on runs of cargo-raze
```
damienmg: This is a generated file, I think there is an open issue that cargo raze is not respecting buildifier format...

Since it seems only to be formatting fix, maybe we can revert that file for now?

rust/private/rustc.bzl
```diff
@@ -115,15 +115,15 @@ def collect_deps(deps, toolchain):
 
     crate_list = transitive_crates.to_list()
     transitive_libs = depset([c.output for c in crate_list]) + transitive_staticlibs + transitive_dylibs
-    indirect_crates = depset([crate for crate in crate_list if crate not in direct_crates])
+    indirect_crates = depset([crate for crate in crate_list if crate not in direct_crates.to_list()])
```
damienmg: I just realize that we flatten depset everywhere here. Maybe we should take a deeper look at performance implication of the DepInfo. /cc @mfarrugi @acmcarther

proto/raze/crates.bzl
laurentlb: Reverted.

rust/private/rustc.bzl
laurentlb: I've added some TODOs

damienmg: Awesome, thanks!

mfarrugi: Of the [usages](https://github.com/bazelbuild/rules_rust/search?q=to_list&unscoped_q=to_list) only https://github.com/bazelbuild/rules_rust/blob/60c362762f2112b4366edc7eeea79ea3122bbae2/rust/private/rustc.bzl#L122 is easy to remove.

I think https://github.com/bazelbuild/rules_rust/blob/60c362762f2112b4366edc7eeea79ea3122bbae2/rust/private/rustdoc_test.bzl#L67-L74 could use Args, but I couldn't figure out how to turn it into a string in the template 🤷‍♂️

PR #166, opened 2018-11-29
Update dependencies, fix a few incompatible issues

laurentlb:
Tested:
  bazel-0.20.0rc4 test //... --all_incompatible_changes --incompatible_disable_deprecated_attr_params=false --incompatible_depset_is_not_iterable=false --incompatible_depset_union=false --incompatible_new_actions_api=false

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -17,6 +17,13 @@ local_repository(
     path = "docs",
 )
 
+http_archive(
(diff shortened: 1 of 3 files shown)


laurentlb:
There are still a few issues with depsets.
Other problems are inside @bazel_tools :(

Issue #140, opened 2018-10-21
Update skydoc and regenerate docs


mfarrugi:
Our `@docs//:rust-docs`  has been broken for a while by a least https://github.com/bazelbuild/skydoc/issues/89. 
Trying to update to skydoc-master produces more errors.

I suspect we should wait for https://github.com/bazelbuild/skydoc/issues/102 before trying to regenerate the docs.

Some additional docs discussion in https://github.com/bazelbuild/rules_rust/pull/137.

mfarrugi:
Another problem is that users should not depend on the docs dependencies at all, at least when we are vendoring them into the repository.

What is the appropriate way to accomplish this?

---

Ran into this trying to upgrade from `f8ad072` to `4a9d0e0 (HEAD)` 

```
>> bazel build //some/rust/library

WARNING: ignoring http_proxy in environment.
ERROR: error loading package '': Encountered error while reading extension file 'lib/versions.bzl': no such package '@bazel_skylib//lib': java.io.IOException: Error downloading [https://github.com/bazelbuild/bazel-skylib/archive/0.5.0.tar.gz] to /home/marco/.cache/bazel/_bazel_marco/4fbaddbffe00c7f1fe1df36e92b33a54/external/bazel_skylib/0.5.0.tar.gz: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
ERROR: error loading package '': Encountered error while reading extension file 'lib/versions.bzl': no such package '@bazel_skylib//lib': java.io.IOException: Error downloading [https://github.com/bazelbuild/bazel-skylib/archive/0.5.0.tar.gz] to /home/marco/.cache/bazel/_bazel_marco/4fbaddbffe00c7f1fe1df36e92b33a54/external/bazel_skylib/0.5.0.tar.gz: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
INFO: Elapsed time: 13.776s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

mfarrugi:
Looks like we should be using `doc` attr's with https://skydoc.bazel.build/docs/writing_stardoc.html

UebelAndre:
@mfarrugi Is there still outstanding work here?

PR #174, opened 2018-12-11
Get stardoc working and update docs. #140

mfarrugi:
Attribute docstrings are now more clearly the canonical documentation, removed redundant copies elsewhere. README now points to generated docs.

Also added @docs//:rust-docs to .bazelci so that it is less likely to fall into disrepair again..

Had various issues with stardoc, but if we comment out loads in `rustc.bzl` it does work and produced `index.md`.. would be nice to not need hacks :)

cc @laurentlb re stardoc


docs/BUILD
```diff
@@ -1,10 +1,24 @@
...
+    symbol_names = [
+        "rust_library",
+        "rust_binary",
+        "rust_benchmark",  # TODO(skydoc): no lint against absent symbols
+        "rust_test",
+        "rust_doc",
+        "rust_doc_test",
+        "rust_toolchain",
+        "rust_proto_toolchain",
+        "rust_proto_library",
```
damienmg: add rust_grpc_library?

proto/proto.bzl
```diff
@@ -190,6 +190,10 @@ rust_proto_library = rule(
     _rust_proto_library_impl,
     attrs = {
         "deps": attr.label_list(
+            doc = """
+                List of proto_library dependencies that will be built.
+                One crate for each proto_library will be created with the corresponding stubs.
+            """,
```
damienmg: I know that this was not document so far, but maybe we should add comment to the rust_deps attribute too (it's the rust compile dependency for the rust_proto_library target, e.g. the proto library itself).

proto/proto.bzl
```diff
@@ -241,12 +239,17 @@ rust_binary(
...
+)
 
 rust_grpc_library = rule(
     _rust_grpc_library_impl,
     attrs = {
         "deps": attr.label_list(
+            doc = """
+                List of proto_library dependencies that will be built.
+                One crate for each proto_library will be created with the corresponding gRPC stubs.
+            """,
```
damienmg: Same here for rust_deps

laurentlb:
> Had various issues with stardoc

Please file bugs for this
Thanks!

PR #178, opened 2018-12-31
Change dash to underscore in generated filenames

mjduijn:
The generated rust file for `my-file.proto` is `my_file.rs`.
Declared outputs for *_proto and *_grpc rules should reflect this.

### DIFF
--- a/proto/toolchain.bzl
+++ b/proto/toolchain.bzl
@@ -14,8 +14,9 @@
 
 """Toolchain for compiling rust stubs from protobug and gRPC."""
 
-def file_stem(f):
+def generated_file_stem(f):
     basename = f.rsplit("/", 2)[-1]
+    basename = basename.replace("-", "_")
     return basename.rsplit(".", 2)[0]
 
 def rust_generate_proto(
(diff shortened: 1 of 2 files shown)


proto/toolchain.bzl
```diff
@@ -16,7 +16,7 @@
 
 def file_stem(f):
```
mfarrugi: This function should remain a helper that takes '/some/path/file.txt' and returns 'file', can you move the `replace` to L52?

`paths = ["%s/%s" % (output_dir, file_stem(i).replace("-", "_") for i in protos.to_list()]`

mfarrugi:
How did you notice this?

proto/toolchain.bzl
mfarrugi: Yes that makes sense. (I didn't notice the other usages.)

mjduijn:
We have some files with dashes as delimiter, so when trying to generate the code, I got the following error

    ERROR: <path-to-build-file> output '<path-to-expected-output>' was not created
    ERROR: <path-to-build-file>: not all outputs were created or valid Target <path-to-target> failed to build

I've written a couple of these kinds of bazel rules for generating gRPC bindings, so I was familiar with the error.

mfarrugi:
Thanks

PR #180, opened 2018-12-31
Establish `CARGO_MANIFEST_DIR`

GregBowyer:
Several popular crates, as well as some source code has a habit of using `env!("CARGO_MANIFEST_DIR")` to find files in a crate. This causes things like google/cargo-raze#71.

Following @acmcarther suggestions this is an attempt at solution 3, which appears to handle (ab)uses of `CARGO_MANIFEST_DIR` I have tested (tera, askuma).

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -235,7 +235,30 @@ def rustc_compile_action(
         out_dir_env = "OUT_DIR=$(pwd)/{} ".format(out_dir.path)
     else:
         out_dir_env = ""
-    command = '{}{} "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd'.format(out_dir_env, toolchain.rustc.path)
+
+    # Similar awkward construction to prepend execroot to the crate-root to set `CARGO_MANIFEST_DIR`
+    #
+    # See https://github.com/google/cargo-raze/issues/71#issuecomment-433225853 for the rationale as
+    # to why
+    #
+    # In brief:
+    #
+    # Certain rust build processes expect to find files from the environment variable
+    # `$CARGO_MANIFEST_DIR`. Examples of this include pest, tera, asakuma.
+    #
+    # The compiler and by extension proc-macros see the current working directory as the Bazel exec
+    # root. Therefore, in order to fix this without an upstream code change, we have to set
+    # `$CARGO_MANIFEST_DIR`.
+    #
+    # As such we attempt to infer `$CARGO_MANIFEST_DIR`.
+    # Inference cannot be derived from `attr.crate_root`, as this points at a source file which may or
+    # may not follow the `src/lib.rs` convention. As such we use `ctx.build_file_path` mapped into the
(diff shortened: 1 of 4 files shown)


rust/private/rustc.bzl
```diff
@@ -235,7 +235,12 @@ def rustc_compile_action(
         out_dir_env = "OUT_DIR=$(pwd)/{} ".format(out_dir.path)
     else:
         out_dir_env = ""
-    command = '{}{} "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd'.format(out_dir_env, toolchain.rustc.path)
+        
+    # Similar awkward construction to prepend execroot to the crate-root to set CARGO_MANIFEST_DIR
+    package_dir = ctx.build_file_path[:ctx.build_file_path.rfind("/")]
+    manifest_dir = "CARGO_MANIFEST_DIR=$(pwd)/{} ".format(package_dir)
+    command = '{}{}{} "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd'.format(
```
mfarrugi: nit manifest_dir_env

mfarrugi:
To be clear, libraries that use this need to declare whatever additional files as data dependencies, and then everything Just Works?

rust/private/rustc.bzl
```diff
@@ -235,7 +235,12 @@ def rustc_compile_action(
         out_dir_env = "OUT_DIR=$(pwd)/{} ".format(out_dir.path)
     else:
         out_dir_env = ""
-    command = '{}{} "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd'.format(out_dir_env, toolchain.rustc.path)
+        
+    # Similar awkward construction to prepend execroot to the crate-root to set CARGO_MANIFEST_DIR
```
acmcarther: Nit: Can you cite either this PR or [my comment](https://github.com/google/cargo-raze/issues/71#issuecomment-433225853) in this code comment? I think this is abstruse enough (specifically the choice of CARGO_MANIFEST_DIR) to confuse later editors/reviewers, and having a back link to the original issue would be helpful.

rust/private/rustc.bzl
GregBowyer: I will take that nit

GregBowyer:
@mfarrugi Yes, `data` attributes will still need to be set. All this does is bind CARGO_MANIFEST_DIR to the crate root so that `env!("CARGO_MANIFEST_DIR")` finds this path.

@acmcarther rather than vendor pest, which would only work for this snapshot / point in time, would a simple rust file that uses `CARGO_MANIFEST_DIR` be suitable? If so thats easy and a test has been added.

I think for pest / tera maybe we should add a test over in cargo-raze as a smoke test too? Thoughts?

GregBowyer:
@acmcarther ... in fact if anything maybe with this we could solve google/cargo-raze#61 which would also function as a fairly solid test

mfarrugi:
I think the test added here is sufficient (not sure if it was added before or after the comment). @acmcarther  wdyt?

GregBowyer:
@mfarrugi added after for your reference.

PR #187, opened 2019-01-30
Create CODEOWNERS

dslomov:
Adding global CODEOWNERS.


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @damienmg @davidzchen
```
damienmg: The current maintainers are:

```suggestion
* @acmcarther @mfarrugi @damienmg
```

PR #188, opened 2019-02-03
Take 'default_edition' as an optional parameter of BUILD_for_rust_toolchain.

mfarrugi:
Noticed this was missing when I went to use it..


mfarrugi:
Yeah, and the argument is optional for the same reason.

For existing users, this would be a huge pain if they aren't using the
latest raze.. but I agree this would go on the list of breaking changes to
make before a 1.0 release.

PR #192, opened 2019-02-11
Pass in correct include directories to bindgen

kornholi:
Surprised it has worked this far :)! Include directories are explicitly defined and can't be derived from the filenames of headers. I ran into this with a pretty complex graph of `cc_library`s.


damienmg:
I'll let Marco and Alex who have more context than me on bindgen review.

PR #194, opened 2019-02-13
Update bazel-toolchains version

hlopko:


hlopko:
CC @meteorcloudy

meteorcloudy:
Should we also send PRs to other projects that're failing on rbe platform?
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/805

PR #195, opened 2019-02-19
Update version of io_bazel

cushon:
https://github.com/bazelbuild/bazel/issues/7463

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -7,6 +7,12 @@ load(
 )
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+git_repository(
+    name = "io_bazel",
+    commit = "62c7aa4f281e61353fdc17513f9b1bf46dd3c597",
+    remote = "https://github.com/bazelbuild/bazel.git",
+)
+
 local_repository(
     name = "examples",
     path = "examples",
@@ -26,6 +32,7 @@ http_archive(
 
 # TODO: Move this to examples/WORKSPACE when recursive repositories are enabled.
 load("@io_bazel_rules_rust//rust:repositories.bzl", "rust_repositories")
+
 rust_repositories()
 
 new_git_repository(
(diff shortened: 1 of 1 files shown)


damienmg:
> Hi Liam,
> 
> I'd like to understand a bit more where it is coming from:
> 
> ```
> $ bazel query 'somepath(@io_bazel//tools/jdk:include/jni.h, //...)'
> INFO: Invocation ID: 192a60e5-33c0-45f4-8472-3dc6c8d7712c
> DEBUG: Rule 'io_bazel_skydoc' modified arguments {"shallow_since": "1544648677 +0100"}
> DEBUG: Rule 'io_bazel_rules_sass' modified arguments {"shallow_since": "1542966119 +0100"}
> DEBUG: Rule 'io_bazel' modified arguments {"shallow_since": "1543255285 -0800"}
> INFO: Empty results
> Loading: 0 packages loaded
> ```
> io_bazel is not defined here, so there is probably something that is adding io_bazel as a repository somewhere in the transitive import.
> 
> Digging a bit into it.

```
$ bazel query 'somepath(@docs//..., @io_bazel//tools/jdk:jni_header)'
INFO: Invocation ID: 0c3917d2-dc84-4481-bb3d-1b988c9ac4fc
DEBUG: Rule 'io_bazel_skydoc' modified arguments {"shallow_since": "1544648677 +0100"}
DEBUG: Rule 'io_bazel' modified arguments {"shallow_since": "1543255285 -0800"}
DEBUG: Rule 'io_bazel_rules_sass' modified arguments {"shallow_since": "1542966119 +0100"}
@docs//:rust-docs
@io_bazel//src/main/java/com/google/devtools/build/skydoc:skydoc
@io_bazel//src/main/java/com/google/devtools/build/skydoc:skydoc_lib
@io_bazel//src/main/java/com/google/devtools/build/skydoc/fakebuildapi/cpp:cpp
@io_bazel//src/main/java/com/google/devtools/build/lib/actions:actions
@io_bazel//src/main/java/com/google/devtools/build/lib:unix
@io_bazel//src/main/native:libunix.so
@io_bazel//src/main/native:jni.h
@io_bazel//src/main/native:copy_link_jni_header
@io_bazel//tools/jdk:jni_header
Loading: 56 packages loaded
```

I got it. This needs to be fixed on skydoc side. Maybe we should just switch to stardoc right now?

damienmg:
Ok so I took a longer look and we should probably just update skydoc to a later version, https://github.com/bazelbuild/skydoc/commit/7dccadb4f2279f66ac6a8b709c0c6e8967dfd071#diff-298c08b112de7c7b76694d16435c15a2 is probably the fix on skydoc side.

damienmg:
I send #196 to update stardoc, I think this should fix the issue as stardoc is green itself.

PR #203, opened 2019-03-08
Make it possible for unit tests to have dev deps

### LINKED ISSUE #202: Cant declare inline tests with dependencies
Sorry if this is already raised

Consider a file like so

```rust
fn some_fn() -> u32 {
    2
}

#[cfg(test)]
mod tests {
    use super::some_fn();
    
    // This following line is not going to work
    use pretty_assertions::assert_eq;
    
    #[test]
    fn test_is_2() {
        assert_eq!(1, some_fn());
    }
}
```

With a build function like so

```python
rust_library(
    name = "example",
    srcs = ["lib.rs"]
)

rust_test(
    name = "example_test",
    deps = [
        ":example",
        "//third_party/rust:pretty_assertion",
    ],
)
```

You will get an error along the lines of 

```
attribute srcs: No lib.rs or example_test.rs source file found.
```

Is there something I am missing?

GregBowyer:
Due to the way unit tests are handled with the rules, it is tricky for a
unit test to have additional deps.

Consider some source code like such

```rust
fn some_fn() -> u32 {
    2
}

mod tests {
    use super::some_fn();

    // This following line is not going to work
    use pretty_assertions::assert_eq;

    #[test]
    fn test_is_2() {
        assert_eq!(1, some_fn());
    }
}
```

With a build function like so

```python
rust_library(
    name = "example",
    srcs = ["lib.rs"]
)

rust_test(
    name = "example_test",
    deps = [
        ":example",
        "//third_party/rust:pretty_assertion",
    ],
)
```

Presently, due to the handling of ‟unit tests” the extra deps will
conflict with rust_rules ideas about how to identify the test, and give
an error along the lines of

```
attribute srcs: No lib.rs or example_test.rs source file found.
```

With this change we make it possible for tests to express ‟dev
dependencies”. This lets us avoid adding deps that are strictly for
testing.

This is somewhat distinct from `testonly` attributes as it does not
always hold that a dependency is always and only ever `testonly` in all
library circumstances, consider for example the usage of
https://github.com/pingcap/fail-rs. There are instances where this needs
to be used both in source code *and* test code, and is not easily
seperated from a cargo-raze perspective.

Closes #202

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -15,6 +15,34 @@
 load("@io_bazel_rules_rust//rust:private/rustc.bzl", "CrateInfo", "rustc_compile_action")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 
+_OLD_INLINE_TEST_CRATE_MSG = """
+--------------------------------------------------------------------------------
+Testing crates with inline `#[cfg(test)]` attributes is now handled with the
+`crate` attribute in `rust_test`:
+
+This means you should migrate from:
+
+```
+rust_test(
+    name = "{name}",
+    deps = ["{dep}"],
+    ...
+)
+```
+
+to the following:
+```
+rust_test(
+    name = "{name}",
+    crate = "{dep}",
+    ...
+```
+
+See https://github.com/bazelbuild/rules_rust/pull/203 for more details
+--------------------------------------------------------------------------------
+
+"""
+
 # TODO(marco): Separate each rule into its own file.
 
 def _determine_output_hash(lib_rs):
@@ -124,19 +152,26 @@ def _rust_test_common(ctx, test_binary):
     """
     toolchain = find_toolchain(ctx)
 
-    if len(ctx.attr.deps) == 1 and len(ctx.files.srcs) == 0:
-        # Target has a single dependency but no srcs. Build the test binary using
-        # the dependency's srcs.
-        parent_crate = ctx.attr.deps[0][CrateInfo]
+    if ctx.attr.crate:
+        # Target is building the crate in `test` config
+        # Build the test binary using the dependency's srcs.
+        crate = ctx.attr.crate[CrateInfo]
         target = CrateInfo(
             name = test_binary.basename,
-            type = parent_crate.type,
-            root = parent_crate.root,
-            srcs = parent_crate.srcs,
-            deps = parent_crate.deps,
+            type = crate.type,
+            root = crate.root,
+            srcs = crate.srcs + ctx.files.srcs,
+            deps = crate.deps + ctx.attr.deps,
             output = test_binary,
-            edition = parent_crate.edition,
+            edition = crate.edition,
         )
+    elif len(ctx.attr.deps) == 1 and len(ctx.files.srcs) == 0:
+        dep = ctx.attr.deps[0].label
+        msg = _OLD_INLINE_TEST_CRATE_MSG.format(
+                name=test_binary.basename,
+                dep=dep if ctx.label.package != dep.package else ":" + dep.name
+              )
+        fail(msg)
     else:
         # Target is a standalone crate. Build the test binary as its own crate.
         target = CrateInfo(
@@ -278,6 +313,18 @@ _rust_library_attrs = {
             The exact output file will depend on the toolchain used.
         """),
         default = "rlib",
+    ),
+}
+
+_rust_test_attrs = {
+    "crate": attr.label(
+        mandatory = False,
+        doc = _tidy("""
+            Target inline tests declared in the given crate
+
+            These tests are typically those that would be held out under
+            `#[cfg(test)]` declarations.
+        """),
     ),
 }
 
@@ -459,7 +506,8 @@ Hello world
 
 rust_test = rule(
     _rust_test_impl,
-    attrs = _rust_common_attrs,
+    attrs = dict(_rust_common_attrs.items() +
+                 _rust_test_attrs.items()),
     executable = True,
     fragments = ["cpp"],
     host_fragments = ["cpp"],
@@ -531,6 +579,18 @@ rust_test(
 ```
 
 Run the test with `bazel build //hello_lib:hello_lib_test`.
+
+To run a crate or lib with the `#[cfg(test)]` configuration, handling inline
+tests, you should specify the crate directly like so.
+
+```
+rust_test(
+    name = "hello_lib_test",
+    crate = ":hello_lib",
+    # You may add other deps that are specific to the test configuration
+    deps = ["//some/dev/dep"],
+)
+```
 
 ### Example: `test` directory
 
--- a/test/inline_test_with_deps/test_with_srcs/BUILD
+++ b/test/inline_test_with_deps/test_with_srcs/BUILD
@@ -1,0 +1,20 @@
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "//rust:rust.bzl",
+    "rust_library",
+    "rust_test",
+)
+
+rust_library(
+    name = "inline",
+    edition = "2018",
+    srcs = ["src/lib.rs"],
+)
+
+rust_test(
+    name = "inline_test",
+    crate = ":inline",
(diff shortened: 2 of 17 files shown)


mfarrugi:
The easiest work sound for this right now is providing srcs, though it is
redundant with the corresponding rust_library.

acmcarther:
Hmm.

I think you may have addressed this in the PR description, but just for extra clarity: is it not currently possible to provide these additional dependencies in `rust_test.deps`? Or does this have unintended side effects?

I expect it might be the case that `rust_test.deps` has the special meaning that "the tests for deps targets will be executed". If that's the case, then I think I might prefer a breaking change of adding a "lib" attr to `rust_test` that accepts the target under test, and repurpose `deps` for test-only dependencies. Those test-only dependencies would be specified in the normal way in the source file e.g

```rust
#[cfg(test)]
extern crate test_only_dep
```

I think this supports the rust-y model of bundling tests with the source while also more closely matching existing rulesets.

GregBowyer:
@acmcarther that works, and might give us some additional flexibilty to do odd things with `srcs` in the future.

If we went down that road would we want to consider some form of deprecation warning for example

```
Single dep given without srcs - Assuming that you are unit-testing inline test code. You should consider moving to `test_lib` as this form will be deprecated on 2020-01-01. See #203 for details
```

acmcarther:
I don't have strong opinions about the best deprecation process. I'm a bit hesitant to be to gradual with this particular feature because (1) most users currently only have one dep in rust_test.deps anyway because of the issue this PR addresses and (2) I'm unsure how to support the needed test dep functionality during the deprecation period.

@mfarrugi do you have any opinion about the suggestion of using a new `lib` single-target attr, the re-usage of deps for test dependencies, and if you're amenable to the approach, what the deprecation process might look like?

mfarrugi:
I mentioned in #202 I thought the options were either new single-target (ie. lib) parameter, or something more thorough.

I disagree about most users with only 1 dep in deps, plenty of my tests look like:

```
rust_library(
    name = "example",
    srcs = ["lib.rs"]
)

rust_test(
    name = "example_test",
    srcs = ["lib.rs"],
    deps = [
        "//third_party/rust:pretty_assertion",
    ],
)
```
---
```
rust_test(
    name = "example_test",
    crate = ":example",
    deps = [
        "//third_party/rust:pretty_assertion",
    ],
)
``` 
looks fine to me, and so does
```
rust_library(
    name = "example",
    srcs = ["lib.rs"],
    dev_deps = [
        "//third_party/rust:pretty_assertion",
    ],
)

rust_test(
    name = "example_test",
    crate = ":example",
)
```
though the latter seems needlessly complicated.

I don't think we should make an excessive effort to avoid breaking changes since many of the changes we have recently made are subtley breaking, and we are pre-1.0 anyway. I think we can look at ironing out the API when users start complaining, or when Bazel hits 1.0.

damienmg:
Hi,

Thanks for doing that!

I personally would prefer something like the second option that @mfarrugi proposed:
```
rust_test(
    name = "example_test",
    crate = ":example",
    deps = [
        "//third_party/rust:pretty_assertion",
    ],
)
```

Maybe also keep the src version too, so either you test a pure test crate (have "srcs") or test inline tests (have "crate").

I don't really like `dev_deps` that sounds like tools dependencies (e.g. deps you use to generate stuff when you use cargo). This are clearly test dependencies so adding them as deps of the `rust_test` just seems logical from my point of view.

GregBowyer:
So I made it like `crate` as suggested above.

Internally this did cause some people pain for migration so I added in a fail message of the following format

```
greg@gregslaptop ~/projects/rules_rust $ bazel test //...
INFO: Invocation ID: e325c15f-c46b-4afb-9214-d3761f633c59
ERROR: /home/greg/projects/rules_rust/test/chained_direct_deps/BUILD:28:1: in rust_test rule //test/chained_direct_deps:mod1_test:
    File "/home/greg/projects/rules_rust/test/chained_direct_deps/BUILD", line 28
        rust_test(name = 'mod1_test')
    File "/home/greg/.cache/bazel/_bazel_greg/f3b74a36572ec13668c86aed8ee3d7cc/external/io_bazel_rules_rust/rust/private/rust.bzl", line 195, in _rust_test_impl
        _rust_test_common(ctx, ctx.outputs.executable)
    File "/home/greg/.cache/bazel/_bazel_greg/f3b74a36572ec13668c86aed8ee3d7cc/external/io_bazel_rules_rust/rust/private/rust.bzl", line 174, in _rust_test_common
        fail(msg)
--------------------------------------------------------------------------------
Testing crates with inline `#[cfg(test)]` attributes is now handled with the
`crate` attribute in `rust_test`:
This means you should migrate from:
```
rust_test(
    name = "mod1_test",
    deps = [":mod1"],
    ...
)
```
to the following:
```
rust_test(
    name = "mod1_test",
    crate = ":mod1",
    ...
```
See https://github.com/bazelbuild/rules_rust/pull/203 for more details
--------------------------------------------------------------------------------
```

Its possibly a reasonable deprecation? It could be killed soon if needed, but I think it will confuse a lot of people in the interim to not have their builds break in a meaningful way.

rust/private/rust.bzl
```diff
@@ -124,19 +152,26 @@ def _rust_test_common(ctx, test_binary):
...
+        crate = ctx.attr.crate[CrateInfo]
         target = CrateInfo(
             name = test_binary.basename,
-            type = parent_crate.type,
-            root = parent_crate.root,
-            srcs = parent_crate.srcs,
-            deps = parent_crate.deps,
+            type = crate.type,
+            root = crate.root,
+            srcs = crate.srcs + ctx.files.srcs,
```
mfarrugi: The use-case for this is additional test-only files?

rust/private/rust.bzl
```diff
@@ -281,6 +316,18 @@ _rust_library_attrs = {
...
 }
 
+_rust_test_attrs = {
+    "crate": attr.label(
+        mandatory = False,
+        doc = _tidy("""
+            Target inline tests declared in the given crate
+
+            These tests are typically those that would be held out under
+            the `#[cfg(test)]` declarations.
```
mfarrugi:
```suggestion
            `#[cfg(test)]` declarations.
```

test/inline_test_with_deps/test_with_srcs/BUILD
```diff
@@ -0,0 +1,20 @@
...
+rust_library(
+    name = "inline",
+    edition = "2018",
+    srcs = ["src/lib.rs"],
+)
+
+rust_test(
+    name = "inline_test",
+    crate = ":inline",
+    srcs = ["src/extra.rs"],
```
mfarrugi: What is the motivation for doing this without making it a separate test target?

mfarrugi:
Looks good to me. I don't feel strongly about rust_test.srcs, but asked about it for context.

This is a bit of a tangent, but this change makes me feel like a `cargo_crate` macro that sets up the relevant set of {`rust_library`, `rust_test`, `rust_doc_test`, `rust_binary`} targets could be useful in the future as well.

rust/private/rust.bzl
GregBowyer: yes

test/inline_test_with_deps/test_with_srcs/BUILD
GregBowyer: primarily to hold out things like `test_ds_helpers.rs` that are test helpers used in inline tests but are not really general code for the resulting library.

I could also go either way

GregBowyer:
Last change was nothing massive, just the typo @mfarrugi suggested

GregBowyer:
Do we want to merge this?

mfarrugi:
It's certainly mergeable. Only question is whether you still want to include the test-only sources support.

GregBowyer:
I would, but thats because I support users who are using that functionality

vitalyd:
I just got hit by this change while upgrading to the latest `rules_rust`.  @GregBowyer, it's probably a good idea to update https://github.com/bazelbuild/rules_rust/blob/master/docs/index.md#rust_test to be consistent with current code?

Issue #204, opened 2019-03-13
Pass linker script through to sandbox


m3rcuriel:
Currently, there is no mechanism by which to forward files into the sandbox for compile-time use.

The specific use-case I ran into differs from #79, because instead of dealing with Rust's mechanisms for including files at compile time, I am attempting to forward a linker script all the way through, such that the following code will work:

```bzl
rust_binary(
  ...
  rustc_flags = [
    "--codegen=link-arg=-Tscript.ld",
  ],
)
```

This serves the usecase of embedded Rust by allowing the author to lay out the binary in specific ways. Additionally, this mirrors https://github.com/bazelbuild/bazel/issues/184.

Based on the "proof of concept" implementation I have (and I use the word loosely), this more or less comes down to needing to filter these out in `rustc.bzl::collect_deps`, wrap them into a new field of `dep_info`, add them into `compile_inputs`, and do something like `args.add_all(["-L" + file.dirname for file in dir_info.link_scripts])`. Furthermore, this would just boil down to checking file extensions for `.ld`, `.lds`, and `.ldscript`, which is how Bazel handles this in Java for CcBinary anyways.

Another way to handle this, of course, is to do it in the toolchain itself but since this is the behavior of `cc_binary` it presumably can be argued for `rust_binary`.

mfarrugi:
This sounds reasonable. 

I'm unfamiliar with this, so you can explain how linker scripts would be used from transitive dependencies? Should `"--codegen=link-arg=-Tscript.ld",` still be declared per binary, or is having a linker script in your transitive dependencies enough? If there are multiple linker scripts, what order are they passed in?

When you say the alternative is putting it in the toolchain, do you mean having an optional `linker_scripts` attr on `rust_toolchain`? I agree it's fair to do the same thing as `cc_binary`, but does it make sense for a target to be able to build with different linker scripts / toolchains?

I think another question is whether we should be more permissive in what files we allow as sources, but that's out of scope for this since providing linker flags in the rules is The Right Thing to do.

m3rcuriel:
`--codegen=link-arg=-Tscript.ld`, according to the `cc_binary` model would be declared only on the `cc_binary` target, although I suppose it's unavoidable for it to be transitive.

[...]

PR #208, opened 2019-03-20
Add linker_script option to rust_binary

m3rcuriel:
This explicitly adds a `linker_script` option to `rust_binary` as suggested in the bottom of #204.

It's fundamentally identical to adding the ldscript to data and then adding `--codegen=link-args=-T$(location //:ldscript)` to `rustc_flags`, but now without the hassle of figuring that out yourself.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -354,9 +354,18 @@ INFO: Elapsed time: 1.245s, Critical Path: 1.01s
 """,
 )
 
+_rust_binary_attrs = {
+    "linker_script": attr.label(
+        doc = _tidy("""
+            Link script to forward into linker via rustc options.
+        """),
+        allow_single_file = True,
+    ),
+}
+
 rust_binary = rule(
     _rust_binary_impl,
-    attrs = _rust_common_attrs,
+    attrs = dict(_rust_common_attrs.items() + _rust_binary_attrs.items()),
     executable = True,
     fragments = ["cpp"],
     host_fragments = ["cpp"],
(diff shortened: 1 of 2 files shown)


m3rcuriel:
Also because I just thought of it does anyone else think we should maybe break `data` into runtime and compile time data?

rust/private/rust.bzl
```diff
@@ -354,9 +354,18 @@ INFO: Elapsed time: 1.245s, Critical Path: 1.01s
 """,
 )
 
+_rust_binary_attrs = {
+    "linker_script": attr.label(
+        doc = _tidy("""
+            Link script to forward into linker via rustc options.
+        """),
```
damienmg: Add `cfg = "host",` so the script is passed as a host transition

damienmg:
> Also because I just thought of it does anyone else think we should maybe break `data` into runtime and compile time data?

In which case is there's a need for compile time data that is not passed as a different attribute? Do you mean a linker script can use data? If so and there is no other case, maybe either add a linker_scripts_data or make linker_script a label_list and have a convention about which one is the script.

m3rcuriel:
> > Also because I just thought of it does anyone else think we should maybe break `data` into runtime and compile time data?
> 
> In which case is there's a need for compile time data that is not passed as a different attribute? Do you mean a linker script can use data? If so and there is no other case, maybe either add a linker_scripts_data or make linker_script a label_list and have a convention about which one is the script.

I was referring to data in general not in the context of linker scripts. `data` in Bazel in general seems to me to refer to runtime data, i.e in https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes.

I understand that compile time dependencies are good for `include_str!` but I was just pointing that out.

mfarrugi:
On the compile_data tangent, yes I think that would be a useful distinction. Bundling it into out_dir_tar makes sense, but I don't recall that being easy/intuitive to use.

Issue #200, opened 2019-03-05
Get all tests passing with --incompatible_disable_legacy_cc_provider for Bazel 0.25.0


oquenchil:
It seems that rules_rust will start breaking with the next Bazel release 0.25 because of an incompatible change (https://github.com/bazelbuild/bazel/issues/7036)

```
ERROR: /home/bazel/.cache/bazel/_bazel_bazel/c922ef9e5f39c5bf7c8601dab4c6fea8/external/examples/ffi/rust_calling_c/BUILD:29:1: in rust_library rule @examples//ffi/rust_calling_c:matrix_dynamically_linked:
--
  | Traceback (most recent call last):
  | File "/home/bazel/.cache/bazel/_bazel_bazel/c922ef9e5f39c5bf7c8601dab4c6fea8/external/examples/ffi/rust_calling_c/BUILD", line 29
  | rust_library(name = 'matrix_dynamically_linked')
  | File "/home/bazel/.cache/bazel/_bazel_bazel/c922ef9e5f39c5bf7c8601dab4c6fea8/external/io_bazel_rules_rust/rust/private/rust.bzl", line 85, in _rust_library_impl
  | rustc_compile_action(ctx = ctx, toolchain = toolchain, cr...)), ...)
  | File "/home/bazel/.cache/bazel/_bazel_bazel/c922ef9e5f39c5bf7c8601dab4c6fea8/external/io_bazel_rules_rust/rust/private/rustc.bzl", line 180, in rustc_compile_action
  | collect_deps(crate_info.deps, toolchain)
  | File "/home/bazel/.cache/bazel/_bazel_bazel/c922ef9e5f39c5bf7c8601dab4c6fea8/external/io_bazel_rules_rust/rust/private/rustc.bzl", line 116, in collect_deps
  | fail(("rust targets can only depend o...)), ...")
  | attribute deps: rust targets can only depend on rust_library, rust_*_library or cc_library targets.<target @examples//ffi/rust_calling_c/c:native_matrix_so>

```

The code is using the old legacy provider `cc`, it can be replaced with CcInfo to get the same functionality as described [here](https://github.com/bazelbuild/bazel/issues/7036).

mfarrugi:
Same as #190. 

Looks like CI is on bazel 0.22 right now, which suggests it's still to early to bump rules_rust minimum bazel version to 0.22.

oquenchil:
[...]

Issue #190, opened 2019-02-08
Update cc api usages before bazel 24


mfarrugi:
https://github.com/bazelbuild/bazel/issues/7036 needs to be fixed to support bazel >= 24, but fixing requires bazel >= 22

Lets wait for at least 23 to release before bumping the minimum version to 22.

PR #209, opened 2019-03-27
#190, #200 :: Migrate to new cc provider.

mfarrugi:
Locally tested this works with and without the flag on bazel v22.

### DIFF
--- a/rust/private/legacy_cc_starlark_api_shim.bzl
+++ b/rust/private/legacy_cc_starlark_api_shim.bzl
@@ -1,28 +1,24 @@
 """
-Utility functions that replace old C++ Starlark API using the new API.
-See migration instructions in https://github.com/bazelbuild/bazel/issues/7036
-
-Pasted from https://gist.github.com/oquenchil/7e2c2bd761aa1341b458cc25608da50c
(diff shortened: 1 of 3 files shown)


mfarrugi:
cc @hlopko @damienmg

rust/private/legacy_cc_starlark_api_shim.bzl
```diff
@@ -0,0 +1,28 @@
+"""
+Utility functions that replace old C++ Starlark API using the new API.
+See migration instructions in https://github.com/bazelbuild/bazel/issues/7036
+
+Pasted from https://gist.github.com/oquenchil/7e2c2bd761aa1341b458cc25608da50c
```
damienmg: There is no license from this code, can we get the author to add one?

hlopko: Author is on vacation, and I'm pretty sure attribution is not needed (in the same way as StackOverflow answers aren't licensed).

mfarrugi: StackOverflow answers are licensed as creative commons. If this not covered by some Google default oss license?

hlopko: Ok, then I'm not a lawyer :) So don't copypaste, write it by your own, and don't attribute?

PR #217, opened 2019-04-23
Add cc toolchain files to rustc dependencies (for linker)

m3rcuriel:
Currently if you're using an artifacted compiler the Rust rules will complain that they can't find the linker script. This is because it is not a dependency.

I'm pretty sure there's a correct way to do this with the new toolchain system but I was looking around and saw that people mostly stuck with this with some TODOs to find a better way.

I tried to get it using `ctx.toolchains` but that doesn't actually have files, it seems.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -194,6 +194,7 @@ def rustc_compile_action(
         dep_info.transitive_libs +
         [toolchain.rustc] +
         toolchain.crosstool_files +
+        ctx.files._cc_toolchain +
         ([] if linker_script == None else [linker_script]),
         transitive = [
             toolchain.rustc_lib.files,
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -194,12 +194,12 @@ def rustc_compile_action(
         dep_info.transitive_libs +
         [toolchain.rustc] +
         toolchain.crosstool_files +
-        ctx.files._cc_toolchain +
         ([] if linker_script == None else [linker_script]),
         transitive = [
             toolchain.rustc_lib.files,
             toolchain.rust_lib.files,
-        ],
+            # TODO wait for CcToolchainInfo.all_files to be available
+        ] + ctx.files._cc_toolchain,
     )
 
     args = ctx.actions.args()
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -188,8 +188,13 @@ def rustc_compile_action(
 
     linker_script = getattr(ctx.file, "linker_script") if hasattr(ctx.file, "linker_script") else None
 
-    # TODO wait for CcToolchainInfo.all_files to be available
-    toolchain_depset = depset(ctx.files._cc_toolchain)
+    cc_toolchain = find_cpp_toolchain(ctx)
+
+    if (len(BAZEL_VERSION) == 0 or
+         versions.is_at_least("0.25.0", BAZEL_VERSION)):
+        linker_depset = find_cpp_toolchain(ctx).all_files
+    else:
+        linker_depset = depset(ctx.files._cc_toolchain)
 
     compile_inputs = depset(
         crate_info.srcs +
@@ -201,7 +206,7 @@ def rustc_compile_action(
         transitive = [
             toolchain.rustc_lib.files,
             toolchain.rust_lib.files,
-            toolchain_depset
+            linker_depset,
         ],
     )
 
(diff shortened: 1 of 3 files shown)


mfarrugi:
Looks fine to me, but I'd like another thumbs up since I know the least
about bazel :)

rust/private/rustc.bzl
```diff
@@ -194,6 +194,7 @@ def rustc_compile_action(
         dep_info.transitive_libs +
         [toolchain.rustc] +
         toolchain.crosstool_files +
+        ctx.files._cc_toolchain +
```
damienmg: Add a comment explaining the why of this addition and with a TODO to find a canonical way. From what I read from this doc theses files should be included in crosstool_files.

side-reflection: this is a pretty huge sum of lists. I am not sure about the resulting complexity (java linked list should be fine but IIRC those are ArrayList, who have O(n+m) complexity). Also the transitive_libs can become very huge, this could be a bottle neck on large list of compilation.

hlopko:
Hi all,

to make sure I understand the issue correctly, you want to do this so the linker is present for the linking action, not the linker script (one is tool that does linking, another one is a script that ld can consume to tweak the linking process, used e.g. to specify symbol visibility etc), correct? 

The best way to access C++ toolchain files is to use [`all_files`](https://docs.bazel.build/versions/master/skylark/lib/CcToolchainInfo.html#all_files) , which is only in 0.25 (https://github.com/bazelbuild/bazel/commit/f59038d2a46937287ab25316676287905ab67009) :). Until then, `ctx.attr._cc_toolchain.files` is good enough.

The best way to access the C++ toolchain is to use [`find_cpp_toolchain`](https://source.bazel.build/bazel/+/8bd7e4822ed96615ccac77cc40ca96eddc74c510:tools/cpp/toolchain_utils.bzl;l=23) which is forward compatible with https://github.com/bazelbuild/bazel/issues/6516. But this makes the full sense only in 0.25 with `all_files`.

rust/private/rustc.bzl
hlopko: Yup I second @damienmg's comment, make a good use of the `transitive` attribute of depset https://docs.bazel.build/versions/master/skylark/lib/globals.html#depset :)

m3rcuriel:
@hlopko correct, this is in order to ensure that the linker itself (and its dependencies!) are available in the sandbox during linking.

I was going to use `all_files` as I am a nerd and run whatever most recent Bazel I can but stuck with `cxt.attr._cc_toolchain.files` because I realized `all_files` is extremely new.

Currently we get the toolchain out of `find_cpp_toolchain`. When we want to migrate to `all_files` we just have to move the depset down about 50 lines below where we get the linker to pass into rustc anyways.

rust/private/rustc.bzl
```diff
@@ -198,7 +198,8 @@ def rustc_compile_action(
         transitive = [
             toolchain.rustc_lib.files,
             toolchain.rust_lib.files,
-        ],
+            # TODO wait for CcToolchainInfo.all_files to be available
```
damienmg:
```suggestion
            # C++ library linked in rust libraries need the linker to be available at compile-time.
            # TODO: Switch to CcToolchainInfo.all_files once available.
```

m3rcuriel:
@mfarrugi I have no idea why this failed CI on one OS, and I don't have OS X to test on.

mfarrugi:
You can get CI to run again by rebasing or making some cosmetic changes.

At a glance the osx error looks like it's not flakey.. but would have
failed on all platforms if that were the case.

rust/private/rustc.bzl
```diff
@@ -198,7 +198,8 @@ def rustc_compile_action(
...
+        ] + ctx.files._cc_toolchain,
```
mfarrugi: Is ctx.files._cc_toolchain a depset of files? Should it just go into the transitive list?

m3rcuriel: Tried that, Linux (at least) complains. I am borrowing a Mac to test against tonight.

m3rcuriel:
So we should probably add an artifacted compiler/linker set up to test now that I've gotten it working.

The reason it was failing is because `ctx.files._cc_toolchain` is an empty list on Linux, but contains `File`(s) on OS X. It needed to be wrapped in a depset. That it didn't fail / that no one but me has encountered the problem this PR should fix means that I should probably take my setup, strip it down, and make it part of the test suite.

m3rcuriel:
Now featuring the required fixes to get this to work with `--incompatible_enable_cc_toolchain_resolution`.

I'm fairly sure `CcToolchainInfo.all_files` is introduced in 0.25.0 since that's when it started working for me but I could be wrong.

PR #239, opened 2019-07-08
Make rust_proto_library work with proto_library rules using import_prefix / strip_import_prefix .

lberki:
In addition, this makes it possible to fix http://github.com/bazelbuild/bazel/issues/7157 (patch upcoming)

### DIFF
--- a/proto/proto.bzl
+++ b/proto/proto.bzl
@@ -52,6 +52,10 @@ RustProtoProvider = provider(
 def _compute_proto_source_path(file, source_root_attr):
     """Take the short path of file and make it suitable for protoc."""
 
+    virtual_imports = "/_virtual_imports/"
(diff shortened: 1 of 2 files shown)


damienmg:
This looks correct, but I don't understand from where "_virtual_imports/" is coming? Who generate this?

lberki:
Wow, that was quick :)

Bazel generates this. At HEAD, when the `proto_library` rule uses the `strip_import_prefix` or the `import_prefix` attribute and I want to make it so that it is also generated when the `proto_library` has a generated source (I added you to the internal change).

The code is here:

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java#L294

lberki:
Done. I'll merge then.

Issue #186, opened 2019-01-21
Add support for Web Assembly


0xADD1E:
The wasm32-unknown-unknown target doesn't really fit in with the current model of physical architectures, but is one of the key selling points for rust to many people

mfarrugi:
What were you trying to do?

I know nothing about wasm. At a glance (reading https://medium.com/@ianjsikes/get-started-with-rust-webassembly-and-webpack-58d28e219635)... getting rustc to emit .wasm shouldn't require too much finagling to the repositories.bzl configurations, but producing something useful requires a bunch of extra steps/tools we would need to teach bazel about as well.

0xADD1E:
I personally was trying to create a website using yew. For me, the logical division would be rules_rust handling the wasm generation alone, with another set of special rules for taking the place of cargo-web

mfarrugi:
Would that be some rules_wasm, or are there still rust specific things that
need to happen?

0xADD1E:
In a perfect world, I think this would have rules_rust handling wasm generation alone, and some rules_wasm_shims creating the JS/HTML bindings needed to get a full site from rust sources. That said, wasm-bindgen/cargo-web are highly integrated into the process, and do need access to the source, but also have the ability to generate multiple output files...

qzmfranklin:
@0xADD1E : Agree that the rules_rust should generate .wasm files and leave the rest to other tools and rules.

Looking at the rules_rust repository and the rust-wasm project, I just do not feel that it is ready to start working on it.

Issue #206, opened 2019-03-14
Target triple parsing incomplete


m3rcuriel:
Target triples are unfortunately much more complicated than "either three or four things separated for a dash." More specifically, GNU explicitly allows you to leave out "unknown". This means that `thumbv7em-none-eabihf` is not only a valid target triple, but a valid target triple that Rust lists in supported platforms (admittedly Tier 3).

m3rcuriel:
Note that #207 admits a partial solution but it is incomplete to force people to override whenever there are non perfectly organized target triples.

sayrer:
Is there a place to read up on how these are parsed? For example "aarch64-fuschia" is also a Rust target triple.

m3rcuriel:
I would describe gcc as a reasonable reference implementation, see

https://github.com/gcc-mirror/gcc/blob/master/config.sub

I don't think duplicating this is necessary since Rust behaves correctly,
but the way rules_rust handles constraints would require something like
this.

sayrer:
hmm, how could that script output "aarch64-fuschia"?

I did mention that directly, and it's listed here: https://forge.rust-lang.org/platform-support.html

m3rcuriel:
https://github.com/gcc-mirror/gcc/blob/master/config.sub#L1365

While I am not studied in the exact mechanisms of this subroutine, I'm going to assume that it works since fuschia is listed as a valid system.

The way this script (I believe) works is by taking the first piece of the triple and saying "does this match any of my architectures" followed by vendors, systems, ABI.

The config.sub script takes `aarch64-fuchsia`, and outputs `aarch64-unknown-fuchsia`, which is correct as the script exists to map the really strange world of target triples into always-sane three or four item ones -- the ones that rules_rust expects, not the ones Rust expects.

Note:
```
$ rustc --target=aarch64-unknown-fuchsia
error: Error loading target specification: Could not find specification for target "aarch64-unknown-fuchsia"
  |
  = help: Use `--print target-list` for a list of built-in targets```

So tl;dr Rust does things its own way, which might beg a ticket for upstream, and config.sub exists to convert "weird" target triples into parse-able ones.

As an aside, a possible mitigation of this problem is to just distribute or pull config.sub, and run it on the target. `rustc` still gets passed the original target passed by the user, which is Rust-valid, but when rules_rust parses constraints out of it it uses the "fixed" form.

[...]

PR #240, opened 2019-07-17
Add support for wasm-bindgen.

johnedmonds:
A few things to note:

- Fixes https://github.com/bazelbuild/rules_rust/issues/186
- It looks big but most of the changes are generated by cargo-raze
- This might be controversial but we added web assembly by having all the rules output an optional `.wasm` file (similar to how `_deploy.jar` works). I chose to do it this way mainly because bazel doesn't yet support multiple target configs. Our use case is mainly being able to use a single bazel build command to generate a docker image containing the server binary and the client code. And also to let us run all the tests together. If you prefer though we could try changing wasm so it would be more like a traditional platform.


WORKSPACE
```diff
@@ -54,8 +57,13 @@ skydoc_repositories()
 load("@io_bazel_rules_sass//:package.bzl", "rules_sass_dependencies")
 rules_sass_dependencies()
 
-load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
+load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "npm_install")
 node_repositories()
+npm_install(
```
acmcarther: Are these are example-only deps? Might be helpful to have a comment above them. I know we haven't done this historically, but I think it will become increasingly more helpful to have documentation as the surface area of rules_rust grows.

rust/platform/triple_mappings.bzl
```diff
@@ -61,6 +65,7 @@ _SYSTEM_TO_DYLIB_EXT = {
     # TODO(acmcarther): To be verified
     "windows": ".dll",
     "emscripten": ".js",
+    "unknown": ".wasm",
```
acmcarther: It seems somewhat presumptive to say that any "unknown" system == wasm. I understand that this is a consequence of the current way target triples are resolved, but I find it regrettable.

rust/private/rustc.bzl
```diff
@@ -33,6 +33,7 @@ CrateInfo = provider(
         "srcs": "List[File]: All source Files that are part of the crate.",
         "deps": "List[Provider]: This crate's (rust or cc) dependencies' providers.",
         "output": "File: The output File that will be produced, depends on crate type.",
+        "wasm_output": "File: The output File that will be produced, depends on crate type. Built for the wasm32-unknown-unknown target if requested.",
```
acmcarther: It strikes me as unusual that the WASM target needs special fields in the standard rust rules and providers.

I suspect this was not your first choice. Can you tell me a bit about the rationale for this arrangement? Is this because the existing toolchain makes too many assumptions (about output, or dependency file types, or something else)?

rust/repositories.bzl
```diff
@@ -303,8 +306,22 @@ def _rust_toolchain_repository_proxy_impl(ctx):
...
+    BUILD_components.append(BUILD_for_toolchain(
+        name = "{toolchain_prefix}_wasm32-unknown-unknown".format(
+            toolchain_prefix = ctx.attr.toolchain_name_prefix,
+        ),
+        exec_triple = ctx.attr.exec_triple,
+        parent_workspace_name = ctx.attr.parent_workspace_name,
+        # Technically we should say it's compatible with wasm32-unknown-unknown but in practice we
+        # want to be able to produce wasm output when targeting the host computer (e.g. for building
+        # a docker image). Thus we make this the exec triple.
+        target_triple = ctx.attr.exec_triple,
```
acmcarther: I see you have a pretty big comment here, but I'll ask anyway.

This defies my original intent for the target_triple field. My hope was that "wasm32-unknown-unknown" could be specified as a target triple in the way of other targets, and I guess, that it would never make sense to compile for host. You mention that there is a docker usecase for doing that, can you elaborate a bit?

WORKSPACE
johnedmonds: Yep that's correct. Added a comment.

rust/platform/triple_mappings.bzl
johnedmonds: Agreed. Let me see if I can find a better way to do this.

rust/private/rustc.bzl
johnedmonds: Right, this wasn't the ideal solution. It all ties together with your next comment about where we set the target triple to the exec triple so I'll try to address it there about why this first draft has multiple outputs for each rule.

rust/repositories.bzl
johnedmonds: Yes this is a good point and I mentioned it might be the controversial part of this PR in the initial request. Basically the use-case is that we have a server binary written in rust (we can call it `//server` for example) and some client code written in TypeScript/JavaScript (let's call the bundle of compiled .js and assets `//web`). Then we have a docker image where we package it all together and ship it to production (e.g. `//:deployment_image` which depends on `//server` and `//web`). The problem is that //server needs to target something like x86_64-linux-gnu while //web needs to target wasm32-unknown-unknown. Thus for now, you wouldn't be able to build //:deployment_image because bazel only supports a single target architecture at a time (see https://stackoverflow.com/questions/51423896/bazel-build-multiple-target-architectures-with-one-invocation). You'd also have trouble running all the tests at once (`bazel test //...`) for similar reasons.

The way we solve it here is that each rule produces its normal output but also a `%name%.wasm` and that .wasm file can be built because it technically targets the host architecture.

Still I agree this is a bit messy so I'll leave it up to you whether this is a valid use-case. In the worst-case scenario we can just call bazel twice with the different constraints and then introduce a mini-build-system on top of bazel to do the final image assembly. So I'd be willing to change this PR so it acts a bit more like you'd expect where you can specify the constraints to pick different toolchains.

johnedmonds: I did some more digging and found https://docs.google.com/document/d/1vc8v-kXjvgZOdQdnxPTaV0rrLxtP2XwnD2tAZlYJOqw/edit#heading=h.pn8nbhno42p5 which says it should allow custom transitions between constraints. So I think you're right and we can just remove all the additional outputs and implement wasm mostly normally.

rust/platform/triple_mappings.bzl
acmcarther: I suspect it wont be possible to fix this with the current structuring. Seems the assumption that components of target triples could be independently resolved was a faulty one.

I think we can go with what's here for now with a note that target triples need their properties resolved "holistically" (as in, with the full triple available). I was going to file an issue for this, but it seems https://github.com/bazelbuild/rules_rust/issues/206 captures it pretty well. I'll mirror this comment into that issue for completeness.

rust/repositories.bzl
acmcarther: Ah I understand the problem. The output target is not wasm32-unknown-unknown, but rather the server platform. The Rust code involved in the output artifact is closer to "data" for the final target than to an executable for the configured Bazel platform.

This seems to be a specific instance of the general problem of building an "executable" that serves "executable". This is clearly an underserved use case for Bazel. My cursory research on the subject indicates that this is typically handled with a "wrapper" build rule over the standard build rules, that uses Bazel's experimental transition feature[1]. I think this is what you linked in your latest comment

This goes pretty far outside my current knowledge of Bazel platform configuration support but I agree with your latest assessment (which agrees with my initial suspicion) -- we should ideally be able to support that use case using the existing toolchain options.

[1] https://docs.bazel.build/versions/master/skylark/lib/transition.html

rust/platform/triple_mappings.bzl
johnedmonds: Makes sense. Then is it okay to leave it like this for now?

rust/repositories.bzl
johnedmonds: I pushed some changes to use Bazel's transition feature. Bazel 0.29.0 was released yesterday with the fix for https://github.com/bazelbuild/bazel/issues/8963 so it should work now. The only thing I'm still trying to figure out is why RBE image fails on the `matrix_dylib_test`.

qzmfranklin:
This is a great PR. Thanks for the work.

@acmcarther @johnedmonds  : Joining your discussions about building for multiple platforms at the same time. The `go_binary` rule from [rules_go](https://github.com/bazelbuild/rules_go) supports exactly the use case that @johnedmonds mentioned with `goos` and `goarch`. I have to admit that they should probably go for the experimental `transition` feature that @acmcarther mentioned. But that is pretty immature as of today (citation needed). Expectedly, it will be while until we get there. Before that, it is possibly better to use an approach similar to how `go_binary` handled this situation.

Our team have an almost identical situation as @johnedmonds . We'd like to use rust-wasm as well. I

qzmfranklin:
I do have a very small point to bring up.

I agree with @0xADD1E mentioned in https://github.com/bazelbuild/rules_rust/issues/186,
> In a perfect world, I think this would have rules_rust handling wasm generation alone, and some rules_wasm_shims creating the JS/HTML bindings needed to get a full site from rust sources. That said, wasm-bindgen/cargo-web are highly integrated into the process, and do need access to the source, but also have the ability to generate multiple output files...

I'd like to grab the raw `.wasm` file as I'd get with `cargo build --target wasm32-unknown-unknown` and use a separate rule to run `wasm-bindgen` on it to get the `pkg/{*.wasm,.js,.d.ts}` files. 

The rationale being that the `cargo` ecosystem is difficult use in China due to slow network connections to AWS S3, where the cargo crates are actually hosted. We'd prefer to vendor the various binaries including `wasm-bindgen` instead of building it from source, which is usually too unreliable and slow for us.

That said, I think a reasonable tradeoff is to have a place in this PR for us to easily swap in our own vendored `wasm-bindgen` binary instead of having to build from source.

What do you think?

johnedmonds:
Thanks for the kind comments @qzmfranklin. With regards to this:

> I'd like to grab the raw .wasm file as I'd get with cargo build --target wasm32-unknown-unknown and use a separate rule to run wasm-bindgen on it to get the pkg/{*.wasm,.js,.d.ts} files.

I think you can already do that with this PR by using `rust_wasm_bindgen_toolchain` to make a new wasm_bindgen toolchain where you specify your vendored executable. Then you can override the default toolchain on the invocation of the wasm bindgen rule to use your new toolchain. Or am I misunderstanding your need?

qzmfranklin:
@johnedmonds Thanks for the informative reply!

I'll definitely try it out before too long, pending other priorities.

BTW, I really like the way this PR landed the `.wasm` support, i.e., similar to `_deploy.jar`. It is probably a path of least resistance given the status.

To truly solve the cross-compilation problem in rules_rust, I'd believe an overhaul is probably necessary.

README.md
```diff
@@ -22,6 +22,15 @@ This repository provides rules for building [Rust][rust] projects with [Bazel](h
   </ul>
 </div>
 
+#### WebAssembly
+
+To build a `rust_binary` for wasm32-unknown-unknown add the `--platforms=//rust/platform:wasm` flag.
```
PiotrSikora: This is a great start, but I think that the requirement to set target platform is quite limiting.

WebAssembly is used outside of browsers / JavaScript host environments, and Wasm modules can be loaded by other build targets in the same Bazel workspace, and those build targets must be built using "native" platforms and not the `//rust:platform:wasm` pseudo-platform.

For example, imagine such `BUILD` file:
```
cc_library(
    name = "runtime",
    srcs = "runtime.cc",
    data = [
        ":wasm",
    ],
)

rust_library(
    name = "wasm",
    srcs = ["wasm.rs"],
    edition = "2018",
    crate_type = "cdylib",
    rustc_flags = ["--target=wasm32-unknown-unknown"],
)
```
where Wasm module compiled using `rust_library` must be loaded by the `runtime` written in C/C++.

The ability to override `--target` is something that I have hacked in my local tree, but ideally we would have something like `rust_wasm_library` that uses `--crate-type=cdylib --target=wasm32-unknown-unknown` to compile Wasm modules.

Basically, we need the ability to indicate Wasm target on a per build target basis, and not globally.

What do you think?

johnedmonds: I think I had a similar problem where I wanted to include the compiled wasm (compiled as wasm32-unknown-unknown) with `include_bytes` and serve it from the binary (compiled as a native Rust binary). We allow it in this PR by using Bazel transitions. See https://github.com/bazelbuild/rules_rust/pull/240/files#diff-9a42a2c125ed40c511198be1665d0faeR346. This allows us to build the entire build with the default target platform (e.g. x86 + Linux) but temporarily transition to `//rust:platform:wasm` when building the wasm code.

The easiest way is to do this in practice is to make a `rust_wasm_bindgen` target which will force the transition for the dependent `rust_library` (and transitively for all its dependencies). I have a note about this here https://github.com/bazelbuild/rules_rust/pull/240/files/7305a12fd276b0c21432d00d4e2ddd8b15115e13#diff-04c6e90faac2675aa89e2176d2eec7d8R31. As I read my note again I realize that perhaps it's a bit unclear since "transition" is a Bazel-specific word. Do you think it would be clearer if I referenced it in this paragraph?

Basically in this paragraph I was trying to explain how you could get the .wasm file if that's the only thing you wanted.

johnedmonds:
@acmcarther Could you take another look?

rust/repositories.bzl
```diff
@@ -293,6 +293,8 @@ def _rust_toolchain_repository_impl(ctx):
     for target_triple in [ctx.attr.exec_triple] + ctx.attr.extra_target_triples:
         BUILD_components.append(_load_rust_stdlib(ctx, target_triple))
 
+    BUILD_components.append(_load_rust_stdlib(ctx, "wasm32-unknown-unknown"))
```
acmcarther: Apologies if this is made obvious in a comment elsewhere in PR:

What's the motivation for specially including "wasm32-unknown-unknown"'s stdlib and toolchain decl instead of using the existing "extra_target_triples" mechanism?

e.g.
```python
    rust_repository_set(
        name = "rust_linux_x86_64",
        exec_triple = "x86_64-unknown-linux-gnu"
        # Specify that wasm32 should be brought in
        extra_target_triples = ["wasm32-unknown-unknown"],	
        version = RUST_VERSION,
    )
```
If that works, I'd (slightly) prefer that approach rather than have everyone automatically get wasm support. If we change to that, it'd probably be prudent to also mention it in the WASM section of the readme.

tools/whitelists/function_transition_whitelist/BUILD
```diff
@@ -0,0 +1,5 @@
+package_group(
+    name = "function_transition_whitelist",
+    packages = ["//..."],
+)
+
```
acmcarther: Nit: Rm extra newline?

rust/private/transitions.bzl
```diff
@@ -0,0 +1,20 @@
+def _wasm_transition(settings, attr):
+    if attr.crate_type == "proc-macro":
+        return {"//command_line_option:platforms": "@local_config_platform//:host"}
+    else:
+        return settings
+
+def _wasm_bindgen_transition(settings, attr):
+    return {"//command_line_option:platforms": "@io_bazel_rules_rust//rust/platform:wasm"}
+
+wasm_transition = transition(
```
acmcarther: Does it make sense to have a test for this functionality (maybe under //test/wasm_transition)?

qzmfranklin:
@johnedmonds : You might've already know about this. Just want to bring it to your attention.

Bazel's support for Android involves the use the `transition`s. More details can be found by searching 'transition' in [this page](https://docs.bazel.build/versions/master/android-ndk.html).

In short, the `deps` attribute of `android_library` causes these deps to use a different build configuration that the workspace-wide exec and host `constrait_value`. In the unlikely case of you not having seen this, it might be helpful as an example illustrating usages of `transition`s in Bazel.

I had another comment about the relative recent (about a month) redesign of `rules_nodejs` API. The providers are heavily redesigned and renamed. I noticed that this PR is still using the old rules_nodejs `ts_library` API. Might want to take a look.

Above and again, this is a great PR. Clear goals, pragmatic approaches, aggressive integrations. Would love to see it merge.

README.md
PiotrSikora: Oh, the transitions are great, I didn't realize that you could do it this way.

I think that `rust_wasm_bindgen` is close to what I want, but after reading that paragraph, I simply ignored it, because I don't want anything related to `wasm-bindgen`, I want pure Rust code that's compiled to `wasm32-unknown-unknown`.

As such, I'm suggesting that we would add `rust_wasm_library` that uses the transitions that you've already added for `rust_wasm_bindgen`, i.e.
```
cc_library(
    name = "runtime",
    srcs = "runtime.cc",
    data = [
        ":wasm",
    ],
)

rust_wasm_library(
    name = "wasm",
    deps = ":library"
)

rust_library(
    name = "library",
    srcs = ["library.rs"],
    edition = "2018",
)
```
Alternatively, since you're already doing it for `proc-macro`, I think that we could use `crate_type = "wasm32"` and simply transition to `wasm32-unknown-unknown` in the `rust_library`, i.e.
```
cc_library(
    name = "runtime",
    srcs = "runtime.cc",
    data = [
        ":library",
    ],
)

rust_library(
    name = "library",
    srcs = ["library.rs"],
    crate_type = "wasm32",
    edition = "2018",
)
```
What do you think?

rust/repositories.bzl
johnedmonds: Good catch. Done.

tools/whitelists/function_transition_whitelist/BUILD
johnedmonds: Done.

rust/private/transitions.bzl
johnedmonds: Perhaps although the existing `hello_world_wasm_test` does cover this because it depends on one of the wasm_bindgen dependencies. Removing the transition will cause the test to fail compilation. Do you think we need a specific test?

acmcarther: Ack, marking resolved.

README.md
PiotrSikora: Friendly ping before this gets merged.

johnedmonds: @PiotrSikora

PiotrSikora: I'm definitely against (4), since this feature has nothing to do with wasm-bindgen.

(3) would still require action from the person building the target (vs person writing `BUILD` file), so it's not much different than requiring extra `--platforms=wasm32` flag, IMHO.

I'm fine with either (1) or (2), with a slight preference for (1), since then you could easily add a target that extends existing `rust_library` to be also compiled as Wasm module (i.e. you could have both native and Wasm build targets built at the same time) by simply adding few lines with `rust_wasm_library` (see the example above). You could achieve the same with (2), but then you'd need to duplicate the complete build target, so it would be more error prone, I think.

Also, there is a precedence for `rust_xxx_library` with `rust_proto_library` and `rust_grpc_library`.

It's fine if you want to add it in a separate PR (i.e. I don't want to block this PR from getting merged), but could you rename it to "Add support for wasm-bindgen", since it's not a complete WebAssembly support?

What do you think?

johnedmonds:
@acmcarther Would you mind merging? I don't have permission to do the merge.

acmcarther:
Oops, my bad, done now.

PR #245, opened 2019-08-06
Add allow_empty to glob, as the patterns don't necessarily match something

laurentlb:
This is needed for the change in Bazel --incompatible_disallow_empty_glob

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -99,6 +99,7 @@ filegroup(
             "lib/rustlib/{target_triple}/lib/*{dylib_ext}",
             "lib/rustlib/{target_triple}/lib/*{staticlib_ext}",
         ],
+        # Some patterns (e.g. `lib/*.a`) don't match anything, see https://github.com/bazelbuild/rules_rust/pull/245
         allow_empty = True,
     ),
     visibility = ["//visibility:public"],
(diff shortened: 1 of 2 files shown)


rust/repositories.bzl
```diff
@@ -93,11 +93,14 @@ def BUILD_for_stdlib(target_triple):
...
-        "lib/rustlib/{target_triple}/lib/*{dylib_ext}",
-        "lib/rustlib/{target_triple}/lib/*{staticlib_ext}",
-    ]),
+    srcs = glob(
+        [
+            "lib/rustlib/{target_triple}/lib/*.rlib",
+            "lib/rustlib/{target_triple}/lib/*{dylib_ext}",
+            "lib/rustlib/{target_triple}/lib/*{staticlib_ext}",
+        ],
+        allow_empty = True,
```
damienmg: In which case does that glob is empty? Can a comment be added to justify it?

laurentlb: That's the error message I got:
`glob pattern 'lib/rustlib/x86_64-unknown-linux-gnu/lib/*.a' didn't match anything, but allow_empty is set to False.`

I don't know if it happens in other cases too.

damienmg: That seems to indicate a potential bug, this is expected that there is no .a file here. I guess the empty glob is acting if any of those is empty.

Can you just add a comment with the pattern that fails to justify this allow_empty? It's not important to fix at the moment but keeping track of why we did that is useful.

laurentlb: Added comment

damienmg: Thanks!

Issue #247, opened 2019-08-29
rules_rust broken by bazel@HEAD


c-parsons:
Example failure:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1169#cc99beed-1b8a-468f-83fa-ffa0475c5b73

Culprit: https://github.com/bazelbuild/bazel/commit/0c8f22e07a6fb7c31b2a9564a9ce4ae1196bb594

Looks like this is because rules_rust depends on an old version of `@io_bazel`, which still uses `PACKAGE_NAME` instead of `package_name()`

alandonovan:
Sorry for the breakage. I still don't really understand how the failing .bzl file becomes part of the test. In any case, consider this error message:

(02:07:26) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/8d833ec1d56a2cd78ffb01dcdaff2cf3/external/io_bazel/third_party/protobuf/3.6.1/protobuf.bzl:270:31: The value 'PACKAGE_NAME' has been removed in favor of 'package_name()', please use the latter (https://docs.bazel.build/versions/master/skylark/lib/native.html#package_name).

The most recent version of the protobuf.bzl file that referred to PACKAGE_NAME on line 270 was over six months ago:
https://github.com/bazelbuild/bazel/commit/04ad00208c2c331e31406d1b07f418a8ebbf1a0a

Why is the test running a new bazel against such old .bzl code? Is that intentional? If so, how long must we preserve compatibility? If not, who knows how to bring the rust test dependencies up to date?

c-parsons:
[Actively working](https://github.com/bazelbuild/rules_rust/pull/248) on updating the dependency now. (It's `rules_rust -> io_bazel_skydoc -> io_bazel`)

c-parsons:
Updating dependency on `io_bazel_skydoc` blocked on https://github.com/bazelbuild/skydoc/issues/233

Will resolve that first then address this.

c-parsons:
Should be fixed. Triggering another Bazel@HEAD + Downstream projects build.
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1172
Will re-open if there are additional issues.

alandonovan:
Thanks, Chris.

PR #248, opened 2019-08-30
Update rules_rust dependency on skydoc

c-parsons:


c-parsons:
Fixes https://github.com/bazelbuild/rules_rust/issues/247

PR #252, opened 2019-09-26
Use C++ environment variables when constructing link command line

hlopko:
This PR makes sure that rustc action also takes declared environment
variables from the C++ toolchain when constructing the action. This way
the C++ toolchain can start using more environment variables without
breaking rust.


hlopko:
Hmm, I see the same failure from my presubmits on master, but buildkite is green.. Something funky is going on?

hlopko:
Ok, I was using old fork, all is fine now, please take a look.

PR #254, opened 2019-10-07
Update setup instructions

laurentlb:
- The commit hash was not compatible with recent versions of Bazel
- Update the dependency on skylib
- Add the load statement for `http_archive`, for easier copy-pasting
- Add an example of use (users shouldn't have to guess the load label)
- Update the version requirement (0.25.0 can't build the code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,19 +65,6 @@ bazel_version(name = "bazel_version")
 ```
 The rules are under active development, as such the lastest commit on the master branch should be used. `master` currently requires Bazel >= 0.26.0.
 
-## Example
-
-Here's an example to build a Hello World:
-
-```python
-load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
-
-rust_binary(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,7 +63,20 @@ rust_repositories()
...
+Here's an example to build a Hello World:
+
+```python
+load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
+
+rust_binary(
+    name = "rs",
+    srcs = ["hello.rs"],
+)
+```
```
damienmg: I am doubtful this is useful. There is link for specific rules that have individual examples that are more explanatory.

damienmg:
Except the example the change LGTM. Maybe we could have the documentation discussion in a separate PR?

README.md
laurentlb: Removed.

When I tried the rule, I didn't find the label to load in the documentation. But now, I see it in the examples.

PR #265, opened 2019-11-04
fix: add all_files to the dummy cc_toolchain

qzmfranklin:
This avoids a problem where rustc_compile_action() tries to use the
`all_files` property of the cc_toolchain.

This problem only occurs after turning on the following Bazel flag:
    --incompatible_enable_cc_toolchain_resolution

The dummy cc toolchain is only used for building for wasm32.


qzmfranklin:
@johnedmonds Since you were the author of the rust-bazel-wasm support, I thought you might also want to take a look.

PR #272, opened 2019-11-20
Rename wasm_transition to relate to what it actually does

### LINKED ISSUE #205: proc-macro crates should build for the host instead of the target
While experimenting with cross compilation using `rules_rust` I noticed something odd: it seems like bazel builds crates marked with the `proc-macro` type for the target platform instead of the host platform. This breaks procedural macros when cross compiling, especially when the target platform does not have a standard library.

I guess that we would want `rules_rust` to convince bazel to compile procedural macro crates and their dependencies for the host platform instead of the target platform, but I do not know enough about Bazel to determine if and how this would be possible.

For now I came up with an ugly but working solution: I added an extra bool field to `rust_library` named `proc_macro_dep`. If this field is set to true, the `--target` argument is omitted from rustc invocations. I can provide the patch in case anyone is interested.

m3rcuriel:
I didn't actually notice that #240 closed #205, so I figure since the transition literally only deals with transitioning on `proc-macro` it should probably be called that.

@acmcarther if the reason for it has nothing to do with #205 please ignore this PR.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -14,7 +14,7 @@
 
 load("@io_bazel_rules_rust//rust:private/rustc.bzl", "CrateInfo", "rustc_compile_action")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
-load("@io_bazel_rules_rust//rust:private/transitions.bzl", "wasm_transition")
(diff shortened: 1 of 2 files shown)


mfarrugi:
lgtm, but I don't know how to rerun the buildkite build.. do you mind rebasing to get it to run again?

mfarrugi:
The build on master was fine, so nevermind that.

PR #285, opened 2019-12-19
Allow deps to be renamed

GregBowyer:


rust/private/rust.bzl
```diff
@@ -275,6 +279,13 @@ _rust_common_attrs = {
             linking a native library.
         """),
     ),
+    "aliases": attr.label_keyed_string_dict(
+        doc = _tidy("""
+            Remap crates to a new name or monikor for linkage to this target
```
damienmg: *monikor* is a typo?, of monitor?

rust/private/rust.bzl
```diff
@@ -275,6 +279,13 @@ _rust_common_attrs = {
...
+
+            These are other `rust_library` targets and will be presented as the new name given
```
damienmg: Unfinished sentence.

rust/private/rust.bzl
GregBowyer: typo, should be moniker

GregBowyer:
> Hi, Thanks for sending out this PR.
> 
> What is the goal of the alias?
> 
> It seems to me one of the goal would be to avoid crate name collision but this PR does not address this case.
> 
> Also could you had an aliasing test case in the examples?

Sorry I dashed this out a little quickly yesterday to pack up just before the holidays.

This is to allow for crate style renames, this is something that occurs very occasionally in crates, the basic idea is to present to a library an external crate under a different name.

The big one where this showed up was in compiling futures at version 0.3.x, with the feature `compat`. This leads to a situation where `futures-util-0.3.x` (a dependency of `futures` as a bill of materials) claims a dep in `Cargo.toml` like so.

```toml
[dependencies.futures_01]
version = "0.1.25"
optional = true
package = "futures"
```

That is, it depends on an earlier version of futures and includes it, from its viewpoint, as the crate `futures_01`

This rename is achieved through changing the name on the `--extern` parameter passed to the compiler.

This is related to https://github.com/google/cargo-raze/pull/123

Without this, its pretty hard to compile crates that perform renames without patching

GregBowyer:
Addressed the comments and added some basic tests

GregBowyer:
Looks like the build error was an unrelated stardoc incompatibility change, I will push that as a seperate PR that this is based on

GregBowyer:
This currently is blocked / deps on !290

GregBowyer:
Rebasing fun, do we want to merge this to unlock cargo-raze?

damienmg:
Thanks for this PR!

PR #290, opened 2020-01-10
Use up to date stardoc to avoid build errors

GregBowyer:
SkyDoc has become Stardoc some time ago, and has a few mistakes in its
older versions regarding workspace paths and contexts. This didnt matter
until bazel 2.0 where these are treated more strictly.

As such we bump to the latest stardoc which allows for building out
rules_rust on bazel 2.0+


GregBowyer:
So this should be fixed, but not sure why the buildkite is stalled out, ideas?

mfarrugi:
Looks like it ran and hit a new (flaky?) error.

GregBowyer:
Ok more fiddling, now we should be fixed. Do we want this merged in

Issue #87, opened 2018-05-25
Make rustfmt accessible (and enforceable?)


mfarrugi:
rustfmt is currently shipped as part of the toolchain tarballs, but not accessible for interactive nor programmatic usage without some finagling.

Some usecases I've run into:

1. rustfmt against generated code
2. enforcing rustfmt as a pre-commit / build step

For enforcement, is it possible for rules_rust to take global configuration, and do this check as an extra-build step? Would that be a reasonable thing to do?

acmcarther:
> rustfmt is currently shipped as part of the toolchain tarballs, but not accessible for interactive nor programmatic usage without some finagling.

Do you know off hand if the packaged rustfmt matches https://github.com/rust-lang-nursery/rustfmt ? I know there are several rustfmt's hanging around and the nightly one is the one users are expected to be using. Either way, exposing the binary should be harmless, though I've historically included rustfmt as a workspace dependency and used the produced binary.

> For enforcement, is it possible for rules_rust to take global configuration, and do this check as an extra-build step? Would that be a reasonable thing to do?

I'm not particularly keen on this idea. It's not conventional to fail a compilation due to formatting in the rust ecosystem, and to my knowledge it's not conventional in the Bazel ecosystem either. 

My mental model of things is that it should be part of your (automated) CI process to enumerate changed files, apply rustfmt to them, and report in the review if there is a notable diff. Seems to me that incremental development (using `ibazel` or like) would be a world of pain if you had to have your formatting just right... unless your editor automatically formatted code. If it does, you still have to make sure that your editor uses the exact same version of rustfmt. [I've griped about this aloud elsewhere](https://github.com/google/cargo-raze/pull/29#issue-175345281)

That said, I suspect you have some context that motivates the suggestion -- or at least describes why including the formatter as a crate is less-than-desirable.

mfarrugi:
My thought was that its in the toolchain and should be available, which
makes user's lives easier. I didn't realize the recommended version wasn't
the same one in the stable tarball; at some point it will be and we can
expose it then.


I think the format check should be precommit, but I buy that bazel doesn't
need to be responsible for this, especially when I think about changes to
the formatting.

colatkinson:
Hey all, not sure if there's continued interest in this issue, but I did a bit of experimentation. From my understanding, it seems that the easiest way to make the `rustfmt` binary accessible is via the `load_arbitrary_tool` macro. A simple (and probably fairly ugly) patch is as follows:

```patch
[...]

PR #291, opened 2020-01-16
Pull/expose rustfmt binaries

colatkinson:
See #87.

* Uses `load_arbitrary_tool()` to download `rustfmt` from its independent package (so no need to pull in the full `rust` archive).
  - Can be invoked via `bazel run @rust_linux_x86_64//:rustfmt -- --check $PWD/greeter.rs`.
  - **TODO**: Is there a way to expose this via a platform-independent workspace?
* Retrieves the `rustfmt` hashes in `fetch_shas.sh`.
  - This is currently pretty ugly due to `rustfmt` and `rustc` being independently versioned.
* Sets `LC_ALL=C` in `fetch_shas.sh`.
  - This was actually causing a different sort order for the hashes on my machine. Probably my fault for not correctly configuring my locales. IMO seems like a quick way to avoid pain for other contributors. Can easily be cherry-picked off, though.

Prebuilt binaries don't seem to be available for FreeBSD, so unfortunately the tool won't be available on that platform.

**TODO**: What's the easiest way to write tests for this?
**TODO**: Can the changes to `fetch_shas.sh` be made nicer?


colatkinson:
@googlebot I signed it!

mfarrugi:
> TODO: Is there a way to expose this via a platform-independent workspace?

I know [`@bazel_tools//tools/cpp:current_cc_toolchain`](https://source.bazel.build/bazel/+/1d205e14e4c069e9199ab71b127c6a6e26a9443b:tools/cpp/BUILD;l=52) exists, but it looks like the aliasing/magic is implemented in Java.

> TODO: What's the easiest way to write tests for this?

You could add a formatting step to one of the codegen rules (ie. protobuf), or write a genrule that invokes rustfmt.

It probably makes sense to make rustfmt available on `rust_toolchain` as part of this so that bindgen, protobuf and any future codegen could make use of it programmatically as well. I think this also simplifies usage from genrules, but not as sure on that.

colatkinson:
Couple of changes:

* `rustfmt` binary is now exposed via the Rust toolchain.
* As suggested, the bindgen rules now fall back on the provided `rustfmt` if available. Tested locally, and it appears to work as intended.
* There's also an explicit test for the binary. A sample rule generates a formatted file from an unformatted one, and `diff` is used to verify that the files differ. Not very comprehensive, so any suggestions on how to beef it up are appreciated.
  - I believe this will fail on BSD, but I haven't yet set up a test environment yet. I haven't been able to determine what the current recommendation is for disabling tests on a platform-specific basis.

kornholi:
Awesome! Is there anything left to do here?

mfarrugi:
lgtm, just need to resolve conflicts.

(sorry for losing track of this, not sure what happened)

colatkinson:
Hey all, I've fixed the conflicts and brought the branch up to date with master. Let me know if there's anything else to do!

PR #295, opened 2020-02-23
Support hyphens in target names

dtolnay:
This PR makes it possible to define and depend on Rust targets whose `name` contains hyphen instead of underscore, primarily intended for case-matching names of third party libraries.

Example:

```bazel
rust_library(
    name = "thiserror-impl",
    srcs = glob(["vendor/thiserror-impl-1.0.11/src/**/*.rs"]),
    crate_type = "proc-macro",
    deps = [
        "//third-party:proc-macro2",
        "//third-party:quote",
        "//third-party:syn",
    ],
)
```

Tested with https://github.com/dtolnay/cxx/pull/37.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -67,7 +67,7 @@ def _determine_lib_name(name, crate_type, toolchain, lib_hash = ""):
               "please file an issue!").format(crate_type))
 
     return "lib{name}-{lib_hash}{extension}".format(
-        name = name,
+        name = name.replace("-", "_"),
         lib_hash = lib_hash,
         extension = extension,
     )
@@ -133,7 +133,7 @@ def _rust_binary_impl(ctx):
     if (toolchain.target_arch == "wasm32"):
         output = ctx.actions.declare_file(ctx.label.name + ".wasm")
     else:
-        output = ctx.actions.declare_file(ctx.label.name)
+        output = ctx.actions.declare_file(ctx.label.name.replace("-", "_"))
 
     return rustc_compile_action(
         ctx = ctx,
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -66,8 +66,9 @@ def _determine_lib_name(name, crate_type, toolchain, lib_hash = ""):
         fail(("Unknown crate_type: {}. If this is a cargo-supported crate type, " +
               "please file an issue!").format(crate_type))
 
+    crate_name = name.replace("-", "_")
     return "lib{name}-{lib_hash}{extension}".format(
-        name = name.replace("-", "_"),
+        name = crate_name,
         lib_hash = lib_hash,
         extension = extension,
     )
@@ -110,12 +111,13 @@ def _rust_library_impl(ctx):
         output_hash,
     )
     rust_lib = ctx.actions.declare_file(rust_lib_name)
+    crate_name = ctx.label.name.replace("-", "_")
 
     return rustc_compile_action(
         ctx = ctx,
         toolchain = toolchain,
         crate_info = CrateInfo(
-            name = ctx.label.name,
+            name = crate_name,
             type = ctx.attr.crate_type,
             root = lib_rs,
             srcs = ctx.files.srcs,
@@ -129,17 +131,18 @@ def _rust_library_impl(ctx):
 
 def _rust_binary_impl(ctx):
     toolchain = find_toolchain(ctx)
+    crate_name = ctx.label.name.replace("-", "_")
 
     if (toolchain.target_arch == "wasm32"):
-        output = ctx.actions.declare_file(ctx.label.name + ".wasm")
+        output = ctx.actions.declare_file(crate_name + ".wasm")
     else:
-        output = ctx.actions.declare_file(ctx.label.name.replace("-", "_"))
+        output = ctx.actions.declare_file(crate_name)
 
     return rustc_compile_action(
         ctx = ctx,
         toolchain = toolchain,
         crate_info = CrateInfo(
-            name = ctx.label.name,
+            name = crate_name,
             type = "bin",
             root = _crate_root_src(ctx, "main.rs"),
             srcs = ctx.files.srcs,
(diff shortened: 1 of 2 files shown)


rust/private/rust.bzl
```diff
@@ -133,7 +133,7 @@ def _rust_binary_impl(ctx):
     if (toolchain.target_arch == "wasm32"):
         output = ctx.actions.declare_file(ctx.label.name + ".wasm")
     else:
-        output = ctx.actions.declare_file(ctx.label.name)
+        output = ctx.actions.declare_file(ctx.label.name.replace("-", "_"))
```
mfarrugi: Is this correct? I believe binary names somewhat commonly contain hyphens

rust/private/rustc.bzl
```diff
@@ -405,7 +405,7 @@ def add_crate_link_flags(args, dep_info):
     )
 
 def _crate_to_link_flag(crate_info):
-    return ["--extern", "{}={}".format(crate_info.name, crate_info.dep.output.path)]
+    return ["--extern", "{}={}".format(crate_info.name.replace("-", "_"), crate_info.dep.output.path)]
```
mfarrugi: If this replacement always happens, it should be done when the CrateInfo struct is created

rust/private/rust.bzl
mfarrugi: Also suspect that wasm scheme is different

rust/private/rustc.bzl
mfarrugi: Maybe this can happen one time, on L105?

rust/private/rust.bzl
dtolnay: > I believe binary names somewhat commonly contain hyphens

That may be the case, but previously rust_binary with a hyphen didn't work at all:

```bazel
rust_binary(
    name = "test-hyphen",
    srcs = ["main.rs"],
)
```

```console
$ bazel run test-hyphen --verbose_failures
INFO: Writing tracer profile to '/home/david/.cache/bazel/_bazel_david/ebce1d0721fb68dda9c70c0dd1405803/command.profile.gz'
INFO: Analyzed target //:test-hyphen (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /git/cxx/BUILD:3:1: error executing shell command: '/bin/bash -c CARGO_MANIFEST_DIR=$(pwd)/BUIL external/rust_1_42_beta/bin/rustc "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd  main.rs --crate-name=test-hyphen --crate-type=bin --codegen=met...' failed (Exit 1) bash failed: error executing command 
  (cd /home/david/.cache/bazel/_bazel_david/ebce1d0721fb68dda9c70c0dd1405803/sandbox/linux-sandbox/1167/execroot/__main__ && \
  exec env - \
    CARGO_CFG_TARGET_ARCH=x86_64 \
    CARGO_CFG_TARGET_OS=linux \
    CARGO_PKG_AUTHORS='' \
    CARGO_PKG_DESCRIPTION='' \
    CARGO_PKG_HOMEPAGE='' \
    CARGO_PKG_NAME=test-hyphen \
    CARGO_PKG_VERSION=0.0.0 \
    CARGO_PKG_VERSION_MAJOR=0 \
    CARGO_PKG_VERSION_MINOR=0 \
    CARGO_PKG_VERSION_PATCH=0 \
    CARGO_PKG_VERSION_PRE='' \
  /bin/bash -c 'CARGO_MANIFEST_DIR=$(pwd)/BUIL external/rust_1_42_beta/bin/rustc "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd' '' main.rs '--crate-name=test-hyphen' '--crate-type=bin' '--codegen=metadata=' '--out-dir=bazel-out/k8-fastbuild/bin' '--codegen=extra-filename=' '--codegen=opt-level=0' '--codegen=debuginfo=0' '--emit=dep-info,link' '--color=always' '--target=x86_64-unknown-linux-gnu' -L external/rust_1_42_beta/lib/rustlib/x86_64-unknown-linux-gnu/lib '--edition=2018' '--codegen=linker=/usr/bin/gcc' --codegen 'link-args=-fuse-ld=gold -Wl,-no-as-needed -Wl,-z,relro,-z,now -B/usr/bin -pass-exit-codes -lstdc++ -lm')
Execution platform: @local_config_platform//:host

error: invalid character `-` in crate name: `test-hyphen`

Target //:test-hyphen failed to build
INFO: Elapsed time: 0.140s, Critical Path: 0.09s
INFO: 0 processes.
```

After this PR rustc will see it as if the user had written the name with an underscore, but the BUILD file and bazel command line will be allowed to use hyphen to match the actual name upstream on crates.io.

I believe it would be a separate feature request to pass the right `-o` flag to rustc to write out the binary with an underscore in the filename.

rust/private/rustc.bzl
dtolnay: Good call, updated. That means we no longer need to touch rustc.bzl for this change, only rust.bzl.

rust/private/rust.bzl
```diff
@@ -110,12 +111,13 @@ def _rust_library_impl(ctx):
         output_hash,
     )
     rust_lib = ctx.actions.declare_file(rust_lib_name)
+    crate_name = ctx.label.name.replace("-", "_")
```
mfarrugi: can remove one more by passing crate_name to L108

rust/private/rust.bzl
mfarrugi: It's strange behavior in rustc I guess, fair enough.

rust/private/rust.bzl
dtolnay: One of them is `ctx.attr.name`, the other is `ctx.label.name`. Are those interchangeable?

mfarrugi: I'm 99% sure they are, but not entirely.. I think the difference is that ctx.label is the full label

PR #301, opened 2020-04-11
Add cfg = proc_macro_host_transition to all rules

### LINKED ISSUE #300: Multiple proto generation/linking causing compilation issue
I've tried to put a minimally reproducible example of the problem I was running into [here](https://github.com/dfreese/rust_bazel_proto_example)

When going from [commit 5c0e90a](https://github.com/bazelbuild/rules_rust/commit/5c0e90a24e9b9e8ab81b183de4d14eb95fb86747) to [commit 7cde1e46](https://github.com/bazelbuild/rules_rust/commit/7cde1e46e9e5a97eb2650ee70a81ec8f9578e3e7), via  #240, I was getting compilation errors, of
```
the trait bound `server::EchoImpl: echo_rust_grpc_proto::Echo` is not satisfied
```
when importing a server implementation from another file that's trying to create a gRPC/protobuf server.  This is using the rust_library, rust_executable, proto_library, and rust_grpc_library rules. 

The compilation between those two different commits seems nearly identical, but in the failing case, the `SUBCOMMAND: # //proto:echo_proto` seems to be run twice, and then linked into the executable twice.

I'm new to rust. Is this something incorrect in that workspace setup, or is something else going wrong?

dfreese:
The rust_library rule specifies a cfg option.  This performs transitions
in order to support WASM.  Other rules, such as rust_binary, do not have
this configuration, which causes Bazel to build and link some libraries
into the final binary twice, resulting in compilation errors.  This is
detailed in pull request #294 and issue #300.  The issue was introduced
in commit 7cde1e46.

This adds the example from @colin353 in PR #294 as an example that would
otherwise fail.

Fixes #300

### DIFF
--- a/examples/proto/BUILD
+++ b/examples/proto/BUILD
@@ -1,8 +1,31 @@
 load("@rules_proto//proto:defs.bzl", "proto_library")
+load("@io_bazel_rules_rust//proto:proto.bzl", "rust_proto_library")
+load("@io_bazel_rules_rust//rust:rust.bzl", "rust_library")
+load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
 
 package(default_visibility = ["//proto:__subpackages__"])
 
 proto_library(
     name = "common",
     srcs = ["common.proto"],
+)
+
+rust_proto_library(
+    name = "common_proto_rust",
+    deps = [":common"],
+)
+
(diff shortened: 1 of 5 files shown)


colin353:
@dfreese thanks for writing this! Looks great!

PR #303, opened 2020-04-15
Add proc macro example, propagate crate-type through rustdoc

smklein:
- Added a no-op proc macro to "examples".
- While adding this macro, the following error message came from
"rust_doc":

    warning: Trying to document proc macro crate without passing
    '--crate-type proc-macro to rustdoc

  To resolve this issue, the rust_doc rule was modified to propagate
  the underlying crate type.


mfarrugi:
Thanks for the fix!

PR #307, opened 2020-04-19
Allow specifying rust versions

kragniz:


mfarrugi:
Before I merge this, can you share what your use-case for this was? It looks like you want to be able to use an older version of rustc, is that right?

The support for beta and nightly is also somewhat disingenuous, since it would only work for the handful of shas listed in `util/fetch_shas_*.txt`

kragniz:
@mfarrugi I wanted to easily build with a newer rust version (1.42.0) than the default. I didn't actually need the nightly or beta versions but it looked like it would be easy to add support for that as well.

mfarrugi:
From what I understand, this only lets you choose between versions that we have sha's for because of Bazel's sha requirement, no? 

I have no problem with this PR, but it doesn't look like it solves your problem.

kragniz:
@mfarrugi if the shas aren't in the list it just doesn't enforce them: https://github.com/bazelbuild/rules_rust/blob/0deef6dd8180cd3bc610878558bb26921b4e8de1/rust/repositories.bzl#L266

Maybe the docs should say that changing the value to one that's not included in the shipped list will make your build non-hermetic.

mfarrugi:
So you've tested this works? The [docs](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download_and_extract) imply it might error on empty string rather than None.

Feel free to update the doc, although I'm not too paranoid about the rust tool distros.

kragniz:
The default is an empty string, right?

>It is optional to make development easier but should be set before shipping.

But yeah, I'm running my local builds with this patch and setting the version to 1.42.0.

mfarrugi:
I don't think the default argument ever makes it through before this change, but didnt look into it.

Thanks!

PR #309, opened 2020-04-20
Remove transition from binary/test/bench rules, add to grpc/proto

### LINKED ISSUE #305: rust_binary does not produce link in bazel-bin
My commit a1d89361 caused rust_binary to no longer link into bazel-bin.  The binary still is built, and can be run by bazel run, but the binary no longer appears in bazel-bin, just bazel-out.  This doesn't appear to be related to depending on a proc-macro, as the following example has this behavior.

```bazel
# //BUILD

rust_binary(
    name = "check",
    srcs = ["main.rs"],
)
```

```rust
// //main.rs
fn main() {
    print!("hello world\n");
}
```

Still trying to track down the exact cause of it here, but wanted to flag it in case anyone has noticed this behavior, or similar, and if it's more clear to others as to why this would happen.

Also, I'm not sure what the revert policy is for this project, but I'm perfectly fine reverting it until a fix is found.  Let me know.

dfreese:
This mostly reverts commit a1d8936, removing the
proc_macro_host_transition from most rules.  This fixes #305, by no
longer having a transition on rust_binary.  It still solves #300 by
adding this transition to rust_proto_library and rust_grpc_library.
Those rules are meant to mirror rust_library, so they should share the
same cfg values.  Sharing the same configuration means that
rust_proto_library or rust_grpc_library can be dependencies to
rust_library or rust_binary and not cause another configuration of the
library to be generated.

The example introduced by #301 is moved in this commit into a subfolder
to provide a more clear directory structure.


damienmg:
Thanks for fixing that, awesome!

PR #312, opened 2020-04-24
add flag to allow putting out binary whenever

Shikugawa:
We had an opportunity to put out binary without specifying `crate_type` as `bin`. 
In this case, we should add configurable crate_type and a flag to put out binary whenever.
With this patch, we can tackle this situation.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -137,12 +137,14 @@ def _rust_binary_impl(ctx):
     else:
         output = ctx.actions.declare_file(crate_name)
 
+    bin_crate_type = getattr(ctx.attr, "bin_crate_type")
+
     return rustc_compile_action(
         ctx = ctx,
         toolchain = toolchain,
         crate_info = CrateInfo(
             name = crate_name,
-            type = "bin",
+            type = bin_crate_type,
             root = _crate_root_src(ctx, "main.rs"),
             srcs = ctx.files.srcs,
             deps = ctx.attr.deps,
@@ -435,6 +437,10 @@ _rust_binary_attrs = {
         cfg = "host",
         allow_single_file = True,
     ),
+    "bin_crate_type": attr.string(
+        default = "bin",
+    ),
+    "always_out_binary": attr.bool(),
 }
 
 rust_binary = rule(
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -137,14 +137,14 @@ def _rust_binary_impl(ctx):
     else:
         output = ctx.actions.declare_file(crate_name)
 
-    bin_crate_type = getattr(ctx.attr, "bin_crate_type")
+    crate_type = getattr(ctx.attr, "crate_type")
 
     return rustc_compile_action(
         ctx = ctx,
         toolchain = toolchain,
         crate_info = CrateInfo(
             name = crate_name,
-            type = bin_crate_type,
+            type = crate_type,
             root = _crate_root_src(ctx, "main.rs"),
             srcs = ctx.files.srcs,
             deps = ctx.attr.deps,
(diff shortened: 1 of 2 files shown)


damienmg:
Hi,

Thanks for contributing. Could you expand a bit more on the purpose of this PR? Maybe by giving an example?

Also the build is failing, you should be able to reproduce with `bazel build @examples//...`.

Shikugawa:
@damienmg Thanks. This functionality is needed on this PR. 
https://github.com/envoyproxy/envoy-wasm/pull/488

I'd like to build Rust code like this. 
https://github.com/Shikugawa/envoy-wasm-rust-playground/blob/master/example/src/lib.rs

This code is build with this build config. 
https://github.com/Shikugawa/envoy-wasm-rust-playground/blob/master/example/Cargo.toml

We want to put out binary but not want to build with `--cargo-type=lib` but `--cargo-type=cdylib`. In the current implementation, we can't put out binary without using `--cargo-type=bin`. This is a critical problem for me so that patched this.

damienmg:
Ok I am not really happy with the interface this PR offer but I cannot think of something better, maybe @mfarrugi or @acmcarther have a better opinion.

acmcarther:
I don't feel particularly strongly about this. It does seem to solve a preexisting problem, though it seems this might be wasm-specific? 

Some ideas of slightly preferred interfaces:
1. out_binary=true should be implicit for a rust_binary rule. Can we just infer this?
2. Its hard to infer the purpose of crate_type on rust_binary if you're not steeped in the wasm context. In fact, I'm not even really sure what this does. Does it emit both a library and a binary?

Reading the relevant [Cargo.toml](https://github.com/Shikugawa/envoy-wasm-rust-playground/blob/master/example/Cargo.toml) it's not obvious to me that this is supposed to produce a "binary" at all. Does rust_library work for this.

I suspect I'm missing something critical here.

mfarrugi:
> It's also hard to infer the purpose of crate_type on rust_binary if you're not steeped in the wasm context. In fact, I'm not even really sure what this does. Does it emit both a library and a binary?
+1

I have no idea why it having rust_library build the cdylib doesn't work, and I can't guess what cdylib means for wasm. Does executable mean something special for wasm?

Shikugawa:
@acmcarther @mfarrugi I think that this is not "binary" but "library" because we emit wasm module. But it seems that `rust_library` doesn't have this functionality, to emit wasm module. So we may have to add this to `rust_library`, shouldn't we? WDYT? 

> it seems this might be wasm-specific?

I think that this is not so, but I have no idea to use this functionality on a normal binary.

> I have no idea why it having rust_library build the cdylib doesn't work, and I can't guess what cdylib means for wasm. Does executable mean something special for wasm?

I also have no idea about this. Maybe cc. @PiotrSikora have some idea.

PiotrSikora:
See previous discussion: https://github.com/bazelbuild/rules_rust/pull/240#discussion_r337676010

`rules_rust` already set the precedent and has `rust_{proto,grpc,wasm_bindgen}_library` targets, so I think that we should add dedicated `rust_wasm_library` (`binary` vs `library` doesn't matter for Wasm, AFAIK).

mfarrugi:
I can't really opine, don't know anything about wasm. To me this stuff looks a little grody, but it's not a lot of code.

Shikugawa:
We may ought to implement `rust_wasm_binary` which spec has to put out code anytime. But It can be good a bit to rename `out_binary` to `wasm_mode`, which put out wasm module anytime if we set this `True`. `out_binary` may cause confusion.

damienmg:
Sorry for taking so long to come back to this one. I don't see strong opinion either way. `rust_wasm_binary` is a good idea but in between I don't see this PR as harmful so LGTM.

PR #313, opened 2020-04-30
When inferring crate root, take the match with shortest filepath

dtolnay:
I had a situation where a rust_library target contained both `src/cratename.rs` (the intended crate root) and `src/some/coincidental/inner/cratename.rs`. It seems fine to prefer the shortest one in this case, rather than a random one.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -97,8 +97,12 @@ def _shortest_src_with_basename(srcs, basename):
     """
     Finds the shortest among the paths in srcs that match the desired basename.
     """
-    files = sorted([f for f in srcs if f.basename == basename], lambda f: len(f.dirname)])
-    return files[0] if len(files) else None
+    shortest = None
+    for f in srcs:
+        if f.basename == basename:
+            if not shortest or len(f.dirname) < len(shortest.dirname):
+                shortest = f
+    return shortest
 
 def _rust_library_impl(ctx):
     # Find lib.rs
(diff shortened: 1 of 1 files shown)


rust/private/rust.bzl
```diff
@@ -81,19 +81,29 @@ def _get_edition(ctx, toolchain):
...
+def _shortest_src_with_basename(srcs, basename):
+    """
+    Finds the shortest among the paths in srcs that match the desired basename.
+    """
+    shortest = None
+    for f in srcs:
+        if f.basename == basename:
+            if not shortest or len(f.dirname) < len(shortest.dirname):
+                shortest = f
+    return shortest
```
damienmg, line 105:
```suggestion
    files = sorted([f for f in srcs if f.basename = basename], lambda f: len(f.dirname)])
    return files[0] if len(files) else None
```

dtolnay:
It looks like lambdas aren't supported:

```console
rust.bzl:100:65: 'lambda' not supported, declare a function instead
rust.bzl:100:65: syntax error at 'lambda': expected expression
rust.bzl:100:90: indentation error
rust.bzl:100:65: contains syntax error(s)
rust.bzl:100:79: name 'f' is not defined
```

Inner functions aren't either, so if we are using `sorted`, it would need to be with a separate top level function to do the `len(f.dirname)`. By that point I find that the original looping implementation is clearer.

damienmg:
Thanks for your contribution!

The spec says lambda are supported :/ I though they added them recently. Agreed with your assessment of your initial implementation being clearer.

Issue #266, opened 2019-11-07
rust_test does not propagate data files from the crate


kornholi:
Data files from the target crate are not collected as part of the test, e.g. with
```bzl
rust_binary(
    name = "foo",
    srcs = ["main.rs"],
    data = ["bar.txt"],
)

rust_test(
    name = "tests",
    crate = ":foo",
)
```

`bar.txt` is not available at either compile or runtime time for the tests.

There are two parts to this:
- `rustc_compile_action` only includes `data` files from the current rule as part of the compilation. Adding a `data = ["bar.txt"]` to the test fixes this as expected.
- `ctx.runfiles(collect_data=True)` does not know about the `crate` attr, and won't use it to collect transitive runfiles. Adding a `deps = [":foo"]` to the test fixes this as expected.

Somewhat related to #222.

djmarcin:
rustc_env is another that does not propagate from rust_library -> rust_test and requires duplication if the env vars affect compilation (e.g. `env!` macros)

PR #314, opened 2020-05-05
rustc: add arbitrary environment variables

tommilligan:
Rules that call `rustc` setup environment variables in line with `cargo`'s behaviour (see the [documentation here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates)). However they don't allow passing in additional environment variables that may be required at compile time, such as the [`PROTOC` variable in `prost-build`](https://github.com/danburkert/prost/blob/2de785aca480779ab0d4442d4bc91a696e3e6c89/prost-build/src/lib.rs#L689).

This PR adds the `rustc_env` attribute to the common attribute set, allowing it to be set in the following rules:
- `rust_benchmark`
- `rust_binary`
- `rust_library`
- `rust_test`

Variables are provided in a string dictionary, and will be merged into the generated environment overriding existing values.

### DIFF
--- a/proto/proto.bzl
+++ b/proto/proto.bzl
@@ -171,6 +171,7 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, grpc,
             aliases = {},
             output = rust_lib,
             edition = proto_toolchain.edition,
+            rustc_env = {},
         ),
         output_hash = output_hash,
     )
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -288,6 +288,11 @@ _rust_common_attrs = {
             These are other `rust_library` targets and will be presented as the new name given.
         """),
     ),
+    "rustc_env": attr.string_dict(
+        doc = _tidy("""
+            Dictionary of additional `"key": "value"` environment variables to set for rustc.
+        """),
+    ),
     "crate_features": attr.string_list(
         doc = _tidy("""
             List of features to enable for this crate.
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -124,6 +124,7 @@ def _rust_library_impl(ctx):
             aliases = ctx.attr.aliases,
             output = rust_lib,
             edition = _get_edition(ctx, toolchain),
+            rustc_env = ctx.attr.rustc_env,
         ),
         output_hash = output_hash,
     )
@@ -149,6 +150,7 @@ def _rust_binary_impl(ctx):
             aliases = ctx.attr.aliases,
             output = output,
             edition = _get_edition(ctx, toolchain),
+            rustc_env = ctx.attr.rustc_env,
         ),
     )
 
@@ -175,6 +177,7 @@ def _rust_test_common(ctx, test_binary):
             aliases = ctx.attr.aliases,
             output = test_binary,
             edition = crate.edition,
+            rustc_env = ctx.attr.rustc_env,
         )
     elif len(ctx.attr.deps) == 1 and len(ctx.files.srcs) == 0:
         dep = ctx.attr.deps[0].label
@@ -194,6 +197,7 @@ def _rust_test_common(ctx, test_binary):
             aliases = ctx.attr.aliases,
             output = test_binary,
             edition = _get_edition(ctx, toolchain),
+            rustc_env = ctx.attr.rustc_env,
         )
 
     return rustc_compile_action(
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -330,6 +330,10 @@ def rustc_compile_action(
     else:
         formatted_version = ""
 
+    # Update environment with user provided variables.
+    if hasattr(ctx.attr, "rustc_env"):
+        env.update(ctx.attr.rustc_env)
+
     ctx.actions.run_shell(
         command = command,
         inputs = compile_inputs,
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -35,6 +35,7 @@ CrateInfo = provider(
         "aliases": "Dict[Label, String]: Renamed and aliased crates",
         "output": "File: The output File that will be produced, depends on crate type.",
         "edition": "str: The edition of this crate.",
+        "rustc_env": """Dict[String, String]: Additional `"key": "value"` environment variables to set for rustc.""",
     },
 )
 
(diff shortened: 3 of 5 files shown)


damienmg:
Hi @tommilligan,

Thanks for your contribution!

Code-wise, this PR looks ok. It is missing the corresponding entry in the doc, but I think it needs to be left away if we move forward with the doc (keeping the thing as undocumented for now).

It seems like what you are trying to achieve can be done with `--action_env` command line flag of Bazel. If you already knew about `--action_env`, could you explain the rationale for adding it vs using `--action_env`?

tommilligan:
Sure - for a concrete example, I've removed the `rustc_env` setting from my test case:
```diff
diff --git a/test/build_env/BUILD b/test/build_env/BUILD
index 55496c9..5d14626 100644
--- a/test/build_env/BUILD
+++ b/test/build_env/BUILD
@@ -14,7 +14,4 @@ rust_test(
 rust_test(
     name = "arbitrary_env_test",
     srcs = ["tests/arbitrary_env.rs"],
-    rustc_env = {
-        "USER_DEFINED_KEY": "USER_DEFINED_VALUE",
-    },
 )
```

I then ran the same test using `--action_env` as you suggested. The compiler cannot access the environment variable:
```bash
$ bazel test --action_env=USER_DEFINED_KEY=USER_DEFINED_VALUE //test/build_env:arbitrary_env_test
DEBUG: <snip>
INFO: Build option --action_env has changed, discarding analysis cache.
INFO: Analyzed target //test/build_env:arbitrary_env_test (3 packages loaded, 439 targets configured).
INFO: Found 1 test target...
ERROR: /home/tom/Documents/personal/rules_rust/test/build_env/BUILD:14:1: error executing shell command: '/bin/bash -c CARGO_MANIFEST_DIR=$(pwd)/test/build_env external/rust_linux_x86_64/bin/rustc "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd  test/build_env/tests/arbitrary_env.rs --crate-name...' failed (Exit 1) bash failed: error executing command /bin/bash -c 'CARGO_MANIFEST_DIR=$(pwd)/test/build_env external/rust_linux_x86_64/bin/rustc "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd' '' test/build_env/tests/arbitrary_env.rs ... (remaining 16 argument(s) skipped)

error: environment variable `USER_DEFINED_KEY` not defined
 --> test/build_env/tests/arbitrary_env.rs:3:18
  |
3 |     let actual = env!("USER_DEFINED_KEY");
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^

Target //test/build_env:arbitrary_env_test failed to build
INFO: Elapsed time: 0.615s, Critical Path: 0.15s
INFO: 0 processes.
//test/build_env:arbitrary_env_test                             FAILED TO BUILD

```

This is due to the fact that in `rustc.bzl` we [explicitly set the `env` attribute](https://github.com/bazelbuild/rules_rust/blob/6835a3c8ed1dcd67040cccd603ff3daf611ce41c/rust/private/rustc.bzl#L337). As [documented here](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run_shell), `env` sets the environment to exactly the values specified (it doesn't update the default shell environment).

This has been discussed in a couple of places:
- https://github.com/bazelbuild/bazel/issues/5980#issuecomment-461135762
- https://github.com/bazelbuild/bazel/issues/7364

A different way to workaround this constraint would be to get the default shell environment (not sure if this is possible), update it with the cargo specific environment variables, then set this as the shell environment. This would mimic the behaviour of `cargo`, which also passes through variables from the outer environment.

If this is possible I'd actually prefer it, but it feels like it could be a breaking change, whereas a `rustc_env` attribute would be completely opt-in.

dfreese:
Taking a page from local_defines in cc_library, perhaps it would be worth indicating if the listed env vars propagate to a dependent rust_library, and if make variable expansion applies or not.

A way to do that could be naming it `local_env` or similar.

https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library_args

Had been dealing with prost-build myself, so this is cool to see!

tommilligan:
@dfreese thanks for the suggestion, if others agree `local_env` is a more appropriate name I'm happy to amend it.

Currently this implementation:
- doesn't propagate variables to dependencies
- doesn't support Make variable expansion

which I'm happy to add to the docstring.

dfreese:
@tommilligan sounds good.  Thank for working on this!

@smklein wanted to point this out since you had been looking at cargo raze's handling of build.rs files.

kornholi:
I've often wanted this myself but usually find better solutions in the end. With prost-build, we don't actually want it to call `protoc`, see my comment here: https://github.com/danburkert/prost/pull/309#issuecomment-624743491

I think `rustc_env` should be added to `CrateInfo`. It's missing a few other attributes as is (e.g. `crate_features`), but this leads to issues similar to #266.

tommilligan:
@kornholi could you elaborate on why this should be added to `CrateInfo`? I'm reasonably new to the `bazel` ecosystem in general, but it looks like `CrateInfo` serves two purposes:
1. to aggregate information about a crate **and** it's dependencies
1. to hold information that is required **after** compilation, when a consumer uses the compiled artefact

I don't think 1. is relevant, as the environment is only relevant to a specific crate's compilation, and does not propagate up or down the build tree.

It's possible that 2. would be useful to inspect the environment when a crate was compiled, but I think in this case the whole of the generated `env` should be stored on `CrateInfo`, not `rustc_env` specifically. This feels like a different issue/PR concern to me.

kornholi:
You have the right idea, but there's already some code that (incorrectly) assumes `CrateInfo` captures everything, e.g. https://github.com/bazelbuild/rules_rust/blob/master/rust/private/rust.bzl#L176-L188. Since it's been broken before your changes, let's not worry about it for now.

dfreese:
Looking at it again, the name and behavior of rustc_env is consistent with the rust_flags, so this is good with me.

damienmg:
I think @kornholi is correct, since we might use the CrateInfo for more than just artifact analysis (e.g. for building compilation database). I would prefer you move it there before we merge or we might forget about it and get bugged by it in the future.

tommilligan:
@damienmg updated. `rustc.bzl` now pulls the values for `rustc_env` from the `CrateInfo` object instead of `ctx.attr`.

As with `aliases`, this field is always set to the empty dictionary in `proto.bzl`, although there is no reason why it couldn't also be exposed there.

damienmg:
I think all concerns have been resolved for this PR, so I will go ahead and merge it.

Thanks a lot @tommilligan for this!

PR #318, opened 2020-05-13
Add "build_script_output" parsing to common attributes.

smklein:
Support parsing the stdout of build scripts, and add
a test which validates this behavior.

Currently only the following build script invocations
are processed:
  cargo:rustc-env=...
  cargo:rustc-cfg=...

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -330,6 +330,20 @@ _rust_common_attrs = {
             ".tar.gz",
         ],
     ),
+    "build_script_output": attr.label(
+        doc = _tidy("""
+            An optional file containing the stdout results of the a build script.
+
+            These typically take the form of multiple lines formatted with
+            the syntax "cargo:COMMAND=VALUE"
+
+            See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for additional contet.
+
+            WARNING: Currently only "cargo:rustc-env" and "cargo:rustc-cfg" are
+                     supported.
+        """),
+        allow_single_file = True,
+    ),
     "_cc_toolchain": attr.label(default = "@bazel_tools//tools/cpp:current_cc_toolchain"),
 }
 
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -275,6 +275,33 @@ def rustc_compile_action(
     args.add_all(getattr(ctx.attr, "rustc_flags", []))
     add_edition_flags(args, crate_info)
 
+    # Process Build Script Output
+    #
+    # The output of the build script is parsed in intermediate files.
+    # Since Bazel does not enable extracting the contents of intermediate files
+    # during rule execution (we'd prefer to load the values into Starlark variables),
+    # these actions produce snippets of bash which load the values during the
+    # invocation of rustc.
+
+    # Input: cargo:rustc-env=...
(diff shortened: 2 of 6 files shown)


smklein:
I'm very open to feedback on this one! I have a corresponding cargo-raze patch which is also ready to emit these files from a real build.rs file if "gen_buildrs" is supplied -- basically, just adding an output parameter which captures stdout, as described by https://doc.rust-lang.org/cargo/reference/build-scripts.html

I think the "test/build_script" test I added really shows the case where I care about these values -- many crates I've encountered rely on similar cargo declarations from a build.rs script.

Super happy to wildly change the implementation / add more tests - as long as we can manage to keep something like the "test/build_script" cases compiling and working.

rust/private/rust.bzl
```diff
@@ -330,6 +330,20 @@ _rust_common_attrs = {
...
         ],
     ),
+    "build_script_output": attr.label(
+        doc = _tidy("""
+            An optional file containing the stdout results of the a build script.
+
+            These typically take the form of multiple lines formatted with
+            the syntax "cargo:COMMAND=VALUE"
+
+            See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for additional contet.
```
acmcarther, line 340: s/contet/content

damienmg:
Hi,

I have an alternative proposal.

At some point in the past where I tried to get a tighter integration with cargo, I implemented a [build script rule](https://github.com/damienmg/rules_rust/commit/9e272e7b2d5848991737bf234a412964b5f730a0) that was executing this build script and generating that output. The main thing was the output was wrapped inside a provider, that avoid having to pass the file to all of them.

Would reviving that change solves your issue? What would be missing otherwise?

smklein:
> Hi,
> 
> I have an alternative proposal.
> 
> At some point in the past where I tried to get a tighter integration with cargo, I implemented a [build script rule](https://github.com/damienmg/rules_rust/commit/9e272e7b2d5848991737bf234a412964b5f730a0) that was executing this build script and generating that output. The main thing was the output was wrapped inside a provider, that avoid having to pass the file to all of them.
> 
> Would reviving that change solves your issue? What would be missing otherwise?

Ah, would this replace https://github.com/google/cargo-raze/blob/master/impl/src/templates/partials/build_script.template (the build script executor which exists in cargo-raze)?

I definitely prefer a rust-first approach to parsing these values, rather than the bash snippets my PR currently uses! I leaned on capturing the output in a file to bridge the gap between raze/rules_rust, but storing the output in a provider seems like a natural fit.

That patch does look a little old though - is there anything I can do to help rebase?

damienmg:
> Ah, would this replace https://github.com/google/cargo-raze/blob/master/impl/src/templates/partials/build_script.template (the build script executor which exists in cargo-raze)?

Correct, I was actually trying totally remove the need for cargo-raze, but bazel wasn't up to the task at the time (needs to implements feature propagation).

> I definitely prefer a rust-first approach to parsing these values, rather than the bash snippets my PR currently uses! I leaned on capturing the output in a file to bridge the gap between raze/rules_rust, but storing the output in a provider seems like a natural fit.
> 
> That patch does look a little old though - is there anything I can do to help rebase?

I will look at what it needs to rebase and let you know but I don't expect the rebase to be too hard so I might be able to do that tonight.

damienmg:
Closing this one in favor for #320

PR #319, opened 2020-05-16
Allow edition to be specified in rust_repository_set

dfreese:
rust_repository_set is a very convenient function to specify all of the
necessary http_repository rules and workspace setup calls.  It however,
forces the edition of rust to be 2015 in the calls of
BUILD_for_rust_toolchain from _load_rust_stdlib.

This change does the following:
 * Adds an edition attr to rust_toolchain_repository (default: 2015)
 * Adds an edition parameter to rust_repository_set (default: None)
 * Adds documentation for rustfmt_version for both
 * Throws in some buildifer fixes of function definitions

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -5,7 +5,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 DEFAULT_TOOLCHAIN_NAME_PREFIX = "toolchain_for"
 
-def rust_repositories(version="1.39.0", iso_date=None, rustfmt_version="1.4.8"):
+def rust_repositories(version = "1.39.0", iso_date = None, rustfmt_version = "1.4.8"):
     """Emits a default set of toolchains for Linux, OSX, and Freebsd
 
     Skip this macro and call the `rust_repository_set` macros directly if you need a compiler for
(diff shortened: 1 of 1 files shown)


rust/repositories.bzl
```diff
@@ -396,6 +397,8 @@ Args:
   toolchain_name_prefix: The per-target prefix expected for the rust_toolchain declarations
   version: The version of the tool among "nightly", "beta', or an exact version.
   iso_date: The date of the tool (or None, if the version is a specific version).
+  rustfmt_version: The version of rustfmt to be associated with the toolchain.
```
damienmg, line 400: nit: Can the order of the args match the attrs list?

rust/repositories.bzl
```diff
@@ -5,7 +5,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 DEFAULT_TOOLCHAIN_NAME_PREFIX = "toolchain_for"
 
-def rust_repositories(version="1.39.0", iso_date=None, rustfmt_version="1.4.8"):
+def rust_repositories(version = "1.39.0", iso_date = None, rustfmt_version = "1.4.8"):
```
damienmg, line 8: Is that a change by the starlark formatter?

rust/repositories.bzl
dfreese: Done.

rust/repositories.bzl
dfreese: Yes it is.  To confirm, I checked out master and ran buildifier on this file, and this was the resulting change.

damienmg: Thanks!

damienmg:
LGTM.

PR #320, opened 2020-05-18
Add a cargo_build_script_run rule

damienmg:
This rule allows to run a build.rs file (cf. https://doc.rust-lang.org/cargo/reference/build-scripts.html)
get its output (environment variable, rust flags and generated flags) and pass it to
`rust_binary` and `rust_library` in order to build crate with build script.


damienmg:
cc @smklein 

I finally got time to update my build script PR.

Would that works for you? We could also add an additional rule to provide the various file directly. If we do so I would be up to deprecate also the out_dir attribute for rules_rust, as they would conflict (and the logic there is bit hairy to be honest).

cargo/cargo_build_script.bzl
```diff
@@ -0,0 +1,108 @@
...
+            allow_files = True,
+            default = Label("//cargo/cargo_build_script_runner:cargo_build_script_runner"),
+            cfg = "host",
+        ),
+    },
+    toolchains = [
+        "@io_bazel_rules_rust//rust:toolchain",
+    ],
+    doc = """
+Execute a rust build script to generate build attributes.
```
smklein, line 61: Thanks for the docs; this helps clarify things a lot.

Do you have a corresponding cargo raze patch already? If not, I'm very happy to start making templates that generate BUILD files in this format.

examples/cargo/BUILD
```diff
@@ -0,0 +1,33 @@
...
+)
+
+rust_binary(
+    name = "build_script",
+    srcs = ["build_script.rs"],
+)
+
+cargo_build_script_run(
+    name = "build_script_run",
+    srcs = ["lib.rs"],
```
smklein, line 19: Is "lib.rs" actually used by the build script anywhere? Seems kinda like a superfluous dependency to me.

cargo/cargo_build_script.bzl
```diff
@@ -0,0 +1,108 @@
...
+Then you want to use the build script in the following:
+
+`hello_lib/BUILD`:
+```python
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary", "rust_library")
+load("@io_bazel_rules_rust//cargo:cargo_build_script.bzl", "cargo_build_script")
+
+rust_binary(
```
smklein, line 85: What are your thoughts on merging the "cargo_build_script_run" and "rust_binary" into one target?

```
cargo_build_script_run(
  name = "build_script",
  srcs = ["build.rs"]
)

rust_library(
  ...
  deps = [":build_script"]
)
```

Couple of reasons why I'm wondering about this:
- Does the current usage of "srcs" in `cargo_build_script_run` -- where the library files are passed -- make sense? I haven't seen build script code that actually depends on corresponding library files. If we wanted to expose files to the build script, perhaps `data` dependencies would be more explicit, as these could be any files, not necessarily source?
- Is there a case where it makes sense to specify `cargo_build_script_run` without the build script binary, or vice versa? I'm struggling to see an example of this case.

cargo/cargo_build_script.bzl
damienmg: No I have not since I have created those orginally trying to get rid totally of cargo raze.

examples/cargo/BUILD
damienmg: Oh right this was meant to test the access, Removed.

cargo/cargo_build_script.bzl
damienmg: * In google/starlark-rust I need to read the sources (in that case the grammar). There are a bunch of preprocessor working on the code themselve too (although those one are mostly macro now).
* I don't know about a case, the logic is separated here because it would require to recreate the logic of the `rust_binary`. I would rather create a shortcut macro rather than making it a separate rules.

cargo/cargo_build_script.bzl
```diff
@@ -0,0 +1,108 @@
...
+    name = "build_script_bin",
+    srcs = ["build.rs"],
+)
+
+cargo_build_script_run(
```
mobileink, line 90: Suggestion: don't name this with a verb, it clashes with convention. A build script is just a tool, you run it because you want its output, just like you run a compiler because you want its output. Hence "rust_library", not "rust_library_compile".  I would call this something like "cargo_builder", "cargo_configuration" or "cargo_config" or similar.

cargo/cargo_build_script.bzl
mobileink: I don't understand the first response, about "in google/starlark-rust".  Can you clarify? Actually I don't follow the second one either.

I'm inclined to favor a single target that builds and runs the build script, rather than separate build and run targets. If that's possible.

damienmg: Rephrasing the first one: In starlark-rust, the build script needs to reads some sources, hence the need for a sources (maybe srcs is not the best place for it though, data would be more logical). There is more case of generative code that want to read the sources input.

For the second point, I am just saying that I don't have a use case. This approach is made for simplicity. I am up for exposing only one rule but I would advise to make it a macro (basically keeping the underlying logic identical).

I'll go ahead and do the change so you can see how you feel about it.

cargo/cargo_build_script.bzl
damienmg: Yeah the thing is the name of a rule normally point to the generated artifacts, in this case this is not really a thing. The most common equivalent would be a `genrule`, neither cargo_builder (this would indicate building a cargo crate to me), `cargo_config*` also sounds like something that read a cargo.toml

Anyway merging the two rule make the exposed name be simply `cargo_build_script`.

mobileink:
A possible alternative: println! is a macro; can we just redefine it to write to a file in OUT_DIR? (I don't know Rust very well so it seems like it should be possible, but I'm not sure how to do it.)

cargo/cargo_build_script.bzl
damienmg: The output of this rule is a set of random files just like a `genrule` not a usable artifacts like a `rust_binary` or a `rust_test`. 

I don't have an issue using `cargo_*` prefix, the rule is now named `cargo_build_script` and is a macro that group the 2 previous rule. WDYT?

damienmg:
> A possible alternative: println! is a macro; can we just redefine it to write to a file in OUT_DIR? (I don't know Rust very well so it seems like it should be possible, but I'm not sure how to do it.)

There is a bit more than that to do. To fix the data access, we have to change also the root directory, and adjust all the path to be absolute. I would expect also to do some wrapping code to redefine the println output, so basically the same as what it is doing (in addition  to the need to parse the output nevertheless).

cargo/cargo_build_script.bzl
damienmg: Change uploaded.

examples/cargo/build_script.rs
```diff
@@ -0,0 +1,31 @@
...
+    let bleh = env::var("CARGO_FEATURE_BLEH").unwrap();
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let data = std::fs::read("cargo/test.txt").unwrap();
+    assert!(!bleh.is_empty());
+    println!(r#"cargo:rustc-env=FOO=BAR
+cargo:rustc-env=BAR=FOO
+cargo:rustc-flags=--cfg=blah="bleh"
+cargo:rustc-flags=--cfg=data="{}"
+cargo:rustc-cfg=foobar"#, std::str::from_utf8(&data).unwrap());
+    assert!(true);
```
smklein, line 28: Seems unnecessary?

examples/cargo/BUILD
```diff
@@ -0,0 +1,28 @@
+load(
+    "@io_bazel_rules_rust//rust:rust.bzl",
+    "rust_binary",
+    "rust_library",
+    "rust_test",
+)
+load(
```
smklein, line 7: Any reason for two separate load statements?

examples/cargo/BUILD
```diff
@@ -0,0 +1,28 @@
...
+    srcs = ["build_script.rs"],
+    data = ["test.txt"],
+    crate_features = ["bleh"],
+)
+
+rust_library(
+    name = "lib",
+    srcs = ["lib.rs"],
+    deps = [":build_script"],
+)
```
smklein, line 23: (No action Item)

I like this format - succinct and clear. Thanks for the updates!

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+            return BuildScriptOutput::None;
+        }
+        let param = split[1].trim().to_owned();
+        match split[0] {
+            "cargo:rustc-link-lib" => BuildScriptOutput::LinkLib(param),
+            "cargo:rustc-link-search" => BuildScriptOutput::LinkSearch(param),
+            "cargo:rustc-cfg" => BuildScriptOutput::Cfg(param),
+            "cargo:rustc-flags" => BuildScriptOutput::Flags(param),
+            "cargo:rustc-env" => BuildScriptOutput::Env(param),
+            _ => BuildScriptOutput::None,
```
smklein, line 56: What are your thoughts on at least printing a warning in this case?

The following look (currently) unsupported, and I imagine build script syntax may be expanded in the future.

- rerun-if-changed
- rerun-if-env-changed
- rustc-cdynlib-link-arg
- warning
- KEY=VALUE (for links scripts)

(from https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script)

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn absolute_path(path: &str) -> PathBuf {
+    let mut p = env::current_dir().unwrap();
+    p.push(path);
+    p
+}
+
+fn main() {
+    let mut args = env::args().skip(1);
+    // Because of the Bazel's sandbox, bazel cannot provide full path, convert all relative path to correct path.
```
smklein, line 34: Rather than relying on this assumption, how about just using: https://doc.rust-lang.org/std/fs/fn.canonicalize.html ?

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+        loop {
+            line.clear();
+            if reader.read_line(&mut line).expect("Valid script output") == 0 {
+                return result;
+            }
+            let bso = BuildScriptOutput::new(&line);
+            if bso != BuildScriptOutput::None {
+                result.push(bso);
+            }
+        }
```
smklein, line 73: Nit: Might be slightly more succinct to do:

```
while reader.read_line(&mut line).expect("Cannot read line") != 0 {
  if let Some(bso) = BuildScriptOutput::new(&line) {
    result.push(bso);
  }
  line.clear();
}
result
```

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+            .join(" ")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Cursor;
+
+    #[test]
```
smklein, line 124: WDYT about adding a test for invalid build script output?

rust/private/rustc.bzl
```diff
@@ -365,19 +387,23 @@ def add_edition_flags(args, crate):
...
-    )
+        return build_info_out_dir
+    else:
+        out_dir = ctx.actions.declare_directory(ctx.label.name + ".out_dir")
+        ctx.actions.run_shell(
+            # TODO: Remove system tar usage
+            command = ";".join([
+                "rm -fr {dir} && mkdir {dir} && tar -xzf {tar} -C {dir}".format(tar = tar_file.path, dir = out_dir.path),
+            ] + [
+                "pushd {dir}; cp -fr {in_dir}; popd {dir}".format(dir = out_dir.path, in_dir = build_info_out_dir.path),
```
smklein, line 401: Aren't the arguments to popd ignored?

examples/cargo/BUILD
damienmg: The path to include is different. Since this is not purely a rust part but a cargo interaction part it lives in a different package.

examples/cargo/BUILD
damienmg: Thanks :)

examples/cargo/build_script.rs
damienmg: Removed

cargo/cargo_build_script_runner/lib.rs
damienmg: Done. (although explicitely ignored rerun*)

cargo/cargo_build_script_runner/lib.rs
damienmg: Done.

cargo/cargo_build_script_runner/lib.rs
damienmg: Done.

rust/private/rustc.bzl
damienmg: Right, removed

cargo/cargo_build_script_runner/bin.rs
damienmg: Done.

damienmg:
Ok this is now passing all the tests having not received any more comment I'll go ahead and merge it.

illicitonion:
This is super exciting, thanks for putting it together! (I found this after writing something very similar and preparing to make a PR for it, but yours is much better than mine was going to be!)

Do you have any plans to modify `cargo-raze` to generate this instead of its existing rust_binary + genrule situation?

PR #321, opened 2020-05-28
add cfg_flags for boolean feature flags

mobileink:
Add support for boolean config flags.  --cfg 'feature="foo"' does not work; --cfg foo does.  This change supports the latter.

Signed-off-by: Gregg Reynolds <601396+mobileink@users.noreply.github.com>

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -307,6 +307,9 @@ _rust_common_attrs = {
             Dictionary of additional `"key": "value"` environment variables to set for rustc.
         """),
     ),
+    "cfg_flags": attr.string_list(
+        doc = "List of source configuration flags passed to `rustc`. Use cfg_opts to pass var=val pairs.",
+    ),
     "crate_features": attr.string_list(
         doc = _tidy("""
             List of features to enable for this crate.
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -308,7 +308,10 @@ _rust_common_attrs = {
         """),
     ),
     "cfg_flags": attr.string_list(
-        doc = "List of source configuration flags passed to `rustc`. Use cfg_opts to pass var=val pairs.",
+        doc = _tidy("""
+           List of boolean configuration flags passed to `rustc` as `--cfg foo`.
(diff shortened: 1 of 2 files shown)


damienmg:
Thanks for the change, I was wondering if that needs an actual extra attribute. I haven't tested but it sounds like this could be done in `rustc_flags`, can't it? I could also make sense of going along as there is the `crate_features` next to it, but it feel kind of bloating the API to add more way to pass rustc flags.

mobileink:
I tried to use rustc_flags but could not make it work.  Maybe I'm doing something wrong?
If I had my druthers I would rename crate_features to cfg_options.  I think it makes sense to partition rustc options into cfg_options (key-val pairs), cfg_flags (booleans), and everything else, since the cfg options indirectly configure the source code, and the rest configure the compiler.  That way the user interface makes a conceptual distinction explicit.

damienmg:
Would you be willing to go ahead and adapt this PR to actually add cfg_options and cfg_flags and emit a warning when using the old options (noting that they are marked for deprecation, and how to switch to new options)?

This is a way larger change so I would understand if you rather not dig into that rabbit hole.

mobileink:
On Tue, Jun 2, 2020 at 3:53 AM Damien Martin-Guillerez <
notifications@github.com> wrote:

> Would you be willing to go ahead and adapt this PR to actually add
> cfg_options and cfg_flags and emit a warning when using the old options
> (noting that they are marked for deprecation, and how to switch to new
> options)?
>
Sure, I'd be happy to do that, but I might not be able to get to it this
week.

c0nk:
Would it make more sense to only add `cfg_flags` for both bools and key-values? This would make it similar to `defines` for C/C++ targets.

UebelAndre:
Any updates here? In Damien's absence, I wonder if @illicitonion might have some thoughts here?

illicitonion:
Sorry, I'm struggling a bit to understand the use-case here that isn't met by the existing `rustc_flags` - would it be possible to add a test or example, in either the `test` or `examples` directory, showing some code which uses this feature, but which doesn't compile without it?

PR #322, opened 2020-05-29
Support CARGO_BIN_EXE env vars

illicitonion:
Since 1.43 cargo will set `CARGO_BIN_EXE`-prefixed env vars for each binary your crate depends on, so you can access them in integration tests.

Also, output binary names with the same name as the target, rather than replacing `-`s with `_`s, to better emulate how Cargo works.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -348,6 +348,13 @@ def rustc_compile_action(
     else:
         formatted_version = ""
 
+    # Make bin crate data deps available to tests.
+    for data in getattr(ctx.attr, "data", []):
+        if CrateInfo in data:
+            dep_crate_info = data[CrateInfo]
+            if dep_crate_info.type == "bin":
+                env["CARGO_BIN_EXE_" + dep_crate_info.output.basename] = dep_crate_info.output.short_path
+
     # Update environment with user provided variables.
     env.update(crate_info.rustc_env)
 
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -320,10 +320,27 @@ def rustc_compile_action(
     # and use `$(pwd)` which resolves the `exec_root` at action execution time.
     package_dir = ctx.build_file_path[:ctx.build_file_path.rfind("/")]
     manifest_dir_env = "CARGO_MANIFEST_DIR=$(pwd)/{} ".format(package_dir)
-    command = '{}{}{} "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd'.format(
+
+    # Handle that the binary name and crate name may be different.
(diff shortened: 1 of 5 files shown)


rust/private/rustc.bzl
```diff
@@ -320,17 +320,34 @@ def rustc_compile_action(
...
+    maybe_rename = ""
+    if crate_info.type == "bin":
+        generated_file = crate_info.name
+        if toolchain.target_arch == "wasm32":
+            generated_file = generated_file + ".wasm"
+        src = "/".join([output_dir, generated_file])
+        dst = crate_info.output.path
+        if src != dst:
+            maybe_rename = " && /bin/mv {src} {dst}".format(src=src, dst=dst)
+    command = '{manifest_dir_env}{out_dir_env}{rustc} "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd{maybe_rename}'.format(
```
damienmg, line 332: This line is long, maybe use a triple quoted string?

rust/private/rustc.bzl
```diff
@@ -320,17 +320,34 @@ def rustc_compile_action(
...
-        toolchain.rustc.path,
+    maybe_rename = ""
+    if crate_info.type == "bin":
+        generated_file = crate_info.name
+        if toolchain.target_arch == "wasm32":
+            generated_file = generated_file + ".wasm"
+        src = "/".join([output_dir, generated_file])
+        dst = crate_info.output.path
+        if src != dst:
+            maybe_rename = " && /bin/mv {src} {dst}".format(src=src, dst=dst)
```
damienmg, line 331: Why do you need that maybe_rename? I cannot make sense of the logic.

illicitonion: If a target name is `foo-bar`, both Cargo and rules_rust will translate this into a crate name of `foo_bar`, and so `rustc` will create an output file (in an intermediate directory) named `foo_bar`.

However, Cargo, when putting the output in your `target/${config}` directory will use the original target name of `foo-bar`.

The env var Cargo sets uses uses the original target name, with `-` not `_`.

If we'd rather not do this, we could instead just set the env var to use `foo-bar` and point at a file named `foo_bar`, but this seemed more compatible with what Cargo does.

If we want to keep doing this, I can gladly add more comments to explain :)

damienmg: I am fine with this behavior but I'd gladly take some comment here :)

rust/private/rustc.bzl
illicitonion: Reverted away from the the fstring-like templates

rust/private/rustc.bzl
illicitonion: Added comment

damienmg:
Thanks!

PR #326, opened 2020-06-07
Update bindgen to 0.54.0

kornholi:
The current version is quite old. This also replaces the manual BUILD
patch with Cargo Raze's new `additional_build_file`.


mfarrugi:
Lovely, thank you.

Issue #327, opened 2020-06-08
Transitions assume host platform == platform where bazel is executing


illicitonion:
This code: https://github.com/bazelbuild/rules_rust/blob/6e5fa2c570ac2f17ac1df840d060fc8aab521a07/rust/private/transitions.bzl#L3

Attempts to transition proc-macros from the target configuration to the host configuration.

Unfortunately, the configuration it transitions _to_ isn't the host configuration (which would be the customisable value of the `--host_platform` flag), but is the platform which Bazel happens to be running on.

I have a set-up where I run Bazel on a Mac, but use a Linux remote execution cluster. I set both the `--platforms` and the `--host_platform` flags to my remote execution platform, because I want all compilations to happen remotely.

Out of the box, if I try to build my code, it tries to invoke a linker locally on my OSX machine, which fails. If I hard-code setting the value of this transition to my host_platform target, it works fine. But this isn't something I can upstream, because it's specific to my execution environment.

Unfortunately, I'm not sure what value this _should_ have. Ideally I think it would be something like `//command_line_option:host_platform`, but that doesn't appear to be supported (I get an error: com.google.devtools.build.lib.packages.BuildFileNotFoundException: no such package 'command_line_option'`).

Is there a meta-value that can be used here for "Whatever the current host platform is"? That's what we should be probably using...

damienmg:
I would guess a select on some configuration values might help. /cc @katre too.

katre:
If you really want a host transition, use `cfg = 'host'` in the attribute declaration. But I want to ask, why specifically the host configuration? I would think you want, instead, the execution configuration, ie using the selected execution platform as the target for the dependency. In that case, you would want to use `cfg = 'exec'`.

damienmg:
@katre  I believe the author just got that part wrong and did not considered the difference between host and execution platform. I think the execution transition should be used there.

illicitonion:
@katre Thanks for the pointer! You're right that `exec` is probably the configuration we want to transition to.

However, with how the code is structured, we need to _conditionally_ transition to the `cfg = "exec"` based on some other attribute of the target, which is why this is hidden behind a full `transition` rather than just a `cfg` value. Is there a way to do that?

katre:
There's not currently a way to do that. We've discussed ways to more fully expose the execution transition, but right now we don't have any plans.

[...]

PR #328, opened 2020-06-09
Add separate proc_macro_deps attr

illicitonion:
Which is always configured in the exec config

Fixes https://github.com/bazelbuild/rules_rust/issues/327
See also https://github.com/google/cargo-raze/pull/164


damienmg:
/cc @dfreese @smklein 
This change looks good, if the API is fine with the user of it, I would be for moving forward with it.

dfreese:
A little concerned about this from a usability perspective, since the providers for both dep attributes are CrateInfo.  Is there a way to reject the incorrect input on either through an aspect, returning different providers, or perhaps something else?

illicitonion:
> A little concerned about this from a usability perspective, since the providers for both dep attributes are CrateInfo. Is there a way to reject the incorrect input on either through an aspect, returning different providers, or perhaps something else?

Good shout - I've added this check, and also fixed up the problems in the examples repo which it found!

damienmg:
Ok I think that's a go so merging :)

djmarcin:
I suspect this PR may be related to #502 where all proc-macro deps become unconditionally compiled in `opt` which breaks crates that detect the presence of `debug_assertions` internally to do different things (e.g. RustEmbed).

PR #329, opened 2020-06-09
Adding basic windows support

jelmansouri:
Adding binary extension support to the rust_binary rule, this allows us to build on windows.


jelmansouri:
I had the change for more than a month, in the meantime this got added:
https://github.com/bazelbuild/rules_rust/pull/322/files
I'm not sure if it breaks my original change

rust/private/rust.bzl
```diff
@@ -143,10 +143,7 @@ def _rust_binary_impl(ctx):
     toolchain = find_toolchain(ctx)
     crate_name = ctx.label.name.replace("-", "_")
 
-    if (toolchain.target_arch == "wasm32"):
-        output = ctx.actions.declare_file(ctx.label.name + ".wasm")
-    else:
-        output = ctx.actions.declare_file(ctx.label.name)
+    output = ctx.actions.declare_file(ctx.label.name + toolchain.binary_ext)
```
damienmg, line 146: I am surprised WASM test still pass as there is no declared extension corresponding: https://github.com/jelmansouri/rules_rust/blob/win_support/rust/platform/triple_mappings.bzl#L39

I don't really understand the WASM change though so I cannot tell if its missing tests or because the extension is not important.

damienmg:
> I had the change for more than a month, in the meantime this got added:
> https://github.com/bazelbuild/rules_rust/pull/322/files
> I'm not sure if it breaks my original change

I don't think it is broken, at least the CI is green.

rust/private/rust.bzl
jelmansouri: So I have a couple changes coming up after to not use run_shell in the subset of rules we are using. (cc_library and cc_binary). Similarly to this change we did on go:
https://github.com/bazelbuild/rules_go/blob/master/go/private/rules/binary.bzl#L120
It's a great thing to add the CI and I'm really happy to do it!
Our goal is not to rely on genrules nor run_shell on Windows.

jelmansouri: So what is using the Windows CI to fail is the use of run_shell, my other that I did in our codebase to not use run_shell, needs to be redone given that the command used to run the compiler is gotten more complicated

jelmansouri: Submitted some WIP to have a green CI, do you think it's ok if I get a similar change as in the go rule linked above in the function I added (will need to shuffle stuff around)?

damienmg: Yes this is fine. I also prefer genrule over run_shell as the later needs a posix shell.

jelmansouri:
The change has two parts to it:
Removing run_shell based rules since rustc doesn't behave when when launched through an msys shell to build an msvc based abi target.
Adding windows extension where necessary

Note that C++ has been chosen to not introduce more dependencies (a cc toolchain is necessary anyways to build rust code), there is a similar wrapper written in python for protoc, I'm planning to migrate it to rust or C++.
out_dir_tar is supported through the tar util posix systems only, I will discuss with @damienmg a clear path to remove out_dir_tar support.

jelmansouri:
The c++ wrapper tool and the cargo build script wrapper change are hard to split, since the wrapper consumes the files generated by it, and I had to do small changes to the format of the output files.
I can split the format from removing the use of the shell command maybe you were referring to that?
The windows changes are overall the small part of the change and I can split them. Are you ok with that?

damienmg:
> The c++ wrapper tool and the cargo build script wrapper change are hard to split, since the wrapper consumes the files generated by it, and I had to do small changes to the format of the output files.
> I can split the format from removing the use of the shell command maybe you were referring to that?
> The windows changes are overall the small part of the change and I can split them. Are you ok with that?

I think we are on the same page. Let me try to rephrase a bit how I see the 3 independent changes:

 - Change in env format, which require a change in cargo_build_script and the rust wrapper
 - Replacement of the shell script by a C++ wrapper
 - Some windows support specifics 

Depending on your preference you can include the drop of the tar file support in the C++ wrapper or as a separate change (it make sense to put it with the C++ wrapper as this comes with the reason for it). Dropping the support is just calling `fail()` if `out_dir_tar` is provided. With a message pointing out to `cargo_build_script` if people use it.

.bazelci/presubmit.yml
```diff
@@ -42,8 +42,9 @@ tasks:
     # rust-lld isn't available on RBE
     - "-@examples//hello_world_wasm:hello_world_wasm_test"
   windows:
-    windows_targets: &windows_targets
+    build_flags:
     - "--enable_runfiles" # this is not enabled by default on windows and is necessary for the cargo build scripts
+    windows_targets: &windows_targets
```
damienmg, line 47: Do we plan to have multiple windows run? Maybe there is no need for the windows_targets indirection?

rust/platform/triple_mappings.bzl
```diff
@@ -39,9 +39,9 @@ _SYSTEM_TO_BUILTIN_SYS_SUFFIX = {
 _SYSTEM_TO_BINARY_EXT = {
     "freebsd": "",
     "linux": "",
-    # TODO(acmcarther): To be verified
-    "darwin": "",
     "windows": ".exe",
+    "darwin": "",
+    # TODO(acmcarther): To be verified
```
damienmg, line 44: Remove this TODO, darwin extension is definitely empty.
```suggestion
```

rust/platform/triple_mappings.bzl
```diff
@@ -63,8 +63,8 @@ _SYSTEM_TO_DYLIB_EXT = {
     "freebsd": ".so",
     "linux": ".so",
     "darwin": ".dylib",
-    # TODO(acmcarther): To be verified
     "windows": ".dll",
+    # TODO(acmcarther): To be verified
```
damienmg, line 67: Same here
```suggestion
```

rust/platform/triple_mappings.bzl
```diff
@@ -53,8 +53,8 @@ _SYSTEM_TO_STATICLIB_EXT = {
     "freebsd": ".a",
     "linux": ".a",
     "darwin": ".a",
-    # TODO(acmcarther): To be verified
     "windows": ".lib",
+    # TODO(acmcarther): To be verified
```
damienmg, line 57: Idem

```suggestion
```

.bazelci/presubmit.yml
jelmansouri: It's to add test_targets, I'm still trying to fix rustdoct_test rules to be able to add them. But I prefer the change to go in first without this.

damienmg: Ok no worries, I am going to merge this PR as-is this is already a huge improvement over the status-quo.

PR #332, opened 2020-06-11
Set a couple more env vars for build scripts

illicitonion:
serde-json relies on CARGO_CFG_TARGET_ARCH

Debugging is easier with backtraces


illicitonion:
As per https://stackoverflow.com/questions/29421727/how-much-overhead-does-rust-backtrace-1-have the env var is only read when actually panicking, so the overhead is none until you actually panic (by which point, your extra few microseconds probably don't matter :))

dfreese:
Sounds good, thanks.

PR #333, opened 2020-06-11
Include rustc deps when running build scripts

illicitonion:
Otherwise rustc crashes on launch

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -18,12 +18,24 @@ def _cargo_build_script_run(ctx, script):
     for f in ctx.attr.crate_features:
         env["CARGO_FEATURE_" + f.upper().replace("-", "_")] = "1"
 
+    tools = depset(
+        direct = [
+            script,
+            ctx.executable._cargo_build_script_runner,
+            toolchain.rustc,
+        ],
+        transitive = [
+            # Needed for rustc to function.
+            toolchain.rustc_lib.files,
+            toolchain.rust_lib.files,
+        ],
+    )
+
     ctx.actions.run(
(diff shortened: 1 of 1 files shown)


cargo/cargo_build_script.bzl
```diff
@@ -18,12 +18,24 @@ def _cargo_build_script_run(ctx, script):
     for f in ctx.attr.crate_features:
         env["CARGO_FEATURE_" + f.upper().replace("-", "_")] = "1"
 
+    tools = depset(
```
dfreese, line 21: do all of these fit the description of tools, which is "inputs with additional runfiles that are automatically made available to the action"

https://docs.bazel.build/versions/master/skylark/lib/actions.html#run

I don't know if there's much harm in putting them all in tools, but it does seem like there was some rationale previously in splitting them.

illicitonion: As far as I can tell, `tools` is strictly better than `inputs` unless you want to explicitly drop runfiles (which doesn't seem appropriate for this kind of runner)... The only thing which was explicitly an `inputs` before (`script` was in both, for some reason?) was `toolchain.rustc`, which happens to be a single file, but I can imagine a world where `toolchain.rustc` had `toolchain.rustc_lib` as runfiles, because these files are required in order to use `rustc` at all.

Happy to move things to `inputs` if there's a reason to (I'm definitely not super familiar with this corner), but I couldn't think of a good one.

damienmg: Thanks for fixing this.

Indeed all those are tools and they need to be in the exec configuration. Inputs on the other hand are IIUC in the target configuration.

PR #334, opened 2020-06-11
Add edition to rust_repositories, and update README

dfreese:
Hopefully this will make it clear that the 2018 edition needs to be
explicitly specified.  This caught me by surprise initially, and seems
to have popped up in a few other areas.

See google/cargo-raze#165

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -5,7 +5,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 DEFAULT_TOOLCHAIN_NAME_PREFIX = "toolchain_for"
 
-def rust_repositories(version = "1.39.0", iso_date = None, rustfmt_version = "1.4.8"):
+def rust_repositories(version = "1.39.0", iso_date = None, rustfmt_version = "1.4.8", edition = None):
     """Emits a default set of toolchains for Linux, OSX, and Freebsd
 
     Skip this macro and call the `rust_repository_set` macros directly if you need a compiler for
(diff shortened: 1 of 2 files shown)


rust/repositories.bzl
```diff
@@ -5,7 +5,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 DEFAULT_TOOLCHAIN_NAME_PREFIX = "toolchain_for"
 
-def rust_repositories(version = "1.39.0", iso_date = None, rustfmt_version = "1.4.8"):
+def rust_repositories(version = "1.39.0", iso_date = None, rustfmt_version = "1.4.8", edition = None):
     """Emits a default set of toolchains for Linux, OSX, and Freebsd
```
smklein, line 9: I'm on board with the direction of this patch, but I think I have a small problem with the way in which we're specifying these default values.

- `BUILD_for_rust_toolchain` specifies `default_edition = "2015"`
- The definition of `rust_toolchain_repository` specified that the `attr` named "edition" is `attr.string(default = "2015")`
- `rust_repository_set` specifies `edition = None`, which takes the "2015" default from `rust_toolchain_repository`.

... and with this patch, `rust_repositories` now specifies `edition = None` - this implies "2015".

Additionally, there are miscellaneous comments specifying that "2015" is the default.

Could we make it more explicitly "default" in one spot?

IMO the behavior of "edition = None" implies "don't override the default" is fine, but it would be great if we could centralize the default to a single location (perhaps the `rust_toolchain_repository` definition) to make this more clear.

What do you think?

dfreese: Those are both public APIs that I would rather mess with in a separate commit from this.  rust_toolchain also specifies a default of 2015, so I think there could be more thought about this specifically.

The comments also need to strike a balance between clarity of use since that's the behavior of the function, and correctness (i.e. telling them to look at the default for a particular rule in case that changes).  Given that changing the default to 2018 would need a pretty concerted effort, I leaned towards the clarity side of things.

smklein: Ack, fair enough re: public APIs - I was just brainstorming the pain points if a "202X" version of Rust ever gets released.

dfreese: #335

Issue #304, opened 2020-04-15
Clippy Integration


smklein:
Hey all, this is more of a feature request than a bug. If discussion would be better served outside a Github issue, lemme know, happy to relocate the conversation.

TL;DR: I have a project depending on rules_rust, and I'd like to be able to use [clippy](https://github.com/rust-lang/rust-clippy) for linting produced libraries, binaries and tests.

It isn't immediately clear what would be the best way to go about doing this integration.

- Given the precedent set by `rust_doc`, would clippy targets be best implemented as a `rust_lint` target, listing all crates as dependencies?
- "clippy-driver" appears to be a non-Cargo mechanism for invoking clippy. Is there a similar target already ya'll would recommend taking a look at for "taking a Cargo command, and running it without Cargo"?

Happy to help contribute here, but I wanted to be oriented correctly before trying anything out :P

mfarrugi:
This is probably most similar to an integration for rust-language-server,
discussed in #71 or an aspect implementation of rustdoc (#152). For clippy
in particular, I think it could also make sense as an optional action
builtin to the crate rules (at the toolchain or crate level).. though I'm
not sure if this is mutually exclusive.

The last time I looked at aspect business was a while ago, and it was bit
complicated.. but the idea is that there is enough information in the
existing crate targets to be able to invoke clippy as an aspect without
adding a bunch of extra `rust_lint` targets (and ostensibly the same would
be possibly for rust_doc etc).

@damienmg I think you know the most about this sort of thing.. any thoughts?

damienmg:
Implementing it as a rule seems to me like a better approach. The rule itself might use some aspect to collect the sources but this does not seems correct for me to call Bazel with a special command-line for doing those clippy test, although this is more a UI question than an implementation one.

smklein:
as of #339 , this is done!

PR #339, opened 2020-06-13
Implement Clippy Aspect & Build Rule

smklein:
Follow-up from https://github.com/bazelbuild/rules_rust/pull/338

Working on issue: https://github.com/bazelbuild/rules_rust/issues/304


test/clippy/BUILD
```diff
@@ -0,0 +1,86 @@
...
+    deps = [":bad_library"],
+    tags = [ "manual" ],
+)
+
+rust_clippy(
+    name = "bad_test_clippy",
+    deps = [":bad_test"],
+    tags = [ "manual" ],
+    testonly = True,
+)
```
smklein, line 86: These targets all throw errors (from clippy) if they're built (i.e., with `bazel build //test/clippy:bad_binary_clippy`) - hence the "manual" tag.

I *really* want to have a way, in Bazel, to specify something like:

```
expect_build_failure_test(
    name = "binary_failing_under_clippy",
    target = ":bad_binary_clippy",
)
```

Which would try to build the target, and assert that an error was thrown.

I tried following: https://docs.bazel.build/versions/master/skylark/testing.html#failure-testing , to see what could be used here. Unfortunately, the failure seems to be arriving at "execution time" -- the *rules* build fine, it's the execution of Clippy itself which causes an error.

I also tried integrating https://github.com/bazelbuild/bazel-integration-testing , but had some issues, as I couldn't figure out how to reference this parent workspace (or even, if I should).

If ya'll have advice here, it would be really appreciated -- I think I'm at the edge of my Bazel knowledge, but I'd like to do what I can to provide an automated tests!

test/clippy/clippy_failure_test.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
smklein, line 1: As an FYI, this test is not integrated into the build system; I've just been using it manually to test. See the discussion thread in the BUILD file for my attempts to have a non-manual solution.

test/clippy/BUILD
damienmg: This has always been a hard thing. In bazel we always ended up testing using an external testing framework (e.g. https://github.com/bazelbuild/bazel-integration-testing). I don't think there is a better approach there.

test/clippy/clippy_failure_test.sh
damienmg: Cf. https://github.com/bazelbuild/bazel-integration-testing if you want to integrate it.

rust/private/clippy.bzl
```diff
@@ -0,0 +1,167 @@
...
+    "rs",
+]
+
+def _is_rust_target(srcs):
+    return any([src.extension in _rust_extensions for src in srcs])
+
+def _all_sources(target, rule):
+    srcs = []
+    if "srcs" in dir(rule.attr):
+        srcs += [f for src in rule.attr.srcs for f in src.files.to_list()]
```
damienmg, line 40: Extract `src.files.to_list()` to avoid converting twice a hash set

rust/private/clippy.bzl
```diff
@@ -0,0 +1,167 @@
...
+    if "hdrs" in dir(rule.attr):
+        srcs += [f for src in rule.attr.hdrs for f in src.files.to_list()]
+    return srcs
+
+def _rust_sources(target, rule):
+    srcs = _all_sources(target, rule)
```
damienmg, line 46: Inline the all sources then the line below can be included in the previous iteration

rust/private/clippy.bzl
```diff
@@ -0,0 +1,167 @@
...
+        "@io_bazel_rules_rust//rust:toolchain",
+        "@bazel_tools//tools/cpp:toolchain_type"
+    ],
+    implementation = _clippy_aspect_impl,
+)
+
+def _rust_clippy_rule_impl(ctx):
+    checks = depset([])
+    for dep in ctx.attr.deps:
+        checks = depset(dep[OutputGroupInfo].clippy_checks.to_list(), transitive = checks.to_list())
```
damienmg, line 159: Overusing `to_list` leads to quadratic behavior. It is better to leverage the depset structure:

```suggestion
        checks = depset(transitive = [dep[OutputGroupInfo], ctx.attr.deps])
```

I am not sure the above works, if not you can simply extract the list:
```suggestion
        checks = depset(transitive = [dep[OutputGroupInfo]] + ctx.attr.deps)
```

rust/private/clippy.bzl
```diff
@@ -0,0 +1,167 @@
...
+        "_cc_toolchain": attr.label(
+            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+        ),
+    },
+    toolchains = [
+        "@io_bazel_rules_rust//rust:toolchain",
+        "@bazel_tools//tools/cpp:toolchain_type"
+    ],
+    implementation = _clippy_aspect_impl,
+)
```
damienmg, line 154: Just a though: 
I don't remember the syntax exact of aspect but I remember they could be propagated along all the edges. If this is the case you are going to flatten all the sources depset of all the targets.

If you understand more how that works, can you expand? Else I'll try to go back to the doc to investigate.

rust/private/clippy.bzl
smklein: So if I expand this out from list comprehension form, we have the following:

```
if "src" in dir(rule.attr):
    for src in rule.attr.srcs:
        for f in src.files.to_list():
            srcs += [f]
```

I can't really pull out `src.files.to_list()`, since this must be accessed on the inner iteration of `rule.attr.srcs`.

That being said, I think it's only being evaluated once? As an example:

```
def foo():
    print("foo")
    return [1, 2, 3]

bar = [f for f in foo()]
```

Only invokes "foo" a single time.

rust/private/clippy.bzl
smklein: The first option throws an error:
`at index 0 of transitive, got element of type OutputGroupInfo, want depset`

The second option also throws the same error:
`at index 0 of transitive, got element of type OutputGroupInfo, want depset`.

rust/private/clippy.bzl
smklein: (warning: this is one of the first aspects I've written, so I am far from an expert, and would appreciate all the guidance I can get)

So I'm aware that aspects *can* propagate along all edges, but I'm not actually certain that is desired behavior here.

In the documentation https://docs.bazel.build/versions/master/skylark/aspects.html , I know that a somewhat recursive notation allows this propagation:

```
def _file_count_aspect_impl(target, ctx):
    ... (redacted)
    # Get the counts from our dependencies.
    for dep in ctx.rule.attr.deps:
        count = count + dep[FileCountInfo].count
    return [FileCountInfo(count = count)]

file_count_aspect = aspect(
    implementation = _file_count_aspect_impl,
    attr_aspects = ['deps'],
    ... (redacted)
 )
```

In this example, accessing the `dep[FileCountInfo]` value accesses the propagated return value of the aspect applied to a dependency.

In the case of the clippy aspect, however, I don't think we actually want to run clippy on all dependencies. For most crates, this would hop into third party dependencies, where clippy would possibly fail, but without really providing actionable feedback to the caller.

rust/private/clippy.bzl
smklein: I've inlined the all sources line, let me know if this is what you meant.

test/clippy/clippy_failure_test.sh
smklein: Are there any examples of integration testing beyond https://github.com/bazelbuild/bazel-integration-testing/blob/master/bazel_integration_test/test_base_test.py ? I see how to set up a python integration test and run bazel commands inside, but I'm struggling to make it reference a pre-existing workspace (e.g., one that references all the `rust_clippy` rules and sources in this directory).

smklein: Actually, I found https://github.com/hchauvin/bazel-cloud-infra/tree/587eb7c11d528148c0243eaa392ef070e44f32b8/rbs/test ... I'll try to base the clippy test off of this example.

rust/private/clippy.bzl
damienmg: Sorry that was meant to be `dep[OutputGroupInfo].clippy_checks` where I put `dep[OutputGroupInfo]`

rust/private/clippy.bzl
damienmg: Sorry I was reading this incorrectly.

rust/private/clippy.bzl
damienmg: I don't see the change (might be an issue with Github). What I meant is to inline `_all_sources` method in `_rust_sources` and then the line 47 can be merged with lines 40 and 42 (just puting the if in those lines)

rust/private/clippy.bzl
damienmg: I think you need to remove the `attr_aspects` attribute to the aspect. The rule declare in which way the aspect propagate along the top edge, so no need for the aspect to propagate further.

Indeed for clippy that would be strange to evaluate the source of all recursive targets.

test/clippy/clippy_failure_test.sh
smklein: I've really tried integrating this, but I ran into multiple issues:

1) Following [the only non-trivial example of the integration test I've found](https://github.com/hchauvin/bazel-cloud-infra/tree/587eb7c11d528148c0243eaa392ef070e44f32b8/rbs/test)... If I create a directory `rules_rust/test/clippy/workspace`, under which I place all these files, I can glob together the sources + WORKSPACE + BUILD file as runfiles, and read them within the test. **However**, although I want to pull in the top-level repo as:

(within `rules_rust/test/clippy/workspace/WORKSPACE`)
```
local_repository(
   name = "io_bazel_rules_rust",
   path = "path_to_io_bazel_workspace_root",
)
```
This doesn't work: The execution root of this "clippy sub-workspace" cannot reference the path to the "rules rust workspace root".

I noticed that the `examples` directory of this repo also uses `local_repository`, but since it isn't loaded via runfiles, it can use relative paths. The same doesn't appear to apply here.

2) After hacking around that issue by injecting an absolute path to the rules_rust WORKSPACE (yuck), the integration test repo failed because it is using extremely old versions of Bazel, which don't seem compatible with `rules_rust`. I tried supplying the `versions` argument to `bazel_py_integration_test`, but this argument doesn't work; it only works for the pre-defined versions of Bazel (0.21.0, 0.22.0, 0.23.2), or else the rule will depend on python libraries which don't exist.

smklein: I have an example of these issues here in a draft PR (not building):
  https://github.com/bazelbuild/rules_rust/pull/340

rust/private/clippy.bzl
smklein: Attempting:
`checks = depset(transitive = [dep[OutputGroupInfo].clippy_checks, ctx.attr.deps])`
Causes the error:
`at index 0 of transitive, got element of type list, want depset`

Attempting:
`checks = depset(transitive = [dep[OutputGroupInfo].clippy_checks] + ctx.attr.deps)`
Causes the error:
`at index 0 of transitive, got element of type Target, want depset`

rust/private/clippy.bzl
smklein: Done

rust/private/clippy.bzl
smklein: Done (squashed the two methods into one)

rust/private/clippy.bzl
damienmg: Ok sorry for making you try several thing. From the error the correct way should be:

```
depset(transitive = ctx.attr.deps, direct = dep[OutputGroupInfo].clippy_checks)
```

test/clippy/clippy_failure_test.sh
damienmg: I went to dive a bit more and the project bazel-integration-testing seems dead, I would suggest to not spend too much time on that.

rust/private/clippy.bzl
```diff
@@ -0,0 +1,161 @@
...
+    checks = depset([])
+    for dep in ctx.attr.deps:
+        checks = depset(dep[OutputGroupInfo].clippy_checks.to_list(), transitive = checks.to_list())
+    return [DefaultInfo(files = checks)]
+
+rust_clippy = rule(
+    implementation = _rust_clippy_rule_impl,
+    attrs = {
+        'deps': attr.label_list(aspects = [rust_clippy_aspect]),
+    },
```
damienmg, line 160: add doc attributes?

rust/private/clippy.bzl
smklein: Using:

```
checks = depset([])
for dep in ctx.attr.deps:
  checks = depset(transitive = ctx.attr.deps, direct = dep[OutputGroupInfo].clippy_checks)
return [DefaultInfo(files = checks)]
```

Results in:

```
File "/home/smklein/repos/rules_rust/rust/private/clippy.bzl", line 156, in _rust_clippy_rule_impl
                depset(<2 more arguments>)
for direct, got depset, want sequence
```

I read through https://docs.bazel.build/versions/master/skylark/depsets.html, and it seems like the following works:

```
files = depset([], transitive = [dep[OutputGroupInfo].clippy_checks for dep in ctx.attr.deps])
return [DefaultInfo(files = files)]
```

damienmg: Oh right sorry I got all mixed up. Yes the solution you gave is the correct one, it avoids creating N depset

smklein: Updated

smklein:
As a final validity test, I ran:
```
cd examples
bazel build --aspects=@io_bazel_rules_rust//rust:rust.bzl%rust_clippy_aspect --output_groups=clippy_checks //...
```

This uncovered a handful of issues -- which have since been fixed. Of note:

- Many of the "compilation-command-building" arguments assume they ingest `ctx`, and operate on `ctx.files` or `ctx.file` directly. When building directly on a target, this is fine, but in the context of an aspect, this is non-ideal, since the **actual** desired file/files exist within `ctx.rule.files` or `ctx.rule.file`, respectively. The functions have been refactored to act on a user-supplied file/files in these cases.
- Rather than trying to "recreate" CrateInfo, I realized I could just access it directly from the Aspect's target, so I'm doing that now.

@damienmg I appreciate the help getting this patch reviewed, but I don't want to slide something in after the "+2" -- would you be willing to take another look?

Also, I'm currently running clippy on the `examples/` directory manually - I think it would be cool if we could actually add that to the automated build. I tried adding a top-level `rust_clippy` targets that includes all `rust_(library,binary,test)` targets, but this would require significant changes to the visibility within  `examples/`.

Any advice on this? How could I go about changing the buildkite script to run this as a check on future PRs?

rust/private/clippy.bzl
```diff
@@ -145,17 +133,83 @@ rust_clippy_aspect = aspect(
...
 
 def _rust_clippy_rule_impl(ctx):
-    checks = depset([])
-    for dep in ctx.attr.deps:
-        checks = depset(dep[OutputGroupInfo].clippy_checks.to_list(), transitive = checks.to_list())
-    return [DefaultInfo(files = checks)]
+    files = depset([], transitive = [dep[OutputGroupInfo].clippy_checks for dep in ctx.attr.deps])
+    return [DefaultInfo(files = files)]
 
 rust_clippy = rule(
```
damienmg, line 172: Missing docstring?

damienmg: Gosh forget that, the documentation is below :P

damienmg:
So for test, I think this is fine as it is.

You could also modify the [preubmit.yml](https://github.com/bazelbuild/rules_rust/blob/master/.bazelci/presubmit.yml#L47) file to add the test you want but I am fine with checking this PR in as-is.

smklein:
> So for test, I think this is fine as it is.
> 
> You could also modify the [preubmit.yml](https://github.com/bazelbuild/rules_rust/blob/master/.bazelci/presubmit.yml#L47) file to add the test you want but I am fine with checking this PR in as-is.

Oh rad, didn't see that file! I added `clippy_examples` and `clippy_failure` as tests to that file.

PR #341, opened 2020-06-15
Support sys-crates build script dependencies.

a1ph:
Sys-crates may produce "cargo:var=value" output that must be translated
into DEP_CRATE_NAME_VAR=VALUE environment variable that is then must be
made visible to all direct dependent crates build files.

This in particular is used by openssl and openssl-sys crates.


cargo/cargo_build_script.bzl
```diff
@@ -1,14 +1,25 @@
...
     out_dir = ctx.actions.declare_directory(ctx.label.name + ".out_dir")
     env_out = ctx.actions.declare_file(ctx.label.name + ".env")
+    dep_env_out = ctx.actions.declare_file(ctx.label.name + ".depenv")
     flags_out = ctx.actions.declare_file(ctx.label.name + ".flags")
     manifest_dir = "%s.runfiles/%s" % (script.path, ctx.label.workspace_name or ctx.workspace_name)
     compilation_mode_opt_level = get_compilation_mode_opts(ctx, toolchain).opt_level
+
+    crate_name = ctx.attr.crate_name
+    # Derive crate name from the rule label which is <crate_name>_build_script
+    # TODO: Remove once cargo_raze starts providing crate_name attribute.
```
damienmg, line 16: I don't think you need to remove this, this is a nice fall back

cargo/cargo_build_script.bzl
```diff
@@ -36,11 +47,21 @@ def _cargo_build_script_run(ctx, script):
         ],
     )
 
-    ctx.actions.run(
-        executable = ctx.executable._cargo_build_script_runner,
-        arguments = [script.path, env_out.path, flags_out.path],
-        outputs = [out_dir, env_out, flags_out],
+    cmd = ""
```
damienmg, line 50: Maybe comment that part to explain for future reader?

a1ph: done

PR #343, opened 2020-06-17
cc works for non-absolute-path toolchains

illicitonion:
Bring along the files, and make sure to absolut-ise the CC path.

Also, add in the env vars required by the linker, because these are
required for proper toolchain detection on the default OSX cc_toolchain.


illicitonion:
I have an example to add to examples which fails without this PR, but it also depends on a second PR I'm about to open.

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -31,18 +34,35 @@ fn main() {
...
     let rustc = canonicalize(&rustc_env).expect(&format!("Failed to canonicalize '{}'", rustc_env));
+
+    let cc = env::var_os("CC").map(|env_var| {
+        let cc_path = Path::new(&env_var);
+        if cc_path.is_relative() {
+            exec_root.join(cc_path).into_os_string()
+        } else {
+            env_var
+        }
+    });
```
dfreese, line 48: Would replacing this with
```rust
let cc = env::var_os("CC").map(|env_var| {
  canonicalize(&env_var).expect(&format!("Failed to canonicalize '{}'", env_var))
});
```
be equivalent?  It seems like command.env takes `Path`s

illicitonion: It would be equivalent assuming the `CC` variable itself isn't a symlink, and none of its parents are. I know that some remote execution platforms have options to use symlinks for input files, so you'd then end up with `CC` pointing at something like `/cas/abcd1234` rather than `/path/to/working/directory/clang` which may be problematic (e.g. if it's a script which tries to look up its own directory to find adjacent files). In generally I've trained myself to avoid `canonicalize` wherever possible - if y'all would be open to it, ideally I'd rather replace all of the `canonicalize` calls in this file with calls to `exec_root.join`, than go the other direction.

`Command::env` is happy taking anything which can be treated as an `OsStr`, and `Path` is just a wrapper around an `OsStr`, so either type is interchangeable and zero-cost here.

damienmg: Do you mind adding a comment about this?

illicitonion: I just sent https://github.com/bazelbuild/rules_rust/pull/347 which has a lengthy comment - if we can get that merged I'll gladly rebase the other PRs so they pull it in :)

illicitonion:
Rebased on the now merged #347

cargo/cargo_build_script_runner/bin.rs
dfreese: Thanks for the clarification.

PR #346, opened 2020-06-17
 Link flags are passed to transitive rdeps

illicitonion:
And absolute paths are redacted into $EXEC_ROOT-relative paths

This mirrors Cargo's behaviour of accumulating link flags and passing
them to the linker - otherwise there's not much point in sys libraries
outputting them.

This currently over-adds out_dirs and link flags to all actions, rather
than just those that will invoke the linker - we can reduce that if we
have a clear indication of what targets will invoke a linker, but I
wanted to go with a safe-but-over-invalidating solution rather than
picking a bad heuristic.

Commits are separately useful to review.

Depends on https://github.com/bazelbuild/rules_rust/pull/343 (the first commit of this PR).

I have a WIP PR which removes both the new usage of system `tar` and the existing usages in the PR by building small implementations of `tar` and `untar` in Rust - will publish it shortly, but it's non-trivial enough to prefer doing as a follow-up :)

### DIFF
--- a/cargo/cargo_build_script_runner/bin.rs
+++ b/cargo/cargo_build_script_runner/bin.rs
@@ -18,8 +18,7 @@ extern crate cargo_build_script_output_parser;
 
 use cargo_build_script_output_parser::BuildScriptOutput;
 use std::env;
-use std::fs::{File, create_dir_all};
-use std::io::Write;
+use std::fs::{create_dir_all, write};
 use std::path::Path;
 use std::process::{exit, Command};
 
@@ -66,18 +65,12 @@ fn main() {
             }
 
             let output = BuildScriptOutput::from_command(&mut command);
-            let mut f =
-                File::create(&envfile).expect(&format!("Unable to create file {}", envfile));
-            f.write_all(BuildScriptOutput::to_env(&output).as_bytes())
-                .expect(&format!("Unable to write file {}", envfile));
-            let mut f =
-                File::create(&depenvfile).expect(&format!("Unable to create file {}", depenvfile));
-            f.write_all(BuildScriptOutput::to_dep_env(&output, &crate_name).as_bytes())
-                .expect(&format!("Unable to write file {}", depenvfile));
-            let mut f =
-                File::create(&flagfile).expect(&format!("Unable to create file {}", flagfile));
-            f.write_all(BuildScriptOutput::to_flags(&output).as_bytes())
-                .expect(&format!("Unable to write file {}", flagfile));
+            write(&envfile, BuildScriptOutput::to_env(&output).as_bytes())
+                .expect(&format!("Unable to write file {:?}", envfile));
+            write(&depenvfile, BuildScriptOutput::to_dep_env(&output, &crate_name).as_bytes())
+                .expect(&format!("Unable to write file {:?}", depenvfile));
+            write(&flagfile, BuildScriptOutput::to_flags(&output).as_bytes())
+                .expect(&format!("Unable to write file {:?}", flagfile));
         }
         _ => {
             eprintln!("Usage: $0 progname crate_name envfile flagfile depenvfile [arg1...argn]");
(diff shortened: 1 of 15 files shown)


cargo/cargo_build_script_runner/bin.rs
```diff
@@ -16,48 +16,77 @@
...
+            env_var
+        }
+    });
+
+    match (args.next(), args.next(), args.next(), args.next(), args.next(), args.next(), args.next(), args.next()) {
+        (Some(progname), Some(crate_name), Some(out_dir_archive), Some(out_dir_path_file), Some(envfile), Some(flagfile), Some(linkflags), Some(depenvfile)) => {
+            let out_dir = Path::new(&out_dir_archive).parent().expect(&format!("out_dir_archive {:?} didn't have a parent", out_dir_archive)).join("out_dir");
+            create_dir_all(&out_dir).expect(&format!("Failed to create directory {:?}", out_dir));
+            write(&out_dir_path_file, out_dir.to_string_lossy().as_bytes()).expect(&format!("Failed to write {:?}", &out_dir_path_file));
+            let out_dir_abs = exec_root.join(&out_dir);
```
damienmg, line 49: Why not use `canonicalize`?

illicitonion: See https://github.com/bazelbuild/rules_rust/pull/343#discussion_r441751274 - resolving symlinks isn't a great thing to implicitly do where what we really want is just an absolute path.

illicitonion:
> Overall the PR looks fine to me except the usage of tar.
> 
> Why tar-ing & detaring?
> 
> Passing directory artifact would work just fine. I disagree with the fact that the directory artifact are not supported well in Bazel (there is an issue that the directory might not be created, just like other output artifact, on some remote platform, but it is not a bazel issue per se).
> 
> Tar is going to add expensive steps in addition to code complexity.

I reverted to using directories :)

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -141,17 +147,23 @@ impl BuildScriptOutput {
...
+                BuildScriptOutput::Cfg(e) => compile_flags.push(format!("--cfg={}", e)),
+                BuildScriptOutput::Flags(e) => compile_flags.push(e.to_owned()),
+                BuildScriptOutput::LinkLib(e) => link_flags.push(format!("-l{}", e)),
+                BuildScriptOutput::LinkSearch(e) => link_flags.push(format!("-L{}", e)),
+                _ => { },
+            }
+        }
+        CompileAndLinkFlags {
+            compile_flags: compile_flags.join(" "),
+            link_flags: link_flags.join(" ").replace(exec_root, "$EXEC_ROOT"),
```
damienmg, line 165: Add `{}` around the exec root
```suggestion
            link_flags: link_flags.join(" ").replace(exec_root, "${EXEC_ROOT}"),
```

It also needs to be adapted in the test.

rust/private/rustc.bzl
```diff
@@ -519,26 +517,38 @@ def add_edition_flags(args, crate):
...
+    # Env vars and build flags which need to be set in the action's command line, rather than on the action's env,
+    # because they rely on other env vars or commands.
+    dynamic_env = {}
+    dynamic_build_flags = []
+
+    # TODO: Remove system tar usage
+    if build_info:
+        prep_commands.append("export $(cat %s)" % build_info.rustc_env.path)
+        # out_dir will be added as input by the transitive_build_infos loop below.
+        dynamic_env["OUT_DIR"] = "$EXEC_ROOT/{}".format(build_info.out_dir.path)
```
damienmg, line 536: here and below:
```suggestion
        dynamic_env["OUT_DIR"] = "${{EXEC_ROOT}}/{}".format(build_info.out_dir.path)
```

cargo/cargo_build_script_runner/lib.rs
illicitonion: Done

rust/private/rustc.bzl
illicitonion: Done

PR #349, opened 2020-06-20
feat: Support downloading nightly rustc-dev components

Arm1stice:
Working with the Rust compiler crates requires the rustc-dev component. This is distributed separately from the standard library. This PR adds a new option to download the rustc-dev component which defaults to False. The component is only distributed with nightly Rust, so the `version` option must be set to "nightly".

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -420,6 +420,7 @@ def _rust_toolchain_repository_impl(ctx):
 
     for target_triple in [ctx.attr.exec_triple]:
         BUILD_components.append(_load_rust_stdlib(ctx, target_triple))
+        # extra_target_triples contains targets such as wasm, which don't have rustc_dev components
         if ctx.attr.dev_components and target_triple not in ctx.attr.extra_target_triples:
             _load_rustc_dev_nightly(ctx, target_triple)
 
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -5,7 +5,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 DEFAULT_TOOLCHAIN_NAME_PREFIX = "toolchain_for"
 
-def rust_repositories(version = "1.44.0", iso_date = None, rustfmt_version = "1.4.8", edition = None):
+def rust_repositories(version = "1.44.0", iso_date = None, rustfmt_version = "1.4.8", edition = None, dev_components = False):
     """Emits a default set of toolchains for Linux, OSX, and Freebsd
 
     Skip this macro and call the `rust_repository_set` macros directly if you need a compiler for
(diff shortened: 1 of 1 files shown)


rust/repositories.bzl
```diff
@@ -392,7 +418,14 @@ def _rust_toolchain_repository_impl(ctx):
...
-    for target_triple in [ctx.attr.exec_triple] + ctx.attr.extra_target_triples:
+    for target_triple in [ctx.attr.exec_triple]:
+        BUILD_components.append(_load_rust_stdlib(ctx, target_triple))
+        if ctx.attr.dev_components:
+            # No BUILD file is needed because components will match the stdlib
+            # BUILD file globs
+            _load_rustc_dev_nightly(ctx, target_triple)
+
+    for target_triple in ctx.attr.extra_target_triples:
         BUILD_components.append(_load_rust_stdlib(ctx, target_triple))
```
Arm1stice, line 429: `ctx.attr.extra_target_triples` contains the wasm target. While there is a stdlib package, there are no rustc-dev components shipped for this target. Because of this, I split this into two different loops. This means the `if` statement in the first loop only has to check if dev_components is True. Another option might be to keep one loop, and have the if statement be:
```python
if ctx.attr.dev_components and target_triple not in ctx.attr.extra_target_triples:
```
I prefer the two loops personally, but I can also switch to the longer `if` statement if there is a strong preference for it by the maintainers.

rust/repositories.bzl
```diff
@@ -392,7 +418,14 @@ def _rust_toolchain_repository_impl(ctx):
     if ctx.attr.rustfmt_version:
         BUILD_components.append(_load_rustfmt(ctx))
 
-    for target_triple in [ctx.attr.exec_triple] + ctx.attr.extra_target_triples:
+    for target_triple in [ctx.attr.exec_triple]:
+        BUILD_components.append(_load_rust_stdlib(ctx, target_triple))
+        if ctx.attr.dev_components:
+            # No BUILD file is needed because components will match the stdlib
+            # BUILD file globs
```
damienmg, line 425: What does that comment means? It is probably not needed.

rust/repositories.bzl
damienmg: After reading the code, I think one loop will make reading the code easier, and adding a comment on why excluding extra_target_triples make sense (e.g. "it contains wasm and the likes, which does not have a dev component")

rust/repositories.bzl
Arm1stice: Okay, I'll remove it. The comment was explaining why the `_load_rustc_dev_nightly` call didn't follow the pattern of being inside of a `BUILD_components.append` call

Arm1stice:
New commits address @damienmg's comments

Arm1stice:
Fixed the broken build

PR #350, opened 2020-06-22
Preserve and tidy stdout on failure

illicitonion:
Rather than discarding lines which are not directives to cargo, print
them.

Also, print a friendly message, rather than a backtrace, if the build
script itself exits non-0.

### DIFF
--- a/cargo/cargo_build_script_runner/bin.rs
+++ b/cargo/cargo_build_script_runner/bin.rs
@@ -20,9 +20,9 @@ use cargo_build_script_output_parser::{BuildScriptOutput, CompileAndLinkFlags};
 use std::env;
 use std::fs::{create_dir_all, write};
 use std::path::Path;
-use std::process::{exit, Command};
+use std::process::Command;
 
-fn main() {
+fn main() -> Result<(), String> {
     // We use exec_root.join rather than std::fs::canonicalize, to avoid resolving symlinks, as
     // some execution strategies and remote execution environments may use symlinks in ways which
     // canonicalizing them may break them, e.g. by having input files be symlinks into a /cas
@@ -64,16 +64,16 @@ fn main() {
                 command.env("CC", cc);
             }
 
-            let output = BuildScriptOutput::from_command(&mut command).unwrap_or_else(|exit_code| {
-                eprintln!(
+            let output = BuildScriptOutput::from_command(&mut command).map_err(|exit_code| {
+                format!(
                     "Build script process failed{}",
                     if let Some(exit_code) = exit_code {
                         format!(" with exit code {}", exit_code)
                     } else {
                         String::new()
-                    });
-                std::process::exit(1)
-            });
+                    }
+                )
+            })?;
 
             write(&envfile, BuildScriptOutput::to_env(&output).as_bytes())
                 .expect(&format!("Unable to write file {:?}", envfile));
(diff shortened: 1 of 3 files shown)


cargo/cargo_build_script_runner/lib.rs
```diff
@@ -50,11 +50,13 @@ impl BuildScriptOutput {
     fn new(line: &str) -> Option<BuildScriptOutput> {
         let split = line.splitn(2, '=').collect::<Vec<_>>();
         if split.len() <= 1 {
+            print!("{}", line);
             return None;
         }
         let param = split[1].trim().to_owned();
         let key_split = split[0].splitn(2, ':').collect::<Vec<_>>();
         if key_split.len() <= 1 || key_split[0] != "cargo" {
+            print!("{}", line);
```
damienmg, line 59: nit: move the line below the command

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -50,11 +50,13 @@ impl BuildScriptOutput {
     fn new(line: &str) -> Option<BuildScriptOutput> {
         let split = line.splitn(2, '=').collect::<Vec<_>>();
         if split.len() <= 1 {
+            print!("{}", line);
```
damienmg, line 53: nit: add the same comment than below: `// Not a cargo directive`

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -57,13 +57,24 @@ fn main() {
...
+                .env("RUST_BACKTRACE", "full");
 
             if let Some(cc) = cc {
                 command.env("CC", cc);
             }
 
-            let output = BuildScriptOutput::from_command(&mut command);
+            let output = BuildScriptOutput::from_command(&mut command).unwrap_or_else(|exit_code| {
+                eprintln!(
+                    "Build script process failed{}",
```
damienmg, line 69: opt: Can we make this error stands out in some way? When I was debugging this code it was kind of hard to figure out if the error was coming from the build script or from the runner, that would be nice to make this really stands out (also to see it from the bazel output).

cargo/cargo_build_script_runner/lib.rs
illicitonion: I'm not quite sure what you mean here - what are you hoping will appear before this?

dfreese:
Does this have output that shows up during normal execution of bazel build?

I've noticed some output from the build script rule that I haven't had a chance to track down yet, but ideally it should be quiet.

illicitonion:
> Does this have output that shows up during normal execution of bazel build?

It will output if the build script unconditionally writes to stdout, similar to how a genrule will output in a normal build if it writes to stdout. Most build scripts do not write to stdout as a matter of course.

> I've noticed some output from the build script rule that I haven't had chance to track down yet, but ideally it should be quiet.

Yeah, I've seen a bunch of `declare -x`s which it would be nice to silence :)

damienmg:
@dfreese Are you fine with that change?

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -57,13 +57,24 @@ fn main() {
...
+                    if let Some(exit_code) = exit_code {
+                        format!(" with exit code {}", exit_code)
+                    } else {
+                        String::new()
+                    });
+                std::process::exit(1)
```
dfreese, line 75: Would it make more sense to have main return an Result with some reasonable sort of error type here?

illicitonion: Done - I'm not sure either is noticeably nicer than the other :)

PR #351, opened 2020-06-23
Clarify error message

achew22:
The error message refers to `proc-macro-deps`, which is a string that only exists in this file. Updating the reference to say `proc_macro_deps` which is a common string in the repo and is well documented.


damienmg:
LGTM, thanks!

PR #352, opened 2020-06-23
build scripts: Set path to AR as well as CC

illicitonion:


illicitonion:
I modified this to not set `AR` by default on OSX because Bazel's default OSX toolchain sets `ar_executable` to `libtool` not `ar`, which doesn't work as `AR` :) CI is now green - PTAL!

PR #353, opened 2020-06-24
Remove python usage in the repo

jelmansouri:
Reducing the area of used languages since c++ has been used and the cc toolchain is a hard dependency for rust.
Only rust code and minimal cc remains to wrap executables when necessary


proto/optional_output_wrapper/bin.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2018 The Bazel Authors. All rights reserved.
```
damienmg, line 1:
```suggestion
// Copyright 2020 The Bazel Authors. All rights reserved.

```

proto/optional_output_wrapper/bin.rs
```diff
@@ -0,0 +1,48 @@
...
+
+fn ensure() -> Result<(),Box<dyn error::Error>> {
+	let index = args().position(|a| a=="--").ok_or("no --")?;
+	let optional_outputs = args().take(index).collect::<Vec<String>>();
+	let exe = args().nth(index+1).ok_or("no exe")?;
+	let exe_args = args().skip(index+2).collect::<Vec<String>>();
+	if exe_args.len() < 1 {
+		return Err("no exe args".into());
+	}
+	let _ = Command::new(exe).args(exe_args).output();
```
damienmg, line 33: If I am reading this correctly this silently capture all the output, the legacy behavior was to simply forward stdout and stdin. Can this behavior be kept?

proto/optional_output_wrapper/BUILD
```diff
@@ -0,0 +1,7 @@
+load("//rust:rust.bzl", "rust_binary")
```
damienmg, line 1: This is a simple 1 file binary, what is the logic for moving it to a new package?

jelmansouri: I just find it's find it's cleaner to have this in a separate directory with it's own build file.
Moved it to the same package as the parent.
Ideally we should move it to //proto/private I feel.

proto/optional_output_wrapper/bin.rs
jelmansouri: Yes you are correct, changed it to status, which inherits stdout, stderr & stdin handles

PR #355, opened 2020-06-25
Remove run shell from build scripts runner

jelmansouri:
As discussed with @damienmg

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -14,7 +14,6 @@ def _cargo_build_script_run(ctx, script):
     compilation_mode_opt_level = get_compilation_mode_opts(ctx, toolchain).opt_level
 
     crate_name = ctx.attr.crate_name
-
     # Derive crate name from the rule label which is <crate_name>_build_script if not provided.
     if not crate_name:
         crate_name = ctx.label.name
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -14,6 +14,7 @@ def _cargo_build_script_run(ctx, script):
     compilation_mode_opt_level = get_compilation_mode_opts(ctx, toolchain).opt_level
 
     crate_name = ctx.attr.crate_name
+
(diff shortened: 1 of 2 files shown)


cargo/cargo_build_script.bzl
```diff
@@ -14,6 +14,7 @@ def _cargo_build_script_run(ctx, script):
     compilation_mode_opt_level = get_compilation_mode_opts(ctx, toolchain).opt_level
 
     crate_name = ctx.attr.crate_name
+
```
damienmg, line 17: nit: unrelated change, remove?

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -53,12 +53,28 @@ fn main() {
...
             let mut command = Command::new(exec_root.join(&progname));
             command
-                .args(args)
                 .current_dir(manifest_dir.clone())
                 .env("OUT_DIR", out_dir_abs)
                 .env("CARGO_MANIFEST_DIR", manifest_dir)
                 .env("RUSTC", rustc);
 
+            while let Some(dep_env_path) = args.next() {
+                if let Ok(contents) = read_to_string(dep_env_path) {
```
damienmg, line 62: opt: add an error handling here? If the file is unreadable, that's probably a worthy failure mode.

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -53,12 +53,28 @@ fn main() {
...
+                    for line in contents.split('\n') {
+                        let mut key_val = line.splitn(2, '=');
+                        match (key_val.next(), key_val.next()) {
+                            (Some(key), Some(value)) => {
+                                command.env(key, value);
+                            }
+                            _ => {
+                                eprintln!("error: Wrong environment file format");
```
damienmg, line 70: This is a "should not happen" error, which means if we fall there we have a serious issue in the build rules, maybe leaves some comment to help the user raise this as a bug?

jelmansouri: Makes sense

PR #359, opened 2020-06-29
Make stardoc our documentation tool

damienmg:
The current documentation was difficult to maintain and not in sync with the inline one.

These changes adjust the setting so we can generate the documentation in markdown format and it will continue to be served on Github pages, this time with an easy way to sync it.

Each commit of this PR have been made to be reviewed separately.

Fixes #337

PR #360, opened 2020-06-29
Deprecate out dir tar

damienmg:
This attribute force the usage of tar as part of the build which
is problematic for various platform. Instead of adding support
for it in native format, just remove this one as we have
the `cargo_build_script` rule.

Depends on #359 

/cc @jelmansouri @illicitonion


rust/private/rustc.bzl
```diff
@@ -518,29 +518,18 @@ def add_edition_flags(args, crate):
         args.add("--edition={}".format(crate.edition))
 
 def _create_out_dir_action(ctx, file, build_info, dep_info):
-    tar_file_attr = getattr(file, "out_dir_tar", None)
-    if build_info and tar_file_attr:
-        fail("Target {} has both a build_script dependency and an out_dir_tar - this is not allowed.".format(ctx.label))
-
     prep_commands = []
     input_files = []
```
jelmansouri, line 522: input_files is note necessary anymore, can be removed I think.

rust/private/rust.bzl
```diff
@@ -48,6 +48,10 @@ See https://github.com/bazelbuild/rules_rust/pull/203 for more details
 def _determine_output_hash(lib_rs):
     return repr(hash(lib_rs.path))
 
+def _deprecated_attributes(ctx):
+    if getattr(ctx.attr, "out_dir_tar", None):
+        fail(ctx, "`out_dir_tar` is deprecated, please use cargo/cargo_build_script.bzl instead.")
```
jelmansouri, line 53: I don't know if it makes sense to mention using cargo raze > 0.3.4. I think most of the uses of this comes from templates predating that version.

rust/private/rustc.bzl
jelmansouri: Nevemind it's used below :)

rust/private/rust.bzl
illicitonion, line 53:
```suggestion
        fail(ctx, "`out_dir_tar` is no longer supported, please use cargo/cargo_build_script.bzl instead.")
```

damienmg: Done.

damienmg: Done.

rust/private/rust.bzl
```diff
@@ -347,13 +357,7 @@ _rust_common_attrs = {
...
     ),
     "out_dir_tar": attr.label(
-        doc = _tidy("""
-            An optional tar or tar.gz file unpacked and passed as OUT_DIR.
-
-            Many library crates in the Rust ecosystem require sources to be provided
-            to them in the form of an OUT_DIR argument. This argument can be used to
-            supply the contents of this directory.
-        """),
+        doc = "__Deprecated__, do not use, refers to [#cargo_build_script] instead.",
```
dfreese, line 360:
```suggestion
        doc = "__Deprecated__, do not use, see [#cargo_build_script] instead.",
```

damienmg: Done.

PR #361, opened 2020-06-30
Export `CcInfo` for `staticlib` & `cdylib` allowing rust outputs to be used in c++ rules

GregBowyer:
Its minor but since `CcInfo` now exists we can state `staticlib` to be a statlclib and `cdylib` to be the relevant `.so`

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -69,6 +69,47 @@ _SYSTEM_TO_DYLIB_EXT = {
     "unknown": ".wasm",
 }
 
+# See https://github.com/rust-lang/rust/blob/master/src/libstd/build.rs
+_SYSTEM_TO_STDLIB_LINKFLAGS = {
+    # TODO(bazelbuild/rules_cc#75):
+    #
+    # Right now bazel cc rules does not specify the exact flag setup needed for calling out system
+    # libs, that is we dont know given a toolchain if it should be, for example,
+    # `-lxxx` or `/Lxxx` or `xxx.lib` etc.
+    #
+    # We include the flag setup as they are _commonly seen_ on various platforms with a cc_rules
+    # style override for people doing things like gnu-mingw on windows.
+    #
+    # If you are reading this ... sorry! set the env var `BAZEL_RUST_STDLIB_LINKFLAGS` to
+    # what you need for your specific setup, for example like so
+    # `BAZEL_RUST_STDLIB_LINKFLAGS="-ladvapi32:-lws2_32:-luserenv"`
+
+    "freebsd": ["-lexecinfo", "-lpthread"],
+    # TODO: This ignores musl. Longer term what does Bazel think about musl?
+    "linux": ["-ldl", "-lpthread"],
+    "darwin": ["-lSystem", "-lresolv"],
+    "uwp": ["ws2_32.lib"],
+    "windows": ["advapi32.lib", "ws2_32.lib", "userenv.lib"],
+    "ios": ["-lSystem", "-lobjc", "-framework Security", "-framework Foundation", "-lresolv"],
+    # NOTE: Rust stdlib `build.rs` treats android as a subset of linux, rust rules treat android
+    # as its own system.
+    "android": ["-ldl", "-llog", "-lgcc"],
+    "emscripten": [],
+    "nacl": [],
+    "bitrig": [],
+    "dragonfly": ["-lpthread"],
+    "netbsd": ["-lpthread", "-lrt"],
+    "openbsd": ["-lpthread"],
+    "solaris": ["-lsocket", "-lposix4", "-lpthread", "-lresolv"],
+    "illumos": ["-lsocket", "-lposix4", "-lpthread", "-lresolv", "-lnsl", "-lumem"],
+    "fuchsia": ["-lzircon", "-lfdio"],
+    # TODO(gregbowyer): If rust stdlib is compiled for cloudabi with the backtrace feature it
+    # includes `-lunwind` but this might not actually be required.
+    # I am not sure which is the common configuration or how we encode it as a link flag.
+    "cloudabi": ["-lunwind", "-lc", "-lcompiler_rt"],
+    "unknown": [],
+}
+
 def cpu_arch_to_constraints(cpu_arch):
     plat_suffix = _CPU_ARCH_TO_BUILTIN_PLAT_SUFFIX[cpu_arch]
 
@@ -112,6 +153,9 @@ def system_to_staticlib_ext(system):
 
 def system_to_binary_ext(system):
     return _SYSTEM_TO_BINARY_EXT[system]
+
+def system_to_stdlib_linkflags(system):
+    return _SYSTEM_TO_STDLIB_LINKFLAGS[system]
 
 def triple_to_constraint_set(triple):
     component_parts = triple.split("-")
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -197,14 +197,9 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
         build_info,
     )
 
-def get_linker_and_args(ctx, rpaths):
-    if (len(BAZEL_VERSION) == 0 or
-        versions.is_at_least("0.18.0", BAZEL_VERSION)):
-        user_link_flags = ctx.fragments.cpp.linkopts
-    else:
-        user_link_flags = depset(ctx.fragments.cpp.linkopts)
-
+def get_cc_toolchain(ctx):
     cc_toolchain = find_cpp_toolchain(ctx)
+
     kwargs = {
         "ctx": ctx,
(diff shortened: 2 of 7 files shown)


rust/platform/triple_mappings.bzl
```diff
@@ -66,6 +66,41 @@ _SYSTEM_TO_DYLIB_EXT = {
...
+# See https://github.com/rust-lang/rust/blob/master/src/libstd/build.rs
+_SYSTEM_TO_STDLIB_LINKFLAGS = {
+    # TODO: Right now bazel cc rules does not specify the exact flag setup needed for calling out
+    # TODO: system libs, that is we dont know given a toolchain if it should be, for example,
+    # TODO: `-lxxx` or `/Lxxx` or `xxx.lib` etc.
+    # TODO: We include the flag setup as they are _commonly seen_ on various platforms
+    # TODO: with a cc_rules style override for people doing things like mingw on windows.
+    # TODO: If you are reading this ... sorry! set the env var `BAZEL_RUST_STDLIB_LINKFLAGS` to
+    # TODO: what you need for your specific setup, for example like so
+    # TODO: `BAZEL_RUST_STDLIB_LINKFLAGS="-ladvapi32:-lws2_32:-luserenv"`
```
damienmg, line 78: TODO format: just put one TODO(#bug): and then you can do a multiline description. having one TODO per line is hard to read and make me think that this is several things to do.

rust/platform/triple_mappings.bzl
```diff
@@ -66,6 +66,41 @@ _SYSTEM_TO_DYLIB_EXT = {
...
+
+    "freebsd": ["-lexecinfo", "-lpthread"],
+    # TODO - This ignores musl although I am not 100% sure what bazel thinks of musl anyhow
```
damienmg, line 81:
```suggestion
    # TODO: This ignores musl. What does Bazel thinks of musl?
```

rust/platform/triple_mappings.bzl
```diff
@@ -66,6 +66,41 @@ _SYSTEM_TO_DYLIB_EXT = {
...
+    # NOTE - rust stdlib `build.rs` treats android as a subset of linux, rust rules treat android
+    # NOTE - as its own system
```
damienmg, line 83:
```suggestion
    # Note: Rust's stdlib `build.rs` treats android as a subset of linux, rust rules treat android
    # as its own system.
```

rust/platform/triple_mappings.bzl
```diff
@@ -66,6 +66,41 @@ _SYSTEM_TO_DYLIB_EXT = {
...
+    "emscripten": [],
+    "nacl": [],
+    "bitrig": [],
+    "dragonfly": ["-lpthread"],
+    "netbsd": ["-lpthread", "-lrt"],
+    "openbsd": ["-lpthread"],
+    "solaris": ["-lsocket", "-lposix4", "-lpthread", "-lresolv"],
+    "illumos": ["-lsocket", "-lposix4", "-lpthread", "-lresolv", "-lnsl", "-lumem"],
+    "fuchsia": ["-lzircon", "-lfdio"],
+    # TODO - feature backtrace is how unwind happens ...
```
damienmg, line 99: What does this TODO means?

rust/private/rustc.bzl
```diff
@@ -198,26 +198,35 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
...
         unsupported_features = ctx.disabled_features,
         **kwargs
     )
+
+    return cc_toolchain, feature_configuration
+
+def get_cc_user_link_flags(ctx):
+    if (len(BAZEL_VERSION) == 0 or
+        versions.is_at_least("0.18.0", BAZEL_VERSION)):
+        return ctx.fragments.cpp.linkopts
```
damienmg, line 223: opt: Drop support for Bazel < 1?

rust/private/rustc.bzl
```diff
@@ -513,6 +528,43 @@ def rustc_compile_action(
...
+        )
+
+    link_input = cc_common.create_linker_input(
+        owner = ctx.label,
+        libraries = depset([library_to_link]),
+        user_link_flags = depset(toolchain.stdlib_linkflags),
+    )
+
+    linking_context = cc_common.create_linking_context(
+        # TODO - What to do for no_std?
```
damienmg, line 559: What do you think no_std has especially?

I only know no_std for embedded code, this also means the C++ toolchain will be different and the libs list will also match, won't it?

GregBowyer: Its mostly that `std` is what brings in these libs, its possible I think for someone to be using `no_std` on for instance linux for something like a kernel module or maybe an NSS hook as well as some things like SGX. I suspect that its not always going to be a toolchain thing it might be a context of compilation instead. I would see it as someone compiling C code with `-nostdlib`.

With that said, its up to how Bazel wants to view such configurations, if its expected that someone makes up a toolchain config for say `no_std` linux x86_64 and its given unique identification against toolchain matching then I fully agree that this todo can be removed.

What do you think? Is the philosophy of Bazel to assume that it is a unique toolchain or is it to allow flag flipping on the current toolchain configuration?

rust/platform/triple_mappings.bzl
GregBowyer: Like this ?

```python
"""TODO(#some_bug)
Right now bazel cc rules does not specify the exact flag setup needed for calling out
system libs, that is we dont know given a toolchain if it should be, for example,
`-lxxx` or `/Lxxx` or `xxx.lib` etc.

We include the flag setup as they are _commonly seen_ on various platforms
with a cc_rules style override for people doing things like mingw on windows.

If you are reading this ... sorry! set the env var `BAZEL_RUST_STDLIB_LINKFLAGS` to
what you need for your specific setup, for example like so
`BAZEL_RUST_STDLIB_LINKFLAGS="-ladvapi32:-lws2_32:-luserenv"`
"""
```
?

rust/platform/triple_mappings.bzl
GregBowyer: Bad comment on my part.

If rust stdlib is compiled for cloudabi with the `backtrace` feature it includes `-lunwind` but if the feature is not compiled in it misses this linkage. I am not sure which is the common configuration or how we encode it as a lib dependency.

I might not matter since currently its a seemingly esoteric environment that I am pretty sure Bazel wont build for all that well, I was mostly blindly translating the stdlib `build.rs`.

rust/platform/triple_mappings.bzl
damienmg: Yes :) docstring is the best for multiline TODOs.

rust/platform/triple_mappings.bzl
damienmg: Ok then maybe remove this TODO?

rust/private/rustc.bzl
damienmg: I would see the no_std as a different toolchain, since it targets a different architecture (kernel mode vs userland is already a different platform in my view), but I am up for both case. I don't think Bazel has any recommendation on that (although it has the concept of platform that include more than the CPU).

For my own curiosity: Is there any actual kernel code in Rust?

rust/platform/triple_mappings.bzl
GregBowyer: ah sorry I already updated it in keeping with the style that was in the file

rust/private/rustc.bzl
GregBowyer: https://ldpreload.com/p/kernel-modules-in-rust-lssna2019.pdf ;)

rust/platform/triple_mappings.bzl
```diff
@@ -66,6 +66,47 @@ _SYSTEM_TO_DYLIB_EXT = {
     "unknown": ".wasm",
 }
 
+# See https://github.com/rust-lang/rust/blob/master/src/libstd/build.rs
+_SYSTEM_TO_STDLIB_LINKFLAGS = {
+    # TODO(https://github.com/bazelbuild/rules_cc/issues/75):
```
damienmg, line 71:
```suggestion
    # TODO(bazelbuild/rules_cc#75):
```

rust/platform/triple_mappings.bzl
damienmg: No worries, using `#` is also fine.

GregBowyer:
Rebased to the latest and made more comment changes as per review

rust/platform/triple_mappings.bzl
GregBowyer: I elborated the comment a little

rust/private/rustc.bzl
GregBowyer: I think in a followup so its more of an atomic commit and we can get all bazel versioning in one swoop

damienmg: Make sense

PiotrSikora:
FYI, this broke the `out_binary = True` flag for Wasm introduced in #312. See: #386.

PR #365, opened 2020-07-01
fix: Download llvm tools for nightly builds after 2020-05-22

Arm1stice:
After nightly Rust builds after 2020-05-22 do not have the `libLLVM-10-*version*-nightly.so` dylib included with the normal rust-nightly gzip due to rust-lang/rust#70838. Therefore, nightly builds after this date need to download and extract the llvm-tools gzip to ensure this dylib is available.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -431,6 +431,25 @@ def _load_rustc_dev_nightly(ctx, target_triple):
 
     return
 
+def _load_llvm_tools(ctx, target_triple):
+    """Loads the llvm tools
+
+    Args:
+      ctx: A repository_ctx.
+      target_triple: The rust-style target triple of the tool
+    """
+
+    load_arbitrary_tool(
+        ctx,
+        iso_date = ctx.attr.iso_date,
+        target_triple = target_triple,
+        tool_name = "llvm-tools",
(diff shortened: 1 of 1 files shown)


rust/repositories.bzl
```diff
@@ -441,6 +460,10 @@ def _rust_toolchain_repository_impl(ctx):
     if ctx.attr.rustfmt_version:
         BUILD_components.append(_load_rustfmt(ctx))
 
+    # Nightly Rust builds after 2020-05-22 need the llvm-tools gzip to get the libLLVM dylib
+    if ctx.attr.version == "nightly" and int(ctx.attr.iso_date.replace("-", "")) > 20200522:
```
damienmg, line 464: There is no need to convert to int, the alphabetical ordering should just work:

```suggestion
    if ctx.attr.version == "nightly" and ctx.attr.iso_date > "2020-05-22":
```

This is more readable and both operations are `O(len(ctx.attr.iso_date))` anyway

Arm1stice: Interesting! Didn't realize that would actually work, but it makes sense.

Arm1stice:
I went ahead and committed your suggestion. Would it be possible to get this merged today?

PR #371, opened 2020-07-17
Update Skylib dependency

laurentlb:
Updating Skylib makes the repository build with --incompatible_disable_depset_items, which is planned for Bazel 4.0.


damienmg:
LGTM

PR #373, opened 2020-07-17
Lint: Pull old warning for `rust_test`

GregBowyer:
Its minor but I think time has passed on this usage, and we should pull the warning for backward compat


damienmg:
Thanks!

PR #374, opened 2020-07-18
Fix rustdoc path in rustdoc_test with --nolegacy_external_runfiles option

oxidase:
Using File.path breaks builds with --nolegacy_external_runfiles option
as the gnerated script has use `external/rust/bin/rustdoc` that
referrers to legacy runfiles and File.short_path uses `../rust/bin/rustdoc`


oxidase:
@googlebot I signed it!

damienmg:
Thanks!

PR #379, opened 2020-07-22
Fix `CcDep` info related to `rust_test`

GregBowyer:
This is a small fix I discovered where a `staticlib` or `cdylib` claims to export `CcInfo` for test crates.

Instead of mapping this all the way through (I am not sure what depending on a `rust_test` would mean) we just ignore / disallow `CcInfo` for test contexts.

### DIFF
--- a/examples/ffi/c_calling_rust/lib.rs
+++ b/examples/ffi/c_calling_rust/lib.rs
@@ -2,3 +2,13 @@
 pub extern fn my_favorite_number() -> i32 {
     4
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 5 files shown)


examples/ffi/c_calling_rust/lib.rs
```diff
@@ -2,3 +2,13 @@
 pub extern fn my_favorite_number() -> i32 {
     4
 }
+
+#[test]
```
dfreese, line 6:
```suggestion
#[cfg(test)]
```

GregBowyer: whoops good catch!

dfreese:
Thanks!

PR #399, opened 2020-09-08
Cleanup platform configurations

UebelAndre:
This is a slight cleanup pull request but also adds the `@io_bazel_rules_rust//rust/platform:unix` and `@io_bazel_rules_rust//rust/platform:wasm32-unknown-unknown` targets


damienmg:
Ok the PR LGTM but I want to find a better title for the commit. Maybe `Cleanup platform configurations`? WDYT?

UebelAndre:
@damienmg Is that preferred over the current commit message?

> Add 'unix' platform

I'm happy to change it. I've definitely noticed I've gotten lazy in my commit messages 😅

damienmg:
It not only add unix platform, but also fix the `wasm32-unknown-unknown`, this is not for the commit message, more for the actual message I'll put when I hit the merge button. I am fine with putting `Add 'unix' platform` but that does not seems to reflect what is done here.

UebelAndre:
Oh man! I totally forgot I added that here, I'll close #402. Sorry about that, haven't been sleeping well 😅

damienmg:
Anyway, what do you think about the name of this PR?

damienmg:
Thanks LGTM!

UebelAndre:
@damienmg Is it the name of the PR that shows as the commit message? I've updated the name and top comment as well as the commit message for this PR. Thoughts?

damienmg:
Yes if you create a PR with a single commit, the title and description of that PR is going to be the title and description of the commit (we can change it when we do the merge). If there is multiple commit, it create a bullet list with all the commit title as the commit description.

PR #402, opened 2020-09-09
Added 'wasm32-unknown-unknown' platform triple

UebelAndre:
Just as the title says, this adds the `@io_bazel_rules_rust//rust/platform:wasm32-unknown-unknown` target. I feel this makes the interactions with wasm more consistent with other platforms.


UebelAndre:
This is implemented by #399

PR #407, opened 2020-09-15
Removed duplicate dependencies of examples in the top level workspace

UebelAndre:
This PR adds `rust_workspace` which is meant to define transitive dependencies as well as cleans up some duplicate files that were in the root of the project but only used by examples.

## Updated documentation
Note the updated documentation in [docs/index.md](https://github.com/UebelAndre/rules_rust/blob/workspace-structure/docs/index.md)

## Added `bazel_skylib` to `rust_repositories`
It appears there are some uses of `bazel_skylib` within these rules but there aren't really any guarantees that that repository is available. This may be my fault with the changes made in #399 to [@io_bazel_rules_rust//rust/platform:platform.bzl](https://github.com/bazelbuild/rules_rust/blob/master/rust/platform/platform.bzl#L3) but I feel the `bazel_skylib` dependency should be conditionally added

## Added `rust_workspace`
I feel this is much nicer on the users's end as it gives them a single place to go to look at dependencies and is nicer on the repo because it means dependencies can more easily updated. I'm not the biggest fan of the name `rust_workspace` so if this was an acceptable concept, I'm happy to take suggestions on what the name should be called.


UebelAndre:
Is [renovate.json](https://github.com/bazelbuild/rules_rust/blob/master/renovate.json) still used?

UebelAndre:
@damienmg Hey, if you get a moment do you think you could check out this PR? 🙏

damienmg:
LGTM.

UebelAndre:
@damienmg i think you'll need to make a new release and update the docs since `rust_workspace` isn't in that version in the docs 😞

damienmg:
Updating the doc is a PR, mind sending one? You just need to run the shellscript. There is no real release on this repository otherwise.

UebelAndre:
> Is [renovate.json](https://github.com/bazelbuild/rules_rust/blob/master/renovate.json) still used?

@damienmg Also, do you know the answer to this question?

damienmg:
I don't think it ever worked, @nlopezgi might have more insights.

nlopezgi:
> I don't think it ever worked, @nlopezgi might have more insights.

iirc (but my memory is fuzzy) It did work for a bit (i do remember we had it for rules_docker and rules_k8s), but then was turned off and/or replaced by some other solution (one provided by Bazel CI, iirc?). But renovate bot has likely not been in use for some time if it ever was used at all.

UebelAndre:
@nlopezgi @damienmg Do you see any value in keeping that there? Or would you like me to open a PR to remove it?

damienmg:
This can be deleted, a PR would be welcomed :)

PR #409, opened 2020-09-17
Extract cargo from rust toolchain tar

illicitonion:
This allows for `cargo` to be used from scripts which may want to invoke
`cargo`, while coupling the version used with the rest of the toolchain
used for the build.


a1ph:
@illicitonion Any plans to make it available in `ToolchainInfo`?

illicitonion:
> @illicitonion Any plans to make it available in `ToolchainInfo`?

I don't need it for my use-case, but feel free to do so if you do! :)

PR #418, opened 2020-09-26
Made examples compatible with cargo-raze

UebelAndre:
Updated examples to be compatible with cargo raze which I then used to regenerate dependencies.


examples/ffi/ffi_examples.bzl
```diff
@@ -0,0 +1,20 @@
+# Copyright 2015 The Bazel Authors. All rights reserved.
```
damienmg, line 1: unless this is branched code.

```suggestion
# Copyright 2020 The Bazel Authors. All rights reserved.
```

examples/hello_runfiles/BUILD
```diff
@@ -1,9 +1,20 @@
-package(default_visibility = ["//visibility:public"])
+# Copyright 2015 The Bazel Authors. All rights reserved.
```
damienmg, line 1:
```suggestion
# Copyright 2020 The Bazel Authors. All rights reserved.
```

damienmg (commented):
I am not sure we want to do such a change. What would be the gain to do so?

UebelAndre:
> I am not sure we want to do such a change. What would be the gain to do so?

I feel it makes the code more representative of what users would actually try to. It seemed like the examples might have used an older version of `raze` to generate the files that are currently committed. All I really did was add the necessary cargo files to be able to regenerate new ones. Can you talk more about why you might not want to make such a change?

damienmg:
> > I am not sure we want to do such a change. What would be the gain to do so?
> 
> I feel it makes the code more representative of what users would actually try to. It seemed like the examples might have used an older version of `raze` to generate the files that are currently committed. All I really did was add the necessary cargo files to be able to regenerate new ones. Can you talk more about why you might not want to make such a change?

Although this is likely that the sys example was indeed make using raze (@illicitonion ?), the FFI ones were not. This adds more complexity for the ffi examples as the libc was handwritten and a simple use-case. I don't see value in putting it in raze format (it would actually remove an example of hand-writting a dependency).

UebelAndre:
> Although this is likely that the sys example was indeed make using raze (@illicitonion ?), the FFI ones were not. This adds more complexity for the ffi examples as the libc was handwritten and a simple use-case. I don't see value in putting it in raze format (it would actually remove an example of hand-writting a dependency).

Hmm, I can remove the use of raze for the FFI example though I don't quite feel any information was lost in updating it to use raze nor do I feel it adds complexity (at least for users). I don't think it's valuable to have any example explicitly demonstrate that you can hand author BUILD files to be able to pull in dependencies. This is something that seems obvious given Bazel's repository rules. Though, that said, the FFI example uses just one dependency. I would care more if it were multiple. What do you think of the current state?

illicitonion:
Yeah, the sys example was generated with Raze and then simplified a little.

I don't have strong feelings either way on this PR :)

UebelAndre:
@damienmg after more thought I agree with you about the did examples 😀 I think it's better now

damienmg (approved):

damienmg:
Nice thanks!

PR #419, opened 2020-10-01
Improved internal and external documentation

UebelAndre:
Changes:
- Added documentation containing type information for a large number of macros
- Cleans up `doc` attributes so stardoc renders cleaner
- Adds new pages to the `docs/update_docs.sh` output
- Rearranged a few macros to reduce code duplication
- Fixed some incompatibility warnings/buildifier defects

@acmcarther @mfarrugi @damienmg @smklein @dfreese
I would hugely appreciate if all code owners would take a look at 548d62ff981ab3569a10d6a515ac4f555d4ff134 and provide suggestions on how the documentations can be improved. I'm very bad at writing docs and this was largely a learning exercise for me so some of the docs may not be as clear as they should be. There are a few I was totally unsure of what to write (namely `get_compilation_mode_opts`, `_determine_lib_name`, `rust_toolchain`, and `RustProtoInfo`).

Again, It'd help out tremendously to have these docs updated. Looking forward to suggestions 😄

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -24,7 +24,7 @@ load(
     "@io_bazel_rules_rust//rust:private/rust.bzl",
     "crate_root_src",
 )
-load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
+load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain", "determine_output_hash")
 
 _rust_extensions = [
     "rs",
@@ -85,7 +85,7 @@ def _clippy_aspect_impl(target, ctx):
         feature_configuration,
         crate_info,
         dep_info,
-        output_hash = repr(hash(root.path)),
+        output_hash = determine_output_hash(root),
         rust_flags = [],
         out_dir = out_dir,
         build_env_file = build_env_file,
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -67,3 +67,19 @@ def _path_parts(path):
     """
     path_parts = path.split("/")
     return [part for part in path_parts if part != "."]
+    
+def get_lib_name(lib):
+    """Returns the name of a library artifact, eg. libabc.a -> abc
+
+    Args:
+        lib (File): A library file
+
+    Returns:
+        str: The name of the library
+    """
+    libname, ext = lib.basename.split(".", 2)
+
+    if libname.startswith("lib"):
+        return libname[3:]
+    else:
+        return libname
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -12,18 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""
-Utility functions not specific to the rust toolchain.
-"""
+"""Utility functions not specific to the rust toolchain."""
 
 def find_toolchain(ctx):
-    """Finds the first rust toolchain that is configured."""
+    """Finds the first rust toolchain that is configured.
+
+    Args:
+        ctx (ctx): The ctx object for the current target.
+
+    Returns:
+        ToolchainContext: A Rust toolchain context.
+    """
     return ctx.toolchains["@io_bazel_rules_rust//rust:toolchain"]
 
-def relative_path(src_path, dest_path):
-    """Returns the relative path from src_path to dest_path."""
-    src_parts = _path_parts(src_path)
-    dest_parts = _path_parts(dest_path)
+# TODO: Replace with bazel-skylib's `path.dirname`. This requires addressing some
+# dependency issues or generating docs will break.
+def relativize(path, start):
+    """Returns the relative path from start to path.
+
+    Args:
+        path (str): The path to relativize.
+        start (str): The ancestor path against which to relativize.
+
+    Returns:
+        str: The portion of `path` that is relative to `start`.
+    """
+    src_parts = _path_parts(start)
+    dest_parts = _path_parts(path)
     n = 0
     done = False
     for src_part, dest_part in zip(src_parts, dest_parts):
@@ -41,11 +56,11 @@ def relative_path(src_path, dest_path):
 def _path_parts(path):
     """Takes a path and returns a list of its parts with all "." elements removed.
 
-    The main use case of this function is if one of the inputs to _relative()
+    The main use case of this function is if one of the inputs to relativize()
     is a relative path, such as "./foo".
 
     Args:
-      path_parts: A list containing parts of a path.
+      path: A list containing parts of a path.
 
     Returns:
       Returns a list containing the path parts with all "." elements removed.
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -83,3 +83,14 @@ def get_lib_name(lib):
         return libname[3:]
     else:
         return libname
+
+def determine_output_hash(crate_root):
+    """Generates a hash of the crate root file's path.
+
+    Args:
+        crate_root (File): The crate's root file (typically `lib.rs`).
+
+    Returns:
+        str: A string representation of the hash.
+    """
+    return repr(hash(crate_root.path))
--- a/proto/proto.bzl
+++ b/proto/proto.bzl
@@ -44,7 +44,7 @@ load(
 load("@io_bazel_rules_rust//rust:private/rustc.bzl", "CrateInfo", "rustc_compile_action")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 
-RustProtoProvider = provider(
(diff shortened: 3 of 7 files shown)


docs/all.bzl
```diff
@@ -56,3 +78,11 @@ rust_wasm_bindgen_toolchain = _rust_wasm_bindgen_toolchain
...
 # including.
 # rust_wasm_bindgen_repositories = _rust_wasm_bindgen_repositories
+
+rust_repositories = _rust_repositories
+rust_repository_set = _rust_repository_set
+rust_toolchain_repository = _rust_toolchain_repository
+rust_toolchain_repository_proxy = _rust_toolchain_repository_proxy
+
+rust_workspace = _rust_workspace
+bazel_version = _bazel_version
```
damienmg, line 88: Do not expose bazel_version. I don't think it is needed.

rust/private/utils.bzl
```diff
@@ -12,43 +12,59 @@
...
       Returns a list containing the path parts with all "." elements removed.
     """
     path_parts = path.split("/")
     return [part for part in path_parts if part != "."]
+
+def relativize(path, start):
+    """Returns the relative path from start to path.
+
+    TODO: Replace with bazel-skylib's `path.dirname`. This requires addressing some
+    dependency issues or generating docs will break.
```
damienmg, line 47: This TODO should not be in the docstring.

rust/private/utils.bzl
```diff
@@ -12,43 +12,59 @@
...
-      path_parts: A list containing parts of a path.
+      path: A list containing parts of a path.
 
     Returns:
       Returns a list containing the path parts with all "." elements removed.
     """
     path_parts = path.split("/")
     return [part for part in path_parts if part != "."]
+
+def relativize(path, start):
```
damienmg, line 43: Why move the position of the method?

damienmg (changes requested):
Comment by change:

- __Renamed RustProtoProvider -> RustProtoInfo__: I don't see the use of this change, Provider is a more common name and I find it more explicit
- __Partially addressed legacy provider syntax issue__: LGTM
- __Renamed AliasableDep -> AliasableDepInfo__: Can we rename as `AliasedDepProvider` instead?
- __Renamed relative_path -> relativize__: Some nit inlined
- __Moved get_lib_name to //rust:private/utils.bzl__: This change is adding an unused function, drop?
- __Updated copyright__: Please do not do this, we should not update existing Copyright lines.
- __Moved _determine_output_hash so it can be used in other places__: LGTM
- __Updating documentation__: this change is too big, please extract to a separate PR.
- __Added more pages to the outward facing docs__: Some nit inlined
- __Regenerate documentation__: LGTM

UebelAndre:
>     * **Renamed RustProtoProvider -> RustProtoInfo**: I don't see the use of this change, Provider is a more common name and I find it more explicit

Buildifier claims this violates naming conventions: https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#name-conventions

UebelAndre:
>     * **AliasableDepInfo**: Can we rename as `AliasedDepProvider` instead?

Again, this violates https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#name-conventions

UebelAndre:
>     * **Moved get_lib_name to //rust:private/utils.bzl**: This change is adding an unused function, drop?

If it was unused, I would have removed it. It's used in `rustc.bzl` and `rustdoc_test.bzl`

rust/private/clippy.bzl
```diff
@@ -180,10 +182,10 @@ rust_clippy = rule(
     attrs = {
         "deps": attr.label_list(aspects = [rust_clippy_aspect]),
     },
-    doc = """
+    doc = """\
```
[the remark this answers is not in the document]
UebelAndre: So there aren't unnecessary new lines when it gets rendered, I thought it would be a less invasive change then moving the first line here. Would you like me to do this instead?

rust/private/utils.bzl
UebelAndre: it reduces the amount of code in `rustc.bzl` which I find valuable since there is a lot going on there.

cargo/cargo_build_script.bzl
```diff
@@ -114,33 +124,46 @@ def _cargo_build_script_run(ctx, script):
         ),
     ]
 
-def _build_script_impl(ctx):
-    return _cargo_build_script_run(ctx, ctx.executable.script)
-
 _build_script_run = rule(
-    _build_script_impl,
+    doc = (
```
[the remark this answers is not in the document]
UebelAndre: I feel it's important to have each rule/macro have it's own documentation. The whole reason I made these changes was because I had to jump around all over the code to figure out what anything was doing. Macros should document what they do as should rules. 

When I wrote 
```
# The source of truth will the `cargo_build_script` macro until stardoc
# implements documentation inheritence. See https://github.com/bazelbuild/stardoc/issues/27
```
I was more trying to inform readers that if the attributes change, then they should look at `cargo_build_script` to ensure the documentation for the wrapper appropriately reflects the changes.

rust/private/utils.bzl
damienmg: In this change the function simply change position inside utils.bzl. Its not changing file. Side-note: a general good methodology when relocating is not changing the code at all in the same change just moving it so we can easily see when we change something in one function.

rust/private/clippy.bzl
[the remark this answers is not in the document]
damienmg: For the first line, it may looks nicer to remove it but keeping the file as is is good for me too.

cargo/cargo_build_script.bzl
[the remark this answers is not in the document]
damienmg: "The source of truth..." is perfectly fine with me but should not be a docstring (it's not the documentation of the API, more like a TODO). I don't think we should have the attribute documented in 5 place, this introduce too much toil and leads to wrong documentation at some point there will be a change that update only of those documentation.

damienmg:
> > ```
> > * **Renamed RustProtoProvider -> RustProtoInfo**: I don't see the use of this change, Provider is a more common name and I find it more explicit
> > ```
> 
> Buildifier claims this violates naming conventions: https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#name-conventions

Meh :/ This is a bad convention, anyway that's the convention so this change LGTM.


> > ```
> > * **Moved get_lib_name to //rust:private/utils.bzl**: This change is adding an unused function, drop?
> > ```
> 
> If it was unused, I would have removed it. It's used in `rustc.bzl` and `rustdoc_test.bzl`

This change only add the function without any call site. Unless the code was broken before this method is not used anywhere. You probably forgot to update the load statements and remove the original place of the code.

rust/private/utils.bzl
UebelAndre: Ah, sorry, I misread the diff just now. I fixed the positioning and that's a good thing to call out. I'll be better about that in future changes 😄

UebelAndre:
> This change only add the function without any call site. Unless the code was broken before this method is not used anywhere. You probably forgot to update the load statements and remove the original place of the code.

Ah I see, I've updated this now. I didn't completely break out this change from the larger one. This has been updated

rust/private/utils.bzl
UebelAndre: The positioning has been updated. Hopefully this is more clear

UebelAndre:
>     * Please do not do this, we should not update existing Copyright lines.

Dropped but can you update it then?

UebelAndre:
>     * **Updating documentation**: this change is too big, please extract to a separate PR.

Moving to another PR

damienmg:
> > ```
> > * Please do not do this, we should not update existing Copyright lines.
> > ```
> 
> Dropped but can you update it then?

We should not touch to the copyright line. The date should be the date of creation of the file as per guidance from the opensource office at google.

rust/private/utils.bzl
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015 The Bazel Authors. All rights reserved.
+# Copyright 2020 The Bazel Authors. All rights reserved.
```
damienmg, line 1:
```suggestion
# Copyright 2015 The Bazel Authors. All rights reserved.
```

cargo/cargo_build_script.bzl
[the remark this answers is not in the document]
UebelAndre: I agree. I've updated this.

I don't think an attribute should be documented in 5 places either. But I do feel strongly about there being at least some documentation that says:

> Go look at `this_specific_function` for details

This change was removed form this PR though so should continue in the next

damienmg (approved):
Just one nit left.

Issue #412, opened 2020-09-20
Persistent worker for Rust rules


nikhilm:
Hello,

I have a simple [persistent worker](https://github.com/nikhilm/rustc-worker) for Bazel that I've been playing with on some small projects.

Would the Rust rules be interested in integrating this? Since rustc does not support Java/TypeScript/Scala like "Compiler as an API", this really only takes advantage of in-crate incremental compilation introduced in rust 1.24. This is achieved by passing `--codegen=incremental=/path/to/dir`. The directory is named using a combination of `hash(path to rustc wrapper)` and the workspace name, with the goal of:
1. Being shared across multiple worker instances when they are spawned by Bazel. This is why one cannot use something like the `tempdir` crate (would be deleted each time a worker shut down) or just a randomly named temp file (would not be shared across worker processes).
2. Not being shared across workspaces.
3. Not being shared across compiler updates. This should probably actually depend on the path to rustc itself, instead of the wrapper. I can fix that.

The questions I have are:
1. Is this kind of cache sharing sound design? I don't know enough of Bazel internals nor rustc internals to have high confidence in answering yes to this.
2. Since this project is itself in Rust, how would something like this be integrated into the rules? Options:
   1. Use pre-built binaries built with Cargo. (What I'm doing now for simplicity.)
   2. Re-write in C++.
   3. Have certain _internal_ switches to the Rust rules to build the worker itself using the rustc wrapper directly.

If this all sounds reasonable, I can send relevant PRs.
In particular, it would also be nice to have the worker and the current process wrapper combine their functionality so that an additional process spawn is saved (Go from `worker -> process wrapper -> rustc` to `worker -> rustc`.

nikhilm:
If we are to follow the behavior of Cargo exactly, we would also want to include the compilation mode in the incremental cache path (to reflect the Cargo default of `target/{debug,release}/incremental`.

damienmg:
That sounds like a great option to offer better build speed.

mfarrugi:
> Is this kind of cache sharing sound design? I don't know enough of Bazel internals nor rustc internals to have high confidence in answering yes to this.

I think 'sound' is a pretty low bar, depending on what you mean. Reproducibility might be a good measure.

[...]

PR #421, opened 2020-10-03
Persistent worker support

nikhilm:
@mfarrugi 
This is the outline of what it would look like to plug in the worker. There are a few thorny issues that need solutions:

1. The need to have `raze` emit `_no_worker_` versions of the rules for _all dependencies_ of the worker. This will need changes to raze to allow it to emit custom rule names and imports.

2. I tried this with a [project that uses Raze+rules_rust](https://github.com/nikhilm/ninja-rs/compare/master..bazel) itself and quickly ran into [the shared deps problem](https://github.com/bazelbuild/rules_rust/blob/169476f772132723b2cc819c96209c6fbce3e366/proto/README.md#customizing-dependencies). Not only does the user of these rules currently need to run the (documented) extra steps to merge their own deps and the protobuf deps, they also need to pull in the protobuf patch (could be mitigated by running the build.rs?) and patch their own copies of the generated BUILD files to use the `_no_worker_` rules. This is a significant barrier to usage. Of course, if raze was modified, there is probably a way to annotate it instead, so that it generates no-worker variants for certain crates and their transitive deps. Ideally it should generate both worker and no-worker variants and then use the appropriate ones, so that any deps shared between the worker and user code (e.g. `protobuf`) can still use the with-worker variant to build the protobuf crate for the user code.

3. The worker actions will need a new mnemonic, otherwise changing the strategy for `Rustc` will make the worker compilation fail, since it uses the same actions, but does not support the `worker` strategy. This is an easy fix.

At this point, it is almost like letting the worker build with Cargo, releasing binaries and then having external repositories for the worker, the way we do for `rustc` itself may be another alternative. What do you think?


nikhilm:
Related to #412

mfarrugi:
> At this point, it is almost like letting the worker build with Cargo, releasing binaries and then having external repositories for the worker, the way we do for rustc itself may be another alternative. What do you think?

That sounds fine, cargo-raze is a much closer analog. We're not currently fetching raze as part of rules_rust, but we probably should.

dae:
@nikhilm unfortunately I'm fairly new to Bazel and thus do not have anything to add on how the problems you listed may be resolved, but I just wanted to say thank you for working on this - waiting ~5 secs for my crate to compile in the trivial change case instead of ~15 secs is a lot nicer. In case it's useful, here is your branch rebased over the latest master: https://github.com/ankitects/rules_rust/tree/persistentworker

nikhilm:
@mfarrugi (or other reviewers)
This seems ready to go for Linux x86-64 host/exec machines. The examples tests pass (I've enabled worker by default for them).

@dae thanks for the macOS build! Would you mind sending a PR on top of this (after this is merged) to add that as a macOS toolchain? I'm happy to assist.

mfarrugi:
> @mfarrugi (or other reviewers)
> This seems ready to go for Linux x86-64 host/exec machines. The examples tests pass (I've enabled worker by default for them).

This looks good to me. Would like for at least one other person to chime in, though.

> I have one CLA issue. I have 2 emails (personal and work) associated with Google CLAs. For some reason, at least in the buildkite runs, the thing seems to be picking up my work email (`BUILDKITE_BUILD_CREATOR_EMAIL`). I have done this on my own time and on my own hardware. Is this something that can be fixed? AFAIK all my commits here are from my personal email.

I don't think I can do anything about this :|

mfarrugi (approved):

UebelAndre:
Should docs be updated for this change?

nikhilm:
Good point! I added some docs.

nikhilm:
friendly bump @UebelAndre

UebelAndre:
Alley-oop @mfarrugi 

(Thanks for adding docs 😃)

UebelAndre:
Any updates here? 😃

nikhilm:
@mfarrugi when you have some time could you merge this? Thanks!

docs/index.md
```diff
@@ -69,6 +69,15 @@ Similarly, `rustfmt_version` may also be configured:
 ```python
 rust_repositories(rustfmt_version = "1.4.20")
 ```
+# Using Bazel Persistent Workers
+
+Iterating on Rust code may benefit from [incremental compilation](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/incremental-compilation-for-faster-compiles.html). This is supported by using a [Bazel Persistent Worker](https://docs.bazel.build/versions/master/persistent-workers.html). While Bazel can determine what needs to be rebuilt at a crate level, the compiler can speed up building a single crate by sharing information across runs. It does this by storing intermediate information in a directory across invocations. This is enabled by default in Cargo. Persistent workers bring this feature to Bazel.
+
+The Rust rules have preliminary support for workers. Pass `use_worker = True` to enable this when available.
```
mfarrugi, line 76: This is linux only, right?

docs/index.md
```diff
@@ -69,6 +69,15 @@ Similarly, `rustfmt_version` may also be configured:
...
+
+```python
+rust_repositories(use_worker = True)
```
mfarrugi, line 79: I don't see why this shouldn't be the default, was there a reason to be conservative about this?

worker/README.md
```diff
@@ -0,0 +1,14 @@
...
+2. Figuring out a good strategy for dependencies. Since Bazel doesn't really have transitive dependencies, attempts to build this worker from source necessarily require users of these rules to register all the external repositories for the worker in their `WORKSPACE`. This could cause collisions with other dependencies. In addition, if the `_no_worker_` approach above is adopted, users will lose the benefits of workers for those dependencies (like `protobuf`) shared between the worker and their code. There is no satisfactory solution for this right now.
+
+## How about rewriting the worker in C++?
```
mfarrugi, line 12: Does the worker benefit from being written in a non-gc'd language? Curious about whether python is a reasonable choice, given other utilities in rules_rust have been written in python.

mfarrugi:
I was really hoping for a second set of eyes to at least take a high-level look. In particular I'm not a huge fan of depending on a linux-only binary download, and while ultimately I think that is fine I can imagine other opinions.

Does anyone care to opine on default-ness or any of the other high level aspects of this addition? cc @dfreese @illicitonion @UebelAndre

docs/index.md
dae: It's pulling in and running a binary hosted on a third-party repo, so off by default is probably the best option.

docs/index.md
dae: I posted a Mac version above - but each platform would need its own binary at the moment.

dae:
Avoiding the need for a binary would be ideal. Python could work, though I don't see any Python in rules_rust at the moment?

Alternatively perhaps the cyclic dependency issue could be solved by having a separate rust_binary_noworker() rule that does not take the worker toolchain as an argument.

Protobuf dependencies could be avoided by using JSON: https://blog.bazel.build/2020/11/11/json-workers.html

docs/index.md
nikhilm: rustc-worker also has a mac build available now. I can update this PR with an entry for that.

docs/index.md
nikhilm: Yes, what @dae said. Plus given that only 2(?) people have actually used it, it seemed very alpha.

worker/README.md
nikhilm: No, there is nothing preventing writing this in Python for someone with such an interest. I did not find any references to Python rules in this repo. I suppose as a bootstrap language (i.e. one that already has Bazel rules for it), that is a better alternative than C++.

nikhilm:
> Alternatively perhaps the cyclic dependency issue could be solved by having a separate rust_binary_noworker() rule that does not take the worker toolchain as an argument.

This did not work out due to the requirement that pops up about not being able to have any cargo-raze based dependencies, without significant changes to cargo-raze itself.

Note that JSON workers are a _very_ new thing at this point, which would necessarily tie rules_rust to Bazel 3.7 or newer. I'm not sure if that is what we want for rules_rust. Perhaps the `use_worker` flag should still be preserved in such a case.

Re: writing this in Python, or modifying raze - I don't have the interest to do those changes, due to incidental complexity (Bazel transition and cycle-checker limitations, modifying another dependency etc.) or choice of language. I appreciate the feedback and encouragement from everyone here. If this is not the best way to proceed, I am OK with closing this PR without merging, but I'm not going to pursue these alternative approaches.

dae:
I've posted an experiment using cargo on #517

UebelAndre:
I actually feel this might be worth keeping around, I'm investigating some ways rust binaries can be bootstrapped as part of https://github.com/bazelbuild/rules_rust/issues/650 and would like to get something basic setup quickly. Maybe the same infrastructure could be applied here?

But I'm also not against a C++ implementation, I'm just more excited about a Rust implementation 😄

PR #424, opened 2020-10-11
rust_repositories() was not honoring edition default on Windows

dae:
The Windows toolchain was ignoring the default provided to rust_repositories(), always using edition 2015, which caused errors about crates being unavailable if extern crate references were not in the code.


damienmg (approved):
Thanks a lot, sorry for the delay I am a bit overloaded at this time.

PR #425, opened 2020-10-11
use -msvc variant on Windows

dae:
use -msvc variant on Windows

Both the standard Rust installer, and rust_repositories() in this repo
default to the -msvc platform. Since it is the default on Windows,
this change comments out the -gnu variant instead.

This change (along with some fixes in cargo raze) allows crates with
msvc-specific build deps to compile. I do not know enough about Bazel
to say if it will have any unintended side effects, but it seems to
work for me, and allows a moderate sized project with about 250
crate dependencies to compile using the default toolchain.

### DIFF
--- a/rust/platform/platform.bzl
+++ b/rust/platform/platform.bzl
@@ -11,16 +11,16 @@ load(
 # All T1 Platforms should be supported, but aren't, see inline notes.
 _SUPPORTED_T1_PLATFORM_TRIPLES = [
     "i686-apple-darwin",
-    "i686-pc-windows-gnu",
+    "i686-pc-windows-msvc",
     "i686-unknown-linux-gnu",
     "x86_64-apple-darwin",
-    "x86_64-pc-windows-gnu",
+    "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-gnu",
     # N.B. These "alternative" envs are not supported, as bazel cannot distinguish between them
     # and others using existing @platforms// config_values
     #
-    #"i686-pc-windows-msvc",
-    #"x86_64-pc-windows-msvc",
+    #"i686-pc-windows-gnu",
+    #"x86_64-pc-windows-gnu",
 ]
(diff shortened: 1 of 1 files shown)


dae:
I experimented with enabling both -gnu and -msvc variants, and using a constraint setting to switch between them. This avoids the "Illegal ambiguous match on configurable attribute deps" errors that arise when both the -gnu variant and -msvc variant are listed in a crate's dependencies. I wasn't able to figure out how this could be made a default though, and if that's not possible, it means Windows users would need to define the environment they wish to use in their BUILD file, eg

```
platform(
    name = "x86_64-windows-gnu",
    constraint_values = [
        "@platforms//cpu:x86_64",
        "@platforms//os:windows",
        "@io_bazel_rules_rust//rust/platform/env:gnu",
    ],
)
```

and then specify the platform when compiling:

bazel build //some:target --platforms=x86_64-windows-gnu --host_platform=x86-windows-gnu

This is a bit of a pain, but then it's already necessary to pass -enable_runfiles in on Windows in order to get crate buildscripts to compile. Thoughts?

damienmg:
I think commenting out the gnu version is preferable.

dae:
Sorry, I had thought I'd commented out the other set already - corrected.

damienmg:
Thanks!

UebelAndre:
@damienmg @dae Hey, this seems to have broken CI. Would one of you guys mind taking a look?

UebelAndre:
It looks like `cargo-raze` will need to be updated and these outputs will need to be regenerated 😞 or a list of `targets` will need to be added to exclude the windows `-gnu` platform declarations.

UebelAndre:
I have a fix in #450

UebelAndre:
Just an FYI: https://github.com/bazelbuild/platforms/issues/38 may potentially lead to support for both `-msvc` and `-gnu`

Issue #423, opened 2020-10-09
cargo_build_script fails on Windows with bazel defaults


dae:
Attempting to build a crate with cargo-raze's gen_buildrs = true fails on Windows:

thread 'main' panicked at 'Unable to start binary: Os { code: 267, kind: Other,
message: "The directory name is invalid." }', external/io_bazel_rules_rust/cargo
/cargo_build_script_runner/lib.rs:104:60

It appears to be happening on this line, as the path it points to does not exist:

https://github.com/bazelbuild/rules_rust/blob/169476f772132723b2cc819c96209c6fbce3e366/cargo/cargo_build_script_runner/bin.rs#L48

I am guessing this is because I have not enabled developer mode/runfiles: https://docs.bazel.build/versions/master/windows.html#enable-symlink-support

While I could do so on my local development machine, I fear this would not be possible when running CI with GitHub actions/Travis/etc. Have I missed something?

dae:
As a follow-up to this, I ran into further issues when trying with --enable_runfiles. When attempting to compile a crate like libsqlite3-sys, the C compiler failed to find includes like stdarg.h. Bazel seems to be able to compile C++ files directly, so my guess is this is an issue with cargo_build_script not passing the relevant envvars into the build script.

dae:
The following seems to help things move along:

```diff
diff --git a/cargo/cargo_build_script.bzl b/cargo/cargo_build_script.bzl
index 3e2a8cc..b18ede4 100644
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
+load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_compile_env", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
@@ -52,6 +52,8 @@ def _cargo_build_script_run(ctx, script):
[...]

PR #426, opened 2020-10-11
pass compile environment into build script

dae:
On Windows, MSVC requires INCLUDE to be set to the path of the standard
include files. Not sure if this is the best solution, but it solved issues with
missing includes mentioned in #423 for me.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -1,5 +1,5 @@
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
-load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
+load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_compile_env", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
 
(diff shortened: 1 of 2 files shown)


cargo/cargo_build_script.bzl
```diff
@@ -52,6 +52,8 @@ def _cargo_build_script_run(ctx, script):
     # Pull in env vars which may be required for the cc_toolchain to work (e.g. on OSX, the SDK version).
     # We hope that the linker env is sufficient for the whole cc_toolchain.
     cc_toolchain, feature_configuration = get_cc_toolchain(ctx)
+    cc_env = get_cc_compile_env(cc_toolchain, feature_configuration)
+    env.update(**cc_env)
```
damienmg, line 56: I am not sure either this is the best solution but I think we are better with it than without.

```suggestion
    # TODO: Trim down the list of environment shipped.
    env.update(**cc_env)
```

damienmg (approved):
I just have a recommendation to add a TODO.

Thanks a lot for all your changes, hopefully I can get to them all now.

cargo/cargo_build_script.bzl
dae: If you'd prefer to keep env changes to a minimum, it seems as if INCLUDE is sufficient for the CC-using crates in my project to compile:

```diff
diff --git a/cargo/cargo_build_script.bzl b/cargo/cargo_build_script.bzl
index 0725c4c..a13826b 100644
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -53,7 +53,9 @@ def _cargo_build_script_run(ctx, script):
     # We hope that the linker env is sufficient for the whole cc_toolchain.
     cc_toolchain, feature_configuration = get_cc_toolchain(ctx)
     cc_env = get_cc_compile_env(cc_toolchain, feature_configuration)
-    env.update(**cc_env)
+    includes = cc_env.get("INCLUDE")
+    if includes:
+        env["INCLUDE"] = includes
     _, _, linker_env = get_linker_and_args(ctx, cc_toolchain, feature_configuration, None)
     env.update(**linker_env)
```

Would you like me to push this change?

dae: I've tentatively made this change and rebased on master.

damienmg (approved):

damienmg:
Thanks that's perfect like this.

Issue #391, opened 2020-08-28
_cargo_build_script_run not executing relative to crate?


parisholley:
i am having issues with crates such as ring/openssl that attempt to compile libraries as part of the build... they try to refer to files that are relative to the crate (when vendor'd) but instead, the working directory is my repository root. if i manually edit the build files to do ./cargo/vendor/crate-name-version/path/to/file.c then it works fine (but obviously isn't a solution).

damienmg:
Hi, sorry for the delay, I am a bit overloaded at the moment,

Would you mind sharing the Cargo.toml file you used so I can repro? Can you also share some details about your operating system?

parisholley:
https://github.com/parisholley/cargobug

damienmg:
Thanks for the repro. Sorry I still don't follow fully, what are you changing in the BUILD file?

parisholley:
if you checkout the repo, and run the instructions in the readme, you can reproduce the error. i'm on Mac OS 10.15.5

damienmg:
Yes I can repro, this was not what I wanted to ask. You said earlier that:

    if i manually edit the build files to do ./cargo/vendor/crate-name-version/path/to/file.c then it works fine (but obviously isn't a solution).

Are you editing the build.rs?

parisholley:
yep

UebelAndre:
Does this happen when using `remote` mode with `cargo raze`?

dae:
@parisholley did #427 solve your problem?

UebelAndre:
Closed by https://github.com/bazelbuild/rules_rust/pull/427

PR #427, opened 2020-10-11
run build scripts in the crate folder instead of the workspace

dae:
Various crates expect to be able to access files relative to the
crate root, and we were giving them the parent directory (eg __main__)
instead. An example crate that fails to compile without this change
is blake3.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -20,7 +20,7 @@ def _build_script_impl(ctx):
     dep_env_out = ctx.actions.declare_file(ctx.label.name + ".depenv")
     flags_out = ctx.actions.declare_file(ctx.label.name + ".flags")
     link_flags = ctx.actions.declare_file(ctx.label.name + ".linkflags")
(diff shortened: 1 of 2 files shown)


dae:
I'm guessing the example script was written with the assumption that it would be running from the parent folder. Presumably there is other code out there that makes such an assumption as well, so this would be a breaking change. There are lots of crates out there, and it would be unfortunate if any crates expecting crate root need to be vendored and modified to work in bazel. If changing the default is out of the question, perhaps an extra argument could be added to cargo_build_script() instead? Cargo raze could then set it automatically.

dae:
#391 may be related

damienmg:
After some though, I am fine with doing this directly. Can you fix the related tests?

dae:
Sure, test fixed.

damienmg:
Thanks!

UebelAndre:
@damienmg @dae Hey, this seems to have broken CI. Would one of you guys mind taking a look?

PR #433, opened 2020-10-14
Regenerate wasm bindgen raze outputs

UebelAndre:
This regenerates the wasm-bindgen raze outputs using `cargo-raze` version `0.6.0`. 

Notably, this change changes the `gen_workspace_prefix` to `rules_rust_wasm_bindgen` so users of these rules also using `raze` don't run into conflicts in the targets they expected to be using vs the targets `rules_rust` is providing.

This also cleans up some of the implementation a bit since `cargo-raze` has gained a bunch of new functionality since it was last run here. 

Additionally `wasm_bindgen_backtrace_sys_0_1_29` was removed since it appeared to be unused.


UebelAndre:
It's worth noting that this PR is intentionally not updating `wasm-bindgen` as a way to first prove `raze` is still able to generate compatible outputs. I'd like to open another PR to update this though.

UebelAndre:
@damienmg @mfarrugi this PR is ready for review 😃

wasm_bindgen/raze/BUILD
```diff
@@ -1,4 +1,5 @@
 """
```
mfarrugi, line 1: delete

wasm_bindgen/raze/remote/backtrace-sys-0.1.29.BUILD
```diff
@@ -1,9 +1,21 @@
 """
```
mfarrugi, line 1: Does raze still generate this when you override the buildfile? Delete if not

wasm_bindgen/raze/BUILD
UebelAndre: I actually meant to keep the `BUILD` file to reduce the amount of deltas in this PR. I've removed `BUILD.bazel` for now. This can be updated by removing `output_buildfile_suffix = "BUILD"` from the `Cargo.toml` file and rerunning `cargo-raze`

wasm_bindgen/raze/remote/backtrace-sys-0.1.29.BUILD
UebelAndre: This is misleading because it's not an "override". It's a file that gets appended to the matching cargo-raze output file

This pointed out though that I forgot to pull a change from another branch I have. Please see  #438

PR #434, opened 2020-10-14
fix deadlock while waiting for build script output

dae:
If the build script outputs enough, it fills up the stdout pipe,
blocks, and wait() never returns. Windows has a smaller pipe buffer,
so when compiling the blake3 crate for example, the compile gets stuck
after a few files have been compiled with the cc crate.


damienmg (approved):
Thanks!

PR #437, opened 2020-10-15
Include wasm-lld in toolchain

lizan:
Before this change, `rust-lld` is not found in RBE environment because they are not part of toolchain definition. This PR adds `rust-lld` and its runtime dependencies into toolchain definition, so they can work in RBE / docker-sandbox.


mfarrugi:
Thanks!

PR #438, opened 2020-10-15
Updated wasm-bindgen backtrace dependency

UebelAndre:
This is something that should have been in #433 

See https://github.com/bazelbuild/rules_rust/pull/433#discussion_r505670820


UebelAndre:
@mfarrugi Thanks for all the reviews! This is one thing I should have added in #433

PR #439, opened 2020-10-15
Add --extern proc_macro for 2018 crates

illicitonion:
This is done implicitly by cargo.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -518,6 +518,10 @@ def construct_arguments(
     add_native_link_flags(args, dep_info)
     add_crate_link_flags(args, dep_info)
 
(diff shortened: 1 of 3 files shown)


rust/private/rustc.bzl
```diff
@@ -518,6 +518,10 @@ def construct_arguments(
     add_native_link_flags(args, dep_info)
     add_crate_link_flags(args, dep_info)
 
+    if crate_info.type == "proc-macro" and not (crate_info.edition == None or crate_info.edition == "2015"):
```
mfarrugi, line 521: edition should never be None, if it is, it's a bug elsewhere

illicitonion: Fabulous, changed to just a `!= "2015"` :) Thanks!

PR #440, opened 2020-10-15
Enable buildifier for CI

UebelAndre:


UebelAndre:
@damienmg @mfarrugi Any assistance in solving for these warnings would be greatly appreciated! 🙏 

I think it'd be really good for the repo to have `buildifier` enabled 😅

mfarrugi:
I don't think either of us are up for making changes. It would be a
step-forward to disable the complaints, then more complaints won't slip
through.

UebelAndre:
I opened #443 and #442 to track things I feel really should be solved

UebelAndre:
I also opened #441 to solve for additional defects not caught in CI (I believe this is due to how the example's workspaces are setup)

PR #447, opened 2020-10-16
support --action_env vars in build scripts

dae:
The existing rustc_env attribute is useful when the provided env var
will not differ between machines, but is not practical when each
user of the build script may require a different value, such as passing
in a custom path to a tool on their machine.

Fixes the issue mentioned in
https://github.com/bazelbuild/rules_rust/pull/314#issuecomment-624524139
by starting from the default environment. I did not make a change in rustc.bzl,
so this only applies to build scripts at the moment.


damienmg (approved):
Thanks! I would have preferred to ship it with a test but this would require a bit too much setup to everybody so LGTM.

PR #448, opened 2020-10-16
add buildscript link flags only to parent crate

dae:
@illicitonion Would you mind looking over this and seeing if it's sane? It's modifying code that was added in a previous pull request of yours: https://github.com/bazelbuild/rules_rust/pull/346

I use a crate for creating Python extension modules called pyo3. It has code in it like this:

```
#[cfg_attr(windows, link(name = "pythonXY"))]
extern "C" {
    pub fn PyNode_Compile(arg1: *mut _node, arg2: *const c_char) -> *mut PyCodeObject;
```

Its build.rs script locates the relevant library name, and spits out a link flag that maps it to the alias:

```
-lpythonXY:python38
```

The pyo3 crate compiles successfully, but when I attempt to compile a crate that depends on pyo3 on Windows, the build fails, complaining that the above alias has been provided but there are no mentions of pythonXY in my crate - because the aliases are only in the pyo3 crate.

When I compile my crate using cargo instead of bazel, I can see that pyo3's build script link flags are not being passed in when compiling my crate - so the current behaviour of rules_rust does not seem to match what cargo is doing.

The change in this PR fixes the issue for me, and hasn't caused any problems here. But your PR specifically mentions transitive deps, so I wonder if I am missing something. Does this change cause a regression in the problem you were originally trying to fix?

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -724,15 +724,10 @@ def _create_extra_input_args(ctx, file, build_info, dep_info):
     if build_info:
         out_dir = build_info.out_dir.path
         build_env_file = build_info.rustc_env.path
-
-        # out_dir will be added as input by the transitive_build_infos loop below.
         build_flags_files.append(build_info.flags.path)
-
-    # This should probably only actually be exposed to actions which link.
-    for dep_build_info in dep_info.transitive_build_infos.to_list():
-        input_files.append(dep_build_info.out_dir)
-        build_flags_files.append(dep_build_info.link_flags.path)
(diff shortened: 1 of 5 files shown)


damienmg:
Can we add an example that would fails without that change?

dae:
Have force-pushed a test without the original fix, and expect it to fail. Will follow up when test run is done.

illicitonion:
Works for me - I added examples in #346 and if those are green, I'm happy :) Thanks!

dae:
Great, this should be good to go now.

PR #449, opened 2020-10-16
Switching "host" configurations to "exec"

UebelAndre:
https://docs.bazel.build/versions/master/skylark/rules.html#configurations

> Note: Historically, Bazel didn’t have the concept of execution platforms, and instead all build actions were considered to run on the host machine. Because of this, there is a single “host” configuration, and a “host” transition that can be used to build a dependency in the host configuration. Many rules still use the “host” transition for their tools, but this is currently deprecated and being migrated to use “exec” transitions where possible.
> 
> There are numerous differences between the “host” and “exec” configurations:
> 
> - “host” is terminal, “exec” isn’t: Once a dependency is in the “host” configuration, no more transitions are allowed. You can keep making further configuration transitions once you’re in an “exec” configuration.
> - “host” is monolithic, “exec” isn’t: There is only one “host” configuration, but there can be a different “exec” configuration for each execution platform.
> - “host” assumes you run tools on the same machine as Bazel, or on a significantly similar machine. This is no longer true: you can run build actions on your local machine, or on a remote executor, and there’s no guarantee that the remote executor is the same CPU and OS as your local machine.

### DIFF
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -197,27 +197,27 @@ rust_bindgen_toolchain = rule(
         "bindgen": attr.label(
             doc = "The label of a `bindgen` executable.",
             executable = True,
-            cfg = "host",
+            cfg = "exec",
         ),
         "rustfmt": attr.label(
             doc = "The label of a `rustfmt` executable. If this is provided, generated sources will be formatted.",
             executable = True,
-            cfg = "host",
+            cfg = "exec",
             mandatory = False,
         ),
         "clang": attr.label(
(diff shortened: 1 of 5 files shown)


mfarrugi (approved):

mfarrugi:
> Tools that are executed as part of the build (e.g., compilers, code generators) should be built for an exec configuration. In this case, specify cfg="exec" in the attribute.
> 
> Otherwise, executables that are used at runtime (e.g. as part of a test) should be built for the target configuration. In this case, specify cfg="target" in the attribute.

Is the more salient bit of the docs.

Issue #428, opened 2020-10-12
How to do metadata-only typecheck build of a rust_library?


dtolnay:
It looks like rules_rust unconditionally passes `--emit=dep-info,link`. This means Bazel builds cannot benefit from the biggest rustc speedup since 1.0.0 (https://blog.rust-lang.org/2017/03/16/Rust-1.16.html#whats-in-1160-stable). :crying_cat_face: 

https://github.com/bazelbuild/rules_rust/blob/e64700dc9b8b3869bce4f77b78c33cb9d088cc4b/rust/private/rustc.bzl#L374

Is there any idiomatic way to perform non-code-generating builds which type-check only? Equivalent to Cargo's `cargo check` or `cargo rustc --profile=check`, rustc's `rustc --emit=metadata`, or Buck's #check flavored builds (`buck build :lib#check`; https://buck.build/concept/flavors.html#analysis_http).

mfarrugi:
We could change the code to react to a flag, eg. ` --compilation_mode` (
https://github.com/bazelbuild/rules_rust/blob/e64700dc9b8b3869bce4f77b78c33cb9d088cc4b/rust/private/rustc.bzl#L370-L374)
or turn `rust_library` into a macro that creates a 'check' target in
addition to the build target. I am not sure whether either approach is
'idiomatic'.

Going a little farther with the latter, rust_library could be changed to
only run check until an output artifact is actually needed by a test or
binary, but this probably requires redundant calls to rustc and more bazel
know-how than I've got :).

dtolnay:
> or turn `rust_library` into a macro that creates a 'check' target in addition to the build target.

This sounds bad if it means `...` builds will build those both. Would there be a way to not do that? I'd expect a way run a whole `...` build in the "check" compilation mode or the non-check mode.

mfarrugi:
I don't think a `build ...` would work because any binaries wouldn't have
the necessary artifacts produced, same for `test ...`

mfarrugi:
The check targets could be tagged manual to be excluded under ..., but that
doesn't let you `check ...`

dtolnay:
For comparison, the behavior of Buck is that `buck build ...` will do codegen of any rust_library that is pulled in transitively by any binary include in the build, and do check-only build of any rust_library that is included in the `...` but *not* pulled in by any binary. That seems like the best behavior. How sure are we that there would be no way to reproduce that in Bazel, or that we wouldn't want to?

[...]

PR #455, opened 2020-10-20
skip link step when running clippy

dae:
I have a crate that uses pyo3 to create a Python extension module.
On macOS, special link flags are required:

    rustc_flags = selects.with_or({
        (
            "@io_bazel_rules_rust//rust/platform:x86_64-apple-darwin",
        ): [
            "-Clink-arg=-undefined",
            "-Clink-arg=dynamic_lookup",
        ],
        "//conditions:default": [],
    }),

Without them, the linker on macOS fails, as the Python API is not
available until runtime.

rules_rust's clippy implementation was passing --emit=dep-info,link
to the clippy invocation, causing a clippy run to fail with linking
errors. This patch changes the invocation to use
--emit=dep-info,metadata instead, which is what cargo uses. It also
shaves a bit of time off the check, as linking no longer needs to happen.

Tangentially related to #428 and #421 - currently the clippy aspect
seems to be falling back on a non-worker compile, so it's still
noticeably slower than running cargo clippy directly when minor
changes have been made.

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -93,6 +93,7 @@ def _clippy_aspect_impl(target, ctx):
         build_flags_files = build_flags_files,
         maker_path = clippy_marker.path,
         aspect = True,
+        emit = "dep-info,metadata",
     )
 
     # Deny the default-on clippy warning levels.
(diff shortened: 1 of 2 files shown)


UebelAndre:
Any updates here? Looks like it didn't even get a review yet

dae:
Patiently awaiting a merge :-)

dfreese (approved):

UebelAndre:
@dfreese Btw there's a similar PR to this

dfreese:
Closed in favor of #526.  Sorry for the delay on the review @dae.

Issue #268, opened 2019-11-11
Unable to load file '@bazel_version//:def.bzl': file doesn't exist


jonhoo:
With this `WORKSPACE`:
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "io_bazel_rules_rust",
    sha256 = "9f38c5fbf37b0dc445fe54558e14ac000cca0be6907251c6df21ba236acd59a8",
    strip_prefix = "rules_rust-ec436b5ff2ab1ddeba6f27a7a1a5d263812981a6",
    urls = [
        # Master branch as of 2019-11-11
        "https://github.com/bazelbuild/rules_rust/archive/ec436b5ff2ab1ddeba6f27a7a1a5d263812981a6.zip",
    ],
)

http_archive(
    name = "bazel_skylib",
    sha256 = "9a737999532daca978a158f94e77e9af6a6a169709c0cee274f0a4c3359519bd",
    strip_prefix = "bazel-skylib-1.0.0",
    url = "https://github.com/bazelbuild/bazel-skylib/archive/1.0.0.tar.gz",
)

load("@io_bazel_rules_rust//rust:repositories.bzl", "rust_repositories")
rust_repositories()

load("@io_bazel_rules_rust//:workspace.bzl", "bazel_version")
bazel_version(name = "bazel_version")
```
And bazel version 1.1.0, `bazel fetch` fails with
```
ERROR: error loading package 'rust': in /home/jon/.cache/bazel/_bazel_jon/66db7c3492b5fc3fa5db2c66360c215b/external/io_bazel_rules_rust/bindgen/bindgen.bzl: in /home/jon/.cache/bazel/_bazel_jon/66db7c3492b5fc3fa5db2c66360c215b/external/io_bazel_rules_rust/rust/rust.bzl: in /home/jon/.cache/bazel/_bazel_jon/66db7c3492b5fc3fa5db2c66360c215b/external/io_bazel_rules_rust/rust/private/rustdoc_test.bzl: in /home/jon/.cache/bazel/_bazel_jon/66db7c3492b5fc3fa5db2c66360c215b/external/io_bazel_rules_rust/rust/private/rustc.bzl: Unable to load file '@bazel_version//:def.bzl': file doesn't exist
```
The crux of this appears to be
```
Unable to load file '@bazel_version//:def.bzl': file doesn't exist
```

jonhoo:
Which I believe stems from:
[...]

PR #460, opened 2020-10-21
Rename `bazel_version` repository to avoid conflict

wchargin:
We use a synthetic `bazel_version` repository to make the Bazel version
available in more contexts (see [bazelbuild/bazel#8305][i8305]). But
some other repositories do so, too, with the same repository name and a
different repository structure. In particular, it is not currently
possible to use `rules_rust` and certain versions of `upb` (downstream
of protobuf) in the same repository, due to their definition here:

https://github.com/protocolbuffers/upb/blob/c1357afb2e39df671d89eaec49033b5329f36a3e/bazel/repository_defs.bzl#L7-L10

An easy workaround is to disambiguate the name. It looks much easier to
change `rules_rust` than to change `upb` and update its long chain of
workspace dependencies, hence this patch. :-)

See my comment on #268 for more details and a full repro:
https://github.com/bazelbuild/rules_rust/issues/268#issuecomment-713920963

[i8305]: https://github.com/bazelbuild/bazel/issues/8305

Test Plan:
Tested by adding the README’s workspace stanza to TensorBoard (at
current master, 8d629954c251). It fails at `rules_rust = 5998baf9016e`,
but succeeds with a `local_repository` that has this patch. Also, in
this repo, `git grep @bazel_version` no longer has any matches.

wchargin-branch: disambiguate-bazel-version-repo

### DIFF
--- a/workspace.bzl
+++ b/workspace.bzl
@@ -41,4 +41,4 @@ def rust_workspace():
 
     bazel_skylib_workspace()
 
-    bazel_version(name = "bazel_version")
+    bazel_version(name = "io_bazel_rules_rust_bazel_version")
--- a/workspace.bzl
+++ b/workspace.bzl
@@ -41,4 +41,8 @@ def rust_workspace():
 
     bazel_skylib_workspace()
 
+    # Give this repository a scoped name to avoid conflicting with other
+    # projects' similar workarounds when used in the same workspace
+    # (issue #268#issuecomment-713920963). TODO(#462): Investigate
+    # whether this can be entirely replaced with `native.bazel_version`.
     bazel_version(name = "io_bazel_rules_rust_bazel_version")
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -7,7 +7,7 @@ bzl_library(
     name = "all_deps",
     srcs = [
         "@bazel_tools//tools:bzl_srcs",
-        "@bazel_version//:def.bzl",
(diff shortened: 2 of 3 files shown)


UebelAndre:
Would it be possible with `native.bazel_version` to remove that repository rule entirely? @mfarrugi would you happen to know as well?

wchargin:
I’m not sure; `rustc.bzl` uses `BAZEL_VERSION` inside macros, like
`get_cc_toolchain`, but bazelbuild/bazel#8305 (open issue) says that
`native.bazel_version` is not available for macros or rules, which is
why this workaround exists in the first place.

workspace.bzl
```diff
@@ -41,4 +41,4 @@ def rust_workspace():
 
     bazel_skylib_workspace()
 
-    bazel_version(name = "bazel_version")
+    bazel_version(name = "io_bazel_rules_rust_bazel_version")
```
dfreese, line 44: can you add a TODO referencing #462 to investigate if this is still necessary, and the original issue where you were seeing naming conflicts (#268)?

dfreese (approved):
Seems reasonable, but I wish bazel had a better solution for this.

workspace.bzl
wchargin: Done.

dfreese: thanks!

dfreese (approved):

PR #461, opened 2020-10-22
Support expanding locations in rustc_env and build_script_env

### LINKED ISSUE #79: Including a genfile at compile time is not possible

### LINKED ISSUE #454: Bazel build with prost-build build.rs fails with "No such file or directory"
I created a demo repository:

https://github.com/ttiurani/bazel-prost-build-bug

to showcase an issue where the [standard `prost-build` example](https://docs.rs/prost-build/0.6.1/prost_build/) works fine with a `cargo build` but running the same with `cargo raze`'d Bazel repository with `bazel build demo` fails with:

```
ERROR: /private/var/tmp/_bazel_ttiurani/0c87cdffcf4741baaa498655bde07714/external/raze__demo_prost_build__0_0_1/BUILD.bazel:36:19: CargoBuildScriptRun external/raze__demo_prost_build__0_0_1/demo_prost_build_build_script.out_dir failed (Exit 1) cargo_build_script_runner failed: error executing command bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/io_bazel_rules_rust/cargo/cargo_build_script_runner/cargo_build_script_runner ... (remaining 8 argument(s) skipped)

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', external/raze__demo_prost_build__0_0_1/cargo/build.rs:2:66
stack backtrace:
   0:        0x10cac9a6e - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h24bb64d98a7e25d6
   1:        0x10caeaf9c - core::fmt::write::h8fdc9cddb01cd8b2
   2:        0x10cac6bd9 - std::io::Write::write_fmt::hcc3030013983bab6
   3:        0x10cacbc35 - std::panicking::default_hook::{{closure}}::h95817712c5ff0736
   4:        0x10cacb972 - std::panicking::default_hook::h34e085f4e0b1062d
   5:        0x10cacc195 - std::panicking::rust_panic_with_hook::haf571858f996ac45
   6:        0x10cacbd62 - rust_begin_unwind
   7:        0x10caf37ef - core::panicking::panic_fmt::h11676ba6a846d9f4
   8:        0x10caf36f5 - core::option::expect_none_failed::h312e48bacb63d8ec
   9:        0x10ca81c1b - demo_prost_build_build_script_script_::main::h399dfdb84fd54273
  10:        0x10ca82986 - std::rt::lang_start::{{closure}}::h0a0a8f1aceef40a7
  11:        0x10cacc45f - std::rt::lang_start_internal::hcd84a36052901671
  12:        0x10ca81c69 - main
Error: "Build script process failed with exit code 101"
```

Where the build.rs is:

```
fn main() {
    prost_build::compile_protos(&["src/items.proto"], &["src/"]).unwrap();
}
```

The error message is quite non-descriptive so was unable to trace this further myself. Note that I'd very much like to use the bundled `protoc` binary and not an external one to make my build easily repeatable in CI environments.

Any ideas what might be wrong? I'm using the latest rules_rust from this morning.

### LINKED ISSUE #459: Can't use `include_str!()` macro with Bazel-generated data files
## The Problem

The `data` attribute appears on each of the core Rust rules (e.g. `rust_binary`). The documentation for `data` says

> This attribute can be used to specify any data files that are embedded into the library, such as via the `include_str!` macro.

I have found that `include_str()!` works only for "source" files (i.e. files included in the Bazel source tree). Unfortunately, the `include_str!()` macro does **not** work if the data files are generated by some other Bazel target.

## Example

To demonstrate this, I've [added a small test to my `rules_rust` fork](https://github.com/dwtj/rules_rust/commit/b67ce49f0bc81735f4e083da560eb02fd6911fbf).

In this test, one data file is a source file and one data file is generated by a `genrule`. The contents of both files are embedded in the `rust_test` via the `include_str!()` macro.

On my linux system, this `rust_test` fails to compile with the following `rustc` error:

```
error: couldn't read test/generated_data/include_str/generated_data.txt: No such file or directory (os error 2)
 --> test/generated_data/include_str/include_str.rs:9:13
  |
9 |             include_str!("generated_data.txt").trim()
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

```

Notice that `rustc` has no problem finding `source_data.txt`.

I believe that the problem can be seen by inspecting this `rustc` action's sandbox. On my Linux system, if I run the test with `bazel test --sandbox_debug`, and then change directory to `<bazel_workspace_output_base>/sandbox/linux-sandbox/<action_id>/execroot/io_bazel_rules_rust`, the root of the problem becomes clear: the library and the source data are sibling files

```
./test/generated_data/include_str/include_str.rs
./test/generated_data/include_str/source_data.txt
```

but the generated file is not

```
./bazel-out/k8-fastbuild/bin/test/generated_data/include_str/generated_data.txt
```

So, the generated data file is an input to the `rustc` Bazel action, but it isn't anywhere that the `rustc` executable can find it. Thus, a compiler error occurs.

## Temporary Workaround

In my project where I actually want this feature I have an awful-dreadful-ugly hack to overcome this problem: I add a symlink from my source tree pointing to where I expect the generated file should be under the Bazel workspace's output base; then I add both the generated file and its symlink to the `data` attribute.

## A Very Rough Idea for a Solution

I have a little bit of experience writing Bazel rules that also needed to deal with this problem. My solution [there](https://github.com/dwtj/dwtj_rules_hugo/blob/8617d3455cf8cc9787aced9ea2a86045efefeb7d/hugo/private/hugo_website/defs.bzl#L57) was for each target to add various `ctx.actions.symlink()` actions to place both generated files and source files into a new directory under their logical paths before calling the tool.

dae:
This makes it possible to pass in the path to generated files and
external tools.

This potentially closes #459, closes #454, and closes #79.

The docs seem to indicate there's precedent for this in rules_cc:
https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables


mfarrugi (approved):

dae:
I've pushed a fix for the failing clippy run; if there was a better way to address it please let me know.

dae:
Ok, this seems to be related to external repos. Changing to the examples folder and running 'bazel test env_locations/...' works, but it fails when running 'bazel test @examples//env_locations/...' from the top level folder. The issue seems to be limited to build scripts - it works correctly in both cases for normal rust_binaries. I'm not yet familiar enough with Bazel to know how to solve this properly - any suggestions would be appreciated.

dae:
Ok, I've got something that is passing CI. Copying in the updated commit notes:

- rustc_env now expands $(rootpath //package:target), for passing
the location of files or tools that are needed at compile time
- build_script_env now expands $(execpath //package:target), for
passing the location of files or tools that are needed at build
script runtime
- cargo_script_build() now passes the data argument to the build
script runner, so the data is available at runtime

Ideally both build and run would work with execpath/location, but
the approach of walking back to the root folder that cargo_build_script()
is using doesn't seem to be possible or necessary at compile time,
so at compile time, we use the relative rootpath instead.

And from the fix:

Rust on Windows seems to struggle with relative pathnames in this case
- I suspect due to the fact that symlinks are being used.

fs::read_to_string("../../../path/to/somewhere")

gives an error

as does calling .canonicalize() on the above. Switching to \ as a
path separator does not seem to help.

If the parent directory references are handled separately, eg

Path::new("../../..").canonicalize(), then the path/to/somewhere
can be joined onto the folder and the file can be read. So instead
of automatically adjusting the path, we just return execroot verbatim,
and rely on the build script to do their own path manipulation.

cargo/cargo_build_script.bzl
```diff
@@ -220,10 +231,19 @@ def cargo_build_script(
...
-        # Data are shipped during execution.
-        data = ["src/lib.rs"],
-        # Environment variables passed during build.rs execution
-        build_script_env = {"CARGO_PKG_VERSION": "0.1.2"},
+        # Optional environment variables passed during build.rs compilation
+        rustc_env = {
+           "CARGO_PKG_VERSION": "0.1.2",
+        },
+        # Optional environment variables passed during build.rs execution.
+        # The path will be relative to the folder above bazel-out, so your
```
illicitonion, line 239: I think this general change of allowing `location` expansion is a really useful one, but I don't love the interactions around path lookups.

In particular, this suggestion to traverse parent directories until you happen to find the right one. The `bazel-out` heuristic is an internal implementation detail of how bazel lays out files, and actually only happens to be true for generated files - if you did `$(execroot //some/checked/in:file)` there wouldn't be a `bazel-out`. `$(location)` and friends were designed assuming actions run in the execroot (or a runfiles tree which mirrors the execroot).

I think a reasonable approach here would be for us to change build scripts so that they executed in the execroot as their working directory (which means that these injected env vars are valid relative paths), and to recommend that crates explicitly add a `env!("CARGO_MANIFEST_DIR")` prefix to any paths they try to read at build time. We changed where build scripts run recently in https://github.com/bazelbuild/rules_rust/pull/427 so I don't feel too bad about changing it again... This approach seems fairly principled, and gives a very clear workaround for crates which face problems with this.

Any required fix pull requests may appear as slightly odd to crate authors (the author may think "the working directory _is_ `env!("CARGO_MANIFEST_DIR")` - why would I need to be explicit?"), but hopefully they'd just find it odd (and understand that alternative build tools may behave differently), rather than objectionable...

dae: I believe it should also work on source files, as we're only referencing "bazel-out" in the build script, which is always generated. But I take your point about this being a bit of a hack, and an approach that doesn't require hunting for execroot would be both more ergonomic and less fragile.

Perhaps one way to solve this would be to add another argument to cargo_build_script() that controls whether the working dir is execroot or the manifest dir? Crates that have bought into the Bazel ecosystem could then consume the locations directly, and third-party crates that assume cargo will set the working dir to the manifest dir for them would continue to function without patching.

illicitonion: That feels like a reasonable compromise to me - we could perhaps also set an additional `BAZEL_EXEC_ROOT` env var in the manifest dir path case, to help folks who are trying to support both well to support conditional behaviour.

One challenge I've run into is crates which use [`prost-build`](https://docs.rs/prost-build), so it may be interesting to run through this as a concrete example:

With https://github.com/danburkert/prost/pull/324 (current unreleased, sadly) and this PR, you can set `build_script_env` to something like `{"PROTOC": "$(execroot @com_google_protobuf//:protoc)"}`:

In exec-root-as-working-directory mode, things should Just Work assuming all other paths (e.g. paths to the protobufs being codegen'd) are using the `$CARGO_MANIFEST_DIR` prefix properly.
In cargo-manifest-dir-working-directory mode, it would require the build script to prepend `${BAZEL_EXEC_ROOT}` to `PROTOC`, which works but is a bit annoying... Alternatively, we could extend `${pwd}`-substitution to the build script runner (which seems pretty reasonable), so that one could simply use:
`{"PROTOC": "${pwd}/$(execroot @com_google_protobuf//:protoc)"}` (or possibly `{"PROTOC": "$${pwd}/$(execroot @com_google_protobuf//:protoc)"}` - I always get lost in the escaping here!)
and the build script runner would do the right thing in both modes...

Actually, the more I talk it through, the more I'm convinced we don't need both modes, we just need `${pwd}`-substitution everywhere that we support `$(location)` substitution - how does that sound to you?

dae: Yeah, solving the prost-build problem was one of the motivations for this PR, as I'm using it in my project too.

Substituting pwd sounds good, I'll look into it.

dae: What do you think of the latest push? It adds the execroot automatically rather than relying on the user to do so - the rationale being that it's consistent with the way things like genrule() behave. But I can change it to require an explicit ${pwd} reference if you'd prefer.

And if you'd like me to squash these commits to tidy things up, please let me know.

dae: (I experimented with resolving execpath/${pwd} for rust_binary/rust_test(), but it didn't work - rustc_env is only available at compile time, and if we bake in the execroot, the execution stage is running in a different sandbox, and the path is no longer valid - so they will need to continue using rootpath instead)

cargo/cargo_build_script.bzl
```diff
@@ -261,5 +287,5 @@ def cargo_build_script(
         crate_features = crate_features,
         version = version,
         build_script_env = build_script_env,
-        deps = deps,
+        data = data,
```
illicitonion, line 290: You need to re-add `deps = deps` - deps still matter! :)

illicitonion (approved):
LGTM - I just tested this out and it worked beautifully.

@dfreese @mfarrugi How do y'all feel about implicitly making `$(execpath)` into an absolute path? On the one hand, it's surprising and un-bazel-y, on the other hand, the fact that we're running actions not in the execpath is also surprising and un-bazel-y, and this feels like a reasonable way to compensate for that weirdness...

cargo/cargo_build_script.bzl
dae: Whoops, good catch - not sure how I managed to do that!

dfreese:
> LGTM - I just tested this out and it worked beautifully.
> 
> @dfreese @mfarrugi How do y'all feel about implicitly making `$(execpath)` into an absolute path? On the one hand, it's surprising and un-bazel-y, on the other hand, the fact that we're running actions not in the execpath is also surprising and un-bazel-y, and this feels like a reasonable way to compensate for that weirdness...

I'm not thrilled by it, particularly since I consider prost-build to be a bad actor by requiring an executable path to be compiled in.  (I ended up patching the env! macros out of the build script), but I don't know if I see a better way around it.

illicitonion:
> > LGTM - I just tested this out and it worked beautifully.
> > @dfreese @mfarrugi How do y'all feel about implicitly making `$(execpath)` into an absolute path? On the one hand, it's surprising and un-bazel-y, on the other hand, the fact that we're running actions not in the execpath is also surprising and un-bazel-y, and this feels like a reasonable way to compensate for that weirdness...
> 
> I'm not thrilled by it, particularly since I consider prost-build to be a bad actor by requiring an executable path to be compiled in. (I ended up patching the env! macros out of the build script), but I don't know if I see a better way around it.

As of https://github.com/danburkert/prost/pull/324 it's at least a runtime rather than compile-time env var :)

dae:
A tidied up version of this is in #468 - I'll close this one in favour of the new one.

PR #464, opened 2020-10-23
Handle external packages using CARGO_MANIFEST_DIR

sitaktif:


illicitonion (approved):
LGTM!

Looks like CI is failing because the root `WORKSPACE` doesn't have a definition for the `hello_cargo_manifest_dir` repository :)

sitaktif:
Fixed!

WORKSPACE
```diff
@@ -79,3 +79,9 @@ docs_deps()
 load("@docs//:docs_transitive_deps.bzl", docs_transitive_deps = "transitive_deps")
 
 docs_transitive_deps(is_top_level = True)
+
+# For the hello_uses_cargo_manifest_dir example.
+local_repository(
```
UebelAndre, line 84: Why does this need to be a repository? And if it needs to be a repository, could it not be put into `example_deps` similar to how `rules_proto` was added to `docs/docs_transitive_deps.bzl`?
https://github.com/bazelbuild/rules_rust/blob/master/docs/docs_transitive_deps.bzl#L20-L31

sitaktif: Great questions!

It needs to be a repository because that's specifically the case we are testing: something that depends on an external target that uses the `CARGO_MANIFEST_DIR` environment variable. If the rules is not external (ie. not in a separate repo) it won't reproduce the issue that this commit is fixing.

I can definitely put the deps in a separate bzl and will do so. Unfortunately I won't be able to reuse that `.bzl` file in `examples/WORKSPACE` because the `path` attribute has to be different, but it will still be a nice improvement to avoid polluting the root `WORKSPACE` 👍

UebelAndre: You could use a bool (`is_top_level`) similar to the snippet I sent you to toggle between the paths. Would that not be sufficient?

sitaktif: You're totally right, I should have read your comment more carefully. Amended as suggested!

examples/examples_transitive_deps.bzl
```diff
@@ -4,14 +4,37 @@ There are some transitive dependencies of the dependencies of the examples'
...
 
     rules_proto_dependencies()
 
     rules_proto_toolchains()
 
     rust_workspace()
+
+    # Needed by the hello_uses_cargo_manifest_dir example.
+    if is_top_level:
+        maybe(
```
UebelAndre, line 28: This should be
```python
        maybe(
            native.local_repository,
            name = "hello_cargo_manifest_dir",
            path = "hello_cargo_manifest_dir",
        )
```
(Sorry if this is obvious, just trying to help 😅)

sitaktif: "just one small change, surely I can't mess that one up". Anyway, thanks for the help :-)

Unrelated, but I noticed that if you build from the `examples/` dir (which creates all the `examples/bazel-*` directories, building from the root directory sometimes fails because of infinite recursion, as if the `.bazelignore` was ignored. I couldn't find how to fix the issue (the issue wasn't introduced in this commit though).

UebelAndre: Yeah... that's what happens when you have nested workspaces that depend on the root workspace like this. `local_repository` should ignore bazel output paths (or accept a pattern to ignore certain directories). 

Hopefully [Bazel's proposal for external dependencies](https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit) resolves this. For now, you'll just have to make sure you `bazel clean` in `./examples`. Or at least that's what I do... 😅

sitaktif:
Ok I think we're good now :-)

UebelAndre:
@sitaktif I know this is a bit late but I was wondering if you'd be willing to merge 
```
examples
├── hello_cargo_manifest_dir
└── hello_uses_cargo_manifest_dir
``` 
to something like:
```
examples
└── manifest_dir
    ├── external_repo
    └── usage
```

Just as a way to bundle the two examples 😅 

(obviously not major, just a drive-by thought)

PR #466, opened 2020-10-26
Add version_file common attr

illicitonion:
This allows folks to generate version files, e.g. using
volatile-status.txt, rather than requiring them to be hard-coded in
BUILD files.

Both ways are still supported, but only one can be used for any
particular target.

I don't love that this moves the version logic into process-wrapper, but the alternatives:
* Need the user to know the 5 different env vars to set in their version_file file
* Write a standalone tool to convert a version_file into a file with the 5 env vars
both seemed worse.

### DIFF
--- a/docs/flatten.md
+++ b/docs/flatten.md
@@ -28,7 +28,7 @@
 
 <pre>
 rust_benchmark(<a href="#rust_benchmark-name">name</a>, <a href="#rust_benchmark-aliases">aliases</a>, <a href="#rust_benchmark-compile_data">compile_data</a>, <a href="#rust_benchmark-crate_features">crate_features</a>, <a href="#rust_benchmark-crate_root">crate_root</a>, <a href="#rust_benchmark-data">data</a>, <a href="#rust_benchmark-deps">deps</a>, <a href="#rust_benchmark-edition">edition</a>,
-               <a href="#rust_benchmark-out_dir_tar">out_dir_tar</a>, <a href="#rust_benchmark-proc_macro_deps">proc_macro_deps</a>, <a href="#rust_benchmark-rustc_env">rustc_env</a>, <a href="#rust_benchmark-rustc_flags">rustc_flags</a>, <a href="#rust_benchmark-srcs">srcs</a>, <a href="#rust_benchmark-version">version</a>)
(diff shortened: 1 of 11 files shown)


dfreese:
Implementation LGTM.  Did you have a use case in mind?  Supporting structopt/clap or similar?  The mix of cargo and bazel seems a little odd, but I remember those crates trying to "just work."

Mostly I'm apprehensive about further expanding the API to accommodate cargo, particularly if it means pulling the bazel side of rules rust further away from, "just calling rustc."

UebelAndre:
Any updates here? Is this still a desired change?

illicitonion:
Sorry for the huge delay here!

> Implementation LGTM. Did you have a use case in mind? Supporting structopt/clap or similar? The mix of cargo and bazel seems a little odd, but I remember those crates trying to "just work."

Yeah, we have a bunch of code which uses libraries which auto-detect versions, and this allows them to get their version numbers from the workspace status command when running with `--stamp`, rather than needing version numbers to be hard-coded in BUILD files.

@wchargin also raised a similar request in https://github.com/bazelbuild/rules_rust/issues/553 wanting to consume from a `Cargo.toml` file - a genrule deriving the information and setting this attribute seems like a reasonable piece of machinery for solving that need (which could be hidden away in a macro, if desired).

See also https://github.com/bazelbuild/rules_rust/pull/556 which I've just raised, which adds a more general mechanism for this, because build scripts can't get workspace status command information from `--stamp` because it's in the wrong configuration to see the data. I think this PR is still worth merging regardless, though, because it avoids the user needing to know the 5 env vars they should be setting.

UebelAndre:
I had need for something like this but found there to be a bit more complications around being able to use volatile workspace status info. So I created https://github.com/bazelbuild/rules_rust/pull/956 which I hope might more-or-less solve for similar issues here.

UebelAndre:
> I had need for something like this but found there to be a bit more complications around being able to use volatile workspace status info. So I created #956 which I hope might more-or-less solve for similar issues here. That said though, it clearly wouldn't solve for usability concerns this PR might be solving.

Actually, this PR is quite nice in how it handles version info. Though, I wonder if anyone else has a need for additional volatile status info. Very torn on what the best path forward is

UebelAndre:
Is this something folks are still working-on/interested-in?

illicitonion:
I don't have a particular need, but can push this forwards if anyone else does - how about you @UebelAndre?

UebelAndre:
> I don't have a particular need, but can push this forwards if anyone else does - how about you @UebelAndre?

I noticed that version info was not correctly being propagated into `rust_doc` rules and will (probably at some point) be looking for a way to do so. Since this would impact similar code paths I wanted to find a conclusion to this PR.

illicitonion:
I'll close, but feel free to revive it, or ask me to, if it'd be useful :)

PR #468, opened 2020-10-29
Support expanding locations in rustc_env and build_script_env #2

### LINKED ISSUE #79: Including a genfile at compile time is not possible

### LINKED ISSUE #454: Bazel build with prost-build build.rs fails with "No such file or directory"

### LINKED ISSUE #459: Can't use `include_str!()` macro with Bazel-generated data files

dae:
This is #461 squashed into fewer commits - please see that PR for previous discussion.

This potentially closes #459, closes #454, and closes #79.

The docs seem to indicate there's precedent for this in rules_cc:
https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables


illicitonion (approved):
Works for me!

One implementation detail worth specifically calling out: This expands to an absolute, rather than relative, path for cargo build scripts, because we don't run them in the execroot. I think this is fine, but it's a bit weird.

I will leave a bit of time for others to chime in before merging :)

cargo/cargo_build_script.bzl
```diff
@@ -4,6 +4,19 @@ load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "ge
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
 
+def _expand_location(ctx, env, data):
+    if env.startswith("$(execpath ") or env.startswith("$(location "):
```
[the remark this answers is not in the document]
dae: I suspect non-prefixed usage won't be common, but it was easy enough to change. Also added a note about it being an absolute path to the docstring.

illicitonion (approved):
Thanks!

dae:
Latest push just corrects some $(execroot) references in the docs to $(execpath)

cargo/cargo_build_script.bzl
```diff
@@ -220,10 +238,19 @@ def cargo_build_script(
...
-        # Environment variables passed during build.rs execution
-        build_script_env = {"CARGO_PKG_VERSION": "0.1.2"},
+        # Optional environment variables passed during build.rs compilation
+        rustc_env = {
+           "CARGO_PKG_VERSION": "0.1.2",
+        },
+        # Optional environment variables passed during build.rs execution.
+        # Note that as the build script's working directory is not execroot,
+        # execpath/location will return an absolute path, instead of a relative
+        # one.
```
mfarrugi, line 248: Does this work under remote-execution? Is it tested by the ci rbe_ubuntu1604 target?

illicitonion: Yes, it should work under remote execution - the expansion is done in the cargo_build_script runner within the action.

The build script could opt to do bad things with the path passing it to dependees (though we have some logic in cargo_build_script to try to normalise that too), but within the action this is valid.

dae:
Rebased on latest master; if there's anything else that needs changing, please let me know.

PR #469, opened 2020-10-30
Correct CARGO_MANIFEST_DIR values for Bazel <3.7.0

sitaktif:
Bazel 3.7.0 introduced a change (301b96b223b0c0) in the way it reports
file paths for external packages. This ensures that rules_rust sets the
right value for CARGO_MANIFEST_DIR both for Bazel < 3.7.0 and >= 3.7.0.

Note that if this changed once, it might change again, so the right
thing to do is to use a bazel API to do that work once it's available.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -386,6 +386,21 @@ def collect_inputs(
     )
     return _process_build_scripts(ctx, file, crate_info, build_info, dep_info, compile_inputs)
 
+def _exec_root_relative_path(ctx, filepath):
+    # Bazel 3.7.0 (commit 301b96b223b0c0) changed how file paths are returned for external packages.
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
This looks good to me to support multiple versions of bazel in the absence of a stable API to get the exec path of the package directory.

@damienmg do you know if there's an API we've overlooked here, or a supported way to look up the exec path of a package directory? It feels like there should be, if these directories are treated as opaque and subject to change (as we saw in bazel 3.7.0)...

UebelAndre:
@damienmg Hey, pinging this PR. Seems pretty important to get in 😅

damienmg:
I am not aware of any API.

Since the breaking change was not marked as breaking that should have been a reason for rollback I guess. It's probably too late now.

Issue #472, opened 2020-11-04
rustdoc: offer web server as a Bazel binary


wchargin:
A `rust_doc` rule spits out a zip file, but that’s not convenient for
actually consulting the docs. It would be great for `rust_doc` to
provide a Bazel binary that can be run to spin up a web server.

Here is one approach. From the Bazel rule, provide this Python file:

```python
import mimetypes
import os
import sys
from wsgiref import simple_server
import zipfile


def main():
    (webfiles_zip_name, default_crate, *args) = sys.argv[1:]
    port = 8000
    if args:
        port = int(args[0])

    webfiles = os.path.join(os.path.dirname(__file__), webfiles_zip_name)
    data = {}
    with open(webfiles, "rb") as fp:
        with zipfile.ZipFile(fp) as zp:
            for path in zp.namelist():
                data[path] = zp.read(path)
    sys.stderr.write("Read %d files from %s\n" % (len(data), webfiles_zip_name))

    default_path = "/%s/index.html" % default_crate

    def app(environ, start_response):
        p = environ.get("PATH_INFO", "/").lstrip("/")
        if not p:
            start_response("302 Found", [("Location", default_path)])
            yield b"302 Found\n"
            return
        if p.endswith("/"):
            p += "index.html"
        blob = data.get(p)
        if not blob:
            start_response("404 Not Found", [])
            yield b"404 Not Found\n"
            return
        (mime_type, encoding) = mimetypes.guess_type(p)
        headers = []
        headers.append(("Content-Type", mime_type))
        if encoding is not None:
            headers.append(("Content-Encoding", encoding))
        start_response("200 OK", headers)
        yield blob

        sys.stderr.write

    server = simple_server.make_server("", port, app)
    # Find which port was actually bound, in case user requested port 0.
    real_port = server.socket.getsockname()[1]
    msg = "Serving %s docs on port %d\n" % (default_crate, real_port)
    sys.stderr.write(msg)
    try:
        server.serve_forever()
    except KeyboardInterrupt:
        print()


if __name__ == "__main__":
    main()
```

Then generate a `py_binary` build target:

```starlark
# Assuming `rust_doc(name = "mylib_doc", dep = ":mylib")`, generate:

py_binary(
    name = "mylib_doc_server",
    srcs = ["mylib_doc_server.py"],
    args = [
        "mylib_doc.zip",  # output of `rust_doc` rule
        "mylib",  # default crate
    ],
[...]

PR #474, opened 2020-11-04
rust_doc: strip directory prefix from archive names

### LINKED ISSUE #471: rustdoc: zip output should not have bazel-out/k8-fastbuild/... prefix
The zip archives emitted by `rust_doc` rules have files all prefixed
with a path derived from the Bazel configuration:

```
$ unzip -l bazel-bin/tensorboard/data/server/rustboard_core_doc.zip | head
Archive:  bazel-bin/tensorboard/data/server/rustboard_core_doc.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/.lock
     1792  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/COPYRIGHT.txt
     4421  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/FiraSans-LICENSE.txt
   186824  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/FiraSans-Medium.woff
   183268  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/FiraSans-Regular.woff
    10847  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/LICENSE-APACHE.txt
     1023  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/tensorboard/data/server/rustboard_core_doc/LICENSE-MIT.txt
```

This makes it hard to work with the archive. Ideally, the whole output
dir prefix would be stripped, so files would be like `LICENSE-MIT.txt`.

This comes from the `args.add_all([input_dir])` call:
https://github.com/bazelbuild/rules_rust/blob/6766c248ef29451d35dfa86eeeef15de6700b7d7/rust/private/rustdoc.bzl#L125

There’s an easy-ish fix for this in pure Starlark:

<details>
<summary>Patch in terms of Starlark `DirectoryExpander`</summary>

```diff
diff --git a/rust/private/rustdoc.bzl b/rust/private/rustdoc.bzl
index 3213e20..154f8cc 100644
--- a/rust/private/rustdoc.bzl
+++ b/rust/private/rustdoc.bzl
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 # buildifier: disable=module-docstring
+load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@io_bazel_rules_rust//rust:private/rustc.bzl", "CrateInfo", "DepInfo", "add_crate_link_flags", "add_edition_flags")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 
@@ -122,7 +123,11 @@ def _zip_action(ctx, input_dir, output_zip):
 
     # Create but not compress.
     args.add("c", output_zip)
-    args.add_all([input_dir], expand_directories = True)
+    args.add_all(
+        [input_dir],
+        expand_directories = False,
+        map_each = _basename_mapper,
+    )
     ctx.actions.run(
         executable = ctx.executable._zipper,
         inputs = [input_dir],
@@ -130,6 +135,14 @@ def _zip_action(ctx, input_dir, output_zip):
         arguments = [args],
     )
 
+def _basename_mapper(root, directory_expander):
+    results = []
+    for leaf in directory_expander.expand(root):
+        rel = paths.relativize(leaf.path, root.path)
+        spec = "%s=%s" % (rel, leaf.path)
+        results.append(spec)
+    return results
+
 rust_doc = rule(
     doc = _rust_doc_doc,
     implementation = _rust_doc_impl,
```

</details>

The catch is that this needs the `DirectoryExpander` API, which was only
added in Bazel 3.4.0. If we’re willing to require that as a minimum
Bazel version, then the fix is easy. But otherwise we need to do a bit
more lifting ourselves. One way to do this would be to write a small
Python tool that performs the same string manipulation as above and then
calls out to the zipper. The approach is similar to `process_wrapper`
but the script would be simpler. Thus:

<details>
<summary>Patch in terms of a new `dir_zipper.py`</summary>

```diff
diff --git a/rust/private/rustdoc.bzl b/rust/private/rustdoc.bzl
index 3213e20..6e1f8d3 100644
--- a/rust/private/rustdoc.bzl
+++ b/rust/private/rustdoc.bzl
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 # buildifier: disable=module-docstring
+load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@io_bazel_rules_rust//rust:private/rustc.bzl", "CrateInfo", "DepInfo", "add_crate_link_flags", "add_edition_flags")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 
@@ -121,13 +122,16 @@ def _zip_action(ctx, input_dir, output_zip):
     args = ctx.actions.args()
 
     # Create but not compress.
-    args.add("c", output_zip)
+    args.add(ctx.executable._zipper)
+    args.add(output_zip)
+    args.add(input_dir.path)
     args.add_all([input_dir], expand_directories = True)
     ctx.actions.run(
-        executable = ctx.executable._zipper,
+        executable = ctx.executable._dir_zipper,
         inputs = [input_dir],
         outputs = [output_zip],
         arguments = [args],
+        tools = [ctx.executable._zipper],
     )
 
 rust_doc = rule(
@@ -159,6 +163,11 @@ rust_doc = rule(
             doc = "File to add in `<body>`, after content.",
             allow_single_file = [".html", ".md"],
         ),
+        "_dir_zipper": attr.label(
+            default = Label("//util/dir_zipper"),
+            cfg = "exec",
+            executable = True,
+        ),
         "_zipper": attr.label(
             default = Label("@bazel_tools//tools/zip:zipper"),
             cfg = "exec",
diff --git a/util/dir_zipper/BUILD b/util/dir_zipper/BUILD
new file mode 100644
index 0000000..49d6d5b
--- /dev/null
+++ b/util/dir_zipper/BUILD
@@ -0,0 +1,7 @@
+py_binary(
+    name = "dir_zipper",
+    srcs = ["dir_zipper.py"],
+    python_version = "PY3",
+    srcs_version = "PY3",
+    visibility = ["//visibility:public"],
+)
diff --git a/util/dir_zipper/dir_zipper.py b/util/dir_zipper/dir_zipper.py
new file mode 100644
index 0000000..80451ae
--- /dev/null
+++ b/util/dir_zipper/dir_zipper.py
@@ -0,0 +1,22 @@
+import os
+import subprocess
+import sys
+
+
+def main():
+    (zipper, outfile, root_dir, *files) = sys.argv[1:]
+    strip_prefix = root_dir + os.path.sep
+    args = []
+    args.append("c")
+    args.append(outfile)
+    for f in files:
+        if not f.startswith(root_dir):
+            raise ValueError("non-descendant: %r not under %r" % (f, root_dir))
+        rel = f[len(strip_prefix) :]
+        spec = "%s=%s" % (rel, f)
+        args.append(spec)
+    subprocess.run([zipper, *args]).check_returncode()
+
+
+if __name__ == "__main__":
+    main()
```

</details>

Either case yields the desired result:

```
$ unzip -l bazel-bin/tensorboard/data/server/rustboard_core_doc.zip | head
Archive:  bazel-bin/tensorboard/data/server/rustboard_core_doc.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2010-01-01 00:00   .lock
     1792  2010-01-01 00:00   COPYRIGHT.txt
     4421  2010-01-01 00:00   FiraSans-LICENSE.txt
   186824  2010-01-01 00:00   FiraSans-Medium.woff
   183268  2010-01-01 00:00   FiraSans-Regular.woff
    10847  2010-01-01 00:00   LICENSE-APACHE.txt
     1023  2010-01-01 00:00   LICENSE-MIT.txt
```

I’m happy to send a PR for either of these patches, and I release them
under the Apache 2.0 license. Please let me know how you’d prefer to
proceed. Thanks!

wchargin:
The Zip archives emitted by `rust_doc` rules used to contain a long
`/bazel-bin/k8-fastbuild/bin/path/to/target` prefix before each archive
entry name. This made it hard to work with the generated archive. This
patch adds a transformation to strip the bin-dir prefix, leaving only
the path to the target. We leave the path to the target so that multiple
`rust_doc` zip outputs within one workspace can be meaningfully expanded
into the same directory.

This could be written in Starlark using the `DirectoryExpander` API, but
that was only introduced in Bazel 3.4.0. Since we have a minimum Bazel
version of 0.17.1, we instead write the helper as a small Rust script.

Fixes #471.

Test Plan:
Unit tests included. As an end-to-end test, run

```
cd examples/ &&
bazel build //hello_world:hello_world_doc &&
unzip -l bazel-bin/hello_world/hello_world_doc.zip
```

to list the contents of the built `hello_world_doc.zip` archive. Before
this patch, the result was like:

```
Archive:  bazel-bin/hello_world/hello_world_doc.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/hello_world/hello_world_doc/.lock
     1792  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/hello_world/hello_world_doc/COPYRIGHT.txt
     4421  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/hello_world/hello_world_doc/FiraSans-LICENSE.txt
...
     2132  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/hello_world/hello_world_doc/storage.js
     1180  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/hello_world/hello_world_doc/theme.js
     3764  2010-01-01 00:00   bazel-out/k8-fastbuild/bin/hello_world/hello_world_doc/wheel.svg
---------                     -------
   890555                     38 files
```

After this patch:

```
Archive:  bazel-bin/hello_world/hello_world_doc.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2010-01-01 00:00   hello_world/hello_world_doc/.lock
     1792  2010-01-01 00:00   hello_world/hello_world_doc/COPYRIGHT.txt
     4421  2010-01-01 00:00   hello_world/hello_world_doc/FiraSans-LICENSE.txt
...
     2132  2010-01-01 00:00   hello_world/hello_world_doc/storage.js
     1180  2010-01-01 00:00   hello_world/hello_world_doc/theme.js
     3764  2010-01-01 00:00   hello_world/hello_world_doc/wheel.svg
---------                     -------
   890555                     38 files
```

wchargin-branch: rustdoc-strip-prefix


util/dir_zipper/dir_zipper_lib.py
```diff
@@ -0,0 +1,26 @@
...
+    strip_prefix = root_dir + os.path.sep
+    args = []
+    args.append("c")
+    args.append(output)
+    for f in files:
+        if not f.startswith(root_dir):
+            raise ValueError("non-descendant: %r not under %r" % (f, root_dir))
+        rel = f[len(strip_prefix) :]
+        spec = "%s=%s" % (rel, f)
+        args.append(spec)
```
dfreese, line 25: Is similar logic not possible within starlark?

rust/repositories.bzl
```diff
@@ -61,6 +61,13 @@ def rust_repositories(
         type = "zip",
     )
 
+    maybe(
+        http_archive,
+        name = "rules_python",
+        url = "https://github.com/bazelbuild/rules_python/releases/download/0.1.0/rules_python-0.1.0.tar.gz",
+        sha256 = "b6d46438523a3ec0f3cead544190ee13223a52f6a6765a29eae7b7cc24cc83a0",
+    )
```
dfreese, line 69: I've run into a lot of problems with rules_python around python versioning and package dependencies.  While it's use in this PR is fairly contained, it does feel like a pretty heavy dependency to bring in.

rust/private/rustdoc.bzl
```diff
@@ -119,15 +119,16 @@ def _zip_action(ctx, input_dir, output_zip):
         output_zip (File): The location of the output archive containing generated documentation
     """
     args = ctx.actions.args()
-
-    # Create but not compress.
-    args.add("c", output_zip)
+    args.add("--zipper", ctx.executable._zipper)
+    args.add("--output", output_zip)
+    args.add("--root-dir", input_dir.path)
```
dfreese, line 124: Should this be the input_dir.path, rather than the workspace root?  It seems like you should be able to unzip all of your docs files into a single folder and have them laid out appropriately.

I'm not super familiar with the structure of rust_docs pages, so feel free to correct me if I'm wrong on that.

util/dir_zipper/dir_zipper_lib.py
wchargin: The logic is possible in Starlark; the tricky bit is expanding the
directory to its list of contents. I haven’t found a way to do that
other than using the `DirectoryExpander` API added in Bazel 3.4.0.

If there is such a way, I would be very happy to switch to it instead of
calling out to this utility.

rust/private/rustdoc.bzl
wchargin: Hmm; I’m not sure that I understand the question?

The directory layout at this stage of the Bazel action is like:

```
bazel-bin/k8-fastbuild/path/to/target/LICENSE.txt
bazel-bin/k8-fastbuild/path/to/target/favicon.ico
bazel-bin/k8-fastbuild/path/to/target/crate_name/index.html
bazel-bin/k8-fastbuild/path/to/target/crate_name/struct.Foo.html
bazel-bin/k8-fastbuild/path/to/target/src/mylib/lib.rs.html
```

and we want to create a zip archive that has files

```
LICENSE.txt
favicon.ico
crate_name/index.html
crate_name/struct.Foo.html
src/mylib/lib.rs.html
```

so that we can point a web server inside this zip file to serve docs.
The value of `input_dir` is `bazel-bin/k8-fastbuild/path/to/target`, so
we do so by stripping off `input_dir.path`.

Is the comment that we instead want to preserve one level of directory
in the output zip, like this?

```
target/LICENSE.txt
target/favicon.ico
target/crate_name/index.html
target/crate_name/struct.Foo.html
target/src/mylib/lib.rs.html
```

If so, this is a reasonable suggestion, and I can make that change if
you like.

rust/repositories.bzl
wchargin: Noted. @UebelAndre and I discussed this a bit in #472. I agree that the
Pip side of `rules_python` is unorthodox, but I figured that this might
be okay since we only need the standard library. Does that seem okay to
you? If not, do you have an alternative suggestion for a toolchain to
handle #471 and #472?

wchargin (commented):
Thanks for the quick review! Happy to discuss further.

rust/repositories.bzl
UebelAndre: Looking at the `dir_zipper` it seems like that could be implemented in Rust using just `std`. The only thing I feel made sense to author in python was the webserver for hosting docs. I don't think `rules_python` is necessary for the goal of this pull request.

rust/private/rustdoc.bzl
dfreese: Thanks for the clear example.  I was attempting to suggest going from:
```
bazel-bin/k8-fastbuild/path/to/target/LICENSE.txt
bazel-bin/k8-fastbuild/path/to/target/favicon.ico
bazel-bin/k8-fastbuild/path/to/target/crate_name/index.html
bazel-bin/k8-fastbuild/path/to/target/crate_name/struct.Foo.html
bazel-bin/k8-fastbuild/path/to/target/src/mylib/lib.rs.html
```

to
```
path/to/target/LICENSE.txt
path/to/target/favicon.ico
path/to/target/crate_name/index.html
path/to/target/crate_name/struct.Foo.html
path/to/target/src/mylib/lib.rs.html
```

which, in my head, would preserve the aggregation of all rustdoc zip files within any given bazel project into a single directory.  Does that make more sense?

rust/repositories.bzl
dfreese: I would agree with @UebelAndre here in that I think rust gives enough support here to handle this.  In general, unless it's particularly prohibitive, it's generally good to keep rules_rust as self-sufficient as possible.  (It's also easier to assume that people will be more familiar with rust/starlark when working within this project)

I think there is some precedent for this with the cargo_build_script_runner
https://github.com/bazelbuild/rules_rust/blob/master/cargo/cargo_build_script_runner/bin.rs

wchargin: Okay, sure: rewritten as a Rust script. (Also wins better handling for
non-UTF-8 path names; loses nice arg parsing and tests, since we don’t
have a stdlib `zipfile` equivalent.)

rust/private/rustdoc.bzl
wchargin: Got it. I think I can work with that. Will address (not done yet).

UebelAndre:
What's the status here? Is this ready for review again?

wchargin:
No; I haven’t gotten around to addressing all comments. Will re-request
review through the GitHub UI once I have.

rust/private/rustdoc.bzl
wchargin: Done.

UebelAndre:
Friendly ping here. This change sounds like a great addition 😃

wchargin:
Okay: I’ve rebased master. This patch seems to work fine. I’m having
trouble with the build setup for #475, but I agree that this patch is
nice to have by itself, so feel free to merge if you like.

dfreese (approved):

PR #477, opened 2020-11-05
Add support for building for wasm32-wasi target.

PiotrSikora:


docs/index.md
```diff
@@ -77,11 +77,17 @@ Currently the most common approach to managing external dependencies is using
 
 ## WebAssembly
 
-To build a `rust_binary` for wasm32-unknown-unknown add the `--platforms=@io_bazel_rules_rust//rust/platform:wasm` flag.
+To build a `rust_binary` for `wasm32-unknown-unknown` target add the `--platforms=@io_bazel_rules_rust//rust/platform:wasm32-unknown` flag.
 
 ```command
-bazel build @examples//hello_world_wasm --platforms=@io_bazel_rules_rust//rust/platform:wasm
+bazel build @examples//hello_world_wasm --platforms=@io_bazel_rules_rust//rust/platform:wasm32-unknown
```
PiotrSikora, line 83: I'm not sure if the new name should be `wasm32-unknown` or `wasm32-unknown-unknown` to match Rust's triple.

UebelAndre: I'd prefer it to match the triple

PiotrSikora: Actually, this doesn't work, since then `@io_bazel_rules_rust//rust/platform:wasm32-unknown-unknown` **platform** collides with `@io_bazel_rules_rust//rust/platform:wasm32-unknown-unknown` **constraint**.

This is actually quite problematic because there is no good way to name `wasm32-wasi` without creating collisions. We need to have `wasm32-wasi` **constraint**, and we need to have `wasi` **os**, but we also want to have **platform** that's either `wasm32-wasi` or simply `wasi`, but we cannot have either right now.

I think triples need to stay where they are, since a lot of external projects depends on them, but since **cpu** and **os** are mostly internal, we could move them elsewhere (e.g. `//rust/platform/cpu:wasm32`, `//rust/platform/os:unknown` and `//rust/platform/os:wasi`).

What do you think? Any preferences?

PiotrSikora: OK, All triple constraints are left in place, I moved the individual constraints to `//rust/platform/cpu:wasm32`, `//rust/platform/os:unknown` and `//rust/platform/os:wasi`, and renamed the new platform to `wasi`.

I think it looks good now, and I cross-tested it with `cargo-raze` and my own project that relies on those.

UebelAndre: This looks good to me. Is there a paired change needed in `cargo-raze`?

UebelAndre (approved):

UebelAndre:
@damienmg @mfarrugi perhaps some more thought could go info the future state of these constraint definitions. I'm wondering if the complications @PiotrSikora ran into might be avoidable given some slight restructuring there.

rust/platform/triple_mappings.bzl
```diff
@@ -140,6 +147,17 @@ def abi_to_constraints(abi):
...
+
+    Args:
+        triple (str): A platform triple. eg: `x86_64-unknown-linux-gnu`
+
+    Returns:
+        str: A system name
+    """
+    if triple == "wasm32-wasi":
+        return "wasi"
+
```
dfreese, line 160: With the name change from wasm32-unknown-wasi to wasm32-wasi, was there a broader change in how targets are specified, or is this still just kind of a snowflake?

What I'm trying to get at is whether or not this function, per-spec could just return the last section of the target or not.

rust/platform/cpu/BUILD
```diff
@@ -0,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
```
dfreese, line 1: nit: can you default to private and set public explicitly?

rust/platform/os/BUILD
```diff
@@ -0,0 +1,11 @@
+package(default_visibility = ["//visibility:public"])
```
dfreese, line 1: nit: can you default to private and set public explicitly?

dfreese (commented):
The mix of platforms and config settings in //rust/platform does leave something to be desired, but this generally looks reasonable to me.

rust/platform/triple_mappings.bzl
PiotrSikora: It looks that there is also a tuple for `{aarch64,x86_64}-fuchsia`, but I don't know whether that's a broader change or not.

https://doc.rust-lang.org/nightly/rustc/platform-support.html

rust/platform/cpu/BUILD
PiotrSikora: Done.

rust/platform/os/BUILD
PiotrSikora: Done.

dfreese (approved):

visortelle:
@PiotrSikora , @dfreese is there any docs or example of how to use rust_library compiled as wasi as a `deps` entry of another rust_library?

PiotrSikora:
@visortelle we're using it in [Proxy-Wasm Rust SDK](https://github.com/proxy-wasm/proxy-wasm-rust-sdk) (see: [BUILD](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/BUILD) for the library, which depends on `hashbrown` and `log` crates,  and [examples/BUILD](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/examples/BUILD) for the example binaries using it). The `BUILD` files for third-party crates are generated using [cargo-raze](https://github.com/google/cargo-raze).

visortelle:
> @visortelle we're using it in [Proxy-Wasm Rust SDK](https://github.com/proxy-wasm/proxy-wasm-rust-sdk) (see: [BUILD](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/BUILD) for the library, which depends on `hashbrown` and `log` crates, and [examples/BUILD](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/examples/BUILD) for the example binaries using it). The `BUILD` files for third-party crates are generated using [cargo-raze](https://github.com/google/cargo-raze).

Thank you!

PR #480, opened 2020-11-06
Links attribute

sitaktif:
* Add the Cargo.toml `links` attribute to our rule to allow inferring depenv environment variable names correctly.
* Replace the sandbox directory with `${pwd}` in the values of the depenv environment variables.
* Add missing transitive dependencies between build script targets.
* Add an example to test this non-trivial case.

This allows building packages with sys transitive dependencies (e.g. when depending on crates such as `git2`).


examples/complex_sys/raze/remote/libgit2-sys-0.12.14+1.1.0.BUILD.bazel
```diff
@@ -0,0 +1,151 @@
...
+        "--cap-lints=allow",
+    ],
+    crate_features = [
+      "https",
+      "libssh2-sys",
+      "openssl-sys",
+      "ssh",
+      "ssh_key_from_memory",
+    ],
+    # LOCALLY MODIFIED
```
illicitonion, line 82: For other reviewers: These files are slightly modified from the cargo-raze generated versions, because they require attributes raze doesn't currently set. This feels like a pretty reasonable approach, as if people re-gen the `git diff` will show the local modifications to re-apply (at least, until Raze supports them)?

cargo/cargo_build_script.bzl
```diff
@@ -165,6 +167,10 @@ _build_script_run = rule(
         "crate_name": attr.string(
             doc = "Name of the crate associated with this build script target",
         ),
+        "links": attr.string(
+            default = "",
+            doc = "The name of the native library this crate links against.",
```
illicitonion, line 172: Can you run `docs/update_docs.sh` to update the docs?

illicitonion (approved):
LGTM

Windows CI is failing - I'm guessing adding a windows condition to the openssl BUILD file which says shared libraries should be `.dll` _may_ fix it? (Also Buildifier CI is failing)

@dfreese How do you feel about the openssl example? I think it's very useful to have...

sitaktif:
Making an attempt at fixing the windows build (using CI as I don't have a Windows machine handy).

examples/third_party/openssl/openssl.BUILD
```diff
@@ -12,6 +12,12 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+config_setting(
+    name = "windows",
+    values = {"cpu": "windows"},
```
illicitonion, line 17: I think possibly this?
```suggestion
    values = {"cpu": "x64_windows"},
```

examples/complex_sys/raze/remote/libgit2-sys-0.12.14+1.1.0.BUILD.bazel
UebelAndre: Is there an issue open on `cargo-raze` for what you ran into here?

cargo/cargo_build_script.bzl
UebelAndre: is `default = "",` necessary?

examples/complex_sys/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "rules_rust_examples_complex_sys"
+path = "src/main.rs"
+
+[dependencies]
+git2 = "=0.13.12"
+
+[raze]
+workspace_path = "//complex_sys/raze"
+genmode = "Remote"
+incompatible_relative_workspace_path = true
```
UebelAndre, line 15: this line should not be necessary anymore. The default is `true` now

examples/third_party/openssl/openssl.BUILD
sitaktif: I ended up doing something that felt cleaner but please check just in case.

illicitonion: Works for me

dae: Rather than defining new config settings, couldn't you use @io_bazel_rules_rust//rust/platform:darwin, :windows, and/or conditions:default?

sitaktif: That's what I started doing... until I undid the changes and ended up skipping CI for `@examples//complex_sys/...` on Windows as plenty of issues arise when building. Note that the `@examples//hello_sys/...` one is already skipped on Windows for similar reasons.

sitaktif: (note: I did not convert the existing ones to preset platforms since I did not touch that part and want to avoid any additional fuss on this PR)

sitaktif:
@UebelAndre @dae @illicitonion I think I addressed your comments, PTAL!

examples/complex_sys/raze/remote/libgit2-sys-0.12.14+1.1.0.BUILD.bazel
sitaktif: @UebelAndre I opened the following two issues:

- https://github.com/google/cargo-raze/issues/287
- https://github.com/google/cargo-raze/issues/288

sitaktif:
Thanks!

UebelAndre:
@sitaktif Hey, sorry to ping this out of the blue, I was trying to regenerate the outputs of the `complex_sys` example using `cargo-raze` version `0.8` but couldn't get it to compile. This does indeed seem complex since you're building OpenSSL 😅. If you have time in the future, could I get you to generate a lockfile and update the `[raze]` settings so future re-runs won't break this? 🙏

sitaktif:
@UebelAndre ACK - looking to fix that right now 👍

sitaktif:
I opened https://github.com/bazelbuild/rules_rust/pull/549 to address (assuming I understood the issue!)

PR #481, opened 2020-11-06
Updated documentation

UebelAndre:
Added some missing documentation from #468 and #469

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -1,21 +1,65 @@
 # buildifier: disable=module-docstring
+load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "C_COMPILE_ACTION_NAME")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
-load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_compile_env", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
+load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
 load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
 
 def _expand_location(ctx, env, data):
-    for directive in "$(execpath ", "$(location ":
+    """A trivial helper for `_expand_locations`
+
+    Args:
+        ctx (ctx): The rule's context object
+        env (str): The value possibly containing location macros to expand.
+        data (sequence of Targets): see `_expand_locations`
+
+    Returns:
+        string: The location-macro expanded version of the string.
+    """
+    for directive in ("$(execpath ", "$(location "):
         if directive in env:
             # build script runner will expand pwd to execroot for us
             env = env.replace(directive, "${pwd}/" + directive)
     return ctx.expand_location(env, data)
 
-def _expand_locations(ctx):
-    "Expand $(execpath ...) references in user-provided env vars."
-    env = ctx.attr.build_script_env
-    data = getattr(ctx.attr, "data", [])
+def _expand_locations(ctx, env, data):
+    """Performs location-macro expansion on string values.
+
+    Note that exec-root relative locations will be exposed to the build script
+    as absolute paths, rather than the ordinary exec-root relative paths,
+    because cargo build scripts do not run in the exec root.
+
+    Args:
+        ctx (ctx): The rule's context object
+        env (dict): A dict whose values we iterate over
+        data (sequence of Targets): The targets which may be referenced by
+            location macros. This is expected to be the `data` attribute of
+            the target, though may have other targets or attributes mixed in.
+
+    Returns:
+        dict: A dict of environment variables with expanded location macros
+    """
     return dict([(k, _expand_location(ctx, v, data)) for (k, v) in env.items()])
+
+def get_cc_compile_env(cc_toolchain, feature_configuration):
+    """Gather cc environment variables from the given `cc_toolchain`
+
+    Args:
+        cc_toolchain (cc_toolchain): The current rule's `cc_toolchain`.
+        feature_configuration (FeatureConfiguration): Class used to construct command lines from CROSSTOOL features.
+
+    Returns:
+        dict: Returns environment variables to be set for given action.
+    """
+    compile_variables = cc_common.create_compile_variables(
+        feature_configuration = feature_configuration,
+        cc_toolchain = cc_toolchain,
+    )
+    return cc_common.get_environment_variables(
+        feature_configuration = feature_configuration,
+        action_name = C_COMPILE_ACTION_NAME,
+        variables = compile_variables,
+    )
 
 def _build_script_impl(ctx):
     """The implementation for the `_build_script_run` rule.
@@ -101,7 +145,11 @@ def _build_script_impl(ctx):
     for f in ctx.attr.crate_features:
         env["CARGO_FEATURE_" + f.upper().replace("-", "_")] = "1"
 
-    env.update(_expand_locations(ctx))
+    env.update(_expand_locations(
+        ctx,
+        ctx.attr.build_script_env,
+        getattr(ctx.attr, "data", []),
+    ))
 
     tools = depset(
         direct = [
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -17,11 +17,16 @@ load("@bazel_skylib//lib:versions.bzl", "versions")
 load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
     "CPP_LINK_EXECUTABLE_ACTION_NAME",
-    "C_COMPILE_ACTION_NAME",
 )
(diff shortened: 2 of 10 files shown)


rust/private/rustc.bzl
```diff
@@ -310,15 +316,6 @@ def get_linker_and_args(ctx, cc_toolchain, feature_configuration, rpaths):
 
     return ld, link_args, link_env
 
-def _expand_locations(ctx, env, aspect):
```
UebelAndre, line 313: @illicitonion @dae Do you know why is this different from the macro that was in`cargo/cargo_build_script.bzl`?

rust/private/utils.bzl
```diff
@@ -123,3 +123,34 @@ def _get_preferred_artifact(library_to_link):
         library_to_link.interface_library or
         library_to_link.dynamic_library
     )
+
+def expand_locations(ctx):
```
UebelAndre, line 127: @illicitonion @dae could you add some suggestions here and for `_expand_location` to help me fill in the documentation?

rust/private/rustc.bzl
dae: - The one in rustc is concerned with the compilation stage (rustc_env), so it needs to work when aspects like clippy are enabled. process_wrapper does not support ${pwd} expansion in environmental variables at the moment, so execpath/location will not work - only an unadorned $(rootpath ...) will work
- The one in cargo applies to the environment passed in when running the build script (build_script_env). $(rootpath ...) won't work here because the working directory is not execroot, and the build script runner has been modified to replace ${pwd} references with execroot so that the execroot-relative paths work correctly.
- env is a string possibly containing $(location/execpath/rootpath) references
- data is an attr list of tools from the calling rule that will be used to expand the location references into actual paths

UebelAndre: Is there a way to unify the functionality? Or do you feel it's more correct to have these separate functions exist?

dae: Because of the differences I'm not sure using a single routine is a big win, as you'd need to document the two different use cases it's trying to cover. It may be possible to get location/execpath to work with rustc_env in the future, but that would require modifying process_wrapper, and my C++ is rusty at best.

UebelAndre: I can update my PR to restore the macros but could you write me some documentation to copy and paste for them? The documentation was really the only thing that bugged me but then thought there might be an easy win to deduplicate code.

rust/private/rustc.bzl
```diff
@@ -395,6 +409,16 @@ def collect_inputs(
     return _process_build_scripts(ctx, file, crate_info, build_info, dep_info, compile_inputs)
 
 def _exec_root_relative_path(ctx, filepath):
+    """[summary]
```
UebelAndre, line 412: @sitaktif Hey, can you add suggestions to this PR to fill out this documentation? 🙏

rust/private/rustc.bzl
UebelAndre: @dae Ok, I've restored the code but made one small change. Can you give me some text to populate the template with?

cargo/cargo_build_script.bzl
```diff
@@ -5,16 +5,33 @@ load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
 load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")
 
 def _expand_location(ctx, env, data):
-    for directive in "$(execpath ", "$(location ":
+    """[summary]
```
illicitonion, line 8:
```suggestion
    """Performs location-macro expansion on string values.
    
    Note that exec-root relative locations will be exposed to the build script as absolute paths, rather than the ordinary exec-root relative paths, because cargo build scripts do not run in the exec root.
```

cargo/cargo_build_script.bzl
```diff
@@ -5,16 +5,33 @@ load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
...
+
+    Args:
+        ctx (ctx): [description]
+        env ([type]): [description]
```
illicitonion, line 12:
```suggestion
        env (str): The value possibly containing location macros to expand.
```

cargo/cargo_build_script.bzl
```diff
@@ -5,16 +5,33 @@ load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
...
+        data ([type]): [description]
```
illicitonion, line 13:
```suggestion
        data (sequence of Targets): The targets which may be referenced by location macros. This is expected to be the `data` attribute of the target, though may have other targets or attributes mixed in.
```

cargo/cargo_build_script.bzl
```diff
@@ -5,16 +5,33 @@ load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
...
+
+    Returns:
+        [type]: [description]
```
illicitonion, line 16:
```suggestion
        string: The location-macro expanded version of the string.
```

cargo/cargo_build_script.bzl
```diff
@@ -5,16 +5,33 @@ load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain")
...
         if directive in env:
             # build script runner will expand pwd to execroot for us
             env = env.replace(directive, "${pwd}/" + directive)
     return ctx.expand_location(env, data)
 
-def _expand_locations(ctx):
-    "Expand $(execpath ...) references in user-provided env vars."
-    env = ctx.attr.build_script_env
-    data = getattr(ctx.attr, "data", [])
+def _expand_locations(ctx, env, data):
```
illicitonion, line 24: As above, except env is a dict whose values we iterate over. I'd probably only bother documenting this function - the other is a trivial helper with identical args.

rust/private/rustc.bzl
illicitonion: The text here should be identical to the one I suggested above, but without the caveat about absolute paths (because `rustc` _does_ run in the exec root, so exec-root-relative paths are fine)

rust/private/rustc.bzl
illicitonion, line 412:
```suggestion
    """Get the path to filepath, relative to the exec root.
```

rust/private/rustc.bzl
```diff
@@ -395,6 +409,16 @@ def collect_inputs(
...
+
+    Args:
+        ctx ([type]): [description]
+        filepath ([type]): [description]
```
illicitonion, line 416:
```suggestion
        filepath (string): Path of the file to get. This should be relative to the root of the workspace containing the file, such as that returned by `Label.package` or `ctx.build_file_path`
```

rust/private/rustc.bzl
```diff
@@ -395,6 +409,16 @@ def collect_inputs(
...
+
+    Returns:
+        [type]: [description]
```
illicitonion, line 419:
```suggestion
        string: The path relative to the exec root.
```

rust/private/rustc.bzl
```diff
@@ -569,7 +593,12 @@ def construct_arguments(
                 env["CARGO_BIN_EXE_" + dep_crate_info.output.basename] = dep_crate_info.output.short_path
 
     # Update environment with user provided variables.
-    env.update(_expand_locations(ctx, crate_info.rustc_env, aspect))
+    # TODO: explain significance of `aspect`
+    env.update(_expand_locations(
+        ctx,
+        crate_info.rustc_env,
+        getattr(ctx.rule.attr, "data", []) if aspect else getattr(ctx.attr, "data", []),
```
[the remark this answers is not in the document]
illicitonion: Have a read of https://docs.bazel.build/versions/master/skylark/aspects.html - it explains why we need to call `ctx.rule.attr` when we're running in an aspect.

How aspects work isn't really something we should be documenting every time we use `ctx`. But maybe we could extract a function which looks like:

```python
def rule_attrs(ctx, aspect):
    """Gets a rule's attributes.

    As per https://docs.bazel.build/versions/master/skylark/aspects.html when we're executing from an aspect we need to get attributes of a rule differently to if we're not in an aspect.

    Args:
        aspect (bool): Whether we're running in an aspect
"""
    return ctx.rule.attr if aspect else ctx.attr
```

And update the places we do this lookup to use that function? That way, we have the documentation in one useful place (and potentially avoid duplication).

rust/private/rustc.bzl
UebelAndre: Thanks friends! 😄 🙏

rust/private/rustc.bzl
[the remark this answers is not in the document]
UebelAndre: I added that macro. Thanks for the info 😄

cargo/cargo_build_script.bzl
dae: Thanks for filling out those docs. Just a minor clarification on the first rustc comment  below - the docs should recommend only using rootpath in that case, as execpath will fail in the case of generated sources (can be reproduced by changing the second reference to GENERATED_DATA or SOME_TOOL in examples/env_locations/BUILD)

UebelAndre:
This PR is ready for review for anyone who has merging powers.

rust/private/rustc.bzl
sitaktif: Sorry I was a bit late to respond, I see that @illicitonion beat me to it (thanks!).

sitaktif:
Thanks @UebelAndre!

UebelAndre:
@illicitonion Hey, you around to merge this real quick? 🙏 😅

illicitonion (approved):

UebelAndre:
Thank you!

PR #486, opened 2020-11-10
Stable way to infer CARGO_MANIFEST_DIR

sitaktif:
A new way that will not break with new versions of Bazel was proposed.

This is cleaner and should get in before the next version of Bazel,
which will change the behaviour of `ctx.build_file_path`.

See https://github.com/bazelbuild/bazel/issues/12344#issuecomment-721815138 - thanks @lberki for the suggestion <3


lberki:
You're welcome! :)

illicitonion (approved):

PR #494, opened 2020-11-14
use target platform for CARGO_CFG_TARGET_*

dae:
build_script_runner was gathering details based on the host platform
instead of the target platform. This was causing a cross-compile of the
ring crate to fail to link, as it was making build-time decisions based
on the host OS instead of the target one.


illicitonion (approved):

illicitonion:
Thanks!

Issue #499, opened 2020-11-20
RFC: Rename workspace from `io_bazel_rules_rust` to `rules_rust`


UebelAndre:
According to https://docs.bazel.build/versions/master/skylark/deploying.html#workspace these rules should be named `rules_rust`. I think they should be renamed. Maybe for backwards compatibility the `rust_workspace` macro could define a [bind](https://docs.bazel.build/versions/master/be/workspace.html#bind) for users to ease the transition but I'd definitely like to see this changed. 

What are the concerns in doing this?

mfarrugi:
+1

I imagine for most cases a `sd 'io_bazel_rules_rust' 'rules_rust'  **/*` would suffice.

UebelAndre:
I've opened https://github.com/bazelbuild/rules_rust/pull/500 for the conversion. Given that there are no releases (which I think there should be, per https://github.com/bazelbuild/rules_rust/issues/415) I'm unsure how the transition should be communicated. I did the same thing `rules_python` did in https://github.com/bazelbuild/rules_python/pull/212 by in the way of messaging for users within the code.

acmcarther:
This should probably not be merged. This ruleset is named "io_bazel" to be consistent with the other rules defined in the bazelbuild org.

See rules go: https://github.com/bazelbuild/rules_go. Their workspace is named "io_bazel_rules_go". I don't know the history behind this naming scheme, but I do know that rules_go has "high-tier" support internally, so we should follow their lead on things.

I'm going to mirror this into the issue.

UebelAndre:
> This should probably not be merged. This ruleset is named "io_bazel" to be consistent with the other rules defined in the bazelbuild org.
> 
> See rules go: https://github.com/bazelbuild/rules_go. Their workspace is named "io_bazel_rules_go". I don't know the history behind this naming scheme, but I do know that rules_go has "high-tier" support internally, so we should follow their lead on things.
> 
> I'm going to mirror this into the issue.

This is in direct violation of the documentation in https://docs.bazel.build/versions/master/skylark/deploying.html#workspace

I would assert either that documentation needs to change or this PR should be merged in a reasonable time.

damienmg:
Yep this documentation is pretty recent and I found some more context in bazel-dev@. Historically we took the decision to recommend reverse qualified domain name as naming convention (like Go) but it seems they prefer shorter name and use github orgs for rules.

[...]

Issue #415, opened 2020-09-24
Releases


UebelAndre:
Why aren't there releases for this repo? I feel like this is something every rules repository should have. Is there anything stopping this repo from having releases?

damienmg:
We are all working in companies that use a work at head model in which release does not really make sense. As this repositories is supposed to be used at source level there is no compelling reason to have release and release induce toil.

mfarrugi:
I think that it is useful to tag the minimum (and maximum?) bazel version
required, since those are the only breaking changes we have made in the
past. This was a source of toil for me in the past, lining things up w/ the
very frequent bazel updates.

I think the 'live at head' approach has only been the case for
contributors, as most people will want to include their changes by using
the latest version.

UebelAndre:
> I think that it is useful to tag the minimum (and maximum?) bazel version required

> most people will want to include their changes by using the latest version.

I agree with both of these points.

UebelAndre:
Would it be possible to get a release? Is there something in CI that tests the oldest supported version?

mfarrugi:
Sure, but what do you need from a release? It is little more than a tag.

It looks like CI can (
https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#using-a-specific-version-of-bazel),
but currently doesn't.

UebelAndre:
A release to me is an indication of what kinds of changes were made to a code base. It's little more than a tag but the value in it is the semantic versioning. If a neat new feature was added, I should be able to infer that from the release number.

I think testing should be done on some modern version and it should be up to the contributors/maintainers to keep up with modern Bazel releases and be aware of potential incompatibilities with older versions. But I don't think this testing is all that valuable without releases to represent these kinds of changes.

UebelAndre:
@mfarrugi @damienmg In the issue linked above, a user had an issue due to change that broke compatibility with older versions of Bazel. I think `rules_rust` should have releases and they should indicate breakages or changes in the API but if nothing else I believe the changes in https://github.com/bazelbuild/rules_rust/pull/463 call for a new release to communicate.

UebelAndre:
[...]

PR #500, opened 2020-11-22
Renamed workspace name from `io_bazel_rule_rust` to `rules_rust`

UebelAndre:
Generated by running
```bash
LANG=C sed -i '' -e 's/io_bazel_rules_rust/rules_rust/' $(find . -type f -not -path "./.git/*")
```

There's some small amount of sorting in here as well (based on the new name)

Closes https://github.com/bazelbuild/rules_rust/issues/499

Blocked by https://github.com/google/cargo-raze/pull/298 and https://github.com/google/cargo-raze/issues/304


UebelAndre:
@mfarrugi with the release of [cargo-raze 0.8.0](https://github.com/google/cargo-raze/commit/4b1bf2225e79ea87639c9b814acd63b3c0c166c6) this PR is ready to go.

UebelAndre:
@mfarrugi Just a heads up, It'd be great to get this in in a timely manner given the messaging in `cargo-raze`:
```
WARNING: The default of `[*.raze.rust_rules_workspace_name]` will soon be set to `"rules_rust"`. Please explicitly set this flag to prevent a change in behavior or upgrade your code to use the latest version of `rules_rust` and change references of `io_bazel_rules_rust` to `rules_rust` in your project.
```
I understand this is has larger implication for users but I feel confident that with the snippet above in this PR (or just a find and replace in common editors), users can easily update the their projects to allow for this switch.

UebelAndre:
@illicitonion Hey, @mfarrugi seems to have been away for some time, would you be able to review this PR?

acmcarther:
This should probably not be merged. This ruleset is named "io_bazel" to be consistent with the other rules defined in the bazelbuild org.

See rules go: https://github.com/bazelbuild/rules_go. Their workspace is named "io_bazel_rules_go". I don't know the history behind this naming scheme, but I do know that rules_go has "high-tier" support internally, so we should follow their lead on things.

I'm going to mirror this into the issue.

UebelAndre:
> This should probably not be merged. This ruleset is named "io_bazel" to be consistent with the other rules defined in the bazelbuild org.
> 
> See rules go: https://github.com/bazelbuild/rules_go. Their workspace is named "io_bazel_rules_go". I don't know the history behind this naming scheme, but I do know that rules_go has "high-tier" support internally, so we should follow their lead on things.
> 
> I'm going to mirror this into the issue.

This is in direct violation of the documentation in https://docs.bazel.build/versions/master/skylark/deploying.html#workspace

> If your rules belong to the bazelbuild organization, you must use rules_<lang> (e.g. rules_mockascript)

I would assert either that documentation needs to change or this PR should be merged in a reasonable time.

UebelAndre:
@mfarrugi re https://github.com/google/cargo-raze/pull/323#issuecomment-749433550

I'd like to talk about potential backwards compatibility shims here. I don't know of many options. To me this change is something users would have to deal with if they wanted to continue to accept changes from master. My stance is still that the changes here have a low impact on users since they can run the command in the description to update. Happy to make a shim if someone can explain what that looks like though. But I feel the sooner these changes are in the better for everyone. Otherwise things are likely to go the way of [rules_go](https://github.com/bazelbuild/rules_go/blob/master/WORKSPACE#L1) where it's acquired so many users that it becomes much more difficult to fix the name.

mfarrugi:
The only option I know of is a bind, but I have never used it and am not
confident that it actually works.

UebelAndre:
[bind](https://docs.bazel.build/versions/master/be/workspace.html#bind) only appears to work on targets. Not entire repositories.

UebelAndre:
@mfarrugi Yeah, I'm not able to get this to work with `bind`. Can you think of any other options?

acmcarther:
In the other link you referenced rules python which already made this change, so I'm not concerned about this being an issue anymore. Discussions about impact on users might still be prudent though.

UebelAndre:
@mfarrugi I tried asking in the [slack channel](https://bazelbuild.slack.com/archives/CA31HN1T3/p1608672195323600) but also wasn't able to find great solutions for backward compatibility. I think the only thing to do would be to pull the trigger. I discussed this with @illicitonion a bit and he suggests that this be coupled a new release model (https://github.com/bazelbuild/rules_rust/issues/415) and official support for Bazel 4.0 where the min supported version follows the new LTS model introduced in that version of Bazel. Since very little changes to add 4.0 as a min version (since the rules, according to bazelisk, is already compatible). So I feel this could be merged and announcements can follow about changes to compatibility.

UebelAndre:
I spent quite a lot of time looking into ways ways to support both workspace names. The two blockers are both related to the use of generated repositories and how they refer to `rules_rust`.

1. TODO
2. TODO

rust/platform/triple_mappings.bzl
```diff
@@ -186,9 +186,15 @@ def triple_to_constraint_set(triple):
         list: A list of constraints (each represented by a list of strings)
     """
     if triple == "wasm32-wasi":
-        return ["@io_bazel_rules_rust//rust/platform/cpu:wasm32", "@io_bazel_rules_rust//rust/platform/os:wasi"]
+        return [
+            "@rules_rust//rust/platform/cpu:wasm32",
```
UebelAndre, line 190: These strings get rendered into the BUILD files for the generated toolchains. There needs to be a way to detect the current `rules_rust` workspace name and render that into there. There may also be some potential for `repo_mapping` but that may also require the ability to dynamically detect the workspace name or force users to explicitly pass this data which to me seems undesirable.

proto/raze/remote/BUILD.base64-0.9.3.bazel
```diff
@@ -7,7 +7,7 @@ DO NOT EDIT! Replaced on runs of cargo-raze
 
 # buildifier: disable=load
 load(
-    "@io_bazel_rules_rust//rust:rust.bzl",
+    "@rules_rust//rust:rust.bzl",
```
UebelAndre, line 10: This BUILD file (as with all the others generated via `cargo-raze` treats `rules_rust` as an external repository. The `rust_rules_workspace_name` setting can be used to control the name but in order to really support multiple names there would need to be multiple versions of these files that have the new name. Again, the blocker here is the inability to detect the workspace name dynamically.

UebelAndre:
@mfarrugi @damienmg @illicitonion This PR is ready to go. I'm not sure what else should be done for this. @illicitonion talked about doing the rename at the same time as the 4.0 migration (which I think is a no-op since the rules are already compatible). Maybe https://github.com/bazelbuild/rules_rust/issues/445 can be done at the same time? And that would conclude "breaking" changes for the near future?

damienmg:
I can take a look at what is needed for #445 next week but I don't think it should block this PR.

UebelAndre:
I'm happy to have this merged 😃  is there anything else I should do here?

UebelAndre:
@damienmg hey, just pinging this again 😅

damienmg (approved):

damienmg:
Sorry I cannot address #445 at the moment, I am fighting fires at work and in my personal life :(

Related to #445, I received a bug that we have `whitelist` somewhere in our codebase, suggesting we replaced it with allowlist (definitely a good idea IMO).

UebelAndre:
Thanks @damienmg! I've really enjoyed having you as a reviewer and would love to keep working with you. I hope everything bounces back for you and am wishing you the best 😄

PR #503, opened 2020-11-24
Allow usage of $(location ...) in rustc_env as well, to include_str!() generated files.

dae:
The ${pwd} resolution in build_script_env is necessary during the compilation stage as well, as otherwise we can't use things like include_str!() on generated files. Starting with a test that demonstrates the issue; will follow up with a fix.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -2,44 +2,8 @@
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "C_COMPILE_ACTION_NAME")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "DepInfo", "get_cc_toolchain", "get_compilation_mode_opts", "get_linker_and_args")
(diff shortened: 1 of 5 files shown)


dae:
For some reason I was labouring under the impression that this would require a change to process_wrapper, but it appears we can just rely on the existing substitution. In any case, the docstring in utils.bzl explains the need for this:

```
    $(execroot ...) and $(location ...) are prefixed with ${pwd},
    which process_wrapper and build_script_runner will expand at run time
    to the absolute path. This is necessary because include_str!() is relative
    to the currently compiled file, and build scripts run relative to the
    manifest dir, so we can not use execroot-relative paths.

    $(rootpath ...) is unmodified, and is useful for passing in paths via
    rustc_env that are encoded in the binary with env!(), but utilized at
    runtime, such as in tests. The absolute paths are not usable in this case,
    as compilation happens in a separate sandbox folder, so when it comes time
    to read the file at runtime, the path is no longer valid.
```

dae:
@UebelAndre this also allows us to move the function into utils as you wanted :-)

illicitonion (approved):
Works for me! Will give others a little time to chime in before merging

dae:
Rebased over master; gentle ping.

illicitonion:
Thanks! Sorry for the delay :)

PR #384, opened 2020-07-28
rust-analyzer support

benbrittain:
#378 This commit adds support for a `rust_analyzer` rule

PR #505, opened 2020-11-26
Add rust_analyzer rule to generate a rust-project.json

djmarcin:
This PR builds on the work in #384 to add a rust_analyzer rule which generates a rust-project.json. This file allows rust-analyzer to function without cargo by describing the code structure, layout, and dependency graph.

I think this PR is ready for review, but looking to get feedback on the following:
- The rule requires the user to manually enter their `bazel info execution_root` in order to locate remote crates and out_dirs. Is there a better way to retrieve this from inside starlark without resorting to genrules? I'd like to leave a genrule rewrite to a later date to unblock the relatively large win this PR brings.
- The rule assumes that the generated rust-project.json will be used directly from `bazel-bin` and generates relative paths for the source code under the workspace root. If there is a way to get the absolute path to the workspace root, this would not be necessary.
- How should this be tested?

Other than that, I think this is good to go.


rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+
+    toolchain = find_toolchain(ctx)
+
+    cfgs = []
+    for feature in ctx.rule.attr.crate_features:
+        cfgs.append("feature=\"" + feature + "\"")
+    for flag in ctx.rule.attr.rustc_flags:
+        # --cfg flags should be passed as well but as an atomic
+        # config, not key/value
+        if flag.startswith("--cfg"):
```
UebelAndre, line 58: Should this not be `--cfg=`?

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+        crate = crate_info,
+        cfgs = cfgs,
+        env = env,
+        deps = deps,
+        transitive_deps = transitive_deps,
+        proc_macro_dylib_path = proc_macro_dylib_path,
+        build_info = build_info,
+    )]
+
+rust_analyzer_aspect = aspect(
```
UebelAndre, line 94: Can you add something for `doc`?

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+                output["crates"].append(crate)
+
+        # Add this crate to the crate mapping and output.
+        crate_id, crate = create_crate(ctx, target[RustAnalyzerInfo], crate_mapping)
+        if crate_id not in crate_mapping:
+            crate_mapping[crate_id] = idx
+            idx += 1
+            output["crates"].append(crate)
+
+    ctx.actions.write(output = ctx.outputs.filename, content = struct(**output).to_json())
```
UebelAndre, line 187: You might want to add a todo or something to replace the `to_json` with https://docs.bazel.build/versions/master/skylark/lib/json.html just to let future authors know about it

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+
+rust_analyzer = rule(
+    attrs = {
+        "targets": attr.label_list(
+            aspects = [rust_analyzer_aspect],
+            doc = "List of all targets to be included in the index",
+        ),
+        "exec_root": attr.string(
```
UebelAndre, line 195: Very interested in avoiding this. Is there no way to get this infer this information?

rust/repositories.bzl
```diff
@@ -493,6 +511,32 @@ def _load_rust_compiler(ctx):
 
     return compiler_build_file
 
+def _load_rust_src(ctx):
+    """Loads the rust source code. Used by the rust-analyzer rust-project.json generator.
```
UebelAndre, line 515: Can you add type annotations to the docstring? 🙏 😅

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+        # Add this crate to the crate mapping and output.
+        crate_id, crate = create_crate(ctx, target[RustAnalyzerInfo], crate_mapping)
+        if crate_id not in crate_mapping:
+            crate_mapping[crate_id] = idx
+            idx += 1
+            output["crates"].append(crate)
+
+    ctx.actions.write(output = ctx.outputs.filename, content = struct(**output).to_json())
+
+rust_analyzer = rule(
```
UebelAndre, line 189: Can you als fill in `doc` for this as well?

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+    )]
+
+rust_analyzer_aspect = aspect(
+    attr_aspects = ["deps", "proc_macro_deps"],
+    implementation = _rust_analyzer_aspect_impl,
+    toolchains = ["@io_bazel_rules_rust//rust:toolchain"],
+)
+
+def create_crate(ctx, info, crate_mapping):
+    """Creates a crate in the rust-project.json format
```
UebelAndre, line 101: Can you add type annotations here as well?

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+load("//rust/platform:triple_mappings.bzl", "system_to_dylib_ext", "triple_to_system")
+
+# We support only these rule kinds.
+_rust_rules = [
+    "rust_library",
+    "rust_binary",
+]
+
+RustAnalyzerInfo = provider(
+    "RustAnalyzerInfo holds rust crate metadata for targets",
```
UebelAndre, line 34: Not, can you use the `doc` keyword here?

UebelAndre (commented):
A few comments from a non maintainer 😀
 
You should also wire all this up to stardoc and run `docs/update_docs.sh`.

This is looking super awesome!!

rust/private/rust_analyzer.bzl
UebelAndre: > The rule requires the user to manually enter their bazel info execution_root in order to locate remote crates and out_dirs. Is there a better way to retrieve this from inside starlark without resorting to genrules? I'd like to leave a genrule rewrite to a later date to unblock the relatively large win this PR brings. 

Can you explain more on why needing `execution_root` is necessary?

rust/private/rust_analyzer.bzl
djmarcin: The `=` is optional, and could be a space. Updated to be explicit about both possibilities.

rust/private/rust_analyzer.bzl
djmarcin: Added a TODO, it's not in any released version yet

rust/private/rust_analyzer.bzl
djmarcin: I agree, this is the biggest issue with this PR. I believe the one real viable alternative is setting a custom `--workspace_status_command` https://docs.bazel.build/versions/master/user-manual.html#flag--workspace_status_command to provide this information via an environment variable, but very interested to know if I missed another way.

UebelAndre: Maybe @damienmg or @mfarrugi might know?

The thing I want from this PR is to be able to push a change to my repo and have enabled `rust-analyzer` support for anyone with no additional work. Having someone need to explicitly set the output of `bazel info execution_root` sounds like it wouldn't support that. I think `--workspace_status_command` might be something that works? Since I could commit a `.bazelrc` file to the repo to enable these rules. But definitely interested in finding something more contained. I'm trying to find such a solution but am having no luck 😞

djmarcin: The reason why it's necessary to get this information is that we must provide a path to various external source files, e.g. the rust sysroot, any third party crates you depend on, etc. All of the paths generated during the action analysis phase are relative to the execution root, and are not available via the bazel-* symlinks, so it's not possible to generate a relative path from the user's source directory to the external sources.

I'm toying with the idea of a simple genrule that can get execution_root and uses sed as a template expander, but it's not portable (e.g. to windows). There is a built-in action `ctx.actions.expand_template`, but unfortunately in our case the substitution would be in an output and that rule requires it to be available as a starlark variable, so I don't think it will work.

rust/repositories.bzl
djmarcin: None of the _functions in this file have type annotations, leaving for another PR

rust/private/rust_analyzer.bzl
djmarcin: I should also mention that I'm really not sure how `--workspace_status_command` or even the genfiles solution would work, because you can't recursively run `bazel info` in order to find the workspace at runtime. So, passing the `exec_root` is still the best solution I have that works at the moment, unsatisfying as it is.

UebelAndre: If there's a path forward with `genrule`, you could use [cmd_ps](https://docs.bazel.build/versions/master/be/general.html#genrule.cmd_ps) or [cmd_bat](https://docs.bazel.build/versions/master/be/general.html#genrule.cmd_bat) to handle the windows specific behavior, right?

rust/repositories.bzl
UebelAndre: That's just because I hadn't gotten to this file when updating all the docs 😅. I don't think anyone should be expected to update an entire file if they touch it, but I do think new macros should have them.

rust/private/rust_analyzer.bzl
UebelAndre: Ah cool 👍 

The `[6:]` was throwing me off and seemed like an oversight. Thanks for updating that 🙏

rust/private/rust_analyzer.bzl
djmarcin: Ok a few updates here:
1) The `bazel-{workspace}` symlink does point at the execution root, I don't know how I missed that before. I'm not sure this actually helps because I'm not sure I can get the workspace name in a normal (non-repository) rule either, and also because of #2

2) Due to how symlinks work, `bazel-bin/..` is **not** your current directory, it is actually a path in the execution root. For example, if `bazel-bin -> /home/vagrant/.cache/bazel/_bazel_vagrant/71ed060a30795ab0736b520a305b19cd/execroot/__main__/bazel-out/k8-fastbuild/bin` then `bazel-bin/..` is actually `.../execroot/__main__/bazel-out/k8-fastbuild`
This has a couple of consequences. First, the way this PR resolves local source code is wrong, but for whatever reason it doesn't look like this is breaking anything. Second, we could drop the exec_root parameter if we were willing to make the assumption that the execroot is at bazel-bin/../../../, which I _think_ is valid and fairly safe. This is assuming that the generated rust-project.json will be used directly from bazel-bin, which is the current assumption made in this PR.

djmarcin: For some reason I don't fully understand yet, switching to entirely relative paths based on this assumption breaks rust-analyzer.

djmarcin:
I've been thinking about this PR and I'm dissatisfied with the way it puts the rust-project.json in the bazel-bin folder. This has several problems: 1) it gets wiped out on clean, 2) it gets wiped out when changing compilation modes, 3) it can't use relative paths to find the real workspace root, 4) it isn't automatically detected by rust-analyzer

I think a better way to go is in the vein of utilities like buildifier or gazelle, where the user runs a target that generates a rust-project.json in the root of their workspace. This also plays well better with things like 1) a CI operation to ensure rust-project.json is up to date, and 2) automatically being detected by rust-analyzer. Using a binary job would also allow us to process the output from build scripts to set env vars.

djmarcin:
Updated the PR. The expected usage is now for users to `bazel run @io_bazel_rules_rust//tools/analzyer` which expects a `rust_analyzer` rule at the repo root named `//:rust_analyzer` with dependencies on all rust targets, runs it to generate a templatized rust-project.json at `bazel-bin/rust-project.json`, and then expand the template with the absolute exec_root path and copies it to the workspace root where it can be checked in.

This eliminates the need to enter an exec_root parameter on the rule and should automatically work with rust-analyzer because the file should generally be auto-detected at the workspace root.

Future work will be to eliminate the need for the `//:rust_analyzer` and auto-detect rust targets in the workspace.

UebelAndre:
Do you know if the `rust-analyzer` plugin can be made to build using Bazel and not cargo?

rust/private/rust_analyzer.bzl
djmarcin: This is fixed by adding the gazelle-like utility run target.

djmarcin:
> Do you know if the `rust-analyzer` plugin for vscode can be made to build using Bazel and not cargo?

I'm doing the following until I figure out a path to land it:
- Fork rules_rust and add --error-format=json to the clippy rule
- Use a vscode task to run `ibazel build --aspects=@io_bazel_rules_rust//rust:rust.bzl%rust_clippy_aspect --output_groups=clippy_checks //...` piped to a horrible looking `jq` filter, with a custom problem matcher. Unfortunately the human readable output of rustc does not contain enough information to be parsed via regexp, and the json format contains input too complicated to be parsed via regexp.

This gives similar problem output as generated by rust-analyzer using `cargo clippy`.

custom tasks.json: https://gist.github.com/djmarcin/bb5a799db6764764777c724dd971187e

UebelAndre:
Oh, I just found https://rust-analyzer.github.io/manual.html#non-cargo-based-projects

aignas (commented):
Disclaimer: An outsider review.

Overall LGTM. I think there is a bug in the rust code in a way that we are not checking the error when we are deleting the file and we are assuming that the removal only fails if the file is not there, which may not be the case in general.

I have left a few other `starlark` related comments, some of them may improve the performance and in general I would suggest using `struct` in order to make it easier to read/understand the current JSON schema.

Can't wait to see this working on my machine!

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,202 @@
...
+# TODO(djmarcin): Run the cargo_build_scripts to gather env vars correctly.
+def _rust_project_impl(ctx):
+    rust_toolchain = find_toolchain(ctx)
+    crate_mapping = dict()
+
+    output = dict()
+    output["sysroot_src"] = "__EXEC_ROOT__/" + rust_toolchain.rust_lib.label.workspace_root + "/lib/rustlib/src/library"
+    output["crates"] = []
+
+    # Gather all crates and their dependencies into an array.
```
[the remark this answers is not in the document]
djmarcin: I appreciate the general idea of using struct, but it does change the output resulting in a substantially larger (10-20%) output file due to a lot of extra `"proc_macro_dylib_path": null` and `"source": null` on every crate. I don't think this causes rust-analyzer's behavior to change, but I'd rather keep the output json clean of null fields.

I changed it to only generate the crate dict when the crate id does not already exist in the mapping, though.

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,202 @@
...
+    cfgs = []
+    for feature in ctx.rule.attr.crate_features:
+        cfgs.append("feature=\"" + feature + "\"")
+    for flag in ctx.rule.attr.rustc_flags:
+        # --cfg flags should be passed as well but as an atomic
+        # config, not key/value
+        if flag.startswith("--cfg ") or flag.startswith("--cfg="):
+            cfgs.append(flag[6:])
+
+    env = {}
```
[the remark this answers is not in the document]
djmarcin: No, just cruft I forgot to clean up.

rust/private/rust_analyzer.bzl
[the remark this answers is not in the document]
aignas: Thanks a lot for doing investigation. Leaving a dict makes sense.

aignas (approved):

tools/analyzer/main.rs
```diff
@@ -0,0 +1,78 @@
...
+        .arg("build")
+        .arg(&analyzer_target)
+        .output()
+        .expect("failed to execute bazel process");
+    if !output.status.success() {
+        panic!(
+            "bazel build failed:({}) of {}:\n{}",
+            output.status,
+            analyzer_target,
+            std::str::from_utf8(&output.stderr).unwrap()
```
illicitonion, line 34: Let's do a lossy conversion here, rather than an unwrap - it'd be more useful to see garbled output than no output.

tools/analyzer/main.rs
```diff
@@ -0,0 +1,78 @@
...
+        );
+    }
+
+    let generated_rust_project = repo_root.join("bazel-bin/rust-project.json");
```
illicitonion, line 38:
```suggestion
    let generated_rust_project = repo_root.join("bazel-bin").join("rust-project.json");
```

In theory we support Windows :)

tools/analyzer/main.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // The generated_rust_project has a template string we must replace with the workspace name.
+    let generated_json = fs::read_to_string(&generated_rust_project)
+        .expect("failed to read generated rust-project.json");
+
+    // It's OK if the file doesn't exist.
+    match fs::remove_file(&workspace_rust_project) {
+        Ok(_) => {}
+        Err(err) if err.kind() == ErrorKind::NotFound => {},
+        Err(_) => panic!("Unexpected error removing old rust-project.json"),
```
illicitonion, line 54:
```suggestion
        Err(err) => panic!("Unexpected error removing old rust-project.json: {}", err),
```

tools/analyzer/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+    fs::write(
+        workspace_rust_project,
+        generated_json.replace("__EXEC_ROOT__", &canonical_exec_root.to_string_lossy()),
+    )
+    .expect("failed to write workspace rust-project.json");
+}
+
+fn workspace_dir() -> Option<PathBuf> {
+    if let Ok(ws_dir) = env::var("BUILD_WORKSPACE_DIRECTORY") {
```
illicitonion, line 64:
```suggestion
    if let Some(ws_dir) = env::var_os("BUILD_WORKSPACE_DIRECTORY") {
```

tools/analyzer/main.rs
```diff
@@ -0,0 +1,78 @@
...
+            "bazel build failed:({}) of {}:\n{}",
+            output.status,
+            analyzer_target,
+            std::str::from_utf8(&output.stderr).unwrap()
+        );
+    }
+
+    let generated_rust_project = repo_root.join("bazel-bin/rust-project.json");
+    let workspace_rust_project = repo_root.join("rust-project.json");
+    let canonical_exec_root = fs::canonicalize(repo_root.join(format!(
```
illicitonion, line 40: As we're already invoking `bazel` in this program, can we run `bazel info execution_root` to get this path, rather than making assumptions about symlinks?

(If not, can you make an `OsString` (with `push`), rather than formatting into a `String`? Avoids needing the lossy conversion (I know we're going to do one later when we JSONify, but minimising their scope to just when we're crossing a serialization/deserialization boundary is nice :)))

tools/analyzer/main.rs
illicitonion, line 38: This path assumes the label isn't overridden - can you add a trivial label parser, so that if `BAZEL_ANALYZER_TARGET=//hello_lib:hello_rust_analyzer` you join `bazel-bin/hello_lib/rust-project.json`?

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+
+rust_analyzer = rule(
+    attrs = {
+        "targets": attr.label_list(
+            aspects = [rust_analyzer_aspect],
+            doc = "List of all targets to be included in the index",
+        ),
+    },
+    outputs = {
+        "filename": "rust-project.json",
```
illicitonion, line 200: This creates a hard requirement that there's at most one `rust_analyzer` target per package. It would be nice to soften that if we can. Does this _need_ to be called exactly `rust-project.json`? If not, can we mix the name of the target in, like `hello_rust_analyzer-rust-project.json`?

illicitonion (commented):
This looks very exciting! Can I make sure I understand your expected user flow here? (Which we should document if/when we're settled on it :)):

Currently, I read it as:
1. I manually create a `rust_analyzer` target for whatever top-level targets I want (its transitive deps will automatically be covered too).
2. I run `BAZEL_ANALYZER_TARGET=//hello_lib:hello_rust_analyzer bazel run @io_bazel_rules_rust//tools/analyzer` to generate a `rust-project.json` in the root of my WORKSPACE (so only one of these can be "active" at a time)

And that maybe in the future, we can get rid of step 1, and skip setting the env var in 2, and auto-detect by running `bazel query` all of the targets in the repo?

If so, I think in general that's a reasonable approach - I don't know if there's a better one (cc @dfreese @mfarrugi @damienmg), but it works for me absent other suggestions. Also, can you update the PR description to show that expected flow?

Assuming my read of this is right, a few high-level thoughts:

1. Do we _need_ the `rust_analyzer` target at all? Could the rust binary instead run something like `bazel build --aspects=rust_analyzer //some:rust_library //some/other:rust_library` (possibly also specifying `--output_groups` if needed) to drive the aspect execution?
2. If we keep the `rust_analyzer` target, I'd consider moving it to a top-level `analyzer` directory of the repo, just like cargo build script support is in a top-level `cargo` directory, rather than merging it into `rust.bzl`
3. Any thoughts on the rust binary taking flags vs env vars?

rust/repositories.bzl
```diff
@@ -493,6 +511,32 @@ def _load_rust_compiler(ctx):
...
+    tool_suburl = produce_tool_suburl("rustc", "src", ctx.attr.version, ctx.attr.iso_date)
+    static_rust = ctx.os.environ.get("STATIC_RUST_URL", "https://static.rust-lang.org")
+    url = "{}/dist/{}.tar.gz".format(static_rust, tool_suburl)
+
+    tool_path = produce_tool_path("rustc", "src", ctx.attr.version)
+    archive_path = tool_path + ".tar.gz"
+    ctx.download(
+        url,
+        output = archive_path,
+        sha256 = ctx.attr.sha256s.get(tool_suburl) or FILE_KEY_TO_SHA.get(tool_suburl) or "",
```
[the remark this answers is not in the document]
djmarcin: Without the empty string you are forced to supply a sha256 in your WORKSPACE file or there is an error. Falling back to the empty string allows you to avoid setting a sha256 (and bazel will tell you the sha it used in an info), and matches the behavior of all the other tools we fetch.

djmarcin:
Thanks for the review! I'll respond quickly to a few things and look more over the next couple days.

> This looks very exciting! Can I make sure I understand your expected user flow here? (Which we should document if/when we're settled on it :)):
> 
> Currently, I read it as:
> 
> 1. I manually create a `rust_analyzer` target for whatever top-level targets I want (its transitive deps will automatically be covered too).
> 2. I run `BAZEL_ANALYZER_TARGET=//hello_lib:hello_rust_analyzer bazel run @io_bazel_rules_rust//tools/analyzer` to generate a `rust-project.json` in the root of my WORKSPACE (so only one of these can be "active" at a time)
> 
> And that maybe in the future, we can get rid of step 1, and skip setting the env var in 2, and auto-detect by running `bazel query` all of the targets in the repo?
> 
> If so, I think in general that's a reasonable approach - I don't know if there's a better one (cc @dfreese @mfarrugi @damienmg), but it works for me absent other suggestions. Also, can you update the PR description to show that expected flow?
> 
> Assuming my read of this is right, a few high-level thoughts:
> 
> 1. Do we _need_ the `rust_analyzer` target at all? Could the rust binary instead run something like `bazel build --aspects=rust_analyzer //some:rust_library //some/other:rust_library` (possibly also specifying `--output_groups` if needed) to drive the aspect execution?
> 2. If we keep the `rust_analyzer` target, I'd consider moving it to a top-level `analyzer` directory of the repo, just like cargo build script support is in a top-level `cargo` directory, rather than merging it into `rust.bzl`
> 3. Any thoughts on the rust binary taking flags vs env vars?

djmarcin:
Thanks for the review! I'll respond to a few things quickly and look at the rest over the next few days.

> This looks very exciting! Can I make sure I understand your expected user flow here? (Which we should document if/when we're settled on it :)):
> 
> Currently, I read it as:
> 
> 1. I manually create a `rust_analyzer` target for whatever top-level targets I want (its transitive deps will automatically be covered too).
> 2. I run `BAZEL_ANALYZER_TARGET=//hello_lib:hello_rust_analyzer bazel run @io_bazel_rules_rust//tools/analyzer` to generate a `rust-project.json` in the root of my WORKSPACE (so only one of these can be "active" at a time)
> 
> And that maybe in the future, we can get rid of step 1, and skip setting the env var in 2, and auto-detect by running `bazel query` all of the targets in the repo?
> 
> If so, I think in general that's a reasonable approach - I don't know if there's a better one (cc @dfreese @mfarrugi @damienmg), but it works for me absent other suggestions. Also, can you update the PR description to show that expected flow?

`BAZEL_ANALYZER_TARGET` is optional, it defaults to //:rust_analyzer. In theory you could put rust-project.json anywhere, rust-analyzer just autodetects it when it's at the root.

> Assuming my read of this is right, a few high-level thoughts:
> 
> 1. Do we _need_ the `rust_analyzer` target at all? Could the rust binary instead run something like `bazel build --aspects=rust_analyzer //some:rust_library //some/other:rust_library` (possibly also specifying `--output_groups` if needed) to drive the aspect execution?

I was hoping we could do this but I couldn't figure a way to. The problem is that the rust_analyzer is able to iterate over each target and pull out the transitive closure of dependencies, and then write the dependency graph in a way that rust-analyzer expects. Without the rule, the aspects can generate output but there's nothing to tie them all together.

> 2. If we keep the `rust_analyzer` target, I'd consider moving it to a top-level `analyzer` directory of the repo, just like cargo build script support is in a top-level `cargo` directory, rather than merging it into `rust.bzl`
> 3. Any thoughts on the rust binary taking flags vs env vars?
Sure.

Some broader thoughts on the "is this a reasonable approach" question:
Since putting up this PR, our team has been using this full time for about a week and rust-analyzer is unfortunately kind of clunky to use via the rust-project.json mechanism. Because the interaction is `bazel run` on a target that doesn't depend on your rust_analyzer target, ibazel can't detect changes, and even if it could, rust-analyzer doesn't appear to respond when rust-project.json changes anyway, so you still need to manually restart it. Even so, it has been pretty useful and unblocked us enough to move our whole repo to bazel.

A radically different idea I've been toying with is to use the bazel build event stream to emulate `cargo metadata` and `cargo check`. If we can do this, then it may be possible to trick rust-analyzer into thinking it's using cargo, which uses codepaths with better UX, by hijacking these commands and issuing bazel commands instead. I haven't made a deep enough exploration here to say this is definitely possible, but our team is currently emulating cargo check via a simple `jq` wrapper around bazel build with the clippy aspect and json format output and it seems to work very well. Unfortunately, rust-analyzer does not currently allow overriding the `cargo metadata` command, so that's a bit of a blocker to more prototyping of this idea.

rust/private/rust_analyzer.bzl
djmarcin: It's a much better UX for it to be named rust-project.json because rust-analyzer will automatically detect it without further configuration. More details on the file format here: https://rust-analyzer.github.io/manual.html

illicitonion:
> > 1. Do we _need_ the `rust_analyzer` target at all? Could the rust binary instead run something like `bazel build --aspects=rust_analyzer //some:rust_library //some/other:rust_library` (possibly also specifying `--output_groups` if needed) to drive the aspect execution?
> 
> I was hoping we could do this but I couldn't figure a way to. The problem is that the rust_analyzer is able to iterate over each target and pull out the transitive closure of dependencies, and then write the dependency graph in a way that rust-analyzer expects. Without the rule, the aspects can generate output but there's nothing to tie them all together.

I guess you could write enough information in the output to be able to infer this... So if you output a list of deps, the rust tool could re-construct the dependency graph based on them... It may mean departing from writing exactly a `rust-project.json` per target, and instead writing a custom JSON/proto file, from which we'd generate the `rust-project.json`.
 
> A radically different idea I've been toying with is to use the bazel build event stream to emulate `cargo metadata` and `cargo check`. If we can do this, then it may be possible to trick rust-analyzer into thinking it's using cargo, which uses codepaths with better UX, by hijacking these commands and issuing bazel commands instead. I haven't made a deep enough exploration here to say this is definitely possible, but our team is currently emulating cargo check via a simple `jq` wrapper around bazel build with the clippy aspect and json format output and it seems to work very well. Unfortunately, rust-analyzer does not currently allow overriding the `cargo metadata` command, so that's a bit of a blocker to more prototyping of this idea.

Folks at JetBrains have done something similar for rules_scala, by implement [BSP](https://build-server-protocol.github.io/) support: https://github.com/JetBrains/bazel-bsp driven by the BEP. It could be nice to try to re-use that if any of it is useful :) (cc @jastice @andrefmrocha)

rust/private/rust_analyzer.bzl
illicitonion: But that's only needed for the one generated in the top-level WORKSPACE directory, right, not for the templatey one generated in bazel-bin?

andrefmrocha:
> > A radically different idea I've been toying with is to use the bazel build event stream to emulate `cargo metadata` and `cargo check`. If we can do this, then it may be possible to trick rust-analyzer into thinking it's using cargo, which uses codepaths with better UX, by hijacking these commands and issuing bazel commands instead. I haven't made a deep enough exploration here to say this is definitely possible, but our team is currently emulating cargo check via a simple `jq` wrapper around bazel build with the clippy aspect and json format output and it seems to work very well. Unfortunately, rust-analyzer does not currently allow overriding the `cargo metadata` command, so that's a bit of a blocker to more prototyping of this idea.
> 
> Folks at JetBrains have done something similar for rules_scala, by implement [BSP](https://build-server-protocol.github.io/) support: https://github.com/JetBrains/bazel-bsp driven by the BEP. It could be nice to try to re-use that if any of it is useful :) (cc @jastice @andrefmrocha)

I'm not particularly familiarized with rust tooling or what would you need to obtain from the BEP, however I would be more than glad to help out figure what would be needed for this :smile: . A first move would definitely be to identify if the BEP already outputs everything you may need. If that is not the case, it is most likely interesting to produce a solution without that dependency (since a change into bazel may take quite a while to be merged down).

UebelAndre:
Friendly ping from a very excited user 😄 Any updates here?

purkhusid:
Is there anything blocking this except an review from the maintainers?

djmarcin:
> > > A radically different idea I've been toying with is to use the bazel build event stream to emulate `cargo metadata` and `cargo check`. If we can do this, then it may be possible to trick rust-analyzer into thinking it's using cargo, which uses codepaths with better UX, by hijacking these commands and issuing bazel commands instead. I haven't made a deep enough exploration here to say this is definitely possible, but our team is currently emulating cargo check via a simple `jq` wrapper around bazel build with the clippy aspect and json format output and it seems to work very well. Unfortunately, rust-analyzer does not currently allow overriding the `cargo metadata` command, so that's a bit of a blocker to more prototyping of this idea.
> > 
> > 
> > Folks at JetBrains have done something similar for rules_scala, by implement [BSP](https://build-server-protocol.github.io/) support: https://github.com/JetBrains/bazel-bsp driven by the BEP. It could be nice to try to re-use that if any of it is useful :) (cc @jastice @andrefmrocha)
> 
> I'm not particularly familiarized with rust tooling or what would you need to obtain from the BEP, however I would be more than glad to help out figure what would be needed for this 😄 . A first move would definitely be to identify if the BEP already outputs everything you may need. If that is not the case, it is most likely interesting to produce a solution without that dependency (since a change into bazel may take quite a while to be merged down).

After looking at the BEP (admittedly not deeply) I don't think it's particularly well suited to replicating the `cargo metadata` style output. We've been using the integration as is and while it's not perfect it's definitely better than not having rust-analyzer support.

I think my preference would be to clean this up for merging mostly as is and make incremental updates to solve some of the rough edges. The top ones that I can think of are:
- When adding new crates it's necessary to regenerate the rust-project.json. We solve this with a VSCode build task that runs `cd ${workspaceFolder} && bazel run @io_bazel_rules_rust//tools/analyzer` and restarting rust-analyzer when things get wonky.
- Hand in hand with the first, you have to maintain a magic top level "rust_analyzer" rule that depends on the output of roughly `bazel query 'kind("rust_library|rust_binary", //...:all)'` and has visibility to all those targets.
- After `bazel clean` rust-analyzer won't be able to find any dependencies until you rerun a `fastbuild` over the targets you're interested in.
- The flags `--remote_download_toplevel` and `--remote_download_minimal` don't work well if you have proc macros, because the proc macro `.so` files are not downloaded.

I probably have some time this coming week to address some of the outstanding comments.

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+
+"""
+Rust Analyzer Bazel rules.
+
+rust_analyzer will generate a rust-project.json file for the
+given targets. This file can be consumed by rust-analyzer as an alternative
+to Cargo.toml files.
+"""
+
+load("@io_bazel_rules_rust//rust:private/rustc.bzl", "BuildInfo", "CrateInfo")
```
UebelAndre, line 23: These should be loaded using workspace relative paths
```
load("//rust:private/rustc.bzl", "BuildInfo", "CrateInfo")
```

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+        deps = deps,
+        transitive_deps = transitive_deps,
+        proc_macro_dylib_path = proc_macro_dylib_path,
+        build_info = build_info,
+    )]
+
+rust_analyzer_aspect = aspect(
+    attr_aspects = ["deps", "proc_macro_deps"],
+    implementation = _rust_analyzer_aspect_impl,
+    toolchains = ["@io_bazel_rules_rust//rust:toolchain"],
```
UebelAndre, line 92: Can you update this to 
```python
toolchains = [str(Label("//rust:toolchain"))],
```
?

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+    doc = "Annotates rust rules with RustAnalyzerInfo later used to build a rust-project.json",
+)
+
+_exec_root_tmpl = '__EXEC_ROOT__/'
+
+def _crate_id(crate_info):
+    '''Returns a unique stable identifier for a crate
```
UebelAndre, line 99: nit inconsistent docstring format, `'''` vs `"""`

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+        "targets": attr.label_list(
+            aspects = [rust_analyzer_aspect],
+            doc = "List of all targets to be included in the index",
+        ),
+    },
+    outputs = {
+        "filename": "rust-project.json",
+    },
+    implementation = _rust_project_impl,
+    toolchains = ["@io_bazel_rules_rust//rust:toolchain"],
```
UebelAndre, line 203: Can you update this to
```python
toolchains = [str(Label("//rust:toolchain"))],
```

rust/repositories.bzl
```diff
@@ -263,6 +263,23 @@ filegroup(
 )
 """
 
+def BUILD_for_rustc_src():
+    """Emits a BUILD file for the rustc src extracted files."""
+
+    return """
```
UebelAndre, line 269: nit, can you move this block into a private variable to match the existing pattern in this file? Or maybe just use a private global variable since there's not templating happening here? Wouldn't be upset if you just left this though.

rust/rust.bzl
```diff
@@ -34,6 +34,11 @@ load(
     _rust_clippy = "rust_clippy",
     _rust_clippy_aspect = "rust_clippy_aspect",
 )
+load(
+    "@io_bazel_rules_rust//rust:private/rust_analyzer.bzl",
```
UebelAndre, line 38: Can you use a workspace relative path here? `"//rust:private/rust_analyzer.bzl",`

tools/analyzer/BUILD
```diff
@@ -0,0 +1,10 @@
+load(
+    "@io_bazel_rules_rust//rust:rust.bzl",
```
UebelAndre, line 2: Can you use a workspace relative path here? `"//rust:rust.bzl",`

UebelAndre (changes requested):
Hey, I'm trying to get https://github.com/bazelbuild/rules_rust/pull/500 merged but don't want it to cause more work for you. So what I did instead was ask you to do more work by updating this PR 😅 I think they're good changes though. If you don't get to them here I'll be doing the same in https://github.com/bazelbuild/rules_rust/pull/500 

Do you think you'll have time to address the bullets you [mentioned](https://github.com/bazelbuild/rules_rust/pull/505#issuecomment-766244817) sometime this week?

tools/analyzer/main.rs
djmarcin: Done. I agree that `bazel info execution_root` is a better way to find this, but it's a bit unsatisfying because the command output is a `Vec<u8>` and there doesn't seem to be a cross-platform way to interpret that as an OsString (unix has an extension to do so, but windows doesn't seem to), so maybe it's better to go the `OsString` with `push` route? What do you think?

djmarcin:
Resolved most of the outstanding comments, still need to hit a few from @illicitonion

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,219 @@
...
+    dep_infos = [dep[RustAnalyzerInfo] for dep in ctx.rule.attr.deps if RustAnalyzerInfo in dep]
+    if hasattr(ctx.rule.attr, "proc_macro_deps"):
+        dep_infos += [dep[RustAnalyzerInfo] for dep in ctx.rule.attr.proc_macro_deps if RustAnalyzerInfo in dep]
+    transitive_deps = depset(direct = dep_infos, order = "postorder", transitive = [dep.transitive_deps for dep in dep_infos])
+
+    crate_info = target[CrateInfo]
+    return [RustAnalyzerInfo(
+        crate = crate_info,
+        cfgs = cfgs,
+        env = ctx.rule.attr.rustc_env if hasattr(ctx.rule.attr, "rustc_env") else {},
```
illicitonion, line 74:
```suggestion
        env = getattr(ctx.rule.attr, "rustc_env", {}),
```

illicitonion (commented):
Looks great to me! Thanks for following up!

UebelAndre:
@djmarcin Hey, can you take a moment to fix the conflicts?

@illicitonion @hlopko Can you guys take a look at this? I think it'd be best to get this merged and allow more people to use it and make incremental improvements to it.

djmarcin:
I merged up to latest HEAD, but it seems that something in the intervening changes (or maybe a bad merge?) busted this PR so that now it gets stuck in an infinite loop inside bazel and eventually OOMs. This happens after the aspect evaluation but before the rust_analyzer rule executes, so I'm not entirely sure how to debug it.

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,205 @@
...
+    crate["display_name"] = info.crate.name
+    crate["edition"] = info.crate.edition
+    crate["env"] = {}
+
+    # Switch on external/ to determine if crates are in the workspace or remote.
+    # TODO: Some folks may want to override this for vendored dependencies.
+    if info.crate.root.path.startswith("external/"):
+        crate["is_workspace_member"] = False
+        crate["root_module"] = _exec_root_tmpl + info.crate.root.path
+        crate_root = _exec_root_tmpl + info.crate.root.dirname + "/../"
```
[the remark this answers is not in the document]
djmarcin: I think this was just a mistake -- what it was trying to do was escape from the `src` directory where most `crate.root` files reside. Unfortunately I don't think there's a way to move up two directories in starlark (dirname returns a string, not a file, so you can't do `file.dirname.dirname`), but actually assuming that a `src` directory exists is probably just a bad idea anyway, and also we probably don't need to escape from it even if it does. So I think I can remove this.

illicitonion (approved):
Looks great - thanks!

djmarcin:
Thanks for merging, here are some quick notes on the latest "how to use" and future work.

How to use:
1) Add the following to your `WORKSPACE`
```
load("@rules_rust//tools/rust_analyzer/raze:crates.bzl", "rules_rust_tools_rust_analyzer_fetch_remote_crates")
rules_rust_tools_rust_analyzer_fetch_remote_crates()
```
2) Add a rule to your root `BUILD` file like the following. You can generate the interior targets list with something like: `bazel query 'kind("rust_library|rust_binary", //...:all)' | grep -v build_script | grep "//" | awk " { print \"'\" \$0 \"',\" }"`. Note that visibility rules apply.
```
rust_analyzer(
    name = "rust_analyzer",
    targets = [ /* all the binary/library targets you want in the rust-project.json */ ],
)
``` 
3) Run `bazel run @rules_rust//tools/rust_analyzer:gen_rust_project` whenever your dependencies change to regenerate `rust-project.json`. We .gitignore this file, because it's large and effectively just a build artifact. We use a VSCode task to run `gen_rust_project` automatically whenever VSCode starts, or manually whenever things get wonky. Once the rust-project.json file exists at your workspace root, rust-analyzer should automatically use it. We sometimes need to force a rust-analyzer restart to pick up changes.

Future work:
1) Since we're already running a separate binary for this, we should be able to autodetect the targets in the workspace without needing the root `rust_analyzer` rule. Eliminate the need for this rule.
2) Potentially, automatically include `rules_rust_tools_rust_analyzer_fetch_remote_crates()` when the user adds `rust_repositories`
3) Improve the documentation once 1 and maybe 2 are done.

Known issues:
- The --remote_download family of flags breaks proc_macro support because the proc_macro `.so` files are not downloaded. rust_analyzer needs these files present so it can load and run them. Solution: remove the --remote_download flags.
- rust-analyzer uses the source code that is downloaded and extracted during a build. If you run `bazel clean`, rust-analyzer will be unable to find the symbols. Solution: rerun `bazel build`

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,220 @@
...
+        crate = crate_info,
+        cfgs = cfgs,
+        env = getattr(ctx.rule.attr, "rustc_env", {}),
+        deps = dep_infos,
+        transitive_deps = transitive_deps,
+        proc_macro_dylib_path = find_proc_macro_dylib_path(toolchain, target),
+        build_info = build_info,
+    )]
+
+def find_proc_macro_dylib_path(toolchain, target):
```
hlopko, line 82: Nit: Why not leading underscore?

hlopko (commented):
I'm sorry, I didn't have time to do the thorough review :/

Rust analyzer is updated very often, do you plan to keep updating our checkout with every release? If not, maybe we should add a mechanism to choose an in-workspace version instead of the one bundled with the rules..

Could you check in the docs you mentioned in the last comment? That's already better than hunting for the PR to know how to use rust-analyzer.

I'll try to look into it more thoroughly soon. Thank you so much for working on this!

djmarcin:
> Rust analyzer is updated very often, do you plan to keep updating our checkout with every release? If not, maybe we should add a mechanism to choose an in-workspace version instead of the one bundled with the rules..

This PR does not bundle rust-analyzer, it merely generates a rust-project.json file, which appears to be a stable format.

rust/private/rust_analyzer.bzl
```diff
@@ -0,0 +1,220 @@
...
+
+"""
+Rust Analyzer Bazel rules.
+
+rust_analyzer will generate a rust-project.json file for the
+given targets. This file can be consumed by rust-analyzer as an alternative
+to Cargo.toml files.
+"""
+
+load("//rust/platform:triple_mappings.bzl", "system_to_dylib_ext", "triple_to_system")
```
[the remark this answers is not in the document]
djmarcin: Yes, I don't see any reason that would be a problem.

rust/private/rust_analyzer.bzl
djmarcin: I can update these to leading underscore in a follow-up.

Issue #462, opened 2020-10-22
Clean up bazel_version usage


dfreese:
The usage of bazel version predates bazel 1.0, so check to see if it is still necessary.    As brought up in #460, native.bazel_version may provide enough of a replacement.

The repository rule in workspace.bzl was causing conflicts with other bazel repositories that pull in a @bazel_version

wchargin:
Just getting back to this. Attempting to replace `BAZEL_VERSION` with
`native.bazel_version` does still fail:

<details>
<summary>Diff and shell session with “Error: no native function or rule 'bazel_version'”</summary>

```
wchargin@wchargin-xenial-42:~/rules_rust/examples$ bazel info release
release 3.7.0
wchargin@wchargin-xenial-42:~/rules_rust/examples$ git diff
diff --git a/rust/private/rustc.bzl b/rust/private/rustc.bzl
index 9968f84..2fd54fc 100644
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -230,12 +230,13 @@ def get_cc_toolchain(ctx):
         tuple: A tuple of (CcToolchain, FeatureConfiguration)
     """
     cc_toolchain = find_cpp_toolchain(ctx)
+    bzv = native.bazel_version()
 
     kwargs = {
         "ctx": ctx,
-    } if len(BAZEL_VERSION) == 0 or versions.is_at_least(
+    } if len(bzv) == 0 or versions.is_at_least(
         "0.25.0",
-        BAZEL_VERSION,
+        bzv,
     ) else {}
     feature_configuration = cc_common.configure_features(
         cc_toolchain = cc_toolchain,
wchargin@wchargin-xenial-42:~/rules_rust/examples$ bazel run //hello_world
ERROR: /home/wchargin/rules_rust/examples/hello_lib/BUILD:11:13: in rust_library rule //hello_lib:hello_lib: 
Traceback (most recent call last):
        File "/home/wchargin/.cache/bazel/_bazel_wchargin/43870155d296519f534e54a7f60aabf1/external/io_bazel_rules_rust/rust/private/rust.bzl", line 163, column 32, in _rust_library_impl
                return rustc_compile_action(
        File "/home/wchargin/.cache/bazel/_bazel_wchargin/43870155d296519f534e54a7f60aabf1/external/io_bazel_rules_rust/rust/private/rustc.bzl", line 600, column 59, in rustc_compile_action
[...]

PR #508, opened 2020-11-29
Remove uses of `BAZEL_VERSION` for detection of pre-1.0 Bazel.

jmillikin:
This PR removes:
* The special cases for pre-1.0 Bazel versions in `rustc.bzl`
* `BAZEL_VERSION` and `@ io_bazel_rules_rust_bazel_version`, which no longer have any references.
* `rust_workspace()` in `//:workspace.bzl`, which was obsolete due to no longer needing to define the Bazel version repository.

The remaining behavior of `rust_workspace()` was to define toolchains for Starlark unit testing, which are used by `rules_rust` itself but not necessary for downstream consumers. I moved that part to `//:WORKSPACE`.

For compatibility, I left the `rust_workspace()` symbol defined but empty (except for a warning message). It ought to be deleted in a future PR once users have removed it from their `WORKSPACE` files.

Fixes https://github.com/bazelbuild/rules_rust/issues/462


jmillikin:
Regarding discussion in the linked issue about whether it's OK to stop supporting Bazel 0.x, I believe that these versions were already blocked due to the following code in `//:workspace.bzl`:
```starlark
_MINIMUM_SUPPORTED_BAZEL_VERSION = "3.0.0"

   elif not versions.is_at_least(_MINIMUM_SUPPORTED_BAZEL_VERSION, bazel_version):
       fail("Bazel {} is too old to use with rules_rust, please use at least Bazel {}, preferably newer.".format(
           bazel_version,
           _MINIMUM_SUPPORTED_BAZEL_VERSION,
       ))
```

UebelAndre:
@damienmg would you be able to take a look here when you get a chance 🙏

damienmg (approved):

PR #509, opened 2020-11-29
Add triple mappings for `i386`, `armv7`, and `openbsd`.

jmillikin:
The `armv7` and `i386` CPU constraints were defined in https://github.com/bazelbuild/platforms/commit/6ad3cffd6797cd69fb2523402fb3b392867a2fa2 (in 2019).

I also noticed that `openbsd` was defined (https://github.com/bazelbuild/platforms/pull/8) so might as well put it in the list.


mfarrugi:
Can/should we depend on bazelbuild/platforms instead?

jmillikin:
> Can/should we depend on bazelbuild/platforms instead?

That dependency already exists. The file being edited is what maps Rust target triplets to `@platforms//` target names.

mfarrugi (approved):

Issue #91, opened 2018-06-01
Versioned dylib .so symlinks don't make it to runfiles


mfarrugi:
It looks like https://github.com/bazelbuild/rules_rust/blob/master/rust/rust.bzl#L342 is not quite right, and we need to fish out the .so.1.0 links from the binary.

@damienmg is there a straight forward way to getthe symlink paths? If not I will spend some time spelunking in whatever the cc_* rules currently expose.

damienmg:
Sorry, what is the issue exactly?

mfarrugi:
```
>> ldd bazel-bin/rust/examples/hello/hello_http 
        linux-vdso.so.1 =>  (0x00007ffc877b1000)
        libcrypto-opt.so.1.0.0 => not found
        libssl-opt.so.1.0.0 => not found
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9e98b0d000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f9e98905000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9e986e9000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9e984d3000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9e98146000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f9e99129000)
        
>> tree bazel-bin/_solib_k8/ 
bazel-bin/_solib_k8
└── _U@openssl_S_S_Clib___Uexternal_Sopenssl_Slib
   ├── libcrypto.so -> /home/marco/.cache/bazel/_bazel_marco/7e11b588968d46a3fb2589ae35c89c95/external/openssl/lib/libcrypto.so
   └── libssl.so -> /home/marco/.cache/bazel/_bazel_marco/7e11b588968d46a3fb2589ae35c89c95/external/openssl/lib/libssl.so
```

The execroot and runfiles tree are missing the versioned symlinks 'libcrypto-opt.so.1.0.0' and 'libssl-opt.so.1.0.0' that the binary is looking for. It looks like the cc_* rules don't create the symlinks in the runfiles tree if they are not depended on.

damienmg:
How are those symlink created in the first place, where is that example coming from?

[...]

PR #513, opened 2020-11-30
Fix linking against versioned shared library

djmarcin:
I was having trouble linking against `libruby.so` because at runtime the linker tries to find a versioned form like `libruby.so.2.7`. Versioned shared libraries were being filtered out of the depset because they don't end with `.so`. `rust_test` rules would then fail at runtime because although `libruby.so` existed in the sandboxed `_solib` directory, `libruby.so.2.7` did not.

starlark does not have regex support, so I cannot directly check for version numbers after the extension. Instead, it uses the assumption (also made elsewhere) that library names do not contain periods before the extension.


djmarcin:
I believe this will likely fix #91

illicitonion (commented):
Seems reasonable - any chance you can add/modify an example so that we have a test for this behaviour?

djmarcin:
I was trying to figure out the right way to do this. I'm not sure if I can
get bazel to create a shared library with the appropriate suffix, so I
might have to include a very small dummy shared object as a binary in order
to write a test, would that be reasonable or do you have a better idea?

dae:
Maybe you could use copy_file to rename it?

illicitonion:
If `copy_file` works, that would be grand, if not, introducing a <1k shared library (ideally for at least two platforms with differing extensions) doesn't feel like the end of the world :)

djmarcin:
I added a test, but I'm having trouble making it pass on non-linux platforms. The problem is that it appears that dynamic linking is not allowed on darwin:
```
(09:49:40) ERROR: /Users/buildkite/builds/bk-imacpro-3/bazel/rules-rust-rustlang/test/versioned_dylib/BUILD:9:12: in rust_binary rule //test/versioned_dylib:versioned_dylib:
  | File "/Users/buildkite/builds/bk-imacpro-3/bazel/rules-rust-rustlang/rust/private/rust.bzl", line 200, column 32, in _rust_binary_impl
  | return rustc_compile_action(
  | File "/Users/buildkite/builds/bk-imacpro-3/bazel/rules-rust-rustlang/rust/private/rustc.bzl", line 613, column 36, in rustc_compile_action
  | args, env = construct_arguments(
  | File "/Users/buildkite/builds/bk-imacpro-3/bazel/rules-rust-rustlang/rust/private/rustc.bzl", line 539, column 33, in construct_arguments
  | rpaths = _compute_rpaths(toolchain, output_dir, dep_info)
  | File "/Users/buildkite/builds/bk-imacpro-3/bazel/rules-rust-rustlang/rust/private/rustc.bzl", line 777, column 13, in _compute_rpaths
  | fail("Runtime linking is not supported on {}, but found {}".format(
  | Error in fail: Runtime linking is not supported on darwin, but found depset([<generated file _solib_darwin_x86_64/_U_S_Stest_Sversioned_Udylib_Sc_Clibreturn_Uzero___Utest_Sversioned_Udylib_Sc/libreturn_zero.2.dylib>, <generated file _solib_darwin_x86_64/_U_S_Stest_Sversioned_Udylib_Sc_Clibreturn_Uzero___Utest_Sversioned_Udylib_Sc/libreturn_zero.dylib>], order = "topological")
``` 

Additionally, there doesn't appear to be a way to have targets that only build on one platform until bazel 4.0 comes out. 3.7.0 has compatible_with on the cc_library which looks like what I want, but it can't be used with contraints:
```
in compatible_with attribute of cc_binary rule //test/versioned_dylib/c:libreturn_zero.so: constraint_value rule '@platforms//os:linux' is misplaced here (expected environment). Since this rule was c
reated by the macro 'cc_binary', the error might have been caused by the macro implementation
```

The last idea I had was to mark the rules manual and use `tags = select({...})` on the test to make it manual on non-linux and automatically run on linux. Unfortunately tags are not configurable:
```//test/versioned_dylib:versioned_dylib_test: attribute "tags" is not configurable

**Comment by illicitonion** [comment]
Sigh. Sorry about the wild chase! Let's make the test linux only, and add it as an excluded test in `.bazelci/presumbit.yml` for now - we can tidy things up when 4.0 happens and if/when the linux-only error you hit gets resolved.

**Comment by UebelAndre** [comment]
> Sigh. Sorry about the wild chase! Let's make the test linux only, and add it as an excluded test in `.bazelci/presumbit.yml` for now - we can tidy things up when 4.0 happens and if/when the linux-only error you hit gets resolved.

Can we please just add the test in a later PR? I'd rather not break the ability to `bazel test //...` for MacOS 🙏

**Comment by illicitonion** [comment]
If we tag it as `manual`, and explicitly include it in the linux version of `.bazelci/presumbit.yml`, we get the best of both worlds :)

File: rust/private/rustc.bzl [hunk 1]
```diff
@@ -185,7 +185,8 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
             transitive_dylibs.append(depset([
                 lib
                 for lib in libs.to_list()
-                if lib.basename.endswith(toolchain.dylib_ext)
+                # Dynamic libraries may have a version number nowhere, or before (linux) or after (macos) the extension.
```
illicitonion, line 188:
```suggestion
                # Dynamic libraries may have a version number nowhere, or before (macos) or after (linux) the extension.
```

illicitonion (approved):
Thanks!

rust/private/rustc.bzl
djmarcin: whoops, thanks :)

PR #526, opened 2020-12-05
Skip linking for clippy aspect for faster builds

djmarcin:
Skipping linking for clippy can dramatically accelerate results. Local testing in my repo shows a speedup anywhere from 30-70% with real time going from ~12-15s -> 4-8s for leaf targets. Larger targets will generally reap greater results from skipping linking.


dae:
This gets my vote; it looks like a more thorough version of #455

dfreese (approved):

PR #528, opened 2020-12-08
Perform - to _ replacement in test output names

illicitonion:
Right now if you have a rust_test/benchmark which has -s in the name,
you get an error that the output file wasn't created by the action.

This renames a `rust_test` in examples to hit the bug, and then fixes it.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -153,7 +153,7 @@ def _rust_library_impl(ctx):
     # Determine unique hash for this rlib
     output_hash = determine_output_hash(crate_root)
 
-    crate_name = ctx.label.name.replace("-", "_")
+    crate_name = name_to_crate_name(ctx.label.name)
     rust_lib_name = _determine_lib_name(
         crate_name,
         ctx.attr.crate_type,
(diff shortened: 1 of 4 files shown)


rust/private/rust.bzl
```diff
@@ -938,3 +938,6 @@ rust_benchmark(
 Run the benchmark test using: `bazel run //fibonacci:fibonacci_bench`.
 """,
 )
+
+def name_to_crate_name(name):
```
UebelAndre, line 942: Can you add a docstring here and give some context as to why it exists?

smklein (approved):

rust/private/rust.bzl
illicitonion: Done :)

rust/private/rust.bzl
```diff
@@ -938,3 +938,15 @@ rust_benchmark(
...
+    """Converts a build target's name into the name of its associated crate.
+
+    Crate names cannot contain certain characters, such as -, which are allowed
+    in build target names. All illegal characters will be converted to
+    underscores.
+
+    This is a similar conversion as that which cargo does, taking a
+    `Cargo.toml`'s `package.name` and canonicalizing it
+    """
```
UebelAndre, line 951: nit: Can you add `Arg` and `Return` sections with type hints 😅 (this one I won't fight anyone over but would still like to see it)

illicitonion: Done :)

PR #519, opened 2020-12-02
Added `proto_compile_deps` and `grpc_compile_deps` to `@io_bazel_rules_rust//proto:toolchain`

UebelAndre:
Closes #483

This PR adds `proto_compile_deps` and `grpc_compile_deps` to the `@io_bazel_rules_rust//proto:toolchain` toolchain to close a comment
```python
# TODO(damienmg): Once bazelbuild/bazel#6889 is fixed, reintroduce
# proto_compile_deps and grpc_compile_deps and remove them from the
# rust_proto_library and grpc_proto_library.
```

This is doable now that https://github.com/bazelbuild/bazel/issues/6889 is closed.

Though, I'm not entirely clear on what the earliest version of Bazel that supports this fix is. It looks like 3.4 as per https://github.com/bazelbuild/bazel/commit/bc4ef8c5ad7bbd274fe5481ca45d8d8a9da4f1d7 from https://github.com/bazelbuild/bazel/issues/10523

Unsure of what to do with these changes in that case.

PR #529, opened 2020-12-08
Updated minimum supported version from 3.0.0 to 3.4.0

UebelAndre:
This unblocks https://github.com/bazelbuild/rules_rust/pull/519 since it introduces the use of `incompatible_use_toolchain_transition` which was only added in `3.4.0` (https://github.com/bazelbuild/bazel/commit/099cf2f9c57936617e912cff73399bbf65f14e64)

Also, version bumps should be their own commits so they're more discoverable IMO.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -20,7 +20,7 @@ tasks:
       - "//test/versioned_dylib:versioned_dylib_test"
   ubuntu2004:
     name: "Minimum Supported Version"
-    bazel: "3.0.0"
+    bazel: "3.4.0"
     build_targets: *default_linux_targets
     test_targets: *default_linux_targets
   macos:


dfreese:
In general, I think the minimum version of 3.0.0 was too aggressive, and I'm looking at moving that backward, so I'm not comfortable with this at the moment.

UebelAndre:
Just calling this out as a requisite of another change. Though I personally think that moving to 4.0 as a min supported version is fine. I hate that there are no releases for this repo and it feels extremely nebulous as to what state the rules are in between different commits. Since 4.0 is a LTS release, it'd at least communicate at what point the rules will work, regardless of what revision they're at.

UebelAndre:
But yeah, not waiting on this change by any means so no pressure to merge.

UebelAndre:
@dfreese what was the change that required 3.0 again?

UebelAndre:
Closing this since I closed https://github.com/bazelbuild/rules_rust/pull/519

Issue #231, opened 2019-05-30
rust_benchmark: expected value of type 'File or NoneType' for parameter 'sibling', for call to method 


prestonvanloon:
```
        File "/home/preston/.cache/bazel/_bazel_preston/3c9e35e32d4309221d1ab3ca894b94ae/external/io_bazel_rules_rust/rust/private/rust.bzl", line 201, in _rust_benchmark_impl
                ctx.actions.declare_file("{}_bin".format(bench_script.bas...), ...)
expected value of type 'File or NoneType' for parameter 'sibling', for call to method declare_file(filename, sibling = None) of 'actions'
```

https://github.com/bazelbuild/rules_rust/blob/adccbbad58411657102ab8168aee7c5f4ff3af4c/rust/private/rust.bzl#L200-L204

prestonvanloon:
`ctx.bin_dir` is not a file, but is a root: https://docs.bazel.build/versions/master/skylark/lib/ctx.html#bin_dir

Is there an example of `rust_benchmark`? I don't think this rule works at all.

prestonvanloon:
Found the example. To reproduce this, run `bazel build //fibonacci:fibonacci_bench` from the examples directory.

mfarrugi:
The benchmark example is skipped for CI because we're running with stable. iirc you are running with nightly, which is a requirement for benchmarks to run at all (ie. the `bazel build` invocation implies a nightly toolchain). 

I would fully believe that either or both:
- The code rotted because it's not tested by CI
- The code requires a particular bazel version

prestonvanloon:
I think this is a starlark code bug, not nightly rust issue

```
Build label: 0.25.3
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu May 23 19:25:07 2019 (1558639507)
Build timestamp: 1558639507
Build timestamp as int: 1558639507
```

Running `bazel build` on a test benchmark actually runs the benchmark? 
Wouldn't it be more idiomatic to run the test on `bazel test`?

mfarrugi:
I'm saying that benchmarks require nightly, which is why the starlark code is untested and likely broken.

`bazel build` should build the benchmark + wrapper script, `bazel run` should run the benchmark. It was a test before, but having a benchmark do a bunch of not-actually-a-test work seemed inappropriate.

a1ph:
[...]

PR #532, opened 2020-12-12
Fix rust_benchmark rule

djmarcin:
I was just looking into fixing #231, but it doesn't seem like the current setup is well suited to criterion benchmarks (which I think has become the preferred method?). Specifically, when delegating to `_rust_test_common`, the criterion code doesn't seem to run at all because criterion wants to disable the test harness in favor of its own `main`.

Since this rule doesn't currently work, it's unlikely anyone depends on it. I propose we repurpose `rust_benchmark` to simply warn when the compilation mode != "opt" and delegate to a rust_binary with `--bench` which is more in line with how criterion expects to be invoked.
```
INFO: Build completed successfully, 60 total actions
Gnuplot not found, using plotters backend
fib 20                  time:   [36.696 us 37.153 us 37.602 us]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

This strategy can optionally be made to work for the nightly `#[bench]` attribute by passing the `--test` flag to rustc to generate the test harness. This is enabled with the `test_harness = True` option.

A test was added using criterion to prevent future bit-rot.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -68,6 +68,7 @@ tasks:
       - "-@examples//ffi/rust_calling_c/simple/..."
       - "-@examples//hello_sys/..."
       - "-@examples//complex_sys/..."
+      - "-@examples//criterion_bench/..."
       - "-@examples//proto/..."
       - "-@examples//wasm/..."
     build_targets: *windows_targets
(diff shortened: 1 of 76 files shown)


UebelAndre:
Hey, sorry about the long delay here. I'd be happy to take a look if you had the time to rebase 😄

UebelAndre:
The changes here were merged in https://github.com/bazelbuild/rules_rust/pull/923, thanks again!

Issue #536, opened 2020-12-16
Correct use of CARGO_CRATE_NAME vs CARGO_PKG_NAME


arlyon:
Currently the CARGO_PKG_NAME env var is passed in to the crate being built, but I am wondering if this is correct. The code https://github.com/bazelbuild/rules_rust/commit/3dffbabb3ab65a41056228b5c387d4b78331eaec#diff-4b0b39fe65c7318b74b94bd271d638a060ef06c2ac9363109b0ea4aa5ca085f2R36 implies that we are setting the crate name to CARGO_PKG_NAME, but there is also CARGO_CRATE_NAME which is used for the same thing.

> `CARGO_PKG_NAME` — The name of your package.
> `CARGO_CRATE_NAME` — The name of the crate that is currently being compiled.

It seems to me, at the very least, that both should be supported, but I am wondering what the specifics of CARGO_PKG_NAME are and if we are using it correctly.

PR #537, opened 2020-12-16
Add the CARGO_CRATE_NAME env var

arlyon:
See #536 for more discussion.


cargo/cargo_build_script.bzl
```diff
@@ -69,6 +69,7 @@ def _build_script_impl(ctx):
     env.update({
         "CARGO_MANIFEST_DIR": manifest_dir,
         "CARGO_PKG_NAME": crate_name,
+        "CARGO_CRATE_NAME": crate_name,
```
illicitonion, line 72: Cargo doesn't actually set these two env vars to exactly the same thing; `CARGO_CRATE_NAME` has `-` replaced with `_`, but `CARGO_PKG_NAME` doesn't.

Probably if `ctx.attr.crate_name` is set, we should use that here, but if it's not, we should use the value _before_ line 55 is run, rather than after.

illicitonion (commented):
Thanks for the contribution! I agree that we should set this env var, but I think we should maybe set it to a slightly different value - let me know what you think :)

To see cargo's behaviour:

```console
$ cargo init crate-with-dashes
$ cd crate-with-dashes
$ cat >src/main.rs <<EOF
fn main() {
    println!("CARGO_CRATE_NAME: {}", env!("CARGO_CRATE_NAME"));
    println!("CARGO_PKG_NAME: {}", env!("CARGO_PKG_NAME"));
}
EOF
$ cargo run --quiet
CARGO_CRATE_NAME: crate_with_dashes
CARGO_PKG_NAME: crate-with-dashes
```

cargo/cargo_build_script.bzl
arlyon: Well I suppose in the face of lacking information online, testing yourself is probably best! :) thanks, I've pushed some changes with similar stuff added to rustc so the env _should_ be propagated.

illicitonion:
Sorry for the delay!

Unfortunately this PR now goes the wrong way around: `CARGO_PKG_NAME` should always be set based on the `name` attribute, and `crate_name` should replace `-`s with `_s`. The `CARGO_PKG_NAME` may contain `_`s, whereas right now this PR replaces all `_`s with `-`.

hlopko:
Friendly ping from a bywalker :)

arlyon:
Thanks for the ping, this fell of my todo list :) I will push changes asap

UebelAndre:
Friendly drive-by ping, any updates here? 😄

arlyon:
We're not far off, but if the crate name in rules_rust is in fact the pkg_name then that differenciation should be reflected in the properties in the attributes and we should rename it. I'm not sure about the implications of this though so some feedback would be helpful.

illicitonion:
> We're not far off, but if the crate name in rules_rust is in fact the pkg_name then that differenciation should be reflected in the properties in the attributes and we should rename it. I'm not sure about the implications of this though so some feedback would be helpful.

That's a really good point! And I _think_ we can answer it with my favourite answer of all: Let's delete the attribute!

As far as I can tell, we currently use the `crate_name` attribute in exactly three places:
1. We use it to set `CARGO_PKG_NAME`
2. We uppercase it and replace -s with _s to set `CARGO_FEATURE_${CRATE_NAME}=1` env vars per feature
3. We use it for some legacy fallback behaviour to guess what the `links` attribute should've been before we had the `links` attribute - https://github.com/google/cargo-raze/pull/400 will allow us to delete the fallback behaviour.

(Worthy of note, we _don't_ use it for the thing it was originally added for in https://github.com/bazelbuild/rules_rust/pull/341, which was to set `DEP_` env vars, they now use the `links` attribute, except where that fallback is used).

So I'd be tempted to delete the `crate_name` attribute entirely, and just treat the `name` attribute like the thing we'd see in a `package.name` of a Cargo.toml. If we did that, we'd (after removing a trailing `_build_script` if present) set `CARGO_PKG_NAME` to exactly `ctx.attr.name`, set `CARGO_CRATE_NAME` to the -s to _s version, set `CARGO_FEATURE_${CRATE_NAME}` using the uppercased version of `CARGO_CRATE_NAME`, and have removed an unnecessary attribute on the way.

hlopko:
SGTM, I'm a big fan of deleting and simplifying code, let's do this!

arlyon:
Closing after a more comprehensive solution here https://github.com/bazelbuild/rules_rust/pull/643 thanks all!

PR #541, opened 2020-12-25
Prefer action_env to set rust_repo when available

cecini:
We set action_env for RUST_VERSION,RUSTFMT_VERSION,RUST_ISO_DATE, RUST_EDITION,RUST_DEVCOMPS in the workspace or bazel cmd line
This pr multiple rust reposities set.  we only call rust_reposities() in the corrosponding repo, no need to sync the rust_reposities() arugment value when update the argument in one repo.


cecini:
@googlebot I signed it!

rust/repositories.bzl
```diff
@@ -528,68 +533,80 @@ def _load_rust_stdlib(ctx, target_triple):
...
 
 def _rust_toolchain_repository_impl(ctx):
     """The implementation of the rust toolchain repository rule."""
 
-    _check_version_valid(ctx.attr.version, ctx.attr.iso_date)
+    version = ctx.os.environ.get("RUST_VERSION") if ctx.os.environ.get("RUST_VERSION") else ctx.attr.version 
+    rustfmt_version = ctx.os.environ.get("RUSTFMT_VERSION") if ctx.os.environ.get("RUSTFMT_VERSION") else ctx.attr.rustfmt_version 
+    iso_date = ctx.os.environ.get("RUST_ISO_DATE") if ctx.os.environ.get("RUST_ISO_DATE") else ctx.attr.iso_date
+    edition = ctx.os.environ.get("RUST_EDITION") if ctx.os.environ.get("RUST_EDITION") else ctx.attr.edition
+    dev_components = True if ctx.os.environ.get("RUST_DEVCOMPS") else ctx.attr.dev_components
```
illicitonion, line 589:
```suggestion
    version = ctx.os.environ.get("RUST_VERSION", ctx.attr.version)
    rustfmt_version = ctx.os.environ.get("RUSTFMT_VERSION", ctx.attr.rustfmt_version)
    iso_date = ctx.os.environ.get("RUST_ISO_DATE", ctx.attr.iso_date)
    edition = ctx.os.environ.get("RUST_EDITION", ctx.attr.edition)
    dev_components = ctx.os.environ.get("RUST_DEVCOMPS") or ctx.attr.dev_components
```

rust/repositories.bzl
```diff
@@ -465,51 +465,56 @@ def _load_rustfmt(ctx):
...
 
-def _load_rust_stdlib(ctx, target_triple):
+def _load_rust_stdlib(ctx, target_triple, iso_date, version, edition):
     """Loads a rust standard library and yields corresponding BUILD for it
 
     Args:
         ctx: A repository_ctx.
         target_triple: The rust-style target triple of the tool
+        version: The rust version
+        iso_date: The rust iso_date
```
illicitonion, line 505:
```suggestion
        iso_date: The rust iso_date
        version: The rust version
```

illicitonion (commented):
I'm not sure about whether this is a change the rules want to make (others will hopefully weigh in), but if so, I'd suggest these modifications :)

UebelAndre:
Would it also make sense to update some documentation that would get rendered in `./docs` describing this new behavior? I think it make sense but also doesn't support having multiple toolchains with different versions. I don't know if that's currently an expected use-case or not but still wanted to call that out.

rust/repositories.bzl
```diff
@@ -465,51 +465,56 @@ def _load_rustfmt(ctx):
...
 
     return BUILD_for_rustfmt(target_triple)
 
-def _load_rust_compiler(ctx):
+def _load_rust_compiler(ctx, version, iso_date):
     """Loads a rust compiler and yields corresponding BUILD for it
 
     Args:
         ctx: A repository_ctx.
+        version: The rust version
```
hlopko, line 478: Nit: please append `for example` with relevant example to version and iso_date here and in other docs as well

hlopko:
@illicitonion it seems you have concerns about this PR? Could you share? Looks reasonable to me, but maybe I'm missing something?

illicitonion:
> @illicitonion it seems you have concerns about this PR? Could you share? Looks reasonable to me, but maybe I'm missing something?

Mostly just that I don't know of any other rulesets which allow environment variables to override this kind of config in repository rules, and that environment variables are somewhat error-prone and hard to document. If someone is using an older version of `rules_rust`, or if we drop support for these overrides at some point, or they make a typo in the env var name, the user won't be given any feedback that their env var is being ignored. Given that there _will_ be a value used, just maybe not the one they expected (because the WORKSPACE file will have some value it uses instead), they may not notice, and this can be a frustrating experience to debug (once you finally notice there's a problem).

It's not a particularly strong objection (hence not taking much of a stance in my original comment), but it makes me pause.

If we do do this, I believe we should also be adding the env vars we look at to the repository rule's `environ` attribute, otherwise it may not be re-run: https://docs.bazel.build/versions/master/skylark/repository_rules.html#when-is-the-implementation-function-executed

dfreese:
> Mostly just that I don't know of any other rulesets which allow environment variables to override this kind of config in repository rules, and that environment variables are somewhat error-prone and hard to document. If someone is using an older version of `rules_rust`, or if we drop support for these overrides at some point, or they make a typo in the env var name, the user won't be given any feedback that their env var is being ignored. Given that there _will_ be a value used, just maybe not the one they expected (because the WORKSPACE file will have some value it uses instead), they may not notice, and this can be a frustrating experience to debug (once you finally notice there's a problem).

I would be nice to use something a little more bazel-like for this, such as [user-defined build settings](https://docs.bazel.build/versions/master/skylark/config.html#user-defined-build-settings), however, this being a repository rule does seem to make that more difficult.

My concern would be that these are a global override on all calls to rust_repository_set, so if you had specified one on 1.49.0 for linux, and then one on nightly for bare-metal work, then this would override both, which might be surprising.

hlopko:
I agree with @dfreese, affecting all toolchains might cause confusion/unintentional incorrect usage. Would it still be useful for you @cecini if `rust_toolchain_repository` had an argument that enabled looking into env to override things, but currently registered toolchains would not set it by default?

UebelAndre:
Hello, just pinging this issue, any updates here?

hlopko:
Friendly ping. Please let us know if you no longer have capacity to pursue this change, we'll try to find a new owner.

PR #542, opened 2020-12-27
Updated cargo-raze outputs

UebelAndre:
Regenerated cargo-raze outputs using version `0.8.0`. A clearer impact of the changes can be seen in 2b2f87cecd64b8c87de8afee6e7f7f5b698b88c8

This change is intended to reduce the footprint of https://github.com/bazelbuild/rules_rust/pull/500


UebelAndre:
It's worth noting that I had to regenerate the `wasm-bindgen` lockfile and set `gen_buildrs = true` for `libc`. I ran into a similar issue documented here: https://github.com/rust-lang/socket2/issues/130

wasm_bindgen/raze/Cargo.toml
```diff
@@ -50,36 +46,30 @@ additional_flags = [
...
-
-[raze.crates.proc-macro2.'1.0.24']
+[package.metadata.raze.crates.proc-macro2.'1.0']
 gen_buildrs = true
 additional_flags = [
     "--cfg=use_proc_macro",
 ]
 
-[raze.crates.syn.'1.0.40']
+[package.metadata.raze.crates.libc.'0.2']
```
UebelAndre, line 55: relevant comment: https://github.com/bazelbuild/rules_rust/pull/542#issuecomment-751433441

bindgen/raze/Cargo.toml
```diff
@@ -9,23 +9,25 @@ path = "fake_rules_rust_bindgen_lib.rs"
...
 
-[raze]
+[package.metadata.raze]
 genmode = "Remote"
 workspace_path = "//bindgen/raze"
 gen_workspace_prefix = "rules_rust_bindgen"
-output_buildfile_suffix = "BUILD"
+rust_rules_workspace_name = "io_bazel_rules_rust"
+package_aliases_dir = "."
+default_gen_buildrs = false
```
illicitonion, line 18: I suspect these can be removed, and most of the per-package overrides removed, across this entire PR, but that can definitely be a follow-up :)

illicitonion (approved):
Thanks!

bindgen/raze/Cargo.toml
UebelAndre: Can you elaborate on what you mean here? What should get removed?

illicitonion: We can probably leave `default_gen_buildrs` to its default of `true`, and stop worrying about it

UebelAndre: Ah, I see, yeah, that should be an explicit decision and didn't want to expand this PR with that change. I think a separate PR is the right thing to do 😄

PR #543, opened 2020-12-27
Use relative labels and `Label` macro wherever possible

UebelAndre:
This PR updates changes the explicit use of `@io_bazel_rules_rust//...` to `//...` and adds the use of [Label](https://docs.bazel.build/versions/master/skylark/lib/Label.html) wherever possible. This should currently have no functional but it provides some better type guarantees.

### DIFF
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 # buildifier: disable=module-docstring
-load("@io_bazel_rules_rust//rust:private/utils.bzl", "find_toolchain", "get_libs_for_static_executable")
-load("@io_bazel_rules_rust//rust:rust.bzl", "rust_library")
+load("//rust:private/utils.bzl", "find_toolchain", "get_libs_for_static_executable")
+load("//rust:rust.bzl", "rust_library")
 
 def rust_bindgen_library(
         name,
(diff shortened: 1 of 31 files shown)


UebelAndre:
This PR came from some exploration i was doing in trying to support either agnostic workspace names for this repository or at least both `rules_rust` and `io_bazel_rules_rust`. For the most part, this would work if it weren't for two things, the way toolchains are currently setup, and the fact that some `cargo-raze` outputs (namely protobuf) have dependencies on the workspace name.

bindgen/bindgen.bzl
```diff
@@ -168,16 +168,16 @@ rust_bindgen = rule(
...
             executable = True,
             allow_single_file = True,
             cfg = "exec",
         ),
     },
     outputs = {"out": "%{name}.rs"},
     toolchains = [
-        "@io_bazel_rules_rust//bindgen:bindgen_toolchain",
-        "@io_bazel_rules_rust//rust:toolchain",
+        str(Label("//bindgen:bindgen_toolchain")),
```
UebelAndre, line 179: I'm shopping around for some thoughts here and generally more insight on [Label](https://docs.bazel.build/versions/master/skylark/lib/Label.html). I'm not sure if this is an appropriate use of `Label`. Anyone have any experiences like this they can share?

UebelAndre:
@dfreese Hey, are you able to take a look at this 🙏

UebelAndre:
@damienmg hey, you wouldn't als have time to look at this would you 😅 🙏

UebelAndre:
@illicitonion Hey, would *you* have time to take a look at this? 🙏 😅

UebelAndre:
@damienmg Hey, spoke with @illicitonion and he thinks you'd be the best reviewer for this change. Would you have a quick moment to take a look at this this week?

bindgen/bindgen.bzl
damienmg: This is perfectly acceptable. I remember having a funny discussion with several senior engineer in the Bazel team about this "hack" and having the manager answer: I don't think this is a hack, which just settled our decision to keep this as an idiom, although I left the team 3 years ago so I don't know what is the status now.

damienmg (approved):

damienmg:
This is the right approach, this repository was created before we made the label behavior consistent and thus was using aboslute label approach, but there is no reason to do this anymore.

UebelAndre:
@damienmg Thanks for the review, that bit of context, and updating the branch 🙏  CI seems to still be passing 😄

PR #545, opened 2020-12-30
Fixed issue where repository_ctx for `load_arbitrary_tool` required `sha256s` attribute.

UebelAndre:
Any repository rule that attempts to use this macro is required to have have a `"sha256s": attr.string_dict()` attribute. This should not be required (and is inconsistent with the current documentation of the macro). This PR fixes that.


illicitonion (approved):
Thanks!

PR #548, opened 2021-01-06
cargo_build_script now only prints warnings to stderr.

UebelAndre:
## Overview
This prevents noisy output from `cargo_build_script` from being printed to the terminal, instead two log files are generated (`stdout`, `stderr`) which contain the raw output of the underlying `build.rs` script

## Details 

Currently users see something similar to the following output when they use `cargo_build_script`:

(note, uplaoded as a file because there's a lot of output)
[big_output.log](https://github.com/bazelbuild/rules_rust/files/5777780/big_output.log)

In general, this masks important information in a sea of arbitrary log lines. The changes in this PR update the output to:
```output
INFO: From CargoBuildScriptRun external/cargo_raze__libssh2_sys__0_2_20/libssh2_sys_build_script.out_dir:
Build Script Warning: libssh2/src/bcrypt_pbkdf.c:84:46: warning: expression does not compute the number of elements in this array; element type is 'uint32_t' (aka 'unsigned int'), not 'uint64_t' (aka 'unsigned long long') [-Wsizeof-array-div]
Build Script Warning:         blf_enc(&state, cdata, sizeof(cdata) / sizeof(uint64_t));
Build Script Warning:                                       ~~~~~  ^
Build Script Warning: libssh2/src/bcrypt_pbkdf.c:65:14: note: array 'cdata' declared here
Build Script Warning:     uint32_t cdata[BCRYPT_BLOCKS];
Build Script Warning:              ^
Build Script Warning: libssh2/src/bcrypt_pbkdf.c:84:46: note: place parentheses around the 'sizeof(uint64_t)' expression to silence this warning
Build Script Warning:         blf_enc(&state, cdata, sizeof(cdata) / sizeof(uint64_t));
Build Script Warning:                                              ^
Build Script Warning: 1 warning generated.
```

And log files are generated for the raw stderr and stdout from the build script, ie:

```output
bazel-out/darwin-fastbuild/bin/external/cargo_raze__libssh2_sys__0_2_20/libssh2_sys_build_script.stderr.log
bazel-out/darwin-fastbuild/bin/external/cargo_raze__libssh2_sys__0_2_20/libssh2_sys_build_script.stdout.log
```

(note, uplaoded as a file because there's a lot of output)
[libssh2_sys_build_script.stderr.log](https://github.com/bazelbuild/rules_rust/files/5777782/libssh2_sys_build_script.stderr.log)

Additionally, the `deps` attribute of `cargo_build_script` now only accepts targets that return the `DepInfo` provider. All other dependencies should be passed via `data`.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -191,7 +191,7 @@ _build_script_run = rule(
         # The source of truth will be the `cargo_build_script` macro until stardoc
         # implements documentation inheritence. See https://github.com/bazelbuild/stardoc/issues/27
         "script": attr.label(
-            doc = "The binary script to run, generally a rust_binary target.",
+            doc = "The binary script to run, generally a `rust_binary` target.",
             executable = True,
             allow_files = True,
             mandatory = True,
@@ -204,7 +204,8 @@ _build_script_run = rule(
             doc = "The name of the native library this crate links against.",
         ),
         "deps": attr.label_list(
-            doc = "The dependencies of the crate defined by `crate_name`",
+            doc = "The Rust dependencies of the crate defined by `crate_name`",
+            providers = [DepInfo],
         ),
         "version": attr.string(
             doc = "The semantic version (semver) of the crate",
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -45,6 +45,11 @@ def _build_script_impl(ctx):
     manifest_dir = "%s.runfiles/%s/%s" % (script.path, ctx.label.workspace_name or ctx.workspace_name, ctx.label.package)
     compilation_mode_opt_level = get_compilation_mode_opts(ctx, toolchain).opt_level
 
+    streams = struct(
+        stdout = ctx.actions.declare_file(ctx.label.name + ".stdout.log"),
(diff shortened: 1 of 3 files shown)


cargo/cargo_build_script.bzl
```diff
@@ -187,7 +204,8 @@ _build_script_run = rule(
             doc = "The name of the native library this crate links against.",
         ),
         "deps": attr.label_list(
-            doc = "The dependencies of the crate defined by `crate_name`",
+            doc = "The Rust dependencies of the crate defined by `crate_name`",
+            providers = [DepInfo],
```
UebelAndre, line 208: After doing a bit more tinkering, I also think this change should be added. The only targets passed to `deps` that end up being used are ones with `DepInfo` so I thought by adding this requirement, it'd be clearer to users that they should be using `data` if they want to add dependencies here.

UebelAndre:
@damienmg @illicitonion hey, do either of you have bandwidth for a review here? 🙏

cargo/cargo_build_script.bzl
damienmg: Good thinking!

damienmg (approved):

Issue #550, opened 2021-01-15
allow for project specific rust file server mirrors


dirkboye:
We mirror all external dependencies to our local network so that we do not need any internet access for building our software.
Using environment variable `STATIC_RUST_URL` is good for our CI but it's cumbersome for developers who need
to have the environment variable set as well.

A better solution can be seen in `go_rules` method `go_download_sdk` ( see https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#go-download-sdk ) where you can set the mirror urls.

I'd like to propose to add something similar to `rust_toolchain_repository` which would make it possible to
have project-specific rust mirrors.

dirkboye:
PR was merged!

UebelAndre:
Oh, one other thing I *just* came accross this morning. You might have been able to add a `.bazelrc` file to your projects and specify a [repo_env](https://docs.bazel.build/versions/master/command-line-reference.html#flag--repo_env). Would that have also worked for you?

dirkboye:
@UebelAndre unfortunately that doesn't work so well and increases build times for our CI as we use `bazel query` in our scripts.

`bazel query` does not support `--repo_env` or `--action_env`

running `bazel build` with repo_env will then result in

```
INFO: Build option --repo_env has changed, discarding analysis cache.
```

UebelAndre:
Oh wow, really good to know. That seems like a bug on Bazel's side though 😞 I otherwise don't understand why `--repo_env` shouldn't be supported by all commands

dirkboye:
@UebelAndre here's an open issue on the bazel side for that:
https://github.com/bazelbuild/bazel/issues/11943

UebelAndre:
@dirkboye Oh wow, you're a hero! Thank you 🙏 😄

PR #551, opened 2021-01-15
#550 rust_toolchain_repository: add urls to enable project specific rust file server mirrors

dirkboye:
environment variable `STATIC_RUST_URL` will work as before.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -21,7 +21,8 @@ def rust_repositories(
         rustfmt_version = None,
         edition = None,
         dev_components = False,
-        sha256s = None):
+        sha256s = None,
+        urls = ["https://static.rust-lang.org/dist/{}.tar.gz"]):
     """Emits a default set of toolchains for Linux, MacOS, and Freebsd
 
     Skip this macro and call the `rust_repository_set` macros directly if you need a compiler for \
@@ -47,6 +48,7 @@ def rust_repositories(
         edition (str, optional): The rust edition to be used by default (2015 (default) or 2018)
         dev_components (bool, optional): Whether to download the rustc-dev components (defaults to False). Requires version to be "nightly".
         sha256s (str, optional): A dict associating tool subdirectories to sha256 hashes. Defaults to None.
+        urls (list, optional): A list of mirror urls containing the tools from the Rust-lang static file server. These must contain the '{}' used to substitute the tool being fetched (using .format). Defaults to ['https://static.rust-lang.org/dist/{}.tar.gz']
     """
 
     if dev_components and version != "nightly":
@@ -84,6 +86,7 @@ def rust_repositories(
         edition = edition,
         dev_components = dev_components,
         sha256s = sha256s,
+        urls = urls,
     )
 
     rust_repository_set(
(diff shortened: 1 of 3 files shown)


dirkboye:
#550

rust/repositories.bzl
```diff
@@ -409,7 +417,7 @@ def load_arbitrary_tool(ctx, tool_name, tool_subdirectories, version, iso_date,
 
     This function sources the tool from the Rust-lang static file server. The index is available
     at: https://static.rust-lang.org/dist/index.html (or the path specified by
-    "${RUST_STATIC_URL}/dist/index.html" if the RUST_STATIC_URL envinronment variable is set).
+    "${STATIC_RUST_URL}/dist/index.html" if the STATIC_RUST_URL envinronment variable is set).
```
UebelAndre, line 420: I'm not sure it's the best idea to change the environment variable name. Whats the justification here?

rust/repositories.bzl
```diff
@@ -656,6 +669,10 @@ rust_toolchain_repository = repository_rule(
         "sha256s": attr.string_dict(
             doc = "A dict associating tool subdirectories to sha256 hashes. See [rust_repositories](#rust_repositories) for more details.",
         ),
+        "urls": attr.string_list(
+            doc = "A list of mirror urls containing the tools from the Rust-lang static file server. These must contain the '{}' used to substitute the tool being fetched (using .format). It defaults to the official repository 'https://static.rust-lang.org/dist/{}.tar.gz'",
```
UebelAndre, line 673: Nit, I think users will see the default in the rendered docs

UebelAndre (changes requested):
Two comments but otherwise looking good to me (as a non maintainer 😅)

rust/repositories.bzl
dirkboye: i think only the docs were wrong as some lines below the variable "STATIC_RUST_URL" is imported from environment.

rust/repositories.bzl
dirkboye: should i remove the default part from the doc?

rust/repositories.bzl
UebelAndre: Ah, good catch!

rust/repositories.bzl
UebelAndre: I would, I know there are a bunch of similar docs for macro parameters, but I think it'd be better to keep them out of attribute `doc`s.

dirkboye: done in fixup!

UebelAndre (approved):

UebelAndre:
Oh, one last thing, can you run `./docs/update_docs.sh` and push the results?

dirkboye:
pushed the re-generated docs!

thanks for the fast review!

UebelAndre:
> pushed the re-generated docs!
> 
> thanks for the fast review!

No problem 😄 Again, I'm not a maintainer so I can't merge it but I like the changes and think they'd be a great addition. Hopefully another maintainer can take a quick look today. @damienmg @illicitonion @dfreese ?

damienmg (approved):

PR #554, opened 2021-01-20
Fixed regression in `load_arbitrary_tool` causing it to require attrs

UebelAndre:
In https://github.com/bazelbuild/rules_rust/pull/545 I fixed an issue where `load_arbitrary_tool` could not be used in repository rules that did not specify particular attributes. This issue was then reintroduced in https://github.com/bazelbuild/rules_rust/issues/551. This PR fixes the issue again and adds a test to prevent this from happening again in the future.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,6 +20,10 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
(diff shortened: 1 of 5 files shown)


UebelAndre:
@damienmg Hey, do you have a quick moment to look at this PR? I feel like it's pretty straight forward and would be a big help 🙏 😄

illicitonion (approved):
LGTM - thanks!

PR #556, opened 2021-01-21
Allow rustc env vars to be set in files

illicitonion:
Build scripts are executed in the exec configuration, not the target
configuration, which means it's currently impossible to use values which
were generated as a result of the `--stamp` flag when creating a
`rust_binary`. This means that embedding things like build timestamps,
git shas, and dynamic version numbers, is impossible.

This change allows you to depend, in the target configuration (because
no transition occurs on the attribute) on the output of a genrule which
consumes data from the stable and volatile workspace status commands.


examples/hello_env/BUILD
```diff
@@ -0,0 +1,26 @@
...
+    "rust_binary",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "hello_env",
+    srcs = ["src/main.rs"],
+    deps = ["@examples//hello_lib"],
+    rustc_env_files = [":generate_rustc_env_file"],
```
hlopko, line 12: What do you think about creating a `tests` directory and moving `hello_env` there? I'd expect `examples` directory to serve as a documentation, inspiration, for the user, and `hello_env` might be more an integration test than an example.

For an actual example it would be cool to see how you typically use `--stamp`.

Orthogonally, would you be fine with naming the directory `rustc_env_files` or something more precise?

examples/hello_env/BUILD
```diff
@@ -0,0 +1,26 @@
+load(
+    "@io_bazel_rules_rust//rust:rust.bzl",
```
hlopko, line 2: I guess we should migrate to `@rules_rust`, but separate PR (and not necessarily you :) sounds good.

examples/hello_env/output_test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -eu
+
```
hlopko, line 2: Nit: add pipefail?

rust/private/rustc.bzl
```diff
@@ -371,7 +371,12 @@ def collect_inputs(
             linker_depset,
         ],
     )
-    return _process_build_scripts(ctx, file, crate_info, build_info, dep_info, compile_inputs)
+    build_env_files = getattr(files, "rustc_env_files", [])
+    compile_inputs, out_dir, build_env_file, build_flags_files = _process_build_scripts(ctx, file, crate_info, build_info, dep_info, compile_inputs)
+    if build_env_file:
+        build_env_files = [f for f in build_env_files] + [build_env_file]
```
hlopko, line 377: Just curious, I guess `[f for f in build_env_files]` is there because `build_env_files` is not a typical list and it doesn't understand `+`, so that's how you create a regular list?

hlopko:
Oh we actually already have a `test` directory..

examples/hello_env/BUILD
illicitonion: Ooh, TIL there's a `test` directory! Moved to `test` and renamed, but also added an example :)

examples/hello_env/output_test.sh
illicitonion: Done

rust/private/rustc.bzl
illicitonion: It's not a magic list type - when it get passed to `_process_build_scripts` the list got frozen, so is no longer modifiable, so we do the copy here to work around the fact that `build_env_files` is frozen :)

hlopko (approved):
LGTM, thank you!

Issue #226, opened 2019-05-16
bazel build //... fails on OSX


johnedmonds:
Seems to be because xcode removed libstdc++ (and replaced it with libc++) but we still depend on it.  Here's the full error message:

```
bazel build //...
INFO: Analysed 48 targets (110 packages loaded, 1680 targets configured).
INFO: Found 48 targets...
ERROR: /Users/jedmonds/rules_rust/tools/runfiles/BUILD:14:1: error executing shell command: '/bin/bash -c CARGO_MANIFEST_DIR=$(pwd)/tools/runfiles external/rust_darwin_x86_64/bin/rustc "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd  tools/runfiles/runfiles.rs --crate-name=runfiles_...' failed (Exit 1) bash failed: error executing command /bin/bash -c 'CARGO_MANIFEST_DIR=$(pwd)/tools/runfiles external/rust_darwin_x86_64/bin/rustc "$@" --remap-path-prefix="$(pwd)"=__bazel_redacted_pwd' '' tools/runfiles/runfiles.rs ... (remaining 14 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
error: linking with `external/local_config_cc/cc_wrapper.sh` failed: exit code: 1
  |
[...]

PR #558, opened 2021-01-25
Fix test target on macOS with Bazel@HEAD

meteorcloudy:
Avoid build artifacts name conflict after `supports_dynamic_linker` feature was removed from cc toolchain for macOS
Context: https://github.com/bazelbuild/bazel/issues/4341#issuecomment-762576370

### DIFF
--- a/examples/ffi/rust_calling_c/c/BUILD
+++ b/examples/ffi/rust_calling_c/c/BUILD
@@ -20,9 +20,9 @@ cc_test(
 
 ## Do the same as above, but with a dynamic c library.
 
-cc_library(
+cc_import(
     name = "native_matrix_so",
-    srcs = [":libnative_matrix_so.so"],
+    shared_library = ":libnative_matrix_so.so",
     hdrs = ["matrix.h"],
 )
 
--- a/examples/ffi/rust_calling_c/c/BUILD
+++ b/examples/ffi/rust_calling_c/c/BUILD
@@ -1,4 +1,4 @@
-load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_import", "cc_library", "cc_test")
 
 package(default_visibility = ["//ffi/rust_calling_c:__subpackages__"])
 


examples/ffi/rust_calling_c/c/BUILD
```diff
@@ -20,9 +20,9 @@ cc_test(
 
 ## Do the same as above, but with a dynamic c library.
 
-cc_library(
+cc_import(
```
UebelAndre, line 23: Can you also update the load statement to include `cc_import`?

UebelAndre (changes requested):
You will also need to remove the section from `.bazelrc/presubmit.yaml` disabling these tests on MacOS so we can confirm the changes actually fixed the issue:
https://github.com/bazelbuild/rules_rust/blob/215a8decfb06525a3f13b23fac5b3124eedabd27/.bazelci/presubmit.yml#L32-L34

When I build locally on your branch, I run into the following error
```output
ERROR: /private/var/tmp/_bazel_user/4affd8360c36081ca80c76cb4d709ed4/external/local_libstdcpp/BUILD.bazel:3:11: SolibSymlink _solib_darwin_x86_64/_U@local_Ulibstdcpp_S_S_Clibstdc++___U/libstdc++.6.dylib failed: missing input file 'external/local_libstdcpp/libstdc++.6.dylib', owner: '@local_libstdcpp//:libstdc++.6.dylib'
ERROR: /private/var/tmp/_bazel_user/4affd8360c36081ca80c76cb4d709ed4/external/local_libstdcpp/BUILD.bazel:3:11 SolibSymlink _solib_darwin_x86_64/_U@local_Ulibstdcpp_S_S_Clibstdc++___U/libstdc++.6.dylib failed: 1 input file(s) do not exist
```

I'm wondering if CI will show the same.

But super thanks for attempting to tackle this issue! I'd love to see all tests working 😄

UebelAndre:
Also wondering if https://github.com/bazelbuild/rules_rust/issues/226 is related to this

meteorcloudy:
https://github.com/bazelbuild/rules_rust/issues/226 seems to be a different issue as this change was only made recently: https://github.com/bazelbuild/bazel/commit/ec5553352f2f661d39ac4cf665dd9b3c779e614c

meteorcloudy:
@UebelAndre I think those tests are disabled for other issues this PR isn't aimed to fix. 
Even with those tests disabled, `bazel build` step still failed on Bazel CI due to a conflict of artifact names:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1870#9d520a08-5923-414d-b7d0-6b253eb959dd
, which is what I want to fix to make our CI green again.

examples/ffi/rust_calling_c/c/BUILD
meteorcloudy: Done.

illicitonion (approved):

PR #562, opened 2021-01-28
Pass C++ runtime lib when C++ toolchain declares it

hlopko:
Currently `rules_rust` won't link against the C++ standard library in case
C++ toolchain is the `static_link_cpp_runtimes` feature (therefore when
it uses `cc_toolchain.static_runtime_lib` and `cc_toolchain.dynamic_runtime_lib`
attributes to declare which version of C++ static library should be
linked into the final binary).

This PR modifies `add_native_link_flags` to also look into the
C++ toolchain and depend on C++ standard library from there.

While there, I moved the `get_cc_toolchain` function to the `utils.rs`,
to be consistent with the handling of the Rust toolchain.

### DIFF
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -14,6 +14,8 @@
 
 """Utility functions not specific to the rust toolchain."""
 
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", find_rules_cc_toolchain = "find_cpp_toolchain")
+
 def find_toolchain(ctx):
     """Finds the first rust toolchain that is configured.
 
@@ -24,6 +26,25 @@ def find_toolchain(ctx):
         rust_toolchain: A Rust toolchain context.
     """
     return ctx.toolchains[Label("//rust:toolchain")]
+
+def find_cc_toolchain(ctx):
+    """Extracts a CcToolchain from the current target's context
+
+    Args:
+        ctx (ctx): The current target's rule context object
+
+    Returns:
+        tuple: A tuple of (CcToolchain, FeatureConfiguration)
+    """
+    cc_toolchain = find_rules_cc_toolchain(ctx)
+
+    feature_configuration = cc_common.configure_features(
+        ctx = ctx,
+        cc_toolchain = cc_toolchain,
+        requested_features = ctx.features,
+        unsupported_features = ctx.disabled_features,
+    )
+    return cc_toolchain, feature_configuration
 
 # TODO: Replace with bazel-skylib's `path.dirname`. This requires addressing some
 # dependency issues or generating docs will break.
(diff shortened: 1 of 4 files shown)


hlopko:
cc @mforster

rust/private/utils.bzl
```diff
@@ -97,17 +118,22 @@ def determine_output_hash(crate_root):
...
+        feature_configuration (FeatureConfiguration): The current `feature_configuration` to use with `cc_toolchain`.
 
     Returns:
         depset: A depset[File]
     """
     linker_inputs = dep[CcInfo].linking_context.linker_inputs.to_list()
-    return depset([_get_preferred_artifact(lib) for li in linker_inputs for lib in li.libraries])
+    return depset(
+        [_get_preferred_artifact(lib) for li in linker_inputs for lib in li.libraries],
+        transitive = [cc_toolchain.static_runtime_lib(feature_configuration = feature_configuration)],
```
dfreese, line 135: The [docs](https://docs.bazel.build/versions/master/skylark/lib/CcToolchainInfo.html) suggest:
>The caller should check whether the feature_configuration enables `static_link_cpp_runtimes` feature

Is there a reason to not do that here?

dfreese: Sounds good

dfreese: Last thought, sorry for missing it earlier.

get_libs_for_static_executable is called for every cc_library target.  This would result in the static_runtime_lib being depended upon N times for a target with N cc_library deps.  Is that what we want?

dfreese (commented):
One last question, and then generally good.  Would appreciate a second set of eyes from someone with more of a bazel background.

hlopko:
Update:

I discovered that when creating an rlib artifact, rustc will open every static library and "copy" its contents into the rlib. This results in a quadratic disk usage growth and that's not acceptable in our workloads. I'll create a separate PR to stop passing native libraries when producing rlibs and wait with this one.

rust/private/utils.bzl
hlopko: Good catch! Actually the static runtime lib is in action inputs already because sad reasons, this PR only deals with passing a command line option. But you correctly notices that we put the option there multiple times, and because it's a  depset, only the first occurence will win. That will cause linking to fail in builds with multiple native libraries. I'll update the PR.

hlopko: I've uploaded a fixed version, PTAL at `add_native_link_flags`.

I'm not sure I see a way to write a test for this without having to create a C++ toolchain with static runtime lib. But I have verified it internally in our codebase. WDYT?

hlopko:
https://github.com/bazelbuild/rules_rust/pull/576 has to be submitted first.

hlopko:
This PR is now ready for review. Blocking PRs have been merged, and I've tested it internally with a C++ toolchain that uses static/dynamic_runtime_lib. Please take another look :)

damienmg (approved):

PR #566, opened 2021-02-01
Support passing arbitrary extra flags to rustc

djmarcin:
Adds a bazel command line flag for passing arbitrary extra arguments to all (non-exec configuration) rustc invocations. This is necessary beyond using `rustc_flags` if you want to enable something that needs to be enabled globally, such as `lto=thin`.

Usage:
```
bazel build --@rules_rust//:extra_rustc_flags=--codegen=lto=thin //some:target
```


rust/private/rustc.bzl
```diff
@@ -857,3 +864,11 @@ error_format = rule(
     implementation = _error_format_impl,
     build_setting = config.string(flag = True),
 )
+
+def _extra_codegen_impl(ctx):
+    return ExtraCodegenInfo(extra_codegen = ctx.build_setting_value)
+
+extra_codegen = rule(
```
UebelAndre, line 871: Nit: can you add a doc and plug this into `./docs`

djmarcin: Updated the docs for this (and `error_format`, while I'm at it)

rust/private/clippy.bzl
```diff
@@ -145,6 +145,7 @@ rust_clippy_aspect = aspect(
             default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
         ),
         "_error_format": attr.label(default = "//:error_format"),
+        "_extra_codegen": attr.label(default = "//:extra_codegen"),
```
hlopko, line 148: So it's quite likely we'll grow the number of these settings that each rule will want to depend on. What do you think about (all or subset of):
* Creating a `config` package at //rust/private/config
* There the BUILD file will define build_setting targets
* There the config.bzl file will define providers, rules, and a list of implicit attributes all the rules can just add
* rust_common struct will have a config sub-struct with functions returning the provider of the particular build_setting, something that instead of `ctx.attr._extra_codegen[ExtraCodegenInfo].extra_codegen` the user will write `rust_common.config.extra_codegen(ctx)`.

The goal is to allow third_party rules to depend on rust specific build settings without hardcoding the name and label of implicit attributes, and without needing to use the provider associated with the setting.

rust/private/rust.bzl
```diff
@@ -543,7 +543,10 @@ _common_attrs = {
         """),
     ),
     "rustc_flags": attr.string_list(
-        doc = "List of compiler flags passed to `rustc`.",
+        doc = _tidy("""
+            List of compiler flags passed to `rustc` for this target. For flags that should be
```
hlopko, line 547: Could you mention that flags specified in the attribute are not applied to dependencies of the target? Folks often get this wrong.

I wonder if we should mention codegen flags in rustc_flags attribute. Ideally we would have full symmetry and full coverage, for example:

* extra_rustc_flags attribute and @rules_rust//:extra_rustc_flags setting
* extra_codegen_flags attribute and @rules_rust//:extra_codegen_flags setting

And then we would only mention the setting in the relevant attribute.

(but also see my bikeshed comment about naming above :)

rust/private/rustc.bzl
```diff
@@ -864,11 +871,31 @@ def _error_format_impl(ctx):
...
+    doc = ("""
+A helper rule for passing [--codegen](https://doc.rust-lang.org/rustc/codegen-options/index.html)
+parameters to rustc for ALL targets, such as `lto=thin`. Codegen parameters that apply only to
```
hlopko, line 892: Specifically for `lto=thin`, I think we could also support the rules_cc way of controlling that - using `--features=thin_lto`. Features are accessible through https://docs.bazel.build/versions/master/skylark/lib/ctx.html#features, and we could query the C++ toolchain to see if it supports thin_lto. In a separate PR is fine, and I'm happy to chat more about this.

hlopko (commented):
Thanks for working on this, this is a very useful PR!!

docs/BUILD
```diff
@@ -36,6 +36,8 @@ PAGES = {
         "rust_binary",
         "rust_benchmark",
         "rust_test",
+        "error_format",
```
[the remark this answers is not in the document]
djmarcin: Sure I can remove it from this PR.

rust/private/clippy.bzl
djmarcin: I'm not sure I understand what you've outlined here. Is the goal that e.g. a `rust_binary` rule could ask for LTO and it would automatically apply down to all dependencies? That would be great, but I didn't think that was possible -- if it is I can certainly try to reimplement it that way.

rust/private/rust.bzl
djmarcin: Ah, because rustc_flags is more general than codegen, specifically -- that makes sense. Do you think instead of an extra codegen flag this should just be extra rustc flags, then?

docs/BUILD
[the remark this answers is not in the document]
UebelAndre: > @UebelAndre is there a way to not have to check in generated docs, but build them on the CI? Have you chatted with Bazel folks about this? Are all the Bazel rules checking in generated docs?

There are a few stragglers which are not being generated if I recall but for the most part, yeah, all major rules are hooked up to generate docs. If CI can run a script and commit something back to the repo then we can have CI update docs no problem. Simply run `./docs/update_docs.sh`. I think this'd be a fantastic improvement. In general, if there's a way to run arbitrary scripts, we can test to prevent generated docs from being modified and forcing users to update the source of truth (which has happened from time to time and documentation was stomped).

rust/private/rustc.bzl
djmarcin: I looked at that, but I thought that --features=lto was specifically a rules_cc thing. Also if we use the same flag as rules_cc, it becomes impossible to control them individually, and there are more options than just "thin" for Rust.

My actual goal was to get `--codegen=linker-plugin-lto` working, but I haven't gotten there yet because it requires linker plugins to work correctly, and `--codegen=lto=thin` is a simpler thing to get working with similar benefit, so merely supporting --features=thin_lto wouldn't suffice.

hlopko: Internally `--features=thin_lto` affects all languages in the build. Do you anticipate the need to say build rust with thinlto but not C++, in a single build?

There are more options than just "thin" in Rust, but only "thin" is compatible with rules_cc, and "thin" is so much faster to build with Bazel thatI don't think full LTO is going to be used often.

All that said, I'm fine with having rust-specific lto switch, I was proposing to in addition also look into rules_cc's thing and enable thin lto when they do.

I'm not sure I follow the last paragraph, why doesn't `--features=thin_lto` suffice to enable thin lto?

rust/private/clippy.bzl
hlopko: Nope nope I only outlined a way to structure the code in a way that protects third_party (for example if rules_ruby wanted to support linking native dependencies from rules_rust) rules from depending on internals of rules rust. Does it make sense? Which part doesn't?

rust/private/rust.bzl
hlopko: Yeah I can imagine having both at some point, but starting with `extra_rustc_flags` makes sense to me. Or `rustcopts` :)

rust/private/rustc.bzl
djmarcin: I was under the impression from https://doc.rust-lang.org/rustc/linker-plugin-lto.html that `--codegen=lto=thin` is not the correct way to do LTO over C++ & Rust together. That page seems to imply that you need to use the `--codegen=linker-plugin-lto` in Rust and the `-flto=thin` flag in C++.

The docs aren't super clear about whether this is mutually exclusive with `--codegen=lto=thin` though, so maybe we can still respond to `--features=thin_lto` by setting the `--codegen=linker-plugin-lto` _and_ `--codegen=lto=thin`, but I haven't tried it. I just found a vague mention here that they might be compatible https://doc.rust-lang.org/rustc/codegen-options/index.html#linker-plugin-lto

rust/private/clippy.bzl
djmarcin: Ok -- I think I follow now, this is a code organization / api surface suggestion, right? I'll take a look.

UebelAndre:
Howdy, any updates here? This seems like an awesome feature 😄

rust/private/clippy.bzl
djmarcin: Sorry -- I wasn't able to get time to work on this. I've rebased the PR on top of the latest commits, but I don't think I have bandwidth for the refactor you're talking about right now.

djmarcin:
This is updated on latest HEAD and tests are passing.

rust/private/rustc.bzl
UebelAndre: It still seems like there's a missing `docs` attribute and it doesn't seem like this is appearing in docs. It'd be nice to have this exposed there so people can easily discover it

UebelAndre (changes requested):

djmarcin:
Added some docs, wasn't completely sure where to put them so I added them to the `defs` section.

UebelAndre (approved):
Small nit but otherwise looks good to me! 😄 Thanks for reviving this PR!!

rust/private/rustc.bzl
```diff
@@ -955,6 +954,12 @@ def _extra_codegen_impl(ctx):
     return ExtraCodegenInfo(extra_codegen = ctx.build_setting_value)
 
 extra_codegen = rule(
+    doc = (
+        "Add additional [--codegen](https://doc.rust-lang.org/rustc/command-line-arguments.html#option-error-format) " +
+        "options from the command line with --@rules_rust//:extra_codegen. See rustc documentation for valid values. " +
```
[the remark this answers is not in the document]
djmarcin: Fixed

sayrer:
Could we land this? :)

djmarcin:
> Could we land this? :)

I'll resolve the conflicts. I think there was some bikeshedding over the flag name. I don't have a strong opinion.

djmarcin:
I guess there's also the question about if we should just make this extra_rustc_flags since that's a more general mechanism. I'll make that update since it seems to solve the bikeshedding question as well.

djmarcin:
@hlopko do you have any advice on handling the issue mentioned here: https://github.com/bazelbuild/rules_rust/pull/566/files#diff-ecce7f6cdb4bc2f884f23fccb9a20174ad2b357dc6745314054e5eb8b784d2e0R500-R506

In particular, adding LTO flags to proc-macro crates breaks them for me. In any case, it seems unlikely that the user's intention here would be to pass these extra flags to targets built for the exec platform -- in my case I only want these to be passed for the target platform. However I don't really know how to detect whether we are in the exec or target platform.

djmarcin:
I retested LTO and it turns out that it's no longer broken (probably something I did with toolchain resolution in my repo lately if I had to guess). In any case, excluding these flags from the exec configuration still seems like the right thing to do to me. I've changed to detecting this based off of the `genfiles_dir` which appears to always contain the string `-exec-` when running in the exec configuration. It's a hack, but I don't see where else to get this information.

rust/private/rustc.bzl
```diff
@@ -492,6 +497,13 @@ def construct_arguments(
     # Set the SYSROOT to the directory of the rust_lib files passed to the toolchain
     env["SYSROOT"] = paths.dirname(toolchain.rust_lib.files.to_list()[0].short_path)
 
+    # extra_rustc_flags are meant to apply across all code being compiled in the target
+    # configuration, not the exec configuration.
+    # TODO(djmarcin): Should these flags apply to cfg=exec targets as well? Or via a
+    # different flag. Also, is there a better way to determine cfg=exec?
+    if hasattr(ctx.attr, "_extra_rustc_flags") and ctx.genfiles_dir.path.find("-exec-") == -1:
```
UebelAndre, line 504: Can you put this in a utility function in `//rust/private/utils.bzl` so it stands out a bit more?

UebelAndre (changes requested):

rust/private/rustc.bzl
djmarcin: Done

hlopko:
I asked about the host/exec detection in https://github.com/bazelbuild/bazel/issues/13308#issuecomment-895860044, let's see what they say.

Could you add a unittest showing that a flag does appear on the command line for a target in target configuration, and that it doesn't appear for a target in exec configuration? Should be fairly simple given https://github.com/bazelbuild/bazel-skylib/blob/main/lib/unittest.bzl#L188. Thank you so much!

hlopko:
Ok, it seems the genfiles_dir path is the only way to detect exec configuration. From me this PR looks good, modulo the unit test (if you don't have capacity to add the test please let us know, maybe somebody will be able to help out).

UebelAndre (approved):

rust/private/rustc.bzl
UebelAndre: Thanks!

djmarcin:
@hlopko merged latest main and added tests

UebelAndre:
Is the PR description up to date? How do we set flags?

djmarcin:
Updated the PR description

docs/symbols.bzl
```diff
@@ -13,8 +13,7 @@ load(
     _rust_bindgen_repositories = "rust_bindgen_repositories",
 )
 load(
-    "@rules_rust//cargo:defs.bzl",
```
UebelAndre, line 16: This seems like a bad merge conflict resolution. Can you restore this and the other deleted things in this file?

UebelAndre (changes requested):
It seems some stuff got removed from docs. If you fix this I can merge give that @hlopko gave a thumbs up pending a test

docs/symbols.bzl
djmarcin: Redid the merge, looks better now, I think. Please take another look.

docs/BUILD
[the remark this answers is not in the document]
djmarcin: Split to https://github.com/bazelbuild/rules_rust/pull/924

PR #569, opened 2021-02-02
Add //rust/private/BUILD (therefore create a package there)

hlopko:
This PR achieves the following:
* buildifier bzl-visibility check was not detecting violations as it assumes `private` package, not just a directory
* adds `//rust:common.bzl` file where all the publicly available and supported API for writing custom rules interacting with Rust will reside
* the private package now sets the stability expectations, and it forces us to think about API layering and abstractions.

This turned out to be much bigger PR than expected :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@
 
 # npm
 **/node_modules/
+
+# vscode
+.vscode
+*.code-workspace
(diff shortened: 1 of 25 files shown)


illicitonion (approved):
Sure, why not!

.gitignore
```diff
@@ -17,3 +17,7 @@
 
 # npm
 **/node_modules/
+
+#vscode
```
UebelAndre, line 21: Nit, space 😅

hlopko: Done :)

UebelAndre (approved):
LGTM! 😄

PR #570, opened 2021-02-02
Run buildifier fix across the workspace

hlopko:


illicitonion (commented):
Could you add a check to the presubmit to enforce this for the future?

I'm on board with one-off disruptive formatting changes, but only if there's only one of them, which is less likely to be the case without a presubmit :)

hlopko:
Oh good point, I thought we already get all the warnings. My bad. Done.

This is how we stand against the full wrath of the buildifier before this PR: https://buildkite.com/bazel/rules-rust-rustlang/builds/2054

illicitonion (approved):
Thanks!

PR #576, opened 2021-02-03
Do not pass native libraries to rlib compilations

hlopko:
When creating an rlib artifact, rustc will open every
native static library and "copy" its contents into the rlib. This
results in a quadratic disk usage growth and that's not necessary for
rules_rust because we track native dependencies and therefore we can pass
them to the final linking actions.

This fixes the build failure when a native dep is actually a linker
script. Rustc will try to open the linker script as if it was a static
archive, and then fail the build.

Creating shared library artifacts (dylib, cdylib) and binaries is not
affected, for those rustc uses the linker, and linker will consume the
linker script just fine. Rustc won't try to open/interpret the linker
script itself.

This also unblocks #562.


hlopko:
CC @UebelAndre @dfreese feel free to add yourselves as reviewers :)

rust/private/rust.bzl
```diff
@@ -230,14 +233,14 @@ def _rust_test_common(ctx, toolchain, output):
...
-
+    crate_type = "bin"
     if ctx.attr.crate:
         # Target is building the crate in `test` config
         # Build the test binary using the dependency's srcs.
         crate = ctx.attr.crate[rust_common.crate_info]
         target = rust_common.crate_info(
             name = crate_name,
-            type = crate.type,
+            type = crate_type,
```
illicitonion, line 243: Should this assert that `crate.type == "bin"`?

illicitonion (approved):

rust/private/rust.bzl
hlopko: I don't think so? Crate under test can be a rust_library with crate type rlib or pretty much any other, it's only the test that will be "bin" in any case.

UebelAndre (approved):
This seems good to me but only have the bandwidth right now for a high level review. 😅

hlopko:
Oki, will merge now, feel free to review afterwards, I'm happy to address any feedback you might have. 🙇

Issue #574, opened 2021-02-03
FR: Add `env` attribute for test rules


UebelAndre:
This is to follow the pattern set by [common attributes: tests](https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests) but would generally be a useful thing to have.

PR #577, opened 2021-02-04
Added `env` attribute to `rust_test` rule

UebelAndre:
Closes https://github.com/bazelbuild/rules_rust/issues/574

I realized in https://github.com/google/cargo-raze/issues/313#issuecomment-772117832 that there is no way to set environment variables for tests. This is something I think should be doable as it can be used to make some tests hermetic. In this case, I have https://github.com/UebelAndre/cargo-raze/tree/hermetic where I'm working on a way to make the `cargo-raze` tests (which require a cargo binary) hermetic. We could use the binary in the toolchain if only it was accessible.

The `env` environment variable seems to solve for that and is consistent with [Bazel common attribute (tests)](https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests) patterns.


illicitonion (approved):
Thanks!

PR #579, opened 2021-02-08
Rust test targets now create a test launcher allow for setting env vars

UebelAndre:
I was mistaken when I originally implemented https://github.com/bazelbuild/rules_rust/pull/577 because I didn't realize that [std::env](https://doc.rust-lang.org/std/macro.env.html) loaded environment variables at compile time. The goal of this PR was to define environment variables at runtime. This appeared to be more complicated than I had thought. It seems there needs to be a launcher/runner that is produced by your rule and invokes your test executable to be able to guarantee the environment is configured for the test.

This PR creates a small shell script that wraps `rust_test` targets to enable the use of the `env` attribute at runtime.


rust/private/test_runner_template.rs
```diff
@@ -0,0 +1,43 @@
...
+/// definition of this map.
+fn environ() -> BTreeMap<&'static str, &'static str> {
+// {environ}
+}
+
+
+/// Simply replace the current process with our test
+#[cfg(target_family = "unix")]
+fn exec(environ: BTreeMap<&'static str, &'static str>) {
+    Command::new("{executable}")
```
illicitonion, line 18: This should also take any args (which may be passed with things like the [`--test_arg`](https://docs.bazel.build/versions/master/command-line-reference.html#flag--test_arg) flag) and pass them as extra `argv`.

illicitonion (commented):
This looks useful, and I'm surprisingly OK with it given it's generating compilable rust code in starlark :D I'm a little worried about not escaping quotes in values, so I'd lean towards using something like:
```rust
"r#####\"{}\"#####"
```

in the templates rather than
```rust
"\"{}\""
```

I'd definitely like others' opinions before merging, though, as this is a significant change, and there are potentially a lot of edge cases :)

rust/private/test_runner_template.rs
UebelAndre: Added an `args` function

rust/private/rust.bzl
```diff
@@ -219,6 +219,97 @@ def _rust_binary_impl(ctx):
...
+        getattr(ctx.attr, "env", {}),
+        getattr(ctx.attr, "data", []),
+    )
+
+    # Generate strings for inlining them in the Rust source file
+    environ_defines = []
+    if environ:
+        environ_defines.append("    let mut environ = BTreeMap::new();")
+        for env, value in environ.items():
+            environ_defines.append("    environ.insert(\"{}\", r#####\"{}\"#####);".format(
```
illicitonion, line 247:
```suggestion
            environ_defines.append("    environ.insert(r#####\"{}\"#####, r#####\"{}\"#####);".format(
```

rust/private/test_runner_template.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// Parse the command line arguments but skip the first element
+fn args() -> Vec<String> {
+    std::env::args().skip(1).collect()
+}
+
+/// Simply replace the current process with our test
+#[cfg(target_family = "unix")]
+fn exec(environ: BTreeMap<&'static str, &'static str>) {
+    Command::new("{executable}")
```
illicitonion, line 24:
```suggestion
    Command::new(r#####\"{executable}\"#####)
```

rust/private/test_runner_template.rs
```diff
@@ -0,0 +1,50 @@
...
+    .envs(environ.iter())
+    .args(args())
+    .exec();
+}
+
+/// On windows, there is no way to replace the current process
+/// so instead we allow the command to run in a subprocess.
+#[cfg(target_family = "windows")]
+fn exec(environ: BTreeMap<&'static str, &'static str>) {
+    let output = Command::new("{executable}")
```
illicitonion, line 34:
```suggestion
    let output = Command::new(r#####\"{executable}\"#####)
```

tools/runfiles/runfiles.rs
```diff
@@ -69,15 +80,31 @@ impl Runfiles {
 fn find_runfiles_dir() -> io::Result<PathBuf> {
     let exec_path = std::env::args().nth(0).expect("arg 0 was not set");
 
-    let mut binary_path = PathBuf::from(&exec_path);
+    let mut binary_path = PathBuf::from(&exec_path)
+        .canonicalize()
```
illicitonion, line 84: Why the call to `canonicalize`? This will plausibly break consumers who use symlinks

illicitonion (commented):
Works for me, but again, would want to hear from others before merging :)

rust/private/test_runner_template.rs
UebelAndre: I figured this and the environment variable would be safe but I'm happy to be consistent 😄

tools/runfiles/runfiles.rs
UebelAndre: Oh yeah... This fails when running tests. I'll remove it to demonstrate the error.

UebelAndre: Or just find an old run: https://buildkite.com/bazel/rules-rust-rustlang/builds/2145#0f14a372-551e-439d-8513-33c1875a894f

Because the path of `binary_path` is relative, we don't traverse far enough to find the `runfiles` directory.

UebelAndre: Would love some suggestions on how to better handle this 🙏

illicitonion: We can absolutise; something like:
```rust
if !binary_path.is_absolute() {
  binary_path = std::env::current_dir().expect("Failed to get current directory").join(binary_path);
}
```

But we should definitely not resolve symlinks.

UebelAndre: That seemed to do the trick 😄

rust/private/test_runner_template.rs
illicitonion: Looks like the wrong comment on this line got resolved :)

illicitonion: Looks like this one's still missing :)

UebelAndre: Fixed 😄

hlopko:
I kept thinking and chatting with people about this yesterday. The Bazel team said they would accept a contribution to allow specifying `env` for test runs the same way they support for native rules. I don't expect that to be trivial, but doable. So that's one avenue.

Second avenue is to not use a rules_rust provided testrunner, but to bake custom one only for the tests that need to have env vars set, only in the project that has the affected tests. It doesn't even have to be a template, the test runner can assume a data file at some well specified location.

Third is the approach of this PR, to have a general Rust test runner that is used for all Rust tests built with rules_rust. As with the second point it doesn't necessarily have to be a template, we can use data files instead. This approach has a potential/handwave-handwave advantage that once we want to produce junit xml test outputs which Bazel can parse and report nicer test results, we will have to have the test runner anyway. The downside is a bit of complexity and my paranoid fear of misusing the test runner for other stuff.

Andre, Daniel, what do you think about this?

UebelAndre:
I don't think users should have to bake environment variables into their tests and it appears that most user's expect to be able to do something like in this PR (as per a recent [slack thread](https://bazelbuild.slack.com/archives/CA31HN1T3/p1612651687394800)). I think for the first and third options, they're more or less invisible to the user. They'll want to set an `env` attribute and will be able to with both approaches. I feel this PR is a good enough start and we can refactor it into something that is more in alignment with upstream Bazel changes or replace it for something entirely.

illicitonion:
> The Bazel team said they would accept a contribution to allow specifying `env` for test runs the same way they support for native rules. I don't expect that to be trivial, but doable. So that's one avenue.

This sounds long-term ideal.

> Second avenue is to not use a rules_rust provided testrunner, but to bake custom one only for the tests that need to have env vars set, only in the project that has the affected tests. It doesn't even have to be a template, the test runner can assume a data file at some well specified location.

This sounds not particularly worth the complexity, particularly if in the long term we're either going to delete the test runner, or run everything under it anyway (for e.g. junit XML).

> Third is the approach of this PR, to have a general Rust test runner that is used for all Rust tests built with rules_rust. As with the second point it doesn't necessarily have to be a template, we can use data files instead. This approach has a potential/handwave-handwave advantage that once we want to produce junit xml test outputs which Bazel can parse and report nicer test results, we will have to have the test runner anyway. The downside is a bit of complexity and my paranoid fear of misusing the test runner for other stuff.

Seems reasonable in the short term, and long-term we can either delete it, or build it up more.

hlopko:
After looking into ways how to make debugging work I yield :) Let's do this. I propose a different approach that I think will result in simpler code:

* I'd make the rust test runner a rust_binary, and make it an implicit attribute of rust_test rule
* rust test runner will look for a file with environment variables and for a test binary in its runfiles (well known locations and symlinks work well enough imho)

I think it's worth doing this because:

* this way we don't have to recompile the runner for every test
* we can make the runner a regular rust_binary target, we don't have to duplicate the building logic/provider merging and so on
* if the future will ask for modifications, we can add dependencies or more sources to the test runner cleanly

What do you think?

illicitonion:
Sounds good!

UebelAndre:
@hlopko @illicitonion How's something like this look?

rust/private/BUILD
```diff
@@ -1,5 +1,7 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
+exports_files(["test_runner_template.rs"])
```
hlopko, line 3: This can be removed, right?

rust/private/rust.bzl
```diff
@@ -219,6 +219,100 @@ def _rust_binary_impl(ctx):
         ),
     )
 
+def _create_test_launcher(ctx, toolchain, output, providers):
```
hlopko, line 222: Do we need to do all of this? What I had in mind was to add this into rust_test rule declaration:

```rust
"_process_wrapper": attr.label(
        default = Label("//util/test_runner"),
        executable = True,
        allow_single_file = True,
        cfg = "exec",
    ),
```

And then:
0) add a rust_binary at //util/test_runner:test_runner
1) in rust_test access the runner using `ctx.executable._test_runner`
2) put that in the DefaultInfo that rust_test provides
3) register action that writes list of env variables into a file
4) put that file into DefaultInfo runfiles
5) register action that symlinks the actual test binary into some path
6) put both the test and the symlink into runfiles
7) modify test_runner to look into runfiles for env-vars file and for the symlink

Does it make sense?

rust/private/rust.bzl
```diff
@@ -511,6 +606,23 @@ _rust_test_attrs = {
...
+    "_launcher": attr.label(
+        executable = True,
+        default = Label("//util/launcher:launcher"),
+        cfg = "exec",
+        doc = _tidy("""
+            A launcher executable for loading environment and argument files passed in via the `env` attribute
+            and ensuring the variables are set for the underlying test executable.
+        """),
+    ),
+    "_launcher_installer": attr.label(
```
hlopko, line 618: I think the rust_binary will deal with the extension automatically, so you won't need the installer

tools/runfiles/runfiles.rs
```diff
@@ -37,6 +37,17 @@ use std::io;
...
 use std::path::PathBuf;
 
+
+/// The common name of Bazel runfiles directory suffix
+const RUNFILES_SUFFIX: &'static str = ".runfiles";
+
+/// Tests are wrapped in a `.runner` executable so we need to 
+/// also know to look for this directory as well
+#[cfg(target_family = "unix")]
+const TEST_RUNFILES_SUFFIX: &'static str = ".runner.runfiles";
```
hlopko, line 47: Could you elaborate on why do we need to look into the runners runfiles from the test?

rust/private/rust.bzl
UebelAndre: This is different from the process wrapper because the process wrapper is used as an executable for an action. However, the launcher/test_runner needs to be the executable returned by the rule so it gets executed when you `bazel test //:my_target`. Bazel will otherwise throw an error if you return an executable that is not created by an action of that rule. Is there another way around this? Aside from having to create the binary in an action in the rule, I what you've written is more or less what I'm doing, right?

rust/private/rust.bzl
UebelAndre: The installer is just a script that helps us create the test runner/launcher in an action in the test rule. I made this target so I can use a batch file on windows and not require the use of `bash`.

tools/runfiles/runfiles.rs
UebelAndre: It's not needed anymore. I've removed this,thanks for catching it 😄

UebelAndre:
Also, I changed the `runner` to be called a `launcher` to be more consistent with upstream bazel: https://github.com/bazelbuild/bazel/tree/master/src/tools/launcher

illicitonion (approved):

rust/private/rust.bzl
```diff
@@ -219,6 +219,100 @@ def _rust_binary_impl(ctx):
...
+
+    args = ctx.actions.args()
+
+    # Because returned executables must be created from the same rule, the 
+    # launcher target is simply copied and exposed. 
+
+    # TODO: Find a better way to detect the configuration this executable
+    # was built in. Here we detect whether or not the executable is built 
+    # for windows based on it's extension.
+    if ctx.executable._launcher.basename.endswith(".exe"):
```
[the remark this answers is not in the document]
UebelAndre: I wasn't sure if the toolchain would always match the configuration of `launcher` but I think in most cases it's safe to use the toolchain. Updated

PR #583, opened 2021-02-10
Set PROFILE env var in cargo build scripts

illicitonion:
Is this exactly correct? No. Is it good enough? Probably.

See https://doc.rust-lang.org/cargo/reference/environment-variables.html


hlopko:
What do you think about using `ctx.var["COMPILATION_MODE"]` to access the value of `--compilation_mode`? That's what native rules use. The value can be 'opt', 'dbg', and 'fastbuild'.

illicitonion:
> What do you think about using `ctx.var["COMPILATION_MODE"]` to access the value of `--compilation_mode`? That's what native rules use. The value can be 'opt', 'dbg', and 'fastbuild'.

Sure! Done :)

cargo/cargo_build_script.bzl
```diff
@@ -82,6 +82,8 @@ def _build_script_impl(ctx):
         "CARGO_PKG_NAME": crate_name,
         "HOST": toolchain.exec_triple,
         "OPT_LEVEL": compilation_mode_opt_level,
+        # This isn't exactly right, but Bazel doesn't have exact views of "debug" and "release", so...
+        "PROFILE": {"opt": "release", "dbg": "debug", "fastbuild": "debug"}.get(ctx.var["COMPILATION_MODE"], "unknown"),
```
hlopko, line 86: LGTM, thanks!

One final nit, is debug build actually faster than release? In C++ we often see that merging debug info is slow, and tests built in debug run prohibitively slow. So fastbuild is more like a release build, maybe with fewer optimizations, but not full debug.

I don't think it matters for build scripts, but it might set the precedence for future changes.

I'll trust your judgement here, feel free to pick what makes sense and merge. Thank you!

hlopko (approved):

cargo/cargo_build_script.bzl
illicitonion: For compile I'm not sure, but the linking stage is definitely a whole lot faster (frequently 1-2 orders of magnitude) in debug than release. [Cargo defaults tests to run in debug](https://doc.rust-lang.org/cargo/reference/profiles.html#test), so presumably it's not a wide problem, but it's tweakable if people need to...

I don't think there's a clearly right or wrong thing to do in terms of this mapping here (cargo similarly allows debug to have optimisations enabled, and release to have them disabled) - in terms of defaults for what optimisations to actually enable/disable, I suspect what we're doing now is reasonable.

PR #586, opened 2021-02-12
Updated min tested version to 3.5.0 and improved some vernacular

UebelAndre:
The only thing outstanding is one thing for `rust_bindgen` which is blocked on https://github.com/rust-lang/rust-bindgen/pull/1990

Very happy to see @hlopko opening that PR 😄 

Additionally, this PR raises the min test version to 3.5.0.


dfreese:
I don't see what in this PR requires a bump in the min tested version.  Can you clarify?

UebelAndre:
`_allowlist_function_transition` is first introduced in `3.5.0`
https://docs.bazel.build/versions/3.5.1/skylark/config.html#user-defined-transitions
https://docs.bazel.build/versions/3.4.0/skylark/config.html#user-defined-transitions

(also, unrelated, while you're here @dfreese, can I get your eyes on some `cargo-raze` PRs 🙏?)

docs/index.md
```diff
@@ -32,7 +32,9 @@ load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 rust_repositories()
 ```
 
-The rules are under active development, as such the lastest commit on the master branch should be used. `master` currently requires Bazel >= 3.0.0.
+The rules are under active development, as such the lastest commit on the `main` branch should be used. `main` is only tested
+against `4.0.0` as the minmum supported version of Bazel. Though previous versions may still be supported in certain
```
hlopko, line 36: s/minmum/minimum

I'd also wrap lines at 80, but anything under 120 (which I guess is the width of the GitHub diff view) works for me.

hlopko:
>  I'm tired of dealing with older versions

I'm not against being pragmatic about our backwards compatibility, but don't you find it a bit ironic that in the PR addressing respectful language we make the rules less usable to people having older Bazel version?

If that's not too much to ask, would you mind lowering the version to 3.5? If there is a technical reason, I'm happy to discuss the options.

Thanks for working on this!

docs/index.md
UebelAndre: wrapped to 80 (for the affected lines)

UebelAndre:
The reason I like 4.0.0 is because it's an LTS release. All rules should always support a LTS release while it's supported. I'm happy to drop to 3.5 but testing on 4.0 doesn't mean we don't support 3.5 and I feel it's a bit arbitrary when support changes specifically because there are no releases for these rules.

But yeah, to be clear, there's no technical reason, It's more of a logistical thing for me (and I feel like I'm constantly making backwards compatibility changes in my PRs).

hlopko:
Agreed on 4.0.0, we should definitely support that.

Testing on 3.5 means we support 3.5. If we only test on 4.0 it's only a question of time when we break compatibility with 3.5. It is ok to intentionaly break compatibility, but it's better to know about it and have a conversation about it.

I'm coming from the perspective of a well behaving user using Bazel 3.5, coming to this PR, and seeing we bumped the min version to 4.0 and we didn't have a strong reason to.

Doing backwards compatibility changes where it makes sense is IMHO the right thing to do. But I agree that we don't have a set policy on that, so if you want to change the status quo let's have a discussion on the mailing list?

UebelAndre:
I'll see if I can find the time to learn how to use the mailing list 😅 

Is there anything more to do for this PR?

UebelAndre:
I forgot to push! So sorry about that 😅 The PR should *now* be addressing all concerns.

hlopko (approved):
LGTM, thank you!

PR #590, opened 2021-02-16
Migrate `cargo-raze` into `rules_rust`

UebelAndre:
This PR moves [cargo-raze](https://github.com/google/cargo-raze) into this repo. There is a lot of functionality that is coupled between the two projects that I feel it makes sense to have the code in one place. 

This is the simplest form of a migration I can think of. It moves all the code into `cargo/cargo_raze` and adds the directory to `.bazelignore` so it doesn't affect existing CI. 

A future change should be made to better integrate this into CI but that can be a separate PR with tracked changes and more realistically reviewable PRs.

cc: @acmcarther @damienmg @illicitonion @mfarrugi @dfreese @hlopko


UebelAndre:
A [slack thread](https://bazelbuild.slack.com/archives/CSV56UT0F/p1613492450073000) was started for this PR

damienmg:
I don't have access to the slack thread. I am curious about the legal implication of such a change and we need to reach out to the Google Open Source Office before merging such a change.

Before that, I'd like to get opinions from David and Alex on wether they think this is the right move or not.

Finally we probably should not duplicate the LICENSE file, but that's still to be discussed with the Open Source Office.

UebelAndre:
@damienmg Can you potentially get that started and link any information about that conversation here?

illicitonion:
I have no objection to this, but I'm also not in a place to advocate for it. If Alex is excited about it, I'm happy to review, but I don't really have a stance on whether this is a good direction for the project overall.

hlopko:
Some thoughts:

* It's a huge PR
* I'm not sure checking in generated files at this scale is the right way forward
* Are there instructions for upgrading dependencies of raze?
* Are we continuing with the `cargo raze` workflow, or is `bazel run @rules_rust//cargo:raze` going to be the only supported approach?

Please do not merge this before we hear from @acmcarther and others, and before we have a green light from the Google Open Source Program Office.

hlopko:
I'll have to digest this PR a bit :)

If we decide to move forward with this, for the process I propose to:
1) check in all the files in the state they are, without any manual modifications. It's ok if stuff doesn't build. We need the PR to be easily verified/reviewed. I want to run diff on 2 repos and see nothing was modified.
2) then slowly and steadily go one PR at a time with reviews and responsible engineering.

hlopko:
What do you plan to do about existing issues?

UebelAndre:
>     * It's a huge PR

The commits in the description are the only notable commits. I've been trying to get the first block merged but there hasn't really been bandwidth for reviews.

>     * I'm not sure checking in generated files at this scale is the right way forward

Yeah, the PR is definitely large... Maybe the examples could be done separately? It's been stated in previous issues that they're not in the ideal state since they're currently more "integration" tests and are acting as replacements for the lack of unittesting.

>     * Are there instructions for upgrading dependencies of raze?

There's a `bazel run //cargo/cargo_raze/tools:bootstrap` target that can build it in cargo. Otherwise if the `cargo_raze` target can compile, it can update it's own dependencies (have tested already in an earlier PR to `cargo-raze`)

>     * Are we continuing with the `cargo raze` workflow, or is `bazel run @rules_rust//cargo:raze` going to be the only supported approach?

The cargo raze workflow is called out in the docs for `rules_rust` currently and as a result I feel the tool has become more prominent than others. But I also don't think this means there can't be other workflows. If you look at the contributors for both `cargo-raze` and `rules_rust` there's a large amount of overlap, if that means anything...

hlopko:
I'd also like to understand the difference between raze and reindeer. In the company I work we don't use raze, the biggest reason is that having a single Cargo.toml file doesn't scale for us.

Without that, I'm not sure we should be "blessing" one solution in the rules_rust.

UebelAndre:
>     1. check in all the files in the state they are, without any manual modifications. It's ok if stuff doesn't build. We need the PR to be easily verified/reviewed. I want to run diff on 2 repos and see nothing was modified.

That's more or less what I did for this PR. 782b3cbb2b2c9ef3279d53e5491208f3536a23f1 is the merge commit from my fork of `cargo-raze` (again, I'd have taken upstream but there are required changes I can't get in without a reviewer).

>     2\. then slowly and steadily go one PR at a time with reviews and responsible engineering.

Agreed, but to be clear are you saying to merge `cargo-raze` over in an "incompatible" state? Maybe where `rules_rust` was conditioned to not fail CI when it's merged?

UebelAndre:
> the biggest reason is that having a single Cargo.toml file doesn't scale for us.

Cargo raze totally supports multiple `Cargo.toml` files 😄 @acmcarther and I spent quite a while discussing support of that.

UebelAndre:
> the biggest reason is that having a single Cargo.toml file doesn't scale for us.

Cargo raze totally supports multiple `Cargo.toml` files 😄 @acmcarther and I spent quite a while discussing support of that. The "single-toml" and "multi-toml" setups, if I recall correctly.

hlopko:
And last thought before I go digesting :)

IMHO raze can be cleanly separated in its own repository, and  only depend on the public interface of rules_rust. If changes to rules_rust break raze too often, we need to discuss our backwards compatibility policy.

UebelAndre:
> IMHO raze can be cleanly separated in its own repository, and only depend on the public interface of rules_rust. If changes to rules_rust break raze too often, we need to discuss our backwards compatibility policy.

It's not that changes to `rules_rust` breaks `cargo-raze`. It's more that someone wants to implement some new functionality that needs to be done in `rules_rust` and then setup in `cargo-raze` or `cargo-raze` was used to generate something but doesn't support some functionality so a change must be made there in order to update/re-generate something in `rules_rust`.

Again, very interested in your workflow. While I think merging the two repos would be a good idea, It's not (currently) my burning passion. I like the idea that the rules has a standardized tool so that all workflows can, to a more thorough degree, be brought to the same place and everyone can benefit from questions/changes.

UebelAndre:
One other thing of note, I think if there's considerations for adding something like `cargo_universe` from https://github.com/bazelbuild/rules_rust/issues/2#issuecomment-738803545 to `rules_rust` then I think it'd also be beneficial to have `cargo_raze` in here as well since it's a dependency of `cargo_universe`.

UebelAndre:
@hlopko Thanks to @dfreese I've been able to drop most of the commits going into this change. There are only two commits that are noteworthy now. The merge commit (c965cee0598b95f8296e0c467c5c191bc9ab0c42) and a commit that adds the new repository to `.bazelignore` (f5dd0ac916c7b67e900dab06c9f716cfb1dbb441).

dfreese:
I've always wondered if it's worth having rules_rust focus explicitly on rustc, and then have other things implemented on top of that.  What has been hard, I think is that rules_rust didn't expose enough of an interface for people to build things like clippy or protobuf integration on top of it.

The architecture that is described in #596, which @hlopko has been working towards with rust_common (providing a more rule_cc-like interface to rules rust) starts to solve some of those issues.

This PR is probably an opportune time to think about drawing some clear lines about what is rust and what is cargo.

Where it has been intermixed has led to some weird results, such as rules_proto being stuck on a particular crate version, where updating it would be a breaking change.  Proper versioning in rules_rust would probably help (though it has a larger API surface than I would like).

UebelAndre:
> I've always wondered if it's worth having rules_rust focus explicitly on rustc, and then have other things implemented on top of that. What has been hard, I think is that rules_rust didn't expose enough of an interface for people to build things like clippy or protobuf integration on top of it.

I 100% agree. Though, I feel there can be a clear separation of responsibility within the same git repository. I think adding this to the `cargo` directory is an appropriate thing to do if that's where we're going to put "Cargo-esque" things.

dfreese:
> > I've always wondered if it's worth having rules_rust focus explicitly on rustc, and then have other things implemented on top of that. What has been hard, I think is that rules_rust didn't expose enough of an interface for people to build things like clippy or protobuf integration on top of it.
> 
> I 100% agree. Though, I feel there can be a clear separation of responsibility within the same git repository. I think adding this to the `cargo` directory is an appropriate thing to do if that's where we're going to put "Cargo-esque" things.

True, but that makes versioning the project a little more tricky if the major sem-ver is attached to both the rust and the cargo, and particularly cargo-raze APIs.

UebelAndre:
> True, but that makes versioning the project a little more tricky if the major sem-ver is attached to both the rust and the cargo, and particularly cargo-raze APIs.

Given that only cargo-raze has any useful notion of versioning, I think that's something we can solve or set the standard for going forward fairly easily as it likely won't disrupt existing users. I personally don't think this is an issue and would imagine we tag releases with the thing actually being released (cargo or rustc rules). Or is that not a reasonable approach? I thought Google was known for monorepos, how is versioning handled internally?

acmcarther:
I haven't had the time to ~~do literally anything recently~~ review the comments PR, but driveby comments anyway. Sorry if this tramples on quality discussion that covers these points already.

This mirrors my response to an email to me from @hlopko that I felt should generally be public.

---

In general I'm against this PR, not because I have some nuanced vision for the future of rules_rust, but primarily because I don't like how cargo-raze deals with the cargo-bazel integration, and would like there to be a vacuum of sorts to leave room for something "better" to fill the space. I think having this moved into rules_rust (by varying degrees, depending on whether it's in the repo "proper", or just lives in a contrib directory, or something else) solidifies this tool as the canonical way of things. I'm not super excited about that possibility.

I think a large part of any compatibility concern can be mitigated by having rules_rust include some test suite that builds a toy application using cargo-raze.

UebelAndre:
I totally get that moving `cargo-raze` into this repo makes that tool the canonical way to interface with external crates, but I think that's a very realistic thing to do and something the rules should provide some solution for. If `cargo-raze` is not going to be a part of that, I'd love to hear proposals for what is. I think `cargo-raze` works fantastically and would be totally fine in the well defined world of "post https://github.com/bazelbuild/rules_rust/pull/596". But I don't quite care what the solution is (even though I've spent a considerable amount of time on cargo-raze), I just care that there's some solution that doesn't expect users to tediously hand author their WORKSPACE dependencies.

damienmg:
Hi,

Side-rant: for some reason github seems to hide some earlier comments :/

I also feel mixed about this PR. While I would love to see a canonical way to handle the cargo eco-system I also feel mixed with the current cargo raze approach which has problem scaling with monorepos (maintain 1000s of cargo files) and requires to write dependencies in a different language than the bazel one.

My view of the next step world would be where bazel is responsible for the download and compilation and maybe doing dependency resolution using a tool integrated with cargo (though I am pretty sure cargo graph resolution will blow up at some point). Cargo raze could be the tool doing the dependency resolution but I am not sure it is needed (could be using just the output of cargo instead).

The final world I see would probably require a custom dependency resolution tooling. I.e. I see a final usable states to have a set of depedency declared in a WORKSPACE file and pull their transitive deps the same way Cargo does but that requires also Bazel to get its story straight regarding external dependencies (and to be honest, I don't feel like there is any hope for a satisfying solution in the forseeable future).

hlopko:
(I wrote to the Google Open Source Program Office, will report when I hear back).

Ok let's try to move forward with the discussion. After some thinking I came to the following conclusions:
* If the decision was solely on me, I'd say no, reason being raze owners don't think raze is the best we can do
* But the decision is not solely on me, and even if I could (which I don't think I do) I wouldn't veto the PR moving raze to rules_rust
* But I will not champion the PR

Andre, if you can find a champion, and OSPO approves my proposal, feel free to move forward.

I'd recommend to:

* Create a new top level package for raze (cargo is there for cargo build script runner and it doesn't make sense to me to put both these into the same directory; I'd be fine with moving cargo build script runner elsewhere, just not both in the single package)
* Split the import into multiple PRs - 1) raw cp, no modifications 2) make it build in one or more PRs, 3) re-upload pending PRs from raze repo into rules_rust
* Ideally we want to maintain git history in the process
* Migrate raze issues into rules_rust issues; create a raze label

UebelAndre:
>     * Create a new top level package for raze (cargo is there for cargo build script runner and it doesn't make sense to me to put both these into the same directory; I'd be fine with moving cargo build script runner elsewhere, just not both in the single package)

I would strongly advocate for this going into the `cargo` package. Put it in a top level package solidifies it as the canonical way to interact with external crates. I see this as purely a cargo tool and I think the `cargo` package is the right place as a result. If there is a rule that has no reliance on cargo but can do the same task, it can go in it's own package

>     * Split the import into multiple PRs - 1) raw cp, no modifications 2) make it build in one or more PRs, 3) re-upload pending PRs from raze repo into rules_rust

There's currently only 1 change other than a straight merge and it's the `.bazelignore`. I'd be happy to do that in a separate PR people feel strongly that there should absolutely be no other changes in the PR.

>     * Ideally we want to maintain git history in the process

I believe this PR is doing that.

>     * Migrate raze issues into rules_rust issues; create a raze label

Makes sense but I also don't think it's a blocker for any merge.

> Andre, if you can find a champion, and OSPO approves my proposal, feel free to move forward.

Given that seemingly all "champions" are here and have said they don't like, it doesn't seem like this change is going to go through. Additionally, I have no idea how to interface with OSPO and don't think that's clear to anyone outside of Google.

I'm just going to close this because of the amount of push back it's getting. I'm happy to continue any conversations here or in some other form for things people would like to clarify or dig deeper into.

damienmg:
Thanks for opening that discussion and driving it @UebelAndre, just to clarify the position of most seems like we would like to see a better approach to handling crate than the one cargo raze offer. If you have a good idea on how to deal with the problem, that would be a very interesting discussion to drive.

tschuett:
rules_go has a different approach:
https://github.com/buchgr/bazel-remote/blob/3aa7b05f1657a647ed11f2104af4897c6cb65df8/WORKSPACE#L55

hlopko:
I heard back from the OSPO, they approved the move. Not relevant anymore here it seems but it can be useful in the future. 

+1 to what Damien said. Thank you @UebelAndre for pushing for this (and for all the other work you're doing for this repository!). It seems there is an opportunity to come up with a better solution. The first step would be to collect all the various (and maybe conflicting) requirements people have and come up with a solution that satisfies all. Maybe you can build on the momentum of this PR to collect more information and see if it would be interesting for you to drive the effort.

Thanks again!

Issue #591, opened 2021-02-17
Split rust_library into smaller rules


hlopko:
As socialized on https://groups.google.com/g/rules_rust/c/kGMg6haEF44, we're going to split `rust_library` into smaller rules. This is the tracking issue for the effort.

Steps:

- [x] Submit defs.bzl with rules in the desired end state
- [x] Regenerate docs
- [x] Provide buildozer migration script
- [x] Migrate all uses of `@rules_rust//rust:rust.bzl` in the repository to `@rules_rust//rust:defs.bzl`
- [x] Update raze and other closely related projects
- [x] Wait for the community to migrate
- [ ] Remove the deprecated rust.bzl file. Currently planned for 2021-10-26.

hlopko:
Buildozer migration scripts:

```bash
buildozer 'fix movePackageToTop' //...:__pkg__
buildozer 'substitute_load //rust:rust.bzl$ //rust:defs.bzl' //...:__pkg__

buildozer 'set kind rust_proc_macro' $(bazel query 'attr(crate_type, proc-macro, (//...))')
buildozer 'new_load @rules_rust//rust:defs.bzl rust_proc_macro' $(bazel query 'attr(crate_type, proc-macro, (//...))')

buildozer 'set kind rust_static_library' $(bazel query 'attr(crate_type, staticlib, (//...))')
buildozer 'new_load @rules_rust//rust:defs.bzl rust_static_library' $(bazel query 'attr(crate_type, staticlib, (//...))')

buildozer 'set kind rust_shared_library' $(bazel query 'attr(crate_type, cdylib, (//...))')
buildozer 'new_load @rules_rust//rust:defs.bzl rust_shared_library' $(bazel query 'attr(crate_type, cdylib, (//...))')

buildozer 'remove crate_type' $(bazel query 'attr(crate_type, .*, (//...))')

buildozer 'fix unusedLoads' //...:__pkg__
```

dae:
> FYI: Cargo-raze v0.13 is released and it produces BUILD files that are compatible with this breaking change.

Are you sure? I'm still seeing rust_library() in output build files, and can't find references to things like 'rust_static_library' in the cargo-raze sources. I see no associated PR, and only an open issue: https://github.com/google/cargo-raze/issues/398

PR #592, opened 2021-02-17
Split rust_library and add //rust:defs.bzl

hlopko:
This PR splits rust_library into multiple smaller rules:

* rust_library: rust_library will represent a non-transitive library
  similar to how cc_library, java_library and others behave. It will
  always provide CrateInfo, and depending on it will always mean you can
  access the crate from source. Once we support dynamic linking we can
  consider producing both rlib and dylib from rust_library, but let's
  leave that for another discussion.
* rust_static_library: this will only provide CcInfo and it will
  represent an archive of all transitive objects, both Rustc-made and
  native.
* rust_shared_library: this will only provide CcInfo and it will
  represent an shared library of all transitive objects, both Rustc-made
  and native.
* rust_proc_macro: similar to rust_library, but with different
  crate_type passed to rustc and different validation logic for its
  attributes.

I this this makes sense based on these observations:
* Right now rust_library covers all possible crate types except `bin`.
* rust_library always provides CrateInfo, rust_libraries can depend on
  other rust_libraries.
* When the crate type is `cdylib` or `staticlib`, rust_library provides
  CcInfo.
* When the crate type is `cdylib` or `staticlib`, Rust code will not be
  able to access the crate by `extern crate Foo` in the source; the
  behavior will be similar to depending on a CcInfo providing rule.

I believe smaller rules will make them less confusing and
will allow us to have more focused implementations.

This PR is mostly backwards compatible. //rust:rust.bzl#rust_library is
a macro. If the crate_type attribute is present, macro dispatches to
the right new rule. If it's not present, macro will choose the new
rust_library. New rules are added, so people can migrate at
their own pace.

defs.bzl is the bzl file that we now expect people to load. Bazel docs
recommend to store public rules in defs.bzl
(https://docs.bazel.build/versions/master/skylark/deploying.html#repository-content).

This change was first socialized at
https://groups.google.com/g/rules_rust/c/kGMg6haEF44.

Progress towards https://github.com/bazelbuild/rules_rust/issues/591.

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -31,9 +31,12 @@ PAGES = {
     "cargo_build_script": [
         "cargo_build_script",
     ],
-    "rust": [
-        "rust_library",
+    "defs": [
         "rust_binary",
+        "rust_library",
+        "rust_static_library",
+        "rust_shared_library",
+        "rust_proc_macro",
         "rust_benchmark",
         "rust_test",
     ],
--- a/docs/all.bzl
+++ b/docs/all.bzl
@@ -30,14 +30,7 @@ load(
     _rust_proto_toolchain = "rust_proto_toolchain",
 )
 load(
-    "@rules_rust//rust:repositories.bzl",
-    _rust_repositories = "rust_repositories",
-    _rust_repository_set = "rust_repository_set",
-    _rust_toolchain_repository = "rust_toolchain_repository",
-    _rust_toolchain_repository_proxy = "rust_toolchain_repository_proxy",
-)
-load(
-    "@rules_rust//rust:rust.bzl",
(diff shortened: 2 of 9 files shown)


hlopko:
I started https://github.com/bazelbuild/rules_rust/issues/591 to track the overall effort.

damienmg (approved):

UebelAndre:
Can you also make sure documentation is updated for this change?

hlopko:
Done. I think :)

UebelAndre (changes requested):
I think you'll need to fix the link to `defs` in `docs/index.md` I don't see that in this change.

UebelAndre (approved):
Nice! Thanks!

PR #593, opened 2021-02-19
Added `//rust/platform:macos` alias

UebelAndre:
Add alias for OSX to "macos" to match the modern name for the OS.

### DIFF
--- a/rust/platform/platform.bzl
+++ b/rust/platform/platform.bzl
@@ -81,6 +81,12 @@ def declare_config_settings():
         actual = ":darwin",
     )
 
+    # Add alias for OSX to "macos" to match the modern name for the OS.
+    native.alias(
+        name = "macos",
+        actual = ":darwin",
+    )
+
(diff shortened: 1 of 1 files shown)


rust/platform/platform.bzl
```diff
@@ -81,6 +81,12 @@ def declare_config_settings():
         actual = ":darwin",
     )
 
+    # Add alias for OSX to "macos" to match the modern name for the OS.
```
dfreese, line 84: can you add a note that this is consistent with the long-term direction of bazelbuild/platforms:

https://github.com/bazelbuild/platforms/commit/43155b81d40765f0d13008bc77cd2cca8ba9fb2a#diff-da1f06c09d1016c6686171f1b0c5d09cfec567fc7200c4c18699d27da5d4257c

dfreese (approved):

rust/platform/platform.bzl
UebelAndre: Updated

dfreese (approved):

PR #596, opened 2021-02-20
Add ARCHITECTURE.md

hlopko:
In this file I hope to describe how we think about the architecture
of `rules_rust`. It's goal is to help contributors orient themselves, and to
document code restrictions and assumptions.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,58 @@
+# Rules Rust - Architecture
+
+In this file we describe how we think about the architecture
+of `rules_rust`. It's goal is to help contributors orient themselves, and to
+document code restrictions and assumptions.
+
+In general we try to follow the common standard defined by
+https://docs.bazel.build/versions/master/skylark/deploying.html.
+
+## //rust
+
+This is the core package of the rules. It contains all the core rules such as
+`rust_binary` and `rust_library`. It also contains `rust_common`, a Starlark
+struct providing all rules_rust APIs that one might need when writing custom
+rules integrating with rules_rust.
+
+`//rust` and all its subpackages have to be standalone. Users who only need
+the core rules should be able to ignore all other packages.
+
+`//rust:defs.bzl` is the file that all users of `rules_rust` will be using.
+Everything in this file can be used (dependend on) and is supported  (though
+stability is not currently guaranteed across commits, see
+[#600](https://github.com/bazelbuild/rules_rust/issues/600)). Typically this
+file reexports definitions from other files, typically from `//rust/private`.
+Also other packages in `rules_rust` should access core rules through the public
+API only. We expect dogfooding our own APIs increases their
+quality.
+
+`//rust/private` package contains code rule implementation. This file can only
+depend on `//rust` and its subpackages. Exceptions are Bazel's builtin packages
+and public APIs of other rules (such as `rules_cc`). Depending on
+`//rust/private` from packages other than `//rust` is not supported, we reserve
+the right to change anything there and not tell anybody.
+
+When core rules need custom tools (such as process wrapper, launcher, test
+runner, and so on), they should be stored in `//rust/tools` (for public tools)
+or in `//rust/private/tools` (for private tools). These should:
+
+* be essential (it does not make sense to have core rules without them)
+* have few or no third party dependencies, and their third party dependencies
+    should be checked in.
+* be usable for cross compilation
+* have only essential requirements on the host system (requiring that a custom
+    library is installed on the system is frowned upon)
+
+## //examples (@examples)
+
+Examples package is actually a local repository. This repository can have
+additional dependencies on anything that helps demonstrate an example. Non
+trivial examples should have an accompanying README.md file.
+
+The goal of examples is to demonstrate usage of `rules_rust`, not to test code.
+Use `//test` for testing.
+
+## //test
+
+Contains unit (in `//test/unit`) and integration tests. CI configuration is
+stored in .bazelci.


hlopko:
Also r? @UebelAndre :)

hlopko:
Comments and contributions are very welcome! This is just a first draft. I haven't written anything about many packages such as proto or bindgen.

UebelAndre:
😄 Thanks for starting this! I think it looks good so far.

Does this mean you're looking to move `./util` into `./rust/tools`?

I'm also particularly curious about `cargo`. I think I like it as it's own package but I can also see it as part of `rust/tools`. Your thoughts?

dfreese (approved):

dfreese:
Looks like a solid start.  Thanks for putting that together.

ARCHITECTURE.md
```diff
@@ -0,0 +1,57 @@
...
+depend on `//rust` and its subpackages. Exceptions are Bazel's builtin packages
+and public APIs of other rules (such as `rules_cc`). Depending on
+`//rust/private` from packages other than `//rust` is not supported, we reserve
+the right to change anything there and not tell anybody.
+
+When core rules need custom tools (such as process wrapper, launcher, test
+runner, and so on), they should be stored in `//rust/tools` (for public tools)
+or in `//rust/private/tools` (for private tools). These should:
+
+* essential (tools that are not essential can be used conditionally for example
```
illicitonion, line 37: I'm not quite sure what this bullet point is trying to tell me?

ARCHITECTURE.md
```diff
@@ -0,0 +1,57 @@
...
+This is the core package of the rules. It contains all the core rules such as
+`rust_binary` and `rust_library`. It also contains `rust_common`, a Starlark
+struct providing all rules_rust APIs that one might need when writing custom
+rules integrating with rules_rust.
+
+`//rust` and all its subpackages have to be standalone. Users who only need
+the core rules should be able to ignore all other packages.
+
+`//rust:defs.bzl` is the file that all users of `rules_rust` will be using.
+Everything in this file can be used (dependend on) and is supported. Typically
```
illicitonion, line 21: This somewhat raises the question of compatibility guarantees - "can be used and is supported"... Until when?

illicitonion (approved):
Looks good!

hlopko:
> Does this mean you're looking to move ./util into ./rust/tools?

Yup (or //rust/private/tools, depending on the tool, I haven't thought about where tools should go).

> I'm also particularly curious about cargo. I think I like it as it's own package but I can also see it as part of rust/tools. Your thoughts?

If you're asking about cargo build script runner, I'm open to suggestions as I haven't looked into how it is used and how the rule using it works. If there's not strong reason why it should be in //rust, I'd vote to not put it there.

ARCHITECTURE.md
hlopko: Rewrote to say `be essential (it does not make sense to have core rules without them)`. What I was trying to say before was that when core rules can be used both with and without the tool, the tool is not essential and therefore should probably not be bundled in the core rules.

ARCHITECTURE.md
hlopko: That's a hard question which I'd like to find answer to in a separate discussion. Would you prefer uncertainty of the current text, or deleting it?

hlopko:
So right now we say //rust is core package, everything else is less so. One alternative is to have a //contrib package that would contain all the "less so" packages. Opinions?

UebelAndre:
> If you're asking about cargo build script runner, I'm open to suggestions as I haven't looked into how it is used and how the rule using it works. If there's not strong reason why it should be in //rust, I'd vote to not put it there.

To my knowledge, the build script runner is there to run [`build.rs` scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html) which is a concept in Cargo, not rustc. So I think it makes more sense to leave it in the `cargo` directory. I think I like the existence of a `cargo` directory because it can act as the separation of `cargo` and `rustc`. Notably, https://github.com/bazelbuild/rules_rust/pull/598 and https://github.com/bazelbuild/rules_rust/pull/590 are also something I feel should go into `cargo` since they build on cargo manifests to generate Bazel repository definitions.

ARCHITECTURE.md
illicitonion: I'd suggest either adding an explicit qualifier ("though stability is not currently guaranteed across commits", and a link to an issue), or removing it.

damienmg (approved):

damienmg:
> > If you're asking about cargo build script runner, I'm open to suggestions as I haven't looked into how it is used and how the rule using it works. If there's not strong reason why it should be in //rust, I'd vote to not put it there.
> 
> To my knowledge, the build script runner is there to run [`build.rs` scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html) which is a concept in Cargo, not rustc. So I think it makes more sense to leave it in the `cargo` directory. I think I like the existence of a `cargo` directory because it can act as the separation of `cargo` and `rustc`. Notably, #598 and #590 are also something I feel should go into `cargo` since they build on cargo manifests to generate Bazel repository definitions.

Yes `cargo_build_script` is named like this and in a different package because it is an interface with cargo not with the rust toolchain.

ARCHITECTURE.md
hlopko: Good point. Done.

hlopko:
I've created https://github.com/bazelbuild/rules_rust/issues/601 and https://github.com/bazelbuild/rules_rust/issues/601 to track cleanups relevant to this PR.

PR #597, opened 2021-02-21
Add Missing link definition in README.md

regisd:
Fix markdown syntax for the Rust programming language.

The link was using a link definition without link.
Since all other links use inline style, this commit also follows this style.


illicitonion (approved):
Thanks!

PR #603, opened 2021-02-23
rules_rust: make it possible to use non-staticlib rust_library targets in c++ deps

durin42:
ld can handle .rlib archives, since they're really just standard ar(1)
archives like a .a file with some extra metadata. As a result, we can avoid
using staticlib crates when building mixed C++/Rust binaries.

This doesn't correctly handle libstd et al (yet), or deal with some generated
symbols that rustc produces when it drives the linking step, but it's a start.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -667,6 +667,7 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
             feature_configuration = feature_configuration,
             cc_toolchain = cc_toolchain,
             static_library = dot_a,
+            pic_static_library = dot_a,
         )
     elif crate_info.type == "cdylib":
         library_to_link = cc_common.create_library_to_link(
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -632,7 +632,7 @@ def rustc_compile_action(
     ]
 
 def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration):
-    """If the produced crate is suitable yield a CcInfo to allow for interop with cc rules 
+    """If the produced crate is suitable yield a CcInfo to allow for interop with cc rules
 
     Args:
         ctx (ctx): The rule's context object
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -645,7 +645,7 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
         list: A list containing the CcInfo provider
     """
 
-    if crate_info.is_test or crate_info.type not in ("staticlib", "cdylib") or getattr(ctx.attr, "out_binary", False):
+    if crate_info.is_test or crate_info.type not in ("staticlib", "cdylib", "rlib", "lib") or getattr(ctx.attr, "out_binary", False):
         return []
 
     if crate_info.type == "staticlib":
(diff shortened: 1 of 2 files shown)


rust/private/rustc.bzl
```diff
@@ -655,6 +655,18 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
             cc_toolchain = cc_toolchain,
             static_library = crate_info.output,
         )
+    elif crate_info.type in ("rlib", "lib"):
+        # bazel hard-codes a check for endswith((".a", ".pic.a",
+        # ".lib")) in create_library_to_link, so we work around that
+        # by creating a symlink to the .rlib with a .a extension.
+        dot_a = ctx.actions.declare_file(crate_info.name + ".a", sibling = crate_info.output)
+        ctx.actions.symlink(output = dot_a, target_file = crate_info.output)
```
UebelAndre, line 663: Is symlink actually safe to use for all platforms? Will windows run into any issues?

dae: Not sure about the core rules, but the cargo build script code requires runfiles already: https://docs.bazel.build/versions/master/windows.html#enable-symlink-support

hlopko: Symlink is safe to use:

ctx.actions.symlink() eventually calls https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/windows/WindowsFileSystem.java;l=87
If the target is directory, it creates a junction on Windows, which works just like symlink
If the target is a file, it copies the target if file symlink is not enabled on your machine, otherwise it creates an actual symlink (https://docs.bazel.build/versions/4.0.0/windows.html#enable-symlink-support)

rust/private/rustc.bzl
```diff
@@ -655,6 +655,18 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
...
+        library_to_link = cc_common.create_library_to_link(
+            actions = ctx.actions,
+            feature_configuration = feature_configuration,
+            cc_toolchain = cc_toolchain,
+            static_library = dot_a,
```
hlopko, line 668: I think rustc produces PIC archives by default, so we could set pic_static_library to dot_a as well. Please add a todo for me to fix it to handle NOPIC correctly.

rust/private/rustc.bzl
```diff
@@ -645,7 +645,7 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
         list: A list containing the CcInfo provider
     """
 
-    if crate_info.is_test or crate_info.type not in ("staticlib", "cdylib") or getattr(ctx.attr, "out_binary", False):
+    if crate_info.is_test or crate_info.type not in ("staticlib", "cdylib", "rlib", "lib") or getattr(ctx.attr, "out_binary", False):
```
hlopko, line 648: Could I bother you with updating docs in //rust/private:rust.bzl in `rust_static_library`, `rust_shared_library` and `rust_library`? For first two I'd add  to "This rule provides CcInfo, so it can be used everywhere Bazel expects rules_cc" that the produced library contains core and stdlib, and it uses either system allocator or the one from rust sources.

For rust_library, I'd say that it provides CcInfo, but care needs to be taken to depend on stdlib/allocator correctly.

durin42: Done - first two in an earlier commit, last one in this commit.

rust/private/rustc.bzl
durin42: I think I did what you asked here. PTAL?

rust/private/rustc.bzl
durin42: AFAICT we're good here.

rust/private/rustc.bzl
hlopko: Yes, thank you!

hlopko (approved):

PR #606, opened 2021-02-25
rustc: correctly handle alwayslink staticlibs

### LINKED ISSUE #325: Linking fails for cc_library targets with "alwayslink = 1"
I ran into a situation with grpc's use of c-ares, where rust_test or binary would fail to link in the appropriate object files.  This was related to the `alwayslink = 1` attribute set on the ares cc_library.

The simplest example I could reduce this to is below (which is also on my [alwayslink branch](https://github.com/dfreese/rules_rust/tree/alwayslink/test/alwayslink))

BUILD
```python
load("//rust:rust.bzl", "rust_library", "rust_test")

cc_library(
    name = "linked",
    srcs = [
        "linked.cc",
    ],
    # Compilation will fail if this is commented out.
    # alwayslink = 1,
)

rust_library(
    name = "use_linked",
    srcs = ["use_linked.rs"],
    deps = [
        ":linked",
    ],
)

rust_test(
    name = "use_linked_test",
    crate = ":use_linked",
)
```
linked.cc
```cpp
extern "C" {
int cc_value() { return 42; }
}
```

use_linked.rs
```rust
extern "C" {
    pub fn cc_value() -> ::std::os::raw::c_int;
}

pub fn value() -> i32 {
    unsafe { cc_value() }
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn value_works() {
        assert_eq!(value(), 42);
    }
}
```

I haven't had a chance to look into it much, but it appears to be a result of the way linker arguments are constructed.  For example, the `if l.basename.endswith(toolchain.staticlib_ext)` [here](https://github.com/bazelbuild/rules_rust/blob/master/rust/private/rustc.bzl#L143) seems to filter out any mention of `linked` completely from the linker invocation.  Changing that to `if not l.basename.endswith(toolchain.dylib_ext)]` causes an `-llinked` linker argument to show up, but this isn't entirely correct, as there should be a reference to linked.cc's object file.

durin42:
In order to do this, we have to inline get_libs_for_static_executable so we
can interact with the LibraryToLink entity as we're categorizing staticlibs.

Fixes #325.

### DIFF
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -16,7 +16,20 @@
 load("//rust:rust.bzl", "rust_library")
 
 # buildifier: disable=bzl-visibility
-load("//rust/private:utils.bzl", "find_toolchain", "get_libs_for_static_executable")
+load("//rust/private:utils.bzl", "find_toolchain", "get_preferred_artifact")
+
+# TODO(hlopko): use the more robust logic you're from rustc.bzl also here, through a reasonable API.
+def _get_libs_for_static_executable(dep):
+    """find the libraries used for linking a static executable.
+
+    Args:
+        dep (Target): A cc_library target.
+
+    Returns:
+        depset: A depset[File]
+    """
+    linker_inputs = dep[CcInfo].linking_context.linker_inputs.to_list()
+    return depset([get_preferred_artifact(lib) for li in linker_inputs for lib in li.libraries])
 
 def rust_bindgen_library(
         name,
@@ -84,7 +97,7 @@ def _rust_bindgen_impl(ctx):
     output = ctx.outputs.out
 
     # libclang should only have 1 output file
-    libclang_dir = get_libs_for_static_executable(libclang).to_list()[0].dirname
+    libclang_dir = _get_libs_for_static_executable(libclang).to_list()[0].dirname
     include_directories = cc_lib[CcInfo].compilation_context.includes.to_list()
     quote_include_directories = cc_lib[CcInfo].compilation_context.quote_includes.to_list()
     system_include_directories = cc_lib[CcInfo].compilation_context.system_includes.to_list()
@@ -112,7 +125,7 @@ def _rust_bindgen_impl(ctx):
     }
 
     if libstdcxx:
-        env["LD_LIBRARY_PATH"] = ":".join([f.dirname for f in get_libs_for_static_executable(libstdcxx).to_list()])
+        env["LD_LIBRARY_PATH"] = ":".join([f.dirname for f in _get_libs_for_static_executable(libstdcxx).to_list()])
 
     ctx.actions.run(
         executable = bindgen_bin,
@@ -120,9 +133,9 @@ def _rust_bindgen_impl(ctx):
             [header],
             transitive = [
                 cc_lib[CcInfo].compilation_context.headers,
-                get_libs_for_static_executable(libclang),
+                _get_libs_for_static_executable(libclang),
             ] + [
-                get_libs_for_static_executable(libstdcxx),
+                _get_libs_for_static_executable(libstdcxx),
             ] if libstdcxx else [],
         ),
         outputs = [unformatted_output],
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -18,7 +18,7 @@ load("//rust:rust.bzl", "rust_library")
 # buildifier: disable=bzl-visibility
 load("//rust/private:utils.bzl", "find_toolchain", "get_preferred_artifact")
 
-# TODO(hlopko): use the more robust logic you're from rustc.bzl also here, through a reasonable API.
+# TODO(hlopko): use the more robust logic from rustc.bzl also here, through a reasonable API.
 def _get_libs_for_static_executable(dep):
     """find the libraries used for linking a static executable.
 
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -600,7 +600,7 @@ def rustc_compile_action(
         ),
     )
 
-    dylibs = [get_preferred_artifact(lib) for lib in dep_info.transitive_noncrates.to_list() if _is_dylib(toolchain, lib)]
+    dylibs = [get_preferred_artifact(lib) for lib in dep_info.transitive_noncrates.to_list() if _is_dylib(lib)]
 
     runfiles = ctx.runfiles(
         files = dylibs + getattr(ctx.files, "data", []),
@@ -622,9 +622,8 @@ def rustc_compile_action(
         ),
     ]
 
-def _is_dylib(toolchain, dep):
-    lib = get_preferred_artifact(dep)
-    return lib.basename.endswith(toolchain.dylib_ext) or lib.basename.split(".", 2)[1] == toolchain.dylib_ext[1:]
+def _is_dylib(dep):
+    return not bool(dep.static_library or dep.pic_static_library)
 
 def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration):
     """If the produced crate is suitable yield a CcInfo to allow for interop with cc rules
@@ -749,7 +748,12 @@ def _compute_rpaths(toolchain, output_dir, dep_info):
     Returns:
         depset: A set of relative paths from the output directory to each dependency
     """
-    dylibs = [get_preferred_artifact(lib) for lib in dep_info.transitive_noncrates.to_list() if _is_dylib(toolchain, lib)]
+    preferreds = [get_preferred_artifact(lib) for lib in dep_info.transitive_noncrates.to_list()]
+
+    # TODO(augie): I don't understand why this can't just be filtering on
+    # _is_dylib(lib), but doing that causes failures on darwin and windows
+    # examples that otherwise work.
+    dylibs = [lib for lib in preferreds if lib.basename.endswith(toolchain.dylib_ext) or lib.basename.split(".", 2)[1] == toolchain.dylib_ext[1:]]
     if not dylibs:
         return depset([])
     if toolchain.os != "linux":
@@ -824,7 +828,7 @@ def _make_link_flags(toolchain, lib):
         pass  # TODO
     elif lib.static_library or lib.pic_static_library:
         args.append("-lstatic=%s" % get_lib_name(f))
-    elif _is_dylib(toolchain, lib):
+    elif _is_dylib(lib):
         args.append("-ldylib=%s" % get_lib_name(f))
     return args
 
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -54,9 +54,8 @@ DepInfo = provider(
         "direct_crates": "depset[CrateInfo]",
         "transitive_build_infos": "depset[BuildInfo]",
         "transitive_crates": "depset[CrateInfo]",
-        "transitive_dylibs": "depset[File]",
         "transitive_libs": "List[File]: All transitive dependencies, not filtered by type.",
-        "transitive_staticlibs": "depset[File]",
+        "transitive_noncrates": "depset[LibraryToLink]: All transitive dependencies that aren't crates.",
     },
 )
 
@@ -149,8 +148,8 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
 
     direct_crates = []
     transitive_crates = []
-    transitive_dylibs = []
-    transitive_staticlibs = []
+    transitive_noncrates = []
+    transitive_noncrate_libs = []
     transitive_build_infos = []
     build_info = None
 
@@ -165,27 +164,17 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
             ))
 
             transitive_crates.append(depset([dep[rust_common.crate_info]], transitive = [dep[DepInfo].transitive_crates]))
-            transitive_dylibs.append(dep[DepInfo].transitive_dylibs)
-            transitive_staticlibs.append(dep[DepInfo].transitive_staticlibs)
+            transitive_noncrates.append(dep[DepInfo].transitive_noncrates)
+            transitive_noncrate_libs.append(depset(dep[DepInfo].transitive_libs))
             transitive_build_infos.append(dep[DepInfo].transitive_build_infos)
         elif CcInfo in dep:
             # This dependency is a cc_library
 
             # TODO: We could let the user choose how to link, instead of always preferring to link static libraries.
             linker_inputs = dep[CcInfo].linking_context.linker_inputs.to_list()
-            libs = depset([get_preferred_artifact(lib) for li in linker_inputs for lib in li.libraries])
-
-            transitive_dylibs.append(depset([
-                lib
-                for lib in libs.to_list()
-                # Dynamic libraries may have a version number nowhere, or before (macos) or after (linux) the extension.
-                if lib.basename.endswith(toolchain.dylib_ext) or lib.basename.split(".", 2)[1] == toolchain.dylib_ext[1:]
-            ]))
-            transitive_staticlibs.append(depset([
-                lib
-                for lib in libs.to_list()
-                if lib.basename.endswith(toolchain.staticlib_ext)
-            ]))
+            libs = [get_preferred_artifact(lib) for li in linker_inputs for lib in li.libraries]
+            transitive_noncrate_libs.append(depset(libs))
+            transitive_noncrates.append(depset([lib for linker_input in linker_inputs for lib in linker_input.libraries]))
         elif BuildInfo in dep:
             if build_info:
                 fail("Several deps are providing build information, only one is allowed in the dependencies", "deps")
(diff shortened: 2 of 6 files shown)


rust/private/rustc.bzl
```diff
@@ -167,24 +169,29 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
...
-                for lib in libs.to_list()
-                # Dynamic libraries may have a version number nowhere, or before (macos) or after (linux) the extension.
-                if lib.basename.endswith(toolchain.dylib_ext) or lib.basename.split(".", 2)[1] == toolchain.dylib_ext[1:]
-            ]))
-            transitive_staticlibs.append(depset([
-                lib
-                for lib in libs.to_list()
-                if lib.basename.endswith(toolchain.staticlib_ext)
-            ]))
+            linker_inputs = dep[CcInfo].linking_context.linker_inputs.to_list()
```
hlopko, line 178: Could you inline get_libs_for_static_executable into other callsites and delete the function ? And while there, review if other callsites should be fixed for alwayslink too.

rust/private/rustc.bzl
```diff
@@ -167,24 +169,29 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
...
+            for li in linker_inputs:
+                always = []
+                static = []
+                dylib = []
+                for lib in li.libraries:
+                    art = get_preferred_artifact(lib)
+                    if lib.alwayslink:
+                        always.append(lib.static_library or lib.pic_static_library)
+                        # Dynamic libraries may have a version number nowhere, or before (macos) or after (linux) the extension.
```
hlopko, line 187: dedent the comment to match the line it relates to

rust/private/rustc.bzl
```diff
@@ -167,24 +169,29 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
...
-                if lib.basename.endswith(toolchain.dylib_ext) or lib.basename.split(".", 2)[1] == toolchain.dylib_ext[1:]
-            ]))
-            transitive_staticlibs.append(depset([
-                lib
-                for lib in libs.to_list()
-                if lib.basename.endswith(toolchain.staticlib_ext)
-            ]))
+            linker_inputs = dep[CcInfo].linking_context.linker_inputs.to_list()
+            for li in linker_inputs:
+                always = []
```
hlopko, line 180: This will break order of linker flags. Ideally we want `-lfoo --whole-archive -lbar --no-whole-archive -lbaz`, but in this PR we get `-lfoo  -lbaz --whole-archive -lbar --no-whole-archive`. We need to have a single depset for both alwayslink and non-alwayslink libraries and use the right flags when adding to args.

rust/private/rustc.bzl
```diff
@@ -822,6 +830,20 @@ def _add_native_link_flags(args, dep_info, crate_type, cc_toolchain, feature_con
     args.add_all(dep_info.transitive_dylibs, map_each = get_lib_name, format_each = "-ldylib=%s")
     args.add_all(dep_info.transitive_staticlibs, map_each = get_lib_name, format_each = "-lstatic=%s")
 
+    # alwayslink=True libraries need special handling so the alwayslink args
+    # make it to the linker. Eventually we might be able to use a first-class
+    # rustc flag for this (https://github.com/rust-lang/cargo/issues/7586),
+    # but until then we have to do it this way.
+    for lib in dep_info.transitive_staticlibs_alwayslink.to_list():
```
hlopko, line 837: I know .to_list() is called in way too many places in this repo, but that's something we will have to fix. Could you instead extract a function that produces the right flags and use args.add_all(libs, map_each=...)? That would defer the depset flattening till the action execution phase thus speed up analysis time and decrease memory use. Thank you!

hlopko (changes requested):
Looks good, only some tweaks missing. Could you add a unit test?

rust/private/rustc.bzl
durin42: Done.

rust/private/rustc.bzl
durin42: This is now (mercifully) gone.

rust/private/rustc.bzl
durin42: Well that was a neat refactor.

rust/private/rustc.bzl
durin42: As far as I can starlark, this can't be done: the map_each function needs information about the OS in play or it can't specify the right format linker flag, and there's no way to set up a partial in starlark. It also looks like there's nothing like __call__ from Python here, so I've tried to structure things as similarly to what you've asked for here so that if I'm wrong it's an easy follow-up but for now this isn't technically done.

durin42:
@hlopko I'm closer, all the tests pass, but it looks like I broke something that's only covered by examples and I'm not running those locally? Is that right?

(If so, what's the right way to run the examples? I'm not figuring it out...)

illicitonion:
> @hlopko I'm closer, all the tests pass, but it looks like I broke something that's only covered by examples and I'm not running those locally? Is that right?
> 
> (If so, what's the right way to run the examples? I'm not figuring it out...)

```shell
$ bazel build @examples//...
```

(Some don't run on non-Linux platforms, see https://github.com/bazelbuild/rules_rust/blob/main/.bazelci/presubmit.yml for a list of exceptions)

durin42:
@illicitonion thanks! Found my bug. :)

illicitonion:
> @illicitonion thanks! Found my bug. :)

Hurrah!

(Hi, by the way! Long time no see!)

durin42:
Hi! It has been a while. Somehow unsurprising this is a corner I'd find
you... :)

rust/private/rustc.bzl
hlopko: You could have 3 different functions and choose one of them depending on the OS when constructing args, right? Or am I missing something?

hlopko:
> @hlopko I'm closer, all the tests pass, but it looks like I broke something that's only covered by examples and I'm not running those locally? Is that right?
> 
> (If so, what's the right way to run the examples? I'm not figuring it out...)

You can run examples by `bazel test @examples//...` (or cd into examples directory and `bazel test //...`)

rust/private/rustc.bzl
durin42: Took a wee bit more refactoring than that, but it's fixed.

durin42:
Alright, this now breaks @examples//ffi/rust_calling_c:matrix_dynamically_linked on windows and darwin, but I'm honestly a little confused if that's my fault. I _think_ this is something that had to be broken all along and I'm merely exposing it?

durin42:
Okay, those broke in 3ad0e99b1ea60c2a5ec2d3facf79ad0bf6dcec90, but it's not super obvious to me that that change is wrong. :(

durin42:
Did some other stuff, realized I could just "bake in" the old behavior in _compute_rpaths() with a TODO, so that's what  did.

rust/private/rustc.bzl
```diff
@@ -754,18 +748,24 @@ def _compute_rpaths(toolchain, output_dir, dep_info):
     Returns:
         depset: A set of relative paths from the output directory to each dependency
     """
-    if not dep_info.transitive_dylibs:
+    preferreds = [get_preferred_artifact(lib) for lib in dep_info.transitive_noncrates.to_list()]
+
+    # TODO(augie): I don't understand why this can't just be filtering on
```
hlopko, line 753: Captain Obvious speaking, but you could just print(dylibs) and build/run the failing example and compare the content?

rust/private/rustc.bzl
```diff
@@ -754,18 +748,24 @@ def _compute_rpaths(toolchain, output_dir, dep_info):
...
+    # _is_dylib(lib), but doing that causes failures on darwin and windows
```
hlopko, line 754: Or you mean that _is_dylib finds dylibs on mac/windows, and the other thing doesn't, and therefore with _is_dylib we trigger `fail` on line 760?

bindgen/bindgen.bzl
```diff
@@ -16,7 +16,19 @@
 load("//rust:rust.bzl", "rust_library")
 
 # buildifier: disable=bzl-visibility
-load("//rust/private:utils.bzl", "find_toolchain", "get_libs_for_static_executable")
+load("//rust/private:utils.bzl", "find_toolchain", "get_preferred_artifact")
+
+def _get_libs_for_static_executable(dep):
```
hlopko, line 21: Pls add a todo for me to use the more robust logic you're adding in rustc.bzl also here, through a reasonable API.

rust/private/rustdoc_test.bzl
```diff
@@ -99,10 +99,15 @@ def _build_rustdoc_flags(dep_info, crate):
     link_flags += ["--extern=" + c.name + "=" + c.dep.output.short_path for c in d.direct_crates.to_list()]
     link_search_flags += ["-Ldependency={}".format(_dirname(c.output.short_path)) for c in d.transitive_crates.to_list()]
 
-    link_flags += ["-ldylib=" + get_lib_name(lib) for lib in d.transitive_dylibs.to_list()]
-    link_search_flags += ["-Lnative={}".format(_dirname(lib.short_path)) for lib in d.transitive_dylibs.to_list()]
-    link_flags += ["-lstatic=" + get_lib_name(lib) for lib in d.transitive_staticlibs.to_list()]
-    link_search_flags += ["-Lnative={}".format(_dirname(lib.short_path)) for lib in d.transitive_staticlibs.to_list()]
+    for lib_to_link in dep_info.transitive_noncrates.to_list():
```
hlopko, line 102: Pls add a todo for me to use the more robust logic you're adding in rustc.bzl also here, through a reasonable API.

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -103,11 +103,44 @@ def _bin_has_native_libs_test_impl(ctx):
     _assert_argv_contains(env, action, "-lstatic=native_dep")
     return analysistest.end(env)
 
+def _extract_linker_args(argv):
+    return [a for a in argv if a.startswith("link-arg=-Wl") or a.startswith("link-arg=-l") or a.startswith("-l") or a.endswith(".lo") or a.endswith(".o")]
+
+def _bin_has_native_dep_and_alwayslink_test_impl(ctx):
```
hlopko, line 109: Could you also add a test for staticlib and cdylib?

rust/private/rustc.bzl
durin42: Yeah, I tried that and it didn't make any sense.

rust/private/rustc.bzl
durin42: Yes. It's super weird, but given that doing it this way (which seems wrong) prevents any regressions I'm willing to live with that for now in the name of making progress.

durin42: (Assuming you are, anyway.)

rust/private/rustc.bzl
```diff
@@ -631,6 +622,9 @@ def rustc_compile_action(
         ),
     ]
 
+def _is_dylib(dep):
+    return not bool(dep.static_library or dep.pic_static_library)
```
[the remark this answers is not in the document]
durin42: I honestly have no idea. In practice, this seemed to produce the right linker invocation and made intuitive sense - other than the weird runtime linking thing below this seems strictly like an improvement.

rust/private/rustc.bzl
```diff
@@ -54,9 +54,8 @@ DepInfo = provider(
         "direct_crates": "depset[CrateInfo]",
         "transitive_build_infos": "depset[BuildInfo]",
         "transitive_crates": "depset[CrateInfo]",
-        "transitive_dylibs": "depset[File]",
         "transitive_libs": "List[File]: All transitive dependencies, not filtered by type.",
-        "transitive_staticlibs": "depset[File]",
+        "transitive_noncrates": "depset[LibraryToLink]: All transitive dependencies that aren't crates.",
```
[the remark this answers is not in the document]
durin42: I'm not sure what line you're commenting on: transitive_noncrates is a depset[LibraryToLink] now?

rust/private/rustc.bzl
```diff
@@ -165,26 +164,17 @@ def collect_deps(label, deps, proc_macro_deps, aliases, toolchain):
...
-            ]))
-            transitive_staticlibs.append(depset([
-                lib
-                for lib in libs.to_list()
-                if lib.basename.endswith(toolchain.staticlib_ext)
-            ]))
+            linker_inputs = dep[CcInfo].linking_context.linker_inputs.to_list()
+            libs = [get_preferred_artifact(lib) for li in linker_inputs for lib in li.libraries]
+            transitive_noncrate_libs.append(depset(libs))
+            transitive_noncrates.append(depset([lib for linker_input in dep[CcInfo].linking_context.linker_inputs.to_list() for lib in linker_input.libraries]))
```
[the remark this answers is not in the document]
durin42: Done. Good catch!

test/unit/native_deps/native_deps_test.bzl
durin42: That is, a staticlib and a cdylib target that have alwayslink deps?

hlopko: Correct

hlopko: Hm, what does it mean to produce a staticlib with an alwayslink dep? Some objects in deps need to be linked as alwayslink, and we put those into one big archive that staticlib is. We can't mark specific objects in the archive as alwayslink.

It seems we only have 2 options:
1) assume that alwayslink deps are marked alwayslink because there is a circular dependency with some other target that is also covered by the staticlib archive. In this case we don't have to do anything, the circular dependency is taken care of by all objects being in the same archive.
2) assume that alwayslink deps are marked alwayslink because they contain static initialization code that needs to make it into the final artifact. The only immediately available solution I can think of is to mark the whole staticlib as alwayslink. That's going to increase the binary size by a lot.

I propose to go with 1), create an issue describing the problem and add todos referencing the issue, and call it a day.

For cdylib I think our current implementation still makes sense.

rust/private/rustc.bzl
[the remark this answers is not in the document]
hlopko: I'm suggesting to make `transitive_nocrates` be `depset[LinkerInput]`, so you don't have to call `.to_list` on line 174 (`dep[CcInfo].linking_context.linker_inputs.to_list()`.

rust/private/rustc.bzl
hlopko: Ok, this PR is already touching many sensitive pieces, let's defer fixing this into a followup PR.

rust/private/rustc.bzl
hlopko: Ok, this PR is already touching many sensitive pieces, let's defer fixing this into a followup PR.

rust/private/rustc.bzl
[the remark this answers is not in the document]
hlopko: I'll take a look. Feel free to move forward with the PR.

rust/private/rustc.bzl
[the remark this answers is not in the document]
durin42: Got it. Done.

test/unit/native_deps/native_deps_test.bzl
durin42: cdylib test added. I'm not sure I follow the staticlib case, but can you file the bug if you want to track it?

hlopko (approved):
Let's get this moving :)

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -103,11 +103,75 @@ def _bin_has_native_libs_test_impl(ctx):
...
 rlib_has_no_native_libs_test = analysistest.make(_rlib_has_no_native_libs_test_impl)
 staticlib_has_native_libs_test = analysistest.make(_staticlib_has_native_libs_test_impl)
 cdylib_has_native_libs_test = analysistest.make(_cdylib_has_native_libs_test_impl)
 proc_macro_has_native_libs_test = analysistest.make(_proc_macro_has_native_libs_test_impl)
 bin_has_native_libs_test = analysistest.make(_bin_has_native_libs_test_impl)
+bin_has_native_dep_and_alwayslink_test = analysistest.make(_bin_has_native_dep_and_alwayslink_test_impl, attrs = {
+    "_macos_constraint": attr.label(default = Label("@platforms//os:macos")),
+    "_windows_constraint": attr.label(default = Label("@platforms//os:windows")),
+})
+cdylib_has_native_dep_and_alwayslink_test = analysistest.make(_cdylib_has_native_libs_test_impl, attrs = {
```
UebelAndre, line 171: Looks like this this function is using the wrong implementation and correcting this causes a failure. @hlopko @durin42 would either of you be able to take a quick look? 🙏 
```suggestion
cdylib_has_native_dep_and_alwayslink_test = analysistest.make(_cdylib_has_native_dep_and_alwayslink_test_impl, attrs = {
```

PR #607, opened 2021-02-25
Add -L but not -l to rlib compilations

illicitonion:
Build scripts may add -l flags to dependees, and without the relevant -L
flags those -l flags may fail to resolve, causing build failures.

### DIFF
--- a/test/unit/native_deps/native_deps_test.bzl
+++ b/test/unit/native_deps/native_deps_test.bzl
@@ -1,6 +1,6 @@
 """Unittests for rust rules."""
 
-load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts", "unittest")
 load("@rules_cc//cc:defs.bzl", "cc_library")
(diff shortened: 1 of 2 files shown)


illicitonion:
https://github.com/bazelbuild/rules_rust/pull/576 removed both the `-L` and `-l` flags, this just re-adds the `-L` ones.

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -18,12 +18,26 @@ def _assert_argv_contains(env, action, flag):
         "Expected {args} to contain {flag}".format(args = action.argv, flag = flag),
     )
 
+def _assert_argv_contains_prefix_suffix(env, action, prefix, suffix):
```
hlopko, line 21: Could you also update all other tests? AFAIU all of create types should get -Lnative flags, right?

hlopko (approved):
LGTM (with updating the tests), thank you for re-fixing this :)

test/unit/native_deps/native_deps_test.bzl
illicitonion: Indeed! Done :)

PR #609, opened 2021-02-25
Removed test launcher's install script in favor of ctx.actions.symlink

UebelAndre:
I found out that [ctx.actions.symlink](https://docs.bazel.build/versions/master/skylark/lib/actions.html#symlink) is safe to use on all platforms 😄 

Cleaning up some cruft from the initial implementation here.


hlopko:
Oh I love merging these kinds of PRs! :) Thank you!

hlopko (approved):

PR #610, opened 2021-02-25
Add -Lnative flags for C++ runtime library

hlopko:
This should have been a part of https://github.com/bazelbuild/rules_rust/pull/576.


illicitonion (approved):
Any tests? :)

hlopko:
Yeah tests for this are one big yak shave for which I'm not sure I'm prepared for mentally :))

It would involve setting up a custom C++ toolchain that uses static_runtime_lib and actually work on our (at least) Linux CI workers. I'd reuse Bazel's autoconfiguration logic, but I'd need to be able to be able to customize it with the runtime lib, which needs Bazel PRs. And then writing a rule that would transition to this custom toolchain for the unit test.

PR #612, opened 2021-02-25
Increase CI coverage

hlopko:
I realized we don't run tests on Windows. This PR fixes that. While there I also enabled tests on RBE.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -57,6 +57,7 @@ tasks:
     - "..."
     - "-//bindgen/..."
     - "-//test/load_arbitrary_tool/..."
+    - "-//test/test_env/..."
     - "@examples//..."
     - "-@examples//ffi/rust_calling_c:matrix_dylib_test"
     - "-@examples//ffi/rust_calling_c:matrix_dynamically_linked"
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -58,6 +58,7 @@ tasks:
     - "-//bindgen/..."
     - "-//test/load_arbitrary_tool/..."
     - "-//test/test_env/..."
+    - "-//tools/rust_analyzer/..."
     - "@examples//..."
     - "-@examples//ffi/rust_calling_c:matrix_dylib_test"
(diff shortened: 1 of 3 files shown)


hlopko:
CC @UebelAndre

.bazelci/presubmit.yml
```diff
@@ -53,6 +54,14 @@ tasks:
     - "--enable_runfiles" # this is not enabled by default on windows and is necessary for the cargo build scripts
     windows_targets: &windows_targets
     - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
+    - "..."
+    - "-//bindgen/..."
+    - "-//test/load_arbitrary_tool/..."
+    - "-//test/test_env/..."
+    - "-//test/proto/..."
+    - "-//tools/rust_analyzer/..."
+    - "-//tools/runfiles/..."
```
illicitonion, line 63: I think runfiles is meant to work on Windows? (Not sure about the others). But also, definitely let's merge and try to shrink this list over time :)

illicitonion (approved):

test/BUILD
```diff
@@ -9,9 +9,17 @@ rule_test(
     rule = "@examples//hello_lib:hello_lib",
 )
 
+config_setting(
```
UebelAndre, line 12: IS there a reason you can't user `//rust/platforms:windows`?

UebelAndre (changes requested):
Sounds good. I'm curious how you actually ensure you're trying to enable more. If there's an automated way to remind you you're not testing everything or try to enable tests and see if something passes based on some change.

test/BUILD
hlopko: There is not, fixed. Thank you!

.bazelci/presubmit.yml
hlopko: Yup, I'll file issues for those.

hlopko:
> Sounds good. I'm curious how you actually ensure you're trying to enable more. If there's an automated way to remind you you're not testing everything or try to enable tests and see if something passes based on some change.

I don't think they will be fixed accidentally, so a system automatically checking if they pass won't be very useful. I'll create issues for them and we should try to fix them soon.

PR #624, opened 2021-03-03
ccinfo: when providing ccinfo, optionally include libstd and alloc

durin42:
The new attribute on RustToolchain is the label of a target that
provides __rust_realloc et al, which allows ld(1) to use the .rlib
files directly without needing to involve rustc in the linking
step. This means Rust and C++ can be mixed in a cc_binary freely
without needing any staticlib-type crates, which avoids problems if
you have a cc_binary -> rust_library -> cc_library -> rust_library
situation.


durin42:
(I expect this to be broken, please don't review just yet.)

durin42:
(Still some more to sort out, please don't merge yet.)

rust/private/rustc.bzl
```diff
@@ -693,6 +693,9 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
         if CcInfo in dep:
             cc_infos.append(dep[CcInfo])
 
+    if crate_info.type in ("rlib", "lib") and toolchain.extra_libstd_and_allocator_ccinfo:
```
hlopko, line 696: Hmm maybe there's a potential optimization, if we depend on an rlib rust_library (a target that provides CrateInfo and CcInfo at the same time), we can assume that it will depend on the allocator ccinfo. If there are no such deps, the current target has to depend on the CcInfo.

Either dismiss on add a todo, let's not implement it in this PR.

rust/toolchain.bzl
```diff
@@ -17,6 +27,42 @@ def _rust_toolchain_impl(ctx):
...
+    link_inputs = []
+    std_rlibs = [f for f in ctx.attr.rust_lib.files.to_list() if f.basename.endswith(".rlib")]
+    cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
+    if std_rlibs:
+        link_inputs.append(cc_common.create_linker_input(
+            owner = ctx.attr.rust_lib.label,
+            libraries = depset(
+                [
+                    cc_common.create_library_to_link(
+                        # TODO(augie): wat? this needs no actions
```
hlopko, line 39: In some situations create_library_to_link will create a symlink for each shared library into one big directory so runtime linker doesn't have to look for libraries in multiple directories.

rust/toolchain.bzl
```diff
@@ -17,6 +27,42 @@ def _rust_toolchain_impl(ctx):
...
+                        actions = ctx.actions,
+                        feature_configuration = feature_configuration,
+                        cc_toolchain = cc_toolchain,
+                        static_library = _make_dota(ctx, f),
```
hlopko, line 43: Nit: also set pic_static_library

rust/toolchain.bzl
```diff
@@ -17,6 +27,42 @@ def _rust_toolchain_impl(ctx):
...
+                    # TODO(augie): wat? this needs no actions
+                    actions = ctx.actions,
+                    feature_configuration = feature_configuration,
+                    cc_toolchain = cc_toolchain,
+                    static_library = f,
+                )
+                for f in ctx.attr.allocator_library.files.to_list()
+            ]),
+        ))
+    extra_libstd_and_allocator_ccinfo = None
```
hlopko, line 63: Nit: I'd drop `extra_` from the name

rust/toolchain.bzl
```diff
@@ -17,6 +27,42 @@ def _rust_toolchain_impl(ctx):
     for k, v in ctx.attr.debug_info.items():
         if not k in ctx.attr.opt_level:
             fail("Compilation mode {} is not defined in opt_level but is defined debug_info".format(k))
+    link_inputs = []
```
hlopko, line 30: I'd extract this whole logic into a function retuning CcInfo or None and call it in the provider constructor.

hlopko:
This looks reasonable so far. WDYT about also submitting your kludge cc_library? It is currently rustc version dependent, but we can make it work now, see how often it breaks, and we can get feedback on it from the community. It is also good to be able to show what we do now when discussing https://github.com/rust-lang/rust/issues/73632 with folks.

rust/toolchain.bzl
```diff
@@ -1,5 +1,15 @@
 """The rust_toolchain rule definition and implementation."""
 
+load(
+    "//rust/private:utils.bzl",
+    "find_cc_toolchain",
+)
+
+def _make_dota(ctx, f):
```
UebelAndre, line 8: Nit: can you add a docstring for any new macros?

UebelAndre (changes requested):

rust/private/rustc.bzl
durin42: added todo

durin42:
Still pending: a test, and I think this ends up providing the rust_lib libraries in the wrong order, causing linker problems. But close.

More later today, I hope.

durin42:
Progress, mostly: for some reason linking is failing to find the hello_from_rust symbol. The linker invocation looks vaguely okay, so I'll stop here for now.

durin42:
Okay, closer. Lingering problems:

1) I'm not sure how to configure the toolchain in the test to tell it about the allocator library

2) core needs`__muloti4`, which is missing during linking.

I'm _really_ stopping for the night this time.

durin42:
Looked into that linker failure more. As far as I can tell, that's a subtle incompatibility between the libcore Rust produces and libgcc (libcore is assuming compiler-rt) when doing 128 bit overflow checks.  https://github.com/rust-lang/rust/issues/8449 is related-ish, at least. It sounds to me like we need some work done in LLVM to make the overflow checks compatible with libgcc again, which will then have to percolate through the Rust ecosystem next time they pick up an LLVM release.

Between that and the fact that MSVC can't seem to cope with the allocator library sources, I'm out of ideas for ways to test this in-tree. I can add a test that validates the presence of the libstd CcInfo on a simple rust_library, but that seems to be about it. Is that enough for folks to be happy?

test/unit/native_deps/alloc_shims.c
```diff
@@ -0,0 +1,24 @@
+#include <stddef.h>
+#include <stdlib.h>
+
+void* __rdl_alloc(size_t, size_t);
```
hlopko, line 4: I'd put this file into generated rust toolchains for linux when cc rules detected clang, but let's do it in a followup PR.

test/unit/native_deps/cc_bin_uses_rust_library.cc
```diff
@@ -0,0 +1,7 @@
+extern "C" {
```
hlopko, line 1: Let's move this test from //test/unit/native_deps to //test/native_deps, I'd say it's not a unit test, it's assertion is that it can link.

rust/toolchain.bzl
```diff
@@ -38,12 +150,17 @@ def _rust_toolchain_impl(ctx):
         default_edition = ctx.attr.default_edition,
         compilation_mode_opts = compilation_mode_opts,
         crosstool_files = ctx.files._crosstool,
+        libstd_and_allocator_ccinfo = _make_libstd_and_allocator_ccinfo(ctx, ctx.attr.rust_lib, ctx.attr.allocator_library),
     )
     return [toolchain]
 
 rust_toolchain = rule(
     implementation = _rust_toolchain_impl,
+    fragments = ["cpp"],
```
hlopko, line 159: Could we add a unit test that CcInfo generated on the rust toolchain orders rlibs in the way you expect? You'd have to create a new package under //test/unit, for inspiration, https://github.com/bazelbuild/rules_rust/pull/576 did something similar.

hlopko:
Regarding testing, I'd be fine with excluding the failing test on windows (also mac?) CI tasks (.bazelci/presubmit.yml). For linux, this still doesn't work with gcc right? In that case I think we can add new tasks for clang on linux (CI workers should have clang installed, I'm not sure if they use libc++ though)

test/unit/native_deps/cc_bin_uses_rust_library.cc
durin42: I worked with @hlopko over chat on this. There doesn't appear to be a way to write this test case so it disappears if clang isn't in use, so we're punting for now and removing the test. :/

hlopko (approved):

Issue #386, opened 2020-08-03
rust_binary(out_binary=True) is broken for Wasm targets.


PiotrSikora:
The `out_binary = True` feature to build Rust binaries for Wasm targets introduced in #312 is broken since #361.
```
$ cat BUILD
```
```
load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary")

rust_binary(
    name = "test_rust",
    srcs = ["test_rust.rs"],
    edition = "2018",
    visibility = ["//visibility:public"],
    crate_type = "cdylib",
    out_binary = True,
)
```
```
$ cat test_rust.rs
```
```
#[no_mangle]
extern "C" fn abort() {
    panic!("abort")
}
```
This works fine when using last commit before #361 (i.e. `fda9a1ce6482973adfda022cadbfa6b300e269c3`):
```
$ bazel build --platforms=@io_bazel_rules_rust//rust/platform:wasm //:test_rust
```
```
Starting local Bazel server and connecting to it...
INFO: Invocation ID: fdecbf32-5468-4c23-9b20-7cf6c993a24c
INFO: SHA256 (https://static.rust-lang.org/dist/rust-std-1.44.0-wasm32-unknown-unknown.tar.gz) = 8e12796a0c2fb083953042218f832bdeb78da1bfaf67b9dfe3d719920084d755
INFO: Analyzed target //:test_rust (24 packages loaded, 161 targets configured).
INFO: Found 1 target...
Target //:test_rust up-to-date:
  bazel-bin/test_rust.wasm
INFO: Elapsed time: 19.079s, Critical Path: 2.90s
INFO: 5 processes: 1 remote cache hit, 4 processwrapper-sandbox.
INFO: Build completed successfully, 11 total actions
```
but is broken starting from #361 until `master` (i.e. `fdf9655ba95616e0314b4e0ebab40bb0c5fe005c`):
```
$ bazel build --platforms=@io_bazel_rules_rust//rust/platform:wasm //:test_rust
```
```
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 3d32152c-bc89-4e2c-98ef-40caa00823a7
INFO: SHA256 (https://static.rust-lang.org/dist/rust-std-1.44.0-wasm32-unknown-unknown.tar.gz) = 8e12796a0c2fb083953042218f832bdeb78da1bfaf67b9dfe3d719920084d755
[...]

PR #626, opened 2021-03-06
Re-add support for building Wasm libraries as executables.

PiotrSikora:
The ability to build Wasm libraries as executables is needed to support
WASI reactors (Wasm executables with multiple entrypoints).

This is a temporary workaround, and we should be able to use crate-type
"bin" when a proper support for WASI reactors (rust-lang/rust#79997) is
stabilised is Rust.

This feature was added in #312, and most recently broken in #592.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -789,13 +789,6 @@ rust_proc_macro = rule(
 )
 
 _rust_binary_attrs = {
-    "linker_script": attr.label(
-        doc = _tidy("""
-            Link script to forward into linker via rustc options.
-        """),
-        cfg = "exec",
-        allow_single_file = True,
-    ),
     "crate_type": attr.string(
         doc = _tidy("""
             Crate type that will be passed to `rustc` to be used for building this crate.
@@ -804,6 +797,13 @@ _rust_binary_attrs = {
             for WebAssembly targets (//rust/platform:wasi and //rust/platform:wasm).
         """),
         default = "bin",
+    ),
+    "linker_script": attr.label(
+        doc = _tidy("""
+            Link script to forward into linker via rustc options.
+        """),
+        cfg = "exec",
+        allow_single_file = True,
     ),
     "out_binary": attr.bool(),
 }
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -255,11 +255,11 @@ def _rust_binary_impl(ctx):
     return rustc_compile_action(
         ctx = ctx,
         toolchain = toolchain,
-        crate_type = "bin",
+        crate_type = ctx.attr.crate_type,
         crate_info = rust_common.crate_info(
             name = crate_name,
-            type = "bin",
-            root = crate_root_src(ctx.attr, ctx.files.srcs, crate_type = "bin"),
+            type = ctx.attr.crate_type,
+            root = crate_root_src(ctx.attr, ctx.files.srcs, ctx.attr.crate_type),
             srcs = ctx.files.srcs,
             deps = ctx.attr.deps,
             proc_macro_deps = ctx.attr.proc_macro_deps,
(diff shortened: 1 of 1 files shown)


rust/private/rust.bzl
```diff
@@ -796,6 +796,15 @@ _rust_binary_attrs = {
         cfg = "exec",
         allow_single_file = True,
     ),
+    "crate_type": attr.string(
+        doc = _tidy("""
+            Crate type that will be passed to `rustc` to be used for building this crate.
+
+            This option is a temporary workaround and should be used only when building
+            for WebAssembly targets (//rust/platform:wasi and //rust/platform:wasm).
+        """),
```
PiotrSikora, line 805: Let me know if you want to document this feature or if I should remove the `doc` field.

PiotrSikora (commented):
@hlopko PTAL

hlopko:
I'm sorry for breaking this /o\. If this is urgent I'm happy to approve the PR, but I'd like to continue discussing the options.

So if I understand things you want to use `rust_binary` rule for reactors, but pass `--crate-type=lib`? I deduce that from https://github.com/rust-lang/rust/pull/79997#issuecomment-759856446. But then from https://github.com/bazelbuild/rules_rust/pull/312 it seems you want to use `--crate-type=cdylib`. Is it an option to use `rust_shared_library` instead of `rust_binary`?

What exactly is going to change after this feature is stabilized in rustc? Won't we still need to have some way of telling the rules that they are building the reactor and that they'll have to set the `DefaultInfo.executable`?

PiotrSikora:
> I'm sorry for breaking this /o. If this is urgent I'm happy to approve the PR, but I'd like to continue discussing the options.

Not urgent, but it would be great if it worked, i.e. a few days is fine, a few months is not.

> So if I understand things you want to use `rust_binary` rule for reactors, but pass `--crate-type=lib`? I deduce that from [rust-lang/rust#79997 (comment)](https://github.com/rust-lang/rust/pull/79997#issuecomment-759856446).

You can build WASI reactors using `--crate-type=bin --target=wasm32-wasi` and `RUSTFLAGS="-Z wasi-exec-model=reactor"`, but that's an unstable nightly feature, so for now we create a "fake" WASI reactor by building a Wasm library with `_initialize` function.

> But then from #312 it seems you want to use `--crate-type=cdylib`. Is it an option to use `rust_shared_library` instead of `rust_binary`?

`rust_shared_library` doesn't work for `.wasm` files (i.e. when using `--platforms=@rules_rust//rust/platform:wasi`):
```
        File ".../rules_rust/rust/private/rust.bzl", line 179, column 32, in _rust_shared_library_impl
                return _rust_library_common(ctx, "cdylib")
        File ".../rules_rust/rust/private/rust.bzl", line 221, column 32, in _rust_library_common
                return rustc_compile_action(
        File ".../rules_rust/rust/private/rustc.bzl", line 614, column 29, in rustc_compile_action
                return establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration) + [
        File ".../rules_rust/rust/private/rustc.bzl", line 671, column 59, in establish_cc_info
                library_to_link = cc_common.create_library_to_link(
Error in create_library_to_link: 'libhello_world-665202032.wasm' does not have any of the allowed extensions .so, .dylib or .dll
```

You can use the repro steps from the previous time it was broken (https://github.com/bazelbuild/rules_rust/issues/386), and adjust accordingly.

But I think it would be better to restore the previously working behavior, otherwise users need to switch from existing:
```
rust_binary(
    name = "foo",
    crate_type="cdylib",
    out_binary=True,
)
```
to:
```
rust_shared_library(
    name = "foo",
)
```
only to switch back again to `rust_binary`, once the WASI reactors are stabilised:
```
rust_binary(
    name = "foo",
    rustc_flags = "-Z wasi-exec-model=reactor",
) 
```

> What exactly is going to change after this feature is stabilized in rustc? Won't we still need to have some way of telling the rules that they are building the reactor and that they'll have to set the `DefaultInfo.executable`?

I think this would work just fine (WASI reactors use `--crate-type=bin`):
```
rust_binary(
    name = "foo",
    rustc_flags = "-Z wasi-exec-model=reactor",
) 
```

Note, that ideally we would have dedicated `rust_wasm_{binary,library}` targets that automatically transition to a Wasm platform (either `@rules_rust//rust/platform:wasm` or `@rules_rust//rust/platform:wasi`), and maybe they would be aware of the WASI exec models (`command` and `reactor`).

Right now, we're using "regular" `rust_{binary,library}` targets and set `--platform` accordingly to build Wasm files, but then the whole workspace transitions to Wasm, and you cannot build mixed projects this way (e.g. have a native `rust_binary` that loads `.wasm` file generated from Wasm `rust_binary`).

hlopko:
Thank you for the thorough explanation! Yeah let's merge this.

hlopko (approved):

PR #628, opened 2021-03-08
migration: include incompatible_use_toolchain_transition globally

durin42:
This is step 2 of
https://github.com/bazelbuild/proposals/blob/master/designs/2020-02-07-toolchain-transition-migration.md,
and was requested by someone at work. We'll eventually want to drop
this (see step 9), but for now this enables more-correct behavior
around toolchain transitions.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -243,6 +243,7 @@ _build_script_run = rule(
         str(Label("//rust:toolchain")),
         "@bazel_tools//tools/cpp:toolchain_type",
     ],
+    incompatible_use_toolchain_transition = True,
 )
 
 def cargo_build_script(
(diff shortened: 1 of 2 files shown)


katre:
Thank you! (I am the requestor internally).

dfreese:
This seems reasonable to me.  It appears to require a minimum of [Bazel 3.4](https://blog.bazel.build/2020/07/13/bazel-3-4.html).  Our current minimum version is 3.5.

The only spot where this might get weird is the configuration is never specified on the toolchain, but I imagine remote build execution users were always specifying a rust repository for the RBE environment.  @illicitonion has more experience using rules_rust with remote build execution, so I'll give him a day to comment.

hlopko:
I've migrated other rules in the repo as well. The PR LGTM, Daniel to concur and merge :)

hlopko (approved):

hlopko:
FTR, we're using the rules with remote execution as well and this PR works for us.

illicitonion (approved):
LGTM - thanks!

Issue #635, opened 2021-03-16
Error: local variable 'cc_toolchain' is referenced before assignment.


iffyio:
Hi! I get this error from `rust_library` targets after updating to commit `76c44206c4d37491767446c6ccdf1a98274a0887`
```
Traceback (most recent call last):
        File "/home/iffy/.cache/bazel/_bazel_iffy/7f16182b13b4a660f93298499007a04c/external/rules_rust/rust/toolchain.bzl", line 154, column 72, in _rust_toolchain_impl
                libstd_and_allocator_ccinfo = _make_libstd_and_allocator_ccinfo(ctx, ctx.attr.rust_lib, ctx.attr.allocator_library),
        File "/home/iffy/.cache/bazel/_bazel_iffy/7f16182b13b4a660f93298499007a04c/external/rules_rust/rust/toolchain.bzl", line 43, column 60, in _make_libstd_and_allocator_ccinfo
                cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
        File "/home/iffy/.cache/bazel/_bazel_iffy/7f16182b13b4a660f93298499007a04c/external/rules_rust/rust/private/utils.bzl", line 43, column 24, in find_cc_toolchain
                cc_toolchain = cc_toolchain,
Error: local variable 'cc_toolchain' is referenced before assignment.
```
it seems like somehow `@bazel_tools//tools/cpp:toolchain_type` was supposed to be set but no longer isnt?

UebelAndre:
Can you provide an example that demonstrates this issue?

iffyio:
The issue occurs when using `--incompatible_enable_cc_toolchain_resolution`

given `WORKSPACE` file

```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "rules_rust",
    strip_prefix = "rules_rust-76c44206c4d37491767446c6ccdf1a98274a0887",
    urls = [
        "https://github.com/bazelbuild/rules_rust/archive/76c44206c4d37491767446c6ccdf1a98274a0887.tar.gz",
    ],
)
load("@rules_rust//rust:repositories.bzl", "rust_repositories")
rust_repositories(version="1.49.0")
```

given the target
```python
load("@rules_rust//rust:rust.bzl", "rust_library")
rust_library(
    name = "hello-world",
    srcs = [
        "src/main.rs",
    ],
    edition = "2018",
    visibility = ["//visibility:public"],
)
```
```bash
[...]

PR #638, opened 2021-03-18
Add cpp toolchain requirement for rust_toolchain

katre:
Part of the work on #635, but not sufficient to fix it.


illicitonion (approved):
Thanks!

PR #639, opened 2021-03-20
Updated rules_foreign_cc to 0.2.0 and migrated rules to new api.

UebelAndre:
[rules_foreign_cc 0.2.0](https://github.com/bazelbuild/rules_foreign_cc/releases/tag/0.2.0) was just cut and contains some API changes. This PR updates the user of `rules_foreign_cc` rules to use the new API.


UebelAndre:
Ah, this change would require updating the min supported version to `3.7.0` so `@platforms//os:windows` could be `select`'ed on

illicitonion:
As far as I know we only use rules_foreign_cc in tests, so it seems like a shame to bump the minimum supported bazel version just for our own tests... Is there any major harm in holding off on this until we happen to bump our minimum bazel version to >= 3.7.0 for some other reason?

UebelAndre:
No harm. This can just sit here until then

UebelAndre:
Since [0.3.0](https://github.com/bazelbuild/rules_foreign_cc/releases/tag/0.3.0) is out this PR should be updated. But, I've since re-forked the repo and this PR is no longer connected to my fork so I'll have to do it in a new PR. Sorry about the noise 😅

PR #640, opened 2021-03-22
The default wasm-bindgen toolchain can now optionally be registered

UebelAndre:
I find it to be the common case that users will want to register their own `rust_wasm_bindgen_toolchain` since they'll be writing code with their own set of crate dependencies and would otherwise be limited to the version of `wasm-bindgen` used in the default toolchain. This PR should make it clearer and easier for users looking to setup their own toolchain.

Notable changes in this PR:
- Added `register_default_toolchain` to `rust_wasm_bindgen_repositories`
- Improved docs
- Raised the default version of `rules_nodejs` from `3.1.0` to `3.2.2` to pick up a few bug fixes.
- Added a `providers.bzl` file for easy access to what `rust_wasm_bindgen` returns


UebelAndre:
@illicitonion Hey, do you have a second to take a look here? 🙏

illicitonion (approved):
Looks reasonable :) Thanks!

PR #641, opened 2021-03-22
Revert accidental bump in minimum version in docs

illicitonion:
https://github.com/bazelbuild/rules_rust/pull/586 bumped the minimum
supported version we test to 3.5.0 but presumably accidentally bumped
the documentation to say 4.0.0 - make those two values match.


hlopko (approved):
Yes 3.5.0 is correct, probably a missed revert.

UebelAndre:
Sorry about that, thanks for fixing this!

PR #642, opened 2021-03-22
Re-gen cargo raze

illicitonion:
These were generated with cargo-raze at
c748b0c4ed7ad6b516f7838a27fee27a9d9c8664.

This removes the need to manually edit
examples/complex_sys/raze/remote/BUILD.libssh2-sys-0.2.20.bazel as well
as adding the `links` attribute to crates which need it, allowing us to
remove `cargo_build_script`'s `crate_name` attribute.


UebelAndre:
link for easy reference: https://github.com/google/cargo-raze/commit/c748b0c4ed7ad6b516f7838a27fee27a9d9c8664

hlopko (approved):
This is practically unreviewable :) I'd also say that raze should generate minimal and buildifier compliant BUILD files, generating these loads and not using them is not great.

PR #643, opened 2021-03-22
Set CARGO_CRATE_NAME and CARGO_PKG_NAME properly

illicitonion:
`CARGO_PKG_NAME` is set to exactly the rule's name, stripping exactly
one trailing `_build_script` if present for build scripts.

`CARGO_CRATE_NAME` is set to exactly `CARGO_PKG_NAME` with `-`s replaced
with `_s`s.

There is a breaking change in here:
`cargo_build_script` no longer has a `crate_name` attribute.

A `cargo_build_script`'s `CARGO_PKG_NAME` will now be set from the
rule's `name` attribute as described above.

A `cargo_build_script`'s `_DEP_ENV` env vars will no longer try to guess
the correct value of the `links` attribute by inferring from the crate
name, and instead the `links` attribute should be set explicitly.


illicitonion:
@arlyon this implements https://github.com/bazelbuild/rules_rust/pull/537 as described in https://github.com/bazelbuild/rules_rust/pull/537#issuecomment-801835272 - would be great to get your thoughts (and thanks for kicking off the discussion)!

arlyon:
Had a look through the code just now and actually exposing rustc env is a nice thouch. Thanks! :)

hlopko (approved):
LGTM, I'd wait for @martinboehme to take a look too if it's fine.

cargo/cargo_build_script.bzl
```diff
@@ -57,14 +54,7 @@ def _build_script_impl(ctx):
...
 
-    crate_name = ctx.attr.crate_name
-
-    # Derive crate name from the rule label which is <crate_name>_build_script if not provided.
-    if not crate_name:
-        crate_name = ctx.label.name
-        if crate_name.endswith("_build_script"):
-            crate_name = crate_name.replace("_build_script", "")
-        crate_name = crate_name.replace("_", "-")
+    pkg_name = _name_to_pkg_name(ctx.attr.name)
```
martinboehme, line 57: I'm new to Bazel, but most places seem to use `ctx.label.name` rather than `ctx.attr.name`. I _think_ they should be equivalent, but is there any reason to prefer one over the other?

cargo/cargo_build_script.bzl
```diff
@@ -79,8 +69,9 @@ def _build_script_impl(ctx):
     env = dict(ctx.configuration.default_shell_env)
 
     env.update({
+        "CARGO_CRATE_NAME": name_to_crate_name(pkg_name),
```
martinboehme, line 72: Nit: `name_to_crate_name` is documented as taking a Bazel target name, but here we're applying it to a Cargo package name. True, it does the right thing, but we distinguish between Bazel target names and Cargo package names in other places (e.g. in the `_name_to_pkg_name` function you're introducing in this change), so using `name_to_crate_name` in this way muddies the distinction a bit.

Here are two alternatives:
- Eliminate `name_to_crate_name` entirely and replace it with `.replace("-", "_")`. (Note that `x.replace("-", "_")` actually has _fewer_ characters than `name_to_crate_name(x)`. For this reason, I would also not suggest simply renaming `name_to_crate_name` to something more generic, such as `hyphens_to_underscores`, because that would say exactly the same thing as `.replace("-", "_")`, just with more characters.)
- If `name_to_crate_name` _is_ a useful function, then it's not because it saves characters (see above) but because it introduces a useful abstraction: It converts one type of name to another without us having to be aware of the rules for doing so. If we believe this abstraction is useful (I do), it should be crisp about the type of name it accepts and returns, so we should provide differently named functions, e.g. `target_name_to_crate_name` and `pkg_name_to_crate_name`, even if they happen to have the same implementation. (Note that `cargo_build_script.bzl` contains another use of `name_to_crate_name` to convert a Cargo feature name to the name of the corresponding `CARGO_FEATURE_` environment variable. I plan to replace this usage with a function that is specific to feature names.)
  If you go for this alternative, I realize it does mean some churn. I don't think all existing uses of `name_to_crate_name` would need to be replaced in this PR -- that could (and probably should) happen in a separate PR. Happy to help out with this too.

cargo/cargo_build_script.bzl
```diff
@@ -316,32 +303,43 @@ def cargo_build_script(
...
-        crate_name (str, optional): Name of the crate associated with this build script target.
+        name (str): The target name for the underlying rule. This should be the name of the crate being compiled, optionally with a suffix of _build_script.
         crate_features (list, optional): A list of features to enable for the build script.
         version (str, optional): The semantic version (semver) of the crate.
-        deps (list, optional): The dependencies of the crate defined by `crate_name`.
+        deps (list, optional): The dependencies of the crate.
         build_script_env (dict, optional): Environment variables for build scripts.
         data (list, optional): Files or tools needed by the build script.
         links (str, optional): Name of the native library this crate links against.
+        rustc_env (dict, optional): Environment variables to set in rustc when compiling the build script.
```
martinboehme, line 313: This doesn't seem to be mentioned in the PR description -- can you add a note there?

Also, what's the reason for adding this here? Is this just for commonality with the other rules?

cargo/cargo_build_script.bzl
```diff
@@ -79,8 +69,9 @@ def _build_script_impl(ctx):
     env = dict(ctx.configuration.default_shell_env)
 
     env.update({
```
martinboehme, line 71: I notice `rustc.bzl` has `_get_rustc_env()`. Can we standardize on that instead of maintaining this separate code for setting environment variables? Or does rustc have different requirements than build scripts in terms of the environment variables that should be set?

cargo/cargo_build_script.bzl
```diff
@@ -316,32 +303,43 @@ def cargo_build_script(
...
         links (str, optional): Name of the native library this crate links against.
+        rustc_env (dict, optional): Environment variables to set in rustc when compiling the build script.
         **kwargs: Forwards to the underlying `rust_binary` rule.
     """
+    rustc_env = dict(rustc_env)
+    if "CARGO_PKG_NAME" not in rustc_env:
+        rustc_env["CARGO_PKG_NAME"] = _name_to_pkg_name(name)
+    if "CARGO_CRATE_NAME" not in rustc_env:
+        rustc_env["CARGO_CRATE_NAME"] = name_to_crate_name(_name_to_pkg_name(name))
+
```
martinboehme, line 321: We already have code for setting these environment variables in `_build_script_impl`. Am I correct in thinking that we do this so they will be available to `rust_binary` as well as `_build_script_run`? Maybe add a comment explaining this or possibly consolidate the code into a single location?

As a side note, when https://github.com/bazelbuild/rules_rust/pull/645 lands, an even nicer way of doing this would be to simply pass the crate name to the `rust_binary` rule through the `crate_name` attribute (which of course means that `_get_rustc_env` would need to be modified to set `CARGO_CRATE_NAME` appropriately -- I can make a note to do this as part of https://github.com/bazelbuild/rules_rust/pull/645). Is just a bit unsatisfying that we can't do the same thing for the package name. So do we go down the rabbit hole of adding a `package_name` attribute as well? On some level, this might actually be the most principled way to do this, as there is currently no way for a Bazel rule to infer the name of the package, other than relying on the convention that the target name is the same as the package name (possibly with a `_build_script` suffix in the case of `cargo_build_script` rules).

WDYT?

cargo/cargo_build_script.bzl
```diff
@@ -316,32 +303,43 @@ def cargo_build_script(
     build script in addition to the file generated by it.
 
     Args:
-        name (str): The target name for the underlying rule
-        crate_name (str, optional): Name of the crate associated with this build script target.
+        name (str): The target name for the underlying rule. This should be the name of the crate being compiled, optionally with a suffix of _build_script.
```
martinboehme, line 306: Since we're being crisp about the distinction between package names and crate names, replace "name of the crate being compiled" with "name of the package being compiled".

martinboehme (changes requested):

cargo/cargo_build_script.bzl
illicitonion: This isn't actually a change in behaviour at all - the `kwargs` are passed through to the underlying `rust_binary` already, this just names one of them so that we can operate on it with a name, rather than as `kwargs.get("rustc_env", {})`. The equivalent code without this would be something like:

```python
if "rustc_env" not in kwargs:
    kwargs["rustc_env"] = {}
rustc_env = dict(kwargs["rustc_env"])
if "CARGO_PKG_NAME" not in rustc_env:
    rustc_env["CARGO_PKG_NAME"] = _name_to_pkg_name(name)
if "CARGO_CRATE_NAME" not in rustc_env:
    rustc_env["CARGO_CRATE_NAME"] = name_to_crate_name(_name_to_pkg_name(name))
```

which seems a little less tidy, but maybe is more clear about what's going on? Either way, the intent I'm trying to convey is "We always pass through a `rustc_env` to the `rust_binary`, you can add to it if you want"...

martinboehme: Got it -- that makes total sense.

So I think the only issue here is my lack of understanding, and there's nothing to be done here.

cargo/cargo_build_script.bzl
illicitonion: I've filed https://github.com/bazelbuild/rules_rust/issues/661 with all of the information I know here - let's tackle it as a follow-up :)

cargo/cargo_build_script.bzl
illicitonion: Done

cargo/cargo_build_script.bzl
illicitonion: Added a comment for now - I suspect solving #661 will get rid of the duplication.

On adding a `package_name` attribute, I think I'm fairly against the idea. My reasoning:
1. Most people don't know that there are separate concepts of a package name and a crate name in the first place, so having these options probably makes them need to do some learning, as well as growing the API surface.
2. Cargo's behaviour is to always derive the package name exactly from the crate name, and given #645 allows for specifying a custom crate name, I'm not sure there's a reasonable use-case for independently specifying a package name.

Do you have a concrete use-case for someone wanting to independently set `crate_name` and `package_name`?

cargo/cargo_build_script.bzl
martinboehme: SG, thanks!

cargo/cargo_build_script.bzl
illicitonion: I can see the value in being specific about what kind of name the function accepts, but in this particular case I'm not so sure... In general `CARGO_PKG_NAME` and `ctx.label.name` are identical, but for build scripts we may optionally strip a suffix, which is a fairly minor transformation.

If we're going to change something here, I'd lean towards renaming `_name_to_pkg_name` to something like `maybe_strip_build_script_suffix`, and continue treating `ctx.label.name` and `CARGO_PKG_NAME` as identical concepts (so maybe renaming `name_to_crate_name` to either `pkg_name_to_crate_name` or `bazel_name_to_crate_name`, but still calling it directly here despite having technically preprocessed).

I can get on board with inlining the `CARGO_FEATURE` one to just a `.replace("-", "_")`, that feels like it's not the same underlying abstraction being replaced there and doesn't really need a named function extracted :)

WDYT?

cargo/cargo_build_script.bzl
illicitonion: These should be exactly equivalent :) I've switched to `ctx.label.name` for consistency

cargo/cargo_build_script.bzl
martinboehme: I don't feel too strongly about this -- it's just something that triggered my  "mental type checker", so I thought I'd point it out. What you're proposing sounds fine to me!

cargo/cargo_build_script.bzl
martinboehme: > Added a comment for now - I suspect solving #661 will get rid of the duplication.
> 
> On adding a `package_name` attribute, I think I'm fairly against the idea. My reasoning:
> 
> 1. Most people don't know that there are separate concepts of a package name and a crate name in the first place, so having these options probably makes them need to do some learning, as well as growing the API surface.

Not growing the API surface is a good argument, and it's the reason I'm not convinced myself that adding a `package_name` attribute is a good idea.

On the issue of people not being aware of the distinction between package names and crate names -- that's also true, though I do wish the two terms got used less interchangeably.

> 2. Cargo's behaviour is to always derive the package name exactly from the crate name,

You mean to derive the crate name from the package name?

> and given #645 allows for specifying a custom crate name, I'm not sure there's a reasonable use-case for independently specifying a package name.
> 
> Do you have a concrete use-case for someone wanting to independently set `crate_name` and `package_name`?

Consider the case where a package defines both a library crate and a binary crate with the same name, which is not just allowed but the default in Cargo when there's both a library and binary crate.

The Bazel rules assume the target name is equal to the package name, but we can't give both the `rust_library` and the `rust_binary` target the same name, so we have no choice but to give one of these a name that is different from the package name. That target will then set its `CARGO_PKG_NAME` environment variable incorrectly.

So that's the use case for a `package_name` attribute. As for having an independent `crate_name` attribute: You need that to set the `CARGO_CRATE_NAME` correctly in the case where people override the `crate_name` in `Cargo.toml`.

Does all of this actually matter? I assume it probably doesn't (though I'm admittedly not sure what people use `CARGO_PKG_NAME` and `CARGO_CRATE_NAME` for in practice), but this is what I mean when I say that there is no reliable way for a Bazel rule to infer the name of the package.

martinboehme: Thanks for the detailed explanation! And good point about being able to do this by explicitly setting `CARGO_PKG_NAME`. I think the original issue that prompted this comment thread has long since been resolved as well, so feel free to mark this as resolved (I think I can't actually do that myself).

martinboehme:
Ready to be merged from my point of view.

wchargin:
To passersby: if this commit broke your build and you use `cargo-raze`,
update to `cargo-raze` version 0.12.0 and re-generate your build files
to fix the problem. (I hit this with `openssl-sys` and `openssl`.)

PR #645, opened 2021-03-25
Add a crate_name attribute to Rust rules

martinboehme:
This allows specifying a different crate name than the one derived by default from the Bazel target name.

This is useful, for example, if you want to use a target name that would not be legal as a crate name, e.g. a target name containing slashes.

This commit introduces a function `_crate_name()` and uses it where we were formerly determining the crate name using `name_to_crate_name(ctx.label.name)`.

There were some additional uses of `name_to_crate_name` in the test rule where we weren't actually determining a crate name but an output file name for the test binary or the test launcher. It's not clear to me why these needed the hyphen-to-underscore conversion that `name_to_crate_name` performs, as hyphens are legal in file names. I think these places should be doing one of the following:

- Create a file name based on the target name, without hyphen-to-underscore substitutions.
- Create a file name based on the (possibly non-default) crate name.

I think one of these two changes should be made, but I don't want to do so in the commit because either of these changes would create a change in behavior for existing uses of the `rust_test` rule. Also, the issue of test binary and test launcher file names is largely orthogonal to the purpose of this commit.

Finally, there is also a use of name_to_crate_name in `cargo_build_script.bzl` to replace hyphens in feature names with
underscores. This has nothing to do with crate names; it's just using an existing function that conveniently happens to be doing the thing the code needs. I think this use of name_to_crate_name should be replaced with a local function that does the same thing, but again, I'll do so in a separate commit.


martinboehme:
@hlopko Can you review?

martinboehme:
I'm new to writing Bazel rules, so I have a few questions on the test failures.

- The "Minimum Supported Version" test is failing with the error
  ```
  Error in fail: fail() accepts no more than 2 positional arguments but got 4.
  ```
  I _believe_ this test is trying to make sure that the rules work on the minimum supported version of Bazel... correct? And it looks as if that version does not yet support arbitrarily many arguments to `fail()`. Should I make sure to pass only a single argument to `fail()` (by just concatenating the strings myself)?
- The Windows test is failing with the error
  ```
  error: could not write output to bazel-out/x64_windows-fastbuild/bin/test/unit/crate_name\custom_name.hjct7k57wmifsnz.rcgu.o: permission denied". 
  ```
  The mix of forward and backward slashes in that path looks strange, but I'm not sure how my changes could have caused this, and I don't have a Windows system immediately available that I could debug this on. Any pointers on what could be going wrong here?

rust/private/rust.bzl
```diff
@@ -118,6 +118,48 @@ def crate_root_src(attr, srcs, crate_type):
...
+    """Returns the crate name to use for the current target.
+
+    Args:
+        label (Label): The label of the current target.
+        attr (struct): The attributes of the current target.
+
+    Returns:
+        str: The crate name to use for this target.
+    """
+    if attr.crate_name:
```
illicitonion, line 143: Can we unify these codepaths? Maybe with a prefix of something like:
```python
crate_name, crate_name_source = attr.name, "name"
if attr.crate_name:
    crate_name = attr.crate_name
    crate_name_source = "crate_name"
...
```

rust/private/rust.bzl
```diff
@@ -118,6 +118,48 @@ def crate_root_src(attr, srcs, crate_type):
...
+                attr.crate_name,
+                " ".join(invalid_chars),
+            ))
+        return attr.crate_name
+
+    crate_name = name_to_crate_name(label.name)
+    invalid_chars = _invalid_chars_in_crate_name(crate_name)
+    if invalid_chars:
+        fail(
+            "Crate name '{}'".format(crate_name),
```
illicitonion, line 156: I think you want to be `+`ing these, rather than `,`ing them together?

illicitonion (commented):
Thanks for the contribution!

> This allows specifying a different crate name than the one derived by default from the Bazel target name.

It'd be nice to avoid this feature if possible by changing names, but if that's not possible, this change generally seems reasonable :)

> Finally, there is also a use of name_to_crate_name in `cargo_build_script.bzl` to replace hyphens in feature names with
> underscores. This has nothing to do with crate names; it's just using an existing function that conveniently happens to be doing the thing the code needs. I think this use of name_to_crate_name should be replaced with a local function that does the same thing, but again, I'll do so in a separate commit.

I have https://github.com/bazelbuild/rules_rust/pull/643 out which overhauls this - would appreciate your thoughts (and it may change how you're thinking about that separate commit :))

hlopko:
Related reading: https://internals.rust-lang.org/t/pre-rfc-unify-dashes-and-underscores-on-crates-io/13216

rust/private/rust.bzl
martinboehme: I've tried this, but I'm not sure it's really better. Here's what I've got:

```
    crate_name, crate_name_source = name_to_crate_name(attr.name), "name"
    if attr.crate_name:
      crate_name = attr.crate_name
      crate_name_source = "crate_name"

    invalid_chars = _invalid_chars_in_crate_name(crate_name)

    if invalid_chars:
        if crate_name_source == "name":
            fail(
                "Crate name '{}'".format(crate_name),
                "derived from Bazel target name '{}'".format(label.name),
                "contains invalid character(s): {}".format(" ".join(invalid_chars)),
                "\nConsider adding a crate_name attribute to set a valid crate name",
            )
        else:
            fail("Crate name '{}' contains invalid character(s): {}".format(
                attr.crate_name,
                " ".join(invalid_chars),
            ))

    return crate_name
```

This is more lines than what I have above and uses "magic" strings for `crate_name_source` (which have the potential for typos).

Side note: I see you using `attr.name` instead of `label.name`. Is there a preference for one or the other in Bazel rules? I do see that in this particular case, using `attr.name` would allow me to eliminate the `label` parameter entirely, which seems attractive, but I wanted to ask first.

rust/private/rust.bzl
martinboehme: Thanks -- done!

martinboehme:
> Thanks for the contribution!
> 
> > This allows specifying a different crate name than the one derived by default from the Bazel target name.
> 
> It'd be nice to avoid this feature if possible by changing names, but if that's not possible, this change generally seems reasonable :)
> 
> > Finally, there is also a use of name_to_crate_name in `cargo_build_script.bzl` to replace hyphens in feature names with
> > underscores. This has nothing to do with crate names; it's just using an existing function that conveniently happens to be doing the thing the code needs. I think this use of name_to_crate_name should be replaced with a local function that does the same thing, but again, I'll do so in a separate commit.
> 
> I have #643 out which overhauls this - would appreciate your thoughts (and it may change how you're thinking about that separate commit :))

Thanks for pointing that out -- I've added comments there (rather many of them in fact...).

I do still feel that using the `name_to_crate_name` function for feature names is a bit dubious -- see detailed rationale in #643. I'd like to submit a PR that adds an `_env_var_from_feature_name` function -- this would also handle adding the `CARGO_FEATURE_` prefix. But I'll postpone this until #643 is in so I can add the corresponding test to the new test files you're adding in #643.

rust/private/rust.bzl
hlopko: I believe `ctx.attr.name` and `ctx.label.name` are identical. I'd default to using `ctx.attr.name` but I don't have a better reason than that I think it's more common.

rust/private/rustc.bzl
```diff
@@ -411,13 +411,13 @@ def construct_arguments(
...
     # have -s if needed.
-    maybe_rename = ""
+    emit_with_paths = emit
     if crate_info.type == "bin" and crate_info.output != None:
         generated_file = crate_info.name + toolchain.binary_ext
         src = "/".join([crate_info.output.dirname, generated_file])
         dst = crate_info.output.path
         if src != dst:
-            args.add_all(["--copy-output", src, dst])
+            emit_with_paths = [("link=" + dst if val == "link" else val) for val in emit]
```
hlopko, line 420: Thanks for doing this here! Good change.

test/unit/crate_name/crate_name_test.bzl
```diff
@@ -0,0 +1,189 @@
...
+def _custom_crate_name_test_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    actions = analysistest.target_actions(env)
+    assert_argv_contains(env, actions[0], "--crate-name=custom_name")
+    return analysistest.end(env)
+
+def _invalid_default_crate_name_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    asserts.expect_failure(env, "contains invalid character(s)")
```
hlopko, line 54: Nit: also cover which characters were invalid?

test/unit/crate_name/main.rs
```diff
@@ -0,0 +1 @@
+fn main() {}
```
hlopko, line 1: FYI, you could use this file also for rust_library srcs, but I actually don't mind having a separate empty file, so it's up to you.

hlopko (approved):
LGTM, thanks for working on this!

rust/private/rust.bzl
illicitonion: I was imagining something like:

```python
crate_name, source = attr.crate_name, "name"
if attr.crate_name:
    crate_name = attr.crate_name
    crate_name_source = "crate_name"

invalid_chars = _invalid_chars_in_crate_name(crate_name)

if invalid_chars:
    fail(
        "Crate name {crate_name} {source_description} contains invalid character(s): {invalid_characters}.{suggestion}".format(
            crate_name = crate_name,
            invalid_characters = " ".join(invalid_chars),
            source_description = "derived from Bazel target name {}".format(label.name) if crate_name_source == "name" else "from crate_name attribute",
            suggestion = "\nConsider adding a crate_name attribute to set a valid crate name" if crate_name_source == "crate_name" else "",
        )
    )
return crate_name
```

But I don't think either way is clearly better :) Happy to merge either way!

test/unit/crate_name/main.rs
martinboehme: I considered this too, but the rules read a bit weird with a `main.rs` in the sources of a `rust_library`, so I'd like to stay with this.

martinboehme:
Rebasing to HEAD now.

rust/private/rust.bzl
martinboehme: Ah, got it. If you're fine with either, I'd prefer to keep the original version -- some duplication, but less logic.

martinboehme: > I believe `ctx.attr.name` and `ctx.label.name` are identical. I'd default to using `ctx.attr.name` but I don't have a better reason than that I think it's more common.

I'm going with `ctx.attr.name` then because it allows me to eliminate the `label` parameter.

martinboehme:
I've added a TODO that the `CARGO_CRATE_NAME` environment variable (added in #643) needs to respect the `crate_name` attribute.

I originally planned to do this in this PR, but then I stumbled across a bug in the Clippy aspect implementation where it's accessing `ctx.attr` instead of `ctx.rule.attr` and so not finding the `crate_name` attribute. I'll fix this bug in a separate PR and will then handle the outstanding TODO.

hlopko (approved):

PR #646, opened 2021-03-25
Update default Rust version to 1.51.0.

PiotrSikora:


illicitonion (approved):
Thanks!

PR #651, opened 2021-03-29
Moved crate_universe into it's own directory

UebelAndre:
Given that there is support for interfacing with this rule without any `Cargo.toml` files, I would say it's not a "cargo" tool and should go elsewhere. It may still use cargo under the hood but I don't think this is relevant enough to warrant this rule living in `./cargo`.

This PR moves the `crate_universe` rule into it's own directory and `workspace.bzl` has been renamed to `defs.bzl` to be consistent with `./rust`


UebelAndre:
cc @illicitonion

crate_universe/CONTRIBUTING.md
```diff
@@ -2,14 +2,15 @@
 
 ## Tour of the codebase
 
-We start at `workspace.bzl`, which invokes the resolver.
+We start at `def.bzl`, which invokes the resolver.
```
illicitonion, line 5:
```suggestion
We start at `defs.bzl`, which invokes the resolver.
```

crate_universe/CONTRIBUTING.md
```diff
@@ -18,16 +19,18 @@ Some areas have unit testing, there are a few (very brittle) integration tests,
...
 ```
 
 Then run a bazel build with this environment variable:
+
 ```console
 RULES_RUST_RESOLVER_URL_OVERRIDE=file:///path/to/resolver/target/debug/resolver bazel build //whatever
 ```
 
-To get verbose logging, edit `workspace.bzl` to set `RUST_LOG` to `debug` or `trace` instead of `info`. In particular, that will print out the generated Cargo.toml, and the path to the generated workspace file.
+To get verbose logging, edit `def.bzl` to set `RUST_LOG` to `debug` or `trace` instead of `info`. In particular, that will print out the generated Cargo.toml, and the path to the generated workspace file.
```
illicitonion, line 33:
```suggestion
To get verbose logging, edit `defs.bzl` to set `RUST_LOG` to `debug` or `trace` instead of `info`. In particular, that will print out the generated Cargo.toml, and the path to the generated workspace file.
```

illicitonion (approved):
Works for me!

integration.rs will need a couple of changes to some hard-coded strings, too

UebelAndre:
@illicitonion I cannot find what you're referring to in the test. They feel quite unmaintainable IMO and would opt to silence it for now. I feel the very next thing that should be done is to update the output to be a well formed, pre-rendered, json file or something. https://github.com/bazelbuild/rules_rust/issues/653

illicitonion:
> I cannot find what you're referring to in the test.
Oh, yeah, I completely lied, no change needed :)

> They feel quite unmaintainable IMO and would opt to silence it for now. I feel the very next thing that should be done is to update the output to be a well formed, pre-rendered, json file or something. #653

Yeah, I agree :) I'm about to send out a PR with some smaller, much more targeted tests (though they don't fully replace the integration tests yet). FWIW the reason that the existing test is failing is because someone released a new version of `which`, and the test uses floating dependencies, which we shouldn't do! I'll fix it up in a follow-up :)

UebelAndre:
>  FWIW the reason that the existing test is failing is because someone released a new version of `which`, and the test uses floating dependencies, which we shouldn't do! I'll fix it up in a follow-up :)

https://github.com/bazelbuild/rules_rust/issues/654 😄

Issue #659, opened 2021-03-29
`build.rs` files don't detect sysroots


sayrer:
There may be a way to do this already, so apologies if I am unaware of it.

In looking at *ring* (https://github.com/briansmith/ring/issues/1242), I think the ideal way to build it in Bazel would be to use a `cc_library` rule instead of running its `build.rs` file. Is there a way to do that?

illicitonion:
FWIW I've been able to successfully build `ring` with bazel using `crate_universe`, both with and without remote execution - it's possible your cc toolchain isn't configured properly to include some include search paths (or it's possible mine is too permissive and is leaking stuff it shouldn't be!)

You can currently add a dependency on the `cc_library` using an `overrides` of `extra_bazel_deps`, but you can't tell it to skip a `build.rs`. https://github.com/bazelbuild/rules_rust/blob/main/examples/crate_universe/has_aliased_deps/WORKSPACE has an example of doing something similar with `openssl` - `openssl-sys`'s build.rs allows you to opt into pointing at a specific location that a built `openssl` can be found, by setting the `OPENSSL_DIR` env var, so we do that, and point at the directory where the dep will be.

If you wanted to build `ring` as a `cc_library` I'd probably recommend taking a similar approach to `openssl` - making the `build.rs` able to detect a pre-built library, and adding it as a dep, rather than flatly skipping the whole `build.rs` - the `build.rs` may do other things that matter, so skipping it may have unforeseen consequences, either now or in the future as `ring` evolves :)

sayrer:
> (or it's possible mine is too permissive and is leaking stuff it shouldn't be!)

That is likely the case. I'm building with https://github.com/grafica/build/blob/main/ubuntu_20_04_bazel_with_cc_toolchain/Dockerfile

along with this option:

`--incompatible_enable_cc_toolchain_resolution`

and main-branch copies of `bazel` and `rules_rust`, using this toolchain:

https://gist.github.com/sayrer/6b62c1e5385c4961635e3aedee90b6e1

The sysroot is the one that Chromium currently uses.

illicitonion:
It looks like the differences in our environments is that we pre-install [libc6-dev](https://packages.debian.org/jessie/amd64/libc6-dev/filelist) and [linux-libc-dev](https://packages.debian.org/jessie/amd64/linux-libc-dev/filelist) and specify their paths as `cxx_builtin_include_directories`s which you don't - how are you supplying those in your toolchain?

[...]

PR #664, opened 2021-03-30
Add sysroot support to build.rs scripts.

sayrer:
If there's a sysroot, it needs to appear before any other flags. See: #659.


illicitonion (approved):
Thanks!

hlopko:
Could you add a unit test? Hopefully https://github.com/bazelbuild/rules_rust/pull/608 is a good starting example?

sayrer:
> Could you add a unit test? Hopefully #608 is a good starting example?

I'm not really certain how I would add a unit test for this, at least in that style. This patch doesn't change the analysis. I think I'd have to run the binary and check that the sysroot appeared in the command output, since all this does is put the sysroot into an environment variable for use with the binary target.

It might be better to check in a copy of *ring*, or something similarly complicated, and build it with a sysroot.

hlopko:
Yeah you're right, using a cc toolchain with a sysroot in a unittest is hard. Ignore my comment. @illicitonion can you merge this PR (since you approved and didn't merge I don't want to step in)?

illicitonion:
I thought I pressed the merge button when I approved! ... Weird.

illicitonion:
Oh, looks like I don't have permission to merge main - @sayrer could you rebase/merge main and I'll merge to main? Thanks!

sayrer:
Huh, now I'm getting Buildkite complaints about files that aren't in my patch. Will take a look later on today.

UebelAndre:
Seems like buildifier was fixed 😄

sayrer:
Put up #675. It seems like I just got unlucky here.

hlopko:
And we'll need a merge again :/

hlopko:
Thank you!

PR #665, opened 2021-03-30
Renamed BUILD and WORKSPACE files to BUILD.bazel and WORKSPACE.bazel

UebelAndre:
The use of `BUILD`/`WORKSPACE` vs `BUILD.bazel`/`WORKSPACE.bazel` was inconsistent across the repo. This PR updates all files without the `.bazel` extension to use it (which has the higher priority over the two styles. see [here](https://docs.bazel.build/versions/master/build-ref.html#packages)).

Additionally, this PR contains buildifier fixes


illicitonion (approved):

hlopko:
Funny story, we relied on the existing naming of these files - files without `.bazel` extension were used by `blaze` (our internal build of `bazel`) and files with `.bazel` were ignored.

But I understand the consistency argument, and it's on us to fix our undocumented assumptions internally. I would have chosen files without extension (support for `.bazel` suffix was added for repos that need to build with bazel and buck at the same time, which is not the case for `rules_rust`). But we use `raze` and there I could be convinced to generate files with the suffix, and then to be consistent we need suffices. So shrug :)

UebelAndre:
I was not aware of any use cases with Blaze nor did I think these rules could be used with Buck. Is any of that documented somewhere? I definitely wasn't trying to break anyone with this change.

fweikert:
@UebelAndre how did you invoke Buildifier? I'm trying to understand why the formatting problems weren't caught on CI.

UebelAndre:
I ran 
```
buildifier -lint=fix -mode=fix -warnings=all -r ./
```

and I'm using
```
buildifier version: 3.5.0
buildifier scm revision: 10384511ce98d864faf064a8ed54cdf31b98ac04
```

fweikert:
Ok, thank you! https://github.com/bazelbuild/continuous-integration/pull/1129 should fix Buildifier problems on CI.

PR #668, opened 2021-03-31
construct_arguments: Always use correct attributes in an aspect context.

martinboehme:
`construct_arguments` was already using the `rule_attrs` helper function in some places to access `ctx.rule.attr` instead of `ctx.attr`, but it wasn't doing this consistently.

I've chosen to eliminate the boolean `aspect` parameter of `construct_arguments` and pass in the correct attributes directly
instead. This also eliminates the need for the `rule_attrs` helper function. Since this function was only being used here, I have eliminated it entirely.

Finally, I have changed `_get_rustc_env` to take only the attributes instead of the entire context.

All tests continue to pass. I haven't added any tests that would break without this change, but in an upcoming PR, I will change `_get_rustc_env` to respect the `crate_name` attribute, and this would break without the changes in this PR.

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -84,6 +84,7 @@ def _clippy_aspect_impl(target, ctx):
 
     args, env = construct_arguments(
         ctx,
+        ctx.rule.attr,
         ctx.file,
         toolchain,
         toolchain.clippy_driver.path,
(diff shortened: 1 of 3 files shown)


martinboehme:
@hlopko Can you review?

hlopko (approved):
LGTM, thanks!

PR #669, opened 2021-03-31
Respect `crate_name` attribute when setting CARGO_CRATE_NAME.

martinboehme:
As part of this change, I've moved the formerly private `_crate_name` function from rust.bzl to utils.bzl and renamed it to `crate_name_from_attr`.

I've also had to add a `hasattr(attr, "crate_name")` check to this function because `rustc_compile_action` (which calls through to `_get_rustc_env`) gets called from the implementation of the Rust proto rules, and those don't have a `crate_name` attribute (and it wouldn't make sense to add one either).


hlopko (approved):
LGTM!

PR #671, opened 2021-03-31
buildifier test - DO NOT SUBMIT

fweikert:


fweikert:
Great, now we're running into https://github.com/bazelbuild/continuous-integration/issues/1089 :(

PR #672, opened 2021-03-31
Added test script for ensuring docs are up to date.

UebelAndre:
Users should now see an error in their builds when they make a change which invalidates docs
Screen Shot 2021-03-31 at 11 52 59 AM

Additionally, I've regenerated docs since they were out of date and fixed a bug in how docs were generated so we can correctly maintain indentations where desired.

Also I deleted the macos docs test since platform should have no impact on generating docs. This is purely a test of stardoc and the visibility of the bzl files which is platform agnostic.


UebelAndre:
@hlopko Do you have a second to review this PR?

rust/private/rust.bzl
```diff
@@ -521,18 +521,31 @@ def _tidy(doc_string):
...
+        return doc_string
+    else:
+        # Remove the leading block of spaces from the current line
+        block = " " * space_count
+        return "\n".join([line.replace(block, "", 1).rstrip() for line in lines])
 
 _common_attrs = {
     "aliases": attr.label_keyed_string_dict(
-        doc = _tidy("""
+        doc = _tidy("""\
```
hlopko, line 541: Why not skip the first line if empty in `_tidy`and not require \? It seems fragile to remember to put \ there.

hlopko (approved):
LGTM, only nits. Thank you!

rust/private/rust.bzl
UebelAndre: Added support for ignoring first lines.

docs/test_docs.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ -n "${BUILD_WORKSPACE_DIRECTORY:-}" ]]; then
+    DOCS_WORKSPACE="${BUILD_WORKSPACE_DIRECTORY}"
+else
+    # https://stackoverflow.com/a/246128/7768383
```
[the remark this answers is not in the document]
UebelAndre: Updated.

PR #673, opened 2021-03-31
Change transitive_libs in DepInfo to a depset

dfreese:
It appears to already be a depset, just happens to be flattened.  It's
not a part of the external API, so, in theory, this should have an
effect.


hlopko (approved):
Good catch! Thank you.

PR #675, opened 2021-04-01
Fix some buildifier errors.

sayrer:
I think this CI check just started working.


hlopko (approved):
Thank you so much for fixing these! Indeed, the CI check started working recently :)

PR #676, opened 2021-04-01
Updated crate_universe examples

UebelAndre:
Simple cleanup of the examples so they can more effectively be run in CI.


UebelAndre:
@illicitonion do you have a quick moment to take a look here?

illicitonion (approved):
I feel like this makes the examples a lot less clear, but I can see the value for making CI simpler... Happy to merge for now, but I can see maybe reverting this in the future to make the examples more clear...

.bazelignore
```diff
@@ -1,4 +1,2 @@
 docs
 examples
-examples/crate_universe
```
illicitonion, line 3: Does this not need to stay?

UebelAndre: it's a sub-directory of `examples` so it should already be ignored. After quite a bit of testing with this change I've noticed no regressions but the documentation for [bazelignore](https://docs.bazel.build/versions/master/guide.html#bazelignore) is a bit lacking.

UebelAndre:
> I feel like this makes the examples a lot less clear, but I can see the value for making CI simpler... Happy to merge for now, but I can see maybe reverting this in the future to make the examples more clear...

It definitely makes CI easier. The only other option I know of is to have one job per workspace which I would assume has an unfortunate impact on the larger pool of workers. This can definitely be reverted in the future once crate_universe stabilizes a bit and we find this is implementation is undesirable.

Issue #483, opened 2020-11-08
Reintroduce proto_compile_deps and grpc_compile_deps to rust_proto_toolchain


UebelAndre:
@damienmg There is a `TODO` in [proto/toolcahin.bzl](https://github.com/bazelbuild/rules_rust/blob/master/proto/toolchain.bzl#L124-L127)
```python
# TODO(damienmg): Once bazelbuild/bazel#6889 is fixed, reintroduce
# proto_compile_deps and grpc_compile_deps and remove them from the
# rust_proto_library and grpc_proto_library.
rust_proto_toolchain = rule(
```

The issue (https://github.com/bazelbuild/bazel/issues/6889) looks closed now so is this something that should be done still?

damienmg:
Probably yes.

PR #678, opened 2021-04-07
Moved proto deps into toolchain to avoid confusing boiler plate code

UebelAndre:
Users looking to register their own toolchain sometimes stumble over the fact that every use of `rust_grpc_library` or `rust_proto_library` now requires them to also add the appropriate targets to the `rust_deps` attributes. This is easily missed and not necessary. Instead, these dependencies should be added to the `@rules_rust//proto:toolchain` being used so there's a single source of truth for all the dependencies required by these rules.

Additionally, this allows users to optionally __not__ register the default toolchain in cases where they want to define their own but would like to ensure other dependencies are brought in.

closes https://github.com/bazelbuild/rules_rust/issues/483

### DIFF
--- a/docs/flatten.md
+++ b/docs/flatten.md
@@ -505,7 +505,6 @@ Example:
 
 ```python
 load("//proto:proto.bzl", "rust_grpc_library")
-load("//proto:toolchain.bzl", "GRPC_COMPILE_DEPS")
 
 proto_library(
     name = "my_proto",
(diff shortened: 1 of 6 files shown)


UebelAndre:
This is a reboot of https://github.com/bazelbuild/rules_rust/pull/519

illicitonion (approved):
Seems pretty reasonable :)

PR #679, opened 2021-04-07
Added `rust_test_suite` macro

UebelAndre:
This macro helps create test rules for users looking define a suite of integration tests. This is most commonly done by users using a [cargo project layout](https://doc.rust-lang.org/cargo/guide/project-layout.html).


rust/private/rust.bzl
```diff
@@ -1139,6 +1139,89 @@ rust_test_binary = rule(
...
+    Args:
+        name (str): The name of the `test_suite`.
+        srcs (list): All test sources, typically `glob(["tests/**/*.rs"])`.
+        **kwargs (dict): Additional keyword arguments. Args from [rust_test](#rust_test) are parsed out and shared
+            between all tests while [common attributes](https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes)
+            are shared between the test targets and `test_suite`.
+    """
+    tests = []
+
+    # Gather any rust attributes
```
illicitonion, line 1197: I don't think we should be filtering these and trying to treat them specially - I think we should add all attributes to the underlying `rust_test`, and call out anything we put on the `test_suite` explicitly. If someone wants to do something like set `tags` on the tests, I think that's reasonable to let them do...

rust/private/rust.bzl
```diff
@@ -1139,6 +1139,89 @@ rust_test_binary = rule(
...
+                rust_attrs.update({attr: value})
+
+    for attr in _rust_test_attrs.keys():
+        if attr in kwargs:
+            value = kwargs.pop(attr)
+            if value:
+                rust_attrs.update({attr: value})
+
+    for src in srcs:
+        test_name = src.replace(".rs", "").replace("/", "_")
```
illicitonion, line 1212: Maybe error if the file name doesn't end with `.rs`, and just trim the right three characters? `.rs` may appear in a file name other than as an extension :)

illicitonion, line 1212: Is the `/`-stripping required, or just cosmetic? If the latter, let's maybe not do it?

illicitonion (commented):
Looks fine in concept, but I'm not quite sure on a few bits of the implementation...

rust/private/rust.bzl
UebelAndre: It's required, either this or we set an explicit `crate_name` but even then we'd be doing the same thing since `.` is not valid in a crate name.

rust/private/rust.bzl
UebelAndre: I think I've updated this accordingly. Let me know

rust/private/rust.bzl
UebelAndre: Updated

dfreese:
This looks helpful, but is there a specific use case in mind where you're planning to use this?

I had started drafting of some ideas in terms of version for rules_rust, and part of that was being pretty clear about what is in //rust:defs.bzl

Because of that I would like to be a little more clear about necessary versus helpful on that interface.

For example, if this is something that could be generated by cargo raze, I would lean going that way, but if this is a common pattern a lot of folks are using everywhere, then it probably makes sense.

UebelAndre:
unless `cargo-raze` starts generating content for the current repository and not just it's dependencies, this is still something that's very useful. Though, I don't know what would qualify as "necessary". Is the concept of a "test suite" necessary? No. Is it useful? Very. 

I found similar macros in a couple different projects I work on and feel this is something that would be a valuable addition to the core rules given that Rust (Cargo?) has such a strong concept of integration tests. Many users won't throw away the Cargo project structure just because they're using Bazel and the lack of a macro like this feels like a forcing function that goes nowhere.

rust/private/rust.bzl
dfreese: Does that end up changing how you'd refer to the individual test, from a bazel perspective?

for example setting the crate name explicitly would result in this being valid, right?
```bash
bazel test //test/rust_test_suite:tests/integrated_test_a
```

versus, this with the current version:
```bash
bazel test //test/rust_test_suite:tests_integrated_test_a
```

I would (slightly) lean toward the former.

rust/private/rust.bzl
```diff
@@ -1139,6 +1139,82 @@ rust_test_binary = rule(
...
+            srcs = [src],
+            tags = tags,
+            **kwargs
+        )
+        tests.append(test_name)
+
+    native.test_suite(
+        name = name,
+        tests = tests,
+        tags = tags,
```
dfreese, line 1215: Would you be alright going with:
```suggestion
        tags = kwargs.getattr("tags", None),
```
And then removing `tags` from the macro definition.  I've found that tends to reduce the overhead on the macro's definition in the long run as more arguments get shared, but need to have a default.

rust/private/rust.bzl
```diff
@@ -1139,6 +1139,82 @@ rust_test_binary = rule(
         """),
 )
 
+def rust_test_suite(name, srcs, tags = [], **kwargs):
```
dfreese, line 1142:
```suggestion
def cargo_test_suite(name, srcs, **kwargs):
```
Since this is more of a cargo pattern, would you be alright with going with this name instead?

(I added in additional changes based on my comment on line 1215)

rust/private/rust.bzl
UebelAndre: I don't think we can use the base name of the test. What happens in the case where there are two tests in different directories named the same?

rust/private/rust.bzl
UebelAndre: I don't think it's appropriate to prefix this with `cargo` since cargo is not being used in any way here. The design does come from cargo but I feel it's an effective pattern particularly for Rust and it's worth continuing in Bazel.

rust/private/rust.bzl
UebelAndre: This seems to be opposite feedback from what @illicitonion gave. Am I misunderstanding this? Could either of you clarify for me?

dfreese: If I'm interpreting correctly, I think it's slightly different.  In that case, it looks like the `tags` attribute was not being passed to `rust_test`.

What I'm suggesting here doesn't remove the tags from the tests, just changes how they're applied to the test_suit.

rust/private/rust.bzl
```diff
@@ -1139,6 +1139,82 @@ rust_test_binary = rule(
...
+    tests = []
+
+    for src in srcs:
+        if not src.endswith(".rs"):
+            fail("srcs should have `.rs` extensions")
+
+        # The test name should not end with `.rs` or have `/`.
+        test_name = src[:-3].replace("/", "_")
+        rust_test(
+            name = test_name,
```
dfreese, line 1205:
```suggestion
        test_name = src[:-3]
        rust_test(
            name = test_name,
            crate_name = test_name.replace("/", "_"),
```

rust/private/rust.bzl
dfreese: so bazel targets can have `/` characters in them, so I think you could leave them in.  I added a suggestion along the lines of what I was thinking.

rust/private/rust.bzl
UebelAndre: The suggestion didn't quite work but I believe I've addressed this.

rust/private/rust.bzl
UebelAndre: Updated

UebelAndre:
@dfreese ready for another review.

rust/private/rust.bzl
UebelAndre: The targets now have forward slashes to match the directory structure.

rust/private/rust.bzl
dfreese: @illicitonion do you have opinions here?

This currently my only hang-up on the PR, but my opinions aren't too strong on this.

The main thing, in my head is that it's a cargo-style pattern, and that's probably the easiest connection for folks to make, and that `rust_test_suite` seems a little too general for this type of pattern.

UebelAndre: I feel pretty strongly that this shouldn't be `cargo_test_suite`. I think that's pretty misleading. Another name would be fine but I also think `rust_test_suite` is totally appropriate.

dfreese: Given that I haven't seen this sort of test pattern used outside of cargo, and we already have cargo_build_script, I'm not sure I understand the strong opposition.

Are there public examples of where this would be use in a context of a project that isn't also building with Cargo?

UebelAndre: I do not have public examples. The difference between this test suite and cargo_build_script is that build.rs is a concept invented by cargo that would never have naturally occurred while integration tests are. Do you have any rust projects with integration tests? How did you implement those?

illicitonion: tl;dr: `rust_test_suite` works for me.

I would definitely feel strange about naming this `cargo_test_suite`, as there nothing cargo-y about it other than that it happens to be useful if you happen to use the cargo file layout - it would equally apply without the `tests` dir, e.g.
```python
rust_test_suite(
    name = "tests",
    srcs = ["*_test.rs"],
)
```

The `${language}_test` -> `${language}_test_suite` convention is pretty common (e.g. [`scala_test_suite`](https://github.com/bazelbuild/rules_scala/blob/master/docs/scala_test_suite.md) for `scala_test`. Apparently `java_test_suite` didn't make it from `blaze` to `bazel`, but my company certainly has a `java_test_suite` macro which looks a lot like this does but for Java)...

rust/private/rust.bzl
```diff
@@ -1139,6 +1141,81 @@ rust_test_binary = rule(
...
+    )
+    ```
+
+    [it]: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html
+    [ts]: https://docs.bazel.build/versions/master/be/general.html#test_suite
+
+    Args:
+        name (str): The name of the `test_suite`.
+        srcs (list): All test sources, typically `glob(["tests/**/*.rs"])`.
+        **kwargs (dict): Additional keyword arguments. Args from [rust_test](#rust_test) are parsed out and shared
```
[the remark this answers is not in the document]
UebelAndre: Good catch, updated.

dfreese (approved):

PR #681, opened 2021-04-07
Fixed error handling in crate_universe build script.

UebelAndre:
https://github.com/bazelbuild/rules_rust/actions/runs/726775754 succeeded but actually failed to push changes to the `main` branch which should have generated a failure. This change fixes that.


illicitonion (approved):
Works for me - may be worth throwing in a `set -e` at the top of the script too :)

UebelAndre:
> Works for me - may be worth throwing in a `set -e` at the top of the script too :)

I'd rather leave it as is since I have the 'retry' logic there. I'll merge this for now just so the issue doesn't continue to happen.

PR #686, opened 2021-04-08
Remove commit and push step of actions

illicitonion:
We need to work out how users should consume these binaries, but for now
this reliably fails so let's stop doing it.

Instead, `cat` out what the the contents of the generated file _would_
be, for easy copy and paste by people working on the rules.


dfreese (approved):

UebelAndre:
I'm still not convinced this is a bad thing to do. This is the reality of having a rust binary being used in a repository rule. Instead of removing it I'd like to discuss alternatives

illicitonion:
> I'm still not convinced this is a bad thing to do. This is the reality of having a rust binary being used in a repository rule. Instead of removing it I'd like to discuss alternatives.
> 
> I'd ask @dfreese for some input since he originally raised concerns.

We should definitely have a discussion about alternatives (and we may re-add this code as a result), but for now we have code which we know is broken (because of permissions issues), and has had reasonable objections raised about it being problematic if it were to start working, so let's remove it while we have that discussion :)

PR #692, opened 2021-04-17
`crate_universe` now renders json formatted lockfiles

UebelAndre:
Notable changes in this PR:
- `crate_universe` `lockfile` attributes are now rendered as json files instead of starlark dumps
- `crate_universe_resolver` now accepts a `repository_dir` argument for the path to the new repository. This should be preferable to a single output file for all dependencies.
- The `cargo-raze` dependenciy has been updated to a pin which implements the required sorting and de-serialization.
- Formatted `use` declarations
- Updated dependencies
- `CARGO` and `RUSTC` environment variables are now ignored in determining the digest. This should be safe given that `cargo version` is an input to generating the lockfile hash.
- `repository_template` has been renamed to `crate_registry_template`, which is more correct.
- The [json](https://docs.bazel.build/versions/master/skylark/lib/json.html) module is now used to handle generating the config `.json` file

### DIFF
--- a/crate_universe/src/main.rs
+++ b/crate_universe/src/main.rs
@@ -1,10 +1,8 @@
-use std::{
-    io::{BufRead, BufReader},
-    path::PathBuf,
-};
+use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context};
-use crate_universe_resolver::config::Config;
+use crate_universe_resolver::{config::Config, renderer::Renderer};
+use indoc::indoc;
 use log::*;
 use structopt::StructOpt;
 
@@ -19,7 +17,9 @@ struct Opt {
     #[structopt(long = "input_path", parse(from_os_str))]
     input_path: PathBuf,
     #[structopt(long = "output_path", parse(from_os_str))]
-    output_path: PathBuf,
+    output_path: Option<PathBuf>,
+    #[structopt(long = "repository_dir", parse(from_os_str))]
+    repository_dir: PathBuf,
     #[structopt(long = "lockfile", parse(from_os_str))]
     lockfile: Option<PathBuf>,
     #[structopt(long = "update-lockfile")]
@@ -30,7 +30,6 @@ fn main() -> anyhow::Result<()> {
     env_logger::init();
 
     let opt = Opt::from_args();
-
     trace!("Parsing config from {:?}", opt.input_path);
 
     let config: Config = {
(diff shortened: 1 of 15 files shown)


UebelAndre:
This lack of 4.0 is really driving me insane....

crate_universe/src/main.rs
```diff
@@ -16,7 +17,9 @@ struct Opt {
     #[structopt(long = "input_path", parse(from_os_str))]
     input_path: PathBuf,
     #[structopt(long = "output_path", parse(from_os_str))]
-    output_path: PathBuf,
+    output_path: Option<PathBuf>,
```
illicitonion, line 20: Let's support one or the other option here, I don't see value in supporting both. Can you either remove `output_path`, or remove `repository_dir`?

It looks like long term you're going to prefer `repository_dir`, so let's maybe drop `output_path`, and just change what we generate in the `repository_dir` in the future?

illicitonion (commented):
Thanks for pushing this forwards!

crate_universe/src/main.rs
UebelAndre: I was going to remove `output_path` but then all the tests started to fail and I didn't want to go down that rabbit hole just yet. I do think using `repository_dir` is the better thing though and have added comments to indicate this. My plan was to address this in a second PR.

crate_universe/src/main.rs
```diff
@@ -36,49 +38,55 @@ fn main() -> anyhow::Result<()> {
...
-        } else {
-            return Err(anyhow!("Invalid lockfile"));
-        }
+        return reuse_lockfile(config, &lockfile, &opt);
+    }
+
+    generate_dependencies(config, &opt)
+}
+
+fn reuse_lockfile(config: Config, lockfile: &Path, opt: &Opt) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
UebelAndre: What you're describing sounds like a good approach to me. I ended up doing this because I thought it was the simpler change and the fastest path to getting well formed lockfiles. Most of the time on this PR was spent trying to figure out what was going on so definitely a more incremental change than a strong directional one.

crate_universe/src/main.rs
illicitonion: In that case, let's maybe drop the `repository_dir` stuff, and fully implement that in one PR, rather than half doing it now? (Sorry, this change already does quite a few things, so it'd be good to avoid half-doing this one :))

crate_universe/src/main.rs
```diff
@@ -36,49 +38,55 @@ fn main() -> anyhow::Result<()> {
...
+    let mut resolver = config.preprocess()?;
+
+    let renderer = Renderer::new_from_lockfile(lockfile)?;
+
+    // TODO: Add lockfile versioning and check that here
+
+    if !renderer.matches_digest(&resolver.digest()?) {
+        return Err(anyhow!(indoc! { r#"
+            "rules_rust_external: Lockfile at {} is out of date, please either:
+            1. Re-run bazel with the environment variable `RULES_RUST_UPDATE_CRATE_UNIVERSE_LOCKFILE=true`, to update the lockfile
```
illicitonion, line 65:
```suggestion
            1. Re-run bazel with the environment variable `RULES_RUST_REPIN=true`, to update the lockfile
```

Again, not your fault, but turns out we changed this env var name but not the warning message!

crate_universe/src/main.rs
UebelAndre: Yeah, I didn't realize how little `repository_dir` was used after reviving `output_path` for tests. Removed it.

illicitonion (approved):
LGTM - would appreciate applying my last suggestion before merging, if you don't mind!

PR #694, opened 2021-04-19
Avoid downloading multiple resolvers

### LINKED ISSUE #693: `crate_universe` currently downloads a resolver for every triple
Rather than setting up a `_resolvers` attribute which points at multiple `http_file` repository targets, we should probably switch to do one `repository_ctx.download` constructing the URL and sha256 to download from attributes on `crate_universe` with defaults in the repo.

illicitonion:
Instead, download just the one that's needed for the host platform.

Fixes #693


illicitonion:
> Does this mean each `crate_universe` repository will download it's own binary?

Ish - if you set the sha256, bazel should deduplicate the downloads. If you don't, it will probably perform multiple downloads.

> it might be worth having a second repository rule for downloading the resolver and have `crate_universe` proper depend on that.
> 
> I feel like this would be a pretty quick change, it's just moving the downloading logic into it's own rule. Your thoughts?

I'd rather not invent a separate "avoid downloading the same file multiple times" mechanism, given the sha256 deduping is designed for that, but can add one if you think it'd be a big problem.

UebelAndre:
> Ish - if you set the sha256, bazel should deduplicate the downloads. If you don't, it will probably perform multiple downloads.

If Bazel will use a cached binary based on the `sha256` value then that works for me 😄 


> I'd rather not invent a separate "avoid downloading the same file multiple times" mechanism, given the sha256 deduping is designed for that, but can add one if you think it'd be a big problem.

No, if there's an existing mechanism for doing this that's all I care about.

UebelAndre (approved):

UebelAndre:
Oh crap, this broke bootstrapping. Totally didn't think about that 😞 

Can you add a followup PR to update the bootstrap script so users can easily iterate on changes?

UebelAndre:
I opened https://github.com/bazelbuild/rules_rust/issues/697 to track the over all plan for CI, but I had setup the bootstrapping to initially solve for this. I wanted to open a PR which runs the bootstrapping script as a [shell_command](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#running-commands-shell-scripts-or-binary-targets) when building [examples/crate_universe](https://github.com/bazelbuild/rules_rust/tree/main/examples/crate_universe) which would have run as it's own job. Maybe that's a bad idea? We should continue this conversation on the issue I opened or some other one if we want to be more granular.

PR #698, opened 2021-04-20
Crate universe now renders separate BUILD files for each dependency

UebelAndre:
This change splits all the crate definitions into unique `BUILD` files. The repository directory is updated from
```
├── BUILD.bazel
├── WORKSPACE
├── defs.bzl
├── resolver
└── uses_sys_crate_deps.resolver_config.json
```
to
```
├── BUILD.bazel
├── BUILD.bzip2-0.3.3.bazel
├── BUILD.bzip2-sys-0.1.10+1.0.8.bazel
├── BUILD.cc-1.0.67.bazel
├── BUILD.libc-0.2.76.bazel
├── BUILD.pkg-config-0.3.19.bazel
├── WORKSPACE
├── defs.bzl
├── resolver
└── uses_sys_crate_deps.resolver_config.json
```

In my experience working on `cargo-raze` this makes testing and debugging much easier.


crate_universe/src/renderer.rs
```diff
@@ -334,6 +341,42 @@ impl Renderer {
...
+                .render("templates/BUILD.crate.bazel.template", &context)?;
+
+            let build_file_path = repository_dir.join(format!(
+                // This must match the format found in the repository rule templates
+                "BUILD.{}-{}.bazel",
+                crate_data.crate_context.pkg_name, crate_data.crate_context.pkg_version
+            ));
+            let mut build_file = File::create(&build_file_path).with_context(|| {
+                format!(
+                    "Could not create lockfile file: {}",
```
illicitonion, line 370:
```suggestion
                    "Could not create BUILD file: {}",
```

illicitonion (commented):
This generally looks good, but I'm concerned that we're dropping down to zero test coverage for BUILD file rendering. Can you add a test which asserts on the output of `renderer`, given a single hard-coded `RenderablePackage`?

This can be nicely reproducible, deterministic, and hard-coded, and I'm ok with deleting `integration.rs` after that (though we should add some unit tests of resolver at some point), but dropping to 0 coverage feels bad.

UebelAndre:
I'll see what I can add. I initially tried adding the unit tests but felt that was a pretty decent rabbit hole and thought it would be better to do that in a dedicated PR. The changes can still be functionally tested using the examples so I wouldn't necessarily say there's zero coverage but definitely the state of testing right now is not ideal.

UebelAndre:
I'm also a bit conflicted because a lot of this logic can be found in cargo-raze so getting deja-vu when looking to add tests.

UebelAndre:
@illicitonion Is this a sufficient replacement of the tests?

crate_universe/src/renderer.rs
```diff
@@ -485,3 +526,285 @@ enum Kind {
...
+            .render_workspaces(&mut output)
+            .expect("Error rendering");
+
+        let output = String::from_utf8(output).expect("Non-UTF8 output");
+
+        // TODO: Don't unconditionally load new_git_repository and http_archive
+        let expected_repository_rule = indoc! { r#"
+            # rules_rust crate_universe file format 1
+            # config hash 598
+
```
illicitonion, line 609: We should drop these lines from the output, they're only relevant for the lockfile which no longer uses them.

crate_universe/tests/integration.rs
```diff
@@ -13,6 +13,7 @@ use crate_universe_resolver::config::{Config, Override, Package};
 use crate_universe_resolver::NamedTempFile;
 
 #[test]
+#[ignore]  // TODO: This test fails because of the way `test` is implemented, see that method for details
```
illicitonion, line 16: Let's delete this whole file, no point keeping around a whole file of broken tests :)

illicitonion (approved):
Thanks!

PR #699, opened 2021-04-20
Fix bootstrapping since #694

illicitonion:


UebelAndre (approved):
Thanks for fixing this!

UebelAndre:
Not super problematic, but wanted to share https://github.com/bazelbuild/bazel/issues/12689

UebelAndre:
And looks like it was addressed in https://github.com/bazelbuild/bazel/commit/c2bdd034014f66ce14529cc353cda18a32320f6c

PR #703, opened 2021-04-22
rustc_src is optional

illicitonion:
This is a tar containing a huge number of files (hundreds of thousands)
which can take double-digit seconds to extract.

It's only used for rust analyzer support, so hide this behind a flag
people can opt in to if they want.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -34,6 +34,7 @@ def rust_repositories(
         edition = None,
         dev_components = False,
         sha256s = None,
+        include_rustc_src = False,
         urls = DEFAULT_STATIC_RUST_URL_TEMPLATES):
     """Emits a default set of toolchains for Linux, MacOS, and Freebsd
 
@@ -60,6 +61,7 @@ def rust_repositories(
         edition (str, optional): The rust edition to be used by default (2015 (default) or 2018)
         dev_components (bool, optional): Whether to download the rustc-dev components (defaults to False). Requires version to be "nightly".
         sha256s (str, optional): A dict associating tool subdirectories to sha256 hashes. Defaults to None.
+        include_rustc_src (bool, optional): Whether to download rustc's src code. This is required in order to use rust-analyzer support. Defaults to False.
         urls (list, optional): A list of mirror urls containing the tools from the Rust-lang static file server. These must contain the '{}' used to substitute the tool being fetched (using .format). Defaults to ['https://static.rust-lang.org/dist/{}.tar.gz']
     """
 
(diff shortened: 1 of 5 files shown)


rust/repositories.bzl
```diff
@@ -34,6 +34,7 @@ def rust_repositories(
         edition = None,
         dev_components = False,
         sha256s = None,
+        experimental_rust_analyzer_support = False,
```
UebelAndre, line 37: Do we consider `rust_analyzer` experimental? But also I feel like this could be `include_rustc_srcs` and we comment that it's useful for when you're using rust analyzer.

UebelAndre (commented):
Perhaps `rust_analyzer` should have it's own toolchain which goes and fetches the rustc sources? That might make configuring it a bit clunky but might be warranted and provide a good way to separate the `rust_analyzer` functionality from core builds. Maybe there's a path forward where rust analyzer is enabled by a `bazelrc` flag and we'd always define the toolchains in `repositories.bzl` knowing they'd otherwise never be used without the aspect enabled.

I do feel the amount of time spent fetching dependencies is quite painful though so this is a good direction.

illicitonion:
I finally got around to addressing these comments - PTAL!

UebelAndre (changes requested):
What is the user story for having this work with `rust-analyzer`? I think if we expect users to set some flag to `True` in their workspace that we should also update documentation for that here at the very least.

illicitonion:
I added a note to the `rust_analyzer` docs to tell users they need to set this.

UebelAndre:
I was actually thinking something should be added to [docs/rust_analyzer.vm](https://github.com/bazelbuild/rules_rust/blob/fdc04e159c295d0f24d55fecf76f95b2034ddcc5/docs/rust_analyzer.vm). If I'm going to setup rust_analyzer in my existing rules_rust project, I may not see the updated note about `rust_repositories`. If you add some documentation there I'm happy with this PR 😄

illicitonion:
Sigh, I edited exactly what you wanted, but I edited the .md file rather than the .vm file so it got reverted by the re-generation. I am looking forward to not putting generated docs in the repo.

Updated the right file this time :)

UebelAndre (approved):
Looks good to me! Thanks!

UebelAndre:
Having thought about this a bit more. Could this also be gated by an `action_env` or something a user could add to their `.bazelrc` files? Basically, I'm never going to set `include_rustc_src = False` if I use `rust_analyzer` in development or I'm gonna awkwardly have a modified WORKSPACE file or something. Maybe [--repo_env](https://docs.bazel.build/versions/master/command-line-reference.html#flag--repo_env) would work here?

PR #705, opened 2021-04-22
Add generated docs for `crate_universe`

UebelAndre:
Additionally, this renames `crate.override.extra_rust_env_vars` to `crate.override.extra_rustc_env_vars` to be more consistent with the behavior of that field

### DIFF
--- a/crate_universe/defs.bzl
+++ b/crate_universe/defs.bzl
@@ -232,6 +232,8 @@ crate_universe = repository_rule(
     doc = """\
 A rule for downloading Rust dependencies (crates).
 
+__WARNING__: This rule experimental and subject to change without warning.
+
 Environment Variables:
 - `REPIN`: Re-pin the lockfile if set (useful for repinning deps from multiple rulesets).
 - `RULES_RUST_REPIN`: Re-pin the lockfile if set (useful for only repinning Rust deps).
@@ -301,6 +303,8 @@ def _spec(
         features = None):
     """A simple crate definition for use in the `crate_universe` rule.
 
+    __WARNING__: This rule experimental and subject to change without warning.
+
     Example:
 
     ```python
@@ -340,6 +344,8 @@ def _override(
         extra_rustc_env_vars = None,
         features_to_remove = []):
     """A map of overrides for a particular crate
+
+    __WARNING__: This rule experimental and subject to change without warning.
 
     Example:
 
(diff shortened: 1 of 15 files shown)


UebelAndre:
These docs aren't surfaced in the top level page, but are included in `flatten.md`. I figured since the rules were still experimental that we might not want to advertise them yet? I'm happy to add that though.

crate_universe/defs.bzl
```diff
@@ -322,20 +298,108 @@ def _spec(
...
         extra_bazel_data_deps = None,
+        extra_bazel_deps = None,
         extra_build_script_bazel_data_deps = None,
+        extra_build_script_bazel_deps = None,
+        extra_build_script_env_vars = None,
+        extra_rustc_env_vars = None,
         features_to_remove = []):
+    """A map of overrides for a particular crate
+
+    example:
```
illicitonion, line 343:
```suggestion
    Example:
```

crate_universe/defs.bzl
```diff
@@ -322,20 +298,108 @@ def _spec(
         name,
         semver,
         features = None):
+    """A simple crate definition
```
illicitonion, line 301:
```suggestion
    """A simple crate definition for use in the `crate_universe` rule.
```

crate_universe/defs.bzl
```diff
@@ -322,20 +298,108 @@ def _spec(
...
+
+    example:
```
illicitonion, line 303:
```suggestion
    Example:
```

crate_universe/defs.bzl
```diff
@@ -236,78 +235,55 @@ A rule for downloading Rust dependencies (crates).
...
-                # ...
-            },
-            extra_build_script_bazel_data_deps = {
-                # ...
-            },
-        ),
-    },
-)
-```
+- `RULES_RUST_CRATE_UNIVERSE_RESOLVER_URL_OVERRIDE`: Override URL to use to download resolver binary
```
illicitonion, line 238: Can you add a note at the top of this docstring highlighting that this rule is experimental and subject to breaking changes at any time?

illicitonion (approved):
Thanks!

crate_universe/defs.bzl
UebelAndre: I can add that here, __or__, I can do so in a header template after https://github.com/bazelbuild/rules_rust/pull/706 what do you think?

UebelAndre: Ah, didn't think about flatten.md. Lame but makes sense

UebelAndre: Updated

illicitonion (approved):

PR #706, opened 2021-04-23
Added support for header_templates in generated docs

UebelAndre:
# Changes
- Pages now render with the title of the page and not the previously hard coded `Rust Rules`
- Users can now optionally pass a [header_template](https://github.com/bazelbuild/stardoc/blob/d93ee5347e2d9c225ad315094507e018364d5a67/docs/stardoc_rule.md#stardoc-header_template)


illicitonion:
Mechanically this seems reasonable, but I don't really know stardoc or anything about our documentation situation, so I'll leave for others to weigh in :)

UebelAndre:
> Mechanically this seems reasonable, but I don't really know stardoc or anything about our documentation situation, so I'll leave for others to weigh in :)

I'll ask David. I figured you'd be a good reviewer since you had just reviewed the crate universe docs PR. But I realize that wasn't a change to docs generation 😅

dfreese:
Will try and take a look at this this evening.

UebelAndre:
For a bit of extra context, I'm wanted this functionality to be able to add something like https://github.com/UebelAndre/rules_rust/blob/docs/docs/index.md#experimental-rules

If we can have custom headers then we can add more contextual documentation about rules there and keep the `doc` attributes of each rule more focused and avoid polluting the landing page.

docs/BUILD.bazel
```diff
@@ -112,7 +134,7 @@ genrule(
     name = "flatten_header_vm",
     outs = ["flatten_header.vm"],
     cmd = ">$@ echo '# Rust rules\n\n%s\n'" % "\n".join(
-        sorted(["* [{rule}](#{rule})".format(rule = v) for k in PAGES.keys() for v in PAGES[k]]),
+        sorted(["- [{rule}](#{rule})".format(rule = v) for k in PAGES.keys() for v in PAGES[k].symbols]),
```
dfreese, line 137: why change `*` to `-` when it renders the same way?

dfreese (commented):
Sorry for the delay, family stuff took precedence.

Generally looks fine to me.  One question.

docs/BUILD.bazel
UebelAndre: It's the default behavior of https://github.com/DavidAnson/markdownlint which seems to have some good justifications for each rule it has so I thought it'd be an improvement.

dfreese: Checked a couple other places throughout the repository, as well as my teams, and `*` is used pretty much exclusively.

Looking at [M0004](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004), that repository's default "consistent" not any particular item (though asterisk is listed next).

For familiarity of folks here and reducing churn, would you mind going with `*`?

UebelAndre:
@dfreese ready for another review

dfreese (approved):

PR #707, opened 2021-04-23
Added crate_universe examples to CI

UebelAndre:
This change introduces a new `crate_universe_bootstrap` macro which defines a repository rule for bootstrapping `crate_universe_resolver` binaries. `crate_universe` can now be made to bootstrap a binary (instead of downloading one) by using the `RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP` environment variable and adding the `crate_universe_bootstrap` macro to your `WORKSPACE` file.

The goal of this PR is to enable testing for `crate_universe` which was originally solved by the `./bootstrap.sh` script but this unfortunately lead to many issues. Using a repository rule to compose all the work needed for bootstrapping basically solved for all of these.


crate_universe/bootstrap.bzl
```diff
@@ -0,0 +1,96 @@
...
+        "--manifest-path",
+        repository_ctx.path(repository_ctx.attr.manifest),
+    ]
+
+    repository_ctx.report_progress("bootstrapping crate_universe_resolver")
+    result = repository_ctx.execute(
+        args,
+        environment = {
+            "RUSTC": str(rustc_path),
+            "RUST_LOG": "info",
```
illicitonion, line 55:
```suggestion
```

IIRC this doesn't do anything for `cargo`/`rustc`?

crate_universe/bootstrap.bzl
```diff
@@ -0,0 +1,96 @@
...
+
+    repository_ctx.file("BUILD.bazel", _BUILD_FILE_CONTENT.format(
+        ext = extension,
+    ))
+
+_crate_universe_resolver_bootstrapping = repository_rule(
+    doc = "A rule for bootstrapping a crate_universe_resolver binary",
+    implementation = _crate_universe_resolver_bootstrapping_impl,
+    attrs = {
+        "lockfile": attr.label(
```
illicitonion, line 75: I'd be tempted to call this `cargo_lockfile` as it conflicts with the current crate_universe concept of a `lockfile`.

crate_universe/defs.bzl
```diff
@@ -24,7 +17,7 @@ def _input_content_template(ctx, name, packages, cargo_toml_files, overrides, re
         overrides (dict): A dict of crate name (`str`) to json encoded `crate.override` data.
         registry_template (str): A crate registry url template
         targets (list): A list of target platform triples
-        cargo_bin_label (Label): The label of a Cargo binary.
+        cargo_bin_path (Label): The label of a Cargo binary.
```
illicitonion, line 20: This doc probably needs a little updating - it's now a string (so doesn't need stringifying later?) rather than a Label?

examples/crate_universe/WORKSPACE.bazel
```diff
@@ -9,6 +9,10 @@ load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 
 rust_repositories()
 
+load("@rules_rust//crate_universe:bootstrap.bzl", "crate_universe_bootstrap")
+
+crate_universe_bootstrap()
```
illicitonion, line 14: I think this will cause a second build of the resolver if `RULES_RUST_CRATE_UNIVERSE_RESOLVER_URL_OVERRIDE` is set? Can we sniff for that env var in the repository rule to make it no-op, rather than building a binary we'll then not use?

.bazelci/presubmit.yml
```diff
@@ -109,7 +109,53 @@ tasks:
...
+    working_directory: examples/crate_universe
+    build_targets:
+      - "//..."
+    test_targets:
+      - "//..."
+  crate_universe_rbe_ubuntu1604:
+    name: Crate Universe Examples
+    platform: rbe_ubuntu1604
+    environment:
+      RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP: true
```
illicitonion, line 126: We should probably remove the checked in lockfile for the `uses_sys_crate` example.

Right now, for a `crate_universe_resolver` binary built in release mode (as the bootstrapper does), the `crate_universe_resolver` version number is what's used in the cache key hash to factor in the resolver's logic. Accordingly, we should either use a unique version number per CI run (e.g. by appending `-${git_hash}` to the version number in the `Cargo.toml`), or we should force the example to regenerate its lockfile every time by setting `RULES_RUST_REPIN=true` for these CI jobs.

But if nothing is going to be automatically updating the lockfile when we change things, it's probably best to just delete it for now, while things are in flux.

illicitonion (commented):
This looks great, surprisingly nice and clean! Thanks for putting so much effort into it, it definitely paid off!

Can you also remove the now-stale bootstrap bazelrc files from the .gitignores? When testing this out locally they tripped me up, and if they'd shown up in my `git status` it would've saved me some head-scratching :)

crate_universe/defs.bzl
UebelAndre: Fixed but `cargo_bin_path` is a [path](https://docs.bazel.build/versions/master/skylark/lib/path.html), not a string
```
DEBUG: /private/var/tmp/_bazel_user/5f9b22deab82162fb4591147bb665052/external/rules_rust/crate_universe/defs.bzl:26:10: path
```

crate_universe/bootstrap.bzl
UebelAndre: I feel given the context (doc and attribute requirements) that it's clear what should go here. I can update it if you feel strongly but I'd like to avoid redundancy if possible.

examples/crate_universe/WORKSPACE.bazel
UebelAndre: Bootstrap will now only build if the environment variable is set.

.bazelci/presubmit.yml
UebelAndre: I don't think we should be regenerating a lockfile and I don't think it's the best idea to bake the version of the resolver into the lockfile. I feel the lockfile should be probably have it's own version and only be updated when changes are made to the resolver which would actually impact the lockfile. But for the purposes of this PR, I'd like to have something which uses the lockfile but doesn't update it. This way we can identify drift in the structure of the lockfiles and keep CI more deterministic.

That said, what do you think I should do?

illicitonion: > I feel the lockfile should be probably have it's own version and only be updated when changes are made to the resolver which would actually impact the lockfile.

I agree with this :)

> But for the purposes of this PR, I'd like to have something which uses the lockfile but doesn't update it.

I think in practice that means every PR which modifies `crate_universe` will need to bump the version number in the `Cargo.toml` of the resolver, and re-generate the lockfile. Which feels like a bad experience, which is why I'm suggesting deleting it for now.

Right now things happen to be green because the checked in lockfile happens to be compatible with the rust code, but I imagine that will stop being the case almost immediately, and the resolver will try to use incompatible values and it will be annoying.

> This way we can identify drift in the structure of the lockfiles and keep CI more deterministic.

I like these as goals, but I don't think the lockfile stuff is mature enough for it to actually work out in practice yet.

crate_universe/bootstrap.bzl
illicitonion: I feel pretty strongly about it - in particular, I can imagine a future bootstrapping world where we used `rules_rust` to build the resolver, and where we actually did use a `crate_universe` lockfile to do so. I don't think the naming/context is sufficiently uniquely disambiguating :)

.bazelci/presubmit.yml
UebelAndre: Removed the lockfiles

crate_universe/bootstrap.bzl
UebelAndre: Updated

UebelAndre:
@illicitonion ready for another pass 😄

illicitonion (approved):
Looks great - thanks!

PR #709, opened 2021-04-25
util/process_wrapper: system_posix.cc: use waitpid correctly.

### LINKED ISSUE #708: Compilation errors hidden on Linux/MacOS
Consider the following malformed `main.rs`:

```rust
fn main() {
  println!("ok"
}
```

and the corresponding BUILD file:

```py
load("@rules_rust//rust:rust.bzl", "rust_binary")

rust_binary(
    name = "main",
    srcs = [
        "main.rs",
    ],
)
```

I expect `bazel build //:main` to fail with an error pointing out the missing closing paren. Instead, on Linux/MacOS, all I see is this:

```
$ bazel build //:main 
ERROR: BUILD:3:12: output 'main' was not created
ERROR: BUILD:3:12: Compiling Rust bin main (1 files) failed: not all outputs were created or valid
Target //main:main failed to build
```

The action correctly fails, but the `rustc` stdout/stderr containing the compilation errors is not printed to the console.

This is a severe usability problem: it means no compiler errors are ever shown to users on Linux/MacOS.

After some debugging, I believe the issue lies in how `system_posix.cc` handles the `rustc` exit status:

https://github.com/bazelbuild/rules_rust/blob/main/util/process_wrapper/system_posix.cc#L177-L182

This code uses `waitpid` incorrectly. The value placed in the `exit_status` variable by `waitpid` is not actually the child's exit status. It is an opaque integer from which the actual exit status can be extracted via the `WEXITSTATUS` macro ([documentation](https://linux.die.net/man/2/waitpid)). As a result, I believe Bazel gets confused about the exit status of the action as a whole, and does not print its stdout/stderr.

When I replace `return exit_status` by `return WEXITSTATUS(exit_status)`, Bazel prints the compilation error to the console as expected.

Ubehebe:
Fixes #708.

Verified that this PR compiles and works as expected on Linux and MacOS.

### DIFF
--- a/util/process_wrapper/system_posix.cc
+++ b/util/process_wrapper/system_posix.cc
@@ -17,6 +17,7 @@
 
 // posix headers
 #include <fcntl.h>
+#include <signal.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
--- a/util/process_wrapper/system_posix.cc
+++ b/util/process_wrapper/system_posix.cc
@@ -174,12 +174,18 @@ int System::Exec(const System::StrType &executable,
     }
   }
 
-  int err, exit_status = -1;
+  int err, exit_status;
   do {
     err = waitpid(child_pid, &exit_status, 0);
   } while (err == -1 && errno == EINTR);
 
-  return exit_status;
+  if (WIFEXITED(exit_status)) {
+    return WEXITSTATUS(exit_status);
+  } else if (WIFSIGNALED(exit_status)) {
+    raise(WTERMSIG(exit_status));
+  } else if (WIFSTOPPED(exit_status)) {
+    raise(WSTOPSIG(exit_status));
+  }
 }
 
 } // namespace process_wrapper


util/process_wrapper/system_posix.cc
```diff
@@ -174,12 +175,18 @@ int System::Exec(const System::StrType &executable,
...
   }
 
-  int err, exit_status = -1;
+  int err, exit_status;
   do {
     err = waitpid(child_pid, &exit_status, 0);
   } while (err == -1 && errno == EINTR);
 
-  return exit_status;
+  if (WIFEXITED(exit_status)) {
```
UebelAndre, line 183: What is the fall-through case? Is it possible that we reach the end of the function without returning anything?

UebelAndre (changes requested):
I've asked another maintainer who I think would have some context here. Otherwise I have one question about the change 😄

util/process_wrapper/system_posix.cc
hlopko: +1, I'd add a fall through case. Otherwise LGTM, great work!

hlopko:
Friendly ping. I'm happy to take over if you want to, just say so please.

hlopko:
Thank you again @Ubehebe for this contribution! I took the liberty of taking over your change in https://github.com/bazelbuild/rules_rust/pull/755. I'll close this PR now.

PR #712, opened 2021-04-27
Run basic crate_universe example on Windows

illicitonion:
We already disable all sys crates from our tests, so this is keeping in
that fashion. Looks like the issues here aren't crate_universe specific.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -145,17 +145,18 @@ tasks:
     environment:
       RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP: true
     working_directory: examples/crate_universe
+    build_flags:
+      - "--enable_runfiles" # this is not enabled by default on windows and is necessary for the cargo build scripts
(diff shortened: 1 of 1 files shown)


.bazelci/presubmit.yml
```diff
@@ -149,10 +149,8 @@ tasks:
       - "//..."
       # TODO: There are windows specific build issues in the generated
       # depednencies. This should be fixed and all testing enabled.
-      - "-//basic:test"
       - "-//has_aliased_deps/..."
       - "-//uses_proc_macro/..."
-      - "-//uses_sys_crate/..."
     build_targets: *crate_universe_windows_targets
     # TODO: fix test targets in `crate_universe_windows_targets`
     # test_targets: *crate_universe_windows_targets
```
UebelAndre, line 156: This should be uncommented too if you wanna see the failures from the test target.

UebelAndre (approved):
Nice, forgot about the run files flag. Thanks!

PR #714, opened 2021-04-29
Escape trailing backslashes in build script env vars

### LINKED ISSUE #710: proc-macro-nested broken on Windows due to processwrapper env escaping
This seems like an unfortunate interaction with process wrapper's file handling.

proc-macro-nested calls this line in its build script:

```rust
    println!("cargo:rustc-env=PATH_SEPARATOR={}", path::MAIN_SEPARATOR);
```

It does this so that it can reference the correct platform separate in a macro:

```rust
include!(concat!(env!("OUT_DIR"), env!("PATH_SEPARATOR"), "count.rs"));
```

While Rust can handle `/` on Windows for a normal path, that apparently does not work on UNC network drives. This is covered on a Rust issue:
https://github.com/rust-lang/rust/issues/75075

Process wrapper reads the env file in and applies special logic to a trailing backslash:

https://github.com/bazelbuild/rules_rust/blob/920256900c367357bf4dae5719b8ebf210e91a0f/util/process_wrapper/utils.cc#L72

Should we be escaping a trailing backslash when present in rustc-env lines? Or is there other code that depends on the existing behaviour?

At least in this case, proc-macro-nested seems to be a workaround for pre-1.45 Rust, and the pinning to the previous 0.1.6 version resolves the issue for now, so maybe it's not a big issue in practice.

illicitonion:
Fixes #710.

### DIFF
--- a/cargo/cargo_build_script_runner/lib.rs
+++ b/cargo/cargo_build_script_runner/lib.rs
@@ -125,7 +125,7 @@ impl BuildScriptOutput {
         v.iter()
             .filter_map(|x| {
                 if let BuildScriptOutput::Env(env) = x {
-                    Some(Self::redact_exec_root(env, exec_root))
+                    Some(Self::escape_for_serializing(Self::redact_exec_root(env, exec_root)))
                 } else {
                     None
                 }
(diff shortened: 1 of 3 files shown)


hlopko (approved):
LGTM, thank you! (I'm back! :)

dae:
Thanks Daniel, appreciate it.

PR #715, opened 2021-04-29
wasm_bindgen: Add support for non-bundler target configs

### LINKED ISSUE #627: Can't build a rust_library as wasm -- output path incorrect with crate_type cdylib
Hello,

I wanted to have a single library file used by both Javascript and rust code so I tried using a rust_library rule wrapped by a rust_wasm_bindgen rule. However, the rust_library with crate_type cdylib doesn't generate the expected output file. It seems this is because rustc does not prefix the file with 'lib' in this case. Altering the code with a hack to set prefix to '' when crate type is cdylib and target arch is wasm32 resolves this issue. E.g., in private/rust.bzl _determine_lib_name():
    if toolchain.target_arch == "wasm32" and crate_type in ("cdylib"):
        prefix = ""

For the rule to succeed it is also necessary to skip the output of the CcInfo, since it won't be generated. This issue also shows up and is resolved when trying to build a rust_library with crate_type=cdylib and --platforms=@rules_rust//rust/platform:wasm.

With these changes you can have a rust_library wrapped by a rust_wasm_bindgen rule and it works as expected. Is this something that could be resolved (perhaps in a different manner) for all? The alternative would seem to be using a rust_binary rule, but for that I believe my wasm code would need to be in the main.rs file to be built, which doesn't work well when using it as a rust library from rust code.

Am I missing an existing way of accomplishing this neatly?

titanous:
This allows specifying a `target` like `web` or `nodejs` in `wasm_bindgen` and ensures that you can build a wasm `rust_shared_library`.

Closes #627

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -90,6 +90,8 @@ def _determine_lib_name(name, crate_type, toolchain, lib_hash = ""):
     prefix = "lib"
     if (toolchain.target_triple.find("windows") != -1) and crate_type not in ("lib", "rlib"):
         prefix = ""
+    if toolchain.target_arch == "wasm32" and crate_type == "cdylib":
+        prefix = ""
 
     return "{prefix}{name}-{lib_hash}{extension}".format(
         prefix = prefix,
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -580,7 +580,7 @@ def rustc_compile_action(
     if hasattr(ctx.attr, "out_binary"):
         out_binary = getattr(ctx.attr, "out_binary")
 
-    return establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration) + [
+    info = [
         crate_info,
         dep_info,
         DefaultInfo(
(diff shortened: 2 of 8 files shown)


rust/private/rustc.bzl
```diff
@@ -580,7 +580,7 @@ def rustc_compile_action(
     if hasattr(ctx.attr, "out_binary"):
         out_binary = getattr(ctx.attr, "out_binary")
 
-    return establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration) + [
+    info = [
```
hlopko, line 583: Nit: rename `info` to `providers` (I could be convinced to other plural words).

hlopko (commented):
LGTM for //rust changes, let's see what folks who own wasm think. Thank you!

rust/private/rustc.bzl
titanous: Done!

rust/private/rust.bzl
```diff
@@ -90,6 +90,8 @@ def _determine_lib_name(name, crate_type, toolchain, lib_hash = ""):
     prefix = "lib"
     if (toolchain.target_triple.find("windows") != -1) and crate_type not in ("lib", "rlib"):
         prefix = ""
+    if toolchain.target_arch == "wasm32" and crate_type == "cdylib":
+        prefix = ""
```
dfreese, line 94: Not your problem, and this doesn't make it any worse, but the logic for "prefix" needs to be cleaned up in this function.

dfreese (approved):
I can't say I "own" wasm, but this looks fine to me.

rust/private/rust.bzl
titanous: Yeah, I wanted to avoid refactoring while fixing things so did it in the existing style.

dfreese:
If you can update this to the latest main, then I'll go ahead an merge.

titanous:
Done!

PR #716, opened 2021-05-01
Removed redundant check for srcs in clippy

UebelAndre:
The `_rust_sources` check seems redundant here. [CrateInfo](https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/rust/private/providers.bzl#L17-L32) has a `srcs` field which should contain the exact same data and because the implementation filters for targets which only contain the `CrateInfo` provider, this should be limited to targets which have a `srcs` attribute that only accept `.rs` files:
https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/rust/private/rust.bzl#L660-L669

Happy to hear a good reason why we shouldn't use `CrateInfo.srcs` though.

Additionally, this adds a top level section for `rust_clippy` to the docs.

### DIFF
--- a/docs/rust_clippy.vm
+++ b/docs/rust_clippy.vm
@@ -1,0 +1,23 @@
+#[[
+## Overview
+]]#
+
+[Clippy][clippy] is a tool for catching common mistakes in [Rust][rust] code and improving it. An
+expansive list of lints and their justification can be found in their [documentation][docs].
+
+[clippy]: https://github.com/rust-lang/rust-clippy#readme
+[docs]: https://rust-lang.github.io/rust-clippy/
+[rust]: https://github.com/rust-lang/rust
+
+#[[
+### Setup
+]]#
+
+Simply add the following to the `.bazelrc` file in the root of your workspace:
+
+```text
+build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
+build --output_groups=+clippy_checks
+```
+
+This will enable clippy on all [Rust targets](./defs.md).
--- a/docs/rust_clippy.vm
+++ b/docs/rust_clippy.vm
@@ -2,12 +2,11 @@
 ## Overview
 ]]#
 
-[Clippy][clippy] is a tool for catching common mistakes in [Rust][rust] code and improving it. An
-expansive list of lints and their justification can be found in their [documentation][docs].
+[Clippy][clippy] is a tool for catching common mistakes in Rust code and improving it. An
+expansive list of lints and the justification can be found in their [documentation][docs].
 
 [clippy]: https://github.com/rust-lang/rust-clippy#readme
 [docs]: https://rust-lang.github.io/rust-clippy/
-[rust]: https://github.com/rust-lang/rust
 
 #[[
 ### Setup
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -12,12 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# buildifier: disable=module-docstring
+"""A module defining clippy rules"""
+
 load("//rust/private:common.bzl", "rust_common")
-load(
-    "//rust/private:rust.bzl",
-    "crate_root_src",
-)
+load("//rust/private:rust.bzl", "crate_root_src")
 load(
     "//rust/private:rustc.bzl",
     "collect_deps",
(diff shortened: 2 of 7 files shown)


docs/rust_clippy.vm
```diff
@@ -0,0 +1,23 @@
+#[[
+## Overview
+]]#
+
+[Clippy][clippy] is a tool for catching common mistakes in [Rust][rust] code and improving it. An
```
hlopko, line 5: Nit: I don't think we need to give a reference for Rust in this repo :)

docs/rust_clippy.vm
```diff
@@ -0,0 +1,23 @@
...
+expansive list of lints and their justification can be found in their [documentation][docs].
```
hlopko, line 6: nit: s/their/the

hlopko (commented):
@smklein @damienmg do you remember why do we look for `hdrs` rule attribute and if present we take `.rs` files from there so clippy can analyze them? AFAIK there are no Rust rules with `hdrs` attribute. Are you ok with removing that logic?

damienmg:
> @smklein @damienmg do you remember why do we look for `hdrs` rule attribute and if present we take `.rs` files from there so clippy can analyze them? AFAIK there are no Rust rules with `hdrs` attribute. Are you ok with removing that logic?

Sorry what is the context of this question? As far as I can remember, we never looked at hdrs attribute unless for c++ interactions rules (which I haven't contributed nor reviewed).

dfreese:
> @smklein @damienmg do you remember why do we look for `hdrs` rule attribute and if present we take `.rs` files from there so clippy can analyze them? AFAIK there are no Rust rules with `hdrs` attribute. Are you ok with removing that logic?

This was lifted pretty heavily from our clang-tidy rules, so I imagine that wasn't strictly intended to make it in.

rust/private/clippy.bzl
```diff
@@ -26,35 +24,20 @@ load(
...
-    rust_srcs = _rust_sources(target, ctx.rule)
 
     toolchain = find_toolchain(ctx)
     cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
     crate_info = target[rust_common.crate_info]
+    rust_srcs = crate_info.srcs.to_list()
     crate_type = crate_info.type
 
     if crate_info.is_test:
         root = crate_info.root
```
dfreese, line 38: is there a reason we wouldn't just always use the `crate_info.root` (i.e. in both the `is_test` and `not is_test` cases)?

UebelAndre: Yeah, that's a good call. In all use cases of `rust_common.crate_info` `root` appears to be set correctly which makes this work redundant too.
https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/rust/private/rust.bzl#L252-L264
https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/rust/private/rust.bzl#L287-L299
https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/rust/private/rust.bzl#L404-L416
https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/rust/private/rust.bzl#L419-L431
https://github.com/bazelbuild/rules_rust/blob/feaeb7ab712da45c1c94f7950e799d79e367ddeb/proto/proto.bzl#L225-L237

rust/private/clippy.bzl
```diff
@@ -26,38 +23,16 @@ load(
...
-
 def _clippy_aspect_impl(target, ctx):
     if rust_common.crate_info not in target:
         return []
-    rust_srcs = _rust_sources(target, ctx.rule)
 
     toolchain = find_toolchain(ctx)
     cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
     crate_info = target[rust_common.crate_info]
+    rust_srcs = crate_info.srcs.to_list()
```
dfreese, line 33: nit: remove this, because it's now unused.

rust/private/clippy.bzl
```diff
@@ -26,38 +23,16 @@ load(
...
     crate_type = crate_info.type
-
-    if crate_info.is_test:
-        root = crate_info.root
-    else:
-        if rust_srcs == []:
-            # nothing to do
-            return []
-        root = crate_root_src(ctx.rule.attr, rust_srcs, crate_info.type)
+    root = crate_info.root
```
dfreese, line 35: nit: can you remove the `root` variable, and just use `crate_info.root` in the one case it is used?

dfreese (approved):
LGTM, with some minor nits.

rust/private/clippy.bzl
UebelAndre: Odd, i thought buildifier would have caught something like this.

Thanks

UebelAndre:
Thanks David! I'll wait for @hlopko to approve too.

hlopko (approved):
LGTM, thank you!

PR #718, opened 2021-05-05
Generate valid file if no crates are found

illicitonion:
Currently we get an error about bad indentation


UebelAndre (approved):

UebelAndre:
Thanks!

UebelAndre (changes requested):
Actually, could you add a new example for this? To catch any regressions 🙏

illicitonion:
Added a unit test instead :)

UebelAndre (approved):
Nice! Thanks for adding the test!

PR #719, opened 2021-05-05
Allow tags to be applied to bindgen libraries.

UebelAndre:
I tried setting the examples to manual while testing another change and found this. `rust_bindgen_library` targets should support tags


hlopko (approved):
LGTM, thanks! :)

Issue #634, opened 2021-03-16
rust_fmt


tschuett:
Are there any plans to add a rust_fmt rule?

```python
rust_fmt(
  name = "format_library",
  deps = [
    ":bar",
    ":foo",
    ],
)
```

There is a basic example:
https://github.com/bazelbuild/rules_rust/tree/main/test/rustfmt

UebelAndre:
https://github.com/bazelbuild/rules_rust/issues/87 sounds relevant to this request

tschuett:
After reading the other issues, I like the idea of modelling it as test:
```Python
rust_test_fmt(
  name = "test_format",
  deps = [
    ":bar",
    ":foo",
   ]
)
```

hlopko:
I haven't looked into other issues (is it just https://github.com/bazelbuild/rules_rust/issues/87 or others as well?), and I only skimmed through the implementation, but my knee jerk proposal would be to implement the rust_fmt as an aspect and then it would be used by for example running `bazel build --aspects @rules_rust//rust:defs.bzl%rust_fmt_aspect` //foo/bar (there will need to be `--experimental_run_validations` also but that's only until Bazel stabilizes that feature).

If the source file is not formatted correctly, the build would fail. Advantage over `deps` approach in my perspective is that you don't have to specify which libraries to cover (which people can forget to do), you just enable the aspect on the CI for the regular test suite run (which usually has //... so it automatically covers everything).

Pardon my ignorance, but does rust-analyzer use rustfmt, or it implements formatting by itself? If it has its own implementation then there will be differences, and that will cause trouble. In that case I'd vote for using rust-analyzer for formatting (if it can be done), not rustfmt (I'm a big fan of rust-analyzer as you can tell :).

tschuett:
There was also: https://github.com/bazelbuild/rules_rust/issues/388.

I use the following to assert that nothing changed
```bash
git diff --exit-code --summary --stat
```

PR #722, opened 2021-05-05
Added `rustfmt` rules and aspects.

UebelAndre:
This pull request implements the following new rules
- `rustfmt_aspect`: An aspect for parsing information from rust targets for use in the `rustfmt` rule.
- `rustfmt_test`: A test rule for checking the formatting of a list of targets.

The intended use of these rules is to have users define a `--@rules_rust//:rustfmt.toml=` flag in their `.bazelrc` file and begin to run the `@rules_rust//tools/rustfmt` target to start formatting rust source files. I strongly feel that formatting errors should be an opt-in failure mode to keep iterating on changes fast and free from distracting (non-functional) failures. This is why there are two outuput groups for `rustfmt_aspect`, `rustfmt_manifest` (non fatal), and `rustfmt_checks` (fatal). To make sure formatting is totally up to date, users can use the `rustfmt_aspect` with the `rustfmt_checks` output group in CI to have all source targets checked. To test targets individually, the `rustfmt_test` rule can be used for more granular control.

Closes https://github.com/bazelbuild/rules_rust/issues/634 
Closes https://github.com/bazelbuild/rules_rust/issues/87


UebelAndre:
@hlopko This is still in a draft phase but would love to get your thoughts on this approach for `rustfmt` support. The thing I care most about is `rustfmt` being something we can optionally enable but always run. As a developer I don't want my build to fail because of `rustfmt` as I iterate on changes. I want to have finished writing my code and tests and have CI remind me that I need to run `rustfmt` or have a commit hook run it for me or something. This is what drew me to aspects and this design of `rustfmt_aspect` vs `rustfmt_check_aspect`. The intent is that users would be able to have a `.bazelrc` for their CI environment which uses `--aspects=@rules_rust//rust:defs.bzl%rustfmt_check_aspect`. There's some documentation I need to write for this and some polish in the runner but am looking for feedback in the overall approach. 🙏

rust/private/rust.bzl
```diff
@@ -714,8 +714,15 @@ _rust_test_attrs = {
     ),
 }
 
+_common_providers = [
```
hlopko, line 717: I believe this is not needed for this PR.

rust/private/rustfmt.bzl
```diff
@@ -0,0 +1,139 @@
...
+)
+
+def _rustfmt_check_aspect_impl(target, ctx):
+    if rust_common.crate_info not in target:
+        return []
+
+    crate_info = target[rust_common.crate_info]
+
+    # Filter out any generated files
+    srcs = [src for src in crate_info.srcs.to_list() if not src.path.startswith("bazel-out")]
```
hlopko, line 41: https://docs.bazel.build/versions/master/skylark/lib/File.html#is_source should work there?

hlopko (commented):
Very brief review only

hlopko:
I don't yet understand why  `rustfmt_aspect` exists. `rustfmt_check_aspect` deals with traversing the graph and generating validation actions, right? I imagine I'd be using it by passing `--aspects=@rules_rust//rust:defs.bzl%rustfmt_check_aspect` locally if I wanted to double check, and I imagine we'll have a CI job passing that aspect too.

When should I use `--aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect`?

I LGTMed the formatting PR, please merge and sync this one so there are only relevant files to review. Thank you!

rust/private/rust.bzl
UebelAndre: I moved it out into another PR I forgot to open yesterday.

rust/private/rustfmt.bzl
UebelAndre: Oh cool, thanks!

UebelAndre:
> When should I use `--aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect`?

This aspect is used to gather information about a target for use with a `rustfmt` target (ie `@rules_rust//tools/rustfmt:rustfmt`). The key information that's being gathered by that aspect is the edition which `cargo fmt` otherwise gets from the `Cargo.toml` file. I could remove it if I were able to get this information in a query or something but I don't think that's possible.

UebelAndre:
@hlopko Maybe https://github.com/UebelAndre/rules_rust/blob/fmt/docs/rust_fmt.md could clear some things up?

UebelAndre:
I've updated the `test/rustfmt` tests to sufficiently test the use of `--edition 2015` vs `--edition 2018`. The use of [async](https://rust-lang.github.io/async-book/) when formatting using edition 2015 produces the following error:
```
error[E0670]: `async fn` is not permitted in Rust 2015
  --> /Users/andrebrisco/Code/rules_rust/test/rustfmt/srcs/2015/formatted.rs:34:1
   |
34 | async fn edition() -> i32 {
   | ^^^^^ to use `async fn`, switch to Rust 2018 or later
   |
   = help: set `edition = "2018"` in `Cargo.toml`
   = note: for more on editions, read https://doc.rust-lang.org/edition-guide
```

rustfmt can parse the edition from a `Cargo.toml` file but in the case of Bazel, we don't use them. This is the purpose of `rustfmt_aspect`. This aspect provides the appropriate edition to use and avoids errors like the one above.

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,207 @@
...
+            }
+            target
+        }
+        None => "//...:all",
+    };
+
+    let query_args = vec![
+        "query".to_owned(),
+        format!(
+            r#"kind('{types}', {scope}) except attr(tags, 'norustfmt|manual', kind('{types}', {scope}))"#,
```
illicitonion, line 38: I'm not sure about removing `manual` targets? Wouldn't we still want to `rustfmt` e.g. a test marked manual?

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub rustfmt: PathBuf,
+
+    /// The rustfmt config file containing rustfmt settings.
+    /// https://rust-lang.github.io/rustfmt/
+    pub config: PathBuf,
+
+    /// An optional command line flag used to control what targets
+    /// to format. Users are expected to either pass a Bazel label
+    /// or a package pattern (`//my/package/...`).
+    pub package: Option<String>,
```
illicitonion, line 153: It would be nice if this were a `Vec<String>` rather then `Option<String>` so that if you have multiple things to fmt, you can do so in one invocation.

I think it would just be a matter of changing the type here, changing the `nth(1)` below to a `skip(1).collect()`, and a small tweak to `query_rustfmt_targets`?

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,207 @@
...
+        workspace: PathBuf::from(
+            env::var("BUILD_WORKSPACE_DIRECTORY")
+            .expect("The environment variable BUILD_WORKSPACE_DIRECTORY is required for finding the workspace root")
+        ),
+        bazel: PathBuf::from(
+            env::var("BAZEL_REAL")
+            .unwrap_or_else(|_| "bazel".to_owned())
+        ),
+        rustfmt: PathBuf::from(env!("RUSTFMT"))
+            .canonicalize()
```
illicitonion, line 169: Reading through symlinks can cause problems in some environments which are symlink-heavy - is there a reason to canonicalize rather than using something like:
```rust
fn absolutify_existing(path: PathBuf) -> std::io::Result<PathBuf> {
    let path = if path.is_absolute() {
        path
    } else {
        std::env::current_dir().expect("Failed to get working directory").join(path)
    };
    std::fs::metadata(&path).map(|_| path)
}
```

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,207 @@
...
+            .expect("Unable to find rustfmt binary"),
+        config: PathBuf::from(env!("RUSTFMT_CONFIG"))
+            .canonicalize()
```
illicitonion, line 172: Ditto

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,207 @@
...
+    let mut lines: VecDeque<String> = content
+        .split("\n")
+        .into_iter()
+        .map(|s| s.to_owned())
+        .collect();
+
+    RustfmtConfig {
+        edition: lines
+            .pop_front()
+            .expect("There should always be 1 line in the manifest"),
```
illicitonion, line 194:
```suggestion
            .expect("There should always be at least 1 line in the manifest"),
```

illicitonion (commented):
This generally looks solid, nice! A couple of high-level questions:
1. I'm not sure why there are two near-identical aspects - could we drive the check vs fix based on which output group you're requesting, rather than by specifying different aspects? One aspect could register both actions, but only the ones driven by the selected `OutputGroup` would actually run.
2. Is there prior art for the `_check` rule-name suffix? In my group we've generally been calling this kind of rule things like `rustfmt_test` so that querying for `_test$` shows them up, because we often want to run them at the same time as tests.

tools/rustfmt/srcs/main.rs
UebelAndre: In order for the `rustfmt_aspect` to work it would need a buildable target. If a user wants to format that target, then to respect the `manual` tag, they should explicitly format that target.

rust/private/rustfmt.bzl
```diff
@@ -0,0 +1,215 @@
...
+    rustfmt(
+        name = "rustfmt",
+    )
+    ```
+
+    This now allows users to run `bazel run //:rustfmt` to format any target which provides `CrateInfo`.
+
+    This binary also supports accepts a [label](https://docs.bazel.build/versions/master/build-ref.html#labels) or
+    pattern (`//my/package/...`) to allow for more granular control over what targets get formatted. This
+    can be useful when dealing with larger projects as `rustfmt` can only be run on a target which successfully
```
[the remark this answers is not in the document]
UebelAndre: Can aspects run on a target without also requiring that target to be built? For the `rustfmt` binaries to work, it needs a manifest of edition and sources. Can that be gathered without building the targets?

rust/private/rustfmt.bzl
```diff
@@ -0,0 +1,215 @@
...
+
+    crate_info = target[rust_common.crate_info]
+
+    # Filter out any generated files
+    srcs = [src for src in crate_info.srcs.to_list() if src.is_source]
+
+    # Only run `rustfmt` if we actually have sources to format. Some rules may produce only
+    # generated sources for `CrateInfo` and these are not necessary to check.
+    if not srcs:
+        return []
```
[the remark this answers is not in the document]
UebelAndre: Updated.

rust/private/rustfmt.bzl
```diff
@@ -0,0 +1,215 @@
...
+            crate_info.edition,
+        ] + srcs),
+    )
+
+    return [
+        DefaultInfo(
+            files = depset([manifest]),
+        ),
+        OutputGroupInfo(
+            rustfmt = depset([manifest]),
```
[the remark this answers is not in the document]
UebelAndre: Updated

tools/rustfmt/srcs/main.rs
UebelAndre: Thanks! I'll burn this to memory. I remember you giving similar feedback on an earlier PR and thought this was a good idea then 😅 

Updated

tools/rustfmt/srcs/main.rs
UebelAndre: Updated

rust/private/rustfmt.bzl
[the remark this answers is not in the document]
illicitonion: Yeah, edition and sources are all statically knowable as part of analysis, and don't require any execution. I think if you remove the `+` from your `--output_groups` flag, it should do what you want without building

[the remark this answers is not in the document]
UebelAndre: >     1. I'm not sure why there are two near-identical aspects - could we drive the check vs fix based on which output group you're requesting, rather than by specifying different aspects? One aspect could register both actions, but only the ones driven by the selected `OutputGroup` would actually run.

I didn't realize I could use the output groups to control what actions were run. I thought it was an all or nothing process. It seems that by having creating `rustfmt_manifest` and `rustfmt_checks` output groups, I'm able to run the formatter without it first failing saying my code is not formatted. This is exactly what I want.

UebelAndre:
>     1. I'm not sure why there are two near-identical aspects - could we drive the check vs fix based on which output group you're requesting, rather than by specifying different aspects? One aspect could register both actions, but only the ones driven by the selected `OutputGroup` would actually run.

I didn't realize I could use the output groups to control what actions were run. I thought it was an all or nothing process. It seems that by having creating `rustfmt_manifest` and `rustfmt_checks` output groups, I'm able to run the formatter without it first failing saying my code is not formatted. This is exactly what I want.

UebelAndre:
>     2\. Is there prior art for the `_check` rule-name suffix? In my group we've generally been calling this kind of rule things like `rustfmt_test` so that querying for `_test$` shows them up, because we often want to run them at the same time as tests.

I dislike the use of `_test` for targets that are not actually Bazel test targets. I've used `check` to signify that this is not actually a test target (that would be picked up by `bazel test //...`). If there's an easy way to turn this rule into a test target, then I'd love to do that.

illicitonion:
> > ```
> > 1. I'm not sure why there are two near-identical aspects - could we drive the check vs fix based on which output group you're requesting, rather than by specifying different aspects? One aspect could register both actions, but only the ones driven by the selected `OutputGroup` would actually run.
> > ```
> 
> I didn't realize I could use the output groups to control what actions were run. I thought it was an all or nothing process. It seems that by having creating `rustfmt_manifest` and `rustfmt_checks` output groups, I'm able to run the formatter without it first failing saying my code is not formatted. This is exactly what I want.

Hurrah! Yeah, the general idea is the aspect registers more optional actions that could be run, and the `OutputGroup`s drive which of those actions actually get run, based on what's needed to produce those outputs :)

illicitonion:
> > ```
> > 2\. Is there prior art for the `_check` rule-name suffix? In my group we've generally been calling this kind of rule things like `rustfmt_test` so that querying for `_test$` shows them up, because we often want to run them at the same time as tests.
> > ```
> 
> I dislike the use of `_test` for targets that are not actually Bazel test targets. I've used `check` to signify that this is not actually a test target (that would be picked up by `bazel test //...`). If there's an easy way to turn this rule into a test target, then I'd love to do that.

If we're having to enumerate the lintable targets in the `rustfmt_check` target anyway, it feels like that could be a test rule, and the test rule implementation could look at the `CrateInfo` provider of each target to get the information it needs... The down-side is that the unit of caching would be the entire `rustfmt_check` target, rather than per-`rust_*` target...

Then again, at least the way my org would be likely to use this is we'd replace `rust_library` and friends with our own macros which would generate a `rust_library` target and a `rustfmt_test` target each time someone makes a `rust_library` (that's what we've done for other languages' linters), so that would work fine for us!

tools/rustfmt/srcs/main.rs
UebelAndre: Ok, updated, you can now pass a list of targets or query scopes.

rust/private/rustfmt.bzl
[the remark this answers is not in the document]
UebelAndre: I've updated this to remove `+` so you can format files without requiring that they get built and as a result removed `manual`.

tools/rustfmt/srcs/main.rs
UebelAndre: the `manual` tag has been removed.

UebelAndre:
> If we're having to enumerate the lintable targets in the `rustfmt_check` target anyway, it feels like that could be a test rule, and the test rule implementation could look at the `CrateInfo` provider of each target to get the information it needs...

I wasn't even sure adding a rule like this was a good idea. How do you pass test targets into it? I'd definitely want to make sure my `rust_test` targets are also formatted.

tools/rustfmt/srcs/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        .split("\n")
+        .into_iter()
+        .map(|s| s.to_owned())
+        .collect();
+
+    RustfmtManifest {
+        edition: lines
+            .pop_front()
+            .expect("There should always be at least 1 line in the manifest"),
+        sources: lines.into(),
```
illicitonion, line 66: This doesn't really matter, and maybe feels super micro-optimisationy, but... If the format of the manifest file were to have the edition as the last line, rather than the first, you could use a `Vec` above, `pop().expect()`-off the edition, and avoid this `into()` :) But I definitely do not care if you ignore this comment.

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,182 @@
...
+    if !output.status.success() {
+        std::process::exit(output.status.code().unwrap_or(1));
+    }
+}
+
+/// Run rustfmt on a set of Bazel targets
+fn apply_rustfmt(options: &Config, targets: &Vec<String>) {
+    // Ensure the targets are first built and a manifest containing `rustfmt`
+    // arguments are generated before formatting source files.
+    build_rustfmt_targets(&options, &targets);
```
illicitonion, line 112:
```suggestion
    build_rustfmt_target_manifests(&options, &targets);
```
Or `generate_` instead of `build_`, or something - otherwise reading this it looks like we're actually building the targets

tools/rustfmt/srcs/main.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// The Bazel executable to use for builds and queries.
+    pub bazel: PathBuf,
+
+    /// Information about the current rustfmt binary to run.
+    pub rustfmt_config: rustfmt_lib::RustfmtConfig,
+
+    /// Optionally, users can pass a list of targets/packages/scopes
+    /// (eg `//my:target` or `//my/pkg/...`) to control the targets
+    /// to be formatted.
```
illicitonion, line 163:
```suggestion
    /// to be formatted. If empty, all targets in the workspace will be formatted.
```

illicitonion (commented):
This looks great!

Am I understanding right that there are now two independent ways of running rustfmt? One is to use an aspect + output_group, and the other is to make a `rustfmt_test` target which depends on all of the targets you want to rustfmt, which will allow `bazel test target` to do the checking with no aspect flag needed?

UebelAndre:
> Am I understanding right that there are now two independent ways of running rustfmt? One is to use an aspect + output_group, and the other is to make a `rustfmt_test` target which depends on all of the targets you want to rustfmt, which will allow `bazel test target` to do the checking with no aspect flag needed?

Correct. I just updated the PR description. You can run `rustfmt` with no setup required in your repo, and you can either use `rustfmt_test` or setup the aspect and `rustfmt_checks` output group to test your sources.

illicitonion (approved):
This looks great to me!

PR #727, opened 2021-05-06
Clean up crate_universe output printing

illicitonion:
This avoids a weird double-printing of errors we used to get.


crate_universe/defs.bzl
```diff
@@ -139,11 +139,10 @@ def _crate_universe_resolve_impl(repository_ctx):
             "RUSTC": str(rustc_path),
             "RUST_LOG": "info",
         },
+        quiet = False,
```
UebelAndre, line 142: Doesn't this make more info print? What does the output look like now?

UebelAndre (changes requested):

UebelAndre (approved):

PR #731, opened 2021-05-10
Use un-released fix to semver-parser

illicitonion:
semver 0.11.0 changed the behaviour of parsing `x.y` requirements - in
semver 0.10 it treated them as `>=x.y.0, <(x+1).0.0` and in semver 0.11
it treated them as `>=x.y.0, <x.(y+1).0`.

`cargo` itself is still using semver 0.10.0, but `cargo_metadata` and
`cargo-lock` have updated to semver 0.11.0, which means that `cargo` and
`crate_universe` currently treat these version ranges differently.

This patch restores the `semver 0.10.0` behaviour, which means that
`cargo` and `crate_universe` will resolve things the same.


UebelAndre (approved):
Good catch and thanks for the link!

PR #733, opened 2021-05-11
Ask rust_toolchain for rustc_src in rust_analyzer.bzl

hlopko:
In this PR we remove hardcoded assumption about where rust-analyzer support code looks for rustc sources, and instead read that information from the `rust_toolchain.rustc_src`. We also handle the case when rustc sources are actually in the main repository, so there is no need to use the `__EXEC_ROOT__/` prefix for its path.

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -2,7 +2,7 @@ workspace(name = "rules_rust")
 
 load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 
-rust_repositories()
+rust_repositories(include_rustc_src = True)
 
 load("@rules_rust//proto:repositories.bzl", "rust_proto_repositories")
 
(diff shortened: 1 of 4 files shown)


hlopko:
@illicitonion only until you'll see I just added rustc srcs back after your PR :))

Now more seriously we'll need some way of including rustc sources for some targets on the CI. I can see how an env var could be used for that, what do you think?

UebelAndre:
> Personally I'm fine with `rules_rust` always fetching the rustc srcs, as long as it doesn't force consumers of `rules_rust` to also always fetch them...

I would love to only fetch Rust srcs when needed (which is locally). I don't think there's a reason to fetch them in CI (of projects using `rules_rust`) so would like to have a mechanism for controlling this. I think there's a lot of value in `rust_analyzer` so I think it's fine to default to downloading the sources and allow for an 'opt-out' flag.

UebelAndre:
> What this PR does is only downloading rustc sources for local development of `rules_rust`, and also for CI of `rules_rust` (because we need test coverage on the CI for the rust analyzer support).

What about local development of another project which uses `rules_rust` but is not actually the rules? What does that workflow look like?

hlopko:
Ah I understand now. Yeah, having an environment variable to include sources that you set locally, or a variable to not include sources, that you set on the CI, sound like a standard solution for this problem. It's an issue a bit orthogonal to this PR, let's address it in a separate PR (do you want to give it a try? :).

PR #736, opened 2021-05-12
Rename rustc_src to rustc_srcs for bazel consistency

hlopko:
This PR renames the existing, but quite recently added attribute of `rust_toolchain` from `rustc_src` to `rustc_srcs`. The latter is more common in the Bazel ecosystem.

I mistyped the attribute name a few times myself, and also got feedback from colleagues that they did so as well :)


hlopko:
CC @djmarcin

UebelAndre (approved):

illicitonion (approved):

PR #737, opened 2021-05-12
Added environment variable to control downloading rustc sources

UebelAndre:
This is a followup to a conversation from https://github.com/bazelbuild/rules_rust/pull/733

rustc sources used in rust analyzer can now be conditionally included in toolchains using the `RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS` enviornment variable. Setting it to `true` or `1` will enable it and any other value will disable it. I feel this gives users the most flexibility in how the decide to enable/disable it locally or in CI.


hlopko (approved):
LGTM thank you! I'd give Daniel some time to take a look too.

illicitonion (approved):
Thanks!

PR #738, opened 2021-05-14
Use allowlist with bindgen

hlopko:


hlopko:
Hi @UebelAndre, I'd like to submit this PR, but I need to update the version of bindgen we use. Do you know if there is a right way to do it?

UebelAndre:
> Hi @UebelAndre, I'd like to submit this PR, but I need to update the version of bindgen we use. Do you know if there is a right way to do it?

You'll need to update the version here:
https://github.com/bazelbuild/rules_rust/blob/76cfae481b49451bc9a9ba24e02041a4a33d0432/bindgen/raze/Cargo.toml#L9

Then run `cargo update && cargo raze` from that directory and commit the results. I believe that's all that's required.

hlopko:
Writing down progress notes so I don't forget:

* `cargo update && cargo raze`
* re-added `cargo_bin_bindgen` target
* `buildifier -lint=fix -warnings=all **/BUILD*bazel`

Goal for tomorrow is to make bindgen build :)

UebelAndre:
I think you just need to update the versions specified in the metadata here:
https://github.com/bazelbuild/rules_rust/blob/main/bindgen/raze/Cargo.toml#L20-L30

Then regenerating the `cargo-raze` outputs should have the appropriate targets.

hlopko:
^ that was indeed true, I missed the second place in that file when updating the version. Thank you!

hlopko:
Wow it worked! Thanks again! :)

hlopko:
@UebelAndre could you review at your convenience? Just making sure you saw the change from a draft to an actual PR.

hlopko:
(and only after sending that comment I realized this PR was still a draft 🤦 )

UebelAndre (approved):
Looks good to me!

PR #739, opened 2021-05-17
Use bazelrc over hard coded env vars for crate_universe examples

UebelAndre:
I find this much easier to work with since I don't have to constantly set this environment variable. Thought others might find this beneficial too.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -112,8 +112,6 @@ tasks:
   crate_universe_examples_ubuntu2004:
     name: Crate Universe Examples
     platform: ubuntu2004
-    environment:
-      RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP: true
     working_directory: examples/crate_universe
     build_targets:
       - "//..."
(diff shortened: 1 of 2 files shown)


illicitonion (commented):
I'm not sure on this one... As far as I can tell, `RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP` doesn't rebuild on source changes to `crate_universe/...`, right?

Right now I don't use `RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP` for local development because I use a `cargo build` and `RULES_RUST_CRATE_UNIVERSE_RESOLVER_URL_OVERRIDE` workflow, which both allows me to re-use the target-dir build cache from my IDE and `cargo test` runs... Do you have a workflow to make sure that you get incremental builds using `RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP` and which makes sure it's up to date?

UebelAndre:
> Do you have a workflow to make sure that you get incremental builds using `RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP` and which makes sure it's up to date?

No, that's a good call. In the stuff I was working on, I wasn't really iterating over the repository rule so I didn't need to build it but then ran into an issue where I expected it to be built but it wasn't. I'm fine with not merging this and allowing the URL variable to be the testing mechanism for local development.

PR #742, opened 2021-05-18
crate_universe: fix example to match path in repo

dae:


dae:
Nevermind, it looks like more than just a simple typo. Will dig around some more :-)

UebelAndre:
I think that `README` is out of date and should be updated to point to https://bazelbuild.github.io/rules_rust/crate_universe.html

PR #743, opened 2021-05-18
Fork `cargo-raze` into `crate_universe` and apply patches

UebelAndre:
There are changes that make sense for `crate_universe` which don't make sense to `cargo-raze` but the change is to `cargo-raze` code. This change forks the behavior and applies some desired patches.


hlopko:
Seems like this draft will be closed?

UebelAndre:
> Seems like this draft will be closed?

I'm still waiting on a resolution to the conversation in #415

PR #746, opened 2021-05-18
Fix clippy warnings for //tools/rust_analyzer:gen_rust_project

mypmc:
The command `bazel run @rules_rust//tools/rust_analyzer:gen_rust_project` fails when run configuring `.bazelrc` like bellow.

```
build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
build --output_groups=+clippy_checks
```


UebelAndre (approved):
Thanks!

Issue #704, opened 2021-04-22
Path to "non-experimental" `rust_analyzer` support


UebelAndre:
This issue tracks all the things required for promoting the `rust_analzyer` rule from "experimental".

## Tasks
- [ ] Remove the requirement for an explicit list of targets
- [ ] Setup a separate toolchain for gathering rustc sources

UebelAndre:
The initial tasks are not set in stone, they're just a starting point. I'll update these with a complete list based on everyone's feedback.

cc @dfreese @illicitonion @djmarcin

UebelAndre:
A lot of good insight can be found on https://github.com/bazelbuild/rules_rust/pull/505#issuecomment-780066923

UebelAndre:
There's now docs for `rust-analyzer`: https://bazelbuild.github.io/rules_rust/rust_analyzer.html

UebelAndre:
>     * [ ]  Setup a separate toolchain for gathering rustc sources

Maybe this is solved by https://github.com/bazelbuild/rules_rust/pull/737 ? It doesn't add a separate toolchain but allows for users to optionally download the the big bundle of sources. Which I think addresses the core issue here. It'd still probably be a usability improvement to have a separate toolchain so the flag would no longer e needed but I think this is more than sufficient for now.

hlopko:
>  Remove the requirement for an explicit list of targets

IMHO that implies not having a rust analyzer target at all. I think this would be a valuable change. I imagine the workflow would then be:

1) `bazel run @rules_rust//tools:gen_rust_project -- //...`. This would invoke bazel behind the scenes with arguments after `--`. Great thing is that we can use `//...` or similar instead of explicitly listing all the top level targets.
2) the inner bazel would attach an aspect that will generate an output per rust library and build all those outputs
3) `gen_rust_project` would collect all those outputs and merge them together into `rust-project.json`

I admit I haven't yet fully read up on #505, if this was already proposed and rejected, just send me away.

UebelAndre:
> I admit I haven't yet fully read up on #505, if this was already proposed and rejected, just send me away.

No, I think this is a great idea and wasn't something rejected (if I recall correctly). If you had the time to write that up I'd be happy to review 😄

djmarcin:
[...]

PR #747, opened 2021-05-19
Added rustfmt config for running rustfmt in CI

UebelAndre:
I only added it to some key packages. I don't think it's necessary to add a `rustfmt_test` to every test target. If that's something we want enforced then we should use the aspect but I think these tests are a more focused change and allow contributors to experience this workflow.


hlopko:
What do you think about only having a single top level target (for example `//:rules_rust_fmt_check` or similar) that contains everything relevant in its `targets` attribute? Or at least one target per top level package (one for //tools, one for //util, one for //cargo, and so on)?

rustfmt is not the only thing that we want to use there. We also have clippy and rust analyzer that use the same approach. Having to add 3 targets to every new added `rust_library` in the repo is a lot of boilerplate. I prefer the alternative where we only have to maintain a list of top level targets that a rustfmt, clippy, and rust analyzer all use.

hlopko:
If we go that way we should also clean up exiting targets, like the one introduced in https://github.com/bazelbuild/rules_rust/pull/746.

illicitonion:
> What do you think about only having a single top level target (for example `//:rules_rust_fmt_check` or similar) that contains everything relevant in its `targets` attribute? Or at least one target per top level package (one for //tools, one for //util, one for //cargo, and so on)?
> 
> rustfmt is not the only thing that we want to use there. We also have clippy and rust analyzer that use the same approach. Having to add 3 targets to every new added `rust_library` in the repo is a lot of boilerplate. I prefer the alternative where we only have to maintain a list of top level targets that a rustfmt, clippy, and rust analyzer all use.

An alternative could be to use a macro so that `rust_library` (or `tidied_rust_library` or something) expands to three targets - this is a common approach to reduce boilerplate while enabling all of the desired checks.

hlopko:
Good point. We'd need something for `rust_binary` too, and I'm not sure it will work that well for `rust_analyzer` rule, which assumes there's only one central target for a workspace.

illicitonion:
This also feels like slightly open-ended Bazel question for how this kind of thing _should_ be done. This comment isn't really scoped to this PR at all, but I figured it would be worth writing down some of the trade-offs in case anyone has any bright ideas... :)

It feels like the Bazel team would like aspects to be The Way to do things like linting, but in practice there are some usability issues with aspects - specifically:
* I think ideally you'd want them to be enabled by default as part of the `test` configuration, but easy to disable for specific runs (a `--nolint` flag or similar), whereas the configuration system makes it much easier to add things for specific invocations than to remove things. Macros make this easy by just giving you targets you can filter out from your bazel invocation.
* If they need per-target or per-package configuration (e.g. to allow exceptions for specific lints for specific targets) this is really easy with a macro (you can just add an attribute and pipe it through to the underlying lint target) and quite hard with an aspect (you probably need to end up encoding anything as an extra `tag` or similar, which e.g. means you couldn't add a replacement config file, as you wouldn't be able to add a dependency on that file based on a string).
* Sharing macros across repos is easy (we could easily make "generate all three targets" the default for all users of `rules_rust` if we wanted), whereas sharing `.bazelrc` config is hard.

UebelAndre:
> What do you think about only having a single top level target (for example `//:rules_rust_fmt_check` or similar) that contains everything relevant in its `targets` attribute? Or at least one target per top level package (one for //tools, one for //util, one for //cargo, and so on)?

I'm not a fan of a top level target. I think these tests should either be individually paried with the things they test or on a per package basis.

> It feels like the Bazel team would like aspects to be The Way to do things like linting, but in practice there are some usability issues with aspects

I agree with these points. I don't want to be bothered with limiting when iterating on changes. That should be a a pre-merge kinda thing. I would love an aspect that appends test actions so I can fully build and test my changes separately of addressing linter issues and not need to hand author a bunch of targets to get to that case.

> An alternative could be to use a macro so that `rust_library` (or `tidied_rust_library` or something) expands to three targets - this is a common approach to reduce boilerplate while enabling all of the desired checks.

I like this idea because it gets me the results I want but dislike the design. I really wish aspects could run as tests with their own results and not impact user's ability to iterate on a change.

UebelAndre:
> Good point. We'd need something for `rust_binary` too, and I'm not sure it will work that well for `rust_analyzer` rule, which assumes there's only one central target for a workspace.

I thought the top level [rust_analyzer](https://bazelbuild.github.io/rules_rust/rust_analyzer.html#rust_analyzer) rule was something we were looking to get rid of? https://github.com/bazelbuild/rules_rust/issues/704 I feel like having to manage a bucket of top level targets is a clunky and error prone practice.

hlopko:
> I thought the top level rust_analyzer rule was something we were looking to get rid of? #704 I feel like having to manage a bucket of top level targets is a clunky and error prone practice.

1) I'm not a fan either. But that's what we have today and I'm happy for it. The ideal (for us at least) rust-analyzer solution is to teach rust analyzer to run Bazel automatically behind the scenes and setup itself based on the output. Not sure anybody tried to pursue that avenue. In the meantime, I'm ok with a top level target.

> I'm not a fan of a top level target. I think these tests should either be individually paried with the things they test or on a per package basis.

2) Again, I agree that the top level target is not great. But I'm also not a fan of separate targets. For my and our internal workflows just having the validation aspect that is configured in `.bazelrc` is good enough ™️. `.bazelci/presubmit.yml` or equivalent is then running bazel with the aspect on specified targets (or `//...` which doesn't work in BUILD files, but it does on Bazel's command line). Internally running rustfmt/clippy is not driven by Bazel, but by the SCM tooling (imagine `git fix-everything` command that usually only fixes lines that are modified).

> > It feels like the Bazel team would like aspects to be The Way to do things like linting, but in practice there are some usability issues with aspects

> I agree with these points. I don't want to be bothered with limiting when iterating on changes. That should be a a pre-merge kinda thing. I would love an aspect that appends test actions so I can fully build and test my changes separately of addressing linter issues and not need to hand author a bunch of targets to get to that case.

In general for validation actions we have 2 options. Either register the action in the library/binary rule (controlled by a build setting), or use an aspect. Bazel team uses both depending on memory usage restrictions (typically C++ rules will use a direct action because we have a lot of C++ and action in the rules usually consumes less memory than an aspect. The price to pay is tight coupling). What I read @UebelAndre is that you would be well served by having a .bazelrc config that enables validations right? On normal development you'd just `bazel test //...`, and before upload (and the CI would run this too) you'd `bazel test --config=validate //...`. FTR there is a way to have .bazelrc checked in the repo and have it load a local .bazelrc file when it exists. 

> > An alternative could be to use a macro so that rust_library (or tidied_rust_library or something) expands to three targets - this is a common approach to reduce boilerplate while enabling all of the desired checks.

> I like this idea because it gets me the results I want but dislike the design. I really wish aspects could run as tests with their own results and not impact user's ability to iterate on a change.

3) I really really dislike macros. The main issue is that not everybody will be using the macro. There will be `rust_library` targets and there will be `tidied_rust_library` targets. Buildozer does not understand macros, and if you're working on a migration tool, you obviously need to deal with `rust_library`, but you will often end up having to have custom logic for individual macros. Second problem is that macros are not visible in `bazel query`, that confuses users and again makes some tooling more complicated. Third problem is that they don't compose very well, you end up with `formatted_rust_library`, `clippied_rust_libray`, `formatted_clippied_rust_library`, `miried_rust_library` and you can see how that goes. I'll stop ranting about macros now :)


Now to your points @illicitonion:

> I think ideally you'd want them to be enabled by default as part of the test configuration, but easy to disable for specific runs (a --nolint flag or similar), whereas the configuration system makes it much easier to add things for specific invocations than to remove things. Macros make this easy by just giving you targets you can filter out from your bazel invocation.

4) I agree with the point, I also think we will have to wait a long time for this to be fixed by the Bazel team. The answer will be use `--config` and .bazelrc. Would that be an acceptable solution to you?

> If they need per-target or per-package configuration (e.g. to allow exceptions for specific lints for specific targets) this is really easy with a macro (you can just add an attribute and pipe it through to the underlying lint target) and quite hard with an aspect (you probably need to end up encoding anything as an extra tag or similar, which e.g. means you couldn't add a replacement config file, as you wouldn't be able to add a dependency on that file based on a string).

5) Yup, typically you'd do this using `tags` or `features` on the package or on the original target. Yup, adding more dependencies is not possible this way.

> Sharing macros across repos is easy (we could easily make "generate all three targets" the default for all users of rules_rust if we wanted), whereas sharing .bazelrc config is hard.

6) Yes I agree sharing .bazelrc configs is hard. External dependencies in general are hard and you end up copy pasting non trivial amount of code into your WORKSPACE file whenever you add a dependency on libraries or rules. Giving users a well commented .bazelrc snippet is IMHO not going to make anybody's life significantly harder.

I feel like I missed some points, did I?

illicitonion:
> > I think ideally you'd want them to be enabled by default as part of the test configuration, but easy to disable for specific runs (a --nolint flag or similar), whereas the configuration system makes it much easier to add things for specific invocations than to remove things. Macros make this easy by just giving you targets you can filter out from your bazel invocation.
> 
> 1. I agree with the point, I also think we will have to wait a long time for this to be fixed by the Bazel team. The answer will be use `--config` and .bazelrc. Would that be an acceptable solution to you?

Yeah, I think in our repo's case people first discovering problems when they make a PR and it's failing because they didn't invoke bazel themselves with the right flags is adequate. But it would be great for the wider Bazel community to not have to make this trade-off in the future :)

UebelAndre:
>     1. I'm not a fan either. But that's what we have today and I'm happy for it. The ideal (for us at least) rust-analyzer solution is to teach rust analyzer to run Bazel automatically behind the scenes and setup itself based on the output. Not sure anybody tried to pursue that avenue. In the meantime, I'm ok with a top level target.

I'd love your feedback on https://github.com/bazelbuild/rules_rust/issues/704 so we can change the docs to reflect it's no longer experimental at some point 😄 

>     1. Again, I agree that the top level target is not great. But I'm also not a fan of separate targets. For my and our internal workflows just having the validation aspect that is configured in `.bazelrc` is good enough ™️. `.bazelci/presubmit.yml` or equivalent is then running bazel with the aspect on specified targets (or `//...` which doesn't work in BUILD files, but it does on Bazel's command line). Internally running rustfmt/clippy is not driven by Bazel, but by the SCM tooling (imagine `git fix-everything` command that usually only fixes lines that are modified).

I do not think it's appropriate to have one single top level target for all of formatting. Or at least have that be the only formatting target. Each package should be responsible for providing the tests it needs to make sure the code there is sufficiently tested and maintained. The only thing I'd expect to find top level would be a `test_suite`. But updating one file shouldn't then require that I go attempt to format the entire world. I'd be happy to apply this change to a `.bazelrc` file instead. Does that sound reasonable?

>     1. I really really dislike macros. The main issue is that not everybody will be using the macro. There will be `rust_library` targets and there will be `tidied_rust_library` targets. Buildozer does not understand macros, and if you're working on a migration tool, you obviously need to deal with `rust_library`, but you will often end up having to have custom logic for individual macros. Second problem is that macros are not visible in `bazel query`, that confuses users and again makes some tooling more complicated. Third problem is that they don't compose very well, you end up with `formatted_rust_library`, `clippied_rust_libray`, `formatted_clippied_rust_library`, `miried_rust_library` and you can see how that goes. I'll stop ranting about macros now :)

I wouldn't have separate macros, instead, just one with some flags on it that enable/disable certain targets. But that seems like something consumers of `rules_rust` will have to do themselves 😅

UebelAndre:
Ooops, didn't mean to delete this

hlopko:
> Yeah, I think in our repo's case people first discovering problems when they make a PR and it's failing because they didn't invoke bazel themselves with the right flags is adequate. But it would be great for the wider Bazel community to not have to make this trade-off in the future :)

@illicitonion If you care about this problem enough (and it's fine if not :) could you create a bazel issue with the problem description and desired end state? I can't do much, but at least I can nudge relevant people into reading the issue.

> I do not think it's appropriate to have one single top level target for all of formatting. Or at least have that be the only formatting target. Each package should be responsible for providing the tests it needs to make sure the code there is sufficiently tested and maintained. The only thing I'd expect to find top level would be a test_suite. But updating one file shouldn't then require that I go attempt to format the entire world. 

@UebelAndre I wasn't clear in my previous reply. What I think is the path of least resistance today is (pardon made up labels):

* don't use Bazel to execute formatters, use a script/tool that uses bazel to build the formatter and invoke the formatter either on the whole repository or on changed lines. In concrete example, don't use `bazel run @rules_rust//:format_the_code`, use `bazel run @rust_toolchain//tools:rustfmt -- **/*.rs`
* don't use any target to validate formatting / ensure no lint errors are present. Use an aspect and invoke bazel with `//...`, for example `bazel build //.. --aspects @rules_rust//:rustfmt.bzl%rustfmt_aspect`. Use the same on the CI.

I don't like having multiple supporting targets along with each `rust_library`. I disagree it's the package's job to ensure it follows the formatting and style policy of the workspace. I think it's the opposite, the workspace dictates those, and packages must comply. I don't see unit (or other kinds of) tests that are an important part of libraries as the same thing as  formatting and linter checks.

> I'd be happy to apply this change to a .bazelrc file instead. Does that sound reasonable?

I'm not sure I understand (it's quite late in here and I'm more of a morning person :) what do you mean by "this change", could you elaborate?

UebelAndre:
> I'm not sure I understand (it's quite late in here and I'm more of a morning person :) what do you mean by "this change", could you elaborate?

Delete the dedicated targets and add a `.bazelrc` with
```
build --aspects=//rust/private:rustfmt.bzl%rustfmt_aspect
build --output_groups=+rustfmt_checks
```

> I don't like having multiple supporting targets along with each `rust_library`. I disagree it's the package's job to ensure it follows the formatting and style policy of the workspace. I think it's the opposite, the workspace dictates those, and packages must comply. I don't see unit (or other kinds of) tests that are an important part of libraries as the same thing as formatting and linter checks.

The thing I don't want to do is have to register my target in multiple locations. If there are formatting rules enforced at the workspace level, then those should be done via aspects, which works quite nicely for that. But the thing I want is to be able to build and separately check that my target meets the formatting/linting standards. My ideal solution would be to have a separate test target so my command line doesn't change as I build and test. But I understand that the thing I want doesn't really exist and requires some macro magic. So as far as solutions go, I think applying formatting in bazel should be done in aspects or handled by a target that will go apply formatting changes. I don't want to have to go setup another tool separately to bazel. But anyway, I can understand your perspective too. This isn't something I'd like to get hung up on and am happy to update the PR to add a `.bazelrc` file but don't want to promote a pattern that forces you to double register all your targets without sufficient buy in that this is the intended way to do things.

hlopko:
> Delete the dedicated targets and add a .bazelrc with
> build --aspects=//rust/private:rustfmt.bzl%rustfmt_aspect
> build --output_groups=+rustfmt_checks

SGTM! I think that makes the aspect be part of the public API (and I'd create a //rust:rustfmt.bzl file for that so people don't get used to seeing `private`). And I'd also create a config for formatting, for example

```
build:rustfmt --aspects=//rust:rustfmt.bzl%rustfmt_aspect
build:rustfmt --output_groups=+rustfmt_checks
```

And I'd use that config in the `.bazelci/presubmit.yml`

hlopko:
> The thing I don't want to do is have to register my target in multiple locations. If there are formatting rules enforced at the workspace level, then those should be done via aspects, which works quite nicely for that. But the thing I want is to be able to build and separately check that my target meets the formatting/linting standards. My ideal solution would be to have a separate test target so my command line doesn't change as I build and test. But I understand that the thing I want doesn't really exist and requires some macro magic. So as far as solutions go, I think applying formatting in bazel should be done in aspects or handled by a target that will go apply formatting changes. I don't want to have to go setup another tool separately to bazel. But anyway, I can understand your perspective too. This isn't something I'd like to get hung up on and am happy to update the PR to add a .bazelrc file but don't want to promote a pattern that forces you to double register all your targets without sufficient buy in that this is the intended way to do things.

I think I understand now, thank you.

> The thing I don't want to do is have to register my target in multiple locations.

+100, fully on board. It wasn't clear to me before that this is your motivation, thank you for bearing with me.

.bazelrc
```diff
@@ -0,0 +1,6 @@
+# `.bazelrc` is a Bazel configuration file.
+# https://docs.bazel.build/versions/master/best-practices.html#using-the-bazelrc-file
+
+# Enable rustfmt for all targets in the workspace
+build --aspects=//rust:defs.bzl%rustfmt_aspect
```
hlopko, line 5: This now applies to all bazel builds, I'd prefer to have it applied only when `--config=rustfmt` is passed by writing:

```
build:rustfmt --aspects=//rust:defs.bzl%rustfmt_aspect
build:rustfmt --output_groups=+rustfmt_checks
```

Also below. I thought that's what we agreed on in the discussion before?

UebelAndre: It is, I'd forgotten about it... Had a busy, distracting week 😅 

Added the config

test/rustfmt/rustfmt_failure_test.sh
```diff
@@ -62,8 +52,20 @@ load("@rules_rust//rust:repositories.bzl", "rust_repositories")
 rust_repositories()
 EOF
 
+  # Drop the 'norustfmt' tags
+  if [ "$(uname)" == "Darwin" ]; then
+    sed -i '' -e 's/"norustfmt"//' "${new_workspace}/test/rustfmt/BUILD.bazel"
```
hlopko, line 57: Nit: use a variable for sed flags (SEDOPTS?) and `if` to initialize the variable. The actual sed invocation can then be present just once.

tools/rustfmt/BUILD.bazel
```diff
@@ -60,7 +60,7 @@ rust_binary(
 )
 
 rust_clippy(
-    name = "rustfmt_clippy",
+    name = "clippy",
```
hlopko, line 63: Why this change?

util/dir_zipper/BUILD.bazel
```diff
@@ -1,4 +1,4 @@
-load("//rust:rust.bzl", "rust_binary")
+load("//rust:defs.bzl", "rust_binary")
```
hlopko, line 1: Why these changes?

BUILD.bazel
```diff
@@ -24,3 +24,9 @@ label_flag(
     build_setting_default = "//tools/rustfmt:rustfmt.toml",
     visibility = ["//visibility:public"],
 )
+
+alias(
```
hlopko, line 28: Why this change?

hlopko (commented):
There are some changes I don't understand. I'm happy to approve a separate PR with rust.bzl -> defs.bzl change. I'd like to keep this PR focused.

util/dir_zipper/BUILD.bazel
UebelAndre: Isn't `//rust:rust.bzl` deprecated/`//rust:defs.bzl` recommended?

BUILD.bazel
UebelAndre: I found it to be more consistent with `//:rustfmt.toml`. Having that setting in a different location than the actual call for running rustfmt like unnecessary burden of knowledge. But I can leave this out if you dislike it.

tools/rustfmt/BUILD.bazel
UebelAndre: I don't think there should be a need for `package:package_thing` targets. They should just be `package:thing`. But I can remove this if you'd prefer.

util/dir_zipper/BUILD.bazel
hlopko: Yup, but it seems unrelated to this PR. I'm happy to approve a separate PR with this change.

BUILD.bazel
hlopko: I'm fine with the change but let's do a separate PR for that? Then you can give motivation for the change in the PR description, whereas in this PR it's a distraction.

tools/rustfmt/BUILD.bazel
hlopko: Makes sense to me, but let's make a separate PR for that?

.bazelrc
hlopko: No problem, thanks for working on this!

util/dir_zipper/BUILD.bazel
UebelAndre: Reverted though.

tools/rustfmt/BUILD.bazel
UebelAndre: Reverted though.

test/rustfmt/rustfmt_failure_test.sh
UebelAndre: Updated.

hlopko (approved):
Please revert test/rustfmt/BUILD.bazel and tools/runfiles/BUILD.bazel, otherwise LGTM, thank you!

BUILD.bazel
UebelAndre: https://github.com/bazelbuild/rules_rust/pull/761

PR #748, opened 2021-05-19
`rustfmt_aspect` now appropriately uses a config

UebelAndre:
I forgot to add this attribute in the initial PR. Now the check should appropriately use the configured `rustfmt.toml` file.

### DIFF
--- a/rust/private/rustfmt.bzl
+++ b/rust/private/rustfmt.bzl
@@ -61,9 +61,9 @@ def _perform_check(edition, srcs, ctx):
 
     ctx.actions.run(
         executable = ctx.executable._process_wrapper,
-        inputs = srcs,
+        inputs = srcs + [config],
         outputs = [marker],
-        tools = [toolchain.rustfmt, config],
+        tools = [toolchain.rustfmt],
         arguments = [args],
         mnemonic = "Rustfmt",
     )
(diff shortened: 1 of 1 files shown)


rust/private/rustfmt.bzl
```diff
@@ -108,6 +110,11 @@ users may tag a target with `norustfmt` to have it skipped. Additionally, genera
 source files are also ignored by this aspect.
 """,
     attrs = {
+        "_config": attr.label(
```
illicitonion, line 113: I'd be inclined to make this configurable (i.e. remove the leading `_`) so that people can override it if they really want - WDYT?

rust/private/rustfmt.bzl
```diff
@@ -61,7 +63,7 @@ def _perform_check(edition, srcs, ctx):
         executable = ctx.executable._process_wrapper,
         inputs = srcs,
         outputs = [marker],
-        tools = [toolchain.rustfmt],
+        tools = [toolchain.rustfmt, config],
```
illicitonion, line 66: This feels like more of an `input` than a `tool`? But don't think it really matters which one it's in tbh

rust/private/rustfmt.bzl
UebelAndre: Is that possible in an aspect?

illicitonion: Oh, yeah, right! Indeed!

illicitonion (approved):
Thanks!

rust/private/rustfmt.bzl
UebelAndre: Updated

PR #750, opened 2021-05-19
Fix rustc-dev

zrlk:
Sometime over the past year the layout of rustc-dev's nightly release has changed.


UebelAndre:
Is there a way to check for when that change happened to make sure we're always doing the right thing based on what dev version is being requested?

zrlk:
Looks like 2020-12-14 was the first nightly that had the directory change (found via manual bisect).

hlopko (approved):
Thanks for fixing this!

PR #751, opened 2021-05-20
Make rust analyzer aspect follow rust_test.crate attribute

hlopko:
Before, generated rust-project.json did not contain dependencies from the `rust_library` when `rust_test` used `.crate` attribute and the test was the only root in the `targets` attribute of `rust_analyzer` rule.

This PR depends on https://github.com/bazelbuild/rules_rust/pull/754 and https://github.com/bazelbuild/rules_rust/pull/760.


hlopko:
CC @djmarcin

rust/private/rust.bzl
```diff
@@ -703,6 +703,13 @@ _rust_test_attrs = {
             ["Make variable"](https://docs.bazel.build/versions/master/be/make-variables.html) substitution.
         """),
     ),
+    "harness": attr.bool(
+        mandatory = False,
+        default = True,
+        doc = _tidy("""\
+            Whether to use libtest (defaults to True).
```
UebelAndre, line 710:
```suggestion
            Whether to use libtest.
```

Stardoc will render the default, we don't need to also write it into the docs.

test/rust_analyzer/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+    testonly = True,
+    targets = [":mylib_test"],
+)
+
+rust_test(
+    name = "rust_project_json_test",
+    srcs = ["rust_project_json_test.rs"],
+    args = ["$(location :rust-project.json)"],
+    data = [":rust-project.json"],
+    harness = False,
```
UebelAndre, line 48: Is there a reason why this couldn't use `harness = True`?

test/unit/rust_analyzer/rust_analyzer_test.bzl
```diff
@@ -0,0 +1,52 @@
+"""Unittests for rust rules."""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+
+# This is an unittest for rust_analyzer.bzl, violating private visibility is necessary.
+# buildifier: disable=bzl-visibility
```
UebelAndre, line 6: Does this violate visibility patterns?

UebelAndre (changes requested):
Sorta feels like there's two changes here? They seem good but curious as to what brought on the `harness` attribute.


Oops, there's another PR 😅 

But yeah, was it necessary for this or just a nicety?

hlopko:
> Sorta feels like there's two changes here? They seem good but curious as to what brought on the `harness` attribute.
> 
> Oops, there's another PR 😅
> 
> But yeah, was it necessary for this or just a nicety?

I pass a path to rust-project.json as a command line arg to the test. 

* libtest doesn't support custom arguments
* It felt weird to have `std::env::args()` in a regular `#[test]`.
* The code was a tiny little bit longer using libtest.

Now that I'm forced to think about this more, I found alternatives:

* pass the path to the `rust-project.json` through environment variable.
* access the file through runfiles.

I'm inclined to pursue the runfiles alternative (but I think the harness PR still makes sense and can go in).

test/unit/rust_analyzer/rust_analyzer_test.bzl
hlopko: Removed

test/rust_analyzer/BUILD.bazel
hlopko: Answered in a top level comment.

hlopko:
@googlebot I fixed it.

hlopko:
And finally we have a green CI. PTAL :)

test/unit/rust_analyzer/rust_analyzer_test.bzl
```diff
@@ -0,0 +1,49 @@
+"""Unittests for rust rules."""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+load("//rust:defs.bzl", "rust_analyzer", "rust_library")
+
+def _rust_analyzer_hello_world_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
```
UebelAndre, line 8: nit: can this just be called `target`?

test/unit/rust_analyzer/rust_analyzer_test.bzl
```diff
@@ -0,0 +1,49 @@
+"""Unittests for rust rules."""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+load("//rust:defs.bzl", "rust_analyzer", "rust_library")
+
+def _rust_analyzer_hello_world_test_impl(ctx):
```
UebelAndre, line 6: General question: What is the value in unit tests like these? I've always felt they were a bit tedious to maintain and didn't offer anything more than what an integration test would.

UebelAndre (approved):
This looks good to me! a small nit and one question that's more of an aside 😅

test/unit/rust_analyzer/rust_analyzer_test.bzl
hlopko: Done.

test/unit/rust_analyzer/rust_analyzer_test.bzl
hlopko: To be honest, this test is there only because I started with the unittest and then realized I need the integration test. I probably should delete the unit test for the reasons you mention, and I will, unless there's a value in having a simple example of using Starlark unit testing, and a place where we can add future rust analyzer unit tests.

UebelAndre: I've not found a place where I value the unittest (when there's an integration test). I think integration tests exercise far more functionality as well as the important parts of a rule. But maybe I should care more about how many actions are in my rules?

hlopko: In general I reach for unit tests whenever I can. I find it more convenient to assert on registered actions, command lines, inputs, that sorts of things. 

I don't see a value in asserting how many actions are registered by a rule though, that was really just a hello world test and i wanted to assert something. I deleted the test now, it really doesn't pull its weight.

UebelAndre:
This looks good to me! Feel free to merge whenever you'd like 😄 and thanks!

dfreese (approved):

scentini (approved):

PR #755, opened 2021-05-27
util/process_wrapper: system_posix.cc: use waitpid correctly

### LINKED ISSUE #708: Compilation errors hidden on Linux/MacOS

hlopko:
All the credit goes to https://github.com/bazelbuild/rules_rust/pull/709 and to @Ubehebe.

I'm barely addressing a nit.

Fixes #708.


gribozavr (approved):

util/process_wrapper/system_posix.cc
```diff
@@ -174,12 +175,28 @@ int System::Exec(const System::StrType &executable,
...
-  return exit_status;
+  if (WIFEXITED(exit_status)) {
+    return WEXITSTATUS(exit_status);
+  } else if (WIFSIGNALED(exit_status)) {
+    raise(WTERMSIG(exit_status));
+#ifdef WCOREDUMP
+  } else if (WCOREDUMP(exit_status)) {
+    std::cerr << "process wrapper error: child process produced a core dump.\n";
+    return -1;
+#endif
```
dfreese, line 191: I don't think this is quite right, given [the man page](https://man7.org/linux/man-pages/man2/wait.2.html):
>WCOREDUMP(wstatus)
              returns true if the child produced a core dump (see
              core(5)).  This macro should be employed only if
              WIFSIGNALED returned true.

dfreese: This seems like it's just additional information, but not strictly necessary.  Can you either incorporated it into the condition on line 185, or remove it?

hlopko: Good catch, removed.

dfreese (approved):

UebelAndre:
@hlopko seems like a new warning has appeared as a result of these changes?

From [builds/3211](https://buildkite.com/bazel/rules-rust-rustlang/builds/3211#f7b64823-f793-4ee8-9ff2-607a0ce65a6c)
```
(21:24:44) INFO: From Compiling util/process_wrapper/system_posix.cc:
util/process_wrapper/system_posix.cc: In static member function 'static int process_wrapper::System::Exec(const StrType&, const Arguments&, const EnvironmentBlock&, const StrType&, const StrType&)':
util/process_wrapper/system_posix.cc:195:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

PR #758, opened 2021-05-28
Fixes for //tools/rust_analyzer:gen_rust_project

csmulhern:
This commit fixes a couple of issues with the gen_rust_project rule.

1) Repository name is ignored from the parsed label, causing targets in external workspaces to not work correctly.

2) The bazel-bin path is relying on the workspace symlink instead of the actual bazel-bin directory. This causes file not found errors when --experimental_convenience_symlinks=ignore is set (see: https://docs.bazel.build/versions/master/command-line-reference.html#flag--experimental_convenience_symlinks).


hlopko (approved):
LGTM, thank you!

PR #760, opened 2021-06-01
Add support for manifest based runfiles.

hlopko:
These are typically used on Windows when using symlinks which are needed
for directory based runfiles is not possible.

Fixes https://github.com/bazelbuild/rules_rust/issues/616.

### DIFF
--- a/tools/runfiles/runfiles.rs
+++ b/tools/runfiles/runfiles.rs
@@ -31,23 +31,54 @@
 //!     // ...
 //!     ```
 
+use std::collections::HashMap;
 use std::env;
 use std::fs;
 use std::io;
 use std::path::Path;
 use std::path::PathBuf;
 
+enum Mode {
+    DirectoryBased(PathBuf),
+    ManifestBased(HashMap<PathBuf, PathBuf>),
+}
+
 pub struct Runfiles {
-    runfiles_dir: PathBuf,
+    mode: Mode,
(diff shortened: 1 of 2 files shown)


illicitonion (approved):

tools/runfiles/runfiles.rs
```diff
@@ -108,10 +148,27 @@ pub fn find_runfiles_dir() -> io::Result<PathBuf> {
...
+        Err(_) => false,
+    }
+}
+
+fn find_manifest_path() -> io::Result<PathBuf> {
+    match std::env::var_os("RUNFILES_MANIFEST_FILE") {
+        Some(path) => Ok(path.into()),
+        None => Err(
+            make_io_error(
+                "RUNFILES_MANIFEST_ONLY was set to '1', but RUNFILES_MANIFEST_FILE was not set. Did Bazel change?"))
```
UebelAndre, line 170: Nit: This function by itself doesn't guarantee that this code path is gated behind `RUNFILES_MANIFEST_ONLY`. Maybe it should check that or this message be updated?

UebelAndre (approved):
One comment but otherwise looks good to me!

tools/runfiles/runfiles.rs
hlopko: Added assertions.

Issue #589, opened 2021-02-16
Breakage with Bazel HEAD due to 00805727b867d33fd922e63ca82b0d9825ad79fe


comius:
This breakage was reported by our continuous integration.

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1915#6900da06-6f7d-4ae0-8005-ea936f851b6e

It is probably caused by bazelbuild/bazel@0080572 that calls test binaries in external repositories not using their legacy external/$REPO/$REPO_RELATIVE_PATH path but the canonical ../$REPO/$REPO_RELATIVE_PATH one.

I haven't looked into what the mechanism of the breakage is; my guess is that test binaries try to find data files relative to $0 and that logic needs to be updated.

hlopko:
Thanks for letting us know! :)

meteorcloudy:
rules_scala has been disabled in Bazel downstream, but you can still check its status with Bazel@HEAD at https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/848

hlopko:
Thanks, I looked at it briefly but couldn't reproduce. Will take a look again.

hlopko:
(misclicked)

djmarcin:
The test is attempting to load a shared library and is now failing to find it. Just a guess, but it might be because the embedded rpath in the test binary no longer points to the right place after the upstream change.
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from @examples//ffi/rust_calling_c/c:native_matrix_test
-----------------------------------------------------------------------------
/b/f/w/rules_rust/bazel-out/k8-fastbuild/bin/external/examples/ffi/rust_calling_c/c/native_matrix_test.runfiles/rules_rust/../examples/ffi/rust_calling_c/c/native_matrix_test: error while loading shared libraries: libexternal_Sexamples_Sffi_Srust_Ucalling_Uc_Sc_Slibnative_Umatrix.so: cannot open shared object file: No such file or directory
```

PR #763, opened 2021-06-02
Fix @examples//ffi/rust_calling_c/c/native_matrix_test on RBE.

hlopko:
Seems the RBE toolchain is not working well with C++ dynamic mode tests.
Rather than struggling through the complicated fixing of the toolchain
we tell Bazel to not use dynamic mode.

Fixes https://github.com/bazelbuild/rules_rust/issues/589.

### DIFF
--- a/examples/ffi/rust_calling_c/c/BUILD.bazel
+++ b/examples/ffi/rust_calling_c/c/BUILD.bazel
@@ -13,6 +13,7 @@ cc_test(
     name = "native_matrix_test",
     srcs = ["matrix_test.c"],
     copts = ["-std=c99"],
+    linkstatic = 1,
     deps = [
         ":native_matrix",
     ],


UebelAndre (approved):
Looks good to me! But can you also open an issue outlining the complications with dynamic mode and link that to this PR?

hlopko:
I'll try to work with the RBE toolchains team to fix the issue internally when I'm back from my vacation. I don't currently have access to RBE so I cannot debug and provide more details than "dynamic mode doesn't work for us". My first guess would be RBE C++ toolchain probably doesn't use EXEC_ORIGIN, but I need access to investigate further.

But frankly, I'd prefer if we didn't use RBE toolchains at all. So maybe I'll just pursue that.

UebelAndre:
> I'll try to work with the RBE toolchains team to fix the issue internally when I'm back from my vacation. I don't currently have access to RBE so I cannot debug and provide more details than "dynamic mode doesn't work for us". My first guess would be RBE C++ toolchain probably doesn't use EXEC_ORIGIN, but I need access to investigate further.
> 
> But frankly, I'd prefer if we didn't use RBE toolchains at all. So maybe I'll just pursue that.

Just out of my own curiosity, why do you prefer we don't use RBE toolchains? It seems like they let us test a unique but not uncommon environment.

hlopko:
To clarify, I'm all for testing on RBE. I believe since we have `remotable` attribute on https://docs.bazel.build/versions/master/skylark/lib/globals.html#repository_rule we don't need RBE toolchains repository. To elaborate a bit more:

* We only use RBE toolchains because we need C++ toolchain (Rust toolchain is remote execution friendly).
* C++ toolchain uses system installed compiler by default, and Bazel tries to detect and autoconfigure C++ toolchain by inspecting the system.
* Configuring the C++ toolchain on RBE docker containers used to be hard, RBE toolchains did it by running bazel inside a container and copying out the autogenerated files. That's the rbe_autoconfig rule we have in the workspace file.
* With remotable=True we can instead tell Bazel to not execute commands locally on the host, but to send invocations to RBE workers. This is much more convenient approach IMHO.

UebelAndre:
Thanks for the extra context! 🙏 😃

Issue #741, opened 2021-05-17
Bindgen should use a downloaded or compiled `libstdc++` shared lib


UebelAndre:
Currently the `rust_bindgen` ([rust-bindgen](https://rust-lang.github.io/rust-bindgen/)) rule depends on a [local_repository](https://docs.bazel.build/versions/master/be/workspace.html#local_repository) to make `libstdc++` binaries available to the rule. 
https://github.com/bazelbuild/rules_rust/blob/eb06a7974f448b3a2199d490282906d698c2d316/bindgen/repositories.bzl#L99-L113

This is pretty non-hermetic and I believe is the reason why we can't use the bindgen rules on MacOS. Instead, the rule should [build it's own version of libstdc++](https://libcxx.llvm.org/docs/BuildingLibcxx.html) or use a pre-compiled one that exists in a shared location (not something that is installed to the host).

UebelAndre:
Maybe this could be done using [rules_foreign_cc](https://github.com/bazelbuild/rules_foreign_cc) if there is no pure Bazel implementation to build what we need.

illicitonion:
I believe it should be fetching these from the C++ toolchain - there is already code for this here:
https://github.com/bazelbuild/rules_rust/blob/d6cf2ec939c1eff9f6601588468146cf72c97e5b/rust/private/rustc.bzl#L876-L901

djmarcin:
On macOS, libstdc++.dylib has been removed in favor of libc++.dylib as far as I can tell. The bindgen error I get is that the symlink to /usr/lib/libstdc++.dylib is actually a broken link, which appears to be why it isn't working by default. If I manually correct it to point to a libc++.dylib on my system then bindgen works again.

```
$ realpath /usr/lib/libstdc++.dylib
/usr/lib/libstdc++.6.0.9.dylib
$ ls /usr/lib/libstdc++.6.0.9.dylib
ls: /usr/lib/libstdc++.6.0.9.dylib: No such file or directory
```

I'm not sure from the above discussion if there's a reason not to use the method that @illicitonion mentioned to pull the correct library from the configured cc_toolchain. If it's a reasonable option, I can try to prep a PR for it.

UebelAndre:
[...]

Issue #621, opened 2021-02-26
Enable @examples//ffi/rust_calling_c/... on MacOS CI


UebelAndre:
Building these targets generates a build failure
```
rust_calling_c % bazel build ...
```
```
INFO: Invocation ID: f981af43-943e-40bb-ba01-ed9362446315
INFO: Analyzed 15 targets (0 packages loaded, 0 targets configured).
INFO: Found 15 targets...
ERROR: /private/var/tmp/_bazel_user/1426ad6c00f8f7b996ddc75bef70871c/external/local_libstdcpp/BUILD.bazel:3:11: SolibSymlink _solib_darwin_x86_64/_U@local_Ulibstdcpp_S_S_Clibstdc++___U/libstdc++.6.dylib failed: missing input file 'external/local_libstdcpp/libstdc++.6.dylib', owner: '@local_libstdcpp//:libstdc++.6.dylib'
ERROR: /private/var/tmp/_bazel_user/1426ad6c00f8f7b996ddc75bef70871c/external/local_libstdcpp/BUILD.bazel:3:11 SolibSymlink _solib_darwin_x86_64/_U@local_Ulibstdcpp_S_S_Clibstdc++___U/libstdc++.6.dylib failed: 1 input file(s) do not exist
INFO: Elapsed time: 0.415s, Critical Path: 0.01s
INFO: 11 processes: 11 internal.
FAILED: Build did NOT complete successfully
```

The examples should be updated (or potentially the underlying [rust_bindgen](https://bazelbuild.github.io/rules_rust/rust_bindgen.html) rules) to support MacOS.

csmulhern:
One issue here (other than the local_libstdcpp issue above) is that no install name is used by bazel when compiling dynamic libraries on macOS. This results in binaries produced by rust_* using relative search paths for dynamic libraries, which are not appropriate for the produced binary. The cc actions take the additional step of patching up their binaries using install_name_tool to work around this (see: https://cs.opensource.google/bazel/bazel/+/master:tools/cpp/osx_cc_wrapper.sh). A new flag has been added to bazel that will fix this behavior: --incompatible_macos_set_install_name. It is currently broken, but will be fixed by https://github.com/bazelbuild/bazel/pull/13427.

[...]

Issue #540, opened 2020-12-21
Runtime linking support for multiple platforms


UebelAndre:
Can anyone explain why this is the case?
https://github.com/bazelbuild/rules_rust/blob/22d67305d80fbc1f0f3e959b3d520e908cf7f9aa/rust/private/rustc.bzl#L758-L762

Looking back at https://github.com/bazelbuild/rules_rust/pull/121 and then https://github.com/bazelbuild/rules_rust/pull/61 there doesn't seem to be anything calling out that this is specific to linux other than the fact that the PR was only scoped to linux.

Is this still correct that only this behavior is only possible on linux? If so, why is linux the only platform that can do this? If not, what would be the risk in expanding this to other platforms?

(hoping @mfarrugi might know since they appear to have been the original author)

mfarrugi:
I didn't have the motivation, knowledge, or resources to implement dynamic
linking support outside of Linux.

PR #766, opened 2021-06-12
Enables runtime linking for darwin platforms

csmulhern:
Runtime linking with rpaths works correctly on darwin with only this change, for dylibs that have the correct extension (dylib) and the correct rpath configuration. Runtime linking against dynamic libraries created by bazel targets remains broken, due to them not having install names set correctly. That is fixed by using --incompatible_macos_set_install_name after https://github.com/bazelbuild/bazel/pull/13427.

I have attached a zip file with a simple example workspace demonstrating this.

[example.zip](https://github.com/bazelbuild/rules_rust/files/6641587/example.zip)

Run:
```bash
(cd lib; ./build.sh)
bazel run //:main
```

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -735,7 +735,12 @@ def _compute_rpaths(toolchain, output_dir, dep_info):
     dylibs = [lib for lib in preferreds if lib.basename.endswith(toolchain.dylib_ext) or lib.basename.split(".", 2)[1] == toolchain.dylib_ext[1:]]
     if not dylibs:
         return depset([])
-    if toolchain.os != "linux":
+
+    # For darwin, dylibs compiled by Bazel will fail to be resolved at runtime
+    # without a version of Bazel that includes
+    # https://github.com/bazelbuild/bazel/pull/13427. This is known to not be
+    # included in Bazel 4.1 and below.
+    if toolchain.os != "linux" and toolchain.os != "darwin":
         fail("Runtime linking is not supported on {}, but found {}".format(
             toolchain.os,
             dep_info.transitive_noncrates,


UebelAndre:
Seems related to https://github.com/bazelbuild/rules_rust/issues/540

UebelAndre (changes requested):
Thanks for the PR! I think this looks good but do you think you could Bazelify the example you have and add it to `examples/runtime_linking` or something so we can be sure to avoid potential regressions in the future 😅 🙏?

csmulhern:
My intention is to fix #621 and so this will be bazel-ified and checked by CI by those targets. The difficulty with adapting the example I have linked here is that it requires using an unreleased version of Bazel so that --incompatible_macos_set_install_name can be set. Otherwise I'd need to write a new rule that post processes cc_library targets to set the install name of the dylib, which I'm not sure is worth the effort. I wanted to submit this PR independently because it is preventing me from linking against a vulkan dylib that I am cc_importing in my project, ahead of a full fix for #621. I can also wait to submit a PR with a full fix for #621, which involves:

1) Setting --incompatible_macos_set_install_name (which will work correctly in an upcoming release of Bazel, hopefully 4.2: https://github.com/bazelbuild/bazel/issues/13558#issuecomment-859973642).
2) This change.
3) The change outlined in my second paragraph here: https://github.com/bazelbuild/rules_rust/issues/621#issuecomment-832382072.

This will specifically solve the issues with the tests in @examples/ffi/rust_calling_c/BUILD.bazel on macOS.

To resolve the issues with the tests in @examples/ffi/rust_calling_c/simple/BUILD.bazel, we also need to resolve #741, which I believe should be resolved by using the dynamic library specified by the cc toolchain. I also plan to take a look at this.

illicitonion (approved):
Works for me - maybe worth a comment above the if statement explaining that darwin doesn't work before (link to PR/commit) and that it's known not to work <4.2

UebelAndre (approved):
I see, thanks! I agree with https://github.com/bazelbuild/rules_rust/pull/766#pullrequestreview-682960848 but otherwise looks good and am excited to see this issue getting fixed 😄

csmulhern:
Done.

UebelAndre (approved):
Thank you so much!

PR #769, opened 2021-06-17
Removed deprecated `out_dir_tar` attribute

UebelAndre:
It's been deprecated for over a year it seems https://github.com/bazelbuild/rules_rust/pull/360

I think it's an acceptable time to remove it.


illicitonion (approved):
100% long enough - thanks!

Issue #604, opened 2021-02-24
What is rust_binary::out_binary?


UebelAndre:
https://bazelbuild.github.io/rules_rust/defs.html#rust_binary-out_binary

And what's the status of the deprecated attributes on the rule?

hlopko:
I just saw we have this attribute 5 minutes ago :)

dfreese:
It's a wasm related thing. I can try dig up the initial issue.

UebelAndre:
@dfreese Can I instead ask that you add documentation for that attribute? Instead of writing anything here? 🙏

UebelAndre:
related https://github.com/bazelbuild/rules_rust/pull/312

Issue #771, opened 2021-06-17
FR: Add `rust_wasm_library`


UebelAndre:
There appears to be a desire for something like this which stems from https://github.com/bazelbuild/rules_rust/pull/312

An insightful bit of context can be found in https://github.com/bazelbuild/rules_rust/pull/240#discussion_r337676010 but essentially, it seems to be `rust_binary` is being used in some (now) [wasm](https://www.rust-lang.org/what/wasm) edge cases but instead, we should probably have a more first class rule for supporting these cases?

I came across this while trying to understand https://github.com/bazelbuild/rules_rust/issues/604 which at first glance just seems to be a quick solution to unblock a larger set of functionality but could ultimately stand to be re-reviewed.

PiotrSikora:
See my previous comments:
- https://github.com/bazelbuild/rules_rust/pull/240#discussion_r337676010
- https://github.com/bazelbuild/rules_rust/pull/240#discussion_r340827459
- https://github.com/bazelbuild/rules_rust/pull/312#issuecomment-624628421
- https://github.com/bazelbuild/rules_rust/pull/626#issuecomment-793582479

We maintain a custom `wasm_rust_binary` target in Envoy with automatic transitions to Wasm/WASI targets:
See: https://github.com/envoyproxy/envoy/blob/f91f79b7a58c9b429a23b2882f67128071fcf495/bazel/wasm/wasm.bzl

And Emscripten did something similar with `wasm_cc_binary`:
https://github.com/emscripten-core/emsdk/tree/eaa49511f401c18fc3b87ef644524936540977d7/bazel#using-wasm_cc_binary

But this should really be supported directly in `rules_rust`.

PiotrSikora:
> @PiotrSikora Could you reiterate why the `rust_library` target needs to return an executable? It seems like you should be able to grab the file from [CrateInfo::output](https://github.com/bazelbuild/rules_rust/blob/176aff74f04f696c09100f4ff5346067a6d31095/rust/private/providers.bzl#L26). Could that be used in place of `out_binary`?

Bazel won't even build library with `.wasm` extension, see this https://github.com/bazelbuild/rules_rust/pull/626#issuecomment-793582479.

UebelAndre:
> Bazel won't even build library with `.wasm` extension, see this [#626 (comment)](https://github.com/bazelbuild/rules_rust/pull/626#issuecomment-793582479).

Sounds like you're referring to this?
```
Traceback (most recent call last):
[...]

PR #772, opened 2021-06-17
Added documentation to `rust_binary::out_binary`

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/604

The description here still feels generic and I'm not really sure how best to communicate what this is for and what it is without just linking it's history. I made https://github.com/bazelbuild/rules_rust/issues/771 since it sounds like a better solution to what seems like an ambiguous issue.


hlopko:
As I remember it, this attribute a temporary bandaid to make some part of wasm build and it will eventually be removed. But I think we shouldn't trust me here, and @dfreese will know more :)

My questions:
* do we need to document the attribute at all?
* should we mark it as experimental and tell people not to use it?
* is presence of this attribute blocking any progress, or making any progress harder? Will creating a `rust_wasm_binary` pay for itself if the only difference is this attribute?

UebelAndre:
>     * do we need to document the attribute at all?

All user facing attributes should be documented. I don't think there should be any exception.



>     * should we mark it as experimental and tell people not to use it?

That's a definite possibility depending on what it's really used for (which is still kinda mysterious to me)



>     * is presence of this attribute blocking any progress, or making any progress harder? Will creating a `rust_wasm_binary` pay for itself if the only difference is this attribute?

I don't think it's blocking any progress but it's a mysterious bit of functionality that may or may not be what someone needs. I care most that this attribute is identified and that there's a plan for it. Either it's something that's going to be removed and we document it as deprecated with some kind of issue link or it's actually removed or replaced by something we think fits better, like `rust_wasm_binary` (which I'm not confident in as a solution but seemed to pop up in more than one place so I wanted to track it as a potential solution).

I would *love* some input on what this is, what it should be, and what to write here. But again, I do think something should be written.

PiotrSikora (approved):
This is a hack that we use to force `crate_type=cdylib` to be built as an executable when targeting Wasm platforms.

`crate_type=cdylib` is needed because we're building Wasm reactors (without `main()` function), and Rust toolchain complains when we're building a binary without it. This is fixed on `nightly` with `RUSTFLAGS="-Z wasi-exec-model=reactor"`, but it's going to be a while before it's stabilized.

`out_binary=True` is needed because we need to mark the output as executable, otherwise Bazel complains that it doesn't have any of the allowed extensions for a library:
```
ERROR: .../envoy/test/extensions/filters/http/wasm/test_data/BUILD:89:17: in rust_binary rule //test/extensions/filters/http/wasm/test_data:_wasm_resume_call_rust_wasm: 
        File ".../external/rules_rust/rust/private/rust.bzl", line 283, column 32, in _rust_binary_impl
                return rustc_compile_action(
        File ".../external/rules_rust/rust/private/rustc.bzl", line 583, column 29, in rustc_compile_action
                return establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration) + [
        File ".../external/rules_rust/rust/private/rustc.bzl", line 640, column 59, in establish_cc_info
                library_to_link = cc_common.create_library_to_link(
Error in create_library_to_link: '_wasm_resume_call_rust_wasm.wasm' does not have any of the allowed extensions .so, .dylib or .dll
ERROR: Analysis of target '//test/extensions/filters/http/wasm:wasm_filter_test' failed; build aborted: Analysis of target '//test/extensions/filters/http/wasm/test_data:_wasm_resume_call_rust_wasm' failed
```

This could be removed if we had `rust_wasm_binary`.

rust/private/rust.bzl
```diff
@@ -885,7 +885,10 @@ _rust_binary_attrs = {
         cfg = "exec",
         allow_single_file = True,
     ),
-    "out_binary": attr.bool(),
+    "out_binary": attr.bool(
+        doc = "Force a target, regardless of it's `crate_type`, to always mark the file as executable.",
```
hlopko, line 889: WDYT about mentioning that this is really only for wasm, how exactly it is supposed to be used in wasm, and that it will be removed in the doc?

UebelAndre: How's this?

hlopko (approved):
LGTM, thank you!

PR #773, opened 2021-06-20
Added util module for rust repository rules.

UebelAndre:
I find it difficult to understand what the repository rules in `//rust:repositories.bzl` are doing with all the noise of the template BUILD files. This is a minor cleanup to make these rules and macros more readable.


UebelAndre:
When I opened this PR I thought it'd just be the `BUILD_*` macros but turned out to be a bit more. I think it's still good cleanup but I can respect the argument that it's too much indirection (but I don't think this crosses a line).

illicitonion (approved):
Looks reasonable to me :) I definitely agree that the file as-is is large and unwieldy!

PR #774, opened 2021-06-20
Fixed compile warning

UebelAndre:
Fixes the following warning
```
warning: function is never used: `native_dep`
 --> test/unit/cc_info/proc_macro.rs:4:5
  |
4 |     fn native_dep() -> isize;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted
```


test/unit/cc_info/proc_macro.rs
```diff
@@ -1,6 +1,7 @@
 use proc_macro::TokenStream;
 
 extern "C" {
+    #[allow(dead_code)]
```
illicitonion, line 4: This is a little odd... Do you know why this `extern "C"` block is here in the first place? I feel like we could remove it without affecting the test at all?

I'd rather either remove it, or modify `use_native_dep` to call `native_dep()`, than keep around the unused thing...

hlopko: So a little story. Starlark unit tests are unit tests, but they require that targets under test are built correctly. So `cc_info_test` will actually compile `bin_using_native_dep.rs` and also the `proc_macro.rs`. Because of that I tried to make their content somewhat adequate (not very successfully).

I see 2 ways forward:

1) Let's make the content of those files the bare minimum. `bin_using_native_dep.rs` will not use native dep, it will not test that the binary is linked correctly with a C++ dep. Some other test should verify that.
2) Make the content more reasonable, and rely on the additional test coverage provided by this. `proc_macro.rs` could look like this:

```
use proc_macro::TokenStream;

extern "C" {
    fn native_dep() -> isize;
}
#[proc_macro_derive(UsingNativeDep)]
pub fn use_native_dep(input: TokenStream) -> TokenStream {
    println!("{}", unsafe { native_dep() });
    input
}

```

Wdyt?

illicitonion: I would lean towards 2, but I don't have a strong preference.

UebelAndre: I can take a stab at 2. I haven't had time to come back to this yet and initially thought this was a quick thing so it might take me a weekend or two to find the time 😅

UebelAndre: Does this test require a native dependency? Seems odd to have this cross-test dependency... But the test is now updated.

illicitonion (approved):

hlopko (approved):

test/unit/cc_info/proc_macro.rs
hlopko: Yup it's testing that proc macro gets the native dependency (CcInfo) correctly.

PR #781, opened 2021-06-22
Merge deps of related crates in Rust Analyzer support

hlopko:
It can happen a single source file is present in multiple crates - there can
  be a `rust_library` with a `lib.rs` file, and a `rust_test` for the `test`
  module in that file. Tests can declare more dependencies than what library
  had. Therefore we had to collect all `RustAnalyzerInfo`s for a given crate
  and take deps from all of them.

  There's one exception - if the dependency is the same crate name as the
  the crate being processed, we don't add it as a dependency to itself. This is
  common and expected - `rust_test.crate` pointing to the `rust_library`.


rust/private/rust_analyzer.bzl
```diff
@@ -116,51 +118,79 @@ def _crate_id(crate_info):
     """
     return "ID-" + crate_info.root.path
 
-def create_crate(ctx, info, crate_mapping):
-    """Creates a crate in the rust-project.json format
+_DepRefInfo = provider(
+    doc = "RustAnalyzerInfo holds rust crate metadata for targets",
+    fields = {
+        "crate": "BuildInfo: build info for this crate if present",
+        "name": "List[String]: features or other compilation --cfg settings",
```
UebelAndre, line 125: nit, maybe this field could be balled `cfg_flags`, `flags` or `settings`?

rust/private/rust_analyzer.bzl
```diff
@@ -116,51 +118,79 @@ def _crate_id(crate_info):
...
     Args:
         ctx (ctx): The rule context
-        info (RustAnalyzerInfo): The crate RustAnalyzerInfo for the current crate
+        infos (list of RustAnalyzerInfos): RustAnalyzerInfos for the current crate
         crate_mapping (dict): A dict of {String:Int} that memoizes crates for deps.
 
     Returns:
         (dict) The crate rust-project.json representation
     """
+    canonical_info = infos[0]
```
UebelAndre, line 150: Can you add something to handle the case where `infos` might be empty?

UebelAndre (changes requested):
Nice one! A couple of small things but otherwise looks good 😄

hlopko:
Related: https://github.com/bazelbuild/rules_rust/issues/704

rust/private/rust_analyzer.bzl
hlopko: I actually realized that even though lists and dicts are not hashable in Starlark, tuples are. Now I'm using an inline dict comprehension instead of collections.dict and I don't need an explicit provider type anymore.

rust/private/rust_analyzer.bzl
hlopko: Done.

scentini (approved):

PR #786, opened 2021-06-23
fix: Set SYSROOT environment variable to the directory of the rust_lib files

wcalandro:
This change fixes some issues that the Kythe team is having building our Rust extractor inside Bazel


hlopko:
Could you elaborate on what issues those are and why is this the right fix? (I honestly don't know)

To fix docs, you need to add `@bazel_skylib//lib:paths` into deps of this target: https://github.com/bazelbuild/rules_rust/blob/main/docs/BUILD.bazel#L13

rust/private/rustc.bzl
```diff
@@ -479,9 +480,8 @@ def construct_arguments(
         getattr(attr, "compile_data", []),
     ))
 
-    # This empty value satisfies Clippy, which otherwise complains about the
-    # sysroot being undefined.
-    env["SYSROOT"] = ""
+    # Set the SYSROOT to the directory of the rust_lib files passed to the toolchain
+    env["SYSROOT"] = paths.dirname(toolchain.rust_lib.files.to_list()[0].short_path)
```
illicitonion, line 484: I'm not sure this is the right path to be using here:

The value of the first file in that depset I'm seeing on my machine is `../rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libaddr2line-6aebf63c63f128f7.rlib` (and that just happens to be the first file).

Outside of Bazel, if I run `rustc --print=sysroot` I'm given:
```
/Users/dwh/.rustup/toolchains/stable-x86_64-apple-darwin
```

which corresponds to `../rust_darwin_x86_64` rather than `../rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib`

I'm not sure the best way to get the path `../rust_darwin_x86_64`.

Either way, though, a test case which fails without this change would be handy :)

wcalandro: Sorry for the late response, I've been moving. This change fixes a bug that comes about when using these rules to build internally, and there isn't really an applicable test case that is external-only. As for the sysroot question, it acts as more of an LD_LIBRARY_PATH that tells the compiler where to find applicable .rlib and .so files. Both the parent sysroot provided by the `rustc --print=sysroot` command and the path we compute in the rules work, but the path we generate is simply more concise since we are indicating the exact path where all the .rlib files are

illicitonion: Could you try to extract an example that points at the real need for setting the sysroot? While I totally believe that this _is_ necessary, we have plenty of examples which don't need it, and I can believe we would break this behaviour in the future without a test/example... Do you know what's special about your code that needs it? Maybe you're using a compiler plugin, or a special kind of proc macro, or a non-standard target-type, or something?

wcalandro: The code for [Kythe's Rust extractor](https://github.com/kythe/kythe/tree/master/kythe/rust/extractor) links to some compiler dynamic libraries, specifically librustc_driver-{version}.so and libstd-{version}.so . Externally, Bazel has no problem finding and linking to these libraries. Internally, Bazel has some intricacies with symlinks that cause path confusions when looking for these libraries as it tries to navigate various symlinks, which causes the linker to fail at finding these libraries. The easiest solution to this problem is to always explicitly tell the compiler the absolute paths to these libraries, so that it doesn't attempt to infer them and chase problematic symlinks. This change has no effect on those utilizing Bazel externally, it simply solves an issue related to our internal usage of these rules.

illicitonion: That sounds pretty reasonable - we may change this to a different value in the future, and hopefully that won't break things for you in the future!

illicitonion (approved):

illicitonion (approved):

PR #789, opened 2021-06-24
`test` package is no longer reliant on `@examples` repository

UebelAndre:
This feels like an oversight. I think it'd be good to keep the `test` module isolated so it's clearer what's being tested and there's less chance for unexpected changes.


illicitonion (approved):
Seems reasonable :)

PR #790, opened 2021-06-24
Deleted deprecated `rustfmt_generator` rule

UebelAndre:
There's now more streamlined [rustfmt rules](https://bazelbuild.github.io/rules_rust/rust_fmt.html)


illicitonion (approved):

rdelfin:
Is there an equivalent for the `rustfmt_generator()` rule? Or can we no longer run rustfmt on generated code using these rules?

UebelAndre:
Rustfmt is still in the `rust_toolchain` so if you have a rule that generates rust code, would it be sufficient to update that rule to also run rustfmt? The issue in running rustfmt on generated code is that most code generator rules don't run rustfmt and the result is that those files are always causing the aspect to fail. May I ask what you're doing for code generation?

PR #791, opened 2021-06-24
Make it easier to use rust_test_suite in macros.

sayrer:
As written, this function produces duplicate test rules if used in macros. This patch aims to make it easy to run integration tests with different `crate_features`, via a macro.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -1210,8 +1210,9 @@ def rust_test_suite(name, srcs, **kwargs):
         if not src.endswith(".rs"):
             fail("srcs should have `.rs` extensions")
 
+        # Prefixed with `name` to allow parameterization with macros
         # The test name should not end with `.rs`
-        test_name = src[:-3]
+        test_name = name + src[:-3]
         rust_test(
             name = test_name,
             crate_name = test_name.replace("/", "_"),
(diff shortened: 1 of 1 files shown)


UebelAndre:
> As written, this function produces duplicate test rules if used in macros. This patch aims to make it easy to run integration tests with different `crate_features`, via a macro.

I'm not sure I understand what the prefixing gets you. Can you demonstrate how this might be used?

sayrer:
```
def mycrate_build(name, crate_features):
    rust_library(
        name = name,
        srcs = native.glob(["src/**/*.rs"]),
        aliases = {
        },
        crate_features = crate_features,
        crate_name = name,
    ...
    )

    # Works
    rust_test(
        name = name + "_unit_tests",
        crate = ":" + name,
        crate_features = crate_features,
    )

    # Does not work due to duplicate rust_test rules
    rust_test_suite(
        name = name + "_integration_tests",
        srcs = native.glob(["tests/**/*.rs"]),
        crate_features = crate_features,
    )
```

sayrer:
I could see other ways to solve this, but it would be nice if the rule made it easy.

illicitonion (approved):
Works for me :) Thanks!

UebelAndre (approved):
Ah, thanks for the example! Looks good to me 😄

PR #792, opened 2021-06-25
Added standalone targets for `rust_toolchain` components.

UebelAndre:
I'm not sure if there's a better way to do this but I certainly feel it's better than:
https://github.com/bazelbuild/rules_rust/blob/8cb0c78cc77be4687f6b50539752329ae8afcfa0/tools/rustfmt/BUILD.bazel#L7-L16

This PR adds targets for parsing specific components of a `rust_toolchain`:
- `@rules_rust//rust/toolchain:current_exec_rustfmt_files`
- `@rules_rust//rust/toolchain:current_exec_rustdoc_files`
- `@rules_rust//rust/toolchain:current_exec_rustc_srcs_files`
- `@rules_rust//rust/toolchain:current_exec_rustc_lib_files`
- `@rules_rust//rust/toolchain:current_exec_rustc_files`
- `@rules_rust//rust/toolchain:current_exec_rust_stdlib_files`
- `@rules_rust//rust/toolchain:current_exec_clippy_files`
- `@rules_rust//rust/toolchain:current_exec_cargo_files`

These targets give users the ability to consume these files at runtime ([cargo-raze](https://github.com/google/cargo-raze/blob/ee2effee5b6c95cfcfd5e496677ce61dcd35c472/tools/BUILD.bazel#L44-L79) is also doing this and would benefit from these targets).


test/toolchain/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
+load("//rust:toolchain.bzl", "toolchain_tool")
+
+toolchain_tool.cargo(
```
hlopko, line 3: Wdyt about putting these targets somewhere in the rules_rust, so users don't have to put them into their BUILD files, but can use `@rules_rust//rust/toolchain:current_rustfmt_files` (name to be discussed)?

rust/toolchain.bzl
```diff
@@ -301,3 +301,170 @@ See @rules_rust//rust:repositories.bzl for examples of defining the @rust_cpuX r
 with the actual binaries and libraries.
 """,
 )
+
+def _cargo_tool_impl(ctx):
```
hlopko, line 305: What do you think about a single rule that gets the name of the tool from an attribute, and then there would be a `if ctx.attr.tool` == "rustfmt":` conditional in the impl? I prefer that to duplicated logic in multiple files, even though they are so close to each other.

rust/toolchain.bzl
```diff
@@ -301,3 +301,170 @@ See @rules_rust//rust:repositories.bzl for examples of defining the @rust_cpuX r
...
+    toolchain = ctx.toolchains[str(Label("//rust:toolchain"))]
+
+    # Executables must be produced within the rule marked executable. To
```
hlopko, line 308: Question - if rustfmt binary that we declare in the rust toolchain is executable, does it stay executable when we provide it in this rule through DefaultInfo? I honestly don't know.

One of the reasons why to make these rules executable is so they work as tools in genrules. Are there other reasons that you care about?

rust/toolchain.bzl
```diff
@@ -301,3 +301,170 @@ See @rules_rust//rust:repositories.bzl for examples of defining the @rust_cpuX r
...
+        target_file = toolchain.cargo,
+        is_executable = True,
+    )
+
+    return [DefaultInfo(
+        files = depset([toolchain.cargo]),
+        executable = cargo,
+    )]
+
+_cargo_tool = rule(
```
hlopko, line 322: Regarding naming, I'm wondering if it makes sense to encode some/all of these things in the name:

* that this is the executable and all its runtime dependencies (data dependencies)
  * we used `_files` suffix for that in C++ rules, but that makes the fact that there is an executable a little bit less clear
* that this tool has been taken from the current Rust toolchain
  * maybe put `current` into the name?
* that this tool has been taken from the current Rust toolchain in the exec configuration (so it's supposed to be runnable on the execution platform, aka my computer, not on the computer where my Rust library I'm building using rules rust runs on)
  * maybe put "exec" in the name?
  
By maybe I really mean let's discuss because maybe there are better ways to communicate these things to the user and/or we shouldn't communicate them at all?

test/toolchain/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+    name = "cargo",
+)
+
+toolchain_tool.clippy(
```
hlopko, line 7: What is this test testing? Wdyt about adding a test that actually executes these files?

test/toolchain/BUILD.bazel
UebelAndre: I really like that. I'll refactor this to do that.

rust/toolchain.bzl
UebelAndre: I suppose they don't have to be executable but I think it's more intuitive since I (and I assume most people) would think of them as binaries and perhaps want to write rules consuming them via [ctx.executable](https://docs.bazel.build/versions/main/skylark/lib/ctx.html#executable). I could make these not executable if you think it's the smoother approach.

rust/toolchain.bzl
UebelAndre: Updated. Sad to say I'd forgotten about that aspect of this PR. The code I wrote this for made more sense to make the separate rules but I really like the introduction of the `current_*` targets since that's ultimately what seems to be in demand here. But the rule is now one rule which takes an additional argument.

rust/toolchain.bzl
UebelAndre: I've updated the PR to have `toolchain_tool` and `toolchain_files`. Maybe it's not valuable to have an executable target? Thoughts and feelings?

rust/toolchain.bzl
UebelAndre: I like `current_exec_{tool}` but I'm not sure about `_files`. I don't think we should guarantee that each target has everything it needs to be fully functional. I don't think the `cargo` tool should come with `rustc` or `rustc` come with `rustc_lib`. If a user is interested in wiring all that up, then I'd want to ensure all the components they're looking for are there, but the tools are really more for edge case stuff where you need a particular binary or source file for some reason. Which would mean being able to ask for `ctx.executable.rustc` in a rule would be the nicest pattern IMO.

test/toolchain/BUILD.bazel
UebelAndre: I've added more involved tests.

rust/toolchain.bzl
UebelAndre: Ok, I've done a complete 180 here and added runfiles so the tools were runnable. This made it actually testable and still maintains some ability to parse just the file/binary requested.

UebelAndre: I was thinking of opening an actual discussion for naming but it feels a bit over kill. I've made the rule private so the only questions now are, do we like the idea of these targets (PR description has been updated) and how should they be named if we like them.

cc @illicitonion @dfreese

rust/toolchain.bzl
UebelAndre: Ok, there's now no longer an executable which I think turns out to be way cleaner.

test/toolchain/BUILD.bazel
UebelAndre: I went with `current_exec_rustfmt_files` per the suggestion in https://github.com/bazelbuild/rules_rust/pull/792#discussion_r659588298

I'll resolve this to consolidate the conversation.

UebelAndre:
Windows is very unfun... but this PR is ready for review again

rust/toolchain.bzl
UebelAndre: For the record, I did end up going with `current_exec_{tool}_files` since the targets may have runfiles depending on which one is requested.

illicitonion: I don't have any strong opinions here :)

test/current_toolchain_tools/BUILD.bazel
```diff
@@ -0,0 +1,43 @@
+# Executable targets will output a pattern similar to the following
+# cargo 1.53.0 (4369396ce 2021-04-27)
+# Also Note, rustc_srcs is too big for this test
+_FILES = {
```
hlopko, line 4: Nit: rename this directory from `current_toolchain_tools` to `current_toolchain_files` to match the `toolchain_files` rule name?

test/current_toolchain_tools/current_exec_files_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -xeuo pipefail
```
hlopko, line 3: Did you leave -x there intentionally? (I typically only use it for debugging)

hlopko (approved):
LGTM, thank you!

test/current_toolchain_tools/BUILD.bazel
UebelAndre: Done

test/current_toolchain_tools/current_exec_files_test.sh
UebelAndre: I thought it was handy for debugging, yeah, but it can be removed.

PR #793, opened 2021-06-25
Promoted internal `_tidy` function to the utils module as `dedent`

UebelAndre:
The functionality here would be useful elsewhere so I thought it should go in `utils.bzl` and I thought it'd be clearer to borrow the term [dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) from python.

### DIFF
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -243,7 +243,10 @@ def crate_name_from_attr(attr):
     return crate_name
 
 def dedent(doc_string):
-    """Tidy excess whitespace in docstrings to not break index.md
+    """Remove any common leading whitespace from every line in text.
+
+    This functionality is similar to python's `textwrap.dedent` functionality
+    https://docs.python.org/3/library/textwrap.html#textwrap.dedent
 
     Args:
         doc_string (str): A docstring style string
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -241,3 +241,32 @@ def crate_name_from_attr(attr):
             "Consider adding a crate_name attribute to set a valid crate name",
         )
     return crate_name
+
+def dedent(doc_string):
+    """Tidy excess whitespace in docstrings to not break index.md
+
+    Args:
+        doc_string (str): A docstring style string
+
+    Returns:
+        str: A string optimized for stardoc rendering
+    """
+    lines = doc_string.splitlines()
+    if not lines:
+        return doc_string
+
+    # If the first line is empty, use the second line
+    first_line = lines[0]
+    if not first_line:
+        first_line = lines[1]
+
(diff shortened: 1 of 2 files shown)


rust/private/utils.bzl
```diff
@@ -241,3 +241,32 @@ def crate_name_from_attr(attr):
             "Consider adding a crate_name attribute to set a valid crate name",
         )
     return crate_name
+
+def dedent(doc_string):
+    """Tidy excess whitespace in docstrings to not break index.md
+
```
dfreese, line 247: Can you add the python docs reference here for future reference?

dfreese (approved):

rust/private/utils.bzl
UebelAndre: done!

hlopko:
Just a drive by - consider adding a test if you still have the behavior of the function cached in brain :) https://docs.bazel.build/versions/main/skylark/testing.html#for-testing-starlark-utilities

UebelAndre:
> Just a drive by - consider adding a test if you still have the behavior of the function cached in brain :) https://docs.bazel.build/versions/main/skylark/testing.html#for-testing-starlark-utilities

I'll add it to my list 😄

UebelAndre:
> Just a drive by - consider adding a test if you still have the behavior of the function cached in brain :) https://docs.bazel.build/versions/main/skylark/testing.html#for-testing-starlark-utilities

I feel like if this is going to be tested then it should go in [bazel_skylib](https://github.com/bazelbuild/bazel-skylib) seems like functionality that would make folks happy. Or maybe even better, the `doc` attribute would be smart enough to perform a dedent, or maybe `stardoc`? It also feels kinda odd `_tidy`/`dedent` exists. We now pay a runtime cost to have the code look nicer? I'm all for maintainability but I think we could get away with writing maintainable code without `dedent` here. Thoughts? Should this be promoted? Should it be removed? Is this the thing that really needs tests (over others)?

hlopko:
First of all, writing a test is independent from (and maybe a prerequisite for) later points.

After we have an implementation that is field tested and of good quality I think folks at https://github.com/bazelbuild/bazel-skylib might accept it. Historically they've been very pedantic and strict about what goes into Skylib and I don't know the exect policy. We should try.

If you want to automatically dedent the string in `doc` attribute, you'll have to fix stardoc (somewhere here: https://osscs.corp.google.com/bazel/bazel/+/master:src/main/java/com/google/devtools/build/skydoc/SkydocMain.java;l=103?q=SkydocMain&ss=bazel). I don't think the runtime cost of dedenting is significant enough to worry about it. And if it is, the short term fix is to extract doc strings to constants and not having extra indentation inside.

PR #794, opened 2021-06-25
Added checksums for additional rust assets to internal list

UebelAndre:
I don't think we should be maintaining a separate list of revisions for `rustfmt` and I think it'd be useful to have additional artifact types being tracked:
- cargo
- clippy


illicitonion (changes requested):
I'm on board with the additions. I'm not so sure about the rustfmt piece - while I agree that _now_ we don't need a separate list, before relatively recently (I would guess 1.48.0, from the contents of `util/fetch_shas_RUSTFMT_VERSIONS.txt`) rustfmt was versioned separately from the rest of the toolchain... I suspect this _may_ break rustfmt support for pre-1.48.0 rust versions, because it will try to download files which don't exist?

UebelAndre:
@illicitonion How would this break rustfmt support? The file I deleted was only used to determine what versions of rustfmt to query for. The fetch tool seems designed to allow for failures when something doesn't exist so the only change for rustfmt here is that we'll try to fetch more versions than there may be available. Nothing has changed about how users define what version of rustfmt they want.

illicitonion (approved):
Aha! Sorry, I was mis-reading - thanks!

PR #796, opened 2021-06-25
Added `rules_rust` prefix to transitive workspace names.

UebelAndre:
I prefer all repositories/transitive workspaces generated by any repository to use the repository name as a prefix (or be user defined). This helps me quickly identify where things came from when crawling `output_base`.

For me this is minor improvement when debugging things.

Requires:
- https://github.com/bazelbuild/rules_rust/pull/803

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -2,17 +2,17 @@
 default_linux_targets: &default_linux_targets
   - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
   - "..."
-  - "@examples//..."
+  - "@rules_rust_examples//..."
   # TODO: Switch manual tag to platform constraint after bazel 4.0.
   - "//test/versioned_dylib:versioned_dylib_test"
   # Bindgen currently only has a working toolchain for 18.04
-  - "-@examples//ffi/rust_calling_c/simple/..."
+  - "-@rules_rust_examples//ffi/rust_calling_c/simple/..."
 tasks:
   ubuntu1804:
     build_targets: *default_linux_targets
     test_targets:
       - "..."
-      - "@examples//..."
+      - "@rules_rust_examples//..."
       # TODO: Switch manual tag to platform constraint after bazel 4.0.
       - "//test/versioned_dylib:versioned_dylib_test"
     build_flags:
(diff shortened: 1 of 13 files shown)


UebelAndre:
I'm pretty confused as to why I'd be getting these errors on windows...
```
error[E0463]: can't find crate for `rustc_std_workspace_core` which `std` depends on

For more information about this error, try `rustc --explain E0463`.
```

UebelAndre:
> It seems this issue stems from the file path on windows being too long. I'm not sure if there's a better way to handle this...
> 
> @hlopko do you have any thoughts? I see you've got a TODO in this code so hopefully you'll have some context 😅

This seems to have been the case. I'm surprised `rustdoc_test` is so prone to this issue but maybe it's related to the batch script that orchestrates running the test?
https://github.com/bazelbuild/rules_rust/blob/26f0b037b8a38b60ee9a69ff99fd2e5a1855f09c/rust/private/rustdoc_test.bzl#L155-L188

This could probably be tracked by https://github.com/bazelbuild/rules_rust/issues/804 but it's likely not specific to sandboxing and really more windows specific. I'm unfortunately no windows expert so issues on that platform are hard for me to interpret.

hlopko:
>  @hlopko do you have any thoughts? I see you've got a TODO in this code so hopefully you'll have some context sweat_smile

I'm not sure I follow, what code is that? Can you send me a link?

UebelAndre:
> I'm not sure I follow, what code is that? Can you send me a link?

https://github.com/bazelbuild/rules_rust/blob/0d98d6b41496c4d66ba60aa3ab9c2dea9a491dba/rust/private/rustdoc_test.bzl#L101

Your name was there which is why I pinged you 😅 

But I've since concluded the issue was [windows max path limitations](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd)

hlopko:
So what is the desired resolution here?

Thinking about the problem you're solving, does it need to be solved? Examples should not be included in `rules_rust` releases, people should not use them from monorepos, people should not put `local_repository` into their WORKSPACE to make examples work.

Therefore, renaming `examples` -> `rules_rust_examples` will only be visible to people who ran bazel in the `rules_rust` checkout locally. From those people, only very few will ever need to debug execroot content, and IMHO to those it will be more expected to see 'examples' repository that covers `/examples` directory, and not `rules_rust_examples` that covers `/examples` directory.

And there should never be a naming conflict in `rules_rust` (we should never have multiple repositories named `examples`). 

What do you think?

UebelAndre:
> So what is the desired resolution here?
> 
> Thinking about the problem you're solving, does it need to be solved? Examples should not be included in `rules_rust` releases, people should not use them from monorepos, people should not put `local_repository` into their WORKSPACE to make examples work.
> 
> Therefore, renaming `examples` -> `rules_rust_examples` will only be visible to people who ran bazel in the `rules_rust` checkout locally. From those people, only very few will ever need to debug execroot content, and IMHO to those it will be more expected to see 'examples' repository that covers `/examples` directory, and not `rules_rust_examples` that covers `/examples` directory.
> 
> And there should never be a naming conflict in `rules_rust` (we should never have multiple repositories named `examples`).
> 
> What do you think?

This to me is a minor readability improvement which I'd only ever expect to impact maintainers. My thought here aligns with what I mentioned on https://github.com/bazelbuild/rules_rust/issues/830 where I'd like `docs` and `examples` to not be something loaded by the top level workspace so the name of the repository should have very little significance in that case and think a prefix would be a good quality of life improvement.

PR #797, opened 2021-06-25
Exposed supported platform triples lists

UebelAndre:
This information really ought to be public or at least consumable by other modules in `//rust`. But I think it's fine to expose publicly. There's a similar list in [cargo-raze](https://github.com/google/cargo-raze/blob/ee2effee5b6c95cfcfd5e496677ce61dcd35c472/impl/src/util.rs#L27-L50) which needs to be in sync with this so having it be part of the public API would be a good improvement.

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -36,24 +36,6 @@ SUPPORTED_T2_PLATFORM_TRIPLES = [
 
 SUPPORTED_PLATFORM_TRIPLES = SUPPORTED_T1_PLATFORM_TRIPLES + SUPPORTED_T2_PLATFORM_TRIPLES
 
-SUPPORTED_CPU_ARCH = [
-    "aarch64",
-    "arm",
-    "i686",
-    "powerpc",
-    "s390x",
-    "x86_64",
-]
-
-SUPPORTED_SYSTEMS = [
-    "android",
-    "darwin",
-    "freebsd",
-    "ios",
-    "linux",
-    "windows",
-]
-
 # CPUs that map to a "@platforms//cpu entry
 _CPU_ARCH_TO_BUILTIN_PLAT_SUFFIX = {
     "aarch64": "aarch64",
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
This works for me, but as we're currently going through "what are our stability guarantees" discussions, it'd be good to get everyone's opinions :)

hlopko:
I'm fine with exposing target triples if you have a need for that, but exposing `SUPPORTED_CPU_ARCH` and `SUPPORTED_SYSTEMS` doesn't seem necessary. And it can be interpreted as if all possible combinations of cpus with those systems are supported, which is not true.

rust/platform/triple_mappings.bzl
```diff
@@ -42,6 +78,7 @@ _SYSTEM_TO_BUILTIN_SYS_SUFFIX = {
 }
 
 _SYSTEM_TO_BINARY_EXT = {
+    "android": "",
```
hlopko, line 81: This is not related to this PR, could you please split it into a standalone PR with a focused description? Not sure there's a good way to add a test for it, but maybe you'll come up with something?

hlopko (approved):
Modulo the above comment LGTM. Thank you!

rust/platform/triple_mappings.bzl
UebelAndre: This is actually a gap in support from whenever `android` was added to the list of supported systems:
https://github.com/bazelbuild/rules_rust/blob/72c591d51311e3eb829220a49cc00700e8a6dc2b/rust/platform/platform.bzl#L54-L61

I definitely don't have bandwidth to setup an android test so I'd say either this would go in as is or support for android be dropped until a test can be added. Thoughts?

hlopko: I'm fine with merging without a test, but please still create a separate PR for that.

Do we support android now? I honestly don't know. And I don't know if we have a single android test in the repo. And I'd say if we don't have a test, then we don't really support it :)

UebelAndre: We have no test 😞 

Creating a separate PR though

UebelAndre: https://github.com/bazelbuild/rules_rust/pull/813

PR #802, opened 2021-06-25
Added `rust_stdlib_filegroup` rule, a helper for creating toolchains

UebelAndre:
This reduces the complexity of the toolchain itself making it lighter weight and allows the rest of the code to otherwise treat the rust stdlib artifact as a general filegroup but benefit from from having additional required files used for compilation/linking.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -1,23 +1,7 @@
 """The rust_toolchain rule definition and implementation."""
 
+load("//rust/private:common.bzl", "rust_common")
 load("//rust/private:utils.bzl", "dedent", "find_cc_toolchain")
-
-RustStdLibInfo = provider(
-    doc = (
-        "A collection of files either found within the `rust-stdlib` artifact or " +
-        "generated based on existing files."
-    ),
-    fields = {
-        "alloc_files": "List[File]: `.a` files related to the `alloc` module.",
-        "between_alloc_and_core_files": "List[File]: `.a` files related to the `compiler_builtins` module.",
-        "between_core_and_std_files": "List[File]: `.a` files related to all modules except `adler`, `alloc`, `compiler_builtins`, `core`, and `std`.",
-        "core_files": "List[File]: `.a` files related to the `core` and `adler` modules",
-        "dot_a_files": "Depset[File]: Generated `.a` files",
-        "srcs": "List[Target]: The original `src` attribute.",
-        "std_files": "Depset[File]: `.a` files associated with the `std` module.",
-        "std_rlibs": "List[File]: All `.rlib` files",
-    },
-)
 
 def _make_dota(ctx, file):
     """Add a symlink for a file that ends in .a, so it can be used as a staticlib.
@@ -75,7 +59,7 @@ def _rust_stdlib_filegroup_impl(ctx):
         DefaultInfo(
             files = depset(ctx.files.srcs),
         ),
-        RustStdLibInfo(
+        rust_common.stdlib_info(
             std_rlibs = std_rlibs,
             dot_a_files = dot_a_files,
             between_alloc_and_core_files = between_alloc_and_core_files,
@@ -122,7 +106,7 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
     cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
     link_inputs = []
 
-    if not RustStdLibInfo in ctx.attr.rust_lib:
+    if not rust_common.stdlib_info in ctx.attr.rust_lib:
         fail(dedent("""\
             {} --
             The `rust_lib` must be a target providing `rust_common.stdlib_info` 
@@ -130,7 +114,7 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
             See https://github.com/bazelbuild/rules_rust/pull/802 for more information.
             
         """).format(ctx.label))
-    rust_stdlib_info = ctx.attr.rust_lib[RustStdLibInfo]
+    rust_stdlib_info = ctx.attr.rust_lib[rust_common.stdlib_info]
 
     if rust_stdlib_info.std_rlibs:
         alloc_inputs = depset(
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -1,6 +1,6 @@
 """The rust_toolchain rule definition and implementation."""
 
-load("//rust/private:utils.bzl", "find_cc_toolchain")
+load("//rust/private:utils.bzl", "dedent", "find_cc_toolchain")
 
 RustStdLibInfo = provider(
     doc = (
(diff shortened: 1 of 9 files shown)


hlopko:
I like this change. It is also a breaking change, because existing projects having their own toolchains (like we are) still use the filegroup, so the search for `RustStdLibInfo` will fail.

I see 3 ways forward:

1) play the "we don't have a backward compatibility policy" card and insist on submitting as is
2) wait for me to write the policy down (which I'll do as soon as people stop uploading a dozen PRs each weekend :P
3) make the change backwards compatible (by extracting a function that does the partitioning, checking if the filegroup has `RustStdLibInfo` and if not, doing the partitioning in place) and following the proposed process at https://github.com/bazelbuild/rules_rust/issues/415#issuecomment-841081403.

Wdyt?

UebelAndre:
> I like this change. It is also a breaking change, because existing projects having their own toolchains (like we are) still use the filegroup, so the search for `RustStdLibInfo` will fail.
> 
> I see 3 ways forward:
> 
>     1. play the "we don't have a backward compatibility policy" card and insist on submitting as is
> 
>     2. wait for me to write the policy down (which I'll do as soon as people stop uploading a dozen PRs each weekend :P
> 
>     3. make the change backwards compatible (by extracting a function that does the partitioning, checking if the filegroup has `RustStdLibInfo` and if not, doing the partitioning in place) and following the proposed process at [Releases #415 (comment)](https://github.com/bazelbuild/rules_rust/issues/415#issuecomment-841081403).
> 
> 
> Wdyt?

I think I'd like to keep the changes more or less as is since the fix is extremely straight forward and, unless someone is doing something I've never heard of, should work with custom toolchains. I could drop the provider requirement and add a very clear message about how to transition? It wouldn't be the end of the world to continue to support filegroups but feels like tech-debt for a not so terrible change on the user's end? Would it be so terrible? Thoughts?

hlopko:
I'm fine with just changing the behavior. Extra points for emitting a nice error message.

However once there is a backwards compatibility policy this PR would have to follow it.

I'll give more context in how this change affects us as an example, to broaden our understanding of our users.

We use https://github.com/google/copybara to import changes in rules_rust into internal workspace. I create a change, I test it by building everything in Google using new rules, I discover a breakage, I investigate the breakage, eventually I find a fix. I rerun tests, get a review, submit the change. Since I'm not a member of the team maintaining the Rust toolchain, and the team is in a different time zone, getting their review can take a day. I usually know about all changes in the project, but toolchain folks not necessarily. So when they want to import the new changes, they will either investigate, or wait a day for me to respond.

So even trivial breaking changes, with a trivial fix, will still take a person one hour or more to fix. Multiply that with the number of users, and compare that number with the expected benefit of the change. Benefits (for our users, not for us) should always outweigh the costs (for our users). I'm not trying to comment on the benefits of this PR, I'm just giving my mental process when justifying breaking changes.

UebelAndre:
> I'm fine with just changing the behavior. Extra points for emitting a nice error message.
> 
> However once there is a backwards compatibility policy this PR would have to follow it.
> 
> I'll give more context in how this change affects us as an example, to broaden our understanding of our users.
> 
> We use https://github.com/google/copybara to import changes in rules_rust into internal workspace. I create a change, I test it by building everything in Google using new rules, I discover a breakage, I investigate the breakage, eventually I find a fix. I rerun tests, get a review, submit the change. Since I'm not a member of the team maintaining the Rust toolchain, and the team is in a different time zone, getting their review can take a day. I usually know about all changes in the project, but toolchain folks not necessarily. So when they want to import the new changes, they will either investigate, or wait a day for me to respond.
> 
> So even trivial breaking changes, with a trivial fix, will still take a person one hour or more to fix. Multiply that with the number of users, and compare that number with the expected benefit of the change. Benefits (for our users, not for us) should always outweigh the costs (for our users). I'm not trying to comment on the benefits of this PR, I'm just giving my mental process when justifying breaking changes.

How's this for the error message?
```
ERROR: /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/BUILD.bazel:72:15: in rust_toolchain rule @rust_darwin_x86_64//:toolchain_for_x86_64-apple-darwin_impl:
	File "/Users/user/Code/rules_rust/rust/toolchain.bzl", line 219, column 72, in _rust_toolchain_impl
		libstd_and_allocator_ccinfo = _make_libstd_and_allocator_ccinfo(ctx, ctx.attr.rust_lib, ctx.attr.allocator_library),
	File "/Users/user/Code/rules_rust/rust/toolchain.bzl", line 126, column 13, in _make_libstd_and_allocator_ccinfo
		fail(dedent("""\
Error in fail: @rust_darwin_x86_64//:toolchain_for_x86_64-apple-darwin_impl --
The `rust_lib` must be a `rust_stdlib_filegroup` target. This rule should be a simple
drop-in replacement for the previous `filegroup` that was expected. If for some reason the
`rust_lib` target is provied from a custom rule, then this rule should return a `RustStdLibInfo`
provider. Both `rust_stdlib_filegroup` and `RustStdLibInfo` can be loaded from
`@rules_rust//rust:toolchain.bzl`. For any additional assistence pleas file an issue on the
GitHub project (https://github.com/bazelbuild/rules_rust) or reach out in the slack channel
(https://bazelbuild.slack.com/archives/CSV56UT0F).
```

rust/toolchain.bzl
```diff
@@ -1,48 +1,48 @@
 """The rust_toolchain rule definition and implementation."""
 
-load(
-    "//rust/private:utils.bzl",
-    "find_cc_toolchain",
+load("//rust/private:utils.bzl", "dedent", "find_cc_toolchain")
+
+RustStdLibInfo = provider(
```
hlopko, line 5: Please move this provider definition into providers.bzl, and expose it through `rust_common` (https://github.com/bazelbuild/rules_rust/blob/2109a0a105baae04148bbe68b9e0f01996ea4e3b/rust/private/common.bzl#L28-L28)

hlopko: I'd name the provider as `StdlibInfo` (providers retain the location where they were declared in their identity, other providers named `StdlibInfo` will not conflict with this one because they will be declared in a different bzl file). If you prefer stricted camel casing I'm fine with that, but please also update the error message below.

rust/toolchain.bzl
```diff
@@ -71,26 +71,91 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
...
+            dot_a_files = dot_a_files,
+            between_alloc_and_core_files = between_alloc_and_core_files,
+            core_files = core_files,
+            between_core_and_std_files = between_core_and_std_files,
+            std_files = std_files,
+            alloc_files = alloc_files,
+        ),
+    ]
+
+rust_stdlib_filegroup = rule(
```
hlopko, line 89: Please expose this rule through `//rust:defs.bzl`. That should be the main entry point for everything users need from rules Rust. The fact that the toolchain is not there yet is sad, but I do plan to move rust_toolchain there soon (definitely before 1.0).

rust/toolchain.bzl
```diff
@@ -71,26 +71,88 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
...
+    Returns:
+        A CcInfo object for the required libraries, or None if no such libraries are available.
+    """
+    cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
+    link_inputs = []
+
+    if not RustStdLibInfo in ctx.attr.rust_lib:
+        fail(dedent("""\
+            {} --
+            The `rust_lib` must be a target providing `rust_common.stdlib_info`
```
hlopko, line 128: I took the liberty of rephrasing the error message. Please take a look and feel free to change further :)

hlopko (commented):
Looking great, only minor administrative changes required.

rust/toolchain.bzl
UebelAndre: How's this look?

PR #803, opened 2021-06-26
Organized examples

UebelAndre:
I think this pattern is much clearer as to what the examples are representing vs the previous `hello_*` and I like the association of the different forms of `sys` crates.

Updates:
`hello_sys` -> `sys/basic`
`complex_sys` -> `sys/complex`
`hello_macro` -> `proc_macro`


UebelAndre:
Perhaps each example should also come with a README?

hlopko (approved):
LGTM.

illicitonion (approved):

PR #805, opened 2021-06-26
`create_universe` no longer assumes the name of rust_repository repos

UebelAndre:
`crate_universe` doesn't work if the user defines their own rust repositories with custom names. The coupling between these repository names is nice to avoid re-downloading assets but is ultimately a bad assumption to make. This change makes `crate_universe` isolated and in the future we can find ways of optimizing this fetch (though, the fetch shouldn't be an issue since we're using Bazel's HTTP downloader which would cache the artifact. The only time spent is in extracting the tar files).


illicitonion:
Is it possible to add arguments which take labels pointing at the right artifacts which can be already-loaded using `rules_rust`'s workspace set-up, rather duplicating it all?

UebelAndre:
> Is it possible to add arguments which take labels pointing at the right artifacts which can be already-loaded using `rules_rust`'s workspace set-up, rather duplicating it all?

Possibly? The thing I wanted to avoid is having `crate_universe` take advantage of internal knowledge of the repository rules or force users to go understand them to be able to write their WORKSPACE files. When you say "which takes labels", do you expect that users would provide maybe a [label_keyed_string_dict](https://docs.bazel.build/versions/main/skylark/lib/attr.html#label_keyed_string_dict) mapping the cargo+rustc binary targets to a triple? Or that they'd provide a repository name for the toolchain repository? Or perhaps something else I'm not interpreting correctly?

illicitonion:
I'm not sure what kind of implementation I'm imagining to be honest... I guess `cargo_universe` could take in a `dict[str, str]` of triple to toolchain name (and default this to `DEFAULT_TOOLCHAIN_TRIPLES`), and we could pass that dict as an additional param to [`get_host_info`](https://github.com/bazelbuild/rules_rust/blob/0768a7f00de134910c3cbdab7bbfdd011d995766/crate_universe/private/util.bzl#L64-L101) to replace the currently hard-coded strings?

UebelAndre:
> I'm not sure what kind of implementation I'm imagining to be honest... I guess `cargo_universe` could take in a `dict[str, str]` of triple to toolchain name (and default this to `DEFAULT_TOOLCHAIN_TRIPLES`), and we could pass that dict as an additional param to [`get_host_info`](https://github.com/bazelbuild/rules_rust/blob/0768a7f00de134910c3cbdab7bbfdd011d995766/crate_universe/private/util.bzl#L64-L101) to replace the currently hard-coded strings?

Would it be acceptable to allow the user to pass a dict of `triple` -> `repository` but have that be an optional parameter where the rule would fallback to using it's own binaries when they're not provided? I feel like knowledge about repository rules is more internal knowledge that the rule shouldn't depend on.

illicitonion:
> > I'm not sure what kind of implementation I'm imagining to be honest... I guess `cargo_universe` could take in a `dict[str, str]` of triple to toolchain name (and default this to `DEFAULT_TOOLCHAIN_TRIPLES`), and we could pass that dict as an additional param to [`get_host_info`](https://github.com/bazelbuild/rules_rust/blob/0768a7f00de134910c3cbdab7bbfdd011d995766/crate_universe/private/util.bzl#L64-L101) to replace the currently hard-coded strings?
> 
> Would it be acceptable to allow the user to pass a dict of `triple` -> `repository` but have that be an optional parameter where the rule would fallback to using it's own binaries when they're not provided?

What does "its own binaries" mean? I'd expect that if you don't specify an explicit value for the arg, it would use whatever `rules_rust` would use if you just called `rust_repositories()` in your `WORKSPACE`...

I guess my answer is: If by "its own binaries" you mean "its own binaries, which happen to be the same ones as `rules_rust` assuming you're not configuring it specially", then yes?

crate_universe/bootstrap.bzl
```diff
@@ -88,11 +89,26 @@ _crate_universe_resolver_bootstrapping = repository_rule(
             allow_single_file = ["Cargo.toml"],
             default = Label("//crate_universe:Cargo.toml"),
         ),
+        "iso_date": attr.string(
+            doc = "The iso_date of cargo binary the resolver should use. Note `version` must be `beta` or `nightly`",
```
illicitonion, line 93:
```suggestion
            doc = "The iso_date of cargo binary the resolver should use. Note: This can only be set if `version` is `beta` or `nightly`",
```

crate_universe/defs.bzl
```diff
@@ -169,6 +168,9 @@ Environment Variables:
             doc = "A template for where to download crates from for the default crate registry. This must contain `{version}` and `{crate}` templates.",
             default = DEFAULT_CRATE_REGISTRY_TEMPLATE,
         ),
+        "iso_date": attr.string(
+            doc = "The iso_date of cargo binary the resolver should use. Note `version` must be `beta` or `nightly`",
```
illicitonion, line 172:
```suggestion
            doc = "The iso_date of cargo binary the resolver should use. Note: This can only be set if `version` is `beta` or `nightly`",
```

crate_universe/private/util.bzl
```diff
@@ -79,23 +86,58 @@ def get_host_info(repository_ctx):
...
-        toolchain_repo = "@rust_windows_{}".format(cpu)
-        extension = ".exe"
     else:
         fail("Could not locate resolver for OS " + repository_ctx.os.name)
 
-    return (resolver_triple, toolchain_repo, extension)
+    return (host_triple, resolver_triple)
+
+def get_cargo_and_rustc(repository_ctx, host_triple):
+    """Download a cargo and rustc binary based on the host triple.
```
illicitonion, line 110:
```suggestion
    """Retrieve a cargo and rustc binary based on the host triple.
```

or locate, maybe... Downloading implies this always causes a download, which I don't think is the case?

examples/crate_universe/has_aliased_deps/Cargo.toml
```diff
@@ -6,3 +6,4 @@ edition = "2018"
 
 [dependencies]
 reqwest = { version = "0.10.7", features = ["blocking", "json"] }
+indexmap = { version = "1.5.2", features = ["std"] }
```
illicitonion, line 9: Why this change?

illicitonion (approved):
Works for me! Thanks

examples/crate_universe/has_aliased_deps/Cargo.toml
UebelAndre: I was running into the following error without it. 
```
error[E0107]: this struct takes 3 type arguments but only 2 type arguments were supplied
  --> external/has_aliased_deps_deps__h2__0_2_7/src/proto/streams/store.rs:15:10
   |
15 |     ids: IndexMap<StreamId, SlabIndex>,
   |          ^^^^^^^^ --------  --------- supplied 2 type arguments
   |          |
   |          expected 3 type arguments
   |
note: struct defined here, with 3 type parameters: `K`, `V`, `S`
help: add missing type argument
   |
15 |     ids: IndexMap<StreamId, SlabIndex, S>,
   |                                      ^^^

For more information about this error, try `rustc --explain E0107`.
```

I came across https://github.com/yewstack/yew/issues/1664 so decided to apply the same.

illicitonion: Can you file an issue for this with repro steps? This shouldn't be necessary

UebelAndre: In attempting to do that, the issue seems to have vanished? Perhaps I had some bad combination of changes before I broke this out into it's own PR. I'll try to find what triggered it again but it now appears to genuinely be unnecessary.

PR #806, opened 2021-06-26
Fix platform mapping for ARM macs downloads

sayrer:


crate_universe/private/util.bzl
```diff
@@ -48,6 +48,9 @@ def _query_cpu_architecture(repository_ctx, expected_archs, is_windows = False):
     else:
         arch = result.stdout.strip("\n")
```
UebelAndre, line 49: Small nit: Could we nest the check in the else statement here?
```suggestion
        arch = result.stdout.strip("\n")
        
        # Correct the arm architecture for macos
        if "mac" in repository_ctx.os.name and arch == "arm64":
            arch = "aarch64"

```

UebelAndre (changes requested):
Thanks! Just one small nit but thanks for fixing this! 😄

UebelAndre (approved):
Nice! Thanks again! 🙏

PR #807, opened 2021-06-26
Fixed inability to use `crate` attribute in wrapper rules with proc-macro targets.

UebelAndre:
It's incorrect to totally disregard the crate type of a target passed to `rust_test::crate` since that info is needed to determine some flags at compile time. There may be more but this was the one I found when trying to flip the `examples/proc_macro` example to use edition `2018` for the default example. I was unable to build with the following issue:
```
error[E0432]: unresolved import `proc_macro`
  --> proc_macro/src/lib.rs:18:5
   |
18 | use proc_macro::TokenStream;
   |     ^^^^^^^^^^ use of undeclared crate or module `proc_macro`

For more information about this error, try `rustc --explain E0432`.
```

By keeping track of the original crate type, we can appropriately include `--extern proc_macro` in the linker flags for these kinds of test targets. I was able to confirm that Cargo is setting this extern flag for the same test target by copying the code from the example into a cargo workspace and adding the following diff
```diff
diff --git a/examples/proc_macro/src/lib.rs b/examples/proc_macro/src/lib.rs
index 87b9d47..08d102a 100644
--- a/examples/proc_macro/src/lib.rs
+++ b/examples/proc_macro/src/lib.rs
@@ -23,3 +23,8 @@ use proc_macro::TokenStream;
 pub fn hello_world(_input: TokenStream) -> TokenStream {
     TokenStream::new()
 }
+
+#[test]
+fn test_test() {
+    println!("test");
+}
```

The workspace structure was:
```
├── Cargo.lock
├── Cargo.toml
├── src
│   └── lib.rs
└── tests
    └── greeting.rs

2 directories, 4 files
```

and the results of `cargo test --verbose` were:
```
   Compiling proc_macro_lib_2018 v0.1.0 (/Users/user/Code/test/proc_macro/2018)
     Running `rustc --crate-name proc_macro_lib_2018 --edition=2018 2018/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type proc-macro --emit=dep-info,link -C prefer-dynamic -C embed-bitcode=no -C split-debuginfo=unpacked -C debuginfo=2 -C metadata=ff213938a7a3dda6 -C extra-filename=-ff213938a7a3dda6 --out-dir /Users/user/Code/test/proc_macro/target/debug/deps -C incremental=/Users/user/Code/test/proc_macro/target/debug/incremental -L dependency=/Users/user/Code/test/proc_macro/target/debug/deps --extern proc_macro`
     Running `rustc --crate-name proc_macro_lib_2018 --edition=2018 2018/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --emit=dep-info,link -C prefer-dynamic -C embed-bitcode=no -C split-debuginfo=unpacked -C debuginfo=2 --test -C metadata=1f84dca61c8a5688 -C extra-filename=-1f84dca61c8a5688 --out-dir /Users/user/Code/test/proc_macro/target/debug/deps -C incremental=/Users/user/Code/test/proc_macro/target/debug/incremental -L dependency=/Users/user/Code/test/proc_macro/target/debug/deps --extern proc_macro`
     Running `rustc --crate-name greeting --edition=2018 2018/tests/greeting.rs --error-format=json --json=diagnostic-rendered-ansi --emit=dep-info,link -C embed-bitcode=no -C split-debuginfo=unpacked -C debuginfo=2 --test -C metadata=4b3cde5d33b9a879 -C extra-filename=-4b3cde5d33b9a879 --out-dir /Users/user/Code/test/proc_macro/target/debug/deps -C incremental=/Users/user/Code/test/proc_macro/target/debug/incremental -L dependency=/Users/user/Code/test/proc_macro/target/debug/deps --extern proc_macro_lib_2018=/Users/user/Code/test/proc_macro/target/debug/deps/libproc_macro_lib_2018-ff213938a7a3dda6.dylib`
    Finished test [unoptimized + debuginfo] target(s) in 0.78s
     Running `/Users/user/Code/test/proc_macro/target/debug/deps/proc_macro_lib_2018-1f84dca61c8a5688`

running 1 test
test test_hi ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `/Users/user/Code/test/proc_macro/target/debug/deps/greeting-4b3cde5d33b9a879`

running 1 test
test test_hello_world_macro ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests proc_macro_lib_2018
     Running `rustdoc --edition=2018 --crate-type proc-macro --crate-name proc_macro_lib_2018 --test /Users/user/Code/test/proc_macro/2018/src/lib.rs -L dependency=/Users/user/Code/test/proc_macro/target/debug/deps -L dependency=/Users/user/Code/test/proc_macro/target/debug/deps --extern proc_macro_lib_2018=/Users/user/Code/test/proc_macro/target/debug/deps/libproc_macro_lib_2018-ff213938a7a3dda6.dylib --extern proc_macro -C embed-bitcode=no --error-format human`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Requires: 
- https://github.com/bazelbuild/rules_rust/pull/803


hlopko:
If it's fine with you, I'll not review this PR until Organized examples #803 is merged, the diff is huge right now. Could you please ping me here once it happens?

illicitonion (approved):
LGTM!

rust/private/rustdoc.bzl
```diff
@@ -85,6 +85,9 @@ def _rust_doc_impl(ctx):
     args.add(crate.root.path)
     args.add("--crate-name", crate.name)
     args.add("--crate-type", crate.type)
+    if crate.type == "proc-macro":
```
hlopko, line 88: Changes in this and the rustdoc_test files are not really relevant to the rest of the PR. I'd prefer splitting these changes into a separate PR with a focused description, and a unit test similar to https://github.com/bazelbuild/rules_rust/blob/main/test/unit/use_libtest_harness/use_libtest_harness_test.bzl. Wdyt?

rust/private/providers.bzl
```diff
@@ -27,6 +27,10 @@ CrateInfo = provider(
         "root": "File: The source File entrypoint to this crate, eg. lib.rs",
         "rustc_env": "Dict[String, String]: Additional `\"key\": \"value\"` environment variables to set for rustc.",
         "srcs": "depset[File]: All source Files that are part of the crate.",
+        "transitive_type": (
+            "str, optional: The original crate type for targets generated using a previously defined " +
```
hlopko, line 31: It seems this is the only optional field in this provider, and you seem to be adding it everywhere explicitly. Are you trying to make the code backwards compatible?

rust/private/providers.bzl
```diff
@@ -27,6 +27,10 @@ CrateInfo = provider(
         "root": "File: The source File entrypoint to this crate, eg. lib.rs",
         "rustc_env": "Dict[String, String]: Additional `\"key\": \"value\"` environment variables to set for rustc.",
         "srcs": "depset[File]: All source Files that are part of the crate.",
+        "transitive_type": (
```
hlopko, line 30: WDYT about naming it "wrapped_crate_type"? If I remember correctly that is how you and Daniel called these rules, "wrapper rules". Or am I misremembering?

hlopko: I'd shy away from "transitive" word though, that usually means everything all the way to leaves, and here we mean something else.

hlopko (commented):
Thanks for working on this! Now that I'm thinking about it I might have hit this bug in the past, I just didn't know it was in the rules.

rust/private/providers.bzl
UebelAndre: I've never introduced an optional parameter on a provider before so I had assumed I needed to plug it in everywhere. But Assuming I don't set it, does it default to `None`?

rust/private/providers.bzl
UebelAndre: I've definitely used the term to describe similar concepts if not this, I can rename

rust/private/rustdoc.bzl
UebelAndre: It's the same core bug though. I don't think a separate PR should be needed for fixing the same class of problem in different rules. If the tech-debt were solved here to have the rules use the same functionality as the core rules for generating flags, there would have been no change to these rules. I think think it's appropriate to keep these changes in this PR and bump the priority on addressing tech debt. Thoughts?

UebelAndre: Also, the functionality here is covered by: 
https://github.com/bazelbuild/rules_rust/blob/d31b77510173f65356a1b56452d7f2ed1d1a68de/examples/proc_macro/BUILD.bazel#L39-L47
so it's not entirely unchecked.

rust/private/providers.bzl
UebelAndre: By leaving this blank, I run into
```
Error: 'CrateInfo' value has no field or method 'wrapped_crate_type'
Available attributes: aliases, deps, edition, is_test, name, output, proc_macro_deps, root, rustc_env, srcs, type
```

It seems the field needs to be explicitly created.

rust/private/rustdoc.bzl
UebelAndre: Though I suppose I only mention `rust_test` in the description....

rust/private/providers.bzl
hlopko: That is very unfortunate. This makes this a backwards incompatible change (and the field is not optional so please fix the documentation).

This could have been prevented if we had `rust_common.create_crate_info` function. I propose following:

1. let's create `rust_common.create_crate_info` function. I'm happy to do it if you want.
2. let's update the docs of `rust_common.crate_info` and say this is not supposed to be used for construction, only for accessing the provider (unfortunately provider constructor is used as both constructor and type declaration that we use in rule definitions or when getting providers from target).
2. let's call this function in all the places in this repo where we currently call `rust_common.crate_info(...)`.
3. let's recommend people migrating for this PR to use `create_crate_info` instead of `crate_info`.

rust/private/rustdoc.bzl
hlopko: Yes you are right. I'd still write an unit test, but you can carry on :)

rust/private/rustc.bzl
```diff
@@ -459,7 +459,8 @@ def construct_arguments(
     # These always need to be added, even if not linking this crate.
     add_crate_link_flags(args, dep_info)
 
-    if crate_info.type == "proc-macro" and crate_info.edition != "2015":
+    is_type_proc_macro = crate_info.type == "proc-macro" or crate_info.wrapped_crate_type == "proc-macro"
```
hlopko, line 462: Nit: rename `-is_type_proc_macro` to `needs_extern_proc_macro_flag. Please move edition check there too, so the if below is only

```
if needs_extern_proc_macro_flag:
```

test/unit/proc_macro/proc_macro_2015.rs
```diff
@@ -0,0 +1,10 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+
+/// This macro is a no-op; it is exceedingly simple as a result
```
hlopko, line 5: Suggestion: I think the comment is not necessary. If you want to make the code more intention revealing, name the macro as Noop?

test/unit/proc_macro/proc_macro_2015.rs
```diff
@@ -0,0 +1,10 @@
...
+/// of avoiding dependencies on both the syn and quote crates.
+#[proc_macro_derive(HelloWorld)]
+pub fn hello_world(_input: TokenStream) -> TokenStream {
+    TokenStream::new()
```
hlopko, line 9: Suggestion: you can `return input;` so the parameter is used. It would then also fit the Noop name better. Up to you.

test/unit/proc_macro/proc_macro_2015.rs
hlopko: Please apply the same change below too.

test/unit/proc_macro/proc_macro_test.bzl
```diff
@@ -0,0 +1,100 @@
...
+        edition = edition,
+    )
+
+    rust_test(
+        name = "proc_macro_lib_{}_unittest".format(edition),
+        crate = ":proc_macro_lib_{}".format(edition),
+        edition = edition,
+    )
+
+    rust_test(
```
hlopko, line 27: Please move the integration test to `//test/proc_macro`. Let's only keep unit tests in //test/unit/...

rust/private/providers.bzl
UebelAndre: Was https://github.com/bazelbuild/rules_rust/pull/818 what you were thinking here?

test/unit/proc_macro/proc_macro_test.bzl
```diff
@@ -0,0 +1,100 @@
...
+    rust_proc_macro(
+        name = "proc_macro_lib_{}".format(edition),
+        srcs = [
+            "proc_macro_{}.rs".format(edition),
+        ],
+        edition = edition,
+    )
+
+    rust_test(
+        name = "proc_macro_lib_{}_unittest".format(edition),
```
hlopko, line 22: Nit: rename unittest -> `wrapper_rule_for_macro_{}`?

test/unit/proc_macro/proc_macro_test.bzl
```diff
@@ -0,0 +1,100 @@
...
+load("//test/unit:common.bzl", "assert_action_mnemonic", "assert_list_contains_adjacent_elements", "assert_list_contains_adjacent_elements_not")
+
+def _proc_macro_test_targets(edition):
+    """Define a set of `rust_proc_macro` targets for testing
+
+    Args:
+        edition (str): The rust edition to use for the new targets
+    """
+    rust_proc_macro(
+        name = "proc_macro_lib_{}".format(edition),
```
hlopko, line 14: Nit: I'd drop the `_lib` from the name

test/unit/proc_macro/proc_macro_test.bzl
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    rust_test(
+        name = "proc_macro_lib_{}_integration_test".format(edition),
+        srcs = ["proc_macro_{}_test.rs".format(edition)],
+        edition = edition,
+        proc_macro_deps = [":proc_macro_lib_{}".format(edition)],
+    )
+
+def _proc_macro_2015_no_extern_flag_impl(ctx):
```
hlopko, line 34: I think we can be more explicit here. Please extract a function with the body of the unittests (besides asserting presence of flags they are identical).

Then I'd write 4 unittests:

extern_flag_not_passed_when_compiling_macro_2015
extern_flag_passed_when_compiling_macro_2018
extern_flag_not_passed_when_compiling_macro_wrapper_rule_2015
extern_flag_passed_when_compiling_macro_wrapper_rule_2015

hlopko (commented):
Looking good, only minor comments and should be good to go.

test/unit/proc_macro/proc_macro_test.bzl
UebelAndre: Why aren't they all just considered tests and go in the root of test? It feels like we're gonna exponentially gain nearly identical rust files if we need to split everything like this.

test/unit/proc_macro/proc_macro_2015.rs
UebelAndre: This was lifted from [the existing examples](https://github.com/bazelbuild/rules_rust/tree/19c74ce6ad55ef07bfeb36af1889c46da36d5a07/examples/proc_macro) should those be updated too?

UebelAndre: I've refactored the code to use code [straight from the book](https://doc.rust-lang.org/reference/procedural-macros.html#function-like-procedural-macros)

test/unit/proc_macro/proc_macro_2015.rs
UebelAndre: [same action](https://github.com/bazelbuild/rules_rust/pull/807#discussion_r662070938)

test/unit/proc_macro/proc_macro_test.bzl
UebelAndre: done

rust/private/rustc.bzl
UebelAndre: done

test/unit/proc_macro/proc_macro_test.bzl
UebelAndre: I think I preferred it before but not strongly enough to not make the change 😅

test/unit/proc_macro/proc_macro_test.bzl
hlopko: We should have order of magnitude more unit tests than integration tests. If they are duplicated, delete the integration test.

rust/private/rustc.bzl
```diff
@@ -459,7 +459,8 @@ def construct_arguments(
     # These always need to be added, even if not linking this crate.
     add_crate_link_flags(args, dep_info)
 
-    if crate_info.type == "proc-macro" and crate_info.edition != "2015":
+    needs_extern_proc_macro_flag = (crate_info.type == "proc-macro" or crate_info.wrapped_crate_type == "proc-macro") and crate_info.edition != "2015"
```
hlopko, line 462: SuperNit: make multiline for readability.

hlopko (approved):

test/unit/proc_macro/proc_macro_test.bzl
UebelAndre: I think the thing that bothers me is that targets and unit tests are declare together. I'd like to think of unit tests as assertions on existing targets but that I could organize it such that I have a bunch of targets in `test/*` and then potentially *only* define unit test targets (which wouldn't be `rust_*` targets) in `test/unit`. But maybe that doesn't make sense (I struggle greatly with starlark unittests and really dislike them 😅)

UebelAndre: Updated. But I think the coupling (decoupling) here will become unmaintainable if we continue with this pattern. I see no reason why there's a need for the `test/unit` directory. I would say each `test/*` directory defines integration tests as the base targets that are used for the relevant unit tests in the same package or a `unit` subpackage.

hlopko: Oh I'd like to go the other way. I'm bothered that the test setup and assertions are too far away from each other (and test setup is below assertions in the file, that is also bad). Ideally I'd like to write a test like:


```
def test_that_yolo_is_foo(ctx):
  ctx.package("//foo", """
    rust_library(name = "asdf", srcs = ["..."])
    rust_test(name="asdf_test", crate=":asdf")
  """
  argv = ctx.target("//foo:asdf_test").actions[0].argv
  assert_list_contains_adjacent_elements("--extern", "yolo")
```

And even more ideally this test would not execute any actions.

rust/private/rustc.bzl
UebelAndre: Wish buildifier would do that for me 😄 

Updated

PR #809, opened 2021-06-27
Expand arg locations

### LINKED ISSUE #801: What's the best way to vary crate_features?
Suppose I want to change `crate_features` arguments on the `bazel` command line. What's the best way?

sayrer:
This fixed #801 for me.


rust/private/utils.bzl
```diff
@@ -179,6 +179,23 @@ def expand_locations(ctx, env, data):
     """
     return dict([(k, _expand_location(ctx, v, data)) for (k, v) in env.items()])
 
+def expand_arg_locations(ctx, args, data):
```
UebelAndre, line 182: I don't know if this over-complicates the implementation but I think it'd be nicer to add an optional `args` filed to `expand_location`? Maybe make `env` optional as well but assert that one *or* the other is set? Could be talked out of this pattern though.

hlopko:
Just a quick drive by - could you add a test?

rust/private/rustc.bzl
```diff
@@ -440,7 +441,14 @@ def construct_arguments(
     args.add_all(rust_lib_paths, before_each = "-L", format_each = "%s")
 
     args.add_all(rust_flags)
-    args.add_all(getattr(attr, "rustc_flags", []))
+    args.add_all(
+        expand_arg_locations(
+            ctx,
+            getattr(attr, "rustc_flags", []),
+            getattr(attr, "data", []) +
+            getattr(attr, "compile_data", []),
```
hlopko, line 449: Nit: Could you put these 2 lines (data and compile_data) on the same line so future me realizes they are added to create a third argument to the expand_arg_locations (if buildifier doesn't allow that, please extract a local variable)? I just starred at this for few minutes not understanding what's going on because I missed that there is a comma at the end of the line 447, and there is a plus on 448 :) Thank you!

illicitonion:
> Just a quick drive by - could you add a test?

An easy way to do this could be to generate in a genrule a file which sets a `--cfg` flag, and then pass `@$(location :generated-file)` as a `rustc_flags` value, and then assert in a main method that that cfg is set. Possibly this could conveniently live in https://github.com/bazelbuild/rules_rust/tree/main/examples/env_locations :)

sayrer:
> > Just a quick drive by - could you add a test?
> 
> An easy way to do this could be to generate in a genrule a file which sets a `--cfg` flag, and then pass `@$(location :generated-file)` as a `rustc_flags` value, and then assert in a main method that that cfg is set. Possibly this could conveniently live in https://github.com/bazelbuild/rules_rust/tree/main/examples/env_locations :)

Oh... maybe you meant I should make a separate test file. Happy to do so if it matters.

rust/private/utils.bzl
sayrer: That seems more complicated than what I have now, and would probably require more documentation too. I'd prefer to leave it as is, unless others want to refactor it.

UebelAndre: If that's the case then what do you think about renaming `expand_location` to `expand_env_locations`? For some reason the `expand_location` + `expand_arg_locations` really irks me 😅

sayrer: Done.

UebelAndre (approved):
> Oh... maybe you meant I should make a separate test file. Happy to do so if it matters.

I personally think it's better to add a small and focused use use of this functionality to the `test` module but I'll defer to @hlopko on this one.

Otherwise looks good to me! 😄

illicitonion (approved):
LGTM - thanks!

rust/private/rustc.bzl
```diff
@@ -438,9 +439,17 @@ def construct_arguments(
 
     # Tell Rustc where to find the standard library
     args.add_all(rust_lib_paths, before_each = "-L", format_each = "%s")
-
     args.add_all(rust_flags)
-    args.add_all(getattr(attr, "rustc_flags", []))
+
+    # Data file paths for location expansion
```
hlopko, line 444: Nit: Unless you feel strongly about it I'd remove the comment, IMHO it doesn't add much.

rust/private/utils.bzl
```diff
@@ -152,7 +152,7 @@ def _expand_location(ctx, env, data):
             env = env.replace(directive, "${pwd}/" + directive)
     return ctx.expand_location(env, data)
 
-def expand_locations(ctx, env, data):
+def expand_env_locations(ctx, env, data):
```
hlopko, line 155: Kind ask: now that you touched it, could you improve the documentation to have Args and Returns, and so there's a symmetry between this function and `expand_arg_locations`? Could you also add that this function is meant to be used to expand `$(location //foo)` macros in dict values, and that the `expand_arg_location` is for list elements?

Now that I'm thinking, what do you think about renaming these functions from `env` to `dict_values` and `arg` to `list_elements`?

rust/private/utils.bzl
```diff
@@ -179,6 +179,23 @@ def expand_locations(ctx, env, data):
     """
     return dict([(k, _expand_location(ctx, v, data)) for (k, v) in env.items()])
 
+def expand_arg_locations(ctx, args, data):
+    """Performs location-macro expansion on a list of string values.
+
+    See `expand_locations` for detailed documentation.
```
hlopko, line 185: Please mention `$(execroot ...) and $(location ...) are prefixed with ${pwd},` there (to keep symmetry with `expand_env_locations`.

examples/flag_locations/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load(
+    "@rules_rust//rust:rust.bzl",
+    "rust_test",
+)
+
+# generate a file containing cfg flags
+genrule(
```
hlopko, line 7: Let me start with saying I know it's a lot of boilerplate to write, but I personally would write a unit test under `//test/unit` (https://docs.bazel.build/versions/main/skylark/testing.html#for-testing-rules). This is an existing unit test that verifies args: https://github.com/bazelbuild/rules_rust/blob/main/test/unit/use_libtest_harness/use_libtest_harness_test.bzl.

hlopko: Extra points for adding tests for that logic adding `pwd` prefix, but I don't insist. I think you'll be suffering way too much already.

hlopko (commented):
Thank you so so much for a solid PR. Could you update docs of rule attributes that are affected by this PR to mention that macros are expanded?

rust/private/utils.bzl
sayrer: I think both of these functions have docs for Args and Returns, if I understand correctly.

I did the rename.

sayrer: > Could you update docs of rule attributes that are affected by this PR to mention that macros are expanded?

This part is yet to be done.

UebelAndre: > Now that I'm thinking, what do you think about renaming these functions from `env` to `dict_values` and `arg` to `list_elements`?

With such verbose names, I wonder if it's better to have kept `expand_locations` and checked for whether a list or dict was passed, but that's just a small gripe 😅

sayrer: Now documented (please unresolve if I misunderstood the Args and Returns situation).

rust/private/rust.bzl
```diff
@@ -620,7 +620,14 @@ _common_attrs = {
         """),
     ),
     "rustc_flags": attr.string_list(
-        doc = "List of compiler flags passed to `rustc`.",
+        doc = dedent("""\
```
UebelAndre, line 623: hmmm, does this render cleanly? Attribute `doc` fields are typically one continuous line of text and expected to be short...

This is typically why you see the less elegant
```python
doc = (
    "List of compiler flags passed to `rustc`. " + 
    "These strings are subject to Make variable expansion for predefined " +
    "source/output path variables like `$location`, `$execpath`, and `$rootpath`. " +
    "This expansion is useful if you wish to pass a generated file of " +
    "arguments to rustc: `@$(location //:package:target)`."
),
```

sayrer: I just copied the format of all of the other ones (?)

UebelAndre: it looks like it renders just fine 😄. Disregard!

examples/flag_locations/BUILD.bazel
sayrer: This actually wouldn't pass without adding `${pwd}` to the expected argument string. Confusingly,  `assert_argv_contains` calls:

`"Expected {args} to contain {flag}".format(args = action.argv, flag = flag)`

on error. I think this expands `${pwd}` to the empty string under test, so it took me a while to figure out that I needed `@${pwd}/bazel-out/...` instead of `@/bazel-out/...`

sayrer: I'm going to leave the example as well. They both cover the same code, but I find the analysis test difficult to read, and I know more about Bazel than most people on the planet :)

hlopko: Sounds good!

examples/flag_locations/main.rs
```diff
@@ -0,0 +1,9 @@
+#[test]
+fn test() {
+    // we should able to read rustc args from a generated file
```
hlopko, line 3: Nit: we should *be* able

rust/private/rust.bzl
```diff
@@ -620,7 +620,14 @@ _common_attrs = {
...
+            List of compiler flags passed to `rustc`.
+
+            These strings are subject to Make variable expansion for predefined
+            source/output path variables like `$location`, `$execpath`, and `$rootpath`.
+            This expansion is useful if you wish to pass a generated file of
+            arguments to rustc: `@$(location //:package:target)`.
```
hlopko, line 629: Please replace `//:package:target` with `//package:target`

rust/private/utils.bzl
```diff
@@ -179,6 +179,27 @@ def expand_locations(ctx, env, data):
...
 
+def expand_list_element_locations(ctx, args, data):
+    """Performs location-macro expansion on a list of string values.
+
+    $(execroot ...) and $(location ...) are prefixed with ${pwd},
+    which process_wrapper and build_script_runner will expand at run time
+    to the absolute path.
+
+    See `expand_locations` for detailed documentation.
+
```
hlopko, line 190: Nit: replace `expand_locations` with

```
[`expand_locations`](https://docs.bazel.build/versions/main/skylark/lib/ctx.html#expand_location)
```

Also above please.

test/unit/location_expansion/location_expansion_test.bzl
```diff
@@ -0,0 +1,54 @@
...
+        name = "flag_generator",
+        outs = ["generated_flag.data"],
+        cmd = "echo --cfg=test_flag > $@",
+    )
+
+    rust_library(
+        name = "mylibrary",
+        srcs = ["mylibrary.rs"],
+        rustc_flags = [
+            "@$(location :flag_generator)",
```
hlopko, line 31: Nit: you could marginally simplify the test by not passing a file with cfg flags, but by passing a cfg flag with value (https://doc.rust-lang.org/rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment).

Something like (hope this works, haven't tried):

```
rustc_flags = [ "--cfg=", "yololo=$(location :foo)"],
```

test/unit/location_expansion/mylibrary.rs
```diff
@@ -0,0 +1,3 @@
+pub fn hello_world() {
```
hlopko, line 1: Nit: this file can be empty.

test/unit/location_expansion/location_expansion_test.bzl
hlopko: I leave the decision what looks simpler to you.

sayrer: This won't work because the analysis fails if the target isn't a dependency of the rule (in compile_data here).

rust/private/utils.bzl
UebelAndre:
```suggestion
See `expand_locations` for detailed documentation.
https://docs.bazel.build/versions/main/skylark/lib/ctx.html#expand_location
```

This isn't a rendered function so I think it'd be nicer to have the doc something that's easy to highlight, copy, and paste. But up to you two 😅

test/unit/location_expansion/mylibrary.rs
sayrer: Well, it can consist of a single newline character, it turns out...

sayrer:
It looks to me like the only issues remaining here might be small documentation / test changes, although I think I've resolved all of the comments.

I don't feel strongly about any of them, so I'd suggest merging this and then changing any of those things in a follow-up, rather than leaving comments here. That will make it easier to work on a California vs. Munich schedule.

hlopko (approved):
Thank you so much! Looks great to me!

PR #810, opened 2021-06-29
Update README community section

UebelAndre:
This is to communicate the transition from Google Groups to GitHub Discussions.


UebelAndre:
@hlopko Should this actually link to discussions? Feels kinda unnecessary?

hlopko:
A link is not necessary, but I'd say we use github discussions for discussions. I'd also keep the link to the (now archived) google group with some note that this is no longer used, and is mentioned only as an archive of past discussions.

UebelAndre:
> A link is not necessary, but I'd say we use github discussions for discussions. I'd also keep the link to the (now archived) google group with some note that this is no longer used, and is mentioned only as an archive of past discussions.

Is this better?

hlopko (approved):

PR #813, opened 2021-06-29
Add missing android extensions

UebelAndre:
Specify extensions for android system.

This is a drive-by fix. We don't currently have any android tests here and I didn't test it manually, but I think it's correct.

(original conversation: https://github.com/bazelbuild/rules_rust/pull/797#discussion_r660884773)

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -42,6 +42,7 @@ _SYSTEM_TO_BUILTIN_SYS_SUFFIX = {
 }
 
 _SYSTEM_TO_BINARY_EXT = {
+    "android": "",
     "darwin": "",
     "emscripten": ".js",
     "freebsd": "",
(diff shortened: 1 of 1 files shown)


hlopko (approved):
Could you add more info into the PR description? Things like is this a bug fix or a new behavior, how did you find it, why do you fix it, how did you test it. 

I'll also say that in my world 

```
Specify extensions for android system.

This is a drive-by fix. We don't currently have any android tests here and I didn't test it manually, but I think it's correct.
```

is an acceptable PR description :) It's better to be explicit.

UebelAndre:
> Could you add more info into the PR description? Things like is this a bug fix or a new behavior, how did you find it, why do you fix it, how did you test it.
> 
> I'll also say that in my world
> 
> ```
> Specify extensions for android system.
> 
> This is a drive-by fix. We don't currently have any android tests here and I didn't test it manually, but I think it's correct.
> ```
> 
> is an acceptable PR description :) It's better to be explicit.

Definitely, just trying to move a million miles an hour to get everything before bed time 😅

Issue #567, opened 2021-02-01
compile_data does not propagate to rust_test targets


djmarcin:
Given a `rust_library` target with `compile_data`, the corresponding `rust_test` will fail to compile unless the `compile_data` attribute is also included on it. It seems like this is extraneous and the `rust_library` `compile_data` should automatically be included.

illicitonion:
Yeah, that seems like a bug - it feels like `compile_data` should be propagated to "wrapper" targets (but not to other dependees, or transitively)

UebelAndre:
What would this fix look like? Should the [CrateInfo](https://github.com/bazelbuild/rules_rust/blob/3974550a12db8b49f4cd2e083337c55b2871779d/rust/private/providers.bzl#L17-L32) provider gain a new field `compile_data`?

cc @hlopko who might also have good insight on the impact there.

hlopko:
just a side question, @illicitonion are there other "wrapper" targets than a `rust_test` using `crate` attr?

I agree that updating `CrateInfo` is the right solution here. Because of the design of Rust unit testing we can't avoid exposing state from all `rust_library` attributes to `rust_test` through `CrateInfo`. This issue is not surprising me, there's a bit of missing functionality there, for example I found https://github.com/bazelbuild/rules_rust/pull/751 the other day.

hlopko:
> just a side question, @illicitonion are there other "wrapper" targets than a rust_test using crate attr?

To partially answer my own question, we also have `rust_doc` as a wrapper rule and the third discovered issue at https://github.com/bazelbuild/rules_rust/issues/689.

illicitonion:
> just a side question, @illicitonion are there other "wrapper" targets than a `rust_test` using `crate` attr?

I don't know of any other than `rust_test` and `rust_doc`, but I'm not confident in saying there aren't others!

PR #814, opened 2021-06-30
The `compile_data` attribute can now be gathered from dependencies

UebelAndre:
The `compile_data` attribute seems like it should propagate all the way through to the final build outputs since it'd be required by wrapper rules (`rust_test`) to be able to successfully compile.

closes https://github.com/bazelbuild/rules_rust/issues/567


UebelAndre:
Not sure if this is an expensive solution but I think it's fine?

hlopko:
Do we really need to propagate `compile_data` all the way to the binary? Do you have a real world example? AFAIU (and it seems docs https://bazelbuild.github.io/rules_rust/defs.html#rust_library-compile_data suggest the same) compile data are only meant to be used to build the library, in other words they are used by rustc when producing the rlib.

Adding `wrapped_crate_compile_data` into `CrateInfo` that allows `rust_test` and other wrapper rules to recreate the original rustc invocation makes total sense and it is a good improvement.

UebelAndre:
> Do we really need to propagate `compile_data` all the way to the binary? Do you have a real world example? AFAIU (and it seems docs https://bazelbuild.github.io/rules_rust/defs.html#rust_library-compile_data suggest the same) compile data are only meant to be used to build the library, in other words they are used by rustc when producing the rlib.

It's also used in [binaries](https://bazelbuild.github.io/rules_rust/defs.html#rust_binary-compile_data) but maybe I misread this as a suggestion that it's only used by libraries.

> Adding `wrapped_crate_compile_data` into `CrateInfo` that allows `rust_test` and other wrapper rules to recreate the original rustc invocation makes total sense and it is a good improvement.

The issue is that at the time it's the responsibility of the crate that's given `compile_data` to forward it. Once we're in a wrapper rule's implementation, we no longer have access to that attribute if it wasn't put in a provider somewhere. It could definitely be the case that we don't want compile data to propagate always and to actually restrict it to wrapper rules, but I wasn't sure.

How would I populate `wrapped_crate_compile_data` only for wrapped targets?

UebelAndre:
Additionally, I do think it might be valuable to be able to transitively provide data without having it end up in the runfiles for runtime.

hlopko:
> It's also used in binaries but maybe I misread this as a suggestion that it's only used by libraries.

Yup it is used in binaries too for the same reason it is in libraries - to make a dependency available during the rustc invocation, not to be propagated.

> The issue is that at the time it's the responsibility of the crate that's given compile_data to forward it. Once we're in a wrapper rule's implementation, we no longer have access to that attribute if it wasn't put in a provider somewhere. It could definitely be the case that we don't want compile data to propagate always and to actually restrict it to wrapper rules, but I wasn't sure.

Oh right that is true, we have to put compile data into the provider in case there is a wrapper rule. But I would still only put the compile data of the current target, not transitive ones.

> Additionally, I do think it might be valuable to be able to transitively provide data without having it end up in the runfiles for runtime.

I'd be interested to learn about an actual, real world use case for this. And even then, I think it would make sense to create a different attribute for transitive propagation. So target author can decide which behavior matches their use case. Propagating by default can end up being too memory costly in large workspaces.

UebelAndre:
> I'd be interested to learn about an actual, real world use case for this. And even then, I think it would make sense to create a different attribute for transitive propagation. So target author can decide which behavior matches their use case. Propagating by default can end up being too memory costly in large workspaces.

I don't think I have a reasonable real-world use case. I've removed the use of transitive compile-data. I think this covers the practical use case now.

rust/private/rust.bzl
```diff
@@ -266,6 +266,7 @@ def _rust_library_common(ctx, crate_type):
             edition = get_edition(ctx.attr, toolchain),
             rustc_env = ctx.attr.rustc_env,
             is_test = False,
+            compile_data = depset(ctx.files.compile_data),
```
hlopko, line 269: Could I kindly nerd snipe you into creating a starlark unit test that in general would inspect providers returned by a `rust_library`, but you'll only have to add a test that `rust_library` using `compile_data` provides `CrateInfo` with that `compile_data`. I think having more coverage for data in providers is valuable, and It'll be great to have this test started, so it is simpler for other contributors to add another test there when needed.

hlopko (commented):
LGTM, thank you! Only a kind unit test nerd snipe attempt remaining :)

test/compile_data/compile_data.rs
```diff
@@ -0,0 +1,9 @@
+/// Data loaded from compile data
+pub const COMPILE_DATA: &str = include_str!("compile_data.txt");
+
+/// A test that is expected to be compiled from a target that does not
+/// directly populate the `compile_data` attribute
+#[test]
```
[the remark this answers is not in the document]
UebelAndre: I didn't think it was just a pattern. In most cases, you'll have test dependencies and extra functions/structs/etc used for tests that you don't want in your binary otherwise. My understanding is the solution is to have a test-only module that contains all this so it can easily be removed. Since there's only one function here, I didn't think it was necessary since I don't think this would compile into a non-test target. I could be wrong but also have no problem adding the module

rust/private/rust.bzl
UebelAndre: Added a test

hlopko (approved):
Thank you so much! Looks great.

PR #817, opened 2021-06-30
Remove ubuntu1604 from presubmit.yml

philwo:
Ubuntu 16.04 is end-of-life, we're going to remove it from Bazel CI.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -8,11 +8,6 @@ default_linux_targets: &default_linux_targets
   # Bindgen currently only has a working toolchain for 18.04
   - "-@examples//ffi/rust_calling_c/simple/..."
 tasks:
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Sounds good to me 😄

UebelAndre:
@philwo what is the [rules-rust-pus-bazelisk-migrate](https://buildkite.com/bazel/rules-rust-plus-bazelisk-migrate/builds/10) pipelie?

hlopko:
> rules-rust-pus-bazelisk-migrate pipeline?

It's a pipeline that was not working until yesterday (I fixed it) that is testing rules rust against upcoming incompatible changes in Bazel. We're discussing with @philwo if this pipeline is still useful or if there are other, better ways to keep up with changes in Bazel.

PR #818, opened 2021-07-01
Added `rust_common.create_crate_info`

UebelAndre:
The `rust_common.create_crate_info` macro creates a `CrateInfo` provider through a macro which allows for the rules to maintain better backwards compatibility when making changes to `CrateInfo`.

### DIFF
--- a/rust/private/common.bzl
+++ b/rust/private/common.bzl
@@ -25,7 +25,11 @@ In the Bazel lingo, `rust_common` gives the access to the Rust Sandwich API.
 
 load(":providers.bzl", "CrateInfo", "DepInfo")
 
+def _create_crate_info(**kwargs):
+    return CrateInfo(**kwargs)
+
 rust_common = struct(
+    create_crate_info = _create_crate_info,
     crate_info = CrateInfo,
     dep_info = DepInfo,
 )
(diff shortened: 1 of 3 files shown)


rust/private/common.bzl
```diff
@@ -25,7 +25,11 @@ In the Bazel lingo, `rust_common` gives the access to the Rust Sandwich API.
 
 load(":providers.bzl", "CrateInfo", "DepInfo")
 
+def _create_crate_info(**kwargs):
+    return CrateInfo(**kwargs)
+
 rust_common = struct(
+    create_crate_info = _create_crate_info,
```
hlopko, line 32: I know it's not exported to https://bazelbuild.github.io/rules_rust/, but could you also add a doc saying that create_crate_info should be used for construction, and crate_info for accessing?

hlopko (approved):

rust/private/common.bzl
UebelAndre: Like, a docstring? (like the one I added?)

hlopko: Oh right, yeah that works. Nit: rename "macro" to "function" in the docstring

UebelAndre: Is the terminology not "macro"? But updated.

PR #820, opened 2021-07-02
Use the libraries included with the clang distribution for bindgen

djmarcin:
Prior to this PR, bindgen is broken on macOS 11.4 because it cannot find the file `/usr/lib/libstdc++.6.dylib` (it's a broken symlink).

As far as I can tell, what this code is trying to do is invoke `clang` with `LD_LIBRARY_PATH` so that it reads the libraries specified in the toolchain. On macOS, this variable name is actually `DYLD_LIBRARY_PATH` so this wasn't really working at all before on macOS -- probably the system libraries were being used instead. I'm not sure if you can detect the OS from inside a rule, so I'm setting both LD_LIBRARY_PATH and DYLD_LIBRARY_PATH because I don't think it hurts anything, but if there's a better way to write it I'm happy to change.

The bindgen toolchain is also kind of weird because it asks for a label to a library like `libstdc++.dylib` but there's really no guarantee that 1) clang was compiled with libstdc++ (on macOS, it isn't -- libc++ is used instead), and 2) there's several other dynamic libraries that are typically required, such as `libc++abi.dylib` and `libunwind.dylib`. Still, this all just happens to work because all of these typically reside in the same directory, and the starlark rule just gets the directory from whatever you pass to the libstdcxx attribute, but I suppose if you were hooking this up to a fully hermetic bazel build of LLVM that might not be true.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -29,7 +29,7 @@ tasks:
       - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
       - "..."
       - "@examples//..."
-      # This test fails under bazel test but not when run directly due to -rpath issues.
+      # This test requires --incompatible_macos_set_install_name and Bazel 4.2.0+
       - "-@examples//ffi/rust_calling_c:matrix_dylib_test"
     build_targets: *osx_targets
     test_targets: *osx_targets
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -29,10 +29,8 @@ tasks:
       - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
       - "..."
       - "@examples//..."
-      # Skip tests for dylib support on osx, since we don't support it yet.
+      # This test fails under bazel test but not when run directly due to -rpath issues.
       - "-@examples//ffi/rust_calling_c:matrix_dylib_test"
-      - "-@examples//ffi/rust_calling_c:matrix_dynamically_linked"
-      - "-@examples//ffi/rust_calling_c/simple/..."
     build_targets: *osx_targets
     test_targets: *osx_targets
     build_flags:
--- a/bindgen/BUILD.bazel
+++ b/bindgen/BUILD.bazel
@@ -19,10 +19,13 @@ rust_bindgen_toolchain(
         "//conditions:default": "@bindgen_clang_linux//:clang",
     }),
     libclang = select({
-        "//rust/platform:osx": "@bindgen_clang_osx//:libclang.so",
+        "//rust/platform:osx": "@bindgen_clang_osx//:libclang.dylib",
         "//conditions:default": "@bindgen_clang_linux//:libclang.so",
     }),
-    libstdcxx = "@local_libstdcpp//:libstdc++",
+    libstdcxx = select({
+        "//rust/platform:osx": "@bindgen_clang_osx//:lib/libc++.dylib",
+        "//conditions:default": "@bindgen_clang_linux//:lib/libc++.so",
+    }),
 )
 
 toolchain(
--- a/bindgen/BUILD.bazel
+++ b/bindgen/BUILD.bazel
@@ -24,7 +24,7 @@ rust_bindgen_toolchain(
     }),
     libstdcxx = select({
         "//rust/platform:osx": "@bindgen_clang_osx//:lib/libc++.dylib",
-        "//conditions:default": "@bindgen_clang_linux//:lib/libc++.so",
+        "//conditions:default": None,
     }),
 )
 
--- a/bindgen/BUILD.bazel
+++ b/bindgen/BUILD.bazel
@@ -19,11 +19,11 @@ rust_bindgen_toolchain(
         "//conditions:default": "@bindgen_clang_linux//:clang",
     }),
     libclang = select({
-        "//rust/platform:osx": "@bindgen_clang_osx//:libclang.dylib",
-        "//conditions:default": "@bindgen_clang_linux//:libclang.so",
+        "//rust/platform:osx": "@bindgen_clang_osx//:libclang",
+        "//conditions:default": "@bindgen_clang_linux//:libclang",
     }),
     libstdcxx = select({
-        "//rust/platform:osx": "@bindgen_clang_osx//:lib/libc++.dylib",
+        "//rust/platform:osx": "@bindgen_clang_osx//:libc++",
         "//conditions:default": None,
     }),
 )
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -125,6 +125,7 @@ def _rust_bindgen_impl(ctx):
 
     if libstdcxx:
         env["LD_LIBRARY_PATH"] = ":".join([f.dirname for f in _get_libs_for_static_executable(libstdcxx).to_list()])
+        # DYLD_LIBRARY_PATH is LD_LIBRARY_PATH on macOS.
         env["DYLD_LIBRARY_PATH"] = env["LD_LIBRARY_PATH"]
 
     ctx.actions.run(
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -123,9 +123,10 @@ def _rust_bindgen_impl(ctx):
         "RUST_BACKTRACE": "1",
     }
 
+    # Set the dynamic linker search path so that clang uses the libstdcxx from the toolchain.
+    # DYLD_LIBRARY_PATH is LD_LIBRARY_PATH on macOS.
     if libstdcxx:
         env["LD_LIBRARY_PATH"] = ":".join([f.dirname for f in _get_libs_for_static_executable(libstdcxx).to_list()])
-        # DYLD_LIBRARY_PATH is LD_LIBRARY_PATH on macOS.
         env["DYLD_LIBRARY_PATH"] = env["LD_LIBRARY_PATH"]
 
     ctx.actions.run(
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -132,6 +132,7 @@ def _rust_bindgen_impl(ctx):
 
     if libstdcxx:
         env["LD_LIBRARY_PATH"] = ":".join([f.dirname for f in _get_libs_for_static_executable(libstdcxx).to_list()])
+        env["DYLD_LIBRARY_PATH"] = env["LD_LIBRARY_PATH"]
 
     ctx.actions.run(
         executable = bindgen_bin,
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -87,13 +87,6 @@ def _rust_bindgen_impl(ctx):
     rustfmt_bin = toolchain.rustfmt or rust_toolchain.rustfmt
     clang_bin = toolchain.clang
     libclang = toolchain.libclang
-
-    # TODO: This rule shouldn't need to depend on libstdc++
-    #  This rule requires an explicit dependency on a libstdc++ because
-    #    1. It is a runtime dependency of libclang.so
-    #    2. We cannot locate it in the cc_toolchain yet
-    #  Depending on how libclang.so was compiled, it may try to locate its libstdc++ dependency
-    #  in a way that makes our handling here unnecessary (eg. system /usr/lib/x86_64-linux-gnu/libstdc++.so.6)
     libstdcxx = toolchain.libstdcxx
 
     # rustfmt is not where bindgen expects to find it, so we format manually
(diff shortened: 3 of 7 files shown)


UebelAndre:
I think this would close https://github.com/bazelbuild/rules_rust/issues/741

djmarcin:
Oops -- one mistake. I didn't notice that bindgen_clang_linux is actually compiled against libstdc++.so and that isn't provided in bindgen_clang_linux, so I think we still need to get the path to libstdc++.so from somewhere else.

djmarcin:
Ok -- I think this is updated and fully working on macOS and Linux and does not break backward compatibility. Some notes:
- It appears that there was confusion over what the use of libstdc++ or libc++ was here. It's essentially used to allow clang itself to run more hermetically by causing it to load the bundled libraries over the system libraries where possible. The library needs to be the library clang itself was compiled against -- it is _not_ used for any binary produced using the generated code, so it should _not_ come from the current C++ toolchain.
- There's really no need for the toolchain to demand that the libraries have `CcInfo`. From its perspective they are just opaque files. We do not need to pass these files to any compilation actions. I did not change this, but I would be open to removing that requirement in this PR.
- On Linux, since we're just assuming the presence of libstdc++.so in the normal system location anyway, it provides no value (and could break things if files are in unusual locations) to have a local repository for this file rather than allow the usual dynamic linker symbol resolution to take place.
- Passing the libs to an sh_binary's data attribute does not cause the binary to attempt to use those libraries first -- I don't think this data attribute is necessary.

UebelAndre (changes requested):
Can you enable the tests in [.bazelci/presubmit.yml](https://github.com/bazelbuild/rules_rust/blob/c9345d33123e49928a021f75725da2f5b609f660/.bazelci/presubmit.yml)?

djmarcin:
Done -- but one of the tests still fails under bazel test because of -rpath problems, I think. It works fine when run directly, but since bazel test has a different working directory, it fails. I left that one disabled but all the others work.

UebelAndre:
> Done -- but one of the tests still fails under bazel test because of -rpath problems, I think. It passes when run directly, but since bazel test has a different working directory, it fails. I left that one disabled but all the others work.

Could you link the build failure you see for the `@examples//ffi/rust_calling_c:matrix_dylib_test` target?

djmarcin:
There's no build failure, it fails at runtime:
```
$ bazel test @examples//ffi/rust_calling_c:matrix_dylib_test
DEBUG: /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14:
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:125:14: buildkite_config not using checked in configs; Bazel version 4.1.0 was picked/selected but no checked in config was found in map {"0.20.0": ["8.0.0"], "0.21.0": ["8.0.0"], "0.22.0": ["8.0.0", "9.0.0"], "0.23.0": ["8.0.0", "9.0.0"], "0.23.1": ["8.0.0", "9.0.0"], "0.23.2": ["9.0.0"], "0.24.0": ["9.0.0"], "0.24.1": ["9.0.0"], "0.25.0": ["9.0.0"], "0.25.1": ["9.0.0"], "0.25.2": ["9.0.0"], "0.26.0": ["9.0.0"], "0.26.1": ["9.0.0"], "0.27.0": ["9.0.0"], "0.27.1": ["9.0.0"], "0.28.0": ["9.0.0"], "0.28.1": ["9.0.0"], "0.29.0": ["9.0.0"], "0.29.1": ["9.0.0", "10.0.0"], "1.0.0": ["9.0.0", "10.0.0"], "1.0.1": ["10.0.0"], "1.1.0": ["10.0.0"], "1.2.0": ["10.0.0"], "1.2.1": ["10.0.0"], "2.0.0": ["10.0.0"], "2.1.0": ["10.0.0"], "2.1.1": ["10.0.0", "11.0.0"], "2.2.0": ["11.0.0"], "3.0.0": ["11.0.0"], "3.1.0": ["11.0.0"], "3.2.0": ["11.0.0"], "3.3.0": ["11.0.0"], "3.3.1": ["11.0.0"], "3.4.1": ["11.0.0"], "3.5.0": ["11.0.0"], "3.5.1": ["11.0.0"], "3.6.0": ["11.0.0"], "3.7.0": ["11.0.0"], "3.7.1": ["11.0.0"], "3.7.2": ["11.0.0"], "4.0.0": ["11.0.0"]}
INFO: Analyzed target @examples//ffi/rust_calling_c:matrix_dylib_test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
FAIL: @examples//ffi/rust_calling_c:matrix_dylib_test (see /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/execroot/rules_rust/bazel-out/darwin-fastbuild/testlogs/external/examples/ffi/rust_calling_c/matrix_dylib_test/test.log)
Target @examples//ffi/rust_calling_c:matrix_dylib_test up-to-date:
  bazel-bin/external/examples/ffi/rust_calling_c/matrix_dylib_test
INFO: Elapsed time: 0.729s, Critical Path: 0.16s
INFO: 2 processes: 2 darwin-sandbox.
INFO: Build completed, 1 test FAILED, 2 total actions
@examples//ffi/rust_calling_c:matrix_dylib_test                          FAILED in 0.1s
  /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/execroot/rules_rust/bazel-out/darwin-fastbuild/testlogs/external/examples/ffi/rust_calling_c/matrix_dylib_test/test.log

INFO: Build completed, 1 test FAILED, 2 total actions

$ cat /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/execroot/rules_rust/bazel-out/darwin-fastbuild/testlogs/external/examples/ffi/rust_calling_c/matrix_dylib_test/test.log
Executing tests from @examples//ffi/rust_calling_c:matrix_dylib_test
dyld: Library not loaded: bazel-out/darwin-fastbuild/bin/external/examples/ffi/rust_calling_c/c/libnative_matrix_so.so
  Referenced from: /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/sandbox/darwin-sandbox/112/execroot/rules_rust/bazel-out/darwin-fastbuild/bin/external/examples/ffi/rust_calling_c/matrix_dylib_test.launcher.runfiles/rules_rust/external/examples/ffi/rust_calling_c/matrix_dylib_test
  Reason: image not found
```

But it passes just fine if you run it from bazel-bin:
```
$ bazel-bin/external/examples/ffi/rust_calling_c/matrix_dylib_test

running 3 tests
test test::test_equal ... ok
test test::test_size ... ok
test test::test_transpose ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ echo $?
0
```

I think the problem is that the -rpath is: `bazel-out/darwin-fastbuild/bin/external/examples/ffi/rust_calling_c/c/libnative_matrix_so.so` which isn't found when it executes within the test sandbox
```
$ otool -L bazel-bin/external/examples/ffi/rust_calling_c/matrix_dylib_test
bazel-bin/external/examples/ffi/rust_calling_c/matrix_dylib_test:
	bazel-out/darwin-fastbuild/bin/external/examples/ffi/rust_calling_c/c/libnative_matrix_so.so (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 905.6.0)
```

djmarcin:
On Linux, the equivalent path to the library is an absolute path instead:
```
$ ldd bazel-bin/external/examples/ffi/rust_calling_c/matrix_dylib_test
        linux-vdso.so.1 (0x00007ffe57fb0000)
        libnative_matrix_so.so => /home/david/rules_rust/bazel-bin/external/examples/ffi/rust_calling_c/../../../../_solib_k8/_U@examples_S_Sffi_Srust_Ucalling_Uc_Sc_Cnative_Umatrix_Uso___Uexternal_Sexamples_Sffi_Srust_Ucalling_Uc_Sc/libnative_matrix_so.so (0x00007f972f165000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f972ef5e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f972f22d000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f972ee0f000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f972edf4000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f972edd1000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f972edc9000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f972ebe7000)
```

.bazelci/presubmit.yml
```diff
@@ -29,10 +29,8 @@ tasks:
       - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
       - "..."
       - "@examples//..."
-      # Skip tests for dylib support on osx, since we don't support it yet.
+      # This test fails under bazel test but not when run directly due to -rpath issues.
```
UebelAndre, line 32: Can you make an issue that shows the runtime error and link that here so it's easier to keep track of? 🙏

bindgen/BUILD.bazel
```diff
@@ -19,10 +19,13 @@ rust_bindgen_toolchain(
...
     }),
     libclang = select({
-        "//rust/platform:osx": "@bindgen_clang_osx//:libclang.so",
-        "//conditions:default": "@bindgen_clang_linux//:libclang.so",
+        "//rust/platform:osx": "@bindgen_clang_osx//:libclang",
+        "//conditions:default": "@bindgen_clang_linux//:libclang",
+    }),
+    libstdcxx = select({
+        "//rust/platform:osx": "@bindgen_clang_osx//:libc++",
+        "//conditions:default": None,
```
UebelAndre, line 27: Could this perhaps be the reason the build succeeds on linux?

UebelAndre (changes requested):
Couple of questions but otherwise I think this is pretty exciting stuff!

bindgen/BUILD.bazel
djmarcin: No, this attribute does not affect the rpath in the resulting binary

.bazelci/presubmit.yml
djmarcin: This is related to this comment https://github.com/bazelbuild/rules_rust/issues/621#issuecomment-832382072. I'm trying to implement the change to generate rpath properly but it will still not work until we use a version of bazel that responds to `--incompatible_macos_set_install_name`

djmarcin:
With https://github.com/bazelbuild/rules_rust/pull/820/commits/385a573e9ad60f2555c878a3582ee2e4855d7812, a HEAD version of bazel, and passing `--incompatible_macos_set_install_name`, the test now works:
```
$ bazel test --incompatible_macos_set_install_name @examples//ffi/rust_calling_c:matrix_dylib_test
2021/07/02 11:37:13 Using unreleased version at commit 348b53232e363e4f303c66cac03ef829e566f147
DEBUG: /private/var/tmp/_bazel_djmarcin/d200bf66822c6cb3ca57630679b9e720/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:59:14:
Current running Bazel is not a release version and one was not defined explicitly in rbe_autoconfig target. Falling back to '4.0.0'
INFO: Analyzed target @examples//ffi/rust_calling_c:matrix_dylib_test (8 packages loaded, 96 targets configured).
INFO: Found 1 test target...
Target @examples//ffi/rust_calling_c:matrix_dylib_test up-to-date:
  bazel-bin/external/examples/ffi/rust_calling_c/matrix_dylib_test
INFO: Elapsed time: 0.709s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
@examples//ffi/rust_calling_c:matrix_dylib_test                 (cached) PASSED in 0.4s

Executed 0 out of 1 test: 1 test passes.
```

This also fixes the issue mentioned in https://github.com/bazelbuild/rules_rust/issues/621#issuecomment-832382072

UebelAndre (approved):
Awesome work! Super excited to finally see this fixed and the tests enabled!

PR #824, opened 2021-07-07
Added support for `noclippy` tag

UebelAndre:
This matches the [rustfmt](https://github.com/bazelbuild/rules_rust/blob/dd1c73e232ef5e1e8bf5c08e93c6fff557e18077/rust/private/rustfmt.bzl#L23-L25) pattern of `norustfmt` and allows users to opt out of running clippy on a particular target.


illicitonion (approved):

UebelAndre:
I ended up adding a regression test for this by enabling clippy in a `.bazelrc` config. This is matches what's being done for `rustfmt` but this also means the code will need to be clippy compliant. I've opened https://github.com/bazelbuild/rules_rust/pull/827 to address that.

illicitonion (approved):
I'm slightly concerned that CI now doesn't test the default config without the extra aspects, and that we may accidentally introduce a dependency on the aspects at some point which shouldn't be there, but as we already have this for rustfmt, if we want to introduce a "no-aspect" CI pipeline that feels like a reasonable separate change for the future.

PR #825, opened 2021-07-09
Fixes for rust_binary that depends on a cc_library that depends on a rust_library

krasimirgg:
Consider a case of a `rust_binary` that depends on a `cc_library` that itself depends
on a `rust_library`:

```
.
├── bin.rs
├── BUILD
└── lib.rs
```
```bazel
load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_library")

rust_library(
    name = "rust_lib",
    srcs = ["lib.rs"],
)

cc_library(
    name = "cc_lib",
    srcs = [],
    deps = [":rust_lib"],
)

rust_binary(
    name = "bin",
    srcs = ["bin.rs"],
    deps = [":cc_lib"],
)
```

```rust
// bin.rs
fn main() {
    println!("hi");
}
```

```rust
// lib.rs: empty
```

Running `bazel build //project:bin` runs into several linker errors, which this PR tries to address:
- When passing an argument like `-lcrate`, the linker looks for a static library named `libcrate.a`.
- the CC link actions get confused by static library names that have multiple dots, e.g., end in "libcrate.rlib.a", so
updated the scripts to produce them as "libcrate-rlib.a" instead.
- The libraries `panic_abort` and `panic_unwind` are alternatives; attempting to pass both to the linker runs into duplicate symbol errors. Updated the rules to prefer `panic_unwind` if both are available. This is since the standard library by default depends on `panic_unwind`. Note that in platforms where `panic_unwind` is not available, `panic_abort` will be picked. In the future we could add a feature that allows users to pick the panic strategy, aligned with how it will be done in upstream rustc (https://github.com/rust-lang/wg-cargo-std-aware/issues/29).

Another thing worth noting is that the way we currently process `.rlib`-s into static libs to pass on to CC rules doesn't quite work when using the current stable `rustc 1.53.0`; it works for the beta `rustc` and beyond: on Linux, a `crate.rlib` is an archive that contains .o files (good) and a Rust metadata file `lib.rmeta` (caution). We symlink `crate.rlib` into `libcrate.a`. The linker generally checks that all members of an archive passed in a `--whole-archive` section are ELF files. The commit https://github.com/rust-lang/rust/commit/c79419af0721c614d050f09b95f076da09d37b0d on Jun 4 updated `rustc` so that the produced `lib.rmeta` is an ELF file (previously it was a binary blob), which satisfies the linker checks. This PR does not attempt to address the case where`lib.rmeta` is not an ELF file, hence even with this PR in, trying to run the example with the current stable `rustc 1.53.0` will produce linker errors complaining that an archive member is not recognized, like this one `/usr/bin/ld.gold: error: rust_lib: plugin failed to claim member lib.rmeta at 2480`.

With this PR, `bazel build //project:bin` works (with sufficiently recent `rustc`, as described above).


UebelAndre:
Hi and thanks for opening this PR! 😄 

Could you also add a regression test to the `test` package?

krasimirgg:
> Could you also add a regression test to the `test` package?

Added some tests.

rust/toolchain.bzl
```diff
@@ -35,7 +41,19 @@ def _rust_stdlib_filegroup_impl(ctx):
         #
         # alloc depends on the allocator_library if it's configured, but we
         # do that later.
-        dot_a_files = [_make_dota(ctx, f) for f in std_rlibs]
+        #
```
UebelAndre, line 44: Small nit since this seems separate from the block comment (but might be just me)
```suggestion

```

UebelAndre (changes requested):
Thanks for adding a test!

I think this generally looks pretty good but I'm wondering if the logic for parsing out `panic_unwind` vs `panic_abort` could be moved out of `rust_stdlib_filegroup` and into `rust_toolchain` (specifically [_make_libstd_and_allocator_ccinfo](https://github.com/bazelbuild/rules_rust/blob/ed64716a87d0ab3a7ad02a3c0742a084b256930a/rust/toolchain.bzl#L105)). I think the `rust_stdlib_filegroup` rule should be gathering everything from a `rust_std` artifact but not making any decisions on what to or not-to include. I would say, either leave both `panic_unwind` and `panic_abort` in `std_rlibs` (and therefor `dot_a_files`) and parse them out later, or update the [StdLibInfo](https://github.com/bazelbuild/rules_rust/blob/ed64716a87d0ab3a7ad02a3c0742a084b256930a/rust/private/providers.bzl#L50-L65) provider to have `panic_unwind_files` and `panic_abort_files` that have the appropriate things parsed out of the various fields there. I think I lean towards the latter but not strongly. How does that sound?

rust/toolchain.bzl
krasimirgg: Thank you! I'll look into moving the `panic_unwind` vs `panic_abort` logic out of `rust_stdlib_filegroup` and report back.

krasimirgg: Moved the panic filtering logic in `_make_libstd_and_allocator_ccinfo`. I think I'm satisfied with this approach until we have a more flexible way to let the user specify the panic strategy.

UebelAndre (approved):
This looks good to me! Thank you! 🙏

I'll give @hlopko a chance to take a look before merging just to be safe 😅

rust/private/rustc.bzl
```diff
@@ -635,7 +635,7 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
         # bazel hard-codes a check for endswith((".a", ".pic.a",
         # ".lib")) in create_library_to_link, so we work around that
         # by creating a symlink to the .rlib with a .a extension.
-        dot_a = ctx.actions.declare_file(crate_info.name + ".a", sibling = crate_info.output)
+        dot_a = ctx.actions.declare_file("lib" + crate_info.name + ".a", sibling = crate_info.output)
```
hlopko, line 638: Since you have this code in your brain, is there a reason why we don't use the same logic here and in the toolchain.bzl to create symlinks? Could we do the same?

rust/toolchain.bzl
```diff
@@ -3,18 +3,24 @@
 load("//rust/private:common.bzl", "rust_common")
 load("//rust/private:utils.bzl", "dedent", "find_cc_toolchain")
 
-def _make_dota(ctx, file):
-    """Add a symlink for a file that ends in .a, so it can be used as a staticlib.
+def _make_dota(ctx, rlib_file, postfix = ""):
```
hlopko, line 6: Could we omit the .rlib/-rlib suffix completely? So `libcore-3e9f2a4814daef68.rlib` will have a symlink named `libcore-3e9f2a4814daef68.a`? I guess we can find a different way to find rust stdlib rlibs in tests (for example by looking at the directory path?)

hlopko (commented):
LGTM, only minor questions

rust/private/rustc.bzl
krasimirgg: That sounds good! I'll extract a helper function and use it from both places.

rust/toolchain.bzl
krasimirgg: Ideally, yes. AFAIK the reason why the suffix exists today is so that the stdlib_ordering unit test can distinguish between standard and user-defined libs:
https://github.com/bazelbuild/rules_rust/blob/404b84bd6f614cfff19d22de9a984d2f9717ca50/test/unit/stdlib_ordering/stdlib_ordering.bzl#L33

This seems pretty fragile. I'll look into sieving out the standard from user-defined libs in a different way there, so we can drop the prefix here.

rust/private/utils.bzl
```diff
@@ -296,3 +296,30 @@ def dedent(doc_string):
         # Remove the leading block of spaces from the current line
         block = " " * space_count
         return "\n".join([line.replace(block, "", 1).rstrip() for line in lines])
+
+def make_static_lib(ctx, rlib_file):
```
UebelAndre, line 300: Nit: Can you move this to [rust/private/toolchain_utils.bzl](https://github.com/bazelbuild/rules_rust/blob/8cca226c6bc3c55444c98a6a08e7d19b6dd028f2/rust/private/toolchain_utils.bzl)?

krasimirgg: Just to clarify: this is referenced from both [rust/toolchain.bzl](https://github.com/bazelbuild/rules_rust/blob/8cca226c6bc3c55444c98a6a08e7d19b6dd028f2/rust/toolchain.bzl) and [rust/private/rustc.bzl](https://github.com/bazelbuild/rules_rust/pull/825/files#diff-ecce7f6cdb4bc2f884f23fccb9a20174ad2b357dc6745314054e5eb8b784d2e0R636), so moving it to rust/private/toolchain_utils.bzl would require adding a dependency for rust/private/rustc.bzl on rust/private/toolchain_utils.bzl, which seems a bit off. Is that expected?

UebelAndre: Ah, I missed that it was in `rustc.bzl`, disregard!

rust/private/rustc.bzl
```diff
@@ -635,8 +636,7 @@ def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configur
         # bazel hard-codes a check for endswith((".a", ".pic.a",
         # ".lib")) in create_library_to_link, so we work around that
         # by creating a symlink to the .rlib with a .a extension.
-        dot_a = ctx.actions.declare_file(crate_info.name + ".a", sibling = crate_info.output)
-        ctx.actions.symlink(output = dot_a, target_file = crate_info.output)
+        dot_a = make_static_lib(ctx, crate_info.output)
```
hlopko, line 639: Nit: make_static_lib_symlink

rust/private/utils.bzl
```diff
@@ -296,3 +296,30 @@ def dedent(doc_string):
...
+    """Add a .a symlink to an .rlib file, so it can be used as a staticlib.
```
hlopko, line 301:
```suggestion
    """Add a .a symlink to an .rlib file.
    
    As of Bazel 4.1.0, C++ providers reject the .rlib, but accept .a as a static library.
```

rust/private/utils.bzl
```diff
@@ -296,3 +296,30 @@ def dedent(doc_string):
...
+    Returns:
+        The symlink's File.
+    """
+    if not rlib_file.basename.endswith(".rlib"):
+        fail("file is not an .rlib: ", rlib_file.basename)
+    basename = rlib_file.basename[:-5]
+    if not basename.startswith("lib"):
+        basename = "lib" + basename
+    dot_a = ctx.actions.declare_file(basename + ".a", sibling = rlib_file)
+    ctx.actions.symlink(output = dot_a, target_file = rlib_file)
```
hlopko, line 324: SuperNit: this function actually only needs ctx.actions, not the whole ctx. Let's pass actions there.

rust/toolchain.bzl
```diff
@@ -140,8 +126,17 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
             transitive = [between_alloc_and_core_inputs],
             order = "topological",
         )
+
+        # The libraries panic_abort and panic_unwind are alternatives.
+        # The std by default requires panic_unwind.
+        # Exclude panic_abort if panic_unwind is present.
+        # TODO: Provide a setting to choose between panic_abort and panic_unwind.
+        filtered_between_core_and_std_files = rust_stdlib_info.between_core_and_std_files
+        has_panic_unwind = [f for f in filtered_between_core_and_std_files if "panic_unwind" in f.basename]
```
hlopko, line 135: Nit: wrap long line

rust/toolchain.bzl
```diff
@@ -140,8 +126,17 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
...
+        if has_panic_unwind:
+            filtered_between_core_and_std_files = [f for f in filtered_between_core_and_std_files if "panic_abort" not in f.basename]
         between_core_and_std_inputs = depset(
-            [_ltl(f, ctx, cc_toolchain, feature_configuration) for f in rust_stdlib_info.between_core_and_std_files],
+            [_ltl(f, ctx, cc_toolchain, feature_configuration) for f in filtered_between_core_and_std_files],
```
hlopko, line 139: Nit: wrap long line

test/unit/stdlib/stdlib.bzl
```diff
@@ -26,17 +26,37 @@ def _dedup_preserving_order(list):
         r.append(e)
     return r
 
+def _stdlibs(tut):
+    """Given a target, return the list of its standard rust libraries."""
+    libs = [lib.static_library for li in tut[CcInfo].linking_context.linker_inputs.to_list() for lib in li.libraries]
```
hlopko, line 31: Nit: wrap long line

hlopko (approved):
Only nits, thank you!

PR #829, opened 2021-07-10
Add support for `self-contained` object files (musl).

UebelAndre:
In newer `rust-std` artifacts there is a `self-contained` directory which contains some `.o` files. These are currently ignored but without them the rules cannot target `musl` and users will see a similar linker error:
```
error: linking with `...`
  |
  = note: `...`
  = note: `...`: error: rct1.o: No such file or directory
          `...`: error: crti.o: No such file or directory
          `...`: error: crtbeginS.o: No such file or directory
          `...`: error: crtendS.o: No such file or directory
          `...`: error: crtn.o: No such file or directory
```

These files appear to have been introduced in Rust `1.46.0`. They seem related to https://github.com/rust-lang/rust/pull/70740/ but I'm still not clear what mechanism actually generates and installs these directories in the artifacts.

Regardless, this change addresses these linker issues without impacting other target triples as far as I can tell.


UebelAndre:
I also wrote this script to find what artifacts contained the `self-contained` directory and what was inside it.
```bash
#!/bin/bash

set -euo pipefail

MUSL=(
rust-std-1.26.0-aarch64-unknown-linux-musl
rust-std-1.26.0-x86_64-unknown-linux-musl
rust-std-1.36.0-aarch64-unknown-linux-musl
rust-std-1.36.0-aarch64-unknown-linux-musl
rust-std-1.45.0-aarch64-unknown-linux-musl
rust-std-1.45.0-x86_64-unknown-linux-musl
rust-std-1.46.0-aarch64-unknown-linux-musl
rust-std-1.46.0-x86_64-unknown-linux-musl
rust-std-1.53.0-aarch64-unknown-linux-musl
)

OTHER=(
rust-std-1.45.0-x86_64-apple-darwin
rust-std-1.45.0-wasm32-unknown-unknown
rust-std-1.45.0-x86_64-pc-windows-msvc
rust-std-1.46.0-x86_64-apple-darwin
rust-std-1.46.0-wasm32-unknown-unknown
rust-std-1.46.0-x86_64-pc-windows-msvc
)

function find_self_contained() {
    temp_dir="$(mktemp -d -t rust-self-contained)"
    pushd "${temp_dir}"
    for art in ${@}; do
        echo "${art}"
        wget --quiet https://static.rust-lang.org/dist/${art}.tar.gz
        tar -xf "${art}.tar.gz"
        find "${art}" -name "self-contained" -exec tree {} \;
    done
    popd && rm -rf "${temp_dir}"
}

echo "Test MUSL"
find_self_contained ${MUSL[@]}

echo "Test OTHER"
find_self_contained ${OTHER[@]}
```

Only `musl` populates this directory
```output
Test MUSL
/var/folders/hf/phjl9q7501gb5wt_qr4ltn3m0000gn/T/rust-self-contained.KEpXUBeb ~/Code/rules_rust
rust-std-1.26.0-aarch64-unknown-linux-musl
rust-std-1.26.0-x86_64-unknown-linux-musl
rust-std-1.36.0-aarch64-unknown-linux-musl
rust-std-1.36.0-aarch64-unknown-linux-musl
rust-std-1.45.0-aarch64-unknown-linux-musl
rust-std-1.45.0-x86_64-unknown-linux-musl
rust-std-1.46.0-aarch64-unknown-linux-musl
rust-std-1.46.0-aarch64-unknown-linux-musl/rust-std-aarch64-unknown-linux-musl/lib/rustlib/aarch64-unknown-linux-musl/lib/self-contained
├── Scrt1.o
├── crt1.o
├── crti.o
├── crtn.o
└── rcrt1.o

0 directories, 5 files
rust-std-1.46.0-x86_64-unknown-linux-musl
rust-std-1.46.0-x86_64-unknown-linux-musl/rust-std-x86_64-unknown-linux-musl/lib/rustlib/x86_64-unknown-linux-musl/lib/self-contained
├── Scrt1.o
├── crt1.o
├── crti.o
├── crtn.o
└── rcrt1.o

0 directories, 5 files
rust-std-1.53.0-aarch64-unknown-linux-musl
rust-std-1.53.0-aarch64-unknown-linux-musl/rust-std-aarch64-unknown-linux-musl/lib/rustlib/aarch64-unknown-linux-musl/lib/self-contained
├── Scrt1.o
├── crt1.o
├── crtbegin.o
├── crtbeginS.o
├── crtend.o
├── crtendS.o
├── crti.o
├── crtn.o
└── rcrt1.o

0 directories, 9 files
~/Code/rules_rust
Test OTHER
/var/folders/hf/phjl9q7501gb5wt_qr4ltn3m0000gn/T/rust-self-contained.0crH80lq ~/Code/rules_rust
rust-std-1.45.0-x86_64-apple-darwin
rust-std-1.45.0-wasm32-unknown-unknown
rust-std-1.45.0-x86_64-pc-windows-msvc
rust-std-1.46.0-x86_64-apple-darwin
rust-std-1.46.0-x86_64-apple-darwin/rust-std-x86_64-apple-darwin/lib/rustlib/x86_64-apple-darwin/lib/self-contained

0 directories, 0 files
rust-std-1.46.0-wasm32-unknown-unknown
rust-std-1.46.0-wasm32-unknown-unknown/rust-std-wasm32-unknown-unknown/lib/rustlib/wasm32-unknown-unknown/lib/self-contained

0 directories, 0 files
rust-std-1.46.0-x86_64-pc-windows-msvc
rust-std-1.46.0-x86_64-pc-windows-msvc/rust-std-x86_64-pc-windows-msvc/lib/rustlib/x86_64-pc-windows-msvc/lib/self-contained

0 directories, 0 files
```

UebelAndre:
Also, I'm not sure how to add a test for `musl`... The changes here work though 😅

illicitonion (commented):
I don't feel like I have enough context to really review this (or indeed to suggest how to test it on CI), but in general the change looks reasonable? If anyone has more context, it'd be great for them to review, otherwise I can somewhat unknowingly approve...

rust/toolchain.bzl
```diff
@@ -126,6 +132,23 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
         """).format(ctx.label, ctx.attr.rust_lib))
     rust_stdlib_info = ctx.attr.rust_lib[rust_common.stdlib_info]
 
+    if rust_stdlib_info.self_contained_files:
+        compilation_outputs = cc_common.create_compilation_outputs(
+            objects = depset(rust_stdlib_info.self_contained_files),
+        )
+
+        linking_context, _linking_outputs = cc_common.create_linking_context_from_compilation_outputs(
```
hlopko, line 140: So this will register an archiving action that will put all objects into the archive. Is it intended? Alternatively you can put those objects into https://docs.bazel.build/versions/3.5.0/skylark/lib/LibraryToLink.html#objects. I'm not sure if ld.exe will be able to use `--start-lib/--end-lib` for those objects though. Gold and lld will be fine with not having a static library.

rust/toolchain.bzl
```diff
@@ -151,22 +174,29 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_lib, allocator_library):
...
+        link_inputs = cc_common.create_linker_input(
             owner = rust_lib.label,
             libraries = std_inputs,
-        ))
+        )
 
-    allocator_inputs = None
-    if allocator_library:
-        allocator_inputs = [allocator_library[CcInfo].linking_context.linker_inputs]
+        allocator_inputs = None
```
hlopko, line 182: Why this change? In general allocator_library is independent from the `rust_stdlib_info`. I can imagine a scenario where people want to use the system installed rust stdlib (therefore without declaring inputs in the `rust_toolchain`, but they still want to add `allocator_library` explicitly.

In general `allocator_library` contains symbols that normally rustc generates on demand when producing a binary. When using clang/lld obviously they don't know how to produce those symbols, so the linking fails. Allocator_library is responsible for defining those symbols and this library is typically part of the toolchain (it currently cannot be system installed because clang needs to be told should use it).

hlopko: (the change I'm talking about is indenting the code to the right, so it is only executed when there are `std_rlibs`)

hlopko:
Is there a way we could have a test?

rust/toolchain.bzl
UebelAndre: The code today only returns a `CcInfo` if `link_inputs` has been populated which only happens when `std_rlibs` is populated.:
https://github.com/bazelbuild/rules_rust/blob/8cca226c6bc3c55444c98a6a08e7d19b6dd028f2/rust/toolchain.bzl#L159-L170

This is not a change in behavior and I thought it'd be clearer to gate this behind the same check. If this is incorrect then I'd say it was originally a bug that I'm happy to do something about but I otherwise think this is a good improvement that makes the new approach more readable.

rust/toolchain.bzl
UebelAndre: I think you'd know better than I here. The goal was to find a mechanism that wires up these object files together with the existing `CcInfo` provider that was being generated using [cc_common](https://docs.bazel.build/versions/4.0.0/skylark/lib/cc_common.html). Is there a more direct path to achieve this?

UebelAndre:
> Is there a way we could have a test?

I had the same question (https://github.com/bazelbuild/rules_rust/pull/829#issuecomment-877664732) and opened https://github.com/bazelbuild/continuous-integration/issues/1190 to try and answer it. Any suggestions?

rust/toolchain.bzl
hlopko: I looked around and it seems what you do is correct :)

rust/toolchain.bzl
hlopko: You're right. Go ahead :)

hlopko (approved):

PR #831, opened 2021-07-12
Allows deps to be specified in rust_bindgen_library

csmulhern:


UebelAndre (approved):
Thanks for catching this!

Issue #600, opened 2021-02-23
Define rules_rust stability policy


hlopko:
The current status quo is that we try to not break compatibility too much. We need to define clearly:
* what is subject to backward compatibility
* how is the compatibility verified
* what's the process for breaking it

illicitonion:
An additional point we should define is our policy around bazel version compatibility.

PR #832, opened 2021-07-14
Define backwards compatibility policy for `rules_rust`

hlopko:
# Introduction

This document proposes a backward compatibility policy for `rules_rust` and tries to find a balance between:

*   being able to add new features and improve the rule implementation through backwards incompatible changes
*   giving reliability guarantees to our existing and new users

As of 2021-07-08 the [Bazel backward compatibility](https://docs.bazel.build/versions/main/backward-compatibility.html) policy can be summarized as:

*   Bazel LTS releases are cut approximately every 9 months and are supported for 3 years. Each LTS release increments the major version number.
*   Bazel rolling releases are cut approximately once a week and can contain breaking changes. There are no guarantees of stability between rolling releases, or between a rolling release and the current LTS. In other words, Bazel LTS version 7.0.0 can be incompatible with 8.0.0pre1, and both can be incompatible with 8.0.0pre2.

# Proposed policy

See the added `COMPATIBILITY.md` file.

# Alternatives Considered

## Match Bazel versioning

Instead of using SemVer, `rules_rust` could follow Bazel's versioning scheme. Major version number will match the Bazel LTS version supported. Minor version number will be increased on breaking changes, patch version number will be increased on backwards compatible changes.

I don't feel strongly about this or SemVer. I picked SemVer because that is what is the standard in the Rust ecosystem, but I'm open to doing the standard thing in the Bazel ecosystem as well :)

## Supporting past releases

Current proposal says the `rules_rust` are developed for the latest Bazel LTS (and the latest Bazel rolling release on the best effort basis). It does not promise old releases will be maintained. \
 \
For users wanting to use Bazel LTS for the whole lifetime of it (not only the active phase that is typically 9 months, but also the inactive phase that is up to 3 years, in which the LTS will receive maintenance updates) supporting LTS for 9 months is not enough.

When a new Bazel LTS becomes active, `rules_rust` could create a branch for the old LTS as they create a new `rules_rust` release with the major number incremented. Even better, when `rules\_rust` drop compatibility with an LTS release we could create a branch.

`rules_rust` would then cherry pick and create patch releases for inactive LTS for critical bug fixes, OS compatibility patches, and security patches. No new features will be cherry picked.

If the community and maintainers of `rules_rust` are willing to make this promise, I'm open to that.

## Supporting all Bazel rolling releases of the current epoch

This approach is inspired by the Bazel LTS releases, but promises more stability. This approach maintains SemVer, and each minor version release of the `rules_rust` aims to support all Bazel rolling releases of the given major version number. In other words, a release of `rules_rust` marked as 4.X would be backwards compatible with all 4.Y releases where X > Y, and always compatible with Bazel 4.0 LTS, and all 5.0.0pre Bazel rolling releases for all Z releases released before release 4.X of `rules_rust`. When Bazel releases 5.0 we'd also release 5.0 which does not promise backwards compatibility with 4.x. New features will be made available to all Bazel versions.

This approach was dismissed because of its maintenance overhead. Since Bazel promises no backwards compatibility between rolling releases, `rules_rust` might find themselves in a situation where they would have to release multiple implementations of the rules, and depending on the Bazel version pick one that works. We have to be prepared to do this for Bazel LTS releases (in the past we used a `BAZEL_VERSION` check in a repository rule that conditionally generated a `.bzl` file with implementations of functions for that Bazel version that were used by the rules in analysis). Having to do the same for each Bazel rolling release is too big of a burden - as it also was for the Bazel team, and the result was the current backwards compatibility policy involving LTS and rolling releases that are backwards incompatible.

Second downside of this approach is the lost control over when we can make backward incompatible releases of `rules_rust`. We want to reserve the right to be pragmatic and decide to make a backward incompatible change as a reaction to a substantial Bazel backward incompatible change.

## Supporting the current Bazel rolling release

Currently the proposal doesn't promise strict compatibility with the current Bazel rolling release. The compatibility is only on the best effort basis.

Bazel rolling releases are planned to be released rather often (approximately weekly), and each can potentially break `rules_rust`. Promising strict compatibility with the current Bazel rolling release can create frequent high priority wor. `rules_rust` maintainers would have to be able to potentially every week stop what they were doing and fix Bazel backward incompatible changes (and do so in a way that is backward compatible with Bazel LTS) to be able to continue regular rules development (to have a green CI) and to be able make new releases.

I believe maintainers of `rules_rust` don't have the capacity to make this promise responsibly.

Fixes https://github.com/bazelbuild/rules_rust/issues/600.

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -55,8 +55,8 @@ are only available in Bazel rolling releases). `rules_rust` will aim that new
 features available to the current Bazel rolling release will be available to the
 next Bazel LTS release at latest.
 
-Whenever there is a new Bazel LTS release, all releases of `rules_rust` in the
-first 3 months will support both Bazel LTS versions unless Bazel doesn't allow this.
+Whenever there is a new Bazel LTS release, all releases of `rules_rust` will maintain
+support for the older LTS version for at least 3 months unless Bazel doesn't allow this.
 
 ## What host platforms are supported?
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -54,9 +54,9 @@ first 3 months will support both Bazel LTS versions.
 
 ## What host platforms are supported?
 
-The only platform subject to backwards compatibility policy is `x86_64-unknown-linux-gnu` on
-`x86_64`, as the only fully supported platform. Process for moving a best effort
-platform to a supported platform is consensus-based.
+The only platform subject to backwards compatibility policy is
+`x86_64-unknown-linux-gnu`, as the only fully supported platform. Process for
+moving a best effort platform to a supported platform is consensus-based.
 
 ## What are the stable APIs of `rules_rust`?
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -50,7 +50,7 @@ features available to the current Bazel rolling release will be available to the
 next Bazel LTS release at latest.
 
 Whenever there is a new Bazel LTS release, all releases of `rules_rust` in the
-first 3 months will support both Bazel LTS versions.
+first 3 months will support both Bazel LTS versions unless Bazel doesn't allow this.
 
 ## What host platforms are supported?
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -54,7 +54,7 @@ first 3 months will support both Bazel LTS versions.
 
 ## What host platforms are supported?
 
-The only platform subject to backwards compatibility policy is `Linux` on
+The only platform subject to backwards compatibility policy is `x86_64-unknown-linux-gnu` on
 `x86_64`, as the only fully supported platform. Process for moving a best effort
 platform to a supported platform is consensus-based.
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -45,7 +45,7 @@ same. We hope red CI with the current rolling release will be rare.
 current Bazel rolling release to be be available to the current Bazel LTS
 release (because Bazel compatibility policy doesn't allow us to make that
 promise, and some new features of `rules_rust` require new Bazel features that
-are only available in Bazel rolling releases). `rules_rust` do promise that new
+are only available in Bazel rolling releases). `rules_rust` will aim that new
 features available to the current Bazel rolling release will be available to the
 next Bazel LTS release at latest.
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -60,9 +60,9 @@ first 3 months will support both Bazel LTS versions unless Bazel doesn't allow t
 
 ## What host platforms are supported?
 
-The only platform subject to backwards compatibility policy is
-`x86_64-unknown-linux-gnu`, as the only fully supported platform. Process for
-moving a best effort platform to a supported platform is consensus-based.
+Platforms subject to backwards compatibility policy are
+`x86_64-unknown-linux-gnu` and `x86_64-apple-darwin` (platforms supported by `rules_rust`).
+Process for moving a best effort platform to a supported platform is consensus-based.
 
 ## What are the stable APIs of `rules_rust`?
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -72,9 +72,11 @@ that anything directly accessible from this file is considered stable.
 `//rust/private/…` is not subject to the backwards compatibility policy. Content
 of this package is an implementation detail.
 
-`//cargo`, `//util`, `//tools`, `//test`, `//examples`, `//bindgen`, `//proto`, `//wasm_bindgen`
-and any packages not mentioned by this document are by default not subject to
-the backwards compatibility policy.
+`//cargo:cargo_build_script.bzl` is subject to the backwards compatibility policy.
+
+`//cargo`, `//util`, `//tools`, `//test`, `//examples`, `//bindgen`, `//proto`,
+`//wasm_bindgen` and any packages not mentioned by this document are by default
+not subject to the backwards compatibility policy.
 
 Experimental build settings are not subject to the backward compatibility
 policy. They should be added to `//rust:experimental.bzl`.
@@ -85,7 +87,9 @@ They should be added to `//rust:incompatible.bzl`.
 
 Bug fixes are not a breaking change by default. We'll use Common Sense (and we
 will pull in more maintainers and the community to discuss) if we see a certain
-bug fix is controversial.
+bug fix is controversial. Incompatible changes to
+`//cargo:cargo_build_script.bzl` that make `cargo_build_script` more accurately
+follow cargo's behaviour are considered bug fixes.
 
 ## How to make a backwards incompatible change?
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,0 +1,102 @@
+# `rules_rust` backwards compatibility policy
+
+This document defines the backwards compatibility policy for rules\_rust and
+defines the process for making compatibility breaking changes. Any exception to
+this process will have to be thoroughly discussed and there will have to be a
+consensus between maintainers that the exception should be granted.
+
+## What is a compatibility breaking change?
+
+A backwards compatible change has the property that a build that is green,
+correct, is using stable APIs of `rules_rust,` and is using a supported version
+of Bazel before the change is still green and correct after the change. In
+practice it means that users should not have to modify their project source
+files, their `BUILD` files, their `bzl` files, their `rust_toolchain`
+definitions, their platform definitions, or their build settings.
+
+A backwards compatible release has the property that it only contains backward
+compatible changes.
+
+`rules_rust` follow [SemVer 2.0.0](https://semver.org/). `rules_rust` promise
+that all minor version number and patch version number releases only contain
+backwards compatible changes.
+
+`rules_rust` don't make any guarantees about compatibility between a released
+version and the state of the `rules_rust` at HEAD.
+
+Backwards incompatible changes will have to follow a process in order to be
+released. Users will be given at least one release where the old behavior is
+still present but deprecated to allow smooth migration of their project to the
+new behavior.
+
+## What Bazel versions are supported?
+
+`rules_rust` support the current Bazel LTS at the time of a `rules_rust`
+release.
+
+Support for the current Bazel rolling release is on the best effort basis. If
+the CI build of `rules_rust` against the current Bazel rolling release is green,
+it has to stay green. If the build is already red (because the new Bazel rolling
+release had incompatible changes that broke `rules_rust`), it is acceptable to
+merge PRs leaving the build red as long as the reason for failure remains the
+same. We hope red CI with the current rolling release will be rare.
+
+`rules_rust` don't promise all new `rules_rust` features available to the
+current Bazel rolling release to be be available to the current Bazel LTS
+release (because Bazel compatibility policy doesn't allow us to make that
+promise, and some new features of `rules_rust` require new Bazel features that
+are only available in Bazel rolling releases). `rules_rust` do promise that new
+features available to the current Bazel rolling release will be available to the
+next Bazel LTS release at latest.
+
+Whenever there is a new Bazel LTS release, all releases of `rules_rust` in the
+first 3 months will support both Bazel LTS versions.
+
+## What host platforms are supported?
+
+The only platform subject to backwards compatibility policy is `Linux` on
+`x86_64`, as the only fully supported platform. Process for moving a best effort
+platform to a supported platform is consensus-based.
+
+## What are the stable APIs of `rules_rust`?
+
+`//rust:defs.bzl` is subject to the backwards compatibility policy. That means
+that anything directly accessible from this file is considered stable.
+
+`//rust/private/…` is not subject to the backwards compatibility policy. Content
+of this package is an implementation detail.
+
+`//cargo`, `//util`, `//tools`, `//test`, `//examples`, `//bindgen`, `//proto`
+and any packages not mentioned by this document are by default not subject to
+the backwards compatibility policy.
+
+Experimental build settings are not subject to the backward compatibility
+policy. They should be added to `//rust:experimental.bzl`.
+
+Incompatible build settings are subject to the backward compatibility policy,
+meaning the behavior of the flag cannot change in a backwards incompatible way.
+They should be added to `//rust:incompatible.bzl`.
+
+Bug fixes are not a breaking change by default. We'll use Common Sense (and we
+will pull in more maintainers and the community to discuss) if we see a certain
+bug fix is controversial.
+
+## How to make a backwards incompatible change?
+
+1. Create a GitHub issue (example:
+[Split rust\_library into smaller rules#591](https://github.com/bazelbuild/rules_rust/issues/591)).
+2. Describe the change, motivation for the change, provide migration
+instructions/tooling.
+3. Add a build setting into `//rust:incompatible.bzl` that removes the old
+behavior (whenever possible) or changes the current behavior (when just
+removing old behavior is not possible). Users should not need to manually
+flip incompatible flags. If for a specific flag they do need to manually flip
+it, then in the first backwards incompatible release we make the flag be a
+noop, and only in the second backwards incompatible release we remove the
+flag.
+4. Mention the link to the GitHub issue in error messages. Do not add a
+deprecation warning (warnings make the deprecation visible to every user
+building a project, not only to the maintainers of the project or the rules).
+5. Mention the issue in the `CHANGELOG` file.
+6. Give the community 3 months from the first release mentioning the issue until
+the flag flip to migrate.
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -29,6 +29,12 @@ released. Users will be given at least one release where the old behavior is
 still present but deprecated to allow smooth migration of their project to the
 new behavior.
 
+### Compatibility before 1.0
+
+All minor version number releases before version 1.0 can be backwards incompatible.
+Backwards incompatible changes still have to follow the the process, but minimum
+time for migration is reduced to 2 weeks.
+
 ## What Bazel versions are supported?
 
 `rules_rust` support the current Bazel LTS at the time of a `rules_rust`
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,6 +1,6 @@
 # `rules_rust` backwards compatibility policy
 
-This document defines the backwards compatibility policy for rules\_rust and
+This document defines the backwards compatibility policy for rules_rust and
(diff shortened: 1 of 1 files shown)


hlopko:
CC @philwo

hlopko:
I will not submit before @illicitonion, @dfreese, and @UebelAndre. @damienmg already approved offline.

hlopko:
Feel free to comment, propose a change, or request more information if something is unclear :)

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
+# `rules_rust` backwards compatibility policy
+
+This document defines the backwards compatibility policy for rules\_rust and
```
UebelAndre, line 3:
```suggestion
This document defines the backwards compatibility policy for rules_rust and
```

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+defines the process for making compatibility breaking changes. Any exception to
+this process will have to be thoroughly discussed and there will have to be a
+consensus between maintainers that the exception should be granted.
+
+## What is a compatibility breaking change?
+
+A backwards compatible change has the property that a build that is green,
+correct, is using stable APIs of `rules_rust,` and is using a supported version
```
UebelAndre, line 11:
```suggestion
correct, is using stable APIs of `rules_rust`, and is using a supported version
```

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+new behavior.
+
+## What Bazel versions are supported?
+
+`rules_rust` support the current Bazel LTS at the time of a `rules_rust`
+release.
+
+Support for the current Bazel rolling release is on the best effort basis. If
+the CI build of `rules_rust` against the current Bazel rolling release is green,
+it has to stay green. If the build is already red (because the new Bazel rolling
```
UebelAndre, line 39: > If the build is already red (because the new Bazel rolling release had incompatible changes that broke `rules_rust`), it is acceptable to merge PRs leaving the build red as long as the reason for failure remains the same

Just out of curiosity, is there a way to manually trigger a build so Bazel incompatibility can be independently verified by a user?

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+are only available in Bazel rolling releases). `rules_rust` do promise that new
+features available to the current Bazel rolling release will be available to the
+next Bazel LTS release at latest.
+
+Whenever there is a new Bazel LTS release, all releases of `rules_rust` in the
+first 3 months will support both Bazel LTS versions.
+
+## What host platforms are supported?
+
+The only platform subject to backwards compatibility policy is `Linux` on
```
UebelAndre, line 57:
```suggestion
The only platform subject to backwards compatibility policy is `x86_64-unknown-linux-gnu` on
```

I think this is more specific (and will be familiar to Rust folks). We don't currently support `musl` and have no way to test it, but it's also an `x86_64` Linux platform. I hope to change this soon though 😅

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+
+## What are the stable APIs of `rules_rust`?
+
+`//rust:defs.bzl` is subject to the backwards compatibility policy. That means
+that anything directly accessible from this file is considered stable.
+
+`//rust/private/…` is not subject to the backwards compatibility policy. Content
+of this package is an implementation detail.
+
+`//cargo`, `//util`, `//tools`, `//test`, `//examples`, `//bindgen`, `//proto`
```
UebelAndre, line 69:
```suggestion
`//cargo`, `//util`, `//tools`, `//test`, `//examples`, `//bindgen`, `//proto`, `//wasm_bindgen`
```

Maybe also `//docs`? And I it was decided that `//crate_universe` would not be here? Just to clarify

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+and any packages not mentioned by this document are by default not subject to
+the backwards compatibility policy.
+
+Experimental build settings are not subject to the backward compatibility
+policy. They should be added to `//rust:experimental.bzl`.
```
UebelAndre, line 74: What is an example of `//rust:experimental.bzl` and `//rust:incompatible/bzl`?

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+They should be added to `//rust:incompatible.bzl`.
+
+Bug fixes are not a breaking change by default. We'll use Common Sense (and we
+will pull in more maintainers and the community to discuss) if we see a certain
+bug fix is controversial.
+
+## How to make a backwards incompatible change?
+
+1. Create a GitHub issue (example:
+[Split rust\_library into smaller rules#591](https://github.com/bazelbuild/rules_rust/issues/591)).
```
UebelAndre, line 87:
```suggestion
[Split rust_library into smaller rules#591](https://github.com/bazelbuild/rules_rust/issues/591)).
```

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+merge PRs leaving the build red as long as the reason for failure remains the
+same. We hope red CI with the current rolling release will be rare.
+
+`rules_rust` don't promise all new `rules_rust` features available to the
+current Bazel rolling release to be be available to the current Bazel LTS
+release (because Bazel compatibility policy doesn't allow us to make that
+promise, and some new features of `rules_rust` require new Bazel features that
+are only available in Bazel rolling releases). `rules_rust` do promise that new
+features available to the current Bazel rolling release will be available to the
+next Bazel LTS release at latest.
```
illicitonion, line 50: > `rules_rust` do promise that new features available to the current Bazel rolling release will be available to the next Bazel LTS release at latest.

I'm not sure we can make this promise as-written. Bazel offer no guarantees that rolling releases will resemble the LTS that follows. Maybe soften:
```suggestion
are only available in Bazel rolling releases). `rules_rust` will aim that new
features available to the current Bazel rolling release will be available to the
next Bazel LTS release at latest.
```

COMPATIBILITY.md
UebelAndre: Maybe this is intentional? Obviously ignore if so 😄

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+
+Whenever there is a new Bazel LTS release, all releases of `rules_rust` in the
+first 3 months will support both Bazel LTS versions.
```
illicitonion, line 53:
```suggestion
first 3 months will support both Bazel LTS versions unless Bazel doesn't allow this.
```

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+correct, is using stable APIs of `rules_rust,` and is using a supported version
+of Bazel before the change is still green and correct after the change. In
+practice it means that users should not have to modify their project source
+files, their `BUILD` files, their `bzl` files, their `rust_toolchain`
+definitions, their platform definitions, or their build settings.
+
+A backwards compatible release has the property that it only contains backward
+compatible changes.
+
+`rules_rust` follow [SemVer 2.0.0](https://semver.org/). `rules_rust` promise
```
UebelAndre, line 20: > `rules_rust` promise that all minor version number and patch version number releases only contain backwards compatible changes.

Does this also apply to sub `1.0.0` releases? Is the plan to release a `1.0.0`?

UebelAndre (changes requested):

COMPATIBILITY.md
illicitonion: I would like to extend this to Darwin x86_64 - is there a reason we don't think this is practical?

COMPATIBILITY.md
illicitonion: I think we should include `//cargo:cargo_build_script.bzl` in the stable APIs, with the caveat that we may make changes to it which are technically backwards-incompatible in order to make `cargo_build_script` more accurate to `cargo`'s behaviour.

Without this, I don't think it's practical for people to build things like `cargo-raze` or `crate_universe` on top of these rules, and I think that `cargo_build_script` is entwined and special-cased enough in `rules_rust` that we can't reasonably separate it out.

COMPATIBILITY.md
```diff
@@ -0,0 +1,102 @@
...
+1. Create a GitHub issue (example:
+[Split rust\_library into smaller rules#591](https://github.com/bazelbuild/rules_rust/issues/591)).
+2. Describe the change, motivation for the change, provide migration
+instructions/tooling.
+3. Add a build setting into `//rust:incompatible.bzl` that removes the old
+behavior (whenever possible) or changes the current behavior (when just
+removing old behavior is not possible). Users should not need to manually
+flip incompatible flags. If for a specific flag they do need to manually flip
+it, then in the first backwards incompatible release we make the flag be a
+noop, and only in the second backwards incompatible release we remove the
```
illicitonion, line 95: Is the idea that the value of the flag that was the _old_ default will now error, but if it's set to the _new_ default it will no-op?

i.e. if we introduced`--incompatible_disable_borrowck=False`, the incompatible release would default `--incompatible_disable_borrowck=True` and setting `--incompatible_disable_borrowck=False` would error?

(The phrase "the flag would be a noop" suggests to me that there would be no error, and that whether you set it to `True` or `False` you'd get the same value)

COMPATIBILITY.md
illicitonion: Similarly, is there a way we can trigger CI when new rolling releases come out to notify us of upcoming failure?

illicitonion (dismissed):
Generally LGTM - thanks for putting this together!

COMPATIBILITY.md
UebelAndre: Related, why don't we support everything Bazel supports? It'd also probably be good to indicate this support is for execution environments.

COMPATIBILITY.md
hlopko: Buildkite builds can only be triggered by users who are explicitly allowed to (buildkite has "teams", and scentini@ and I are the Bazel's "rust" team there, so we can trigger builds manually). For random users this will not be allowed, but maintainers of rules_rust quite likely yes.

> Similarly, is there a way we can trigger CI when new rolling releases come out to notify us of upcoming failure?

Yes, I hope to have such pipeline.

COMPATIBILITY.md
hlopko: Oh good point, I should talk about pre-1.0 policy. In short I would make 0.X releases backwards incompatible, I'd ask breaking changes to follow the process, but maybe have shorter minimum time for migration than 3 months. WDYT?

COMPATIBILITY.md
hlopko: Let's move this to a separate issue? (In short I have no objections but at least from the start I'd like to have a champion with available capacity to fix things (for example enabling all tests on mac and windows that are currently disabled because reasons). Or making sure all features work on mac and windows just like they do on linux (can we use ld.exe to link?, can we pass env vars to test executions? that kind of stuff)

COMPATIBILITY.md
hlopko: Oh right! What do you think about moving cargo_build_script to //rust? In any case, could you suggest a wording you'd like?

COMPATIBILITY.md
UebelAndre: Yeah, until there's some concurrence on the current API in its entirety being stable, I'd say `0.X` releases are backwards incompatible and should have a short migration period. I'd love to get to a `1.0` ASAP and to do so I think we need the ability to relatively rapidly make changes. A note about the pre-`1.0` phase would be good.

COMPATIBILITY.md
hlopko: So `--incompatible_disable_borrowck` is a well behaving flag, you can migrate your repo (= stop relying on borrowck) without having to flip the flag. Because you don't have to flip it, you won't, and therefore we can remove the flag silently when it is not needed (without changing the default value of the flag).

If you had a flag like `--incompatible_use_new_orphan_rule` that forces you to either have old behavior with the flag set to false, or new behavior with flag set to true, but not new behavior with flag set to false, only then we would have to first have a release when the behavior in rules is hardcoded for the new orphan rule, but the flag still exists but does nothing (so you can simply remove it from your bazelrc). Am I making any sense? Can you think of a different wording that is clearer?

COMPATIBILITY.md
hlopko: Those files will contain `build_settings` (https://docs.bazel.build/versions/main/skylark/config.html) for experimental/incompatible flags. I expect that this file will be potentially loaded from all private bzl files we have (therefore it cannot depend on them; the content of those 2 files and their loaded bzl files has to be standalone).

I expect that people will just duplicate what is already there, and for the first examples we will have a discussion on the PR that introduces them. Do you think we should give concrete examples here in the policy?

COMPATIBILITY.md
hlopko: Added a section, please take a look :)

COMPATIBILITY.md
UebelAndre: >  Do you think we should give concrete examples here in the policy?

No, but I think it'd be good to include a template of sorts so whoever ends up needing to add the first one (and I have a feeling I know who that is 😄) knows exactly what the expectation is. I don't think I've seen any other rules implement incompatibility flags so interested to know if there's something I missed that we might be able to refer users to in the description of this PR or something.

COMPATIBILITY.md
hlopko: I created separate issues for this, let's continue there.

COMPATIBILITY.md
hlopko: (resolving now, feel free to reopen)

COMPATIBILITY.md
hlopko: Actually we can resolve moving cargo_build_script elsewhere. Added cargo_build_script into the policy, please take a look.

COMPATIBILITY.md
illicitonion: I think the bit I'm not clear on is that I'm not sure what the value is in keeping the flag around when it doesn't do anything? Why not delete it as soon as the old behaviour is no longer supported?

philwo (approved):
Thanks!

COMPATIBILITY.md
hlopko: Imagine user migrated for `incompatible_use_new_orphan_rule`, and they added the flag to their bazelrc. If we remove the flag, their build will break.

But thinking about it now, it is a bit too complicated, and the value provided is rather small. I'll remove this bit from the text.

COMPATIBILITY.md
hlopko: I'll upload a PR soon.

hlopko:
https://github.com/bazelbuild/rules_rust/pull/847 for an example of an incompatible flag.

COMPATIBILITY.md
```diff
@@ -0,0 +1,109 @@
...
+2. Describe the change, motivation for the change, provide migration
+instructions/tooling.
+3. Add a build setting into `//rust:incompatible.bzl` that removes the old
+behavior (whenever possible) or changes the current behavior (when just
+removing the old behavior is not possible). Ideally, users should not need to
+manually flip incompatible flags.
+4. Mention the link to the GitHub issue in error messages. Do not add a
+deprecation warning (warnings make the deprecation visible to every user
+building a project, not only to the maintainers of the project or the rules).
+5. Mention the issue in the `CHANGELOG` file.
```
UebelAndre, line 107: Are users expected to update this file? I'd like to avoid that requirement if possible or automate it away somehow.

COMPATIBILITY.md
```diff
@@ -0,0 +1,109 @@
...
+release.
+
+Support for the current Bazel rolling release is on the best effort basis. If
+the CI build of `rules_rust` against the current Bazel rolling release is green,
+it has to stay green. If the build is already red (because the new Bazel rolling
+release had incompatible changes that broke `rules_rust`), it is acceptable to
+merge PRs leaving the build red as long as the reason for failure remains the
+same. We hope red CI with the current rolling release will be rare.
+
+`rules_rust` don't promise all new `rules_rust` features available to the
```
UebelAndre, line 50: Is it intentional that `rules_rust` is from the perspective of self? This is a very minor thing but reads a bit funny to me.
```suggestion
`rules_rust` doesn't promise all new `rules_rust` features available to the
```

UebelAndre (changes requested):
A couple of small questions but looking really good!

COMPATIBILITY.md
hlopko: If by users you mean contributors to rules_rust who introduce incompatible changes, then yes I thought so.

If you have plans to automate the release notes generation process and maybe scrape commit descriptions for some RELNOTES tag or what not, we can always update the text of the policy. Or do you want to update the text right now?

COMPATIBILITY.md
hlopko: I just assumed `rules_rust` is plural, hence `don't`, not `doesn't`. In general I shouldn't be trusted with English.

UebelAndre: lol nor I. This is minor and I'm not at all confident which way is right. Just read funny to me but I don't see a problem with changing things like this in the future since it doesn't change the policies 😄

COMPATIBILITY.md
UebelAndre: > If by users you mean contributors to rules_rust who introduce incompatible changes, then yes I thought so.

Yeah, that's what I meant.

I suppose it's fine that users introducing those changes would need to update the `CHANGELOG`. For some reason I had it in my head that the changelog would contain "notable" changes in a release which are not limited to incompatible changes. If `CHANGELOG` requires that then I'd like to dive into the details there a bit more but if it's only required for incompatibility changes, then I'm a happy camper 😄

UebelAndre (changes requested):
I only have one more quesion-ish thing open but I like what's here and can approve once that conversation is resolved 😄

COMPATIBILITY.md
hlopko: In my mind, CHANGELOG contains both new features, notable bug fixes, new incompatible changes, or flipped incompatible changes, or anything useful. So would you like to dive into the details here in this issue?

UebelAndre: No, I can agree to making that part of the `CHANGELOG`. My only concern is about making sure that's effortlessly updated. I don't want to run into situations where a change that should have been in `CHANGELOG` wasn't added and now it's up to whoever is making a release to check that. Nor do I want users to feel like they need to add that for any change they make. A manually managed `CHANGELOG` file, in my experience, has added lots of churn across PRs and can turn into a time sink or a reason not to perform a release. Are there other projects that can more-or-less automate updating that file? Maybe something maintainers can comment on a ticket to have that updated pre-merge? This can also be a separate thread.

UebelAndre (approved):
This looks good to me, thanks for all the work you put into this @hlopko! Very much appreciated!

dfreese (approved):
I'd consider it just a nit, but otherwise LGTM.  thanks.

hlopko:
I believe this was baking for long enough. Thank you all for your contributions and review!

PR #835, opened 2021-07-14
Add support for clippy.toml

djmarcin:
This allows users to override clippy settings via the usual config file in their workspace.


rust/private/clippy.bzl
```diff
@@ -116,9 +116,14 @@ def _clippy_aspect_impl(target, ctx):
     if crate_info.is_test:
         args.add("--test")
 
+    valid_config_file_names = [".clippy.toml", "clippy.toml"]
```
UebelAndre, line 119: Does clippy care about the file name? Why can't this be arbitrary?

tools/clippy/BUILD
```diff
@@ -0,0 +1 @@
+exports_files(["clippy.toml"])
```
UebelAndre, line 1: Can you rename this file `BUILD` file to `BUILD.bazel` for consistency with the rest of the repo?

UebelAndre (changes requested):
Nice! Just a few questions but a very welcomed change! 😄

rust/private/clippy.bzl
djmarcin: Yes, clippy cares about the filename. https://github.com/rust-lang/rust-clippy/blob/cadb93b6de1b5ba5aca53c5c4c7987f3c09962a8/clippy_lints/src/utils/conf.rs#L220-L222

tools/clippy/BUILD
djmarcin: Renamed

UebelAndre (changes requested):
It looks like this was added in https://github.com/rust-lang/rust-clippy/commit/86c513e605e7982c00587e05a33a1062ca9a5334 which was first released with Rust `1.34.0`. Maybe that's something we should add to the documentation?

UebelAndre:
Additionally, can you also add a comment at the top of the default `clippy.toml`
```toml
# clippy options: https://rust-lang.github.io/rust-clippy/
```

Similar to the [rustfmt.toml](https://github.com/bazelbuild/rules_rust/blob/86a653f32535f75f0dedf432cdbe9309739e4db1/tools/rustfmt/rustfmt.toml) equivalent?

djmarcin:
> Additionally, can you also add a comment at the top of the default `clippy.toml`
> 
> ```toml
> # clippy options: https://rust-lang.github.io/rust-clippy/
> ```
> 
> Similar to the [rustfmt.toml](https://github.com/bazelbuild/rules_rust/blob/86a653f32535f75f0dedf432cdbe9309739e4db1/tools/rustfmt/rustfmt.toml) equivalent?

Done.

rust/private/clippy.bzl
```diff
@@ -116,6 +116,12 @@ def _clippy_aspect_impl(target, ctx):
...
+    if ctx.file._config.basename not in valid_config_file_names:
+        fail("The clippy config file must be named one of: %s" % valid_config_file_names)
+    env["CLIPPY_CONF_DIR"] = "${pwd}/%s" % ctx.file._config.dirname
```
UebelAndre, line 122:
```suggestion
    env["CLIPPY_CONF_DIR"] = "${pwd}/{}".format(ctx.file._config.dirname)
```

rust/private/clippy.bzl
```diff
@@ -116,6 +116,12 @@ def _clippy_aspect_impl(target, ctx):
     if crate_info.is_test:
         args.add("--test")
 
+    valid_config_file_names = [".clippy.toml", "clippy.toml"]
+    if ctx.file._config.basename not in valid_config_file_names:
+        fail("The clippy config file must be named one of: %s" % valid_config_file_names)
```
UebelAndre, line 121:
```suggestion
        fail("The clippy config file must be named one of: {}".format(valid_config_file_names))
```

UebelAndre (changes requested):
Two small nits and, while this probably isn't a fantastic example, could you also setup something similar to [//test/rustfmt/rustfmt_failure_test](https://github.com/bazelbuild/rules_rust/blob/86a653f32535f75f0dedf432cdbe9309739e4db1/test/rustfmt/rustfmt_failure_test.sh#L79) which actually uses the config? I realize the `rustfmt` case is kinda bad because it doesn't enforce that the particular setting is actually used, but at least the starlark code path would be exercised in a workspace that's not `rules_rust`.

rust/private/clippy.bzl
UebelAndre:
```suggestion
    # Clippy itself looks for specifically named files so the same
    # enforcement is applied here.
    valid_config_file_names = [".clippy.toml", "clippy.toml"]
```

test/clippy/too_many_args/clippy.toml
```diff
@@ -0,0 +1 @@
+too-many-arguments-threshold = 2
```
UebelAndre, line 1: nit: can you add a newline here?

UebelAndre (approved):
Awesome! Some tiny nits if you don't mind but otherwise LGTM!

UebelAndre (approved):

Issue #770, opened 2021-06-17
Support for custom targets


davidskidmore:
I'm having difficulty using bazel with a [custom target](https://doc.rust-lang.org/rustc/targets/custom.html) based on `x86_64-unknown-linux-gnu`. In theory, given a simple library:

```rust
#![no_std]

#[no_mangle]
pub extern "C" fn rust_function() {}
```

After [bootstrapping](https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html) libcore and friends I can now build the library for my new target:

```sh
rustc --crate-type=lib foo.rs --target x86_64-unknown-none.json \
  -L dependency=~/rust_libs/ \
  --extern 'noprelude:core=~/rust_libs/libcore-7363d577ba15c1d7.rmeta' \
  --extern 'noprelude:compiler_builtins=~/rust_libs/libcompiler_builtins-b774e5dde03423fb.rmeta' \
  -Z unstable-options
```

I'm aware of how to accomplish the cross build with cargo, but I'm not sure how to do this with bazel. There doesn't seem to be a way for me to pass my target JSON through a new `rust_toolchain` definition.

hlopko:
So AFAIU the request is to add a way to pass `target.json` to the `rust_toolchain`. That's quite a simple change, and I'd say not very controversial. @davidskidmore, do you want to give creating a PR a try?

davidskidmore:
I've tried hacking something together, but haven't been able to get something working that will both accept the new json as a target and successfully use my prebuilt libraries.

davidskidmore:
I have successfully built and linked my own libraries, and created #836 to show my approach in adding support for passing a custom target specification. I can't tell if this is the "right" way, so let me know if you have any feedback.

PR #836, opened 2021-07-14
Add support for passing a custom target specification.

davidskidmore:
A [label](https://docs.bazel.build/versions/main/skylark/lib/attr.html#label) doesn't seem to express both a string (as in a triple like `x86_64-unkown-none`) and a path to a custom target specification (as in something like `//path/to/target.json`), so this change is propose treating `target_json` as an override of `target_triple`.


UebelAndre (changes requested):
Is there a way that this code path can be tested in CI? I think the rules should definitely be supportive to custom targets but I'd like to know when a change impacts support for custom code paths. Perhaps something can be added with the existing [unit tests](https://github.com/bazelbuild/rules_rust/tree/86a653f32535f75f0dedf432cdbe9309739e4db1/test/unit)?

UebelAndre:
Also, to follow up on https://github.com/bazelbuild/rules_rust/issues/770#issuecomment-880268493

I think this approach makes sense. I would probably add some protections against trying to set both `target` and `target_json` but I like the idea of two dedicated attributes for defining the targets, especially after having looked at the [Embedonomicon](https://docs.rust-embedded.org/embedonomicon/custom-target.html). To me it seems like a natural translation in Bazel and a great addition to the toolchain. But others may know better than I.

rust/toolchain.bzl
```diff
@@ -296,6 +297,11 @@ rust_toolchain = rule(
             ),
             mandatory = True,
         ),
+        "target_json": attr.label(
```
hlopko, line 300: Is it possible you're testing this code locally without sandboxing? I don't see where the target_json file is passed to the rustc compile action, I only see the path to the file used in command line arguments. We definitely need to pass the file as an input to an action, otherwise remote execution and sandboxing won't work.

rust/toolchain.bzl
```diff
@@ -193,6 +193,7 @@ def _rust_toolchain_impl(ctx):
         rustfmt = ctx.file.rustfmt,
         cargo = ctx.file.cargo,
         clippy_driver = ctx.file.clippy_driver,
+        target_json = ctx.file.target_json,
```
hlopko, line 196: So there are multiple places where we compute the the value of the argument to the `--target` flag. How about we computed it only once here in the toolchain, and we added the string into the ToolchainInfo? I'd call it `target_flag_value`.

You'll still need a separate argument for the File instance so you can pass it as input to actions, or you can make the argument be part of the `rustc` depset. I prefer the latter.

hlopko (commented):
Thanks for working on this!

rust/toolchain.bzl
davidskidmore: I agree, I'll pass the file as a dep, probably during `collect_inputs` as a `compile_input`. I think the reason this was working is I am using a very kludgy script to package the json and use it during build time in order to work around an interesting behavior in rustc.

If you pass a custom target specification then rustc will append a hash of the absolute path to the JSON, turning something like `x86_64-unknown-none` into `x86_64-unknown-none-9187537878920592665`. This is done to differentiate specs with the same name but different paths. It also has the effect of making it impossible to compile with a prebuilt set of standard libraries. At least for me, my remote workers use temporary, nondetermistic working directories and so I have been copying my specification to `/tmp` using my kludge script.

rust/toolchain.bzl
davidskidmore: Done.

hlopko (approved):
LGTM, thank you!

hlopko:
Oh actually, +1 to @UebelAndre's comment, could you please add a check in the `toolchain.bzl` ensuring that `target_json` and `target_triple` are not both specified at the same time, and if yes, then `fail()` with a nice error message?

davidskidmore:
> Is there a way that this code path can be tested in CI? I think the rules should definitely be supportive to custom targets but I'd like to know when a change impacts support for custom code paths. Perhaps something can be added with the existing [unit tests](https://github.com/bazelbuild/rules_rust/tree/86a653f32535f75f0dedf432cdbe9309739e4db1/test/unit)?

I'm a little out of my depth here, being pretty new to writing bazel build logic. Do you have any suggestions on where the new unit tests would go?

UebelAndre:
> I'm a little out of my depth here, being pretty new to writing bazel build logic. Do you have any suggestions on where the new unit tests would go?

We have unit tests located in [test/unit](https://github.com/bazelbuild/rules_rust/tree/main/test/unit). But this one might be pretty unique since the change is to the toolchain? I think a [transition](https://docs.bazel.build/versions/main/skylark/lib/transition.html) might be needed to build a target using a newly registered toolchain which exercises this code path. Maybe someone else might have a simpler path forward though?

davidskidmore:
Sorry, but I don't really follow. I've tried throwing together a bare bones unit test:

```bazel
"""Unit tests for rust toolchains."""

load("@bazel_skylib//lib:unittest.bzl", "analysistest")
load("//rust:toolchain.bzl", "rust_stdlib_filegroup", "rust_toolchain")

def _toolchain_specifies_target_triple_test_impl(ctx):
    env = analysistest.begin(ctx)
    target_under_test = analysistest.target_under_test(env)

    # TODO: assert something

    return analysistest.end(env)

toolchain_specifies_target_triple_test = analysistest.make(_toolchain_specifies_target_triple_test_impl)

def _toolchain_test():
    rust_stdlib_filegroup(
        name = "std_libs_none",
        srcs = [],
    )

    rust_toolchain(
        name = "rust_x86_64_none_impl",
        binary_ext = "",
        dylib_ext = ".so",
        os = "linux",
        rust_lib = ":std_libs_none",
        staticlib_ext = ".a",
        stdlib_linkflags = [],
    )

    toolchain_specifies_target_triple_test(
        name = "toolchain_specifies_target_triple_test",
        target_under_test = ":rust_toolchain",
    )

def toolchain_test_suite(name):
    _toolchain_test()

    native.test_suite(
        name = name,
        tests = [":toolchain_specifies_target_triple_test"],
    )
```

But bazel isn't handling handle references to `rust_toolchain` correctly:

```
ERROR: ~/rules_rust/test/unit/toolchain/BUILD:3:21: in target_under_test attribute of toolchain_specifies_target_triple_test rule //test/unit/toolchain:toolchain_specifies_target_triple_test: rule '//test/unit/toolchain:rust_toolchain' does not exist. Since this rule was created by the macro 'toolchain_test_suite', the error might have been caused by the macro implementation
ERROR: Analysis of target '//test/unit/toolchain:toolchain_specifies_target_triple_test' failed; build aborted: Analysis of target '//test/unit/toolchain:toolchain_specifies_target_triple_test' failed
```

hlopko:
Oh wow, you actually almost did make it work. In your `target_under_test` you have `":rust_toolchain"`, but you actually want `":rust_x86_64_none_impl"`. Since you started, let's endure :)

So I guess you want to get the `ToolchainInfo` from the target under test in `_toolchain_specifies_target_triple_test_impl` and verify that the `target_flag_value` is as expected, and then add another test with a toolchain that sets `target_json` and assert different `target_flag_value`, and presence of `target_json` file in the provider. That all makes a lot of sense, thank you for the excellence!

davidskidmore:
Thanks for the advice, that was it! I was quite able to get `expect_failure` working for a unit test that could demonstrate passing both triple and json at the same time, but I think we're in a pretty good place testing wise.

UebelAndre (changes requested):
Would it be possible to rename the `x86_64-unknown-linux-gun.json` file to something like `toolchain-test-triple.json`? I feel like it'd be better to set values here that would never match naturally to confirm we're not benefiting from any coincidental assumptions. I think the cross platform tests will probably catch this but I still feel it's worth going the extra step here? Could be convinced otherwise though. Let me know what you think 😅 

Otherwise, I think this is looking awesome! Thanks for all the work you put into this 🙏

UebelAndre:
Also, docs can be fixed by running `./docs/update_docs.sh`, The error message in CI should describe the same but hoping this message is easier to find 😄

davidskidmore:
The target specifications are usually named after the target they're for, but I think it makes sense for the tests to use something more unique.

UebelAndre:
> The target specifications are usually named after the target they're for, but I think it makes sense for the tests to use something more unique.

I agree and would definitely maintain this behavior for any target I define, but yeah, tests should be able to afford the flexibility.

test/unit/toolchain/toolchain_test.bzl
```diff
@@ -0,0 +1,80 @@
...
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+load("//rust:toolchain.bzl", "rust_stdlib_filegroup", "rust_toolchain")
+
+def _toolchain_specifies_target_triple_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    toolchain_info = analysistest.target_under_test(env)[platform_common.ToolchainInfo]
+
+    asserts.equals(env, None, toolchain_info.target_json)
+    asserts.equals(env, "x86_64-unknown-linux-gnu", toolchain_info.target_flag_value)
```
UebelAndre, line 11: If these values are also being pulled from the `.json` file, can you rename them to `toolchain-test-triple` as well?

UebelAndre (changes requested):
One final nit then otherwise looks good to me!

test/unit/toolchain/toolchain_test.bzl
davidskidmore: In this case, `toolchain_specifies_target_triple_test` is referencing one of the builtin targets. That is, if you're building on Linux today you're probably targeting `x86_64-unknown-linux-gnu`. Any of the (builtin) targets that come out of `rustup target list` are valid to use as `target_triple`. I think we should keep this one as-is.

UebelAndre: Yeah, that I understand but the thing I'm worried about is if I were to make a change to this code, would it sufficiently catch regressions if I were to also have set some values somewhere that were sane looking values? I think it's good to have tests contain all obviously test values so nothing accidentally passes. By replacing a real triple with a fake one for the test, I think it'd give anyone high confidence that the tests is doing the expected behavior and would be obvious if they saw a failure about `toolchain-test-triple`. Maybe I'm just being paranoid and you actually added sufficient coverage here, what are your thoughts? 😅

davidskidmore: Since `_rust_toolchain_impl` contains no logic on the actual contents of the `target_triple` string, a fake triple will definitely work for this unit test. In practice though, you can only pass a string corresponding to a builtin target or rustc will fail to build. I'm not convinced we'll gain anything in the test with a fake triple, but I can add it if you like.

UebelAndre: I see, one thing I missed is that the `json` file is not checked (other than it's a file that exists at a path). One of the things I was hoping to see was some target that was built using a `json` file. I think the implementation of that would be tricky though... The change looks functionally correct so I'm fine with this going in but I'll leave it to @hlopko to decide if there should be that kind of integration test or not (unless you just wanted to write it 😅, but again, I'm not sure what that should look like)

davidskidmore: I'm familiar with how to write a bare metal rust binary that would benefit from this change, but I agree it's not clear how to [associate](https://docs.bazel.build/versions/main/toolchains.html) it with the toolchain/platform during a unit test. Locally I tried a unit test that uses a binary but bazel skips it, probably because of the target constraint.

```python
load("//rust:defs.bzl", "rust_binary")

native.toolchain(
    name = "rust_x86_64_none",
    exec_compatible_with = [
        "@platforms//cpu:x86_64",
        "@platforms//os:linux",
    ],
    target_compatible_with = [
        "@platforms//cpu:x86_64",
        "@platforms//os:none",
    ],
    toolchain = ":rust_json_toolchain",
    toolchain_type = "@rules_rust//rust:toolchain",
)

rust_binary(
    name = "bare",
    srcs = ["bare.rs"],
    target_compatible_with = [
        "@platforms//cpu:x86_64",
        "@platforms//os:none",
    ],
)
```

hlopko (approved):

hlopko:
LGTM, I think this is good to go.

Issue #840, opened 2021-07-16
linking a `rust_binary` that depends on two `cc_libraries` with the same name fails


krasimirgg:
This depends on https://github.com/bazelbuild/rules_rust/pull/825. In particular, trying out this example without https://github.com/bazelbuild/rules_rust/pull/825 will run into the linker errors described there and this example assumes using at least the beta `rustc`. I'm using Linux with ld.

Consider a case where a `rust_binary` (transitively) depends on two `cc_library`es that have the same name. In the wild targets with common names like `utils`, or `status` could occur in unrelated packages.

Like in this example where `//same_names:rbin` depends on `//same_names/x:exc` and `//same_names/y:exc`: 
```
same_names
├── BUILD
├── rbin.rs
├── x
│   ├── BUILD
│   └── exc.cc
└── y
    ├── BUILD
    └── exc.cc
```
```bazel
# same_names/BUILD
load("@rules_rust//rust:defs.bzl", "rust_binary")

rust_binary(
    name = "rbin",
    srcs = ["rbin.rs"],
    deps = [
        "//same_names/x:exc",
        "//same_names/y:exc",
    ],
)
```
```rust
// same_names/rbin.rs
use std::os::raw::c_int;

extern "C" {
    pub fn cx() -> c_int;
    pub fn cy() -> c_int;
}

fn main() {
    println!("hi {} {}", unsafe { cx() }, unsafe { cy() });
}
```
```bazel
# same_names/x/BUILD
package(default_visibility = ["//visibility:public"])

cc_library(
    name = "exc",
    srcs = ["exc.cc"],
)
```
```cpp
// same_names/x/exc.cc
extern "C" int cx() {
  return 17;
}
```
```bazel
# same_names/y/BUILD
package(default_visibility = ["//visibility:public"])

cc_library(
    name = "exc",
    srcs = ["exc.cc"],
)
```
```cpp
// same_names/y/exc.cc
extern "C" int cy() {
  return 113;
}
```

Running `bazel build //same_names:rbin` fails with a linker error like:
```
bin::main: error: undefined reference to 'cy'
```

[...]

PR #841, opened 2021-07-16
Use library paths instead of `-Lnative/-lstatic` when linking native dependencies

krasimirgg:
This fixes https://github.com/bazelbuild/rules_rust/issues/840 by updating the rules to use library paths instead of `-Lnative/-lstatic` when linking native dependencies.


hlopko (approved):
LGTM, thank you!

hlopko:
Oh, CI doesn't like the dylibs change :/

krasimirgg:
> Oh, CI doesn't like the dylibs change :/

I'm sorry, I tried it out, noticed the failure and then left the PR in a broken state for a while.
So, as shown by the CI, trying to use this pattern for dylibs causes tests at runtime to fail to find the dynamic libs with errors like [this one](https://buildkite.com/bazel/rules-rust-rustlang/builds/3850#d19409bc-a30a-4944-9f81-e5da5e9d06a6):
```
/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/1458/execroot/rules_rust/bazel-out/k8-fastbuild/bin/external/examples/ffi/rust_calling_c/matrix_dylib_test.launcher.runfiles/rules_rust/external/examples/ffi/rust_calling_c/matrix_dylib_test: error while loading shared libraries: bazel-out/k8-fastbuild/bin/_solib_k8/_U@examples_S_Sffi_Srust_Ucalling_Uc_Sc_Cnative_Umatrix_Uso___Uexternal_Sexamples_Sffi_Srust_Ucalling_Uc_Sc/libnative_matrix_so.so: cannot open shared object file: No such file or directory
```
I'm still interested in pursuing this change for dylibs, but I'll need to dig deeper in the dynamic linking process to figure this out. I left a TODO and will try resolve it in a follow-up PR.

aaliddell:
Has this changed the order that args get passed to the linker to produce the final executable?

When using rules_rust after this commit, I get link failures when using a `rust_binary` that transitively depends on a `cc_library` which itself has dependencies. The `cc_library` has dependencies on standard libraries, with linker args like `-lpthread` etc, but rather than the link args being interleaved between objects I instead get `-lpthread` repeated consecutively many times in the args, followed by the object `.a` paths all at the end. This causes linking to fail, as the objects from the `cc_library` need to come before their deps to resolve the symbols, since the order of linking args matters.

It appears that the link opts and the objects they come from are losing order relative to each other. e.g rather than `-lobj_a -lobj_a_dep -lobj_b -lobj_b_dep -lobj_c -lobj_c_dep` we instead get `-lobj_a_dep -lobj_b_dep -lobj_c_dep path/to/obj_a path/to/obj_b path/to/obj_c`, which breaks the required linking order.

krasimirgg:
I did not intend to change the order that args get passed to the linker, just how each native dep is passed.

In your example, how are the dependencies of the `cc_library` on standard libraries declared? If you can share a sample failing example that would be great. In my setup stuff like `-lpthread` is handled via a `rust_toolchain.stdlib_linkflags` declaration, like in this example: https://github.com/bazelbuild/rules_rust/blob/75c63c36ab6a268d4de773437400d1d853723e84/rust/toolchain.bzl#L392

Also which linker are you using? I might have instructed my linker to ignore backreferences to the standard libraries via [–warn-backrefs-exclude](https://lld.llvm.org/ELF/warn_backrefs.html) and hence missed this.

aaliddell:
So the example I found this with was pretty complex, using the `grpc-sys` crate, which wraps `grpc` C/C++ project which depends on `pthread` etc. The error I was getting was that it couldn't even find the symbol for `stat`.

A build log is linked below, from [this build](https://buildkite.com/bazel/rules-proto-grpc-rules-proto-grpc/builds/1159#22a079e5-1eb3-4198-bd70-0539596836ef). The rust specific failure is right at the bottom.

However, since this is such a weird setup, I'll see if I can create a small reproduction that doesn't depend on such big projects.

Build log: [rules-proto-grpc-rules-proto-grpc_build_1159_build-and-test-all-on-ubuntu-18-dot-04-openjdk-11.log](https://github.com/bazelbuild/rules_rust/files/7221076/rules-proto-grpc-rules-proto-grpc_build_1159_build-and-test-all-on-ubuntu-18-dot-04-openjdk-11.log)

bsilver8192:
+1 for this breaking builds using `ld`. My situation has `-pthread` and `-lstdc++` coming from the C++ toolchain, and I'm using [cxx](https://github.com/dtolnay/cxx) via [autocxx](https://github.com/google/autocxx). That means I have this dependency order:
1. My Rust code
2. Generated C++ wrappers
3. C++ code from `cxx` (this might be redundant if my Rust->C++ API was more complicated and causes the former item to rely directly on the latter item)
4. Rust code from `cxx`

This means C++ code must be sandwiched between Rust code in the link order, or the whole thing has to be wrapped in `--start-lib`/`--end-lib` or `--start-group`/`--end-group`. I think this is what `rustc` effectively does with this PR reverted, which I confirmed makes my example build.

I don't see a way to fix it with the approach used by this PR: any way of passing a generic linker argument through `rustc` is going to put the files either before or after all the Rust libraries. If the .a filenames are a problem, maybe generating a folder with them all symlinked/copied with sanitized filenames would work?

You can see my problematic toolchain at https://github.com/frc971/971-Robot-Code/blob/c34b3ea019461766d70b43b41f509faa6b8001f3/tools/cpp/BUILD#L122. I can work on pulling out a smaller example if that would help, but I think the CI build referenced in https://github.com/dtolnay/cxx/pull/947 is a more convenient example.

For reference, I've got some other `lld`-based toolchain which do similar things with linker arguments at https://github.com/frc971/971-Robot-Code/tree/1e7c897dc01b85ce773c5334a728057621a486b5/third_party/bazel-toolchain (based on https://github.com/grailbio/bazel-toolchain with some changes I haven't upstreamed yet) which do work with this PR.

krasimirgg:
> +1 for this breaking builds using `ld`

Thank you for the detailed report! I'm looking into this.

bsilver8192:
Adding a note here for anybody from the future: #1148 (and related PRs) fix this for me.

PR #847, opened 2021-07-21
Add incompatible flag for splitting rust_library

hlopko:
This PR adds some elementary infrastructure for incompatible flags.

### DIFF
--- a/rust/settings/incompatible.bzl
+++ b/rust/settings/incompatible.bzl
@@ -1,0 +1,39 @@
+"""This file contains definitions of all current incompatible flags.
+
+See COMPATIBILITY.md for the backwards compatibility policy.
+"""
+
+IncompatibleFlagInfo = provider(
+    doc = "Set the --error-format flag for all rustc invocations",
+    fields = {
+        "enabled": "(bool) whether the flag is enabled",
+        "issue": "(string) link to the github issue associated with this flag",
+    },
+)
+
+def _incompatible_flag_impl(ctx):
+    return [IncompatibleFlagInfo(enabled = ctx.build_setting_value, issue = ctx.attr.issue)]
+
+incompatible_flag = rule(
+    doc = "A rule defining an incompatible flag.",
+    implementation = _incompatible_flag_impl,
+    build_setting = config.bool(flag = True),
+    attrs = {
+        "issue": attr.string(),
+    },
+)
+
+def _fail_when_enabled_impl(ctx):
+    flag = ctx.attr.flag
+    flag_info = getattr(ctx.attr, "_" + flag)[IncompatibleFlagInfo]
+    if flag_info.enabled:
+        fail("Incompatible flag {} has been flipped, see {} for details.".format(flag, flag_info.issue))
+
+fail_when_enabled = rule(
+    doc = "A rule that will fail analysis when a flag is enabled.",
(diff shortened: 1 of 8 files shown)


hlopko:
Unsurprisingly, I realized the text of compatibility policy is incorrect. I'll update it once this PR is reviewed, once we settle down on things.

rust/settings/incompatible.bzl
```diff
@@ -0,0 +1,39 @@
+"""This file contains definitions of all current incompatible flags.
+
+See COMPATIBILITY.md for the backwards compatibility policy.
+"""
+
+IncompatibleFlagInfo = provider(
+    doc = "Set the --error-format flag for all rustc invocations",
```
UebelAndre, line 7: Some copy/pasta?

rust/settings/incompatible.bzl
```diff
@@ -0,0 +1,39 @@
...
+    flag = ctx.attr.flag
+    flag_info = getattr(ctx.attr, "_" + flag)[IncompatibleFlagInfo]
+    if flag_info.enabled:
+        fail("Incompatible flag {} has been flipped, see {} for details.".format(flag, flag_info.issue))
+
+fail_when_enabled = rule(
+    doc = "A rule that will fail analysis when a flag is enabled.",
+    implementation = _fail_when_enabled_impl,
+    attrs = {
+        "flag": attr.string(),
```
UebelAndre, line 36: I'm bad with words but maybe something like this could be added? Just to have each public attribute documented.
```suggestion
        "flag": attr.string(
            doc = "The build flag to check",
            mandatory = True,
        ),
```

rust/settings/incompatible.bzl
```diff
@@ -0,0 +1,39 @@
...
+
+def _incompatible_flag_impl(ctx):
+    return [IncompatibleFlagInfo(enabled = ctx.build_setting_value, issue = ctx.attr.issue)]
+
+incompatible_flag = rule(
+    doc = "A rule defining an incompatible flag.",
+    implementation = _incompatible_flag_impl,
+    build_setting = config.bool(flag = True),
+    attrs = {
+        "issue": attr.string(),
```
UebelAndre, line 22: Should this section also be mandatory?
```suggestion
        "issue": attr.string(
            doc = "The link to the GitHub issue associated with this flag",
        ),
```

UebelAndre (changes requested):
The new rules will also need to be added to `docs/symbols.bzl` for them to show up in docs... lame to have this duplication, I know...

Otherwise, some small things but is looking good so far!

rust/settings/incompatible.bzl
hlopko: Good point, done.

rust/settings/incompatible.bzl
hlopko: Done.

rust/settings/incompatible.bzl
hlopko: /o\

hlopko (commented):
> The new rules will also need to be added to docs/symbols.bzl for them to show up in docs... lame to have this duplication, I know...

Done. Please take another look :)

UebelAndre (approved):
Thanks! Looks good to me! Just needs to look good to buildifier now 😄

PR #264, opened 2019-11-03
Propagate linkopts from native deps

hlopko:
This PR ensures that `linkopts` attributes from dependencies such as `cc_library` make it to the `rust_binary` linker invocation.

Not-so-great but still motivating example is that with this PR we can make openssl-sys crate work with system installed openssl just by pushing additional_deps `cc_libraries`:

```
cc_library(
  name = "crypto",
  linkopts = ["-lcrypto"],
)

cc_library(
  name = "ssl",
  linkopts = ["-lssl"],
  deps = [":crypt"],
)
```

I updated skylib and made use of its unittest.bzl framework for adding the unittest.

And since I was there, I added one TODO regarding libraries to link ordering.

Issue #78, opened 2018-03-23
Rust link arguments are not passed transitively to dependant rust targets


SirVer:
Simple repro case is [in this branch](https://github.com/bazelbuild/rules_rust/compare/master...SirVer:link_example?expand=1).

`bazel run @examples//hello_world` exposes the problem.

The example is a simple dependency chain: `rust_binary` ("hello_world") depends on `rust_library` ("hello_from_c_sys") which depends on `cc_library` ("hello_from_c"). The c library requires c++, therefore needs linking with `-lstdc++`. It should be sufficient to specify this in the `rust_library` rule, but right now all `rust_binary` rules that transitively depend on the lib need to specify it as well: https://github.com/bazelbuild/rules_rust/compare/master...SirVer:link_example?expand=1#diff-e85472ad14a7cde9a8510d9beae332c5R18

mfarrugi:
Afaiu #63 handles the build step on osx, but doesn't correctly set the rpath. 

You can try commenting out compute_rpaths from the pr branch (I don't have a mac to test on) to get it to build, then using LD_LIBRARY_PATH to get it to run. 

At this point we should be asking @mhlopko for advice on not reinventing cross platform dylib linking, in the context of the cc Skylark project.

(I have no idea if collecting link args is a generally correct thing to do.)

SirVer:
@mfarrugi Just to clarify, the posted bug also doesn't link on Linux for me without commenting in the `"-Clink-args=-lstdc++"` in the `rust_binary`.

mfarrugi:
On the PR branch? If so, I'll take a look as transitive dylibs ought to be handled.

SirVer:
@mfarrugi Yes, the branch linked in OP: https://github.com/bazelbuild/rules_rust/compare/master...SirVer:link_example?expand=1#diff-e85472ad14a7cde9a8510d9beae332c5R18

acmcarther:
Do you happen to know off hand what Cargo's current treatment of `Clink-args` is? If you don't, I can whip up a mock crate and verbose build it to see. I could easily believe that this is an oversight on our (rules_rust) part.

If it is indeed an oversight, we'll probably need to promote this flag to a rule-level attr so that the transitives can be tracked.

mfarrugi:
[...]

PR #849, opened 2021-07-22
propagate linkopts from (transitive) native dependencies

krasimirgg:
This revives https://github.com/bazelbuild/rules_rust/pull/264 and is aiming to address https://github.com/bazelbuild/rules_rust/issues/78. A lot has changed since the original PR in a good way: currently, I believe we don't need to do anything special about non-direct transitive deps; we just have to iterate over all direct `deps`' `CcInfo`s and collect their `linkopts` since they include merged `CcInfo`s from their transitive deps.

### DIFF
--- a/test/unit/native_deps/native_deps_test.bzl
+++ b/test/unit/native_deps/native_deps_test.bzl
@@ -3,7 +3,7 @@
 load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
 load("@rules_cc//cc:defs.bzl", "cc_library")
 load("//rust:defs.bzl", "rust_binary", "rust_library", "rust_proc_macro", "rust_shared_library", "rust_static_library")
-load("//test/unit:common.bzl", "assert_argv_contains", "assert_argv_contains_not", "assert_argv_contains_prefix_suffix")
+load("//test/unit:common.bzl", "assert_argv_contains", "assert_argv_contains_not", "assert_argv_contains_prefix_suffix", "assert_list_contains_adjacent_elements")
 
 def _native_dep_lib_name(ctx):
     if ctx.target_platform_has_constraint(ctx.attr._windows_constraint[platform_common.ConstraintValueInfo]):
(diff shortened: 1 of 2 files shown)


test/unit/native_deps/native_deps_test.bzl
```diff
@@ -243,13 +254,56 @@ def _native_dep_test():
...
 
+def _linkopts_propagate_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    actions = analysistest.target_actions(env)
+    action = actions[0]
+
+    # Ensure linkopts from direct (-Llinkoptdep1) and transitive
+    # (-Llinkoptdep2) dependencies are propagated.
+    linkopt_1_found = False
```
hlopko, line 265: Could you also assert the order in which linkopts appear? You maybe can use test/unit/common.bzl `assert_list_contains_adjacent_elements`?

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -243,13 +254,56 @@ def _native_dep_test():
         target_under_test = ":cdylib_has_native_dep_and_alwayslink",
     )
 
+def _linkopts_propagate_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    actions = analysistest.target_actions(env)
```
hlopko, line 260: Nit: pls use tut.actions instead of `taget_actions`, I believe that's the current recommended style.

hlopko (commented):
Only nits, LGTM otherwise. Thank you!

test/unit/native_deps/native_deps_test.bzl
krasimirgg: Thank you! I'll also update the rest of unit tests to use this style in a follow-up PR.

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -243,13 +254,62 @@ def _native_dep_test():
         target_under_test = ":cdylib_has_native_dep_and_alwayslink",
     )
 
+def _linkopts_propagate_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    action = tut.actions[0]
+
+    # Ensure linkopts from direct (-Llinkoptdep1) and transitive
```
hlopko, line 262: Kudos ;)

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -243,13 +254,62 @@ def _native_dep_test():
...
+    # (-Llinkoptdep2) dependencies are propagated.
+    # Consistently with cc rules, dependency linkopts take precedence over
```
hlopko, line 264: Kudos ;)

hlopko (approved):

PR #850, opened 2021-07-22
Internal macro `construct_arguments` now generates separate Args sets

UebelAndre:
`construct_arguments` now generates separate sets of arguments, one for the process wrapper, another to launch `rustc` from the process wrapper, and a final set for Rust flags. This allows rules like clippy to pass flags to the process wrapper and not have this function need to be aware of clippy (making the rules more contained).

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -329,7 +329,6 @@ def construct_arguments(
         out_dir,
         build_env_files,
         build_flags_files,
-        maker_path = None,
         emit = ["dep-info", "link"]):
     """Builds an Args object containing common rustc flags
 
@@ -349,12 +348,16 @@ def construct_arguments(
         out_dir (str): The path to the output directory for the target Crate.
         build_env_files (list): Files containing rustc environment variables, for instance from `cargo_build_script` actions.
         build_flags_files (list): The output files of a `cargo_build_script` actions containing rustc build flags
-        maker_path (File): An optional clippy marker file
         emit (list): Values for the --emit flag to rustc.
 
     Returns:
         tuple: A tuple of the following items
-            - (Args): An Args object with common Rust flags
+            - (struct): A struct of arguments used to run the `Rustc` action
+                - process_wrapper_flags (Args): Arguments for the process wrapper
+                - rustc_path (Args): Arguments for invoking rustc via the process wrapper
+                - rustc_flags (Args): Rust flags for the Rust compiler
+                - all (list): A list of all `Args` objects in the order listed above.
+                    This is to be passed to the `arguments` parameter of actions
             - (dict): Common rustc environment variables
     """
     output_dir = getattr(crate_info.output, "dirname") if hasattr(crate_info.output, "dirname") else None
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
I am a little surprised this works, but it seems pretty reasonable considering it does! :)

jfgoog (commented):
This seems like a reasonable abstraction, and makes the handling of --touch-file cleaner.

But I don't see how this helps with _error_format, which I think is the bigger problem for me in https://github.com/bazelbuild/rules_rust/pull/848

You can probably also do this much more simply, and without having to modify all the callers of construct_arguments, if you just prepend --touch-file in clippy.bzl:

```
args = ["--touch-file", clippy_marker.path, args]
```

I appreciate your help on this, but I think a complicating factor is that aspect-on-aspect handling in Bazel just doesn't work right at the moment, so it may be best for me to hold off on this until that's been fixed, because I am basically unable to test my changes right now.

UebelAndre:
> You can probably also do this much more simply, and without having to modify all the callers of construct_arguments, if you just prepend --touch-file in clippy.bzl:
> 
> ```
> args = ["--touch-file", clippy_marker.path, args]
> ```

True, but there's only two callers so far and this is a method in a private package so I feel like the impact is minimal.

> But I don't see how this helps with _error_format, which I think is the bigger problem for me in #848

> I appreciate your help on this, but I think a complicating factor is that aspect-on-aspect handling in Bazel just doesn't work right at the moment, so it may be best for me to hold off on this until that's been fixed, because I am basically unable to test my changes right now.

Is there an issue tracking what's broken? Also, don't mean to ping pong you but gonna continue this thread in https://github.com/bazelbuild/rules_rust/pull/848#discussion_r674059185 😅

UebelAndre:
Testing for this would be covered by https://github.com/bazelbuild/rules_rust/pull/851

rust/private/rustc.bzl
```diff
@@ -341,12 +340,16 @@ def construct_arguments(
...
 
     Returns:
         tuple: A tuple of the following items
-            - (Args): An Args object with common Rust flags
+            - (struct): A struct of arguments used to run the `Rustc` action
+                - process_wrapper (Args): Arguments for the process wrapper
+                - launcher (Args): Arguments for invoking rustc via the process wrapper
+                - rustc (Args): Rust flags for the Rust compiler
+                - all (list): A list of all `Args` objects in the order listed above.
+                    This is to be passed to the `arguments` parameter of actions
```
hlopko, line 352: I wonder if there is naming that will show that the first and the third args contain command line flags, and launcher contains pretty much a path to the rustc binary?

What do you think about:

* process_wrapper_flags (or _opts, probably not _args as it is loaded term)
* rustc_path (launcher is also a bit loaded term, in java/python rules the launcher is a binary that starts the runtime and bundles all the code in a single file)
* rustc_flags (or _opts)

hlopko:
Overall I like this change a lot. Thank you Andre!

rust/private/rustc.bzl
UebelAndre: Updated 😄 

I, kinda arbitrarily, went with `_flags` but I wonder if there could there be a vocabulary/language guide made for Bazel so decisions like this are clearer across various rules?

PR #851, opened 2021-07-22
Added unit tests for `rust_clippy_aspect`

UebelAndre:
This should ensure the extra flags to the clippy action are always set and the marker is always created.

### DIFF
--- a/test/unit/clippy/clippy_test.bzl
+++ b/test/unit/clippy/clippy_test.bzl
@@ -1,0 +1,112 @@
+"""Unittest to verify properties of clippy rules"""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest")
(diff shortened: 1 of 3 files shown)


test/unit/clippy/clippy_test.bzl
```diff
@@ -0,0 +1,112 @@
...
+    # Ensure each flag is present in the clippy action
+    for flag in flags:
+        assert_argv_contains(
+            env,
+            clippy_action,
+            flag,
+        )
+
+    clippy_checks = target[OutputGroupInfo].clippy_checks.to_list()
+    if len(clippy_checks) > 1:
```
illicitonion, line 31: Worth handling it being 0 as well?

illicitonion (approved):
Thanks!

PR #855, opened 2021-07-23
ensures rust binary tools are run in an exec configuration

benbrittain:
When a tool is needed to help build the target, the corresponding attribute should specify a transition to an exec configuration. This enables the rules to work in certain cross-compilation environments.

related documentation: https://docs.bazel.build/versions/main/skylark/rules.html#configurations


UebelAndre (approved):
Looks good to me, thanks!

hlopko:
Shouldn't we also mark `rustc_lib` as 'exec'?

UebelAndre:
> Shouldn't we also mark `rustc_lib` as 'exec'?

Yeah, that'd be a good thing to do. My bad for missing that: https://github.com/bazelbuild/rules_rust/pull/856

PR #857, opened 2021-07-26
Split Rust repositories into separate repos per static asset.

UebelAndre:


UebelAndre:
This draft is additional exploration of more solutions to https://github.com/bazelbuild/rules_rust/discussions/815

I thought I had something close but alas, I see many build failures 😅

PR #858, opened 2021-07-28
propagate additional inputs from native dependencies

krasimirgg:
Pass [additional inputs from native dependencies](https://docs.bazel.build/versions/3.3.0/skylark/lib/LinkerInput.html#additional_inputs) (e.g., linker scripts) as inputs to the compile action.


rust/private/rustc.bzl
```diff
@@ -290,6 +290,12 @@ def collect_inputs(
 
     linker_depset = cc_toolchain.all_files
 
+    additional_transitive_inputs = depset([
+        additional_input
+        for linker_input in dep_info.transitive_noncrates.to_list()
```
hlopko, line 295: This is introducing quadratic behavior, transitive noncrates are now flattened on each rust_library transitively depending on the cc library.

Because of https://github.com/bazelbuild/bazel/issues/13132, and the current design of CcInfo, we cannot really avoid flattening the depset at some points. WDYT about:

* Flattening and constructing "expanded" additional_inputs depset (and propagating it through CrateInfos) when we collect_deps from CcInfo
* or only add additional inputs when you know this is a linking-like action (so if the only output is rlib, we don't add them)

hlopko: Hmm linkopts are a bit relevant here, if we don't add additional inputs, we cannot add linkopts as well.

krasimirgg: Updated to only add the additional inputs for linking-like actions (crate_type `bin`).

rust/private/rustc.bzl
```diff
@@ -268,6 +268,7 @@ def collect_inputs(
         files,
         toolchain,
         cc_toolchain,
+        crate_type,
```
UebelAndre, line 271: Is `crate_type` not already a part of `crate_info`?
https://github.com/bazelbuild/rules_rust/blob/ce0291cb5ffa5b2ba27820e225d03f00a51ec1d5/rust/private/providers.bzl#L31

Why does this need to explicitly be added?

UebelAndre (dismissed):
I had one question but think @hlopko would probably be the better reviewer here.

rust/private/rustc.bzl
krasimirgg: Chatted with @hlopko, and it seems currently indeed `crate_type` is the same as `crate_info.type`, hence redundant. For some historical context, we think that [crate_info.wrapped_crate_type](https://github.com/bazelbuild/rules_rust/blob/e10ab210f34799aa6bffd7a0f11f587967be9587/rust/private/providers.bzl#L32) obsoleted the need for the crate_type arg. I'll update this PR to use crate_info.type and send a follow-up PR to update the rest of the functionality in rules_rust to use `crate_info.type` instead of passing an additional `crate_type` argument, where it makes sense.

rust/private/rustc.bzl
```diff
@@ -290,6 +290,18 @@ def collect_inputs(
 
     linker_depset = cc_toolchain.all_files
 
+    # Pass linker additional inputs (e.g., linker scripts) only for linking-like
+    # actions, not for example where the output is rlib. This avoids quadratic
+    # behavior where transitive noncrates are flattened on each transitive
+    # rust_library dependency.
+    additional_transitive_inputs = []
+    if crate_info.type == "bin":
```
hlopko, line 298: I think you also want to flatten when crate type is "cdylib" and "dylib"

hlopko (commented):
Looks good, thank you! One final comment.

rust/private/rustc.bzl
krasimirgg: Sounds good! Added both.

hlopko (approved):

PR #860, opened 2021-07-29
Enabled `//test/load_arbitrary_tool/...` tests for windows

UebelAndre:


illicitonion (approved):
Works for me - I suspect at some point we _may_ want an alias on Windows so that `//:bin/cargo` points to `//:bin/cargo.exe`, but 🤷

UebelAndre:
> Works for me - I suspect at some point we _may_ want an alias on Windows so that `//:bin/cargo` points to `//:bin/cargo.exe`, but 🤷

We do have that: 
https://github.com/bazelbuild/rules_rust/blob/ce0291cb5ffa5b2ba27820e225d03f00a51ec1d5/rust/private/repository_utils.bzl#L65-L72

I suppose I could update this test to use the same BUILD file but I liked the encapsulation here and don't think it matters too much if we use a select here or not. I also kinda like that it checks the file directly. I think it's clearer over all.

PR #864, opened 2021-07-30
cleanup: replace crate_type argument with crate_info.type

krasimirgg:
This removes the `crate_type` argument in internal implementations, replacing uses with `crate_info.type` where appropriate, as a followup to https://github.com/bazelbuild/rules_rust/pull/858#discussion_r679825635.


hlopko (approved):
Thank you for this!

PR #866, opened 2021-07-30
Expose `DepInfo` provider

tiziano88:
This is useful when developing rules that need to collect Rust dependencies from existing targets.

### DIFF
--- a/rust/rust_common.bzl
+++ b/rust/rust_common.bzl
@@ -14,6 +14,7 @@
 
 """Module with Rust definitions required to write custom Rust rules."""
 
-load("//rust/private:providers.bzl", _CrateInfo = "CrateInfo")
+load("//rust/private:providers.bzl", _CrateInfo = "CrateInfo", _DepInfo = "DepInfo")
 
 CrateInfo = _CrateInfo
+DepInfo = _DepInfo


tiziano88:
We have a use case already internally in http://cl/387770348 , which is currently blocked on this, but feel free to suggest an alternative approach if this is not viable. Thanks!

UebelAndre:
Just FYI: This should already be exposed by [@rules_rust//rust:defs.bzl%rust_common](https://github.com/bazelbuild/rules_rust/blob/c2f5701dad159c3ce2b5997e6599445b1d21a123/rust/defs.bzl#L104-L105)
https://github.com/bazelbuild/rules_rust/blob/c2f5701dad159c3ce2b5997e6599445b1d21a123/rust/private/common.bzl#L44-L49

Does using `rust_common.dep_info` work for your use case?

tiziano88:
Thanks, let me try that again; when I tried first, buildifier was complaining about the dependency on a private file, but maybe it's just a linter I can disable, not a hard rule.

UebelAndre:
You wanna load `rust_common` from `//rust:sets.bzl` then you should have no complaints. I hope 😃

tiziano88:
Thanks, that worked, I had to follow the chain of imports and re-exports all the way through. There must be a better way to figure this stuff out though :)

PR #868, opened 2021-07-30
Update `//proto` rules to provide all required dependencies.

UebelAndre:
Not all dependencies are currently provided by the `@rules_rust//proto` modules. This PR updates the proto rules to provide these missing dependencies via a new `rules_proto_transitive_repositories` macros.

The setup instructions have been updated accordingly in the generated docs.

This change may contain a breaking change for some users who've only loaded `rules_proto_repositories` and not called `@com_google_protobuf//:protobuf_deps.bzl%protobuf_deps` in their workspace. The fix for this is to simply call `rules_proto_transitive_repositories` as documented in the updated docs.

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -8,6 +8,10 @@ load("@rules_rust//proto:repositories.bzl", "rust_proto_repositories")
 
 rust_proto_repositories()
 
+load("@rules_rust//proto:transitive_repositories.bzl", "rust_proto_transitive_repositories")
+
+rust_proto_transitive_repositories()
+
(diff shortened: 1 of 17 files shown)


UebelAndre:
I've opened the following PRs to pair with the patches used to generate docs:
- https://github.com/bazelbuild/rules_proto/pull/94
- https://github.com/protocolbuffers/protobuf/pull/8847

UebelAndre:
This PR requires the following PRs:
- https://github.com/bazelbuild/rules_rust/pull/873

illicitonion (approved):
Seems reasonable :)

PR #870, opened 2021-07-31
Reduce rustc flag duplication

UebelAndre:
A follow up to https://github.com/bazelbuild/rules_rust/pull/849, I noticed in a build that there were a bunch of duplicate rustc flags
```
(cd /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/87f8cc6ac2ea5ec14312b4b2ac21c9fc/sandbox/linux-sandbox/189/execroot/rules_rust_examples_crate_universe && \
  exec env - \
    CARGO_CFG_TARGET_ARCH=x86_64 \
    CARGO_CFG_TARGET_OS=linux \
    CARGO_CRATE_NAME=h2 \
    CARGO_MANIFEST_DIR='${pwd}/external/has_aliased_deps_deps__h2__0_2_7' \
    CARGO_PKG_AUTHORS='' \
    CARGO_PKG_DESCRIPTION='' \
    CARGO_PKG_HOMEPAGE='' \
    CARGO_PKG_NAME=h2 \
    CARGO_PKG_VERSION=0.2.7 \
    CARGO_PKG_VERSION_MAJOR=0 \
    CARGO_PKG_VERSION_MINOR=2 \
    CARGO_PKG_VERSION_PATCH=7 \
    CARGO_PKG_VERSION_PRE='' \
    SYSROOT=bazel-out/k8-fastbuild/bin/external/rules_rust/rust/toolchain/current \
  bazel-out/k8-opt-exec-2B5CBBC6/bin/external/rules_rust/util/process_wrapper/process_wrapper --subst 'pwd=${pwd}' -- bazel-out/k8-fastbuild/bin/external/rules_rust/rust/toolchain/current/bin/rustc external/has_aliased_deps_deps__h2__0_2_7/src/lib.rs '--crate-name=h2' '--crate-type=rlib' '--error-format=human' '--codegen=metadata=-680607353' '--out-dir=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__h2__0_2_7' '--codegen=extra-filename=-680607353' '--codegen=opt-level=0' '--codegen=debuginfo=0' '--remap-path-prefix=${pwd}=.' '--emit=dep-info,link' '--color=always' '--target=x86_64-unknown-linux-gnu' -L bazel-out/k8-fastbuild/bin/external/rules_rust/rust/toolchain/current/lib/rustlib/x86_64-unknown-linux-gnu/lib -L bazel-out/k8-fastbuild/bin/external/rules_rust/rust/toolchain/current/lib -L bazel-out/k8-fastbuild/bin/external/rules_rust/rust/toolchain/current/lib/rustlib/x86_64-unknown-linux-gnu/bin '--cap-lints=allow' '--edition=2018' '--codegen=linker=/usr/bin/gcc' --codegen 'link-args=-fuse-ld=gold -Wl,-no-as-needed -Wl,-z,relro,-z,now -B/usr/bin -pass-exit-codes -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -ldl -lpthread -lstdc++ -lm' --extern 'bytes=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__bytes__0_5_6/libbytes--225745396.rlib' --extern 'fnv=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__fnv__1_0_7/libfnv--351998681.rlib' --extern 'futures_core=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_core__0_3_16/libfutures_core-938126416.rlib' --extern 'futures_sink=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_sink__0_3_16/libfutures_sink--280424228.rlib' --extern 'futures_util=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_util__0_3_16/libfutures_util--200256755.rlib' --extern 'http=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__http__0_2_4/libhttp-2068069658.rlib' --extern 'indexmap=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__indexmap__1_7_0/libindexmap-1829724378.rlib' --extern 'slab=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__slab__0_4_3/libslab-703676971.rlib' --extern 'tokio=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tokio__0_2_25/libtokio-214378357.rlib' --extern 'tokio_util=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tokio_util__0_3_1/libtokio_util-1949648639.rlib' --extern 'tracing=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tracing__0_1_26/libtracing-425662279.rlib' --extern 'tracing_futures=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tracing_futures__0_2_5/libtracing_futures-106518904.rlib' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__bytes__0_5_6' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__fnv__1_0_7' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_core__0_3_16' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_sink__0_3_16' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_io__0_3_16' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_task__0_3_16' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__memchr__2_4_0' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__pin_project_lite__0_2_7' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__pin_utils__0_1_0' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__slab__0_4_3' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__futures_util__0_3_16' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__bytes__1_0_1' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__itoa__0_4_7' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__http__0_2_4' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__hashbrown__0_11_2' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__indexmap__1_7_0' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__libc__0_2_98' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__iovec__0_1_4' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__lazy_static__1_4_0' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__cfg_if__0_1_10' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__cfg_if__1_0_0' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__log__0_4_14' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__net2__0_2_37' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__mio__0_6_23' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__num_cpus__1_13_0' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__pin_project_lite__0_1_12' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tokio__0_2_25' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tokio_util__0_3_1' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tracing_core__0_1_18' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tracing__0_1_26' '-Ldependency=bazel-out/k8-opt-exec-2B5CBBC6/bin/external/has_aliased_deps_deps__unicode_xid__0_2_2' '-Ldependency=bazel-out/k8-opt-exec-2B5CBBC6/bin/external/has_aliased_deps_deps__proc_macro2__1_0_28' '-Ldependency=bazel-out/k8-opt-exec-2B5CBBC6/bin/external/has_aliased_deps_deps__quote__1_0_9' '-Ldependency=bazel-out/k8-opt-exec-2B5CBBC6/bin/external/has_aliased_deps_deps__syn__1_0_74' '-Ldependency=bazel-out/k8-opt-exec-2B5CBBC6/bin/external/has_aliased_deps_deps__pin_project_internal__1_0_8' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__pin_project__1_0_8' '-Ldependency=bazel-out/k8-fastbuild/bin/external/has_aliased_deps_deps__tracing_futures__0_2_5' --sysroot '${pwd}/bazel-out/k8-fastbuild/bin/external/rules_rust/rust/toolchain/current')
```

(note the repeated `-ldl -lpthread -ldl -lpthread -ldl -lpthread`)

I'm not sure if this is the right solution here but it does feel like something is incorrectly being passed multiple times or not being collapsed.

This solves for the duplication.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -214,7 +214,7 @@ def get_linker_and_args(ctx, cc_toolchain, feature_configuration, rpaths):
         cc_toolchain = cc_toolchain,
         is_linking_dynamic_library = False,
         runtime_library_search_directories = rpaths,
-        user_link_flags = user_link_flags,
+        user_link_flags = depset(user_link_flags).to_list(),
     )
     link_args = cc_common.get_memory_inefficient_command_line(
         feature_configuration = feature_configuration,


UebelAndre:
cc @hlopko  @krasimirgg
Do either of you have any ideas?

krasimirgg:
Nice catch!

Just to clarify: I observe these repeated `-ldl -lpthread`-s also before #849. I think the duplication is a result from this: we add `toolchain.stdlib_linkflags` (that tend to include `-ldl -lpthread`) as `user_link_flags` to each `rust_library`'s `CcInfo`:
https://github.com/bazelbuild/rules_rust/blob/c2f5701dad159c3ce2b5997e6599445b1d21a123/rust/private/rustc.bzl#L692

The deduplication in this PR looks good to me.

hlopko:
We unfortunately cannot do this. We expect duplicated flags on the linker command lines (`-Wl,--start-group`, `-Wl,--whole-archive`, `-Wl,--start-lib`...) and by deduplicating them you change linking behavior. For example I found this in one of our projects:

```
linkopts = [
    "-Wl,--whole-archive",
    "-lpthread",
    "-Wl,--no-whole-archive",
]
```

If there was some other target that had similar linkopts, deduplicating would result in failed build.

The principled fix to this general problem is to not use linkopts for linking system installed libraries such as pthread, and instead use a target for them. Using a target would do the "deduplication" for us on the Bazel's dependency graph level. https://docs.bazel.build/versions/main/be/c-cpp.html#cc_import was supposed to be the rule to use for this, but I'm not sure if it allows linkopts-only, it used to require a file artifact (so it can easily detect modifications to that artifact, whereas using linkopts only is problematic because if the system got an update, bazel wouldn't notice).

In any case, I'm afraid we cannot accept this fix.

UebelAndre:
https://github.com/bazelbuild/rules_rust/pull/870#issuecomment-891577561

That all makes sense. This PR wasn't really a desire for this specific implementation vs something to identify an issue and potential fix which I definitely felt was questionable 😄 

I do think there's probably some updates to be made to the default linker flags here since there seem to have been quite a few changes to how they're set in Rust itself. Maybe one day someone can try removing them or do a pass to update them. Though, that then raises the question about versioned flags since from what I've seen, these flags definitely existed at one point in this form.

briansmith:
> The principled fix to this general problem is to not use linkopts for linking system installed libraries such as pthread, and instead use a target for them. Using a target would do the "deduplication" for us on the Bazel's dependency graph level. https://docs.bazel.build/versions/main/be/c-cpp.html#cc_import was supposed to be the rule to use for this, but I'm not sure if it allows linkopts-only, it used to require a file artifact (so it can easily detect modifications to that artifact, whereas using linkopts only is problematic because if the system got an update, bazel wouldn't notice).

Is there an issue tracking this principled fix? In my case, the duplication causes the build to fail because the arguments are duplicated 9,102 times and the linker command line becomes too long.

hlopko:
So I looked into the implementation of the starlark version of the cc_import (https://osscs.corp.google.com/bazel/bazel/+/master:src/main/starlark/builtins_bzl/common/cc/cc_import.bzl;bpv=1) and it seems this can be used for your usecase. Can you try it out? (maybe on a small repro case, not on all 9102 targets :)

1. you need to use very recent Bazel version (4.2.1 didn't work, I had to use their rolling release)
2. for every system installed library you depend on add equivalent of:

```
cc_import(
	  name = "zlib",
	  linkopts = ["-lz"],
	  system_provided = True,
)
```

3. remove linkopts from cc_libraries, and depend on this cc_import from cc_libraries
4. build and notice things don't work :) I just discovered that cc_import doesn't propagate linkopts if there is no artifact to link. If you could file an issue on bazelbuild/bazel that would be great. As a workaround, create an empty cc_library, and depend on the static library produced by this library from all cc_imports through their `static_library` attribute. This way there should be something to link and linkopts will be propagated.

As you can see this is a problem with Bazel C++ rules, not with rules_rust. You'll have to talk to those folks if you want more qualified answers, and you'll have to look into and file issues into bazelbuild/bazel if you have problems.

sayrer:
> So I looked into the implementation of the starlark version of the cc_import (https://osscs.corp.google.com/bazel/bazel/+/master:src/main/starlark/builtins_bzl/common/cc/cc_import.bzl;bpv=1)

It seems this is a Google-internal link.

sayrer:
Also, I suspect the duplicated lines are from the default flags per triple mapping:
https://github.com/bazelbuild/rules_rust/blob/ed0c6e9dbf3e6646531d9ee253dc7dcd0bf4a86a/rust/platform/triple_mappings.bzl#L149

PR #872, opened 2021-07-31
Fix //test/rustfmt/... on Windows

minor-fixes:
This change aims to fix `//test/rustfmt/...` tests under Windows.

The root cause of test failures is, generally speaking, attempting to locate runfiles with logic that assumes directory-based runfiles, which is not available on Windows. This change attempts to find all files through the `Runfiles` utility object so that it can work cross-platform, adding helper methods where necessary to facilitate higher-level logic.

Fixes: #619


tools/rustfmt/srcs/lib.rs
```diff
@@ -5,16 +5,20 @@ use std::path::{Path, PathBuf};
...
-    let absolute_path = if path.as_ref().is_absolute() {
-        path.as_ref().to_owned()
-    } else {
-        std::env::current_dir()
-            .expect("Failed to get working directory")
-            .join(path)
-    };
-    std::fs::metadata(&absolute_path).map(|_| absolute_path)
+pub fn current_dir_name() -> PathBuf {
+    PathBuf::from(std::env::current_dir().unwrap().file_name().unwrap())
```
minor-fixes, line 9: General runfiles question:

Runfiles seem to be passed in/specified as e.g. `external/rust_windows_x86_64/bin/rustfmt.exe`, but sometimes (concretely, when using manifest-based runfiles) they're expected to be resolved with a workspace name prefix (`rules_rust/external/rust_windows_x86_64/bin/rustfmt.exe`, in this example, with a `rules_rust` prefix)

Where is that prefix supposed to come from? Is client code/the runfiles library supposed to assume the workspace name, or divine it somehow? This happens to coincide with the basename of the CWD for both runfiles implementations - hence this sketchy function here - but I'd like to better understand how runfiles are supposed to work r.e. this workspace prefix.

tools/rustfmt/srcs/lib.rs
```diff
@@ -5,16 +5,20 @@ use std::path::{Path, PathBuf};
...
+pub fn from_slash(p: PathBuf) -> PathBuf {
+    let s = p
+        .to_string_lossy()
+        .chars()
+        .map(|c| match c {
+            '/' => std::path::MAIN_SEPARATOR,
+            c => c,
+        })
+        .collect::<String>();
+    PathBuf::from(s)
```
minor-fixes, line 21: Essentially https://docs.rs/path-slash/0.1.4/src/path_slash/lib.rs.html#302; I would have also not liked to re-implement a directory walking function. Can this repo have third-party dependencies on these crates?

UebelAndre: There are ways to use 3rd party crates but for this particular tool I'd say it's not desirable since it'd require users to load the dependencies before the target can be used. Maybe there's an easy path forward there that I'm not thinking about but my current thought is that this target should remain free of 3rd party deps 😅

tools/rustfmt/srcs/lib.rs
UebelAndre: This is a good question. I don't think the prefix is expected to be passed anywhere. Bazel uses a [launcher](https://github.com/bazelbuild/bazel/blob/23f171e2208700dc2f2f65f5ccb32bacd7aec87d/src/tools/launcher/launcher.cc#L239-L249) to set some environment variables dictating the locations of runfiles related paths. My experience is that bazel suggests paths to be treated as relative paths from the runfiles directory which is also suggested to be a relative path from the binary. I say "suggest" here since Bazel produces outputs this way but it's obviously up to the author to maintain this if the outputs are distributed in some way.

... This very likely doesn't answer your question. Maybe rephrase and I can try to be more direct 😅

tools/runfiles/runfiles.rs
```diff
@@ -103,10 +103,25 @@ impl Runfiles {
                 .clone(),
         }
     }
+
+    /// Returns a list of paths to all files in runfiles.
+    ///
+    /// Does not return directories.
+    /// The returned path may not be valid. The caller should check the path's
+    /// validity and that the path exists.
+    pub fn list_files(&self) -> Vec<PathBuf> {
```
UebelAndre, line 112: Would you be willing to add a test for this? And perhaps add a few more sample text files to `//tools/runfiles/data` to make the test a bit more interesting?

UebelAndre (changes requested):
Sorry for the delay, I like the changes here! 😄 

From my end there's just one request for a test and I hope I was able to answer some of your questions.

tools/rustfmt/srcs/lib.rs
hlopko: cc @meteorcloudy

tools/runfiles/runfiles.rs
hlopko: Hmm I actually would like to see if we could get away with not having this function at all. Other runfiles libraries didn't need it it seems (https://osscs.corp.google.com/bazel/bazel/+/master:tools/cpp/runfiles/runfiles_src.h, https://osscs.corp.google.com/bazel/bazel/+/master:tools/java/runfiles/Runfiles.java, https://osscs.corp.google.com/bazel/bazel/+/master:tools/bash/runfiles/runfiles.bash). 

It seems the main motivation is to be able to find rustfmt manifests in the test data. Is there a reason why we cannot hardcode paths to manifests in the test?

UebelAndre: Unless a unique binary is compiled for each use of the rustfmt test binary, there isn't really a way to hard code the values. I think it'd be more desirable to continue to use a shared binary and locate manifests in runfiles than to compile unique binaries but if there's a well defined pattern for runfiles libraries then I'd happily respect that. If there is though, I'd question whether or not the implementations for each language could live all in the same place.

UebelAndre:
Hey, friendly ping on this PR 😄 

Let me know if I can do something to help!

UebelAndre:
Closing out in favor of https://github.com/bazelbuild/rules_rust/pull/1375

PR #874, opened 2021-08-03
Make url overrides for downloading toolchains work.

### LINKED ISSUE #867: tools fetched from upstream rust static repo even when override URL template is provided
My WORKSPACE for the rules_rust looks like this:

```

local_repository(
    name = "rules_rust",
    path = "/Users/wturkal/projects/rust/rules_rust",
)

load("@rules_rust//rust:repositories.bzl", "rust_repositories")

rust_repositories(
    version="1.54.0",
    urls=[
        "https://{internal_mirror_site}/generic-3rdparty-local/bazel/build-support/rust/toolchains/1.54.0/{}.tar.gz"
    ],
)
```

When running a build, it still fetches from the default rust upstream location (i.e. "https://static.rust-lang.org/..."). This happens because the urls provided in my workspace config are just appended to a list whose first entry is always the default site. See https://github.com/bazelbuild/rules_rust/blob/main/rust/private/repository_utils.bzl#L529-L535 for the relevant code.

I think that the code should only add the default location iff urls are not specified.

wt:
Previously, the default url template for downloading toolchains
was always present as the first place to download tools from. This
prevents overriding the default location for the tools so that an
internal mirror can be used.

Fixes #867

### DIFF
--- a/rust/private/repository_utils.bzl
+++ b/rust/private/repository_utils.bzl
@@ -526,8 +526,11 @@ def load_arbitrary_tool(ctx, tool_name, tool_subdirectories, version, iso_date, 
     # View the indices mentioned in the docstring to find the tool_suburl for a given
     # tool.
     tool_suburl = produce_tool_suburl(tool_name, target_triple, version, iso_date)
-    static_rust = ctx.os.environ.get("STATIC_RUST_URL", "https://static.rust-lang.org")
-    urls = ["{}/dist/{}.tar.gz".format(static_rust, tool_suburl)]
+    urls = []
+
+    static_rust_url_from_env = ctx.os.environ.get("STATIC_RUST_URL")
+    if static_rust_url_from_env:
+        urls.append("{}/dist/{}.tar.gz".format(static_rust_url_from_env, tool_suburl))
 
     for url in getattr(ctx.attr, "urls", DEFAULT_STATIC_RUST_URL_TEMPLATES):
         new_url = url.format(tool_suburl)


rust/private/repository_utils.bzl
```diff
@@ -526,8 +526,7 @@ def load_arbitrary_tool(ctx, tool_name, tool_subdirectories, version, iso_date,
     # View the indices mentioned in the docstring to find the tool_suburl for a given
     # tool.
     tool_suburl = produce_tool_suburl(tool_name, target_triple, version, iso_date)
-    static_rust = ctx.os.environ.get("STATIC_RUST_URL", "https://static.rust-lang.org")
-    urls = ["{}/dist/{}.tar.gz".format(static_rust, tool_suburl)]
+    urls = []
```
illicitonion, line 529: This code used to allow adding a URL by setting an env var, and this PR removes that ability - how would you feel about making this look something like:
```py
urls = []
static_rust_url_from_env = ctx.os.environ.get("STATIC_RUST_URL")
if static_rust_url_from_env:
    urls.append("{}/dist/{}.tar.gz".format(static_rust_url_from_env, tool_suburl))
```

that way if the env var isn't set, it still behaves exactly as you've written, but if it is we still pay attention to it.

illicitonion (commented):
Thanks for putting this together! Just one question/suggestion :)

rust/private/repository_utils.bzl
wt: Thats a good point. +1

illicitonion (approved):
LGTM - thanks!

PR #876, opened 2021-08-04
Expose documentation for `rust_test_binary`

UebelAndre:
I didn't know this rule existed. Since it's exposed in `//rust:defs.bzl`, I think it should be included in documentation.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -62,6 +62,7 @@ PAGES = dict([
             "rust_proc_macro",
             "rust_benchmark",
             "rust_test",
+            "rust_test_binary",
             "rust_test_suite",
         ],
     ),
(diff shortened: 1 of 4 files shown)


UebelAndre:
Though, I'm wondering if this is something that should just be deprecated.

UebelAndre:
https://github.com/bazelbuild/rules_rust/pull/316 appears to be the origins of this rule.

hlopko:
Oh wow I also didn't know it exists. I wonder if it's actually needed even for the person who contributed it now since rust_test can setup environment variables directly.

@tommilligan do you still use `rust_test_binary`? Could you instead use `rust_test` with declaring env variables?

tommilligan:
I don't have anything active in flight with Bazel currently, so happy for you to remove if you want to.

UebelAndre:
Created https://github.com/bazelbuild/rules_rust/pull/880

PR #879, opened 2021-08-07
Updated examples using openssl to no longer assume a relative path to it

UebelAndre:
This updates the way openssl is located for certain examples which relies on it. The [rules_foreign_cc example](https://github.com/bazelbuild/rules_foreign_cc/blob/0.2.0/examples/third_party/openssl/BUILD.openssl.bazel#L51-L56) exposes the same target which has proven to be more accurate in more places (at least in my experience).


UebelAndre:
Odd that CI would succeed on in [builds/4029](https://buildkite.com/bazel/rules-rust-rustlang/builds/4029) but then after running `git commit --amend --date=now && git push -f`, (no changes made), CI would fail on windows: [builds/4046](https://buildkite.com/bazel/rules-rust-rustlang/builds/4046#73d9f1e8-780c-4833-bb42-85b1c29c1fa8)

illicitonion (approved):
Works for me if it works for CI!

PR #880, opened 2021-08-09
Delete `rust_test_binary`

UebelAndre:
Now that `rust_test` supports an `env` attribute for setting environment variables for test targets, the `rust_test_binary` rule should no longer be necessary. This PR removes it.


hlopko (approved):
LGTM, thanks for cleaning it up!

Issue #884, opened 2021-08-10
cargo_build_script uses target version of data instead of exec


djmarcin:
I'm trying to pass `PROTOC` to a `cargo_build_script` in order to use tonic. However the protoc path is `PROTOC='${pwd}/bazel-out/darwin-fastbuild/bin/external/com_google_protobuf/protoc'` when it should be `PROTOC='${pwd}/bazel-out/darwin-opt-exec-123456/bin/external/com_google_protobuf/protoc'`

The cargo_build_script config looks like:
```
cargo_build_script(
    name = "build_script",
    srcs = ["build.rs"],
    build_script_env = {
        "PROTOC": "$(execpath @com_google_protobuf//:protoc)",
    },
    data = [
        "//proto/backend",
        "@com_google_protobuf//:protoc",
        "@com_google_protobuf//:well_known_protos",
    ],
    deps = [
        "//third_party/cargo:tonic_build",
    ],
)
```

Because of this, when cross compiling the script attempts to invoke a protoc binary compiled for the target, rather than the host.

UebelAndre:
I believe this was satisfied by https://github.com/bazelbuild/rules_rust/pull/885

Please re-open if I'm mistaken 😄

alealv:
Hi

I'm still having this problem when I build for `iOS` on an `x86_64` machine.

`bazel build //apps/bertie/ios:libbertieios --config ios_arm64 --compilation_mode=fastbuild`

This is my `.bazelrc` file

```bazel
# .bazelrc
build:ios --apple_platform_type=ios
build:ios --copt=-Wno-c++11-narrowing
build:ios --noenable_platform_specific_config
build:ios --xcode_version_config=//bazel/xcode:host_xcodes
# Disable C++ warnings
build:ios --copt=-w
build:ios --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
# iOS CPU specific settings
build:ios_arm64 --config=ios
build:ios_arm64 --cpu=ios_arm64
build:ios_arm64 --platforms=//bazel/platforms:ios_arm64_platform
```

My platforms file

```bazel
platform(
    name = "ios_arm64_platform",
    constraint_values = [
        "@platforms//os:ios",
        "@platforms//cpu:aarch64",
    ],
)
```

And this is the rule failing

```bazel
cargo_build_script(
    name = "build_script",
    srcs = ["build.rs"],
    build_script_env = {
[...]

PR #885, opened 2021-08-10
Add exec transition for cargo_build_script data

djmarcin:
If tools are passed in `data`, they must be built in the `exec` configuration in order to run on the host machine when the target != host.

I _think_ this should be a fairly safe change, unless someone is doing something with the tools passed to the build script that would require them to run on the target configuration, but short of copying these tools into the out_dir I don't know what that would be. And -- that would be better handled by depending directly on these tools in the data attr of the library or binary anyway.


UebelAndre:
This is actually a very interesting find. Lets consider the case of `openssl-sys`. Which we currently have an example of that builds openssl and passes it to the build script:
https://github.com/bazelbuild/rules_rust/blob/fe657ed7ac4db6f9e9b4d67900d0dea76ee167f1/examples/sys/complex/raze/remote/BUILD.openssl-sys-0.9.60.bazel#L39-L76

Assuming `rules_foreign_cc` supports cross compilation, would this crate expect `openssl` to be built in the `target` or `exec` configuration?

djmarcin:
I suppose the question is what is the openssl-sys build script doing with the openssl binary? For what it's worth, we do use openssl-sys in our build and it doesn't seem broken to me (yet...)

djmarcin:
This (or some other mechanism for passing tools in the exec configuration) would solve https://github.com/bazelbuild/rules_rust/issues/884

illicitonion:
How about if we added a `compile_data` attribute to `cargo_build_script` as per https://github.com/bazelbuild/rules_rust/pull/516 ?

illicitonion:
Yeah, `tools` works for me!

hlopko:
`tools` sgtm too :)

djmarcin:
Updated to add a tools attr instead. This is potentially more fixup work in other tools such as cargo raze if they are using data to pass tools that should be built for the exec platform, but certainly better from the perspective of ensuring backwards compatibility.

illicitonion:
Fantastic - LGTM

One last piece - any chance of a test? I know that in our current execution environments on CI we won't see a difference, but it'd be handy for the future. I imagine adding a one-line `rust_binary` as a `tools` to a target and running it in the build script? @hlopko I'm not sure if the unit testing stuff is set up well to test this kind of thing?

UebelAndre (approved):
Looks good to me! I wonder if `tools` should be wired into `compile_data` for the `rust_binary`? Leaning towards not but would like some other opinions as well 😄 

Either way, thank you!

hlopko:
The PR looks good to me, and as for the test, I guess it would be enough to have a genrule that touches a file (using for example this rule: https://github.com/bazelbuild/bazel-skylib/blob/main/rules/write_file.bzl), and then in the unit test we would check the exec path of the file passed in data and in tools. Like you do in a different PR, exec path of the file in tools should have "exec" in it.

djmarcin:
Test added & merged main, and documentation re-updated.

illicitonion (approved):
LGTM, thanks! Over to @hlopko

hlopko (approved):
Looks good, thank you!

hlopko:
(please rebase and somebody will merge this right away, thank you!)

PR #886, opened 2021-08-11
`rustc_compile_action` now requires an explicit `attr` param

UebelAndre:
In addition to some slight cleanup of using keyword arguments in more places, `rustc_compile_action`, an internal function, now requires that attributes be passed explicitly and no longer uses `ctx.attr`.


rust/private/clippy.bzl
```diff
@@ -80,15 +80,15 @@ def _clippy_aspect_impl(target, ctx):
...
-        ctx.rule.attr,
-        ctx.file,
-        toolchain,
-        toolchain.clippy_driver.path,
-        cc_toolchain,
-        feature_configuration,
-        crate_info,
-        dep_info,
+        ctx = ctx,
+        attr = ctx.rule.attr,
```
hlopko, line 84: Let's name the attribute the same here and in get_linker_and_args

rust/private/rustc.bzl
```diff
@@ -185,26 +185,31 @@ def get_cc_user_link_flags(ctx):
     """
     return ctx.fragments.cpp.linkopts
 
-def get_linker_and_args(ctx, cc_toolchain, feature_configuration, rpaths):
+def get_linker_and_args(ctx, cc_toolchain, feature_configuration, rpaths, attrs = None):
```
hlopko, line 188: Probably just FYI at this moment, but I plan to refactor the rules to not access ctx.attr behind rustc_compile_action. Everything will have to be passed explicitly. This will make rustc_compile_action less context dependent, and more reusable for example from aspects.

hlopko: So actually this PR is a step in the right direction, and my further change would make attrs parameter non-optional (and very likely unpacked, so for example there will be a non optional 'deps' parameter)

hlopko (approved):

rust/private/clippy.bzl
UebelAndre: Done

rust/private/rustc.bzl
UebelAndre: I wasn't sure if you wanted me to do that in this change or not but I added an additional commit to make it so.

cargo/cargo_build_script.bzl
```diff
@@ -92,7 +92,7 @@ def _build_script_impl(ctx):
     # Pull in env vars which may be required for the cc_toolchain to work (e.g. on OSX, the SDK version).
     # We hope that the linker env is sufficient for the whole cc_toolchain.
     cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
-    _, _, linker_env = get_linker_and_args(ctx, cc_toolchain, feature_configuration, None)
+    _, _, linker_env = get_linker_and_args(ctx, ctx.attr, cc_toolchain, feature_configuration, None)
```
hlopko, line 95: Oh thanks for this cleanup. I didn't want to force you to do it in this PR, but thank you! :) I'd recommend going a bit further and only pass attributes that are really needed. Here get_linker_and_args actually only needs deps, so I'd pass those instead of the full ctx.attr.

No need to do it in this PR, but I'd also try (as much as reasonable) to not pass ctx around unnecessarily. Here get_linker_and_args only uses it for getting `ctx.fragments.cpp.linkopts`, so I'd probably pass `linkopts`. In some situations it may be more reasonable to pass the cpp fragment.

proto/proto.bzl
```diff
@@ -220,6 +220,7 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, is_gr
 
     return rustc_compile_action(
         ctx = ctx,
+        attr = ctx.attr,
```
hlopko, line 223: Yeah changing this to only needed attributes is a huge change, and I agree that it should be done in a separate PR, not in this one. :+1:

rust/private/rustc.bzl
```diff
@@ -204,7 +206,7 @@ def get_linker_and_args(ctx, cc_toolchain, feature_configuration, rpaths):
 
     # Add linkopt's from dependencies. This includes linkopts from transitive
     # dependencies since they get merged up.
-    for dep in ctx.attr.deps:
+    for dep in getattr(attr, "deps", []):
```
hlopko, line 209: And by making `deps` separate function parameter we can make it mandatory, and we can skip gettattr.

rust/private/rustc.bzl
```diff
@@ -613,9 +617,9 @@ def rustc_compile_action(
         collect_data = True,
     )
 
-    out_binary = False
-    if hasattr(ctx.attr, "out_binary"):
-        out_binary = getattr(ctx.attr, "out_binary")
+    # TODO: Remove after some resolution to
+    # https://github.com/bazelbuild/rules_rust/issues/771
+    out_binary = getattr(attr, "out_binary", False)
```
hlopko, line 622: Nice!

rust/private/rustc.bzl
```diff
@@ -628,18 +632,19 @@ def rustc_compile_action(
...
-        providers += establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration)
+        providers += establish_cc_info(ctx, attr, crate_info, toolchain, cc_toolchain, feature_configuration)
 
     return providers
 
 def _is_dylib(dep):
     return not bool(dep.static_library or dep.pic_static_library)
 
-def establish_cc_info(ctx, crate_info, toolchain, cc_toolchain, feature_configuration):
+def establish_cc_info(ctx, attr, crate_info, toolchain, cc_toolchain, feature_configuration):
```
hlopko, line 642: Again, this can be a separate PR, but here I'd pass ctx.actions separately, as well as deps, as well as deps, label, and out_binary

hlopko (approved):
Only nits/ramblings :) Thank you for doing this!

UebelAndre:
I've opened https://github.com/bazelbuild/rules_rust/issues/889 to track the additional comments here. Seems like there's a decent amount of things that could be updated so to make sure that gets done and things are as consistent as possible going forward, we should discuss what we want the code to look like 😄

Issue #887, opened 2021-08-11
`rust_doc_test` targets are currently broken on windows


UebelAndre:
A change to the Bazel CI windows runners broke the `rust_doc_test` targets currently defined in the repo. This issue tracks fixing these.

Related to:
- https://github.com/bazelbuild/continuous-integration/issues/1201

UebelAndre:
I've opened https://github.com/bazelbuild/rules_rust/pull/888 to disable the tests while a solution is found. This way other PRs are not blocked.

UebelAndre:
Currently RBE also does not work with `rust_doc_test`. I suspect that a solution to making RBE work would also fix the issues we're seeing with windows.

sayrer:
Yes, see also: https://github.com/bazelbuild/rules_rust/issues/845

sayrer:
I suspect the problem is here:

https://github.com/bazelbuild/rules_rust/blob/dd1c73e232ef5e1e8bf5c08e93c6fff557e18077/rust/private/rustdoc.bzl#L101

but I haven't had time to investigate further.

UebelAndre:
This rule is unlike any rule in that it is compiling rust within a test. To try and make it clearer:
| build | execution |
| --- | --- |
| generate bash/batch script | run script -> build+run doc tests |

I suspect the issue is that the test environment is not the same as the sandboxed action environment that `rustc` normally runs in (or even `rustdoc` for the `rust_doc` rule).

I think the fix is either compile the tests in an action (but not run them) using the [--no-run](https://github.com/rust-lang/rust/issues/87022) feature so that the test execution is just running the test binary. I'd expect this to shape out like:
| build | execution |
| --- | --- |
| rustdoc --no-run --persist-doctests ${CRATE} | run ${CRATE}.doc_test |

Unfortunately, this is a nightly feature so not something that can be done now. The alternative is to figure out how to create a process-wrapper for the test that ensures the environment is sufficient for building and running the tests. This is probably the better path forward since it'd be backwards compatible (not needing the nightly flag) but I don't know how to do that...

sayrer:
Given #845, I suspect we could do 

```
ctx.actions.run {
  ...
  env = {
    "CC": "...something from the toolchain..."
  }
  ...
}
```

[...]

PR #888, opened 2021-08-11
Temporarily disable `rust_doc_test` targets on windows

UebelAndre:
It's by no means ideal to disable these tests but there's currently no known solution and this is blocking all other PRs to the project. As far as I'm aware, `rust_doc_test` continues to work on some (the previous) windows environments but additional work will need to be done to update this rule to consistently work everywhere.

related to:
- https://github.com/bazelbuild/rules_rust/issues/887


hlopko:
I'll take the liberty to merge this right away, to fix the CI.

PR #892, opened 2021-08-14
Update formatting of `--extern` flags passed to `rustc`

UebelAndre:
In an effort to make the the rules more testable, the `--extern` flags, which were previously:
```
['--extern', 'crate=some/path/to/libcrate.rlib']
```
have been updated to: 
```
['--extern=crate=some/path/to/libcrate.rlib']
```

Additionally, a test has been added to test regressions in accessibility to symbols in dependencies and transitive dependencies of a target. [re-exports](https://doc.rust-lang.org/reference/items/use-declarations.html#use-visibility) (ie `pub use ...`) for example.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -866,7 +866,7 @@ def _crate_to_link_flag(crate_info):
     Returns:
         list: Link flags for the current crate info
     """
-    return ["--extern", "{}={}".format(crate_info.name, crate_info.dep.output.path)]
+    return ["--extern={}={}".format(crate_info.name, crate_info.dep.output.path)]
 
 def _get_crate_dirname(crate):
     """A helper macro used by `add_crate_link_flags` for getting the directory name of the current crate's output path
--- a/test/unit/exports/lib_b/src/lib.rs
+++ b/test/unit/exports/lib_b/src/lib.rs
@@ -1,0 +1,6 @@
+// The functionality from `lib_a` should be usable within this crate
+pub use lib_a::{greeting_a, greeting_from};
+
+pub fn greeting_b() -> String {
+    greeting_from("lib_b")
+}
--- a/test/unit/common.bzl
+++ b/test/unit/common.bzl
@@ -29,6 +29,18 @@ def assert_argv_contains_prefix_suffix(env, action, prefix, suffix):
         ),
     )
 
+def assert_argv_contains_prefix(env, action, prefix):
+    for found_flag in action.argv:
+        if found_flag.startswith(prefix):
+            return
+    unittest.fail(
+        env,
+        "Expected an arg with prefix '{prefix}' in {args}".format(
+            prefix = prefix,
+            args = action.argv,
+        ),
+    )
+
 def assert_action_mnemonic(env, action, mnemonic):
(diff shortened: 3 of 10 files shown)


test/unit/exports/lib_b/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+// The functionality from `lib_a` should be usable within this crate
```
hlopko, line 1: Functionality from`lib_a` will be available within this crate even without `pub use`. Did you want to say lib_b re-exports symbols from lib_a (so users of lib_b can use symbols from lib_a directly)?

hlopko (commented):
I think I'd expect a test that checks that we pass all transitive dependencies as `-L` and as `--extern` flags. If we didn't do that, your test would fail even if it didn't use `pub use`. In that sense mentioning re-exporting is a distraction, a red-herring, that is only slightly related to the test at hand. The fact that there is re-exporting happening is only known to rustc internally, and rules_rust don't know about it (and they don't need to know about it because of how rustc compilation model works).

What do you think about renaming the test to something like transitive deps flags test?

What do you think

test/unit/exports/lib_b/src/lib.rs
UebelAndre: correct

UebelAndre:
> I think I'd expect a test that checks that we pass all transitive dependencies as `-L` and as `--extern` flags. If we didn't do that, your test would fail even if it didn't use `pub use`. In that sense mentioning re-exporting is a distraction, a red-herring, that is only slightly related to the test at hand. The fact that there is re-exporting happening is only known to rustc internally, and rules_rust don't know about it (and they don't need to know about it because of how rustc compilation model works).
> 
> What do you think about renaming the test to something like transitive deps flags test?
> 
> What do you think

I only care that there's a test that checks this specific functionality since it's the behavior I care about. If the unittest code should actually be checking different flags then that's fine, but I'd run into an issue where it was believed this functionality was not supported so I'm hoping for something that very clearly states it is. Does that change the ask here?

rust/private/rustc.bzl
```diff
@@ -866,7 +866,7 @@ def _crate_to_link_flag(crate_info):
     Returns:
         list: Link flags for the current crate info
     """
-    return ["--extern", "{}={}".format(crate_info.name, crate_info.dep.output.path)]
+    return ["--extern={}={}".format(crate_info.name, crate_info.dep.output.path)]
```
hlopko, line 869: Nit: this is a change in behavior, and the PR claims it's only adding a test. I understand that this is done to make testing easier, but let's document that at least in the PR description.

Btw, I'd be happy to accept a PR that normalizes all flags we pass to rustc to use consistent format :)

hlopko (approved):

hlopko:
I don't know the details of the discussion you were in, but can it be that the person unaware of this already working was rather new to Rust/Bazel/rules_rust? For those people adding an unit test is not the best way to show how Rustc (not rules_rust, rustc) handles reexports. Maybe we need a text file documenting things like this in a approachable way.

For everybody else, the unit test with reexports distracts from the actual thing the test is testing.

PR #896, opened 2021-08-17
Moved examples and docs repo definitions into their WORKSPACE files

UebelAndre:
This is primarily a minor cleanup.


illicitonion (approved):
Much cleaner :)

PR #897, opened 2021-08-17
Added `default_edition` and `default_version` to `rust_common`

UebelAndre:
This should help keep the use of these variables consistent everywhere.

### DIFF
--- a/rust/private/common.bzl
+++ b/rust/private/common.bzl
@@ -25,6 +25,15 @@ In the Bazel lingo, `rust_common` gives the access to the Rust Sandwich API.
 
 load(":providers.bzl", "CrateInfo", "DepInfo", "StdLibInfo")
 
+# These constants only represent the default value for attributes and macros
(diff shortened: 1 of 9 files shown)


rust/private/common.bzl
```diff
@@ -25,6 +25,10 @@ In the Bazel lingo, `rust_common` gives the access to the Rust Sandwich API.
 
 load(":providers.bzl", "CrateInfo", "DepInfo", "StdLibInfo")
 
+# Note: Code in `.github/workflows/crate_universe.yaml` looks for this line, if you remove it or change its format, you will also need to update that code.
+DEFAULT_RUST_VERSION = "1.54.0"
```
hlopko, line 29: Please add a comment saying that this is a default that can be overriden by the toolchain (and in the future I can see we will have a build setting that will override even the toolchain). Same for the edition.

UebelAndre: Updated. Hope it's what you were thinking 😄

hlopko (approved):
Thank you!

PR #901, opened 2021-08-19
Update default edition to 2018

UebelAndre:
I've come across multiple issues where a build fails for some obscure issue and after I or some one spends a bunch of time digging around for something that *should* work, finds that the edition being used is `2015` and not `2018`, which is the default in Cargo. I feel this leads to a bunch of wasted time when getting a Rust project setup using these rules. If possible, I think the default should be `2018` so folks don't run into these issues if they forgot to set `edition = "2018"` on [rust_repositories](https://github.com/bazelbuild/rules_rust/blob/fc396284a11c90bb0fe72451045faf39d2408599/rust/repositories.bzl#L34-L71)


hlopko (approved):
LGTM

PR #903, opened 2021-08-20
Allow setting visibility on cargo_build_scripts

illicitonion:


cargo/cargo_build_script.bzl
```diff
@@ -244,6 +244,7 @@ def cargo_build_script(
         data = [],
         links = None,
         rustc_env = {},
+        visibility = None,
```
UebelAndre, line 247: Why not pull `visibility` out of `kwargs`? 

Also, what about `tags` if you're already in this neck of the woods 😄 I'm never sure how that should best be handled in macros which generate multiple targets. Though, it's fine if that isn't included in this PR

UebelAndre (changes requested):
Good idea!

One question, and buildifier/docs seems unhappy

cargo/cargo_build_script.bzl
illicitonion: I don't have valid use-cases in mind for anyone wanting to set visibility for the internal generated target whose only purpose is to run the build script, so I don't see why we should let people set that visibility unless someone _does_ have a use-case for it :)

Added tags under the same idea - we can introduce a separate `binary_tags` or similar if we need to in the future.

UebelAndre: I think at the very least, the underlying target should be tagged as `manual` then. It seems like it should only ever be build if it's used in the runner target. Is that something you'd be willing to add to this PR?

illicitonion: Done

UebelAndre: Sorry, what was done? Feels like a commit is missing, maybe?

illicitonion: Yeah, there's a whole "maybe add a manual tag to tags" thing?

illicitonion: Oh which I squashed into the regenerate documentation commit. Ugh.

UebelAndre: I still don't see that change. Looking at the latest commit on your [cbs-visibility](https://github.com/illicitonion/rules_rust/blob/c982010184d94a6504ee434718ff67bcdcf0c723/cargo/cargo_build_script.bzl#L330-L338) branch...

illicitonion: Oops, weirdness happened :) Should look better now!

UebelAndre (approved):
Awesome! Thanks!!

PR #907, opened 2021-08-23
Rust Analyzer: Generate rust-project.json without explicit target list

djmarcin:
This PR refactors the generation of rust-project.json significantly. It's not fully ready to merge and I need some advice on how to test this new design. (@hlopko)

In broad strokes:
* The aspect now emits a JSON representation of each crate in the dependency graph
* Crate deduplication has moved into Rust because we no longer aggregate crates in starlark
* The logic formerly in the rust_analyzer rule has been moved into `gen_rust_project`

`gen_rust_project` now manages the parsing of individual crate specs and creating the rust-project.json file. The basic logic is:
1) Call `bazel build` using the aspect and a list of targets.
1) Call `bazel aquery` using the list of targets to retrieve the list of generated spec files
1) Parse the individual spec files
1) Iterate through the list of crates, merging crates into rust-project.json if all dependencies have already been merged into rust-project.json (required due to the self-referential array data structure that rust-project.json uses).
1) Write the final data structure to rust-project.json in the workspace root.

Compatibility with the existing rust_analyzer rule has been dropped, save for an empty implementation to prevent immediate breakages upon updating rules_rust. In the common integration case where a user just used the defaults, no changes are required and the rust_analyzer rule can be deleted at their convenience.

Unfortunately, the existing tests rely on the rust_analyzer rule, so they are broken by this PR. Additionally, I don't know if this PR can be tested inside bazel because it repeatedly invokes bazel while running. A proper test given this approach would probably require some additional flags, so I'd like to get some feedback before I go down a rabbit hole adding hooks for testing that might end up not being what folks want to see.

/cc @hlopko @UebelAndre


UebelAndre (changes requested):
This looks awesome! You're an absolute legend!!! 😄 

I think testing for this one is one would make sense to live more in Rust than in Bazel? I'd expect to see some tests for the functions in `gen_rust_project_lib` and probably have some rule which explicitly generates a `.rust_analyzer_sysroot_src` file from the aspect use in the `rust_test` target for schema/deserialization validation. I feel like that'd provide sufficient coverage for the functionality here?

For testing the starlark code though, I did write some [unit tests for Clippy](https://github.com/bazelbuild/rules_rust/tree/8ae83f07c268352aa845699d7a0afbc73e3c7f07/test/unit/clippy) which were the first I'd ever seen written for aspects. Hopefully they can serve as some baseline? I'll keep peeking at this throughout the day and see if I have any new ideas.

rust/private/rust_analyzer.bzl
```diff
@@ -204,39 +195,25 @@ def _rust_analyzer_impl(ctx):
...
-    ).to_json())
+    sysroot_src_file = ctx.actions.declare_file(ctx.label.name + ".rust_analyzer_sysroot_src")
+    ctx.actions.write(
+        output = sysroot_src_file,
+        content = sysroot_src,
+    )
+
+    return [OutputGroupInfo(rust_analyzer_sysroot_src = depset([sysroot_src_file]))]
+
+rust_analyzer_detect_sysroot = rule(
```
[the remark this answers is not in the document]
djmarcin: The name comes from the field in the rust-project.json, but essentially it's the root under which all the rust source can be located (e.g. std::, core::, test::, etc). I'll add a comment.

hlopko:
I'm super interested in this work, but I'm also on vacation until 6th. I'll try to find time on evenings, but I apologize in advance for slowness. Thank you!

djmarcin:
@hlopko No rush, I’m on vacation this week. 

I am particularly curious if you (or anyone else) has immediate concerns about this breaking backward compatibility with the rule? Since it’s an experimental feature I’m hoping that we can just give migration instructions and not worry about attempting to interpret the existing rule. I think it’s probably possible to use the existing rule to override the default `--targets` if it’s important, though.

UebelAndre:
Hello friends, any updates here? I think at this point everyone is back from vacation but no worries if I miscalculated 😅.

rust/private/rust_analyzer.bzl
```diff
@@ -23,22 +23,26 @@ to Cargo.toml files.
...
         "crate": "rust_common.crate_info",
         "deps": "List[RustAnalyzerInfo]: direct dependencies",
         "env": "Dict{String: String}: Environment variables, used for the `env!` macro",
         "proc_macro_dylib_path": "File: compiled shared library output of proc-macro rule",
-        "transitive_deps": "List[RustAnalyzerInfo]: transitive closure of dependencies",
     },
 )
 
 def _rust_analyzer_aspect_impl(target, ctx):
+    if OutputGroupInfo in target:
```
hlopko, line 42: Why do we have to "re-provide" output groups here?

hlopko (commented):
I like the approach! Great work!

Correct me if I'm wrong, but this could be trivially scaled down to analyzing a single file (for example currently opened file in the editor). The `gen_rust_project` can compute the target of the file, and then generate rust-project.json from that.

Well that's not that exciting you may say, and you would be right :)

But imagine rust-analyzer itself detected that the project is using Bazel. It could use the `gen_rust_project` tool repeatedly for each opened file to fetch all the information it needs, invalidating what needs to be invalidated on changes to BUILD files and bzl files.

IMHO it's worth a shot talking to rust-analyzer folks once this PR lands.

Anyway, this is a great work and I'd be happy to have it merged.

rust/private/rust_analyzer.bzl
djmarcin: This is a little bit of a hack to ensure that the rust_analyzer aspect generates the rust_analyzer_sysroot_src, which doesn't have crate_info so will be skipped by the check on L46. We could use a separate aspect for it though.

More generally, we need to re-provide the output groups because otherwise if you did something like `bazel build --aspects=@rules_rust//rust:defs.bzl%rust_analyzer_aspect --output_groups=rust_analyzer_crate_spec //path/to/single:target` then only the crate_spec file for `//path/to/single:target` will be generated, but none of its deps. Re-providing OutputGroupInfo ensures that all the dependent crate_spec files are generated as well.

djmarcin (commented):
Trying to find some time to circle back on these comments and tests sometime this week.

UebelAndre:
Hello friends, friendly ping here 😅 This feature would be awesome!

purkhusid:
@djmarcin Does this improve the performance of the analyzer rules as well?

hlopko:
@purkhusid is the current performance problematic? Could you file an issue for that with more information?

purkhusid:
@hlopko I created an issue here: https://github.com/bazelbuild/rules_rust/issues/962

hlopko:
Thank you! Yeah this PR may be an (constant-factor) improvement. We'll need to measure/profile. But this PR doesn't change the fact that we have to visit all transitive crates...

UebelAndre:
Is there anything I can do to help drive this PR forward? What are the remaining steps?

rust/defs.bzl
```diff
@@ -107,6 +108,9 @@ rust_common = _rust_common
 rust_analyzer_aspect = _rust_analyzer_aspect
 # See @rules_rust//rust/private:rust_analyzer.bzl for a complete description.
 
+rust_analyzer_detect_sysroot = _rust_analyzer_detect_sysroot
```
UebelAndre, line 111: Does this rule need to be public?

UebelAndre (changes requested):

UebelAndre:
These changes were merged in https://github.com/bazelbuild/rules_rust/pull/1010

djmarcin:
Thanks @UebelAndre !

PR #909, opened 2021-08-25
Fix docs to use up-to-date rule name

vmax:
As the rule now seems to be loaded with name `rust_proto_library`, docs should use the up-to-date name when giving usage examples.


UebelAndre (approved):
Oh wow, good catch! Thank you for fixing this 😃

PR #914, opened 2021-08-27
Improved `rust_wasm_bindgen` error messaging

UebelAndre:
I was recently working through getting a use for this setup and there were some issues which made this difficult for non-frequent Bazel users. This change aims to improve messaging to make this experience better.


illicitonion (approved):
Seems reasonable to me :)

PR #916, opened 2021-08-29
Adjust env for bindgen compilation.

### LINKED ISSUE #899: rust_bindgen fails on macOS due to fatal error: 'TargetConditionals.h' file not found
I have a Bazel project where I want to use `rules_rust` with `rust_bindgen` to be able to use [Duktape](https://github.com/svaarala/duktape) in rust.

I have a `cc_library(...)` which can successfully compile Duktape, but the `rust_bindgen(...)` target fails with the following error:

```
duktape-sys/src/duk_config.h:407:10: fatal error: 'TargetConditionals.h' file not found
duktape-sys/src/duk_config.h:407:10: fatal error: 'TargetConditionals.h' file not found, err: true
thread 'main' panicked at 'Unable to generate bindings: ()', external/rules_rust_bindgen__bindgen__0_58_1/src/main.rs:54:36
```

[duk_config.h:407](https://github.com/svaarala/duktape/blob/5252b7a50611a3cb8bfcd6856b6a4a899bc622ed/config/platforms/platform_apple.h.in#L5)

When I use [`clang` from `_bindgen_clang_repositories`](https://github.com/bazelbuild/rules_rust/blob/main/bindgen/repositories.bzl#L87) (`$(bazel info execution_root)/external/bindgen_clang_osx/bin/clang -c duktape-sys/src/duktape.c `)  I get indeed this error:

```
duktape-sys/src/duk_config.h:407:10: fatal error: 'TargetConditionals.h' file not found
```

However, when using `xcrun clang  -c duktape-sys/src/duktape.c`, it compiles without problems.

Is there a way to teach `rust_bindgen` to use `clang` with the includes directories used by `xcrun clang`?

sayrer:
This fixes #899 by putting  `XCODE_VERSION_OVERRIDE`, `APPLE_SDK_VERSION_OVERRIDE`, `APPLE_SDK_PLATFORM` in the env.


illicitonion (approved):
Thanks!

bindgen/bindgen.bzl
```diff
@@ -126,6 +129,9 @@ def _rust_bindgen_impl(ctx):
         "LIBCLANG_PATH": libclang_dir,
         "RUST_BACKTRACE": "1",
     }
+    cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
+    _, _, linker_env = get_linker_and_args(ctx, ctx.attr, cc_toolchain, feature_configuration, None)
+    env.update(**linker_env)
```
illicitonion, line 134: Does this need any special handling in the case of clashes with things we explicitly set? e.g. should we be prepending/appending to the `LD_LIBRARY_PATH` rather than overwriting it?

sayrer: There's that issue, and whether the the system include directories from the cc_toolchain should be added. I'm inclined to leave it as I've written it, and adjust if people have problems with it, since this patch is clearly a step forward.

illicitonion: Yeah, sounds reasonable - let's see how we go!

illicitonion:
Can you push a rebase/merge-main, and I'll merge to main? :)

sayrer:
We should have the user in #899 contribute their test repo as an example, to get some coverage on this.

PR #918, opened 2021-08-29
Created bindgen examples directory

UebelAndre:
The `rust_bindgen` examples should be separate from others to not have it be confused with other core rules in the `rust` package.


UebelAndre:
I guess the bindgen rules don't work in all places. Opened https://github.com/bazelbuild/rules_rust/issues/919

illicitonion (approved):

PR #920, opened 2021-08-31
Don't pass along unnecessary dep info for staticlibs

djmarcin:
While debugging another issue, I noticed that staticlibs appear to include all their dependent libraries in linking. This does not seem to be necessary since the staticlib archive already includes all the dependencies.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -715,6 +715,11 @@ def establish_cc_info(ctx, attr, crate_info, toolchain, cc_toolchain, feature_co
     )
 
     cc_infos = [CcInfo(linking_context = linking_context)]
+
+    # Staticlibs don't need deps in the linking_context.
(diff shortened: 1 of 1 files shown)


hlopko:
This is a good change!

I guess when there is a shared library in deps, this PR would incorrectly throw it away?

Could we have a test for both situations (when there are only static library deps, and when there is a shared library dep)?

djmarcin:
I think that might be right. I'll try to get a test case up and fix for that if necessary.

hlopko:
Friendly ping :)

hlopko:
cc @krasimirgg @scentini this is another issue where we "leak" some transitive dependencies where we don't have to.

scentini:
Closing, as https://github.com/bazelbuild/rules_rust/pull/1299 took care of this.

PR #921, opened 2021-08-31
Enable disabled target

### LINKED ISSUE #881: Enable `@examples//ffi/rust_calling_c:matrix_dylib_test` on Darwin once we are on Bazel >= 4.2
Once we set install name on Darwin (which happens on Bazel 4.2.0) we can enable this test on the CI.

https://github.com/bazelbuild/rules_rust/blob/main/.bazelci/presubmit.yml#L34

illicitonion:
The required flag hit 4.2 which has been released

Closes #881


hlopko (approved):

UebelAndre:
Seems like there's something up with CI? But definitely a fan of enabling this target 😄

UebelAndre:
I think this will need to remove the constraints from the target now that https://github.com/bazelbuild/rules_rust/pull/1107 has been merged.

PR #923, opened 2021-09-01
Fix rust benchmark

UebelAndre:
This is a revival of https://github.com/bazelbuild/rules_rust/pull/532 (Thanks @djmarcin!)

# Changes
- Bring in the fixes outlined in https://github.com/bazelbuild/rules_rust/pull/532#issue-538007318
- Added the `use_libtest_harness` attribute to `rust_benchmark` allowing for the use of [criterion](https://bheisler.github.io/criterion.rs/book/criterion_rs.html)
- Added a small wrapper `rust_binary` for invoking benchmarks with a controlled set of arguments
- Added an example of both [criterion](https://bheisler.github.io/criterion.rs/book/criterion_rs.html) and [libtest](https://doc.rust-lang.org/unstable-book/library-features/test.html)


util/bench/bench_runner.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    // Get args but skip the binary
+    let args: Vec<String> = std::env::args().skip(1).collect();
+
+    // If there are arguments passed to the binary, use those
+    // otherwise, use that the binary was compiled with.
+    if !args.is_empty() {
+        cmd.args(args);
+    } else {
+        cmd.args(env!("BENCH_ARGS").split(" "));
```
illicitonion, line 14: Could we use https://crates.io/crates/shlex here? I know dependencies here are fiddly, but the crate itself has no dependencies, so it's just n=1...

util/bench/bench_runner.rs
```diff
@@ -0,0 +1,26 @@
+use std::process;
+
+fn main() {
+    let mut cmd = process::Command::new(env!("BENCH_BINARY"));
+
+    // Get args but skip the binary
+    let args: Vec<String> = std::env::args().skip(1).collect();
```
illicitonion, line 7: In case args contain non-utf8

```suggestion
    let args: Vec<_> = std::env::args_os().skip(1).collect();
```

examples/benchmarking/private/bench_runner.rs
```diff
@@ -0,0 +1,21 @@
+use std::process;
+fn main() {
+    // `rust_benchmark` targets produce two rust binaries. We want
+    // to get the benchmark runner executable
+    let bin = env!("BENCH_FILES")
```
illicitonion, line 5: Does this mean we compile the runner once per benchmark? Could this use `std::env::var` and get this path at runtime rather than compile time?

examples/benchmarking/libtest/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
...
+# this as manual as the examples do not use a nightly toolchain.
+rust_benchmark(
+    name = "bench",
+    srcs = ["benches/fibonacci.rs"],
+    edition = "2018",
+    tags = ["manual"],
+    use_libtest_harness = True,
+)
+
+example_bench_runner_test(
```
illicitonion, line 31: Could we remove the macro indirection here? Feels like it makes the example less clear

examples/benchmarking/criterion/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+    name = "bench",
+    srcs = ["benches/fibonacci.rs"],
+    edition = "2018",
+    use_libtest_harness = False,
+    deps = [
+        "//benchmarking/criterion/raze:criterion",
+    ],
+)
+
+example_bench_runner_test(
```
illicitonion, line 30: Could we remove the macro indirection here? Feels like it makes the example less clear

illicitonion (commented):
Generally looks reasonable - a few suggestions

util/bench/bench_runner.rs
UebelAndre: I'd like to avoid using external dependencies if possible. There's no prior art for it and if I was going to add one at all, I'd love to this rule to have a built in way of running `criterion` instead of just hoping users know how set that up themselves. Instead, could we try to get something mostly working and free of dependencies and then discuss how to add them so we have that option readily available should an issue come up?

examples/benchmarking/private/bench_runner.rs
UebelAndre: I think it's difficult to define runtime variables like this. That's kinda what the `//units/launcher` is meant to handle but would require some work to make it more generically consumable. Which could be done but figured that could happen in another PR where we want to optimize this rule. What do you think?

examples/benchmarking/libtest/BUILD.bazel
UebelAndre: Done 😄

examples/benchmarking/criterion/BUILD.bazel
UebelAndre: Done 😄

examples/benchmarking/criterion/BUILD.bazel
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 load("@rules_rust//rust:rust.bzl", "rust_benchmark")
-load("//benchmarking/private:bench_runner.bzl", example_bench_runner_test = "bench_runner_test")
+load("//benchmarking/private:example_bench_runner.bzl", "example_bench_runner_test")
```
illicitonion, line 16: Sorry, what I meant was to literally remove the `example_bench_runner_test` macro, and have a `rust_test` each place it's used...

UebelAndre: Ah, I feel like the existence of it kinda crosses the line between example and integration test. I'd like it to be clear that the target there has nothing to do with the use of `rust_benchmark` but since I suspect the common use case is going to be `criterion`, which I kinda don't want to put in the `test` package, I opted to make this target. With that being said, I don't mind deleting the macro, I just want it to be clear that it's not something users should need to care about

UebelAndre: Though, I can probably move the tests outside each package...

illicitonion: In that case I'd be pretty tempted to move this to the `test` package - I don't really mind if we end up duplicating the actually `rust_benchmark` between `examples` and `test`...

UebelAndre: I ended up moving it. Should there still be something in `examples`? Also, I'm kinda unhappy about having a test with a bunch of external dependencies. I'd really hate for a test to fail because something happened to some network resource and don't want to have iteration loops where your test failed but to thoroughly test your changes you need to build a massive application. This particular case wouldn't qualify as "massive" to me but it is in comparison to the existing tests and introducing dependencies like this can be a slippery slope.

illicitonion (approved):

PR #925, opened 2021-09-02
Bindgen can include TargetConditionals on Apple

kiron1:


examples/bindgen/simple.h
```diff
@@ -1,3 +1,7 @@
 #include <stdint.h>
 
+#if defined(__APPLE__)
```
UebelAndre, line 3: I think it'd be nice to have something which actually uses a symbol from this header. Nervous about this being optimized out for whatever reason. Though that fear may be irrational here.

One thing I would like is maybe a comment or and link to the issue describing why this is here?

UebelAndre (changes requested):

examples/bindgen/simple.h
```diff
@@ -1,3 +1,12 @@
...
+// https://github.com/bazelbuild/rules_rust/issues/899
+#include <TargetConditionals.h>
+
+const int8_t IS_MACOS = TARGET_OS_OSX;
+#else
+const int8_t IS_MACOS = 0;
```
UebelAndre, line 9: Could you also add a test that uses this value in the `main.rs` file? Where in `#[cfg(target_os = "macos")]` would test that the value is maybe not some invalid default (`-1`?)

UebelAndre (changes requested):

examples/bindgen/simple.h
sayrer: This is a good test--but `__APPLE__` will only be defined if you're using Apple's copy of clang (something that ships with XCode). If you download something from llvm.org and build on macOS, `__APPLE__` will not be defined.

kiron1: For me all versions of `clang` have `__APPLE__` defined:

```sh
/Library/Developer//CommandLineTools/usr/bin/clang -x c /dev/null -dM -E | grep __APPLE__
#define __APPLE__ 1
```

Using `clang` from [pkgsrc](https://www.pkgsrc.org).

```sh
/opt/pkg/bin/clang -x c /dev/null -dM -E | grep __APPLE__
#define __APPLE__ 1
```

Using `clang` form [LLVM.org](https://releases.llvm.org/):

```sh
/usr/local/bin/clang -x c /dev/null -dM -E | grep __APPLE__                      
#define __APPLE__ 1
```

But only the version from Apple will contain `TargetConditionals.h`. Maybe this is not a good example after all and we should leave it out?

sayrer: Sorry, I was thinking of the define `__apple_build_version__`.

PR #928, opened 2021-09-06
add CXX var to configure cpp compiler

DXist:
I use [cpp crate](https://docs.rs/cpp/0.5.6/cpp/) to integrate Rust and C++ code in my project which depends on `cc` crate.

`cc` relies on [CXX env variable](https://github.com/alexcrichton/cc-rs#c-support) to find `cpp` compiler.

This PR sets this env variable along with `CC`.


illicitonion (approved):
Seems very reasonable :) If we can get CLA bot happy, this is should be good to go!

DXist:
fixed commiter email, signed the CLA

PR #929, opened 2021-09-06
Update deps

illicitonion:


UebelAndre:
I think this is dropping the `binary_deps` feature which reintroduces dependencies/complications around OpenSSL. Is that intentional?

illicitonion:
Yes - I don't want to maintain a long-term fork, and I want to start pulling in `cargo-raze` updates, so if we want to remove those features we should work on pulling out a reusable core library :)

crate_universe/bootstrap.bzl
```diff
@@ -1,14 +1,18 @@
 """A module for declaraing a repository for bootstrapping crate_universe"""
 
+load("@examples//third_party/openssl:openssl_repositories.bzl", "openssl_repositories")
```
UebelAndre, line 3: I feel strongly that we should not be crossing boundaries into examples from the top level rules. Would you be willing to add a `crate_universe/3rdparty` directory of sorts for put the definitions for OpenSSL there?

cargo/cargo_bootstrap.bzl
```diff
@@ -55,12 +55,17 @@ def cargo_bootstrap(
     if build_mode == "release":
         args.append("--release")
 
+    env = {}
+
+    if repository_ctx.attr.include_perl_on_windows and "win" in repository_ctx.os.name:
```
UebelAndre, line 60: Instead of always using setting `PERL` here, could the `cargo_bootstrap_repository` rule take an `env` section which allows users to set these?

UebelAndre (changes requested):

cargo/cargo_bootstrap.bzl
illicitonion: I added a `windows_env_vars_to_labels` which does this _just on Windows_, but it feels still pretty similarly specific...

crate_universe/bootstrap.bzl
illicitonion: Yeah, I'm with you - fixed!

cargo/cargo_bootstrap.bzl
UebelAndre: `windows_env_vars_to_labels` to me is an improvement but not quite what I had in mind. Would something like https://github.com/bazelbuild/rules_rust/pull/932 work for you?

illicitonion: Works for me! Though exactly which triple I should be using for "Whatever CI is on Windows" I'm not sure :D

UebelAndre: I'd probably just use:
```
x86_64-pc-windows-gnu
x86_64-pc-windows-msvc
```

Perhaps that section could be expanded to support matching on components of a triple, `system`, `vendor`, etc... Host detection is not the easiest thing to do sadly... I wish there was more in Bazel to make this easy.

.bazelci/presubmit.yml
```diff
@@ -223,6 +225,7 @@ tasks:
     platform: windows
     environment:
       RULES_RUST_CRATE_UNIVERSE_BOOTSTRAP: true
+      OPENSSL_STATIC: "1"
```
UebelAndre, line 228: Could this potentially be passed as an `env` attribute to the bootstrap definition to avoid having two places where environment variables are set?

UebelAndre (changes requested):
One last question 😅

cargo/cargo_bootstrap.bzl
```diff
@@ -132,7 +132,12 @@ def _collect_environ(repository_ctx, host_triple):
             fail("File for key '{}' does not exist: {}", key, env_labels[key])
     env_labels = {key: str(value) for (key, value) in env_labels.items()}
 
-    return dict(env_vars.items() + env_labels.items())
+    if host_triple == "*":
```
UebelAndre, line 135:
```suggestion
    # In addition to platform specific environment variables, a common set (indicated by `*`) will always
    # be gathered.
    if host_triple == "*":
```

UebelAndre: Maybe the use of `*` should also make it's way into user facing docs (maybe the attribute on `cargo_bootstrap_repository`)?

UebelAndre (approved):
Two more comments but I leave it up to you to decide what to do with them. Looks good to me! Thanks for updating this 🙏

crate_universe/bootstrap.bzl
```diff
@@ -1,14 +1,30 @@
...
+
     cargo_bootstrap_repository(
         name = "rules_rust_crate_universe_bootstrap",
         cargo_lockfile = Label("//crate_universe:Cargo.lock"),
         cargo_toml = Label("//crate_universe:Cargo.toml"),
         srcs = [Label("//crate_universe:resolver_srcs")],
         version = rust_common.default_version,
         binary = "crate_universe_resolver",
+        env = {
+            "*": crago_env({
```
UebelAndre, line 18:
```suggestion
            "*": cargo_env({
```

UebelAndre (approved):

PR #933, opened 2021-09-09
Fix a unit test for Bazel 5.0

hlopko:
unittest.bzl does a configuration transition, and with Bazel 5.0
repository-relative labels will be interpreted not as in the main
repository, but as in the repository of the transition. For us it means
we have to explicitly specify which repository we mean in our test.


hlopko:
CC @gregestren, hopefully this was an intentional breaking change in Bazel 5.0.

illicitonion (approved):

PR #935, opened 2021-09-09
Add soft-failing steps that test with Bazel Rolling Releases

philwo:
With this PR each run of the CI will also run a current Bazel rolling release. Since this run can be red according to rules_rust compatibility guarantees, it will be non-blocking - the PR will be mergeable even if those runs fail.


hlopko:
LGTM from me, @UebelAndre, you've been neck deep in this stuff lately, please take a look :)

hlopko (approved):

UebelAndre (commented):
Oh awesome! Very excited about this new job!

The only thing I'd suggest is adding a job for testing examples. This way we have coverage for the rules as a consumer in another workspace. But otherwise looks good!

hlopko:
I'll go ahead and merge this, and will work on examples pipeline in a followup.

PR #937, opened 2021-09-10
Optionally capture clippy lints as a build output

jfgoog:
Writing clippy lints as a build output allows tools integrated with the build system to consume them. This is most useful when the output is in JSON format, which is already configurable.


UebelAndre (changes requested):
Seems reasonable to me 😄 

A couple of questions though:
- Can you add a test in the `test` package for this change?
- Can you update the docs to make sure others can learn about this functionality?

rust/private/clippy.bzl
```diff
@@ -75,9 +81,13 @@ def _clippy_aspect_impl(target, ctx):
         build_info,
     )
 
-    # A marker file indicating clippy has executed successfully.
-    # This file is necessary because "ctx.actions.run" mandates an output.
-    clippy_marker = ctx.actions.declare_file(ctx.label.name + ".clippy.ok")
+    capture_output = ctx.attr._capture_output[ClippyConfigInfo].capture_output
```
hlopko, line 84: Could you please move this block below after `construct_arguments`, and merge all 3 `if capture_output` blocks into one? I would move `--test` above clippy flags, below `construct_arguments`

rust/private/clippy.bzl
```diff
@@ -127,15 +145,16 @@ def _clippy_aspect_impl(target, ctx):
...
+        outputs = [clippy_out],
         env = env,
         tools = [toolchain.clippy_driver],
         arguments = args.all,
         mnemonic = "Clippy",
     )
 
     return [
-        OutputGroupInfo(clippy_checks = depset([clippy_marker])),
+        OutputGroupInfo(clippy_checks = depset([clippy_out])),
```
hlopko, line 156: I presume the infrastructure that reads ClippyInfo provider cannot read OutputGroupInfos, is that correct?

rust/private/clippy.bzl
```diff
@@ -146,6 +165,10 @@ rust_clippy_aspect = aspect(
     fragments = ["cpp"],
     host_fragments = ["cpp"],
     attrs = {
+        "_capture_output": attr.label(
+            doc = "If true, write clippy output as a build output, rather than printing it",
```
hlopko, line 169: Could you please also mention that the clippy output respects the configured `error_format`?

hlopko (changes requested):

rust/private/clippy.bzl
jfgoog: It can read it, but since it's always present we can't depend on it in a useful way.

rust/private/clippy.bzl
```diff
@@ -260,3 +284,20 @@ rust_clippy(
...
+def _capture_clippy_output_impl(ctx):
+    """Implementation of the `capture_clippy_output` rule
+
+    Args:
+        ctx (ctx): The rule's context object
+
+    Returns:
+        list: A list containing the ClippyConfigInfo provider
+    """
+    return [ClippyConfigInfo(capture_output = ctx.build_setting_value)]
```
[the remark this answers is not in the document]
jfgoog: Done.

rust/private/clippy.bzl
```diff
@@ -22,6 +22,12 @@ load(
     "construct_arguments",
 )
 load("//rust/private:utils.bzl", "determine_output_hash", "find_cc_toolchain", "find_toolchain")
+load(":providers.bzl", "ClippyInfo")
```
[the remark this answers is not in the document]
jfgoog: Done.

rust/private/clippy.bzl
jfgoog: Done

rust/private/clippy.bzl
jfgoog: Done. I did it that way to minimize the diff, but now that's clear, your suggestion is better.

jfgoog:
> * Can you add a test in the `test` package for this change?

Done.

> * Can you update the docs to make sure others can learn about this functionality?

Done.

rust/private/clippy.bzl
```diff
@@ -146,6 +160,10 @@ rust_clippy_aspect = aspect(
     fragments = ["cpp"],
     host_fragments = ["cpp"],
     attrs = {
+        "_capture_output": attr.label(
+            doc = "If true, write clippy output as a build output, rather than printing it, using the configured error_format",
```
hlopko, line 164: Nit: let's not duplicate the docs of the setting, we could say "Value of the `capture_clippy_output` setting" or something like that?

rust/private/clippy.bzl
```diff
@@ -23,6 +24,11 @@ load(
 )
 load("//rust/private:utils.bzl", "determine_output_hash", "find_cc_toolchain", "find_toolchain")
 
+CaptureClippyOutputInfo = provider(
+    doc = "Configures how to run clippy",
```
hlopko, line 28: Nit: WDYT about "Value of the `capture_clippy_output` build setting"?

rust/private/clippy.bzl
```diff
@@ -23,6 +24,11 @@ load(
...
+    fields = {"capture_output": "If true, write clippy lints as a build output rather than printing them, using the configured error_format"},
```
hlopko, line 29: Nit: wrap long line at 100-ish please

rust/private/clippy.bzl
```diff
@@ -23,6 +24,11 @@ load(
 )
 load("//rust/private:utils.bzl", "determine_output_hash", "find_cc_toolchain", "find_toolchain")
 
+CaptureClippyOutputInfo = provider(
```
hlopko, line 27: I see we are a bit inconsistent with placing providers. It makes sense to me to put all providers (including CaptureClippyOutputInfo) into providers.bzl. WDYT? (I don't mean to fix pre-existing problems in this PR)

rust/private/clippy.bzl
hlopko, line 164: TBH I think the doc here is not needed. Will buildifier complain if we leave it out? If yes, or if you prefer to have a doc set, I'd go with "Value of the capture_clippy_output build setting".

test/clippy/clippy_failure_test.sh
```diff
@@ -27,8 +27,11 @@ function check_build_result() {
   local ret=0
   echo -n "Testing ${2}... "
   ADDITIONAL_ARGS=()
-  if [[ -n "${3-}" ]]; then
-    ADDITIONAL_ARGS+=("--@rules_rust//:clippy.toml=${3}")
+  if [[ "${3-False}" != "False" ]]; then
```
hlopko, line 30: Please update function comment for added parameter

test/clippy/clippy_failure_test.sh
```diff
@@ -77,9 +80,15 @@ EOF
   check_build_result $BUILD_FAILED bad_library_clippy
   check_build_result $BUILD_FAILED bad_test_clippy
 
+  # When capturing output, clippy errors are treated as warnings and the build
+  # should succeed.
+  check_build_result $BUILD_OK bad_binary_clippy True
```
hlopko, line 85: Nit: using the local -r to declare CAPTURE_OUTPUT=True and DONT_CAPTURE_OUTPUT=False and using that in the test instead of True/False directly would make the tests more readable IMHO.

I'd also make all arguments to check_build_result mandatory and always passed.

hlopko (approved):
Only nits, overall looks great, thank you for working on this! @UebelAndre do you want to take another look?

rust/private/clippy.bzl
jfgoog: The linter complains. Changed to "Value of the capture_clippy_output build setting"

test/clippy/clippy_failure_test.sh
jfgoog: I made it so arguments >=3 are just passed to `bazel build` as-is. I also used `local -r`, which I agree improves readability.

The reason I chose not to pass extra args to bazel in the default case is because I think it better reflects how it will be actually be invoked. The vast majority of the time, you won't set `--@rules_rust//:capture_clippy_output`, but will depend on the default value. If you feel strongly I'm happy to change it, though.

UebelAndre (approved):
This looks good to me! I leave it to @hlopko to make the final call 😃

test/clippy/clippy_failure_test.sh
hlopko: Even better, awesome!

PR #939, opened 2021-09-13
Add CI tasks for rolling bazel version

hlopko:
This PR adds tasks for running our standard battery using current Bazel
rolling release. These tasks are only informative, they will not stop a
PR from submitting, but it's good to keep an eye on these and fix
quickly so we are prepared to incoming incompatible changes in Bazel.

This PR:

* adds Windows task
* adds Macos task
* adds Ubuntu 20.04 with Clang task
* adds RBE with Ubuntu 16.04 task (although I'm not sure why we still are on 16.04 there :)
* removes Ubuntu 20.04 without aspects task (not necessary)


UebelAndre (approved):
Thanks @hlopko!

PR #942, opened 2021-09-15
Support additional registries

illicitonion:


wt:
Don't be sorry. And keep being awesome.

crate_universe/defs.bzl
```diff
@@ -6,9 +6,9 @@ load("//rust:defs.bzl", "rust_common")
 load("//rust:repositories.bzl", "DEFAULT_TOOLCHAIN_TRIPLES")
 load("//rust/platform:triple_mappings.bzl", "system_to_binary_ext", "triple_to_system")
 
-DEFAULT_CRATE_REGISTRY_TEMPLATE = "https://crates.io/api/v1/crates/{crate}/{version}/download"
+DEFAULT_DEFAULT_CRATE_DOWNLOAD_URL_TEMPLATE = "https://crates.io/api/v1/crates/{crate}/{version}/download"
```
UebelAndre, line 9: `DEFAULT_DEFAULT`?

UebelAndre (changes requested):
Can you explain what this change is doing? I think I misunderstood it initially. Is this adding plumbing to add a [Cargo config](https://doc.rust-lang.org/cargo/reference/config.html) to the `cargo metadata` call?

Also one nit 😄

illicitonion:
> Can you explain what this change is doing? I think I misunderstood it initially. Is this adding plumbing to add a [Cargo config](https://doc.rust-lang.org/cargo/reference/config.html) to the `cargo metadata` call?

Yes - this is the interface `cargo-raze` expects in https://github.com/google/cargo-raze/pull/412 so we take alternate registry name-URL pairs specified as kwargs, and generate the relevant cargo config so that cargo-raze will use them.

crate_universe/defs.bzl
illicitonion: Intentional (but definitely looks weird!)

`DEFAULT_CRATE_DOWNLOAD_URL_TEMPLATE` is the crate download URL template for crates from the default registry (i.e. crates.io)

This is the default value for that

UebelAndre: Maybe add a comment above this then?

UebelAndre:
> Yes - this is the interface `cargo-raze` expects in [google/cargo-raze#412](https://github.com/google/cargo-raze/pull/412) so we take alternate registry name-URL pairs specified as kwargs, and generate the relevant cargo config so that cargo-raze will use them.

I see, I think this looks fine then, pending my other request. Nice job with the paired change!

crate_universe/defs.bzl
illicitonion: Fair - added comment :)

PR #945, opened 2021-09-17
Updated `rules_cc` and `rules_proto`

UebelAndre:


UebelAndre:
The deleted `rules_proto` patch is no longer needed since it's already included in https://github.com/bazelbuild/rules_proto/commit/fcad4680fee127dbd8344e6a961a28eef5820ef4

illicitonion (approved):

PR #946, opened 2021-09-17
Allow passing files to `rustc_env_files`

UebelAndre:


UebelAndre:
I think this is correct. Is there a particular limitation that prevents enabling this?

illicitonion (approved):

PR #953, opened 2021-09-22
fix: load providers from //:providers.bzl

manekinekko:
**BREAKING CHANGE**: `@build_bazel_rules_nodejs//:declaration_provider.bzl` has been deleted; load from `@build_bazel_rules_nodejs//:providers.bzl` instead. See https://github.com/bazelbuild/rules_nodejs/commit/cc648183160832b1540c3893963e8ef4a45e4220

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -8,7 +8,7 @@ bzl_library(
     name = "docs_deps",
     srcs = [
         "@bazel_tools//tools:bzl_srcs",
-        "@build_bazel_rules_nodejs//internal/providers:bzl",
+        "@build_bazel_rules_nodejs//:providers:bzl",
         "@com_google_protobuf//:bzl_srcs",
     ],
     deps = [
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -8,7 +8,7 @@ bzl_library(
     name = "docs_deps",
     srcs = [
         "@bazel_tools//tools:bzl_srcs",
-        "@build_bazel_rules_nodejs//:providers:bzl",
+        "@build_bazel_rules_nodejs//:providers.bzl",
         "@com_google_protobuf//:bzl_srcs",
     ],
     deps = [
--- a/wasm_bindgen/providers.bzl
+++ b/wasm_bindgen/providers.bzl
@@ -1,11 +1,8 @@
 """A module for re-exporting the providers used by the rust_wasm_bindgen rule"""
 
 load(
-    "@build_bazel_rules_nodejs//internal/providers:declaration_info.bzl",
+    "@build_bazel_rules_nodejs//:providers.bzl",
     _DeclarationInfo = "DeclarationInfo",
-)
-load(
-    "@build_bazel_rules_nodejs//internal/providers:js_providers.bzl",
     _JSEcmaScriptModuleInfo = "JSEcmaScriptModuleInfo",
     _JSModuleInfo = "JSModuleInfo",
     _JSNamedModuleInfo = "JSNamedModuleInfo",


UebelAndre:
Thanks for fixing this! I think you'll also need to update the `docs/BUILD.bazel`:
https://github.com/bazelbuild/rules_rust/blob/e589105b4e8181dd1d0d8ccaa0cf3267efb06e86/docs/BUILD.bazel#L7-L19

docs/BUILD.bazel
```diff
@@ -8,7 +8,7 @@ bzl_library(
     name = "docs_deps",
     srcs = [
         "@bazel_tools//tools:bzl_srcs",
-        "@build_bazel_rules_nodejs//internal/providers:bzl",
+        "@build_bazel_rules_nodejs//:providers:bzl",
```
UebelAndre, line 11: Typo maybe?
```suggestion
        "@build_bazel_rules_nodejs//:providers.bzl",
```

UebelAndre (changes requested):

UebelAndre:
It seems the rolling release builds are unable to start? Doesn't seem like this change would impact that.
cc @hlopko @philwo

philwo:
macOS workers do not yet have a new enough version of Bazelisk yet to support rolling releases. Will fix that next week.

RBE Rolling Bazel seems like some issue with the container config maybe. @coeuvre any idea?

docs/BUILD.bazel
manekinekko: Good catch! Thank you.

UebelAndre:
@manekinekko Seems like `providers.bzl` is not publicly visible?
```
in bzl_library rule //:docs_deps: target '@build_bazel_rules_nodejs//:providers.bzl' is not visible from target '//:docs_deps'. Check the visibility declaration of the former target if you think the dependency is legitimate. To set the visibility of that source file target, use the exports_files() function
```
Maybe that's an oversight in `rules_nodejs`?

manekinekko:
Weird! I am using v4.2.0. Maybe that's why it's working for me.

```bazel
http_archive(
    name = "build_bazel_rules_nodejs",
    sha256 = "4e1a5633267a0ca1d550cced2919dd4148575c0bafd47608b88aea79c41b5ca3",
    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/4.2.0/rules_nodejs-4.2.0.tar.gz"],
)
```

cc @alexeagle @gregmagolan

alexeagle:
Hi @manekinekko nice to see you again :)

```
ERROR: /workdir/docs/BUILD.bazel:7:12: in bzl_library rule //:docs_deps: target '@build_bazel_rules_nodejs//:providers.bzl' is not visible from target '//:docs_deps'. Check the visibility declaration of the former target if you think the dependency is legitimate. To set the visibility of that source file target, use the exports_files() function
```
this is because you're referencing the providers.bzl file directly, instead you should use some `filegroup` target in that package that exposes all the files (probably named "bzl")

manekinekko:
Thank you for the hint @alexeagle ❤️

@UebelAndre looks the `bzl_library` rule needs to be updated as well. Do you need to fix that?

https://github.com/bazelbuild/rules_rust/blob/e589105b4e8181dd1d0d8ccaa0cf3267efb06e86/docs/BUILD.bazel#L07-L12

coeuvre:
RBE Rolling Bazel issue is fixed by https://github.com/bazelbuild/rules_rust/pull/958.

UebelAndre:
Closing this in favor of https://github.com/bazelbuild/rules_rust/pull/984

PR #954, opened 2021-09-22
Add test case for OUT_DIR in tests.

sayrer:
This doesn't pass right now, but it does pass under `cargo test`. I see the `rustc` rule has a way to pass `OUT_DIR`, but it looks like tests aren't getting this right now. Posting this now, before I dig further, in case someone knows of a quick fix here.

### DIFF
--- a/test/out_dir_in_tests/src/lib.rs
+++ b/test/out_dir_in_tests/src/lib.rs
@@ -1,0 +1,13 @@
+#[cfg(test)]
+mod tests {
+    use std::env;
+    use std::fs;
+    use std::path::PathBuf;
+
+    #[test]
+    fn can_find_the_out_dir_file() {
+        let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+        let contents = fs::read_to_string(out_path.join("test_content.txt")).unwrap();
+        assert_eq!("Test content", contents);
+    }
+}
--- a/test/out_dir_in_tests/tests/out_dir_reader.rs
+++ b/test/out_dir_in_tests/tests/out_dir_reader.rs
@@ -1,0 +1,10 @@
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+
+#[test]
+fn can_find_the_out_dir_file() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let contents = fs::read_to_string(out_path.join("test_content.txt")).unwrap();
+    assert_eq!("Test content", contents);
+}
--- a/test/out_dir_in_tests/BUILD.bazel
+++ b/test/out_dir_in_tests/BUILD.bazel
@@ -1,0 +1,26 @@
+load(
+    "//rust:rust.bzl",
+    "rust_library",
+    "rust_test",
+)
+load("//cargo:cargo_build_script.bzl", "cargo_build_script")
+
+cargo_build_script(
+    name = "build_script",
+    srcs = ["build.rs"],
+)
+
+rust_library(
+    name = "demo_lib",
+    srcs = [
+        "src/lib.rs",
+    ],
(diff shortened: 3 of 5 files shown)


test/out_dir_in_tests/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+    use std::env;
+    use std::fs;
+    use std::path::PathBuf;
+
+    #[test]
+    fn can_find_the_out_dir_file() {
+        for (key, value) in env::vars() {
+            println!("{}: {}", key, value);
+        }
+        let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
```
briansmith, line 12: https://doc.rust-lang.org/cargo/reference/environment-variables.html says:
> OUT_DIR — If the package has a build script, this is set to the folder where the build script should place its output. See below for more information. (**Only set during compilation.**)

So I think this should be using `env!` instead of `env::var`.

test/out_dir_in_tests/tests/out_dir_reader.rs
```diff
@@ -0,0 +1,10 @@
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+
+#[test]
+fn can_find_the_out_dir_file() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
```
briansmith, line 7: Ditto.

sayrer: Nevertheless, the tests pass under `cargo test`, and I think the behavior should be the same.

briansmith: > For integration tests, I am not sure it makes sense for this env variable to be something only available at compile time.

It is available at compile time when run under Cargo and it is available at compile time when compiling unit tests. Further, the Cargo "spec" is explicit that the only way to get it is through `env!` during compilation and not at runtime of the test. It seems Cargo does also set it at runtime although I think that's accidental and maybe target-specific. 

But the idea that a unit test should be able to do `env!("OUT_DIR")` but an integration test shouldn't is strange. Arguably the only difference between a unit test and an integration test is that integration tests only have access to the public API of the crate.

Note in particular that not all targets have `std:::env` available; `#[no_std]` targets don't, and that's probably why the Cargo rules are the way they are.

sayrer: > Further, the Cargo "spec" is explicit that the only way to get it is through `env!` during compilation

I am not sure the "spec"/docs make a ton of sense here. Previously discussed on slack here: https://app.slack.com/client/TA4K1KQ87/CSV56UT0F/thread/CSV56UT0F-1632336998.082200

The goal is really just to match what Cargo does to a reasonable degree--I'll add some tests that use the macro as well.

> But the idea that a unit test should be able to do `env!("OUT_DIR")` but an integration test shouldn't is strange.

I didn't intend to suggest this. What I meant is that an integration test might reasonably need access to `OUT_DIR` at runtime, in addition to compile time.

sayrer:
I decided it's probably not worth it to emulate cargo here, after thinking about it some more.

This patch should probably land for test coverage and documentation purposes. I searched the repo for `OUT_DIR` while trying to figure out how to make it work in integration tests.

hlopko:
I'm a bit confused, is the test supposed to fail with bazel at the moment? Because the CI seems to be green. Do we have a bug in the CI configuration?

sayrer:
> I'm a bit confused, is the test supposed to fail with bazel at the moment? Because the CI seems to be green. Do we have a bug in the CI configuration?

No--it all works, it's just undocumented (although implied by `rust_test` `kwargs`) and uncovered by tests, I think. I had to write this PR to figure it out.

Issue #951, opened 2021-09-21
Deduplicate platform-specific linker arguments


sayrer:
It looks like the platform default linker arguments, such as:

https://github.com/bazelbuild/rules_rust/blob/ed0c6e9dbf3e6646531d9ee253dc7dcd0bf4a86a/rust/platform/triple_mappings.bzl#L149

are accumulating in the build graph, per rlib. At the limit, this issue can cause the link line to become too long for the command line to run. See #870.

Issue #889, opened 2021-08-12
Reevaluate uses of `ctx` in `//rust` package


UebelAndre:
This is a follow-up of https://github.com/bazelbuild/rules_rust/pull/886

It seems there's an inconsistent style of parsing attributes to generate the actions used to compile Rust code. The maintainers should find consensus on what that should look like going forward and provide some rationale.

PR #955, opened 2021-09-23
Only add stdlib_linkflags once.

sayrer:
More testing needed here--trying to track down #951.


sayrer:
Looks like there are some errors in CI unrelated to this patch:

```
+ /rbe_autoconf/bazelisk/bazelisk build @local_config_cc//...
  | 2021/09/23 16:39:08 Downloading https://releases.bazel.build/5.0.0/release/bazel-5.0.0-linux-x86_64...
  | 2021/09/23 16:39:08 could not download Bazel: HTTP GET https://releases.bazel.build/5.0.0/release/bazel-5.0.0-linux-x86_64 failed with error 404
```

rust/private/rustc.bzl
```diff
@@ -716,7 +720,6 @@ def establish_cc_info(ctx, attr, crate_info, toolchain, cc_toolchain, feature_co
     link_input = cc_common.create_linker_input(
         owner = ctx.label,
         libraries = depset([library_to_link]),
-        user_link_flags = depset(toolchain.stdlib_linkflags),
```
UebelAndre, line 719: How do the linker flags get passed to the action with this omitted?

UebelAndre (changes requested):
Couple of questions but thanks for working on this!

UebelAndre:
> Looks like there are some errors in CI unrelated to this patch:
> 
> ```
> + /rbe_autoconf/bazelisk/bazelisk build @local_config_cc//...
>   | 2021/09/23 16:39:08 Downloading https://releases.bazel.build/5.0.0/release/bazel-5.0.0-linux-x86_64...
>   | 2021/09/23 16:39:08 could not download Bazel: HTTP GET https://releases.bazel.build/5.0.0/release/bazel-5.0.0-linux-x86_64 failed with error 404
> ```

This is a non critical failure. Builds using the rolling release can fail (and are currently broken) so don't worry about them for now. In the future, they'll hopefully let us know when we've made a change that breaks compatibility with the rolling releases but that's work yet to be done.

rust/private/rustc.bzl
sayrer: They are added when the linker is actually run: https://github.com/bazelbuild/rules_rust/pull/955/files#diff-ecce7f6cdb4bc2f884f23fccb9a20174ad2b357dc6745314054e5eb8b784d2e0R222 (just above, you can see that code gathering the linker args from all of the dependencies)

rust/private/rustc.bzl
```diff
@@ -191,12 +191,13 @@ def get_cc_user_link_flags(ctx):
     """
     return ctx.fragments.cpp.linkopts
 
-def get_linker_and_args(ctx, attr, cc_toolchain, feature_configuration, rpaths):
+def get_linker_and_args(ctx, attr, toolchain, cc_toolchain, feature_configuration, rpaths):
```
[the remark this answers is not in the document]
sayrer: I could, but I tried to match other functions that use both the Rust toolchain and the CC toolchain. I don't feel strongly here.

rust/private/rustc.bzl
UebelAndre: Sorry, the link there is hard to tell what line it's referring to. Could you link to the branch in your fork? 🙏 😅

sayrer: https://github.com/grafica/rules_rust/blob/deduplicate_platform_args/rust/private/rustc.bzl#L222

sayrer: Both of these links seem to highlight the exact same line via a fragment identifier, so, sorry if I'm being dense. Previously, the nested loop below this comment:

```
    # Add linkopt's from dependencies. This includes linkopts from transitive
    # dependencies since they get merged up.
```

duplicated these platform default link arguments. With this patch, they get added once at the end.

rust/private/rustc.bzl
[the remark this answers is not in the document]
UebelAndre: I think this would fall under https://github.com/bazelbuild/rules_rust/issues/889 where some concerns were raised about passing larger structs (like `ctx.attr` and `ctx.fragments`) to functions. Unless something out of our control or difficult to refactor requires that the entire toolchain be passed, I think the desire is to pass smaller components. Maybe @hlopko can speak to that 😄

rust/private/rustc.bzl
UebelAndre: Ah, I see now. Thanks for bearing with me while I learn to GitHub 😅 

This seems reasonable but I'm not sure off the top of my head what [cc_common.create_link_variables](https://docs.bazel.build/versions/4.0.0/skylark/lib/cc_common.html#create_link_variables) does in terms of propagating flags to upstream targets.

I think if you were to add some before+after evidence and a regression test that I'd be happy with this 😄

sayrer: > add some before+after evidence

With the patch, `-ldl -lpthread` appears once at the beginning of `--codegen 'link-args=...`. The status quo does this:

Screen Shot 2021-09-25 at 12 46 41 PM

rust/private/rustc.bzl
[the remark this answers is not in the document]
sayrer: Once again, don't feel strongly, but I do think passing the toolchain here is better so that the arguments are well-documented and type-checked.

rust/private/rustc.bzl
sayrer: OK, I added a test below. It only checks for deduplication on Linux, but it fails without this patch (in that case, the test targets insert three copies of the platform-default link flags).

I could add more platforms and flags, but I'm not sure how to fish around in ctx to get the information I would need under test.

rust/private/rustc.bzl
```diff
@@ -217,6 +218,9 @@ def get_linker_and_args(ctx, attr, cc_toolchain, feature_configuration, rpaths):
             for linker_input in dep[CcInfo].linking_context.linker_inputs.to_list():
                 for flag in linker_input.user_link_flags:
                     user_link_flags.append(flag)
+
+    user_link_flags.extend(toolchain.stdlib_linkflags)
```
hlopko, line 222: There are 2 things that we need to take care of:

1) we need to make sure we pass `rust_toolchain.stdlib_linkflags` to `rustc` invocations doing linking. I think we're handling this use case already in this PR.
2) we need to put `rust_toolchain.stdlib_linkflags` to `CcInfo` provider. When we have a build like `cc_binary` depending on `rust_library`, linking action in `cc_binary` needs to contain all the `stdlib_linkflags`. I think we are not handling this use case currently.

I'm thinking how to best handle the 2). One approach is to provide `CcInfo` right from the `rust_toolchain`, and make sure we merge it in with the `CcInfo` of the current target. This will work, I'm not sure how good/bad idea it is to provide an additional provider from a toolchain, it's not done very often. Alternative is to add a rule that depends on the Rust toolchain, provides the `CcInfo`, and all other Rust rules implicitly depend on it to provide the `CcInfo`.

I lean towards the Rust toolchain providing `CcInfo`, but I'll try to nerd snipe @scentini into finding all the holes in that solution.

In any case though, could you add a test that:

* checks that `cc_binary` -> `rust_library` use case is handled
* checks that `stdlib_linkflags` appear after all `-l` flags for Rust crates
* checks that standalone `rust_binary` contains the flags?

And just to be sure, please make sure that all these tests still pass with the current implementation (without changes from this PR)

hlopko:
Thanks for working on this!!!

rust/private/rustc.bzl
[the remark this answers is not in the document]
UebelAndre: Yeah, it's fine to leave the PR as is. I think the toolchain won't cause issues in existing aspects. Happy to continue the conversation on the other ticket as well.

rust/private/rustc.bzl
sayrer: > * checks that `stdlib_linkflags` appear after all `-l` flags for Rust crates

On my machine, I'm getting this with the current `rust_shared_library` and `rust_binary` tests:

```
"--codegen", "link-args=-fuse-ld=gold -Wl,-no-as-needed -Wl,-z,relro,-z,now -B/usr/bin -pass-exit-codes -ldl -lpthread -lstdc++ -lm", 
"--extern=library_one=bazel-out/k8-fastbuild/bin/test/unit/platform_link_flags/liblibrary_one--1261616740.rlib", 
"--extern=library_two=bazel-out/k8-fastbuild/bin/test/unit/platform_link_flags/liblibrary_two--1109861386.rlib", 
"--extern=library_three=bazel-out/k8-fastbuild/bin/test/unit/platform_link_flags/liblibrary_three--1716324700.rlib", 
"-Ldependency=bazel-out/k8-fastbuild/bin/test/unit/platform_link_flags"
```

sayrer: Current status:

- [ ] checks that cc_binary -> rust_library use case is handled
- [ ] checks that stdlib_linkflags appear after all -l flags for Rust crates
- [x] checks that standalone rust_binary contains the flags?

hlopko:
Thanks for working on this (and I apologize for the delay here).

So I managed to implement a change so that rust toolchain provides cc info in https://github.com/bazelbuild/rules_rust/compare/main...hlopko:stdlib_linkopts?expand=1. I based the change on rules_rust@main, not on this PR. I stole your test, and discovered that currently it seems some of the tests (they're left commented out in my branch) of this PR are not passing at HEAD for the wrong reasons. The right reason would be that there are duplicated stdlib linkopts, but what I observe is that there are none. This is the flag I see for `platform_link_flags_rust_binary_native_deps_test`:

```
link-args=-fuse-ld=gold -Wl,-no-as-needed -Wl,-z,relro,-z,now -B/usr/bin -pass-exit-codes -lc -lrt -lstdc++ -lm
```

I couldn't investigate this further, but it may be that this PR is doing 2 things - adding stdlib linkopts in situations where we didn't do so, and removing duplicates. I'd recommend to focus only on the latter in this PR, since the former is a different problem (and it's not clear to me if this is actually a problem).

If I could kindly ask you to double check all that I wrote :) If it makes sense, what do you think about the change in my branch? If reasonable, feel free to integrate that into this PR.

hlopko:
Friendly ping :)

hlopko:
https://github.com/bazelbuild/rules_rust/pull/1045

PR #956, opened 2021-09-23
Added common `rustc_env_stamp_template` attribute

UebelAndre:
When building targets, it's sometimes desirable to include some information about the build itself into the resulting binaries (commit hash, time stamp, etc). At first it seemed `rustc_env_files` could be used to solve for this in combination with some `genrule` but on further investigation, it seemed that because `stamp = True` for `genrule` also included the `stable-status.txt`, the target would constantly force my rust targets to rebuild even when no changes to other inputs were made. I'm looking for a way to embed volatile-status info such that only changes to my rust target would trigger a rebuild and any change would include an updated stamp file.

The change here solves for that by running an action to produce a `rustc_env` file where the action includes all the same inputs as the `Rustc` action.

## Example
An example of this attribute can be seen below:

`BUILD.bazel`
```python
rust_binary(
    name = "stamp_printer",
    srcs = ["src/main.rs"],
    # Simply pass a file to the new attribute
    rustc_env_stamp_template = "stamp.env",
)
```
`stamp.env`
```
BUILD_TIMESTAMP={BUILD_TIMESTAMP}
```

## New rule: `internal_rust_binary`
To implement this, a new private rule was introduced called `internal_rust_binary`. Since the executable which processes the `volatile-status.txt` and template files is itself a `rust_binary`, it would have been invalid to have an attribute on `rust_binary` which depended on a `rust_binary`. To avoid this circular dependency, `internal_rust_binary` was created to be a copy of `rust_binary` but without this dependency. This is the rule used to build `//utils/env_stamper`.


UebelAndre:
I think this may be the wrong approach. Instead. This should probably be changed to introduce a `stamp` attribute that would similarly resolve stamps in `rustc_env_files`

hlopko:
I'll try to review soon, in the meantime just to be sure, https://docs.bazel.build/versions/main/command-line-reference.html#flag--workspace_status_command and https://kchodorow.com/2017/03/27/stamping-your-builds/ are relevant.

UebelAndre:
> I think this may be the wrong approach. Instead. This should probably be changed to introduce a `stamp` attribute that would similarly resolve stamps in `rustc_env_files`

As I mentioned here, I there's a simpler, more canonical approach in creating a `stamp` attribute and allowing `rustc_env_files` to contain templates. Then, similar to https://github.com/bazelbuild/rules_rust/pull/466 it'd probably be better to handle the stamping in the process wrapper as it'd keep everything in one action. Open to other ideas though.

UebelAndre:
Yeah, I think I'll rewrite this to simply introduce a `stamp` attribute.

hlopko:
https://github.com/bazelbuild/rules_rust/pull/961 landed today :)

hlopko:
Ok let me explain this more in detail now that I have a moment to do so. I've pushed a branch right now with details:

https://github.com/hlopko/rules_rust/commit/bd9730683a90d0c64ba0e17c8caaff02629c1e01

Hope it is useful. If you continue investigating this, it would be incredible if you could document all you discover somewhere, I feel like most people (including me) don't really know how stamping works, and docs are bad.

hlopko:
In the commit description I write:

> if there is a change only to the linkstamping headers, it is "ignored"

By that I mean that regenration of nonvolatile and volatile header will not invalidate the binary. There has to be an actual change to some other input in order for the binary to be rebuilt/relinked. Bazel has custom magic logic in it detecting the situation when only workspace status related files are changed, and it stops propagation.

Lastly, when the user specifies --nostamp on the bazel command line, bazel will generate "fake", nonchanging nonvolatile/volatile headers or other workspace status related files. So I believe it would work exactly like you need, with a little annoyance with having to deal with C++.

PR #958, opened 2021-09-29
Use rbe_preconfig for RBE toolchain config

coeuvre:
`rbe_autoconfig` doesn't support Bazel 5.0.0+. Use `rbe_preconfig` instead.


coeuvre:
`rbe_preconfig` uses `json` module internally which is only available in Bazel 4.0.0+. Hence, it is disabled by default and only enabled on RBE platforms.

hlopko (approved):
Looks good, thank you!

PR #961, opened 2021-10-04
Support native dependencies with linkstamps

krasimirgg:
This adds support for linking in native dependencies with linkstamps into Rust binaries. The necessary Skylark features to support this were added recently to Bazel in bazelbuild/bazel@9894cba, which has only landed in pre-release versions. This implementation performs feature detection and only enables dependency linkstamps if Bazel supports it, so this is not intended to be a functional change for versions of Bazel that do not support the Skylark linkstamp APIs.


rust/private/rustc.bzl
```diff
@@ -310,13 +317,50 @@ def collect_inputs(
...
+                    linkstamp_out = ctx.actions.declare_file(out_name)
+                    linkstamp_outs.append(linkstamp_out)
+                    cc_common.register_linkstamp_compile_action(
+                        actions = ctx.actions,
+                        cc_toolchain = cc_toolchain,
+                        feature_configuration = feature_configuration,
+                        grep_includes = ctx.file._grep_includes,
+                        source_file = linkstamp.file(),
+                        output_file = linkstamp_out,
+                        compilation_inputs = linkstamp.hdrs(),
```
hlopko, line 350: Oh I just realized one thing. We need to make sure this action is reexecuted in all the situations when we would re-link the binary. That can be achieved by adding dependencies on all linker inputs (including the toolchain inputs for Rust linking and Rust compilation), and all source files of the current target. AFAIK the only way to test this is in a shell test that invokes Bazel inside, and we don't have good support for that here in rules_rust (and the support we have internally is incredibly complex). So I'd be fine with manually veryfing that linkstamps are recompiled when:

* any transitive Rust dependency is modified
* any transitive C++ dependency is modified
* linkstamp is modified (this should be true already)
* Rust toolchain is modified
* source file of the binary is modified

hlopko: Other than that this PR LGTM, thank you!

hlopko: Actually I can be a little less dramatic and just say to make sure that linkstamp action has all the inputs that the rust compile action does (of course rust compile action will have compiled linkstmaps as input in addition).

rust/private/rust.bzl
```diff
@@ -651,6 +651,12 @@ _rust_test_attrs = {
             Whether to use libtest.
         """),
     ),
+    "_grep_includes": attr.label(
+        allow_single_file = True,
+        cfg = "host",
```
UebelAndre, line 656: @krasimirgg @hlopko just out of curiosity. Should this be `host`? Why not `exec`?

hlopko: Hmm good catch. https://github.com/bazelbuild/rules_rust/pull/1099

PR #964, opened 2021-10-08
Add and expand `env` attributes to the `Rustc` compile action

UebelAndre:
I noticed when writing a `rust_test` that if I use `env` to set environment variables, they're not available in the compile action. This sees incorrect to me I'm trying to use an environment variable to expose the `rootpath` of some file I pass to `data` and `data` is available to the compile action. The change here makes the experience more consistent.

### DIFF
--- a/test/test_env/tests/run.rs
+++ b/test/test_env/tests/run.rs
@@ -12,7 +12,7 @@ fn run() {
         "Hello fellow Rustaceans!"
     );
 
-    // Ensure environmnet variables set using `env` are also available at compile time
+    // Ensure environmnet variables set using `env` rule attribute are also available at compile time
     assert_eq!(env!("FERRIS_SAYS"), "Hello fellow Rustaceans!");
 
     // Test the behavior of `rootpath` and that a binary can be found relative to current_dir


UebelAndre:
> Are the places where you want to do this intentional (the information is really needed both at compile time and runtime), or is it the result of code being written which wasn't really thinking about the difference?

There is no *need* for this functionality as `rustc_env` would correctly set variables at compile time and `env` for runtime. You're likely correct in questioning whether or not the code really thought about `env!` vs `std::env::var` and if what's written is ideal. 

> Yeah, I don't think it's the end of the world for us to do this, and I can see the convenience, I'm just trying to fully understand the use-case :) (i.e. whether this is "because it's useful to do" or "because it papers over bugs in existing code" :))

👍 

My feeling here is that it's useful to have consistent behavior through the rules. Hard to say whether or not something like this is a bug. Maybe it's a bug in Bazel. Again, the thing I care most about is avoiding confusion. I'm happy to close this PR but want to know that that core problem will be addressed.

UebelAndre:
> And specifically I think the place the image of coupling between `data` and `env` comes from is from make-variable expansion of `$(location)` and similar macros...
> 
> I think probably the right thing for us to do here is to firm up what targets we allow in location expansion (as we are in control of this). Right now we mix in all of the expandable targets into one set, and allow them all to be expanded in all attributes - instead we should make it so that `env` only expands from the `data` attribute, and `rustc_env` only expands from the `compile_data` attribute - WDYT?

I do feel it will be hard to break the feeling that this (pull-request) is how the attributes are expected to behave. But as long as the behavior is clear and consistent I'm fine with that.

PR #981, opened 2021-10-21
Allow explicitly setting Cargo package name in `cargo_build_script`

UebelAndre:
Cargo build scripts are (to my knowledge) all represented as targets named `build_script_build` but it seems `cargo_build_script` has some components that are closely related to specific naming conventions of [cargo-raze](https://github.com/google/cargo-raze/blob/v0.12.0/impl/src/rendering/templates/partials/build_script.template#L9). I think it's slightly more correct to allow for setting an explicit package name when defining build scripts to get closer 1-1 translations of Cargo targets to Rust targets.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -53,7 +53,7 @@ def _build_script_impl(ctx):
         stderr = ctx.actions.declare_file(ctx.label.name + ".stderr.log"),
     )
 
-    pkg_name = _name_to_pkg_name(ctx.label.name)
+    pkg_name = getattr(ctx.attr, "pkg_name", _name_to_pkg_name(ctx.label.name))
 
     toolchain_tools = [
         # Needed for rustc to function.
@@ -169,6 +169,7 @@ def _build_script_impl(ctx):
         tools = tools,
         inputs = build_script_inputs,
         mnemonic = "CargoBuildScriptRun",
+        progress_message = "Running Cargo build script {}".format(pkg_name),
         env = env,
     )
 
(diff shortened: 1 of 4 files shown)


UebelAndre:
> I'm not sure what's wrong with the general "name == package_name" convention we enforce? For an area with already a lot of subtlety and nuance, it feels like introducing a third attribute mostly adds to confusion?

I think having an explicit attribute solves for the nuance here. Right now, it's pretty unclear that this "constraint" exists. There's no mention of package naming conventions anywhere and the only note about the existence of `_build_script` doesn't explain it's behavior at all.

> The name for the underlying rule. This should be the name of the package being compiled, optionally with a suffix of _build_script.

In this scenario, by trying to do more behind the scenes, I feel it's actually costing users more time than saving (since I spent some time scratching my head as to why my build scripts weren't working).

> But more broadly: why do we need this level of configurability?

I "need" this because I'm trying to map Cargo targets to Bazel and in my efforts I've found it to be more correct to treat the Cargo target name the same as the Bazel target name. I could accept using `crate_name` so long as the behavior about the package naming is also made clearer.

UebelAndre:
Opened https://github.com/bazelbuild/rules_rust/pull/985 to get one of the changes here I'd very much like to get in.

UebelAndre:
Gonna reject this. I don't think there's a strong enough consensus here to merge but I do feel package naming is in an inconsistent state (or is at least unclear). Issues here can largely be solved with an [alias](https://docs.bazel.build/versions/main/be/general.html#alias)

PR #983, opened 2021-10-21
`rustc_env_file` can now render workspace status stamp format strings

UebelAndre:
This change allows for `rust_*` targets to specify the `stamp` attribute to determine if or how format strings (`{VAR}`) will be rendered in files passed to the `rustc_env_files` attribute.

```
load("@bazel_skylib//rules:write_file.bzl", "write_file")

write_file(
    name = "env",
    out = "stamp.env",
    content = ["BUILD_TIMESTAMP={BUILD_TIMESTAMP}"],
)

rust_binary(
    name = "bin",
    srcs = ["main.rs"],
    rustc_env_files = [":env"],
    stamp = 1, # forced stamping
)
```


test/unit/stamp/stamp.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[cfg(feature = "force_stamp")]
+    #[test]
+    fn stamp_resolved() {
+        assert!(!build_timestamp().contains("BUILD_TIMESTAMP"));
```
illicitonion, line 18: Can you assert that it's numeric?

illicitonion (approved):
Someone with more C++ than me should probably review the C++, but otherwise looks great to me! :)

test/unit/stamp/stamp.rs
UebelAndre: Done!

test/unit/stamp/stamp_test.bzl
```diff
@@ -0,0 +1,121 @@
...
+    "assert_argv_contains",
+    "assert_argv_contains_not",
+)
+
+def _stamp_test_impl(ctx, force_stamp):
+    env = analysistest.begin(ctx)
+    target = analysistest.target_under_test(env)
+
+    action = target.actions[0]
+    asserts.equals(env, action.mnemonic, "Rustc")
```
hlopko, line 16: Nit: use assert_action_mnemonic

test/unit/stamp/stamp_test.bzl
```diff
@@ -0,0 +1,121 @@
...
+
+    if force_stamp:
+        assert_argv_contains(env, action, "--volatile-status-file")
+        assert_argv_contains(env, action, "bazel-out/volatile-status.txt")
+    else:
+        assert_argv_contains_not(env, action, "--volatile-status-file")
+        assert_argv_contains_not(env, action, "bazel-out/volatile-status.txt")
+
+    # Note that use of stable status invalidates targets any time it's updated.
+    # this is undesirable behavior so it's intended to be excluded from the
```
hlopko, line 26: Nit: s/this/This

util/process_wrapper/process_wrapper.cc
```diff
@@ -76,8 +80,18 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
         value = System::GetWorkingDirectory();
       }
       subst_mappings.push_back({std::move(key), std::move(value)});
+    } else if (arg == PW_SYS_STR("--volatile-status-file")) {
+      if (!volatile_status_file.empty()) {
+        std::cerr << "process wrapper error: \"--volatile-status-file\" can "
+                     "only appear "
```
hlopko, line 86: Nit: join these two short lines please.

rust/private/stamp.bzl
```diff
@@ -0,0 +1,75 @@
+"""A small utility module dedicated to detecting whether or not the `--stamp` flag is enabled
```
hlopko, line 1: Could you please move this file into //rust/settings/?

rust/private/stamp.bzl
```diff
@@ -0,0 +1,75 @@
...
+    attrs = {
+        "value": attr.bool(
+            doc = "The default value of the stamp build flag",
+            mandatory = True,
+        ),
+    },
+)
+
+def stamp_build_setting(name, visibility = ["//visibility:public"]):
+    native.config_setting(
```
hlopko, line 42: I don't have any issue with this implementation, but here https://docs.bazel.build/versions/main/skylark/lib/transition.html#transition I see that `//command_line_option:cpu` works for transitions for accessing `--cpu`, so maybe `//command_line_option:stamp` works as well and can be used as any other build settings. May be worth trying out.

rust/private/rust.bzl
```diff
@@ -623,6 +629,25 @@ _common_attrs = {
...
+            [--nostamp](https://docs.bazel.build/versions/main/user-manual.html#flag--stamp) builds. \
+            This setting should be avoided, since it potentially kills remote caching for the target and \
+            any downstream actions that depend on it.
+
+            - `stamp = 0`: Always replace build information by constant values. This gives good build result caching.
+
+            - `stamp = -1`: Embedding of build information is controlled by the \
+            [--[no]stamp](https://docs.bazel.build/versions/main/user-manual.html#flag--stamp) flag.
+
+            Stamped targets are not rebuilt unless their dependencies change.
```
hlopko, line 646: Could you please add an example situation to give more clarity into the impact of this behavior: For example if a `rust_library` is stamped, and a `rust_binary` depends on that library, the stamped library won't be rebuilt when we change sources of the `rust_binary`. This is different from how [`cc_library.linkstamps`](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library.linkstamp) behaves.

hlopko (approved):
Only minor things, but looks good to me. I apologize for the delay.

rust/private/stamp.bzl
UebelAndre: I feel strongly that this file should not need to exist in the first place and that the value of `--stamp` should be easily accessible within rules. https://github.com/bazelbuild/bazel/issues/11164 should be implemented and this file deleted so wherever it goes, I'd like it not to be a part of the public API.

rust/private/stamp.bzl
UebelAndre: This implementation was inspired by [rules_pkg](https://github.com/bazelbuild/rules_pkg/blob/0.5.1/pkg/private/BUILD#L48-L51) which has been in place for some time. I'd rather stick with the same approach since it seems to have withstood the test of time (so far).

rust/private/rust.bzl
UebelAndre: Correct, each target does it's own stamping, if there are no updates to the `rust_library`, it will not be rebuilt and will maintain the stamp it was cached with. If that's not desirable, stable stamping can always be used to generate the `rustc_env_files` targets so both the library and binary are guaranteed to have the same stamp. What would you like in terms of an example? Just some more documentation?

UebelAndre:
@hlopko Thanks for the review! Some quick follow up questions if you don't mind 😅

rust/private/stamp.bzl
hlopko: Sounds good.

rust/private/stamp.bzl
hlopko: Ack. But by trying to use `//command_line_option:stamp` you may discover that it's already possible to access value of `--stamp` flag from Starlark today, negating the need for the workaround.

rust/private/rust.bzl
hlopko: Oh yeah I'm sorry for not being clear in my comment. I was merely asking for more documentation (everything from the colon to the end of my comment was a suggestion for added documentation - and that was absolutely not clear from the comment :).

rust/private/stamp.bzl
hlopko: (But if I had to bet, I'd bet on `//command_line_option` thing not working as a normal build setting, I'd expect the label only exists in the context of configuration transitions)

UebelAndre: These two options may not be what you were thinking but demonstrating that they don't work.

```diff
diff --git a/rust/private/stamp.bzl b/rust/private/stamp.bzl
index 1a7cab6..3292225 100644
--- a/rust/private/stamp.bzl
+++ b/rust/private/stamp.bzl
@@ -48,7 +48,7 @@ def stamp_build_setting(name, visibility = ["//visibility:public"]):
     _stamp_build_setting(
         name = name,
         value = select({
-            ":stamp_detect": True,
+            "//command_line_option:stamp": True,
             "//conditions:default": False,
         }),
         visibility = visibility,
```
```
ERROR: Analysis of target '//rust/private:stamp' failed; build aborted: no such package 'command_line_option': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
```

```diff
diff --git a/rust/private/stamp.bzl b/rust/private/stamp.bzl
index 1a7cab6..7f0f963 100644
--- a/rust/private/stamp.bzl
+++ b/rust/private/stamp.bzl
@@ -41,7 +41,7 @@ _stamp_build_setting = rule(
 def stamp_build_setting(name, visibility = ["//visibility:public"]):
     native.config_setting(
         name = "stamp_detect",
-        values = {"stamp": "1"},
+        values = {"//command_line_option:stamp": "1"},
         visibility = visibility,
     )
```
```
ERROR: /rules_rust/rust/private/BUILD.bazel:15:20: in values attribute of config_setting rule //rust/private:stamp_detect: error while parsing configuration settings: unknown option: '//command_line_option:stamp'. Since this rule was created by the macro 'stamp_build_setting', the error might have been caused by the macro implementation
```

I've used `//command_line_option` with transitions (exactly what you linked to earlier) but I feel the build setting is simpler than transitions and would opt to keep this as is.

UebelAndre: This is something I'm happy to make another PR for if I can understand the appropriate change. Since this is an exciting change for some, I'm gonna merge this since ultimately this is an implementation detail that does not impact the API.

PR #1000, opened 2021-10-29
Fix outdated instructions

SUPERCILEX:
I'm pretty sure `rust.bzl` isn't supposed to be used directly anymore.

> PS: please squash.


dfreese (approved):
Looks good, thanks!

Issue #1003, opened 2021-11-02
Forward CXX env and arguments from cargo_build_script


ikalchev:
Objects produced with `cargo_build_script` may be linked in `cc_*` targets. For this to succeed, the options for building with a cargo build script and with `cc_*` should be (mostly) the same.

When using a `cargo_build_script` that compiles a C++ object, the env and command line arguments from the `cc_toolchain` are not forwarded to `cargo_build_script_runner/bin.rs`.

(This is in continuation of my comments here https://github.com/bazelbuild/rules_rust/pull/969#issuecomment-955198208)

PR coming soon.

# Example
Suppose the `cc_toolchain` defines `-D_GLIBCXX_USE_CXX11_ABI=0`. Now suppose we use a `cargo_build_script` that compiles a `cc_library` that has in its sources:

```
void f(const std::string& str);
```

If we now try to link this library in a `cc_binary` which calls `f`, we will get a link-time error for `undefined reference to f`. Upon inspecting the symbols of the library produced by `cargo_build_script` we can find `f`. However, upon further inspection, the mangling of the parameter for the function is different - `std::__cxx11:basic_string...`, whereas the cc_binary looks for `std::basic_string...`.

UebelAndre:
I believe we already forward CXX args. 
https://github.com/bazelbuild/rules_rust/blob/af24d277c91a69a0eae05b4ade444dae4e2784c4/cargo/private/cargo_build_script.bzl#L467-L474

If this is not the case please feel free to reopen!

PR #1004, opened 2021-11-02
Forward CXX env and arguments from cargo_build_script

ikalchev:
We need to forward the C++ command line arguments and env to ``bin.rs``
to make sure cargo_build libraries are compiled with the same options
as `cc_*` targets.

Fixes https://github.com/bazelbuild/rules_rust/issues/1003

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@
 
 Google Inc.
 Spotify AB
+VMware Inc.
 Damien Martin-Guillerez <dmarting@google.com>
 David Chen <dzc@google.com>
 Florian Weikert <fwe@google.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -20,3 +20,4 @@ Philipp Wollermann <philwo@google.com>
 Ulf Adams <ulfjack@google.com>
 Justine Alexandra Roberts Tunney <jart@google.com>
 John Edmonds <jedmonds@spotify.com>
+Ivan Kalchev <kivan@vmware.com>
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -1,5 +1,9 @@
 # buildifier: disable=module-docstring
-load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "C_COMPILE_ACTION_NAME")
+load(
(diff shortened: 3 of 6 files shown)


illicitonion (commented):
@ikalchev Thanks for putting this together!

@hlopko Do we have any prior art on wiring up a custom cc_toolchain in a unit analysis test, so that we could test that args get properly propagated from there?

JasperDeSutter:
These arguments can contain paths relative to the exec root, i.e. "external/androidndk/...' when cross-compiling for android (and perhaps other targets).
The cargo build script runner doesn't resolve these without a symlink or rewriting the paths to be absolute. I think a symlink would be the best option because a heuristic for detecting paths to external seems very fragile.

illicitonion:
@JasperDeSutter What kind of symlink are you suggesting?

JasperDeSutter:
> @JasperDeSutter What kind of symlink are you suggesting?

A directory symlink mapping exec_root/external to CARGO_MANIFEST_DIR/external before running the build script. I'm not sure how many bazel rules this breaks, but I hacked it into cargo_build_script_runner's bin.rs and that seems to work.
The reason we would need this is because the build script is executed from CARGO_MANIFEST_DIR instead of exec_root.

illicitonion:
> > @JasperDeSutter What kind of symlink are you suggesting?
> 
> A directory symlink mapping exec_root/external to CARGO_MANIFEST_DIR/external before running the build script. I'm not sure how many bazel rules this breaks, but I hacked it into cargo_build_script_runner's bin.rs and that seems to work. The reason we would need this is because the build script is executed from CARGO_MANIFEST_DIR instead of exec_root.

I see! Yeah, this could work, and would I imagine cover most of the use-cases people actually hit - it's possible that toolchains can point at files in-repo rather than in a separate workspace (hidden behind `external`), but it's probably rare...

I had been considering doing some on-the-fly flag editing to basically for each flag, if resolving it from the exec-root it happens to be an existing file/directory, to absolutify the flag, but that feels pretty invasive. The symlink feels like a pretty reasonable middle-ground to make most things work most of the time without breaking much (and we can add a "don't create external symlink" attribute to `cargo_build_script` if we really need)...

UebelAndre:
Closing per https://github.com/bazelbuild/rules_rust/issues/1003#issuecomment-3917286463

PR #1006, opened 2021-11-03
Avoid Win32 MAX_PATH issues by shortening `cargo_build_script` bin names

UebelAndre:
I'm running into issues on Windows where `cargo_build_script` targets are filing to find files. After an investigation, I've found the cause to be paths exceeding the [Windows max path length limitation](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd). This change would shorten the runfiles path, potentially keeping some runfiles paths within the 260 character cap

before:
```
C:\bzl\czfrdgju\execroot\cargo_bazel_examples\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\multi__libnghttp2-sys-0.1.7-1.45.0\libnghttp2-sys_build_script_.exe.runfiles\multi__libnghttp2-sys-0.1.7-1.45.0\nghttp2/lib/includes/nghttp2/nghttp2ver.h.in
```
after:
```
C:\bzl\czfrdgju\execroot\cargo_bazel_examples\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\multi__libnghttp2-sys-0.1.7-1.45.0\libnghttp2-sys_build_script_script_.exe.runfiles\multi__libnghttp2-sys-0.1.7-1.45.0\nghttp2/lib/includes/nghttp2/nghttp2ver.h.in
```

Understandably this doesn't solve for all issues since another target may exceed the character limit again, but every character counts and given that this hasn't been a huge issue, I suspect [libnghttp2-sys](https://crates.io/crates/libnghttp2-sys) might just be the extreme case (or we don't have many windows users).


illicitonion (approved):
Sure :)

PR #1009, opened 2021-11-11
Better handling of versioned library filenames

ddeville:
#513 fixed an issue that prevented linking against a versioned library (using a version suffix on linux) but didn’t address a problem with `get_lib_name` that prevents from linking against certain libraries that happen to include the version number **before** the extension, such as `libpython3.8.dylib` on macos.

Before this patch, we would attempt to link `-lpython3` (since it would cut the filename up to the first `.` in the library name) and thus failing whereas after the diff it correctly links against `-lpython3.8`.

Also removed the `_has_dylib_ext` function that seems to be a leftover that was added in #808 but left unused after #820.

### DIFF
--- a/test/unit/versioned_libs/versioned_libs_analysis_test.bzl
+++ b/test/unit/versioned_libs/versioned_libs_analysis_test.bzl
@@ -1,0 +1,233 @@
+"""Analysis tests for getting the link name of a versioned library."""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_import")
+load("//rust:defs.bzl", "rust_shared_library")
+
+LIBNAMES = ["sterling", "cheryl", "lana", "pam", "malory", "cyril"]
+
+def _is_in_argv(argv, version = None):
+    return any(["-ldylib={}{}".format(name, version or "") in argv for name in LIBNAMES])
+
+def _no_version_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    argv = tut.actions[0].argv
+
+    asserts.true(env, _is_in_argv(argv))
+
+    return analysistest.end(env)
+
+def _prefix_version_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    argv = tut.actions[0].argv
+
+    asserts.true(env, _is_in_argv(argv, "3.8"))
+
+    return analysistest.end(env)
+
+def _suffix_version_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    argv = tut.actions[0].argv
+
+    asserts.true(env, _is_in_argv(argv))
+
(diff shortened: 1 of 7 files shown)


ddeville:
I've signed the CLA, feel free to re-run.

UebelAndre:
It seems `buildifier` needs to be run. Also would it be possible to add a unit test asserting that the action outputs match the new patter and specifically avoids the previous issues?

UebelAndre (changes requested):
Thanks! Sorry for the delay here. One more request: 

Would you be able to maybe use `rust_library` (or similar) rules to also test this code path? I'd be good to have a clear understand of, based on the target name of the `crate_name` attribute, what to expect.

UebelAndre:
After that the changes look good to me!

ddeville:
Sure I can look into that.

test/unit/versioned_libs/versioned_libs_analysis_test.bzl
```diff
@@ -0,0 +1,235 @@
...
+        name = "malory.dll",
+        srcs = ["b.c"],
+        linkshared = True,
+    )
+    native.filegroup(
+        name = "malory_interface_lib",
+        srcs = [":malory.dll"],
+        output_group = "interface_library",
+    )
+    native.genrule(
```
UebelAndre, line 171: Small nit: could `copy_file` be used here in this function? Is there a difference in it's behavior on windows than on linux?

UebelAndre (approved):
Thank you so much for putting this together and for putting up with all my requests! Looks good to me!

PS: I had one clarifying question for my own sake, if you don't mind 😅

test/unit/versioned_libs/versioned_libs_analysis_test.bzl
ddeville: Ah yes I believe it should work just the same. I was switching between `genrule` and `copy_file` while trying to debug a naming issue with the interface library (I think `cc_library` generates the lib file as `something.dll.lib` that Windows doesn't recognize since it expects `something.lib` to live next to `something.dll`, hence the renaming).

UebelAndre: Ah, ok, Sounds good. Could you move the `genrule` to `copy_file` for consistency if it works?

ddeville: Yep I just did. Let's see if CI is happy with it :)

ddeville: Oh it doesn't and now I remember why it is a `genrule`. Looks like the `interface_library` `filegroup` generates multiple outputs (even though by the time the `genrule` runs `$<` is a single input 🤔) and `copy_file` chokes on it.

UebelAndre (approved):

test/unit/versioned_libs/versioned_libs_analysis_test.bzl
UebelAndre: That's fine. Maybe just add a comment about if you've still got the code open? 🙏

ddeville: For sure, doing now.

ddeville: ok I actually had a typo and `copy_file` works fine, sorry for the confusion 😄

PR #1010, opened 2021-11-11
Rust Analyzer: Generate rust-project.json without explicit target list

UebelAndre:
This is a follow up of https://github.com/bazelbuild/rules_rust/pull/907 (special thanks to @djmarcin!!) but with some comments addressed and an integration test added.

This change increases the min supported Bazel version from `3.5.0` to `4.0.0` (this is a factor from the original PR).


UebelAndre:
f1b887ade55229a33daf0518985293afee6c3572 should be a squashed commit of all changes in https://github.com/bazelbuild/rules_rust/pull/907 for easier reviewing

UebelAndre:
@hlopko Hey, I'd really love to get these changes in so we can continue to get more feedback on `rust-analyzer`. Having played around with it

rust/private/rust_analyzer.bzl
```diff
@@ -160,38 +163,29 @@ def _create_crate(ctx, infos, crate_mapping):
...
-
-    crate["deps"] = [{"crate": d[0], "name": d[1]} for d in deps]
-    crate["cfg"] = canonical_info.cfgs
+
+    # TODO: The only imagined use case is an env var holding a filename in the workspace passed to a
+    # macro like include_bytes!. Other use cases might exist that require more complex logic.
+    expand_targets = getattr(ctx.rule.attr, "data", []) + getattr(ctx.rule.attr, "compile_data", [])
+    crate["env"].update({k: ctx.expand_location(v, expand_targets) for k, v in info.env.items()})
+
+    # Omit when a crate appears to depend on itself (e.g. foo_test crates).
```
hlopko, line 180: Please add the big comment back, I find it useful context:

```
It can happen a single source file is present in multiple crates - there can
    be a `rust_library` with a `lib.rs` file, and a `rust_test` for the `test`
    module in that file. Tests can declare more dependencies than what library
    had. Therefore we had to collect all RustAnalyzerInfos for a given crate
    and take deps from all of them.

    There's one exception - if the dependency is the same crate name as the
    the crate being processed, we don't add it as a dependency to itself. This is
    common and expected - `rust_test.crate` pointing to the `rust_library`.
```

tools/rust_analyzer/main.rs
```diff
@@ -148,13 +108,10 @@ struct Config {
...
-    // If not specified, uses the result of `bazel info bazel-bin`.
-    #[structopt(long)]
-    bazel_bin: Option<PathBuf>,
-
     #[structopt(long, default_value = "bazel")]
     bazel: PathBuf,
 
-    #[structopt(long, default_value = "//:rust_analyzer")]
-    bazel_analyzer_target: String,
+    /// Comma-separated list of target patterns
```
hlopko, line 114: I'd keep the flag syntax here identical to bazel - "space separated" list of target patterns without a dedicated --flag.

tools/rust_analyzer/rust_project.rs
```diff
@@ -0,0 +1,304 @@
...
+    sysroot_src: &str,
+    crates: &[CrateSpec],
+) -> anyhow::Result<RustProject> {
+    let mut project = RustProject {
+        sysroot_src: Some(sysroot_src.into()),
+        crates: Vec::new(),
+    };
+
+    let mut unmerged_crates: Vec<&CrateSpec> = crates.iter().collect();
+    let mut skipped_crates: Vec<&CrateSpec> = Vec::new();
```
hlopko, line 101: It's getting too late here for my brain to think, but it seems this algorithm is going to great lengths to process a crate after its deps and do it all in a single loop. But the only reason for that seems to be to get the index of the dep. And that sounds like unnecessary requirement. Maybe the index could be precomputed in one linear pass, and then the crates could be generated in another linear pass?

Also I'm not sure I see where the merging happens? As noted below, could you add a test that shows the scenario of rust_library and rust_test using `crate` attribute (therefore they have the same id I believe)?

tools/rust_analyzer/rust_project.rs
```diff
@@ -0,0 +1,304 @@
...
+        assert_eq!(project.crates.len(), 1);
+        let c = &project.crates[0];
+        assert_eq!(c.display_name, Some("example".into()));
+        assert_eq!(c.root_module, "example/lib.rs");
+        assert_eq!(c.deps.len(), 0);
+    }
+
+    #[test]
+    // An example with a one crate having two dependencies.
+    fn generate_rust_project_with_deps() {
```
hlopko, line 244: Could you add tests that exercise merging? Showing that for example rust_library and rust_test have their deps merged (and potentially showing that if either of those doesn't have deps or the test comes before the library the algorithm still works)

hlopko (changes requested):
Thank you for taking this over! By luck I see you're fixing one bug I was meaning to look into :P

rust/private/rust_analyzer.bzl
```diff
@@ -61,21 +65,32 @@ def _rust_analyzer_aspect_impl(target, ctx):
...
         env = getattr(ctx.rule.attr, "rustc_env", {}),
         deps = dep_infos,
-        transitive_deps = transitive_deps,
+        crate_specs = depset(direct = [crate_spec], transitive = [dep.crate_specs for dep in dep_infos]),
         proc_macro_dylib_path = find_proc_macro_dylib_path(toolchain, target),
         build_info = build_info,
-    )]
+    )
+
+    ctx.actions.write(
```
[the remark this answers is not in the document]
UebelAndre: `rust_analyzer_info` is written to the file produced by `ctx.actions.write`.

tools/rust_analyzer/main.rs
UebelAndre: Are you familiar enough with [StructOpt](https://crates.io/crates/structopt) to know how to do that easily?

UebelAndre:
@hlopko for context on my involvement here. I've been interested in these changes for a while and saw there was no activity on the original PR. I spent a few late nights to test the changes out and to write some integration tests, which to my understanding were the biggest missing component to the last PR. I agree with some of the changes you're proposing but since my experience with this code is mostly as a user, some of the proposals are not quick changes for me. Are these things that can be done post-merge, given that the `rust-analzyer` rules are still experimental? Or what would be the core changes required to get this merged? Not trying to skip the review process here but don't want these changes to continue to be blocked behind our increasingly limited availability.

UebelAndre:
That said, I can vouch for these changes in my workspaces and they're awesome. Even enabled it for `rules_rust` and it was night and day in the developer experience

UebelAndre:
@djmarcin would you be able to speak to some of the comments left by @hlopko?

tools/rust_analyzer/main.rs
djmarcin: I think the idea being proposed is to use the final varargs to do that, and it looks like you might be able to do that by specifying neither short nor long in the annotation, like this from the example:
```
    /// Files to process
    #[structopt(name = "FILE", parse(from_os_str))]
    files: Vec<PathBuf>,
```

tools/rust_analyzer/rust_project.rs
djmarcin: Yes, that's the goal. It iteratively processes any crate with zero dependencies that have not yet been populated into the list. I had to do some things that felt a little goofy to satisfy the borrow checker, so there may be ways to simplify it. The complexity is worst case n^2, but only occurs when the dependency tree is entirely linear and reverse ordered. In most cases the complexity is closer to n*d where d is the maximal depth of the dependency tree, which is generally a fairly small constant, and n collapses very quickly after 1-3 passes as most library crates have a fairly shallow dependency tree.

I am not sure that you can do this in two linear passes, because I think that rust-analyzer may assume that dependencies always appear prior to use, but I am not sure of this. I _believe_ that's why I chose this route, but I am not certain of that -- there were a few bugs during development, and I may have been dealing with a different bug that turned out to be that some dependencies were not getting populated at all. So, the two linear passes approach may actually be fine, it just needs to be tested to ensure out-of-order dependencies work with rust-analyzer.

djmarcin (commented):
Thanks so much for picking this up @UebelAndre, let me know if you need any assistance finalizing this PR.

UebelAndre:
@djmarcin I'm hoping to get these changes in ASAP so more folks can use them and updates become more targeted. If any of @hlopko's feedback would be a quick thing for you, I'd hugely appreciate it if you were able to make those changes 🙏 😅.

purkhusid:
Super pumped about trying this out, good work!

tools/rust_analyzer/rust_project.rs
hlopko: Discussed with Andre offline, I'm fine with leaving this investigation/change to a follow-up PR.

tools/rust_analyzer/rust_project.rs
hlopko: Now that I reviewed the tests I think I'd be fine with bringing the reversed tests back, since that one covers the behavior I'm interested in.

test/rust_analyzer/rust_analyzer_test_runner.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
+# Runs Bazel build commands over rustfmt rules, where some are expected
```
djmarcin, line 3: I think this mentions rustfmt incorrectly?

test/rust_analyzer/rust_analyzer_test_runner.sh
```diff
@@ -0,0 +1,94 @@
...
+rust_analyzer_deps()
+EOF
+
+cat << EOF > "${new_workspace}/.bazelrc"
+build --keep_going
+test --test_output=errors
+build:strict --aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect
+build:strict --output_groups=+rustfmt_checks
+build:strict --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
+build:strict --output_groups=+clippy_checks
```
djmarcin, line 44: Do you need the rustfmt / clippy stuff here?

test/rust_analyzer/rust_analyzer_test_runner.sh
```diff
@@ -0,0 +1,94 @@
...
+    
+    pushd "${workspace}" &> /dev/null
+    echo "Generating rust-project.json..."
+    bazel run "@rules_rust//tools/rust_analyzer:gen_rust_project"
+    echo "Building..."
+    bazel build //...
+    echo "Testing..."
+    bazel test //...
+    echo "Building with Aspects..."
+    bazel build //... --config=strict
```
djmarcin, line 74: Do you need the build, test, and build --config=strict?

test/rust_analyzer/rust_analyzer_test_runner.sh
```diff
@@ -0,0 +1,94 @@
...
+    if [ "$(uname)" == "Darwin" ]; then
+        SEDOPTS=(-i '' -e)
+    else
+        SEDOPTS=(-i)
+    fi
+    sed ${SEDOPTS[@]} 's/"manual"//' "${workspace}/BUILD.bazel"
+    
+    pushd "${workspace}" &> /dev/null
+    echo "Generating rust-project.json..."
+    bazel run "@rules_rust//tools/rust_analyzer:gen_rust_project"
```
djmarcin, line 68: This seems to just validate that gen_rust_project runs at all and doesn't validate the output, is that correct?

test/rust_analyzer/rust_analyzer_test_runner.sh
djmarcin: Oh I think I see -- it copies in the `WORKSPACE` and `BUILD.bazel` files and then the `test //...` ends up running the rust-analyzer manual tests.

test/rust_analyzer/rust_analyzer_test_runner.sh
UebelAndre: No but it's the only context that these targets would get rustfmt and clippy applied to them so wanted the integration test to be stricter by including them. Do you think they'd be problematic?

test/rust_analyzer/rust_analyzer_test_runner.sh
UebelAndre: The test targets the selves consume a `rust-project.json` which should be doing that validation. Otherwise what are the tests testing?

test/rust_analyzer/rust_analyzer_test_runner.sh
UebelAndre: It does, I'll fix later this afternoon

tools/rust_analyzer/rust_project.rs
UebelAndre: I think I've solved for the concern here by adding tests to the `aquery` module.

test/rust_analyzer/rust_analyzer_test_runner.sh
hlopko: Sounds good to me, but please mention this in a code comment, I think this is not obvious otherwise.

hlopko (approved):
Thank you so much!

test/rust_analyzer/rust_analyzer_test_runner.sh
UebelAndre: Done

PR #1012, opened 2021-11-12
Replaced uses of `@bazel_tools` with `@rules_cc`

UebelAndre:
This is kinda a minor cleanup but really just wondering if it makes sense to use `rules_cc` for more things vs `bazel_tools`.

### DIFF
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -195,7 +195,7 @@ rust_bindgen = rule(
             allow_single_file = True,
         ),
         "_cc_toolchain": attr.label(
-            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+            default = Label("@rules_cc//cc:current_cc_toolchain"),
         ),
         "_process_wrapper": attr.label(
             default = Label("//util/process_wrapper"),
@@ -209,7 +209,7 @@ rust_bindgen = rule(
     toolchains = [
         str(Label("//bindgen:bindgen_toolchain")),
         str(Label("//rust:toolchain")),
-        "@bazel_tools//tools/cpp:toolchain_type",
+        "@rules_cc//cc:toolchain_type",
     ],
     incompatible_use_toolchain_transition = True,
 )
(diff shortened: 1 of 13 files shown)


hlopko:
Hmm, honestly, I think Bazel folks abandoned the `rules_cc`. We may be better off not using that. If it was up to me I'd probably remove our current usage of rules_cc instead.

UebelAndre:
> Hmm, honestly, I think Bazel folks abandoned the `rules_cc`. We may be better off not using that. If it was up to me I'd probably remove our current usage of rules_cc instead.

Can you confirm, I don't recall ever hearing that `rules_cc` was abandoned and there's still a [buildifier-lint](https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#native-cc) demanding users load `cc` rules from it and it only mentions the work was postponed.

illicitonion:
Last I heard too was that the shim rules were being vaguely deprecated, e.g. https://github.com/bazelbuild/bazel/issues/8746#issuecomment-704209400

scentini:
I advocate for sticking to `@bazel_tools` and native rules until the Bazel team starts advertising `@rules_cc` again. The plan was for everyone to start using `@rules_cc` with Bazel 1.0; when that didn't work out, work on `@rules_cc` stagnated and the ruleset is currently in a state of flux.
The buildifier lint was disabled by default in https://github.com/bazelbuild/buildtools/pull/952, and its [docs](https://github.com/bazelbuild/buildtools/blob/4.2.3/WARNINGS.md#all-c-build-rules-should-be-loaded-from-starlark) now say:

> Update: the plans for disabling native rules have been postponed, at the moment it's not required to load Starlark rules.

UebelAndre:
Well then. Now the question is where to find the time to go rip out `rules_cc` 😅 

Thanks for the insights everyone

UebelAndre:
Hey @scentini @illicitonion @hlopko, catching up on Bazelcon now and saw https://youtu.be/L3xiMvOIeuM . Does that impact anyone's stance to stick with `@bazel_tools`?

illicitonion:
Thanks for sharing! The vibe I got from that talk is "Maybe we should do this in 6-12 months, we'll see" :)

scentini:
From what I understood is they plan to transfer the rules into their own repositories at some point. IMO we should switch to `rules_cc` after the transfer has happened.

Issue #947, opened 2021-09-18
`rust_doc` does not define `CARGO_MANIFST_DIR`


UebelAndre:
I've had success using the following to consistently access files in both Cargo and Bazel builds (something prescribed in the [examples](https://github.com/bazelbuild/rules_rust/blob/75c63c36ab6a268d4de773437400d1d853723e84/examples/cargo_manifest_dir/external_crate/src/lib.rs)):
```rust
include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/path/to/file"));
```

Unfortunately, `CARGO_MANIFEST_DIR` is not defined in actions under [rust_doc](https://github.com/bazelbuild/rules_rust/blob/75c63c36ab6a268d4de773437400d1d853723e84/rust/private/rustdoc.bzl#L148-L198). I'm also not sure how to infer that information from a [CrateInfo](https://github.com/bazelbuild/rules_rust/blob/75c63c36ab6a268d4de773437400d1d853723e84/rust/private/providers.bzl#L17-L40) provider. Would love some ideas in how to define this or would like to know alternatives I could use for [include_str](https://doc.rust-lang.org/std/macro.include_str.html).

dae:
For normal binaries/libs/tests, include_str!() references are relative to the file they fall in, so foo/bar.rs can reference baz/quux.txt with include_str!("../baz/quux.txt"). Is the doc code behaving differently?

illicitonion:
This is how we set it for other things: https://github.com/bazelbuild/rules_rust/blob/75c63c36ab6a268d4de773437400d1d853723e84/rust/private/rustc.bzl#L396-L413

UebelAndre:
> This is how we set it for other things:
> 
> https://github.com/bazelbuild/rules_rust/blob/75c63c36ab6a268d4de773437400d1d853723e84/rust/private/rustc.bzl#L396-L413

Yeah, It seems simple enough to expose `CARGO_MANIFEST_DIR` to the action but what is the value of it? What would represent the manifest dir? Seems like it'd be something on `CrateInfo` but I'm not sure what.

UebelAndre:
> For normal binaries/libs/tests, include_str!() references are relative to the file they fall in, so foo/bar.rs can reference baz/quux.txt with include_str!("../baz/quux.txt"). Is the doc code behaving differently?

The issue is that Cargo and Bazel behave differently for those strings and it's valuable to me to support both (makes transitioning easy or even possible in some of my cases).

[...]

Issue #804, opened 2021-06-26
`rustdoc` sandboxing issues


UebelAndre:
It seems rustdoc is currently disabled on [RBE](https://docs.bazel.build/versions/main/remote-execution-ci.html) environments. This rule likely needs to be updated to appropriately account for this.

Issue #980, opened 2021-10-21
`rust_doc_test` Failed to link crates when proc-macro is used in the dependencies


vbkaisetsu:
To describe the situation, I created the following small repository:
https://github.com/vbkaisetsu/rules-rust-link-fails-repro

# Steps to reproduce

1. Clone the repository.
    ```
    git clone https://github.com/vbkaisetsu/rules-rust-link-fails-repro.git
    ```

2. Run the test.
    ```
    cd rules-rust-link-fails-repro
    bazel run //repro:my_lib_doc_test
    ```

3. The following error occurs:
    ```
    exec ${PAGER:-/usr/bin/less} "$0" || exit 1
    Executing tests from //repro:my_lib_doc_test
    -----------------------------------------------------------------------------
    error[E0460]: found possibly newer version of crate `fnv` which `tonic` depends on
     --> repro/my_lib.rs:3:5
      |
    3 | use tonic::Status;
      |     ^^^^^
      |
      = note: perhaps that crate needs to be recompiled?
      = note: the following crate versions were found:
              crate `fnv`: /home/koichi.akabe/.cache/bazel/_bazel_koichi.akabe/65e8d7f25191e794777d18f078967787/execroot/link_fail_repro/bazel-out/k8-opt-exec-2B5CBBC6/bin/external/raze__fnv__1_0_7/libfnv--1978688209.rlib
              crate `tonic`: /home/koichi.akabe/.cache/bazel/_bazel_koichi.akabe/65e8d7f25191e794777d18f078967787/execroot/link_fail_repro/bazel-out/k8-fastbuild/bin/external/raze__tonic__0_5_2/libtonic--506703014.rlib
    
    error: aborting due to previous error
    ```

# Workarounds

* Run the test with optimization.
  ```
  bazel run -c opt //repro:my_lib_doc_test
  ```

* Change `cfg` of `proc_macro_deps` from `exec` to `target`.
https://github.com/bazelbuild/rules_rust/blob/82b650d5d0709ae4c0ee8584f4ed92112ba11d67/rust/private/rust.bzl#L575

jyn514:
I don't know whether this is related at all, but I remember seeing https://github.com/rust-lang/rust/issues/13983 a while ago.

PR #1015, opened 2021-11-14
Rustdoc rules now share a code path with Rustc rules for more consistency

UebelAndre:
## `rust_doc`
The first issue I faced with `rust_doc` was https://github.com/bazelbuild/rules_rust/issues/947 but quickly found there were many more variables being defined for `Rustc` actions but not `Rustdoc`, leading to my inability to use the rule. The changes here introduce a macro `@rules_rust//rust/private:rustdoc.bzl%rustdoc_compile_action` designed to create a near identical action `@rules_rust//rust/private:rustc.bzl&rustc_compile_action` for improved consistency in generating docs to solve for the issues I ran into.

## `rust_doc_test`
Additionally, the `rust_doc_test` was similarly unusable due to the fact that `rustdoc --test` builds and tests binaries as a test execution but the test environment was quite different from the build environments of `Rustdoc` and `Rustc` actions. This PR shares the same `rustdoc_compile_action` functionality to determine action args, environment vars, and inputs and write an executable to rebuild an execution environment similar to the build actions. 

Notably, this fixes `rust_doc_test` in RBE and Windows environments.

## Related Issues
closes https://github.com/bazelbuild/rules_rust/issues/887
closes https://github.com/bazelbuild/rules_rust/issues/947
closes https://github.com/bazelbuild/rules_rust/issues/804
closes https://github.com/bazelbuild/rules_rust/issues/1016


rust/private/rustdoc_test.bzl
```diff
@@ -15,7 +15,71 @@
 """Rules for performing `rustdoc --test` on Bazel built crates"""
 
 load("//rust/private:common.bzl", "rust_common")
-load("//rust/private:utils.bzl", "dedent", "find_toolchain", "get_lib_name", "get_preferred_artifact")
+load("//rust/private:rustdoc.bzl", "rustdoc_compile_action")
+load("//rust/private:toolchain_utils.bzl", "find_sysroot")
+load("//rust/private:utils.bzl", "dedent", "find_toolchain")
+
+def _construct_writer_arguments(ctx, test_runner, action, crate_info, rust_toolchain):
```
illicitonion, line 22: I'm kind of surprised at home much of this is actual logic vs re-calling functions defined elsewhere... What's unique about this that means that e.g. we need to set up the SYSROOT specially?

illicitonion (commented):
Looks like a nice clean-up, thanks!

rust/private/rustdoc_test.bzl
UebelAndre: This is largely solving for the fact that tests run from a runfiles directory where actions run in an execroot. But it also tracks what environment variables were explicitly added to the action. Happy to simplify if possible 😀

illicitonion: Makes sense, and seems like it's a hard thing to share around - probably worth a comment to that effect :)

tools/rustdoc/rustdoc_test_writer.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    content.extend(env.iter().map(|(key, val)| format!("{}='{}' \\", key, val)));
+
+    let argv_str = argv
+        .iter()
+        // Remove any substrings found in the argument
+        .map(|arg| {
+            match strip_substrings
+                .iter()
+                .find(|substring| arg.contains(substring.as_str()))
```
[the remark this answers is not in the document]
UebelAndre: It's intentional that this looks for one match then continues. Since the substrings are reverse sorted by the length of the string, it means the first match should be the correct match for this particular argument. I was worried about shorter strings getting in and there being an incorrect replacement. Though, I don't think there's much risk in a double replacement in how this turned out. Do you still think I should be trying to do replacements on the entire list?

UebelAndre:
I was unfortunately not able to get a solution for https://github.com/bazelbuild/rules_rust/issues/980 in this PR (which is a hard problem to solve, for me 😅).

tools/rustdoc/rustdoc_test_writer.rs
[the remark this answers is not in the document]
illicitonion: I think it's fine if we document that the idea is these are non-overlapping paths to roots, at at most one of them is expected to match - right now that assumption is a bit scattered around the code... But just replacing all of them would avoid needing to document that assumption at all :)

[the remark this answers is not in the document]
UebelAndre: Updated, now all substrings are replaced 😄

illicitonion (approved):
Thanks!

rust/private/rustc.bzl
```diff
@@ -460,7 +460,9 @@ def construct_arguments(
         build_flags_files,
         emit = ["dep-info", "link"],
         force_all_deps_direct = False,
-        stamp = False):
+        force_link = False,
```
hlopko, line 463: Would it be possible to add a unit test demonstrating the usage of this parameter and its expected effect (probably on the generated rustc command line?) Same for `remap_path_prefix`.

UebelAndre: Sorry about the delay on this one: https://github.com/bazelbuild/rules_rust/pull/1078

PR #1022, opened 2021-11-17
Make action_env work for rustc actions

djmarcin:
Currently, action_env is not passed through to rustc actions, which means that variables like `DEVELOPER_DIR` needed in some cases on macOS cannot be passed through. This PR passes action_env through to rustc actions.

Similar to existing code at https://github.com/bazelbuild/rules_rust/blob/e2f0fccda912daac686b533ad77c5bc5d2f2ddb7/cargo/cargo_build_script.bzl#L68

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -729,7 +729,7 @@ def rustc_compile_action(
         stamp = stamp,
     )
 
-    args, env = construct_arguments(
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thanks! Looks good to me but would like another person to look too 😃

illicitonion (approved):

Issue #992, opened 2021-10-23
Disabling formatting of generated bindgen code is not possible


csmulhern:
The bindgen rule falls back on the rustfmt target specified by the rust_toolchain if the rust_bindgen_toolchain has rustfmt set to None. See:
https://github.com/bazelbuild/rules_rust/blob/d974e0ea173f5b5dc0ebd56946881b24ebe23248/bindgen/bindgen.bzl#L94

If rustfmt_version is set to None when calling rust_repositories, a rust version is used instead. See:
https://github.com/bazelbuild/rules_rust/blob/d974e0ea173f5b5dc0ebd56946881b24ebe23248/rust/repositories.bzl#L77

Because of this, it's not possible to disable formatting when generating bindgen code. It would be nice to be able to disable this for performance reasons.

It seems like the project should either:
1) Not try to intelligently fallback on the rustfmt of the rust toolchain if the rust bindgen toolchain rustfmt target is None in _rust_bindgen_impl (my recommendation).
2) Allow rustfmt to be unset on the rust toolchain itself (I'm not sure of the implications of this, but I'm not sure why it would be required).

djmarcin:
This is particularly problematic when the nightly version of rustfmt is broken. Right now, the output of the nightly rustfmt on my generated code is an empty file.

PR #1024, opened 2021-11-18
Allow rust_bindgen rules to disable rustfmt with an attr

djmarcin:
Fixes https://github.com/bazelbuild/rules_rust/issues/992, but it is worth noting that this behavior has existed for a long time and could result in files that were previously being formatted becoming unformatted if toolchains are not specifying the rustfmt parameter when they do want formatted code. The unformatted code is still perfectly valid, and this behavior change matches [the documentation](https://bazelbuild.github.io/rules_rust/rust_bindgen.html#rust_bindgen_toolchain).

### DIFF
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -40,6 +40,7 @@ def rust_bindgen_library(
         cc_lib,
         bindgen_flags = None,
         clang_flags = None,
+        rustfmt = True,
         **kwargs):
     """Generates a rust source file for `header`, and builds a rust_library.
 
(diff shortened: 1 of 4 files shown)


UebelAndre:
Hmm, would it not be worth it to instead add some testing for the available versions of Rust to ensure each toolchain works? It sounds like the real issue is we add untested nightly toolchains that break users. Am I understanding that correctly?

djmarcin:
I don’t think so. The request in #992 is to disable rustfmt for speed reasons. Additionally, users can add arbitrary nightly toolchains and there doesn’t seem to be a compelling reason to force bindgen code to be formatted.

An alternative solution may be to add a parameter to the bindgen rule that would allow explicitly disabling formatting on a per-rule basis (e.g. rustfmt = False)

UebelAndre:
I think i like the flag to disable rustfmt. I'd had some issues with the proto toolchains in this repo where things I'd expect to come from the rust toolchain weren't and caused confusion for folks as a result. I think it'd be good to keep the fallback but allow for opting out of rustfmt. Which do you think is more ergonomic?

djmarcin:
I was just going with what the documentation currently states, but I think either is fine honestly. I'll take another pass to update the docs to reflect reality and add the parameter to the rust_bindgen rule.

UebelAndre (approved):
Awesome! Thank you! 

edit: Oh, CI hadn't finished 😅

djmarcin:
> Awesome! Thank you!
> 
> edit: Oh, CI hadn't finished 😅

Should be good now, I verified locally, just waiting on CI

UebelAndre (approved):

PR #1025, opened 2021-11-18
declare CARGO_CFG_(WINDOWS|UNIX) in build scripts

dae:
These variables are documented as being available:
https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

The latest rusqlite depends on CARGO_CFG_WINDOWS to be defined
in order to link correctly: https://github.com/rusqlite/rusqlite/pull/961

### DIFF
--- a/cargo/cargo_build_script_runner/bin.rs
+++ b/cargo/cargo_build_script_runner/bin.rs
@@ -236,6 +236,10 @@ fn get_target_env_vars<P: AsRef<Path>>(rustc: &P) -> Result<BTreeMap<String, Str
     let stdout = std::str::from_utf8(&output.stdout)
         .map_err(|err| format!("Non-UTF8 stdout from rustc: {:?}", err))?;
 
+    Ok(parse_rustc_cfg_output(stdout))
+}
+
+fn parse_rustc_cfg_output(stdout: &str) -> BTreeMap<String, String> {
     let mut values = BTreeMap::new();
 
     for line in stdout.lines() {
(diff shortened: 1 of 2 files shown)


cargo/cargo_build_script_runner/bin.rs
```diff
@@ -250,6 +250,8 @@ fn get_target_env_vars<P: AsRef<Path>>(rustc: &P) -> Result<BTreeMap<String, Str
                     .or_insert_with(Vec::new)
                     .push(value[1..(value.len() - 1)].to_owned());
             }
+        } else if ["windows", "unix"].contains(&line) {
```
UebelAndre, line 253: Can you explain this a bit? What exactly is this catching and how does it relate to `CARGO_CFG_(WINDOWS|UNIX)` (maybe in the form of a comment)?

UebelAndre (changes requested):
One question probably for my own sake. Also, could you add a test that asserts `CARGO_CFG_(WINDOWS|UNIX)` is set in the expected cases?

cargo/cargo_build_script_runner/bin.rs
dae: 'unix' or 'windows' are one of the lines output by rustc:

```
$ rustc --print=cfg
debug_assertions
target_arch="x86_64"
target_endian="little"
target_env=""
target_family="unix"
target_feature="fxsr"
target_feature="sse"
target_feature="sse2"
target_feature="sse3"
target_feature="ssse3"
target_os="macos"
target_pointer_width="64"
target_vendor="apple"
unix
```

The resulting lines are uppercased and prefixed directly below.

UebelAndre: Ah! I see now, thank you! Would you be willing to add some comment like the following to help future @UebelAndre out? 😄 
```suggestion
        // `rustc --print=cfg` may include these exact values
        } else if ["windows", "unix"].contains(&line) {
```

UebelAndre (changes requested):

UebelAndre (approved):
Awesome work! Thanks for putting this together 😄

PR #1028, opened 2021-11-19
Don't filter out pre-release versions for git deps

### LINKED ISSUE #1018: pre-release dependency ignored by crate_universe in member_packages_version_mapping
The dependency in question from `Cargo.toml`:
```
tantivy = { git = "https://github.com/quickwit-inc/tantivy", rev="72cef12db1697a230a0e51170d9801af410f529b", default-features = false }
```

The version in there is `0.17.0-dev`. In diffing the `crate_universe` lockfile vs before when the dependency was a non-pre-release version `0.16.1`, I see it is being removed altogether from the `member_packages_version_mapping` section.

The build fails for the package where this dependency is required, but it succeeds if I add a mapping manually as `0.17.0`.

illicitonion:
Fixes #1018

### DIFF
--- a/crate_universe/src/resolver.rs
+++ b/crate_universe/src/resolver.rs
@@ -372,10 +372,20 @@ impl Resolver {
         for package in &merged_cargo_metadata.packages {
             for dep in &package.dependencies {
                 if dep.kind == kind {
-                    member_package_version_reqs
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Sound good to me!

PR #1030, opened 2021-11-22
Don't duplicate stdlib link flags in rustc compile action arguments 

ddeville:
I've been migration a semi large Rust codebase from our own rules to `rules_rust` and frequently run into errors when spawning clang due to the argument list being too large.

It looks like the argument list is mostly filled with hundreds or even thousands duplicated entries for `stdlib_linkflags` (since they appear in each transitive Rust dependency).

This patch attempts to prevent the duplication of `stdlib_linkflags` in the linker args for a given compile action.

### DIFF
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -133,7 +133,7 @@ def _rust_bindgen_impl(ctx):
         "RUST_BACKTRACE": "1",
     }
     cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
-    _, _, linker_env = get_linker_and_args(ctx, ctx.attr, cc_toolchain, feature_configuration, None)
+    _, _, linker_env = get_linker_and_args(ctx, ctx.attr, cc_toolchain, feature_configuration, None, rust_toolchain.stdlib_linkflags)
     env.update(**linker_env)
 
     # Set the dynamic linker search path so that clang uses the libstdcxx from the toolchain.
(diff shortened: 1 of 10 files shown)


UebelAndre:
This seems to be solving for the same thing as https://github.com/bazelbuild/rules_rust/pull/955
@sayrer, Since the other PR seems to have stalled, do the changes here work for you and/or do you have any feedback? I think it'd be cool to get either PR in but also wanna try to make sure everyone's needs are met 😄

UebelAndre:
@hlopko would you be able to take a look since you seemed to already have some good questions on https://github.com/bazelbuild/rules_rust/pull/955 (a similar PR)? 🙏

hlopko:
So, I'm still thinking what to do and experimenting with options locally. So far there is not a solution I'm excited about. But with https://github.com/bazelbuild/rules_rust/pull/1042 and https://github.com/bazelbuild/rules_rust/pull/1033 I think this problem is far less pressing. Would you agree?

ddeville:
I'll try running my test suite with #1033 and without this change to confirm, but yes I think it should work.
#1042 also look like a more correct change than mine so yay! :)

ddeville:
Hmm I'm actually still seeing this on macos with a very large number of duplicated `"-lSystem"` and `"-lresolv"` entries (it's a fairly large crate dependency tree but there are 7,064 of each so 14,128 stdlib link args in total!) when calling into `external/local_config_cc/cc_wrapper.sh` that leads to a failure to `posix_spawn`.

This is happening when building a `cargo_build_script` though which made me think that [this](https://github.com/bazelbuild/rules_rust/blob/e2f0fccda912daac686b533ad77c5bc5d2f2ddb7/cargo/cargo_build_script.bzl#L144) also needed a param file but it doesn't seem to help (likely because the call to `cc_wrapper.sh` *from* bazel rather than from the rules *to* bazel?)

hlopko:
Let's continue here: https://github.com/bazelbuild/rules_rust/pull/1045

PR #1033, opened 2021-11-25
Use a param file when invoking rustc since the argument list can become too large

ddeville:
Even with #1030 I still seem to experience build errors due to the argument list being too long. This change ensures that arguments are written to a param file should they exceed the maximum length on the current platform.


hlopko (approved):

hlopko:
Thank you!

PR #1035, opened 2021-11-25
tools/runfiles: use $TEST_SRCDIR when available

TimNN:
Bazel sets the `$TEST_SRCDIR` environment variable telling tests the "absolute path to the base of the runfiles tree", see [the test encyclopedia](https://docs.bazel.build/versions/main/test-encyclopedia.html#initial-conditions).

That should in general be the more reliable way to determine the runfiles directory (when available), I think, but the current logic also does not always work when running tests internally at Google.

### DIFF
--- a/tools/runfiles/runfiles.rs
+++ b/tools/runfiles/runfiles.rs
@@ -111,6 +111,14 @@ pub fn find_runfiles_dir() -> io::Result<PathBuf> {
         std::env::var_os("RUNFILES_MANIFEST_ONLY").unwrap_or_else(|| OsString::from("0")),
         "1"
     );
+
+    // If bazel told us about the runfiles dir, use that without looking further.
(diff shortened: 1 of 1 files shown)


UebelAndre:
This change seems good to me but wanna make sure we're not missing anything else while focus is on this section of the rules 😄. @TimNN Are you able to talk about the specifics of what doesn't work with the current implementation? I requested @hlopko join the review (as, to my knowledge, the only Googler actively maintaining these rules) in case there's some sensitive details here.

TimNN:
I don't think the details are particularly sensitive. The issue is that the directory structure used by our CI system does not follow the `<binary name>.runfiles` pattern that the current implementation expects, so it cannot find the runfiles directory.

hlopko (approved):
Thanks for this! LGTM.

Issue #1032, opened 2021-11-25
rust-analyzer: Choose the library's `display_name` as much as possible.


reiyw:
Thank you for the Rust Analyzer support without an explicit target list https://github.com/bazelbuild/rules_rust/commit/8c05ac765f61ba735ad243bb826f92e45ae9bd6f. This greatly reduces our workload.

https://github.com/bazelbuild/rules_rust/blob/8c05ac765f61ba735ad243bb826f92e45ae9bd6f/tools/rust_analyzer/aquery.rs#L347-L348
As left as a TODO, when there is a `CrateSpec` with the same ID, the information of the first `CrateSpec` added to the `BTreeSet` takes precedence, regardless of whether it is a test or a library. The comment says "probably better if the display name matches the library target", but in fact, I had to make the display name matches the library target or Rust Analyzer returned an unresolved import error. I haven't looked at the source code of Rust Analyzer in detail, but from my experiments, it seems that `display_name` is used to match crate entries in `rust-project.json` against symbols in the source file.

In order to choose the library's `display_name` as much as possible when consolidating crate specs, I made the following change.
```diff
@@ -167,6 +167,9 @@ fn consolidate_crate_specs(crate_specs: Vec<CrateSpec>) -> anyhow::Result<BTreeS
         log::debug!("{:?}", spec);
         if let Some(existing) = consolidated_specs.get_mut(&spec.crate_id) {
             existing.deps.extend(spec.deps);
+            if existing.display_name.ends_with("_test") {
+                existing.display_name = spec.display_name;
+            }
         } else {
             consolidated_specs.insert(spec.crate_id.clone(), spec);
         }
```
This seems to have eliminated the unresolved import error in our repository and everything seems to be working fine, but I'm not sure how general this solution is. Can anyone come up with a better, more general solution? If this solution would suffice, I will submit a PR.

UebelAndre:
Hi! Thanks for trying out the new `rust-analyzer` functionality! I think for this TODO it might be better to allow the aspect to generate an additional field that tracks whether or not the crate is a test target and allow non-test display names to always be written to the `BTreeMap`, overriding whatever was there before. Do you know if binary targets would share the same ID as a Library target?

reiyw:
Thank you for your quick response! Your suggestion seems much more reasonable.

> Do you know if binary targets would share the same ID as a Library target?

[...]

PR #1039, opened 2021-11-26
rust-analyzer: Set a library's `display_name` when consolidating crate specs

reiyw:
This PR addresses the issue https://github.com/bazelbuild/rules_rust/issues/1032.

When there were multiple build targets sharing the same source, their crate specs were consolidated into one, and their `display_name` depended on the result of aquery. Since the `display_name` is actually a target name, it caused a mismatch between the `display_name` and the actual crate name referenced by the Rust source, leading to unresolved import errors.

To eliminate the possibility of the name mismatch, this PR changes to set a library's `display_name` when consolidating crate specs. This change resolved the unresolved import errors of Rust Analyzer in [my reproduction repository of the issue](https://github.com/reiyw/rules_rust_1032_repro).

### DIFF
--- a/rust/private/rust_analyzer.bzl
+++ b/rust/private/rust_analyzer.bzl
@@ -145,6 +145,7 @@ def _create_single_crate(ctx, info):
     crate["display_name"] = crate_name
     crate["edition"] = info.crate.edition
     crate["env"] = {}
+    crate["crate_type"] = info.crate.type
 
     # Switch on external/ to determine if crates are in the workspace or remote.
     # TODO: Some folks may want to override this for vendored dependencies.
--- a/tools/rust_analyzer/BUILD.bazel
+++ b/tools/rust_analyzer/BUILD.bazel
@@ -41,6 +41,9 @@ rust_library(
 rust_test(
     name = "gen_rust_project_lib_test",
     crate = ":gen_rust_project_lib",
+    deps = [
+        "//tools/rust_analyzer/raze:itertools",
+    ],
 )
 
 rust_clippy(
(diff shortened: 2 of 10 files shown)


hlopko:
Hi @reiyw, thanks so much for working on this!

So, what do you think about instead of checking the crate type, we would check for presence of `wrapped_crate_type` - that would imply that the crate at hand is actually a wrapper like rust_test using `crate` attribute or potentially other rules that are not part of the rules_rust, but use the integratoin APIs of rules_rust. In case of a merge event, we pick the display name of the crate that does not have `wrapped_crate_type`.

For a merge event of `rust_binary` and `rust_library`, I agree that library should win, but is this actually a situation that happens in practice? I'd expect that rust_binary will have its root in `main.rs` file, and rust_library in `lib.rs` file, so there won't be a conflict. I'd be fine with discussing/solving this in a followup PR, since the rust_library/rust_test problem is far more pressing.

rust/private/rust_analyzer.bzl
```diff
@@ -145,6 +145,7 @@ def _create_single_crate(ctx, info):
     crate["display_name"] = crate_name
     crate["edition"] = info.crate.edition
     crate["env"] = {}
+    crate["crate_type"] = info.crate.type
```
hlopko, line 148: See the suggestion in the toplevel PR comment about `wrapped_crate_type`

hlopko:
Hmm, I changed my mind, let's go with the current PR approach.

hlopko (approved):

hlopko:
Thank you again!!

PR #1045, opened 2021-11-27
Provide CcInfo with stdlib_linkopts from rust_toolchain

### LINKED ISSUE #951: Deduplicate platform-specific linker arguments

hlopko:
I believe this is the most principled solution to the problem discussed in #870, #951, #955, and #1030.

The problem was that `rust_library`-like rules need to provide `CcInfo`, they need to pass `stdlib_linkflags` to the linker, and they didn't know if there was another `rust_library`-like rule that did it already, so they had to add those flags themselves.

This PR solves the problem by providing a `CcInfo` from the `rust_toolchain`. All `rust_library`-like rules can depend on this provider, and luckily, `user_link_flags` will then be expanded only once.

Closes #951, #955, #1030

If you were using the `rust_toolchain.stdlib_linkflags` field before, please replace it with `rust_toolchain.cc_info.linking_context.linker_inputs.to_list()[0].user_link_flags`.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -231,6 +231,18 @@ def _rust_toolchain_impl(ctx):
         fail("Do not specify both target_triple and target_json, either use a builtin triple or provide a custom specification file.")
 
     make_rust_providers_target_independent = ctx.attr._incompatible_make_rust_providers_target_independent[IncompatibleFlagInfo]
+    linking_context = cc_common.create_linking_context(
+        linker_inputs = depset([
+            cc_common.create_linker_input(
+                owner = ctx.label,
+                user_link_flags = depset(ctx.attr.stdlib_linkflags),
+            ),
+        ]),
+    )
+    stdlib_linkflags_cc_info = CcInfo(
+        compilation_context = cc_common.create_compilation_context(),
+        linking_context = linking_context,
+    )
 
     toolchain = platform_common.ToolchainInfo(
         rustc = ctx.file.rustc,
@@ -246,7 +258,15 @@ def _rust_toolchain_impl(ctx):
         binary_ext = ctx.attr.binary_ext,
         staticlib_ext = ctx.attr.staticlib_ext,
         dylib_ext = ctx.attr.dylib_ext,
+        # Contains linker flags needed to link Rust standard library.
+        # These need to be added to linker command lines when the linker is not rustc
+        # (rustc does this automatically). Only use these flags for direct actions and
+        # Use `stdlib_linkflags_cc_info for propagation (otherwise `stdlib_linkflags`
+        # may end up being unnecessailry duplicated).
         stdlib_linkflags = ctx.attr.stdlib_linkflags,
+        # Empty `CcInfo` whose only purpose is to provide `stdlib_linkflags` in a way
+        # that doesn't duplicate them per target providing a `CcInfo`.
+        stdlib_linkflags_cc_info = stdlib_linkflags_cc_info,
         target_triple = ctx.attr.target_triple,
         exec_triple = ctx.attr.exec_triple,
         os = ctx.attr.os,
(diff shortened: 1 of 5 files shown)


hlopko:
cc @ddeville

hlopko:
cc @sayrer

rust/toolchain.bzl
```diff
@@ -246,7 +258,15 @@ def _rust_toolchain_impl(ctx):
         binary_ext = ctx.attr.binary_ext,
         staticlib_ext = ctx.attr.staticlib_ext,
         dylib_ext = ctx.attr.dylib_ext,
+        # Contains linker flags needed to link Rust standard library.
+        # These need to be added to linker command lines when the linker is not rustc
+        # (rustc does this automatically). Only use these flags for direct actions and
+        # Use `stdlib_linkflags_cc_info for propagation (otherwise `stdlib_linkflags`
+        # may end up being unnecessailry duplicated).
         stdlib_linkflags = ctx.attr.stdlib_linkflags,
```
UebelAndre, line 266: Kinda surprised this isn't available via [CcInfo](https://docs.bazel.build/versions/4.0.0/skylark/lib/CcInfo.html) somehow (just looking at `compilation_context` and `linking_context`) or [cc_common](https://docs.bazel.build/versions/4.0.0/skylark/lib/cc_common.html). Do you know if there actually is a way to get the linker flags again in Starlark? Not a huge deal if not but I think it'd be nice to allow the `stdlib_linkflags_cc_info` attribute to be the single source of truth.

UebelAndre (approved):
Pretty slick! LGTM and Thanks for working on this! 🚀

rust/toolchain.bzl
hlopko: Good idea. You can access the flags via `cc_info.linking_context.linker_inputs[0].user_link_flags`. I feel like people don't rely directly on `ToolchainInfo` from `rust_toolchain`, so I'd just replace the old field with the new one. What do you think?

hlopko: (done in the latest commit)

UebelAndre: Since it's part of the public API, I think it might be worth having some kind of deprecation notice. Though, I think if you were to add the migration story to the PR description then it wouldn't be the end of the world if anyone decides to pick up this change.


Also, Taking a second look at this, I wonder if it'd be worth just having a `cc_info` field on `rust_toolchain` to allow the toolchain to be more flexible? Would there be additional uses for `CcInfo` aside from housing link flags?

hlopko: Added a note to the PR description.

Re one big `CcInfo` from `rust_toolchain`, that's the vision yes. But I'd leave that to a followup PR as I expect that will need a bit more investigation and added test coverage.

UebelAndre: Sounds good! Last question just to double check, `cc_info.linking_context.linker_inputs` [is reported to be a `depset`](https://docs.bazel.build/versions/4.0.0/skylark/lib/LinkingContext.html#linker_inputs) which I thought couldn't be indexed. Am I mistaken?

hlopko: Nice catch, yeah you have to `.to_list()` first. Updated the description.

ddeville:
Very nice, thank you for fixing this! Also I can confirm that I'm not seeing a large amount of duplicates and spawn errors anymore when running my project 😃

hlopko:
You can check the build speed before https://github.com/bazelbuild/rules_rust/commit/57243dc7286b79eecd012809d446beef1fd0042e and at HEAD. We see 4-5x speedup for some very nonstandard workloads (and near zero effect on standard workloads :)

PR #1047, opened 2021-11-30
Expose crate versions in CrateInfo

vmax:
We're trying to automatically collect crates.io dependencies (exposed with the help of `cargo raze`) of a `rust_library` target (see vaticle/bazel-distribution#323 for exact context)

While names of those crates are available by reading `target[CrateInfo].name`, versions were not. In order to help address this, we're extending `CrateInfo` provider to also contain `version` field

### DIFF
--- a/proto/proto.bzl
+++ b/proto/proto.bzl
@@ -229,6 +229,7 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, is_gr
         toolchain = find_toolchain(ctx),
         crate_info = rust_common.create_crate_info(
             name = crate_name,
+            version = ctx.attr.version,
             type = "rlib",
             root = lib_rs,
             srcs = depset(srcs),
--- a/rust/private/providers.bzl
+++ b/rust/private/providers.bzl
@@ -23,6 +23,7 @@ CrateInfo = provider(
         "edition": "str: The edition of this crate.",
         "is_test": "bool: If the crate is being compiled in a test context",
         "name": "str: The name of this crate.",
+        "version": "str: The version of this crate.",
         "output": "File: The output File that will be produced, depends on crate type.",
         "owner": "Label: The label of the target that produced this CrateInfo",
         "proc_macro_deps": "depset[DepVariantInfo|Target]: This crate's rust proc_macro dependencies' providers.",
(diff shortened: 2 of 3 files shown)


UebelAndre:
Version may not always be represented by a string. It can sometimes be generated via environment variables and can change if stamping is involved. I think if version is going to be tracked then it should be tracked as a file the contains the rendered output. Would that work for your uses?

hlopko:
As @UebelAndre wrote, this is a bit more complicated all corner cases considered. What do you think about using an [aspect](https://docs.bazel.build/versions/main/skylark/aspects.html) to traverse all `rust_library` rules while inspecting their `version` attribute (aspects can see attributes of the target they are visiting simply by `ctx.rule.attr.version`)? I think that could work very simply and would not require changes to rules.

vmax:
@UebelAndre @hlopko thank you for highlighting the issues with the approach we took.

We don't really expect the versions to change (they are statically generated by `cargo raze`), so it would be fine for us to retrieve them from the `version` attribute.

As suggested, we're handling it on our side with an aspect; if anybody is interested on how to do it, this is the snippet:

```
CrateInformation = provider(
    fields = {
        "name": "Crate name",
        "version": "Crate version",
        "deps": "Crate dependencies",
    },
)

aggregate_crate_information = aspect(
    attr_aspects = [
       "deps",
    ],
    doc = "Collects the Crate coordinates of the given rust_library and its direct dependencies",
    implementation = _aggregate_crate_information_impl,
    provides = [CrateInformation],
)

def _aggregate_crate_information_impl(target, ctx):
    return CrateInformation(
        name = ctx.rule.attr.name,
        version = ctx.rule.attr.version,
        deps = [target for target in getattr(ctx.rule.attr, "deps", [])]
    )
```

and then later you can read them in the rule implementation:

```
def _rule_impl(ctx):
    deps = {}
    for dependency in ctx.attr.target[CrateInformation].deps:
        deps[dependency[CrateInformation].name] = dependency[CrateInformation].version

...

example_rule = rule(
    implementation = _rule_impl,
    attrs = {
        "target": attr.label(
            mandatory = True,
            aspects = [aggregate_crate_information]
        ),
)
```

PR #1049, opened 2021-12-01
toolchain: support location expansion for stdlib_linkflags

krasimirgg:
This adds location expansion of stdlib_linkflags with respect to the
rust_lib attribute. This allows us to, e.g., have a linker flag that
refers to a native dependency target label declared in rust_lib.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -60,6 +60,7 @@ def _rust_stdlib_filegroup_impl(ctx):
             std_files = std_files,
             alloc_files = alloc_files,
             self_contained_files = self_contained_files,
+            all_targets = ctx.attr.srcs,
         ),
     ]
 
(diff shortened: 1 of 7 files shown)


rust/toolchain.bzl
```diff
@@ -60,6 +60,7 @@ def _rust_stdlib_filegroup_impl(ctx):
             std_files = std_files,
             alloc_files = alloc_files,
             self_contained_files = self_contained_files,
+            all_targets = ctx.attr.srcs,
```
hlopko, line 63: Nit: I'd go with `all_files` to be consistent with other attributes here (and also to be consistent with `cc_toolchain`).

hlopko (approved):

rust/toolchain.bzl
krasimirgg: As discussed offline, reverted back to using the original `srcs` attribute for this since it's a list of targets, which makes it a bit different from the other attributes of this provider.

hlopko (approved):
Thank you!

PR #1052, opened 2021-12-01
Don't add dylibs to runfiles of rlibs

scentini:
They are not needed, and flattening `transitive_noncrates` is expensive


scentini:
We actually don't need to manually add dylibs to runfiles at all, as we create runfiles with `collect_data = True`.

hlopko (approved):

Issue #1051, opened 2021-12-01
incompatible_remove_transitive_libs_from_dep_info


scentini:
Currently `DepInfo` has a `transitive_libs: List[File]` field that contains a list of all transitive dependencies of the crate. In order to populate this field, at analysis time we need to flatten `DepInfo.direct_crates`, `DepInfo.transitive_noncrates`, as well as any `CcInfo.linking_context.linker_inputs`. This leads to quadratic behavior.

The idea is to deprecate `DepInfo.transitive_libs` in favor of a new field - `DepInfo.transitive_crate_outputs : Depset[File]` that will keep the transitive crate outputs that can then be passed as inputs to the rust compile action without flattening. For the non-crate transitive outputs we will still have to flatten `DepInfo.transitive_noncrates`, however we do not need those for compiling `rlib`s, thus we should get significant memory improvements.

PR #1054, opened 2021-12-02
Remove uses of `DepInfo.transitive_libs`

scentini:
This is a step further towards https://github.com/bazelbuild/rules_rust/issues/1051.

With this change, `rules_rust` doesn't need `DepInfo.transitive_libs` anymore. The only place where this field was used was to pass the files it contains as inputs to rust compile actions. We use `DepInfo.transitive_crate_outputs` for that now. For non-rlib compile actions, which need non-crate inputs as well, we add those inputs to the `additional_transitive_inputs` variable.


test/unit/native_deps/native_action_inputs_test.bzl
```diff
@@ -0,0 +1,152 @@
...
+    "rust_library",
+    "rust_proc_macro",
+    "rust_shared_library",
+    "rust_static_library",
+)
+
+def _native_action_inputs_present_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    action = tut.actions[0]
```
hlopko, line 16: Nit: use assert_action_mnemonic from //test/unit:common.bzl to double check we are looking into the right action.

hlopko (approved):

test/unit/native_deps/native_action_inputs_test.bzl
scentini: Done.

Issue #1013, opened 2021-11-12
rename_first_party_crates flag


cfredric:
This is a tracking issue for the idea described in https://github.com/bazelbuild/rules_rust/discussions/1008.

cfredric:
More detail:

Google has some Rust in its source repository, and we have realized that rules_rust's current approach to naming crates (i.e., using the target name as the crate name) will not work for us. We need to use a naming scheme that uniquely identifies each first-party crate, and associates it with the entire Bazel label that it was built from; for third-party crates, we can use the crate's actual name (without modification). The motivations for this are the following:

* Facilitate automated updates of BUILD files based on .rs source file content, without requiring `bazel query`.
  * E.g., when the developer adds an import to a Rust source file and saves it, we can run tooling that scans the source, identifies the new dependency, and adds the dep to the correct target in the BUILD file.
* Avoid the need for the `aliases` attribute to resolve naming conflicts.
  * E.g., if a crate has dependencies `//foo:utils` and `//bar:utils`, this will cause a conflict unless the deps are renamed via `aliases` in the BUILD file. Use of the `aliases` attribute requires repetition of each label being renamed, and is in the BUILD file. We expect developers to spend more time reading .rs files than BUILD files, so we'd prefer to move that mapping into the Rust source, and eliminate the need for the developer to manually update both occurrences of the dependency's label (i.e. in the `deps` and `aliases` attributes).
* Facilitate tool-driven large-scale refactorings.
  * I.e., we'd like to be able to update relevant BUILD files without having to scan through all of them.

To achieve those goals, we need the following properties:
 * References to first-party crates must encode their whole Bazel label, in an invertible fashion. (This doesn't necessarily need to be human readable, since `extern crate` statements can be generated in Rust source by a proc macro. But it's nice to make it human readable for easier debugging, etc.)
 * References to third-party crates (in first-party code) must also encode the whole Bazel label, to enable automated BUILD file updates.
 * References to third-party crates (in third-party code) must not need to be changed. I.e., vendored third-party crates must not be renamed to encode the full Bazel label.

We're experimenting with a proc_macro to achieve this. Example usage of the macro could be the following:

```rust
import! {
  "//some_project:utils";
  "//some_project:ugly_name" as something_else;
  "//third_party/rust/serde";
  "//third_party/rust/clap" as aliased_clap;
}
```

This would expand to:

```rust
extern crate some_project_COLON_utils as utils;
[...]

PR #1056, opened 2021-12-03
Add rename_first_party_crates and third_party_dir flags

cfredric:
This PR adds a flag to enable renaming of first-party crates (to encode their package and target, rather than just the target) and a flag to control which crates are considered third-party vs. first-party. This PR addresses https://github.com/bazelbuild/rules_rust/issues/1013, related to https://github.com/bazelbuild/rules_rust/discussions/1008.

There's a TODO still in the code I added: I wasn't able to figure out how to get end-to-end tests to override the values of those flags, so the end-to-end tests are mostly commented out for the moment. But I tried to get good unit test coverage (both analysistest and unittest) to compensate.


rust/private/BUILD.bazel
```diff
@@ -1,6 +1,7 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("//rust/private:rust_analyzer.bzl", "rust_analyzer_detect_sysroot")
 load("//rust/private:stamp.bzl", "stamp_build_setting")
+load(":utils_test.bzl", "utils_test_suite")
```
hlopko, line 4: Could you please move this file to `//test/unit`? That's where we typically leave all our tests. There is a potential future where we will be creating `rules_rust` releases and those won't have tests bundled. If we separate all tests to a side it will be easie achieve. If you're fighting with buildifier visibility, it is fine to disable buildifier visibility check for the test (`# buildifier: disable=bzl-visibility`)

rust/private/BUILD.bazel
```diff
@@ -19,3 +20,5 @@ rust_analyzer_detect_sysroot(
     name = "rust_analyzer_detect_sysroot",
     visibility = ["//visibility:public"],
 )
+
+utils_test_suite(name = "utils_tests")
```
hlopko, line 24: (ditto, please move to `//test/unit`)

rust/private/utils.bzl
```diff
@@ -252,11 +252,13 @@ def _invalid_chars_in_crate_name(name):
 
     return dict([(c, ()) for c in name.elems() if not (c.isalnum() or c == "_")]).keys()
 
-def crate_name_from_attr(attr):
+def compute_crate_name(attr, label, toolchain):
```
hlopko, line 255: I think we can simplify these parameters a little. We use `attr` to read `name` and `crate_name`. However, `name` can also be read from the `label`. And, passing `ctx.attr` around is a bit of an antipattern (unfortunately it happens a lot in `rules_rust`). 

So what I'd do is pass `label`, `toolchain` as mandatory parameters, and `name_override` as an optional parameter defaulting to `None`. Callers of this function having `crate_name` attribute can call `getattr(ctx.attr, "crate_name", None)` and pass result to `compute_crate_name` without too much hassle.

rust/private/utils.bzl
```diff
@@ -270,7 +272,14 @@ def crate_name_from_attr(attr):
             ))
         return attr.crate_name
 
-    crate_name = name_to_crate_name(attr.name)
+    if toolchain and label and toolchain._rename_1p_crates:
+        if is_third_party_package(label.package, toolchain._third_party_dir):
+            crate_name = attr.name
```
hlopko, line 277: `attr.name` can be `label.name`, also below.

rust/private/utils.bzl
```diff
@@ -379,3 +388,80 @@ def transform_deps(deps, make_rust_providers_target_independent):
         build_info = dep[BuildInfo] if BuildInfo in dep else None,
         cc_info = dep[CcInfo] if CcInfo in dep else None,
     ) for dep in deps] if make_rust_providers_target_independent else deps
+
+def is_third_party_package(package, third_party_dir):
```
hlopko, line 392: Nit: WDYT about `should_encode_label_in_crate_name`?

rust/private/utils.bzl
```diff
@@ -379,3 +388,80 @@ def transform_deps(deps, make_rust_providers_target_independent):
...
+    """Determines if the given package is considered third-party.
```
hlopko, line 393: Please add brief comment on what does it mean "third-party" -> "names of third-party crates do not contain the full label" or something in that spirit.

rust/private/utils.bzl
```diff
@@ -379,3 +388,80 @@ def transform_deps(deps, make_rust_providers_target_independent):
...
+
+    Args:
+        package (string): The package in question.
+        third_party_dir (string): The directory in which third-party packages are kept.
+
+    Returns:
+        True if the package is considered third-party, False otherwise.
+    """
+    return ("//" + package + "/").startswith(third_party_dir + "/")
```
hlopko, line 402: Please add `TODO(hlopko): This code assumes a monorepo; make it work with external repositories as well.

rust/private/utils.bzl
```diff
@@ -379,3 +388,80 @@ def transform_deps(deps, make_rust_providers_target_independent):
...
+    ("}", "rbrace"),
+    ("~", "tilde"),
+    ("/", "slash"),
+    (".", "dot"),
+)
+
+# For each of the above encodings, we generate two substitutions: one that
+# ensures any occurrences of the encodings themselves in the package/target
+# aren't clobbered by this translation, and one that does the encoding itself.
+_substitutions = [
```
hlopko, line 445: (no action needed :) I'm wondering if we should get https://crates.io/crates/aho-corasick Starlark API into Bazel :)

rust/settings/BUILD.bazel
```diff
@@ -9,6 +10,26 @@ incompatible_flag(
     issue = "https://github.com/bazelbuild/rules_rust/issues/966",
 )
 
+# A flag controlling whether to rename first-party crates such that their names
+# encode the Bazel package and target name, instead of just the target name.
+#
+# First-party vs. third-party crates are identified using the value of
+# //settings:third_party_dir.
+incompatible_flag(
```
hlopko, line 18: I assumed we will maintain both naming strategies for some time, so using an `incompatible_flag` for this right away may send a wrong impression to users (typically users have to migrate for `incompatible_flags` rather promptly). Would it make sense to you if we made this a bool_flag for the time being, and only added incompatible flag after long time and proper discussion with other owners?

That said, if we set `third_party_dir` to `//` by default, and we made sure it works for external repositories, we could enable the behavior of `rename_1p_crates` even without the incompatible flag today, it's just that most (all? except us?) users would still be using the third-party like behavior for all their workspaces. Let's not do this in this PR though.

test/rename_1p_crates/BUILD.bazel
```diff
@@ -0,0 +1,60 @@
+load("//rust:defs.bzl", "rust_library", "rust_test")
+
+rust_library(
```
hlopko, line 3: The ideal would be if the import! macro could deal with both states of `rename_1p_crates`, when the flag is disabled it would just assume crate names as `rules_rust` do today. With that we could have a passing integration test both ways.

But the macro is not yet upstreamed, so I think it's fine to delete these tests for the time being. What do you think?

hlopko (commented):
Great work @cfredric thank you!

UebelAndre (changes requested):
Sorry for chiming in so late in the game, I think the PR is coming along pretty well! 

I just wanted to echo a point brought up and also chat about https://github.com/bazelbuild/rules_rust/issues/1013#issuecomment-986900290 (wasn't sure if that would happen on the issue since there was a PR open now).

rust/private/BUILD.bazel
cfredric: Done.

rust/private/BUILD.bazel
cfredric: Done.

rust/private/utils.bzl
cfredric: Done.

rust/private/utils.bzl
cfredric: That's probably better (/more semantically accurate) in the context of multi-repo projects. Done.

rust/private/utils.bzl
cfredric: Done.

rust/private/utils.bzl
cfredric: Done. Would supporting external repositories be as simple as checking if the package is in a different repo, and not renaming in that case? If so, I can add that case in this PR.

rust/settings/BUILD.bazel
cfredric: Thanks, I wasn't sure if `incompatible_flag` was appropriate for this, but in hindsight that should have been clear. I didn't mean to propose this as "the only option, in the future", but as an option that will be supported. I'll make it a `bool_flag`, and rename it to `rename_first_party_crates` per https://github.com/bazelbuild/rules_rust/issues/1013#issuecomment-986900290, while I'm at it.

It's an interesting thought to make this the default/only behavior, and rely on people to use the third-party opt-out. But I agree, I think that kind of change should not be part of this PR.

test/rename_1p_crates/BUILD.bazel
cfredric: The macro I've been prototyping does take a `bool` argument that disables the renaming, so that it is effectively a no-op; that way all the relevant files in the repo can be migrated, then the macro definition can be modified so that it starts renaming everything all at once.

I think that could help in the integration test, though I'm not sure how to (implicitly) plumb that bool from the .bazelrc file into the macro invocation. Would that mean the macro has to read the .bazelrc file at compilation time?

Also - I'm happy to start upstreaming the macro prototype as well, if there's a way to depend on external crates here (aho_corasick, quickcheck, etc).

rust/private/utils.bzl
```diff
@@ -379,3 +388,85 @@ def transform_deps(deps, make_rust_providers_target_independent):
...
+
+    # TODO(hlopko): This code assumes a monorepo; make it work with external
+    # repositories as well.
+    return ("//" + package + "/").startswith(third_party_dir + "/")
+
+# This is a list of pairs, where the first element of the pair is a character
+# that is allowed in Bazel package or target names but not in crate names; and
+# the second element is an encoding of that char suitable for use in a crate
+# name.
+_encodings = (
```
mfarrugi, line 413: Is the encoded name supposed to be decodable, or just human readable?

rust/private/utils.bzl
hlopko: I think supporting external repos means mangling the repo name into the crate name, and checking the repo in the label of the third party package, and writing a bunch of tests :)

rust/settings/BUILD.bazel
hlopko: Agreed on all points.

test/rename_1p_crates/BUILD.bazel
hlopko: I think we can talk about it in the PR that adds the macro :) By the way external crates are possible (we use cargo raze for this). The potential idea for plumbing the boolean is to use an environment variable and read its value from the macro.

rust/private/utils.bzl
hlopko: I think only hard requirement is unique for a given label, not conflicting with a different label. Decodable/human readable is nice to have, but not required.

mfarrugi: This encoding lets degenerate naming clash, eg. 
```
//x_slash_y:z => x_slash_y_colon_z
//x/y:z       => x_slash_y_colon_z
```

I feel like a more naive encoding (with potentially more degenerate naming clashes) would be just as good and removes the need for a macro to resolve labels.
```
//x_slash_y:z => x__slash__y_z
//x/y:z       => x_y_z
//x/y/z       => x_y_z
//x/y/z:a     => x_y_z_a
```


A bit separately, I also feel like supporting a bunch of special characters in the label names is unnecessary. We may as well go the same route as python, which requires labels to be valid python module names (eg. dashes are not importable).

If the encoding must be unambiguous, and support the additional characters, then this should use something like base64 encoding instead.

cfredric: > This encoding lets degenerate naming clash, eg.

That's not true, actually. This encoding is invertible (in the [mathematical sense](https://en.wikipedia.org/wiki/Inverse_function)). There are unit tests in utils_test.bzl to demonstrate this (though I'm aware that a test is not a proof :)). I've added the cases you mentioned, to show that those namings do not clash under this encoding.

> I feel like a more naive encoding (with potentially more degenerate naming clashes) would be just as good and removes the need for a macro to resolve labels.

For our use case, avoiding name collisions is a hard requirement. Supporting the full range of possible Bazel labels is also a hard requirement. So for us, the encoding/decoding steps are both necessary.

> If the encoding must be unambiguous, and support the additional characters, then this should use something like base64 encoding instead.

I did consider base64 encoding, but decided against it, for a few reasons:
* Rust identifiers cannot start with a digit, so we'd have to artificially modify the base64-encoded name to make sure it started with a letter.
* Rust identifiers cannot include `=` characters, so we'd have to strip out any padding bytes.
* Rust identifiers cannot include `+`, `-`, or `/` characters (which are used by the [Base64 and Base64Url alphabets](https://datatracker.ietf.org/doc/html/rfc4648#section-5)), so we'd have to do a global search/replace for those chars after encoding anyway, with some better alternative. But note that all the other characters that are legal in a crate name are already in use in the base64 alphabet, so we'd have to do some multi-character encoding. (So the base64 approach is starting to look more like the current approach.)
* Base64-encoded crate names would show up when viewing the macro-expanded version of the Rust code, but would be entirely opaque to the user. It would be better to show the user something easier to understand.

For those reasons, I decided the current approach was simpler and better from a user experience perspective as well.

hlopko:
(Feel free to continue the discussion, I figured since this is an experimental code anyway it's OK to submit and let people try it locally)

PR #1061, opened 2021-12-09
Support `cargo:rustc-link-arg=FLAG` instruction in build script runner

ddeville:
Cargo 1.50 added support for `cargo:rustc-link-arg` and `cargo:rust-link-arg-bins` instructions in build script output (`rustc-cdylib-link-arg` has been supported since 1.35 and `cargo:rustc-link-arg-bin` was later added in 1.54).

The build script runner currently prints a warning when `rustc-cdylib-link-arg` is used but entirely ignores the other ones (likely setting them as `BuildScriptOutput::DepEnv` since this is the default branch).

This patch adds support for the `cargo:rustc-link-arg` instruction and prints an warning when `cargo:rust-link-arg-bins` and `cargo:rust-link-arg-bin` are used, just like `rustc-cdylib-link-arg`.

Note that `cargo:rustc-link-arg` is important because `cargo` errors when `cargo:rustc-flags` is used for anything that isn't a `-l` or `-L` flag (nb: this is enforced by cargo so it'd work with Bazel). There are cases (such as linking a `-weak_framework` or `-weak_library` on macos) that cannot use `cargo:rustc-link-lib` and passing a custom `-Clink-arg` is required)

I think that we should also support `cargo:rust-link-arg-bins`, `rustc-cdylib-link-arg` and `cargo:rustc-link-arg-bin` but that's a bigger task since it requires knowing the crate type the build output will apply to. I've made issue https://github.com/bazelbuild/rules_rust/issues/1062 for it.

Ref: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

### DIFF
--- a/cargo/cargo_build_script_runner/lib.rs
+++ b/cargo/cargo_build_script_runner/lib.rs
@@ -34,6 +34,8 @@ pub enum BuildScriptOutput {
     Cfg(String),
     /// cargo:rustc-flags
     Flags(String),
+    /// cargo:rustc-link-arg
+    LinkArg(String),
     /// cargo:rustc-env
     Env(String),
     /// cargo:VAR=VALUE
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Thanks!

PR #1064, opened 2021-12-13
use absolute value of hash function to determine crate name

krasimirgg:
Previously the library names of rlibs sometimes had two dashes, like
libslib--389267823.rlib.  This happens because the hash() function may
return a negative value, so the second dash is the minus sign. However
such names are a bit confusing since they look like there's a missing
component between the two dashes that didn't get properly propagated,
which is not true.
For the unit test, since the hash() function is the standard Java String hashCode
function, I chose "slib"  since that produces a negative hash code.


test/unit/crate_name/crate_name_test.bzl
```diff
@@ -56,6 +56,15 @@ def _invalid_custom_crate_name_test_impl(ctx):
     asserts.expect_failure(env, "contains invalid character(s): -")
     return analysistest.end(env)
 
+# Regression test to check that the extra hash value appended to the library
+# filename only contains one dash. Previously, the hash for `slib` was negative,
+# resulting in an extra dash in the filename (--codegen_extra_filename=--517943325).
+def _slib_library_name_test_impl(ctx):
```
UebelAndre, line 62: nit: can the comment here be a docstring?

UebelAndre (approved):
The changes look good to me! Thanks!

PR #1065, opened 2021-12-13
Provide pdb file for cydlib/bin on Windows in a `pdb_file` output group

ddeville:
I've chose the `pdb_file` name for the output group since it matches what `cc_binary` uses
(https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java;l=707)

There is one edge case that is not covered here: if the user passes `/DEBUG:NONE` in the `rustc_flags` no pdb will be generated and I imagine the build will fail since it's now missing an output. I guess we could check the `rustc_flags` for it but I don't know if it's the right place to do so?


ddeville:
Looks like the test launcher adds its own `OutputGroupInfo` provider, let me see if those can be combined...

UebelAndre:
The launcher is actually something I think we need to remove. I can ad an incompatibility flag for it

UebelAndre:
> The launcher is actually something I think we need to remove. I can ad an incompatibility flag for it

Sorry this took so long. I've opened https://github.com/bazelbuild/rules_rust/pull/1070

UebelAndre (approved):
Thanks for putting this together, Awesome work! 😄

test/unit/win_interface_library/win_interface_library_analysis_test.bzl
```diff
@@ -36,6 +36,12 @@ def win_interface_library_analysis_test_suite(name):
         target_compatible_with = ["@platforms//os:windows"],
     )
 
+    native.filegroup(
```
UebelAndre, line 39: Can you add a filegroup for a `rust_binary` target as well?

UebelAndre (approved):
Oh, one last change, if you don't mind 😅

test/unit/win_interface_library/win_interface_library_analysis_test.bzl
ddeville: sure can do! now let's see if the tests still pass 🤞🏼

ddeville: ah derp I added it to the wrong binary type, give me a sec.

UebelAndre (approved):
Thank you so much!

Issue #405, opened 2020-09-15
Consistent Library Names


UebelAndre:
I've come across this enough and feel like I should post an issue here. 

Can someone provide some justification for why libraries produced by `rust_library` include a hash in them ([see @io_bazel_rules_rust//rust/private:rust.bzl](https://github.com/bazelbuild/rules_rust/blob/master/rust/private/rust.bzl#L30-L58)). My expectation is that the name would simply be the name of the target plus the appropriate extension. 

```rust
// Current
"libmy_target--451414955.so"
```

vs

```rust
// Expectation
"libmy_target.so"
```

I can appreciate the hash but I feel like that should instead be a symlink target or maybe a symlink should be created that uses the simple name. Maybe the hash can just be some metadata file generated with the library. Either way, I find this to be more cumbersome than helpful. 

Interested in hearing other peoples thoughts though. Again, I'd love to hear why this is the way that it is.

acmcarther:
Yo, undead maintainer here. Thanks for your comments and PRs. This change dates back to this PR. 
https://github.com/bazelbuild/rules_rust/pull/47

I can't remember the exact problem this solves, but it's something like "rustc can't handle libraries with the exact same filename as build inputs". Before that PR, nobody knew it was an issue because nobody could compile crates from the cargo ecosystem. In the early version of raze, I found this issue, and implemented this hack.

This comment is probably the most specific framing of it
https://github.com/bazelbuild/rules_rust/pull/47#issuecomment-365492191

---

Side thing. In this case, I knew I was probably the only person that could answer this, but didn't know exactly what the answer was. Here's how I found the answer from my past self:

1. From [your link](https://github.com/bazelbuild/rules_rust/blob/master/rust/private/rust.bzl#L30-L58) I found the "determine_output_hash", which I knew was the most relevant snippet. I hit the `blame` button and saw Marco's PR here: https://github.com/bazelbuild/rules_rust/pull/123 with this commit: https://github.com/bazelbuild/rules_rust/commit/d33bf1d2d9f7f66f4c3ba854949bcf88162e36ee
2. I knew a priori that this wasn't the initial PR (you'd have to determine this yourself in a similar situation), so I went to the commit, clicked the "parent", which is this commit: https://github.com/bazelbuild/rules_rust/commit/eb4681914a5c7abd99b48c600125c3f1794594fe
3. From the parent, I browsed files
[...]

PR #1066, opened 2021-12-15
Don't use output hash in filename for crates of type cdylib

ddeville:
Cdylib crates, similarly to binaries, are meant for consumption outside of the rust ecosystem. For this reason, there is no need to attempt making their name unique to disambiguate them from other crates with the same name but different features/versions in the Rust/Cargo dependency graph (as discussed in https://github.com/bazelbuild/rules_rust/issues/405).

In fact, when distributing dynamic libraries or embedding them into applications, it is *critical* for their name to be deterministic since it ends up embedded into the executable itself, that the linker uses to link/resolve the library. On macOS, this requires the use of `install_name_tool` to tweak the id/name after it's build. On Windows, one has to re-generate the LIB file to embed the new name (which afaict can only be done via the hacky intermediary export to a DEF file) and live with the fact that the associated pdb will embed the previous library name.

This patch just doesn't use the hash for cdylib library crates. It solves the problem above and all tests still pass so I assume it's not breaking anything crucial that was depending on this before 😃


hlopko:
Looks good, could you please add a test?

rust/private/rust.bzl
```diff
@@ -61,14 +61,14 @@ def _assert_correct_dep_mapping(ctx):
...
-def _determine_lib_name(name, crate_type, toolchain, lib_hash = ""):
+def _determine_lib_name(name, crate_type, toolchain, lib_hash = None):
     """See https://github.com/bazelbuild/rules_rust/issues/405
 
     Args:
         name (str): The name of the current target
         crate_type (str): The `crate_type`
         toolchain (rust_toolchain): The current `rust_toolchain`
-        lib_hash (str, optional): The hashed crate root path. Defaults to "".
+        lib_hash (str, optional): The hashed crate root path. Defaults to None.
```
UebelAndre, line 71: No need to double document the default 😅 (this is my fault for using an auto-docs tool)
```suggestion
        lib_hash (str, optional): The hashed crate root path.
```

UebelAndre (approved):
Small nit from me but otherwise looks good!

test/unit/cdylib_name/cdylib_name_analysis_test.bzl
```diff
@@ -0,0 +1,42 @@
...
+    target = analysistest.target_under_test(env)
+
+    file = target[DefaultInfo].files.to_list()[0]
+
+    if file.extension == "so":
+        asserts.equals(env, file.basename, "libsomething.so")
+    if file.extension == "dylib":
+        asserts.equals(env, file.basename, "libsomething.dylib")
+    if file.extension == "dll":
+        asserts.equals(env, file.basename, "something.dll")
```
UebelAndre, line 17: These would be mutually exclusive, right?
```python
    if file.extension == "so":
        asserts.equals(env, file.basename, "libsomething.so")
    elif file.extension == "dylib":
        asserts.equals(env, file.basename, "libsomething.dylib")
    elif file.extension == "dll":
        asserts.equals(env, file.basename, "something.dll")
    else:
        analysistest.fail(env, "unexpected file extension for file: {}".format(file.basename))
```

test/unit/cdylib_name/cdylib_name_analysis_test.bzl
```diff
@@ -0,0 +1,42 @@
+"""Analysis tests for the name we assign to cdylib libraries."""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+load("//rust:defs.bzl", "rust_shared_library")
+
+def _cdylib_name_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    target = analysistest.target_under_test(env)
+
+    file = target[DefaultInfo].files.to_list()[0]
```
UebelAndre, line 10:
```suggestion
    # Expecting the `cdylib` to be the one and only file.
    files = target[DefaultInfo].files.to_list()
    asserts.equals(env, len(files), 1)
    file = files[0]
```

UebelAndre (changes requested):
Sorry to flip-flop 😅 I noticed two things before merging

test/unit/cdylib_name/cdylib_name_analysis_test.bzl
ddeville: on windows there will also be an interface library but I can do the appropriate check in each branch.

UebelAndre: I'm a big fan of checks that extract what's expected as explicitly as possible. If the files you're after aren't in an `OutputGroupinfo` provider or something, then I think there should be an assert that you can find a unique file from a depset. In general I've found asserting on the number of elements in the depset to cover most edge cases but can be annoying for volatile rules (don't think that's the case here).

ddeville: ok I've shuffled things a bit so that we check for the file count and all files in the group are what they are expected.

test/unit/cdylib_name/cdylib_name_analysis_test.bzl
```diff
@@ -0,0 +1,46 @@
...
+    # contain a pair of `.dll` and `.lib` files.
+    files = target[DefaultInfo].files.to_list()
+    if len(files) == 1:
+        asserts.true(env, files[0].extension in ("so", "dylib"))
+        if files[0].extension == "so":
+            asserts.equals(env, files[0].basename, "libsomething.so")
+        elif files[0].extension == "dylib":
+            asserts.equals(env, files[0].basename, "libsomething.dylib")
+    elif len(files) == 2:
+        asserts.equals(env, files[0].basename, "something.dll")
```
UebelAndre, line 20: This might be a minor thing but I'd like to avoid creating a dependency on the ordering of the files returned by `DefaultInfo`. I think everyone should do due-diligence to ensure the files they're trying to get out of a list is what they want. Depending on `depset().to_list()[INDEX]` is something I'm hoping we can avoid
```suggestion
        expected_extensions = ["dll", "lib"]
        for file in files:
            asserts.true(env, file.extension in expected_extensions)
            expected_extensions.pop(file.extension)
```

UebelAndre (changes requested):

test/unit/cdylib_name/cdylib_name_analysis_test.bzl
ddeville: oh right that's fair. I'll change it to `expected_filenames` so that we can actually test the cdylib name too in one pass.

UebelAndre: Sorry, my understanding of `pop` seems to be wrong. I think I meant [remove](https://docs.bazel.build/versions/4.0.0/skylark/lib/list.html#remove)

ddeville: oops, fixed (that'll teach me not to test my windows changes on linux, I guess that's why we have CI 😄)

UebelAndre (approved):

PR #1071, opened 2021-12-16
Define remote repositories with maybe

illicitonion:
This allows individual repositories to be overridden by binding
replacement repositories as long as the same name is used.

The names we give to repositories are not currently part of the API
contract of crate_universe, but this can still be useful.


UebelAndre (approved):
🚀

PR #1072, opened 2021-12-17
Provide `.dSYM` package in `OutputGroupInfo` for cdylib and bin targets on darwin

ddeville:
rustc emits a `.dSYM` package/folder when the `split-debuginfo=packed` codegen option is set (currently the default) so support that.
My original idea was to run `dsymutil` in my project on the produced binary from bazel but unfortunately it relies on some of the object files still being present in the build tree, which doesn't work since the absolute paths it records point to the bazel sandbox that is by then long gone. For this reason, relying on rustc to generate this file is the only correct option.

Note that while the cc rules use `pdb_file` in the `OutputGroupInfo` I wasn't able to find anything similar for macos. So I opted `dsym_folder` and added some extra docs to `rust_shared_library` and `rust_binary` so that it's easier to find.

Also reorganized the tests a bit so that the logic between `pdb_file` and `dsym_folder` is shared.


UebelAndre (approved):
🤯 You can have multiple `OutputGroupInfo` providers? And they get merged? Awesome!

Looks good to me! Thanks for updating the documentation too 🙏

Issue #251, opened 2019-09-20
The provided `example-bindgen-toolchain` rule does not find `libstdc++.so.6` on ArchLinux because of a broken symlink


dwtj:
I've made [this minimum working example Bazel workspace](https://github.com/bazelbuild/rules_rust/files/3634145/bazel_rust_bindgen.zip) (`.zip`) to demonstrate the problem I'm having building `rust_bindgen_library` targets on ArchLinux.

It just has a `rust_bindgen_library` rule pointing to a trivial `cc_library` and its header file. `bindgen_toolchain` resolves to the `example-bindgen-toolchain` (i.e. the one registered while calling the `rust_bindgen_repositories()` workspace rule).

I cannot `bazel build //...` this workspace on my ArchLinux system. I get the following error.

```
ERROR: missing input file '@local_libstdcpp//:libstdc++.so.6'
ERROR: /home/dwtj/.cache/bazel/_bazel_dwtj/b8ac5acf45adefed9426377fa8b7244a/external/local_libstdcpp/BUILD.bazel:2:1: @local_libstdcpp//:libstdc++: missing input file '@local_libstdcpp//:libstdc++.so.6'
ERROR: /home/dwtj/.cache/bazel/_bazel_dwtj/b8ac5acf45adefed9426377fa8b7244a/external/local_libstdcpp/BUILD.bazel:2:1 1 input file(s) do not exist
```

The problem appears to be because `@local_libstdcpp`'s symlink to `libstdc++.so` appears to be broken.

```
lrwxrwxrwx 1 dwtj dwtj 40 Sep 19 15:37 /home/dwtj/.cache/bazel/_bazel_dwtj/b8ac5acf45adefed9426377fa8b7244a/external/local_libstdcpp/libstdc++.so.6 -> /usr/lib/x86_64-linux-gnu/libstdc++.so.6
```

But `libstdc++.so.6` is not located at that location on my ArchLinux system. It is at `/usr/lib/libstdc++.so.6`.

The problem appears to be with an assumption about Linux made here:

https://github.com/bazelbuild/rules_rust/blob/f727669b8ac3c9d237ed9bc7833b8e1eeec90506/bindgen/repositories.bzl#L76-L79

Presumably this is the usual location of `libstdc++.so.6` on other distros.

I realize that I can work around this issue by creating and registering my own `bindgen_toolchain` instance, following `example-bindgen-toolchain` as a guide. That's my next move.

[...]

PR #1076, opened 2021-12-22
Always use a param file when calling into rustc

ddeville:
Transitive crate dependencies are passed to rustc by specifying the folder they live in as a library search path (an `-Ldependency=path` flag). Given that bazel builds all libraries into their own folders, this means passing potentially hundreds of paths to rustc which can lead to overflowing the rather short limit for command arguments on Windows.

For this reason, this patch creates symlinks to all the transitive crates into a single directory that it then sets as the `-Ldependency=` flag to rustc. This dramatically shortens the command arguments length when calling rustc. Note that we only do that on Windows since it introduces some prep actions that need to do IO and there is no need to regress other platforms that support much longer command args.

I opted for declaring these files as inputs and creating the symlinks via `ctx.actions.symlink` but another option would have been to do that in the process wrapper instead.

For context, here is the size returned by `getconf ARG_MAX` on my following machines:
```
macos (12 x86_64):      1,048,576
linux (arch x86_64):    2,097,152  // obviously somewhat lower on 32-bit but
I don't have a machine handy
windows (10 x86_64):    32,000
```

### DIFF
--- a/rust/private/rustdoc.bzl
+++ b/rust/private/rustdoc.bzl
@@ -49,7 +49,8 @@ def rustdoc_compile_action(
         crate_info,
         output = None,
         rustdoc_flags = [],
-        is_test = False):
+        is_test = False,
+        use_param_file_for_rustc_args = True):
     """Create a struct of information needed for a `rustdoc` compile action based on crate passed to the rustdoc rule.
 
     Args:
@@ -59,6 +60,7 @@ def rustdoc_compile_action(
         output (File, optional): An optional output a `rustdoc` action is intended to produce.
         rustdoc_flags (list, optional): A list of `rustdoc` specific flags.
         is_test (bool, optional): If True, the action will be configured for `rust_doc_test` targets
+        use_param_file_for_rustc_args(bool, optional): Whether Bazel should use a param file to pass the arguments to rustc.
 
     Returns:
         struct: A struct of some `ctx.actions.run` arguments.
@@ -118,6 +120,7 @@ def rustdoc_compile_action(
         emit = [],
         remap_path_prefix = None,
         force_link = True,
+        use_param_file_for_rustc_args = use_param_file_for_rustc_args,
     )
 
     return struct(
--- a/util/process_wrapper/process_wrapper.cc
+++ b/util/process_wrapper/process_wrapper.cc
@@ -21,14 +21,41 @@
 #include "util/process_wrapper/system.h"
 #include "util/process_wrapper/utils.h"
 
-using CharType = process_wrapper::System::StrType::value_type;
+using namespace process_wrapper;
+using CharType = System::StrType::value_type;
+
+static int process_param_file(System::StrType& path,
+                              System::Arguments& collected_args,
+                              const std::vector<Subst>& subst_mappings) {
+  System::StrVecType args;
+  if (!ReadFileToVec(path, args)) {
+    std::cerr << "Cannot read param file " << ToUtf8(path) << "\n";
+    return -1;
+  }
+  for (System::StrType& arg : args) {
+    // Expand tokens into each argument (since some paths might embed `${pwd}`
+    // that we need to resolve from the subsitution mappings).
+    ReplaceTokens(arg, subst_mappings);
+
+    // If this is a param file, expand it into the top-level arguments by
+    // recursing into its content.
+    if (arg.size() > 0 && arg.substr(0, 1) == PW_SYS_STR("@")) {
+      System::StrType path = arg.substr(1);
+      int res = process_param_file(path, collected_args, subst_mappings);
+      if (res != 0) {
+        return res;
+      }
+    } else {
+      collected_args.push_back(arg);
+    }
+  }
+  return 0;
+}
 
 // Simple process wrapper allowing us to not depend on the shell to run a
 // process to perform basic operations like capturing the output or having
 // the $pwd used in command line arguments or environment variables
 int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
-  using namespace process_wrapper;
-
   System::EnvironmentBlock environment_block;
   // Taking all environment variables from the current process
   // and sending them down to the child process
@@ -37,8 +64,6 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
   }
 
   System::EnvironmentBlock environment_file_block;
-
-  using Subst = std::pair<System::StrType, System::StrType>;
 
   System::StrType exec_path;
   std::vector<Subst> subst_mappings;
@@ -53,7 +78,7 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
   System::Arguments file_arguments;
 
   // Processing current process argument list until -- is encountered
-  // everthing after gets sent down to the child process
+  // everything after gets sent down to the child process
   for (int i = 1; i < argc; ++i) {
     System::StrType arg = argv[i];
     if (++i == argc) {
@@ -79,7 +104,8 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
       if (value == PW_SYS_STR("${pwd}")) {
         value = System::GetWorkingDirectory();
       }
-      subst_mappings.push_back({std::move(key), std::move(value)});
+      subst_mappings.push_back({PW_SYS_STR("${") + key + PW_SYS_STR("}"),
+                                std::move(value)});
     } else if (arg == PW_SYS_STR("--volatile-status-file")) {
       if (!volatile_status_file.empty()) {
         std::cerr << "process wrapper error: \"--volatile-status-file\" can "
@@ -90,11 +116,11 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
         return -1;
       }
     } else if (arg == PW_SYS_STR("--env-file")) {
-      if (!ReadFileToArray(argv[i], environment_file_block)) {
+      if (!ReadFileToVec(argv[i], environment_file_block)) {
         return -1;
       }
     } else if (arg == PW_SYS_STR("--arg-file")) {
-      if (!ReadFileToArray(argv[i], file_arguments)) {
+      if (!ReadFileToVec(argv[i], file_arguments)) {
         return -1;
       }
     } else if (arg == PW_SYS_STR("--touch-file")) {
@@ -143,7 +169,7 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
       for (++i; i < argc; ++i) {
         arguments.push_back(argv[i]);
       }
-      // after we consume all arguments we append the files arguments
+      // After we have consumed all arguments add the files arguments the them.
       for (const System::StrType& file_arg : file_arguments) {
         arguments.push_back(file_arg);
       }
@@ -154,13 +180,39 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
     }
   }
 
+  // Go through the arguments and make sure that we process param files so that
+  // we can flatten potential recursive param files within them.
+  for (System::StrType& arg : arguments) {
+    if (arg.size() > 0 && arg.substr(0, 1) == PW_SYS_STR("@")) {
+      // This is a param file, process it separately.
+      System::StrType path = arg.substr(1);
+      System::Arguments processed_args;
+      if (process_param_file(path, processed_args, subst_mappings) != 0) {
+        std::cerr << "Failure to process arguments list\n";
+        return -1;
+      }
+
+      // Create a new param file so that we don't need to edit the original one
+      // that Bazel created.
+      System::StrType new_path = path + PW_SYS_STR(".expanded");
+      if (!WriteVecToFile(new_path, processed_args)) {
+        std::cerr << "Cannot write params to file " << ToUtf8(new_path) << "\n";
+        return -1;
+      }
+
+      // Replace the params file arg with the new one.
+      arg = PW_SYS_STR("@") + new_path;
+    } else {
+      // Expand tokens into each argument (since some paths might embed `${pwd}`
+      // that we need to resolve from the subsitution mappings).
+      ReplaceTokens(arg, subst_mappings);
+    }
+  }
+
   // Stamp any format string in an environment variable block
-  for (const Subst& stamp : stamp_mappings) {
-    System::StrType token = PW_SYS_STR("{");
-    token += stamp.first;
-    token.push_back('}');
+  if (stamp_mappings.size()) {
     for (System::StrType& env : environment_file_block) {
-      ReplaceToken(env, token, stamp.second);
+      ReplaceTokens(env, stamp_mappings);
     }
   }
 
@@ -171,19 +223,9 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
                            environment_file_block.begin(),
                            environment_file_block.end());
 
-  if (subst_mappings.size()) {
-    for (const Subst& subst : subst_mappings) {
-      System::StrType token = PW_SYS_STR("${");
-      token += subst.first;
-      token.push_back('}');
-      for (System::StrType& arg : arguments) {
-        ReplaceToken(arg, token, subst.second);
-      }
-
-      for (System::StrType& env : environment_block) {
-        ReplaceToken(env, token, subst.second);
-      }
-    }
+  // Expand tokens in environment
+  for (System::StrType& env : environment_block) {
+    ReplaceTokens(env, subst_mappings);
   }
 
   // Have the last values added take precedence over the first.
--- a/util/process_wrapper/utils.h
+++ b/util/process_wrapper/utils.h
@@ -21,16 +21,20 @@
 
 namespace process_wrapper {
 
+using Subst = std::pair<System::StrType, System::StrType>;
+
 // Converts to and frin the system string format
 System::StrType FromUtf8(const std::string& string);
 std::string ToUtf8(const System::StrType& string);
 
-// Replaces a token in str by replacement
-void ReplaceToken(System::StrType& str, const System::StrType& token,
-                  const System::StrType& replacement);
+// Replaces tokens in str by replacement
+void ReplaceTokens(System::StrType& str, const std::vector<Subst>& substs);
 
 // Reads a file in text mode and feeds each line to item in the vec output
-bool ReadFileToArray(const System::StrType& file_path, System::StrVecType& vec);
+bool ReadFileToVec(const System::StrType& file_path, System::StrVecType& vec);
+
+// Write the content of an array file as lines in a file
+bool WriteVecToFile(const System::StrType& file_path, const System::StrVecType& vec);
 
 // Reads a workspace status stamp file to an array of key value pairs
 bool ReadStampStatusToArray(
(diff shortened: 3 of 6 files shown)


illicitonion:
`rustc` accepts passing arguments in a newline-delimited file:

```console
% rustc --help -v | grep '@path'
    @path               Read newline separated options from `path`
```

Would it work to just create a single text file with contents along the lines of:
```
-L
foo
-L
bar
```

rather than having to set up a whole directory which will be a lot of overhead (particularly on Windows where symlinks are often actually copies), and allow us to use a consistent command line across platforms?

I think we're possibly currently bypassing Bazel's built-in ability to do this [which we set up here](https://github.com/bazelbuild/rules_rust/blob/8c0ece8cbe0e246d3a87e8d9d4567cb7582be548/rust/private/rustc.bzl#L586-L588) because that ends up being an argfile to the process-wrapper rather than rustc itself?

ddeville:
Oh wow I looked around for this but completely missed it. Let me give this a try and see how that works, I'd also much rather avoid having to do all the work setting up symlinks…

ddeville:
So I'm doubly embarrassed because 1) I added this `use_param_file` for the rustc args a few weeks ago and 2) I managed to forget about the `@param_file` option to rustc since then...

Anyway, with that out of the way 😆, I took a look at how it gets propagated from the run action to rustc via the process wrapper. Turns out it just works, it's passed to the process wrapper as `libhello_lib-1405984315.rlib-0.params` and it forwards it to `rustc`, as expected.

What I realized however is that we have to set `use_always` to `True`. `use_param_file` is documented to automatically switch to a param file when needed if `use_always` is `False` but that's clearly not the case since I've been encountering args list way past the 32k character limit in the process wrapper on Windows.
However I was able to get my pathological large project to build fine on Windows by setting `use_always` to `True` so we should probably just change it.

This patch is gonna get a lot smaller :)

ddeville:
Hmm not sure where all these test failures come from, I'll take a look tomorrow morning.

illicitonion:
Sounds great - thanks so much for looking into it!

ddeville:
Quick update: I think I've got something working for the general case but am having trouble with `rust_doc_test` since it creates a shell script for the test that contain the arguments and in case they are written into a param file, it won't be in the runfiles folder. I can strip the substring but in this case the param file is not part of the runfiles folder at all (rather next to it) so I need to find a way to move it in there 🤔

ddeville:
I think I'm probably going to opt out the rustdoc tests from generating a param file when constructing the rustc arguments.

UebelAndre:
> I think I'm probably going to opt out the rustdoc tests from generating a param file when constructing the rustc arguments.

Yeah, `rust_doc_test` is currently in a weird state and I think there are a few contexts where it doesn't work. I think it'd be ok to defer updating it as long as the changes don't reduce the amount of places it currently works.

ddeville:
Looks like there are some build failures when building the examples.. I'll check on my macos machine since I can't seem to build the examples on my linux box due to failures to find libclang/libinfo as discussed in https://github.com/bazelbuild/rules_rust/issues/251.

ddeville:
Alright, it was missing a call to `ReplaceTokens`, I *think* everything should now build fine.

illicitonion (approved):
LGTM - thanks!

@UebelAndre @hlopko How's your C++? The C++ could probably do with better eyes than mine giving it a quick pass :)

UebelAndre:
I can take a look later today. Sorry for the delay 😅

util/process_wrapper/process_wrapper.cc
```diff
@@ -79,7 +77,8 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
       if (value == PW_SYS_STR("${pwd}")) {
         value = System::GetWorkingDirectory();
       }
-      subst_mappings.push_back({std::move(key), std::move(value)});
+      subst_mappings.push_back({PW_SYS_STR("${") + key + PW_SYS_STR("}"),
```
UebelAndre, line 80: Why is this necessary?

util/process_wrapper/process_wrapper.cc
```diff
@@ -143,24 +142,72 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
...
-        arguments.push_back(file_arg);
-      }
     } else {
       std::cerr << "process wrapper error: unknow argument \"" << ToUtf8(arg)
                 << "\"." << '\n';
       return -1;
     }
   }
 
+  for (System::StrType& file_arg : file_arguments) {
```
UebelAndre, line 152: nit: Can you add a small comment about why we do this?

rust/private/rustdoc.bzl
```diff
@@ -48,7 +48,8 @@ def rustdoc_compile_action(
         toolchain,
         crate_info,
         output = None,
-        rustdoc_flags = []):
+        rustdoc_flags = [],
+        use_param_file_for_rustc_args = True):
```
UebelAndre, line 52: Can you add a small unittest ensuring that when this flag is `True` prama files prefixed with `@` are found in argv of the action?

UebelAndre (changes requested):
Sorry for the delay, had to do a second pass. I think this is looking good! I had a couple questions and one request for a test but then I think it's ready to go 😄

util/process_wrapper/process_wrapper.cc
ddeville: We used to do this below and I've opted to just store the mapping as `${key}` rather than `key` since it's never used as `key` anyway.

https://github.com/bazelbuild/rules_rust/blob/82059917e2c7e69244d422478bdb73f8a6d5a7b9/util/process_wrapper/process_wrapper.cc#L176-L180

rust/private/rustdoc.bzl
ddeville: Good catch, will do.

util/process_wrapper/process_wrapper.cc
UebelAndre: I like it!

rust/private/rustdoc.bzl
ddeville: I just tried doing that but it looks like `Action.argv` returns the actual arguments for the action before they are turned into a file, rather than the arguments passed to the executable...

UebelAndre: Wow, that's kinda crazy. I feel like this is a bug in Bazel... Is there a way we can at least test two different actions with and without this flag?

UebelAndre (changes requested):

rust/private/rustdoc.bzl
UebelAndre: I suppose not. That's the point of the "always" in this PR

UebelAndre (approved):
This PR looks good to me. I'll leave it to @illicitonion to merge as a final sanity check 😅

hlopko:
Hello everybody, I'm back from a lovely long vacation and I'm also curious about this. Reviewing right now :)

hlopko:
> What I realized however is that we have to set use_always to True. use_param_file is documented to automatically switch to a param file when needed if use_always is False but that's clearly not the case since I've been encountering args list way past the 32k character limit in the process wrapper on Windows.
However I was able to get my pathological large project to build fine on Windows by setting use_always to True so we should probably just change it.

@meteorcloudy do you know if `use_param_file` works correctly on Windows? This is where we use it: https://cs.opensource.google/bazel/rules_rust/+/main:rust/private/rustc.bzl;l=588?q=rustc.bzl. Is this something we should file a bug over at bazelbuild/bazel?

hlopko:
@meteorcloudy  So one thing that is not so common in the ecosystem is that we actually pass 3 instances of args to the action, and only one of them has `use_param_file` set (https://cs.opensource.google/bazel/rules_rust/+/main:rust/private/rustc.bzl;l=690?q=rustc.bzl). Maybe `rules_rust` are the first such user?

rust/private/rustdoc.bzl
hlopko: Yup this is an unfortunate and a known limitation of unittesting args. You can access the original `args` value from the unittest (https://docs.bazel.build/versions/main/skylark/lib/Action.html#args), but you cannot do anything with it other than setting it as args into some other action. So the only way we could test this is by writing all these args into a file and registering an action that reads the file and makes assertions. In other words, you cannot use unittest, you need an integration test.

util/process_wrapper/utils.h
```diff
@@ -21,17 +21,21 @@
...
-void ReplaceToken(System::StrType& str, const System::StrType& token,
-                  const System::StrType& replacement);
+// Replaces tokens in str by replacement
+void ReplaceTokens(System::StrType& str, std::vector<Subst>& substs);
 
 // Reads a file in text mode and feeds each line to item in the vec output
 bool ReadFileToArray(const System::StrType& file_path, System::StrVecType& vec);
 
+// Write the content of an array file as lines in a file
+bool WriteArrayToFile(const System::StrType& file_path, System::StrVecType& vec);
```
hlopko, line 37: Nit: rename to WriteVecToFile?

util/process_wrapper/utils.h
```diff
@@ -21,17 +21,21 @@
...
+
 // Converts to and frin the system string format
 System::StrType FromUtf8(const std::string& string);
 std::string ToUtf8(const System::StrType& string);
 
-// Replaces a token in str by replacement
-void ReplaceToken(System::StrType& str, const System::StrType& token,
-                  const System::StrType& replacement);
+// Replaces tokens in str by replacement
+void ReplaceTokens(System::StrType& str, std::vector<Subst>& substs);
```
hlopko, line 31: I think `substs` (and also `vec` in `WriteArrayToFile`) can be `const`.

util/process_wrapper/process_wrapper.cc
```diff
@@ -143,24 +142,74 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
...
-      }
     } else {
       std::cerr << "process wrapper error: unknow argument \"" << ToUtf8(arg)
                 << "\"." << '\n';
       return -1;
     }
   }
 
+  // Expand tokens into the file arguments (since some paths might embed
+  // `${pwd}` that we need to resolve it from the subsitution mappings).
```
hlopko, line 153: Nit: remove `it`?

util/process_wrapper/process_wrapper.cc
```diff
@@ -143,24 +142,74 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
...
+  for (System::StrType& file_arg : file_arguments) {
+    ReplaceTokens(file_arg, subst_mappings);
+  }
+
+  // Process the arguments, making sure that we don't have paths to param files
+  // within another param file (that is an arg composed of `@` + a path), which
+  // rustc doesn't support.
+  // (We process them in reverse so that we can append the param files to the
```
hlopko, line 161: Nit: I think we can replace "append to the front" with "prepend"

util/process_wrapper/process_wrapper.cc
```diff
@@ -143,24 +142,74 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
...
+      if (!ReadFileToArray(path, params)) {
+        std::cerr << "Cannot read param file " << ToUtf8(path) << "\n";
+        return -1;
+      }
+
+      for (auto par_it = params.rbegin(); par_it != params.rend(); par_it++) {
+        System::StrType& param = *par_it;
+        ReplaceTokens(param, subst_mappings);
+
+        // If we find a path to a param file, move it to the top-level file
```
hlopko, line 183: Do we need this behavior? Could we just fail? I'm imagining a bug in rules_rust that produces nested param files and we wouldn't notice because process wrapper would recover from that. I also worry that moving param file args around can change the order of flags in a surprising way so I try to avoid it in general.

util/process_wrapper/process_wrapper.cc
```diff
@@ -143,24 +142,74 @@ int PW_MAIN(int argc, const CharType* argv[], const CharType* envp[]) {
...
+
+      // Create a new param file so that we don't need to edit the original one
+      // that Bazel created.
+      System::StrType new_path = path + PW_SYS_STR(".expanded");
+      if (!WriteArrayToFile(new_path, params)) {
+        std::cerr << "Cannot write params to file " << ToUtf8(new_path) << "\n";
+        return -1;
+      }
+
+      arguments.erase(std::next(arg_it).base());
```
hlopko, line 199: Oh this looks complicated. Would using `remove` or `remove_if` (https://en.cppreference.com/w/cpp/algorithm/remove) and https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom after the for loop ends be simpler to reason about?

I know it doesn't matter in our workloads, but even hypothethically repeatedly prepending into a vector makes me nervous. How about collecting new file arguments into a new vec and merging them once after the end of the loop?

hlopko (commented):
Overall looks good, but I'd like to wait for @meteorcloudy to get back to us about the root problem here. However the process wrapper change is necessary in any case - without it we don't handle param files correctly today. So thank you so much!

meteorcloudy:
> @meteorcloudy do you know if use_param_file works correctly on Windows?

I'm not quite familiar with this API, but I don't see any reason why it shouldn't work on Windows. You do have test coverage on Windows (and it seems passing), right?

oquenchil:
> > What I realized however is that we have to set use_always to True. use_param_file is documented to automatically switch to a param file when needed if use_always is False but that's clearly not the case since I've been encountering args list way past the 32k character limit in the process wrapper on Windows.
> > However I was able to get my pathological large project to build fine on Windows by setting use_always to True so we should probably just change it.
> 
> @meteorcloudy do you know if `use_param_file` works correctly on Windows? This is where we use it: https://cs.opensource.google/bazel/rules_rust/+/main:rust/private/rustc.bzl;l=588?q=rustc.bzl. Is this something we should file a bug over at bazelbuild/bazel?

It might be a bug where on Windows it doesn't decide properly. Maybe just use_always=True for now and you can file a bug on bazelbuild/bazel.

ddeville:
Welcome back @hlopko, I hope you had a relaxing break 😃.

util/process_wrapper/utils.h
ddeville: I was matching `ReadFileToArray` above but agreed, we should probably rename both.

util/process_wrapper/process_wrapper.cc
ddeville: That was also my first intuition but it turns out that we have a test that specifically checks that extra `@param_file_path` arguments are supported, by adding such an arg to the `rustc_flags`:
https://github.com/bazelbuild/rules_rust/blob/b16c26ba5faf1c58ebe94582afd20567ce676e6d/test/unit/location_expansion/location_expansion_test.bzl#L31

I wasn't able to gather the rationale for supporting this in the PR #809 or the associated issue #801, but there's a comment from you on that review that might be worth re-discussing here 😉 https://github.com/bazelbuild/rules_rust/pull/809/files#r662057446

util/process_wrapper/process_wrapper.cc
ddeville: Yeah that's a fair callout. I'll look at simplifying that.

hlopko:
Thanks @meteorcloudy and @oquenchil, @ddeville, will you have the capacity to investigate if this is a bazel bug? I guess trying to create a minimal repro case as a standalone repository would be the most useful. There we can try a single args instance with an extremely long command line, if that works correctly then try multiple args instances in a single action, if that works, then I don't know :) I'm happy to help but I'm hopeless on Windows..

util/process_wrapper/process_wrapper.cc
hlopko: I think this explains the scenario: https://github.com/bazelbuild/rules_rust/issues/801#issuecomment-869226649. There are ways to reimplement that use case to not use param files, but I think I'm changing my mind, we should support nested param files. But I think the correct support should keep the relative order of flags. What do you think about flattening the param files, as in, if we find a path to a param file, we read the content of that file and use that instead of the `@nested_file` flag?

meteorcloudy:
@hlopko Feel free to file a bug to Bazel, I'm always happy to fix bugs. I minimal repo would be indeed very helpful! @oquenchil is also gaining a lot of Windows knowledge now, he'll have insight on both side. ;)

util/process_wrapper/process_wrapper.cc
ddeville: Ah that makes sense. I like the idea of flattening params files, I've update the PR to do that.

ddeville:
Hey everyone, sorry for the delay here, the last few weeks have been particularly busy...

I'd be happy to help test this on Windows but since my project is still stuck on 4.1.0 it might not be the best place to start. I could try setting up a minimal repro, I'll just have to prep for some Windows pain, as is always the case 😄.

However ideally I'd like to do this independently of the work here since the fix in the process wrapper needs to happen regardless. What do you think?

hlopko:
Yeah I totally understand. Just to get it out of the way, I'm trying to figure out if there is a way we could still use the only-when-needed param files (maybe on Linux only), since those command lines are easier to work with when using a remote execution service.

But how I'd start would be to write a test that uses our process wrapper and fails on Linux today (custom Starlark rule that generates a command line that has more than 2097152 (my `getconf ARG_MAX` :) elements in ctx.actions.args instance. It should fail. If it fails because process wrapper doesn't handle param files correctly, I'd upload a PR with the test and with the fix in the process wrapper from this PR (and I'm happy to merge that in quickly).

In the presubmit of that PR we will see what happens on Mac and on Windows. We may have to disable the test if the params file was not created correctly there. 

Then we can discuss what options we have (either the rest of this PR, or only doing the fix of this PR when running on failing OS, or using ctx.actions.args in a way that works since I find it unlikely that the feature is fully broken).

Does that sound reasonable?

ddeville:
Yep that makes sense.

As an update, I've also been doing more testing with this PR on my project and I'm still able to reproduce the bug on Windows :(. It's really not super clear what's going on, the manifestation of the bug is that some imports in code just don't resolve. When I symlink all the dependencies into one directory and pass a single `-Ldependency` that points to the directory, things work fine.

Since the project question is super large and entirely proprietary code I've been trying to put together a self-contained sample workspace that could become the base of the testing required here.
I started by checking whether this might be an issue with rustc itself but wasn't able to repro by making the compiler invocations myself (and https://github.com/rust-lang/rust/pull/44094 seems to imply that it hasn't been a problem over there for a while).

I'm gonna try making a bazel workspace using the rules and see where that takes me. It's definitely been a puzzling endeavor so far unfortunately :(

ddeville:
Ok some update here. I've been playing around with this some more and I don't think there is any issue with the automatic use of a param file on Windows (see hacky https://github.com/ddeville/lots-of-rust-deps for generating an insane amount of Rust dependencies 😄). Bazel correctly turns large argument lists that would overflow into param files as expected, on all platforms.

I am still hitting that weird issue where rustc fails to find some extern when a large (I think > 400) number of `-Ldependency=` paths are passed, but only on Windows (it fails with an odd error message, something like `error: LoadLibraryExW failed`). If I tweak things to symlink all these deps into a single folder and pass that to rustc (same thing that Cargo does) then it works fine. I'll have to debug this some more but I don't think it has anything to do with the argument list/param file but potentially something in rustc itself.

It'd still be nice to fix the nested param files issue here but now that the process wrapper was rewritten in Rust that'll take some more work. I'll see if I have some time to migrate this PR.

PR #1081, opened 2022-01-01
Populate CFLAGS and CXXFLAGS when invoking cargo build script.

ddeville:
Downstream tools (in particular cc-rs) depend on them being set.

The particular use case that I noticed was broken was the macos deployment target (specified via the --macos_minimun_os Bazel option) wasn't being applied on binaries being built via cc-rs in a build.rs script. This was causing all sort of bugs such as symbols not being weakly linked since it was using the SDK version (the default) for the deployment target and bazel-built Rust binaries wouldn't load on older OS versions due to failures to resolve symbols at runtime, even though the deployment target was correctly set in Bazel.

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -1,5 +1,5 @@
 # buildifier: disable=module-docstring
-load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "C_COMPILE_ACTION_NAME")
+load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "CPP_COMPILE_ACTION_NAME", "C_COMPILE_ACTION_NAME")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("//rust:defs.bzl", "rust_binary", "rust_common")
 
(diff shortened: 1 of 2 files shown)


illicitonion:
Thanks for putting this together! It's pretty similar to #1004 but the differences are pretty interesting, so it's probably worth lingering on them for a moment:
1. This one gets `CFLAGS` as well as `CXXFLAGS` - this seems good.
2. This one gets env from `C_COMPILE_ACTION_NAME` and the other from `CPP_COMPILE_ACTION_NAME` - should we be getting one, the other, or both and somehow merging them?
3. That one skips `--sysroot` because it's already set on the `CC`/`CXX` env vars. But maybe we can actually stop adding `--sysroot` to `CC`/`CXX` if it's set properly here? Or maybe we can always set it as part of `CFLAGS`/`CXXFLAGS` even if it's not set by the toolchain?

Either way, there's an open question for @hlopko - Do we have any prior art on wiring up a custom cc_toolchain in a unit analysis test, so that we could test that args get properly propagated from there?

(cc @ikalchev - sorry we've left your PR open for so long! I don't really have a preference for which PR we end up merging, but it'd be great to come to a consensus on the open questions above :))

hlopko:
We don't have an existing example, and creating even a fake cc toolchain is quite an endeavor (docs are here: https://docs.bazel.build/versions/main/cc-toolchain-config-reference.html, and I'd be surprised if they were up to date). I think I cannot ask anybody to try doing that without feeling like the pure evil. So I can commit to sending an PR with this fake cc toolchain next week. Hope that's quickly enough.

PiotrSikora:
I can verify that this fixes some build issues for me. Thanks!

illicitonion (approved):
Let's merge this - we can address the c/cpp and sysroot questions in the future as/when needed :)

Sorry for the delay here!

PR #1083, opened 2022-01-03
mix in label in crate hash computation

krasimirgg:
Consider the situation where we have two variants of the same crate:
```
rust_library(
    name = "foo",
    srcs = ["foo.rs"],
    # ... features = ["std"],
)

rust_library(
    name = "foo2",
    crate_name = "foo",
    srcs = ["foo.rs"],
    # ... features = ["no-std"],
)
```

A use case for this is for example when we want to define different variants
of a crate using different feature sets, e.g., define a rustc_private non-std
variant of a vendored crate of the standard library together with a more
standard version of the vendored crate that requires the standard library.

Currently, this is a fragile definition since the output rlib filename of both
of these is the same, e.g., `libfoo-$HASH.rlib`, where `$HASH` is
derived from the crate root source filename, in this case `foo.rs`. So whenever
we have a Bazel query that includes both of these, the build may fail with
errors like:
```
ERROR: file 'test/unit/crate_variants/libfoo-717083168.rlib' is generated by these conflicting actions:
Label: //test/unit/crate_variants:foo2, //test/unit/crate_variants:foo
```

This patch reduces the risk of such output filename collisions by mixing in
the rule label in the output hash calculation. Alternatively we could add an `output_hash` attribute to `rust_library` to make it possible to disambiguate output filenames manually but I thought mixing in the label is simpler for now.


UebelAndre (approved):
Great idea! Thanks!

PR #1089, opened 2022-01-11
Fix proc_macro_dylib_path when they are built in both opt and debug

djmarcin:
For workspace members, a crate_spec is generated for both opt and debug configurations, but when a build is executed, proc-macros are built and used only in the opt configuration. This can result in rust-analyzer looking for the proc macro in the fastbuild location when it was only created in the opt location. Rust analyzer will then be unable to find and use the dylib and code generated by the proc-macro will not have annotations.

This PR fixes the issue by preferring the proc_macro_dylib_path with the `-opt-exec-` path component when merging CrateSpecs.

### DIFF
--- a/tools/rust_analyzer/aquery.rs
+++ b/tools/rust_analyzer/aquery.rs
@@ -525,11 +525,9 @@ mod test {
 
     #[test]
     fn consolidate_proc_macro_prefer_exec() {
-        // mylib.rs is a library but has tests and an entry point, and mylib2.rs
-        // depends on mylib.rs. The display_name of the library target mylib.rs
-        // should be "mylib" no matter what order the crate specs is in.
-        // Otherwise Rust Analyzer will not be able to resolve references to
-        // mylib in mylib2.rs.
+        // proc macro crates should prefer the -opt-exec- path which is always generated
+        // during builds where it is used, while the fastbuild version would only be built
+        // when explicitly building that target.
         let crate_specs = vec![
             CrateSpec {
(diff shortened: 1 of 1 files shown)


tools/rust_analyzer/aquery.rs
```diff
@@ -177,6 +177,16 @@ fn consolidate_crate_specs(crate_specs: Vec<CrateSpec>) -> anyhow::Result<BTreeS
                 existing.display_name = spec.display_name;
                 existing.crate_type = "rlib".into();
             }
+
+            // For proc-macro crates that exist within the workspace, there will be a
+            // generated crate-spec in both the fastbuild and opt-exec configuration.
+            // Prefer proc macro paths with an opt-exec component in the path.
+            if let Some(dylib_path) = spec.proc_macro_dylib_path.as_ref() {
+                const OPT_PATH_COMPONENT: &str = "-opt-exec-";
```
UebelAndre, line 185: Out of curiosity, is `const` doing anything here? I've only ever used it in a global scope.

UebelAndre (approved):
Awesome! One question just for my own education, if you don't mind 😅

tools/rust_analyzer/aquery.rs
djmarcin: `const`s are just compiler placeholders that get inlined wherever they are used. It doesn't actually "do" anything each iteration through this loop because the &str is just inlined, but it's only used in this one function so it seemed better to keep the definition close to use than to disconnect it by putting elsewhere like the top of the file.

PR #1091, opened 2022-01-12
optimization: switch fastbuild opt default to 1 from 0

durin42:
This will very slightly increase build times but applies basic
optimizations that will do things like make iterators _much_
faster. Since fastbuild is how people iterate most of the time, this
seems like a sensible default value.


durin42:
The windows test failure seems like it might be environmental?

UebelAndre:
> The windows test failure seems like it might be environmental?

The windows build is non-fatal but good to know about. The CI failure impacting you is docs being out of date. I wonder if you fix that and push a new commit if the windows build will work.

hlopko (approved):

hlopko:
LGTM, @UebelAndre, I'd just like to have at least 2 approvals on this one, feel free to merge once you approve.

UebelAndre (approved):
Looks good to me as well! I don't expect this to be problematic but just as a heads up. I'd probably be quick to revert it and do some profiling to get a clear picture of the impact if someone complains 😅. But given that my team has a ton of targets we use in tools that are probably suffering from this while iterating, I thank you!

PR #1096, opened 2022-01-19
Add DEBUG env var for running build scripts

djmarcin:
This variable is set for the default profiles as described in https://doc.rust-lang.org/cargo/reference/profiles.html -- true for dev and false for release.

This is an environment variable cargo sets for build scripts documented here https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts


UebelAndre (approved):
Seems good to me! I'd like @illicitonion to take a look too though

Also, I think we should probably add some test for expected environment variables... but would those go in `//test` or would there be a `//cargo/test`?

illicitonion:
LGTM - I'm not sure how easy a test would be, so let's merge this - @hlopko any idea how we'd go about compilation-mode-option-value-specific tests?

PR #1099, opened 2022-01-21
Replace host with exec configurations for grep_includes

hlopko:
This is to be consistent with C++ rules: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CppRuleClasses.java;l=478?q=ccincludescanning.


hlopko:
cc @krasimirgg

UebelAndre (approved):
Thanks for fixing this!

PR #1101, opened 2022-01-22
Allow users to configure the timeout of `cargo_bootstrap_repository`.

UebelAndre:


UebelAndre:
Seems there's an issue in adding `timeout` attributes to repository rules: https://github.com/bazelbuild/bazel/issues/14624

UebelAndre:
Ah, turns out `timeout` is actually an `int` value and it can totally be set. PR has been updated

illicitonion (approved):

PR #1105, opened 2022-01-25
Added `rules_rust_dependencies` and `rust_register_toolchains` macros

UebelAndre:
This allows users to have more control over what repositories are created when setting up `rules_rust` in their workspace and conforms to the [bazel-contrib](https://github.com/bazel-contrib/rules-template/blob/f26177bcbdd9eeab373f23ca568c98704efddd3f/mylang/repositories.bzl) guide for Bazel rules.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -127,12 +127,14 @@ PAGES = dict([
     page(
         name = "rust_repositories",
         symbols = [
+            "rules_rust_depednencies",
+            "rust_register_toolchains",
             "rust_repositories",
             "rust_repository_set",
-            "rust_toolchain",
-            "rust_toolchain_repository",
-            "rust_toolchain_repository_proxy",
             "rust_stdlib_filegroup",
+            "rust_toolchain_repository_proxy",
+            "rust_toolchain_repository",
+            "rust_toolchain",
         ],
     ),
     page(
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -31,52 +31,8 @@ DEFAULT_TOOLCHAIN_TRIPLES = {
     "x86_64-unknown-linux-gnu": "rust_linux_x86_64",
 }
 
-# buildifier: disable=unnamed-macro
-def rust_repositories(
-        version = rust_common.default_version,
-        iso_date = None,
-        rustfmt_version = None,
-        edition = None,
-        dev_components = False,
-        sha256s = None,
-        include_rustc_srcs = False,
-        urls = DEFAULT_STATIC_RUST_URL_TEMPLATES):
-    """Emits a default set of toolchains for Linux, MacOS, and Freebsd
-
-    Skip this macro and call the `rust_repository_set` macros directly if you need a compiler for \
-    other hosts or for additional target triples.
-
-    The `sha256` attribute represents a dict associating tool subdirectories to sha256 hashes. As an example:
-    ```python
-    {
-        "rust-1.46.0-x86_64-unknown-linux-gnu": "e3b98bc3440fe92817881933f9564389eccb396f5f431f33d48b979fa2fbdcf5",
-        "rustfmt-1.4.12-x86_64-unknown-linux-gnu": "1894e76913303d66bf40885a601462844eec15fca9e76a6d13c390d7000d64b0",
-        "rust-std-1.46.0-x86_64-unknown-linux-gnu": "ac04aef80423f612c0079829b504902de27a6997214eb58ab0765d02f7ec1dbc",
-    }
-    ```
-    This would match for `exec_triple = "x86_64-unknown-linux-gnu"`.  If not specified, rules_rust pulls from a non-exhaustive \
-    list of known checksums..
-
-    See `load_arbitrary_tool` in `@rules_rust//rust:repositories.bzl` for more details.
-
-    Args:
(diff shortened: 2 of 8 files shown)


rust/repositories.bzl
```diff
@@ -40,7 +40,8 @@ def rust_repositories(
         dev_components = False,
         sha256s = None,
         include_rustc_srcs = False,
-        urls = DEFAULT_STATIC_RUST_URL_TEMPLATES):
+        urls = DEFAULT_STATIC_RUST_URL_TEMPLATES,
+        register_toolchains = True):
```
illicitonion, line 44: I would expect this param to just bubble down to select whether `native.register_toolchains` gets run inside `rust_repository_set` - either way I would expect the repositories to exist, just not necessarily be registered as toolchains. WDYT?

illicitonion (changes requested):

rust/repositories.bzl
UebelAndre: Maybe the wording is wrong but I feel like there would be little value in adding repositories to fetch toolchains but not registering them. I'd like something that prevents calling `rust_repository_set` all together unless there's a good reason to? Open to ideas

illicitonion: I can imagine there are cases where you'd want to be able to access "a cargo for linux" binary or "a rustc for linux" binary without it affecting toolchain resolution. This is how things like [rules_foreign_cc](https://github.com/bazelbuild/rules_foreign_cc/blob/50ee9979e60e8db38e10de45d2c60873a210bf55/foreign_cc/private/framework/toolchain.bzl#L89-L111) works (and I've seen the pattern elsewhere).

Why are you calling `rust_repositories` at all if you don't want any rust repositories registered?

illicitonion: > Though, I still want to load "rust repositories" because I want to make sure repositories used in the rules (Bazel-skylib, rules_cc, and anything else that may be added) get defined.

Sounds like we maybe could pull out a separate `rules_rust_deps` function, which `rust_repositories` would call, but which you could call instead of calling `rust_repositories`? It's all idempotent, so multiple calls should be fine...

UebelAndre: Sure but I'd then expect to find a `register_rust_toolchains` since I don't think it's immediately clear what the difference between `rules_rust_deps` and `rust_repositories` would be.

illicitonion: Can we bikeshed a name which makes clear "the ruleset dependencies required to use rules_rust" vs "the repositories created by rules_rust" vs "the toolchains registered for those repositories"?

https://github.com/bazel-contrib/rules-template/blob/main/mylang/repositories.bzl would suggest:

* `rules_rust_dependencies`
* `rust_toolchains`, and I'd suggest we make that with an optional `register_toolchains` boolean kwarg

UebelAndre: But that's a larger change than I expected. Should I add a print to `rust_repositories` mentioning that the macro is deprecated?

illicitonion: I think let's just leave `rust_repositories` as a wrapper that calls both for now, and we can consider deprecating in the future :)

UebelAndre: Wait, but why wouldn't we deprecate this? What's the rationale behind keeping all 3 macros?

UebelAndre: Also, while we're making this change, I've noticed quite a few rules use `deps.bzl` over `repositories.bzl` ([rules_go](https://github.com/bazelbuild/rules_go/blob/master/go/deps.bzl) [rules_dotnet](https://github.com/bazelbuild/rules_dotnet/blob/master/dotnet/deps.bzl), [rules_pkg](https://github.com/bazelbuild/rules_pkg/blob/main/pkg/deps.bzl)). While others do use `repositories.bzl`, I think `deps.bzl` is the more common (there are other variants too). Would this be a good time to introduce that or are we happy with `repositories.bzl`?

UebelAndre: And how happy are we with the names? [rules_go](https://github.com/bazelbuild/rules_go/blob/master/go/deps.bzl#L34-L35) uses `go_rules_dependencies` and `go_register_toolchains`, I'm fine with either but want to make sure we've explored all options before making a decision.

cc @hlopko

illicitonion: > Wait, but why wouldn't we deprecate this? What's the rationale behind keeping all 3 macros?

Primarily avoiding churn for people.

> Would this be a good time to introduce that or are we happy with repositories.bzl?
> And how happy are we with the names?

I'm happy with repositories.bzl, rules_rust_dependencies, and rust_register_toolchains, as it's consistent with https://github.com/bazel-contrib/rules-template but I don't have a strong opinion.

UebelAndre: Updated!

illicitonion (approved):
Thanks!

hlopko (approved):
Thank you! This is also conforming to https://docs.bazel.build/versions/main/skylark/deploying.html \o/ :)

docs/BUILD.bazel
```diff
@@ -127,12 +127,14 @@ PAGES = dict([
     page(
         name = "rust_repositories",
         symbols = [
+            "rules_rust_depednencies",
```
luiz10x, line 130: Drive-by comment, but should it be `rules_rust_dependencies` here?

UebelAndre: Oops! Thanks! https://github.com/bazelbuild/rules_rust/pull/1110

PR #1109, opened 2022-01-27
Replace Targets and lists of files in `rust_toolchain` with depsets

UebelAndre:
This should simplify the API of `rust_toolchain` and help avoid confusion for users defining their own toolchains and rules.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 # buildifier: disable=module-docstring
+load("@bazel_skylib//lib:paths.bzl", "paths")
 load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
     "CPP_LINK_EXECUTABLE_ACTION_NAME",
@@ -20,7 +21,6 @@ load(
 load("//rust/private:common.bzl", "rust_common")
 load("//rust/private:providers.bzl", _BuildInfo = "BuildInfo")
 load("//rust/private:stamp.bzl", "is_stamping_enabled")
-load("//rust/private:toolchain_utils.bzl", "find_sysroot")
 load(
     "//rust/private:utils.bzl",
     "expand_dict_value_locations",
@@ -680,7 +680,7 @@ def construct_arguments(
     ))
 
     # Set the SYSROOT to the directory of the rust_std files passed to the toolchain
-    env["SYSROOT"] = find_sysroot(toolchain)
+    env["SYSROOT"] = paths.dirname(toolchain.rust_std.to_list()[0].short_path)
 
     # extra_rustc_flags apply to the target configuration, not the exec configuration.
     if hasattr(ctx.attr, "_extra_rustc_flags") and not is_exec_configuration(ctx):
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -408,8 +408,8 @@ def collect_inputs(
         ([toolchain.target_json] if toolchain.target_json else []) +
         ([] if linker_script == None else [linker_script]),
         transitive = [
-            toolchain.rustc_lib.files,
-            toolchain.rust_std.files,
+            toolchain.rustc_lib,
+            toolchain.rust_std,
             linker_depset,
             crate_info.srcs,
             dep_info.transitive_crate_outputs,
@@ -617,7 +617,7 @@ def construct_arguments(
         rustc_flags.add(linker_script.path, format = "--codegen=link-arg=-T%s")
 
     # Gets the paths to the folders containing the standard library (or libcore)
-    rust_std_paths = depset([file.dirname for file in toolchain.rust_std.files.to_list()]).to_list()
+    rust_std_paths = depset([file.dirname for file in toolchain.rust_std.to_list()]).to_list()
 
     # Tell Rustc where to find the standard library
     rustc_flags.add_all(rust_std_paths, before_each = "-L", format_each = "%s")
@@ -675,7 +675,7 @@ def construct_arguments(
     ))
 
     # Set the SYSROOT to the directory of the rust_std files passed to the toolchain
-    env["SYSROOT"] = paths.dirname(toolchain.rust_std.files.to_list()[0].short_path)
+    env["SYSROOT"] = paths.dirname(toolchain.rust_std.to_list()[0].short_path)
 
     # extra_rustc_flags apply to the target configuration, not the exec configuration.
     if hasattr(ctx.attr, "_extra_rustc_flags") and is_exec_configuration(ctx):
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 # buildifier: disable=module-docstring
-load("@bazel_skylib//lib:paths.bzl", "paths")
 load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
     "CPP_LINK_EXECUTABLE_ACTION_NAME",
@@ -21,6 +20,7 @@ load(
 load("//rust/private:common.bzl", "rust_common")
 load("//rust/private:providers.bzl", _BuildInfo = "BuildInfo")
 load("//rust/private:stamp.bzl", "is_stamping_enabled")
+load("//rust/private:toolchain_utils.bzl", "find_sysroot")
 load(
     "//rust/private:utils.bzl",
     "expand_dict_value_locations",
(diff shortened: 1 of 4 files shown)


UebelAndre:
Not sure if this is good toolchain hygiene or not to try to reduce the toolchain values to files and depsets. It feels like it but happy to learn better practices.

Also, not sure if this should have an incompatibility flag. Happy to add that as well if the change makes sense.

rust/private/rustc.bzl
```diff
@@ -617,7 +617,7 @@ def construct_arguments(
         rustc_flags.add(linker_script.path, format = "--codegen=link-arg=-T%s")
 
     # Gets the paths to the folders containing the standard library (or libcore)
-    rust_std_paths = depset([file.dirname for file in toolchain.rust_std.files.to_list()]).to_list()
+    rust_std_paths = depset([file.dirname for file in toolchain.rust_std.to_list()]).to_list()
```
hlopko, line 620: Since you're thinking about the toolchain these days, WDYT about preparing this depset when creating the toolchain (in other words adding `rustc_std_paths` field to the `RustToolchainInfo`), and not when creating a rustc action? Seems wasteful.

rust/private/rustc.bzl
```diff
@@ -675,7 +675,7 @@ def construct_arguments(
     ))
 
     # Set the SYSROOT to the directory of the rust_std files passed to the toolchain
-    env["SYSROOT"] = paths.dirname(toolchain.rust_std.files.to_list()[0].short_path)
+    env["SYSROOT"] = paths.dirname(toolchain.rust_std.to_list()[0].short_path)
```
hlopko, line 678: Same here, this could be precomputed once in the toolchain, not recomputed for each action. And while at it, it may be a good idea to check that all the paths in `rust_std` have the same parent directory/or they conform to the sysroot layout rustc expects. The latter thing sounds like a separate PR though.

rust/toolchain.bzl
```diff
@@ -268,6 +268,11 @@ def _rust_toolchain_impl(ctx):
         linking_context = linking_context,
     )
 
+    rust_std_files = depset(transitive = [
+        rust_std[DefaultInfo].files,
+        rust_std[DefaultInfo].default_runfiles.files,
```
hlopko, line 273: Why this change with `default_runfiles`? This is surprising. It's also surprising that it seems `rust_std` has some runfiles that are not in `files`, and here we want to make them available at build time. I'm really curious to know the story now :)

rust/toolchain.bzl
```diff
@@ -276,10 +281,10 @@ def _rust_toolchain_impl(ctx):
         clippy_driver = ctx.file.clippy_driver,
         target_json = ctx.file.target_json,
         target_flag_value = ctx.file.target_json.path if ctx.file.target_json else ctx.attr.target_triple,
-        rustc_lib = ctx.attr.rustc_lib,
+        rustc_lib = depset(ctx.files.rustc_lib),
         rustc_srcs = ctx.attr.rustc_srcs,
-        rust_std = rust_std,
-        rust_lib = rust_std,  # `rust_lib` is deprecated and only exists for legacy support.
+        rust_std = rust_std_files,
```
hlopko, line 286: I'm surprised that no test fails because it seems we didn't change this line, https://cs.opensource.google/bazel/rules_rust/+/main:rust/toolchain.bzl;l=190?q=toolchain.bzl, and `rust_std_files` now doesn't have `.label`

hlopko:
In general, why people end up using `Target` vs a depset of files is that there is the label, and that there are the providers. I think only having a depset of files in the `RustToolchainInfo` is a good idea and it will simplify how rules use the toolchain (it will force us to extract all the necessary information from providers and expose only simple depset). So I think I'm in favor of this change. Thank you!

rust/toolchain.bzl
hlopko: Well, of couse it doesn't fail - that code is in toolchain.bzl, we have access to `ctx.attr.rust_std` there. Ignore this thread please :)

rust/private/rustc.bzl
UebelAndre: Yeah, good call. Also, is there `RustToolchainInfo`? Is that is that what's created by `platform_common.ToolchainInfo` or is this a suggestion for a new provider?

rust/toolchain.bzl
UebelAndre: I have no story around `rust_std` using runfiles but had the thought when I wrote `DefaultInfo` and leaned toward the less breaking change. I expect most users to be using what's defined in `rust_register_toolchains` which would not have runfiles. I can remove this if you think it's better 😄

rust/private/rustc.bzl
UebelAndre: This seems wrong also, I'm pretty sure the sysroot should be a directory, not a path to an arbitrary `rlib` (I think that's what they commonly are). But agreed the toolchain should provide this info.

rust/private/rustc.bzl
UebelAndre: Wait, can [dirname](https://docs.bazel.build/versions/main/skylark/lib/File.html#dirname) be calculated like this in the toolchain? 

> The name of the directory containing this file. It's taken from [path](https://docs.bazel.build/versions/main/skylark/lib/File.html#path) and is always relative to the execution directory. 

Couldn't that potentially be different based on the target? Even if the toolchain was the same? Maybe that would imply a loosely constrained toolchain? Does any of that make any sense? 😅

UebelAndre:
@hlopko I have the changes ready, just want to double check they won't introduce a regression

UebelAndre:
@hlopko I opened https://github.com/bazelbuild/rules_rust/pull/1115 just because I'm not sure how the changes related to `dirname` are gonna pan out. It should make it easier to revert if there's an issue.

rust/private/rustc.bzl
hlopko: * Whoops, no there is no `RustToolchainInfo`, it's just `ToolchainInfo`. Bazel toolchains and platforms don't have the ability to specify which provider type is returned from `ctx.toolchains`, it's always `ToolchainInfo`, with custom fields.
* If you provide a `File` from a target, its `dirname` will be constant, no matter which target depends on it (no matter what target is "asking"). So yup, you can calculate that dirname in the toolchain. I'm not really following the issue you're  anticipating to the detail, but I'd say we're safe here anyway.

rust/private/rustc.bzl
hlopko: :+1:

rust/toolchain.bzl
hlopko: > leaned toward the less breaking change

But the original code didn't check runfiles, so it's strictly less breaking change to omit that line from this PR. If you think this is something that needs to be fixed, I'd recommend a separate PR, justification, and a test if possible..

rust/toolchain.bzl
```diff
@@ -276,10 +276,10 @@ def _rust_toolchain_impl(ctx):
         clippy_driver = ctx.file.clippy_driver,
         target_json = ctx.file.target_json,
         target_flag_value = ctx.file.target_json.path if ctx.file.target_json else ctx.attr.target_triple,
-        rustc_lib = ctx.attr.rustc_lib,
+        rustc_lib = depset(ctx.files.rustc_lib),
         rustc_srcs = ctx.attr.rustc_srcs,
-        rust_std = rust_std,
-        rust_lib = rust_std,  # `rust_lib` is deprecated and only exists for legacy support.
+        rust_std = rust_std[DefaultInfo].files,
```
hlopko, line 281: Nit: Why not `rust_std.files`? Also with `rust_lib`.

rust/private/rustc.bzl
```diff
@@ -675,7 +675,7 @@ def construct_arguments(
     ))
 
     # Set the SYSROOT to the directory of the rust_std files passed to the toolchain
-    env["SYSROOT"] = paths.dirname(toolchain.rust_std.files.to_list()[0].short_path)
+    env["SYSROOT"] = find_sysroot(toolchain)
```
hlopko, line 678: I'd prefer if this change was in a separate PR, the logic in `find_sysroot` is non-trivial and it differs from what was here before. I'd like to separate changes in behavior from refactorings. And looking into the implementation of `find_sysroot` I seem to have more questions, and I'd value discussing them in a focused change.

hlopko (approved):
I don't think we need an incompatible flag here, because the actual API of the Rust `ToolchainInfo` is only used by the implementation of `rustc_compile_action`. Thanks for this change! Only nits remaining.

rust/private/rustc.bzl
UebelAndre: Continued in https://github.com/bazelbuild/rules_rust/pull/1115

PR #1115, opened 2022-01-29
Calculate path values in the rust_toolchain construction

UebelAndre:
This change updates the `rust_toolchain` to calculate some paths instead of having each `Rustc` action do it.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -268,14 +268,14 @@ def _rust_toolchain_impl(ctx):
         linking_context = linking_context,
     )
 
-    # Calculate the rustc sysroot path by using a file from the rust-std bundle
-    rust_std_files_list = rust_std.files.to_list()
-    sysroot_anchor = rust_std_files_list[0]
-    sysroot_path_parts = sysroot_anchor.path.split("/")
-    sysroot_path = ""
-    if len(sysroot_path_parts) > 0:
-        # Drop the last part of the split which is expected to be the filename
-        sysroot_path = "/".join(sysroot_path_parts[:-1])
+    # In cases where the toolchain uses the Rust standard library, calculate sysroot path
+    sysroot_path = None
+    if rust_std:
+        # Calculate the rustc sysroot path by using a file from the rust-std bundle
+        rust_std_files_list = rust_std.files.to_list()
+        if not rust_std_files_list:
+            fail("The `rust_std` cannot be represented by an empty list")
+        sysroot_path = rust_std_files_list[0].dirname
 
     toolchain = platform_common.ToolchainInfo(
         rustc = ctx.file.rustc,
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -268,6 +268,15 @@ def _rust_toolchain_impl(ctx):
         linking_context = linking_context,
     )
 
+    # Calculate the rustc sysroot path by using a file from the rust-std bundle
+    rust_std_files_list = rust_std.files.to_list()
+    sysroot_anchor = rust_std_files_list[0]
+    sysroot_path_parts = sysroot_anchor.path.split("/")
+    sysroot_path = ""
+    if len(sysroot_path_parts) > 0:
+        # Drop the last part of the split which is expected to be the filename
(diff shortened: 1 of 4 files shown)


UebelAndre:
Continuing https://github.com/bazelbuild/rules_rust/pull/1109#discussion_r795454460
> * If you provide a `File` from a target, its `dirname` will be constant, no matter which target depends on it (no matter what target is "asking"). So yup, you can calculate that dirname in the toolchain. I'm not really following the issue you're  anticipating to the detail, but I'd say we're safe here anyway.

I'm interested in knowing if `path` will always be the same for a particular file or if it can change based on the current configuration, perhaps after a transition. If it doesn't change then I think the suggestion from the other PR is good.

rust/toolchain.bzl
```diff
@@ -268,6 +268,15 @@ def _rust_toolchain_impl(ctx):
         linking_context = linking_context,
     )
 
+    # Calculate the rustc sysroot path by using a file from the rust-std bundle
+    rust_std_files_list = rust_std.files.to_list()
```
hlopko, line 272: Can this be empty? If so, let's add a nice `fail` error message?

rust/toolchain.bzl
```diff
@@ -268,6 +268,15 @@ def _rust_toolchain_impl(ctx):
...
+    sysroot_anchor = rust_std_files_list[0]
+    sysroot_path_parts = sysroot_anchor.path.split("/")
+    sysroot_path = ""
```
hlopko, line 275: Is `""` a valid sysroot?

rust/toolchain.bzl
```diff
@@ -268,6 +268,15 @@ def _rust_toolchain_impl(ctx):
...
+    if len(sysroot_path_parts) > 0:
```
hlopko, line 276: I think the current behavior is more cleanly expressed using `rpartition`?

rust/toolchain.bzl
UebelAndre: Technically yes. We used to use [here](https://github.com/bazelbuild/rules_rust/blob/5bba04c7d38465b0f5f5982ad4ef76d445b24a4a/rust/private/rustc.bzl#L384-L386) (https://github.com/bazelbuild/rules_rust/pull/339)). Though I do think it might be better to attempt to find the sysroot and otherwise default the path to `None` to indicate there is no sysroot.

rust/toolchain.bzl
UebelAndre: I don't know if it'd be valid for this to be empty but I do think it's valid for `rust_std` to be `None` in the case of `no_std` builds when building for more obscure targets. I can update the code to express that if it makes sense to you too.

rust/toolchain.bzl
UebelAndre: Oh, thanks! Updated

rust/toolchain.bzl
UebelAndre: I ended up adding the changes to the last commit but the question still stands and can update accordingly

rust/toolchain.bzl
UebelAndre: Though, I don't know exactly what it means to `rustc` and co. for `sysroot` to be an empty string.

rust/toolchain.bzl
UebelAndre: Some extra context. To my knowledge the sysroot is only useful for locating `rust_std`. `rustc` should otherwise be able to function the same without it being explicitly set.

hlopko:
> I'm interested in knowing if path will always be the same for a particular file or if it can change based on the current configuration, perhaps after a transition. If it doesn't change then I think the suggestion from the other PR is good.

* The path to a given source file is constant across the whole build and doesn't depend on the current configuration (it's the same in all configurations)
* The path to a given generated file depends on the current configuration, but again, it is constant across the whole build - there may be other identical files with a different path created in a different configuration. A target can depend on files from multiple configurations, but they appear as different files. They do not appear as one file changing its path

Is this making sense?

hlopko (approved):
Thank you!

UebelAndre:
> Is this making sense?

Yes, Thank **you**!

PR #1118, opened 2022-02-01
Deduplicate input to ctx.expand_location

illicitonion:
Works around https://github.com/bazelbuild/bazel/issues/14681

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -628,7 +628,9 @@ def construct_arguments(
     rustc_flags.add_all(rust_std_paths, before_each = "-L", format_each = "%s")
     rustc_flags.add_all(rust_flags)
 
-    data_paths = getattr(attr, "data", []) + getattr(attr, "compile_data", [])
+    # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
+    data_paths = depset(direct = getattr(attr, "data", []) + getattr(attr, "compile_data", [])).to_list()
+
     rustc_flags.add_all(
         expand_list_element_locations(
             ctx,


rust/private/rustc.bzl
```diff
@@ -628,7 +628,12 @@ def construct_arguments(
     rustc_flags.add_all(rust_std_paths, before_each = "-L", format_each = "%s")
     rustc_flags.add_all(rust_flags)
 
-    data_paths = getattr(attr, "data", []) + getattr(attr, "compile_data", [])
+    # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
+    data_paths = getattr(attr, "data", [])[:]
```
UebelAndre, line 632: What's the `[:]` for?

rust/private/rustc.bzl
```diff
@@ -628,7 +628,12 @@ def construct_arguments(
...
+    for extra_data in getattr(attr, "compile_data", []):
+        if extra_data not in data_paths:
+            data_paths.append(extra_data)
```
UebelAndre, line 635: Would it be more efficient to put these into a depset?

UebelAndre (changes requested):
Thanks! Just a few questions

rust/private/rustc.bzl
illicitonion: It was to copy, because an attr is immutable, but I like the depset thing so I've done that instead :)

rust/private/rustc.bzl
illicitonion: We need to pass a sequence, not a depset, to the function, but we might as well use depset to do the deduping for us! Done.

UebelAndre (approved):
Awesome! Thank you!

PR #1119, opened 2022-02-03
`rust_toolchain` now generates a Rust sysroot

UebelAndre:
Currently, `rust_toolchain` does not support the use of toolchain components that are not already a part of a valid sysroot structure. This means that if I have a [`rustc` artifact](https://static.rust-lang.org/dist/rustc-1.58.1-x86_64-unknown-linux-gnu.tar.gz) and a [`rust-std`(Rust standard library) artifact](https://static.rust-lang.org/dist/rust-std-1.58.1-x86_64-unknown-linux-gnu.tar.gz), I cannot use them if they were not downloaded into the same repository rule and follow a specific directory structure. This makes toolchain management hard and leads to unnecessary data being downloaded if users want to have toolchains for multiple target platforms as before any 1 toolchain could be used, all `rust-std` artifacts (or any other target specific assets) would all need to be downloaded regardless of whether or not that target platform is being built for. This PR aims to solve for allowing users to use individual Rust static assets without needing to do additional maintenance to maintain asset bundles specific to all targets or waste time fetching components that will not be used.


examples/crate_universe/WORKSPACE.bazel
```diff
@@ -1,4 +1,4 @@
-workspace(name = "rules_rust_examples_crate_universe")
+workspace(name = "crate_universe_examples")
```
UebelAndre, line 1: I've opened https://github.com/bazelbuild/rules_rust/issues/1120 based on this change.

hlopko:
Adding @krasimirgg who is doing something similar on our end :)

examples/crate_universe/WORKSPACE.bazel
hlopko: Could you please upload this change as a separate PR? I'm happy to approve right away.

krasimirgg:
We'd like to do some more extensive internal testing of this PR. 
I will aim to share our findings in a few business days.

examples/crate_universe/WORKSPACE.bazel
UebelAndre: Opened https://github.com/bazelbuild/rules_rust/pull/1122

krasimirgg:
We hit a tricky issue with this PR internally. This PR causes some unstable clients that use `extern crate rustc_driver` to fail to build under a configuration with remote caching. An example is Kythe: https://osscs.corp.google.com/kythe/kythe/+/master:kythe/rust/extractor/src/lib.rs;l=16?q=rustc_driver&ss=kythe%2Fkythe:kythe%2Frust%2F,
but I've got a smaller example [here](https://github.com/bazelbuild/rules_rust/pull/1126/files#diff-edc482d1d82d786763e347733d20713d77777e0ae0e98c368c94de99936c8e24).
```
#![feature(rustc_private)]
extern crate rustc_driver;
fn main() {}
```

Compiling something like this in linux that refers to the `librustc_driver-$HASH.so` distributed with the toolchain causes `rustc` to invoke the linker with a `-L + -l` combination e.g.:
`-L<stuff>/.cache/bazel/_bazel_krasimir/c736e0c317c9884eaf6f814e37ef83da/external/rust_linux_x86_64/lib/rustlib/x86_64-unknown-linux-gnu/lib -lrustc_driver-cbd31b920a2b2b82`
The linker is happy, as it finds the `librustc_driver-$HASH.so` in the `-L` dir.

Now, suppose that instead `librustc_driver-$HASH.so` happened to be a symlink to a file with a different name in a different directory, e.g., `librustc_driver-$HASH.so -> /tmp/dir/FILE`. Importantly the symlink it points to doesn't follow the `lib<name>.so` naming convention (something like this happens in our configuration with remote caching). In a case like this, the linker sees something like `-L/tmp/dir -lFILE`, which doesn't work since the linker can't find `libFILE.so` under `/tmp/dir`. Note that this by itself is a pre-existing issue, it didn't work before this PR. With Bazel, I can repro this by manually copying `librustc_driver-$HASH.so` from the bazel cache sysroot to `/tmp/dir/FILE` and making the `librustc_driver-$HASH.so` in the bazel cache sysroot to be a symlink to that `FILE`. We'll investigate this further and possibly bring this to the attention to upstream rustc (e.g., for rustc to specify the location `rustc_driver` literally).

Now this PR seems to trigger this pre-existing issue in remote caching builds in certain configurations where the `librustc_driver-$HASH.so` is provided via the `data` attribute of some `sh_binary`. Prior to this PR, the paths resolve to a temporary directory where the `librustc_driver-$HASH.so` is extracted to with exactly this filename. After this PR, the paths resolve to a remote cache directory where the `librustc_driver-$HASH.so` exists under an unrelated name (hash of the contents). I tried but unfortunately couldn't get to reproduce the problem using the RBE rules_rust buildbots.

@hlopko had an idea that may help us out: In this PR we always generate symlinks for sysroot files. Instead, could this PR be enchanted so that if the inputs are already in a directory tree that resembles sysroot, we use them directly without creating symlinks?

Alternatively as a bit of a workaround, wdyt about adding a new attribute, say `copy_sysroot_artifacts_pattern` that allows us to specify a pattern for sysroot artifacts that should be copied instead of just symlinked?

krasimirgg (changes requested):
We hit a tricky issue with this PR internally, caused by a misfortunate combination of factors. This PR causes some unstable clients that use `extern crate rustc_driver` to fail to build under a configuration with remote caching. An example is Kythe: https://osscs.corp.google.com/kythe/kythe/+/master:kythe/rust/extractor/src/lib.rs;l=16?q=rustc_driver&ss=kythe%2Fkythe:kythe%2Frust%2F,
but I've got a smaller example [here](https://github.com/bazelbuild/rules_rust/pull/1126/files#diff-edc482d1d82d786763e347733d20713d77777e0ae0e98c368c94de99936c8e24).

So "-L" "/usr/local/google/home/krasimir/.cache/bazel/_bazel_krasimir/c736e0c317c9884eaf6f814e37ef83da/external/rust_linux_x86_64/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-lrustc_driver-cbd31b920a2b2b82"

UebelAndre:
> @hlopko had an idea that may help us out: In this PR we always generate symlinks for sysroot files. Instead, could this PR be enchanted so that if the inputs are already in a directory tree that resembles sysroot, we use them directly without creating symlinks?

I don't think I like the forked behavior. I'd rather the toolchains work the same way all the time and not need to do any work to detect a directory structure. Creating symlinks should be very fast but alternatively a directory could be created where files are copied into place if symlinks are a concern. WDYT?

rust/toolchain.bzl
```diff
@@ -358,6 +365,10 @@ rust_toolchain = rule(
             ),
             mandatory = True,
         ),
+        "llvm_tools": attr.label_list(
+            doc = "llvm-tools",
```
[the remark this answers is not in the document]
UebelAndre: Not sure how much documentation to put here. The tracking issue (https://github.com/rust-lang/rust/issues/85658) is about as descriptive as I've updated it to. The [embedded book](https://docs.rust-embedded.org/book/intro/tooling.html) has some extra info but this is good enough for me.

rust/private/rustc.bzl
```diff
@@ -681,8 +678,9 @@ def construct_arguments(
         data_paths,
     ))
 
-    # Set the SYSROOT to the directory of the rust_std files passed to the toolchain
+    # Ensure the sysroot is set for the target platform
     env["SYSROOT"] = toolchain.sysroot
+    rustc_flags.add("--sysroot", "${pwd}/" + toolchain.sysroot)
```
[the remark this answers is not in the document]
UebelAndre: This change appears to be unnecessary at this time. Reverted.

rust/toolchain.bzl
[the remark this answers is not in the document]
krasimirgg: Looks good now. It appears I can't resolve github comments here :D

rust/private/rustdoc_test.bzl
```diff
@@ -56,6 +56,11 @@ def _construct_writer_arguments(ctx, test_runner, action, crate_info, rust_toolc
     for var in action.env.keys():
         writer_args.add("--action_env={}".format(var))
 
+    # Ensure the rustdoc binary is always accessed via a relative path
```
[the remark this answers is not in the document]
UebelAndre: Good call. The path to rustdoc moved from being just a path to some file in `external` to a `bazel-out/*` directory since the file was now generated from some action. I've corrected the `rustdoc_compile_action` macro to take a flag `is_test` so the action can correctly be configured for `rust_doc_test` targets.

krasimirgg:
> Creating symlinks should be very fast but alternatively a directory could be created where files are copied into place if symlinks are a concern. WDYT?

Sounds roughly to the symlink pattern approach from my previous comment:
>Alternatively as a bit of a workaround, wdyt about adding a new attribute, say copy_sysroot_artifacts_pattern that allows us to specify a pattern for sysroot artifacts that should be copied instead of just symlinked?

Are you thinking about something like this, or something where we always copy everything instead of using symlinks ever? I personally don't have a sense if copying all the toolchain artifacts (some of which are quite large files) like that may become prohibitively expensive in some situations; @hlopko may have an opinion on this.

UebelAndre:
> Are you thinking about something like this, or something where we always copy everything instead of using symlinks ever? I personally don't have a sense if copying all the toolchain artifacts (some of which are quite large files) like that may become prohibitively expensive in some situations; @hlopko may have an opinion on this.

I'm thinking that we would always copy the same files into the sysroot. My expectation is that instead of having an extra attribute for determining what to copy (or symlink), that would be determined by the attributes themselves. Meaning if you don't want something in your sysroot, don't include it in the toolchain (behavior may vary with specific attributes but this would generally be the rule).

krasimirgg:
>... My expectation is that instead of having an extra attribute for determining what to copy (or symlink), that would be determined by the attributes themselves. Meaning if you don't want something in your sysroot, don't include it in the toolchain

The subtlety is in the distinction between the [what to copy] and [what to symlink] cases. If we go with the approach to copy everything coming from all attributes, we don't have the `rustc_driver` problem above since we're not dealing with symlinks.

If we go the approach to use symlinks (for some of the stuff coming from the attributes), we would still want `librustc_driver-$HASH.so` to exist, but as a real file, not as a symlink, for cases like that example: since it uses `#![feature(rustc_private)] extern crate rustc_driver;` (part of the `dev_components` bundle), it refers to `librustc_driver-$HASH.so` from the sysroot. If we don't include `librustc_driver-$HASH.so` in the sysroot, building the binary will just fail with a `cannot find crate rustc_driver` error. We want to have it in the sysroot, just ensure it's a real file, since the way rustc passes it to the linker breaks if it's a symlink to a random other file [in some configurations].

UebelAndre:
I guess, it doesn't matter to me if symlinks or copies are used, I don't want the behavior to be configurable outside of what is passed to what attribute (e.g. [rust_stdlib_filegroup](https://github.com/bazelbuild/rules_rust/blob/48f97b33c6750f479b2d256f3461426406942df5/rust/toolchain.bzl#L67-L77)). Though, I'm confused at how a situation like this occurs. I won't be able to test on linux till later today but I'm surprised the solution to a problem like this is not "include the missing file in the sysroot" which I'd expect to be done by expanding the globs [here](https://github.com/bazelbuild/rules_rust/blob/48f97b33c6750f479b2d256f3461426406942df5/rust/private/repository_utils.bzl#L26-L39).

krasimirgg:
Update: the `rustc_driver` issue was triggered by the additional `--sysroot` in the older version of this PR. It appears that `rustc` changes how it canonizes filepaths with this flag.

The current version of this PR looks good (using symlinks for everything, no need to copy anything, no need for a pattern-knob workaround) IMO.

UebelAndre:
> Update: the `rustc_driver` issue was triggered by the additional `--sysroot` in the older version of this PR. It appears that `rustc` changes how it canonizes filepaths with this flag.
> 
> The current version of this PR looks good (using symlinks for everything, no need to copy anything, no need for a pattern-knob workaround) IMO.

Glad to hear! But still seems odd that that flag would trigger this behavior. I'll have to dig into that some more but would be glad to see these changes merged 😄

UebelAndre:
The [last commit](https://github.com/bazelbuild/rules_rust/pull/1119/commits/e2fa3ef8dfac5b9cefe36cd3ba45148efc20f690) fixes the issue for me but does rely on the `--sysroot` flag for `rustdoc`...

UebelAndre:
> Though, I'm confused at how a situation like this occurs. I won't be able to test on linux till later today but I'm surprised the solution to a problem like this is not "include the missing file in the sysroot" which I'd expect to be done by expanding the globs [here](https://github.com/bazelbuild/rules_rust/blob/48f97b33c6750f479b2d256f3461426406942df5/rust/private/repository_utils.bzl#L26-L39).

@krasimirgg do you know why the issues with `rustc_driver` are not fixed by updating the glob pattern for `rustc_lib`?

krasimirgg:
> The [last commit](https://github.com/bazelbuild/rules_rust/pull/1119/commits/97a142cb29d63a828e01f4380b2a3048bace3413)[](https://github.com/UebelAndre) fixes the issue for me but does rely on the --sysroot flag for rustdoc...

Using the --sysroot flag for rustdoc sounds good to me (just using it with rustc seems problematic).

> do you know why the issues with rustc_driver are not fixed by updating the glob pattern for rustc_lib?

I don't understand the mechanism that causes the issues with --sysroot while using the rustc_driver in our remote build configurations. I'll double check, but I believe the [equivalent of the] glob patterns we use are correct. 
I suspect it's something to do with symlink canonization in rustc (e.g., there's some non-trivial canonization [here](https://github.com/rust-lang/rust/blob/0c292c9667f1b202a9150d58bdd2e89e3e803996/compiler/rustc_session/src/filesearch.rs#L92)), but that's pure speculation. I'll poke around this more and share what I discover.

UebelAndre:
@hlopko @krasimirgg If there are no outstanding concerns, could I get a final review from each of you? 🙏 😅

rust/private/toolchain_utils.bzl
```diff
@@ -1,5 +1,171 @@
...
+        name (str): A common name for the output directory
+        directory (str): The directory under `name` to put the file in
+        target (File): A File object to symlink to
+
+    Returns:
+        File: A newly generated symlink file
+    """
+    symlink = ctx.actions.declare_file("{}/{}/{}".format(
+        name,
+        directory.lstrip("/"),
```
krasimirgg, line 51: nit: the lstrip looks unnecessary now

krasimirgg (commented):
Looks good to me. I just noticed a small nit.
Please wait for @hlopko's review/ approval.

rust/private/toolchain_utils.bzl
UebelAndre:
```suggestion
        directory,
```

rust/private/toolchain_utils.bzl
```diff
@@ -1,5 +1,171 @@
...
+    Args:
+        ctx (ctx): The toolchain's context object
+        name (str): The name of the sysroot directory (typically `ctx.label.name`)
+        target (Target): A target owning files to symlink
+
+    Returns:
+        depset[File]: A depset of the generated symlink files
+    """
+    tree_files = []
+    for file in target.files.to_list():
```
hlopko, line 15: So this creates one symlink action per file in sysroot. How many files are there, and how large are they (extracted)?

Symlink action does a symlink (linux, mac, sometimes windows) or a copy (usually windows). On linux this is quite fast, on mac slower, on Windows very slow.

To be aware of the impact of this change, would it be an option to benchmark the impact of this change on cpu and memory use on a small project with full toolchain? I hope https://github.com/bazelbuild/bazel-bench works already :)

rust/private/toolchain_utils.bzl
```diff
@@ -1,5 +1,171 @@
...
+        rustc_lib,
+        cargo = None,
+        clippy = None,
+        llvm_tools = None,
+        rust_std = None,
+        rustfmt = None):
+    """Generate a rust sysroot from an exec and target toolchain
+
+    Args:
+        ctx (ctx): A context object from a `rust_toolchain` rule.
```
hlopko, line 76: toolchain.bzl is the only caller of this function, right? If so, I'd move the function to toolchain.bzl, make it private there, and then it will become obvious that ctx is the context of the toolchain. And users won't be tempted to call this function from their code.

rust/private/toolchain_utils.bzl
```diff
@@ -1,5 +1,171 @@
...
+    # Define runfiles
+    direct_files = []
+    transitive_file_sets = []
+
+    # Rustc
+    sysroot_rustc = _symlink_sysroot_bin(ctx, name, "bin", rustc)
+    direct_files.extend([sysroot_rustc, rustc])
+
+    # Rustc dependencies
+    sysroot_rustc_lib = _symlink_sysroot_tree(ctx, name, rustc_lib) if rustc_lib else None
```
hlopko, line 100: Nit: I find the following format more readable, and more common

```
sysroot_rustc_lib = None
if rustc_lib:
	sysroot_rustc_lib = _symlink_sysroot_tree(ctx, name, rustc_lib)
        transitive_file_sets.extend([sysroot_rustc_lib, rustc_lib.files])
```

hlopko (commented):
Only minor things, in general looks good to me, thank you!!

rust/private/toolchain_utils.bzl
UebelAndre: Always nervous about the number of lines changed in a PR. I feel more people look at that number and immediately ignore it after it hits their threshold. But since you asked I'm happy to update this 😄

rust/private/toolchain_utils.bzl
UebelAndre: Moved

rust/private/toolchain_utils.bzl
UebelAndre: Just to start dumping some data, This is what's produced for a compile action:
```
bazel-out/darwin-fastbuild/bin/external/rust_darwin_x86_64/toolchain_for_x86_64-apple-darwin_impl
├── bin
│   ├── cargo -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/bin/cargo
│   ├── clippy-driver -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/bin/clippy-driver
│   ├── rustc -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/bin/rustc
│   ├── rustdoc -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/bin/rustdoc
│   └── rustfmt -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/bin/rustfmt
└── lib
    ├── librustc-stable_rt.asan.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/librustc-stable_rt.asan.dylib
    ├── librustc-stable_rt.lsan.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/librustc-stable_rt.lsan.dylib
    ├── librustc-stable_rt.tsan.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/librustc-stable_rt.tsan.dylib
    ├── librustc_driver-410e8e582906abf2.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/librustc_driver-410e8e582906abf2.dylib
    ├── libstd-91f0e69a7ebd6e00.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/libstd-91f0e69a7ebd6e00.dylib
    ├── libtest-cf5ce37eca3c5b4f.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/libtest-cf5ce37eca3c5b4f.dylib
    └── rustlib
        └── x86_64-apple-darwin
            ├── bin
            │   └── rust-lld -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/bin/rust-lld
            └── lib
                ├── libaddr2line-e4462cb73bc7c44d.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libaddr2line-e4462cb73bc7c44d.rlib
                ├── libadler-63585096303d8a9b.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libadler-63585096303d8a9b.rlib
                ├── liballoc-38142aee14781386.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/liballoc-38142aee14781386.rlib
                ├── libcfg_if-f9d3de4d6bdbf64b.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libcfg_if-f9d3de4d6bdbf64b.rlib
                ├── libcompiler_builtins-870fc7f82e4bde6f.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-870fc7f82e4bde6f.rlib
                ├── libcore-7eec28d9e637d9ae.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libcore-7eec28d9e637d9ae.rlib
                ├── libgetopts-c1e0b00ce77325b8.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libgetopts-c1e0b00ce77325b8.rlib
                ├── libgimli-37047837b923d0f0.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libgimli-37047837b923d0f0.rlib
                ├── libhashbrown-b493ccc9a0b06169.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libhashbrown-b493ccc9a0b06169.rlib
                ├── liblibc-b524a7407cf68766.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/liblibc-b524a7407cf68766.rlib
                ├── libmemchr-cadc9c82d273f933.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libmemchr-cadc9c82d273f933.rlib
                ├── libminiz_oxide-782208c813e9a91c.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libminiz_oxide-782208c813e9a91c.rlib
                ├── libobject-e31c8111c698d163.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libobject-e31c8111c698d163.rlib
                ├── libpanic_abort-e519a6588aa0d119.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libpanic_abort-e519a6588aa0d119.rlib
                ├── libpanic_unwind-36050f194975374d.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libpanic_unwind-36050f194975374d.rlib
                ├── libproc_macro-e654089d7f69e271.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libproc_macro-e654089d7f69e271.rlib
                ├── libprofiler_builtins-8ca8198cfac47c8f.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libprofiler_builtins-8ca8198cfac47c8f.rlib
                ├── librustc-stable_rt.asan.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc-stable_rt.asan.dylib
                ├── librustc-stable_rt.lsan.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc-stable_rt.lsan.dylib
                ├── librustc-stable_rt.tsan.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc-stable_rt.tsan.dylib
                ├── librustc_demangle-54ab9662306ce2de.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_demangle-54ab9662306ce2de.rlib
                ├── librustc_std_workspace_alloc-7a86533803c69aa4.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_std_workspace_alloc-7a86533803c69aa4.rlib
                ├── librustc_std_workspace_core-c27d31671763ac4a.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_std_workspace_core-c27d31671763ac4a.rlib
                ├── librustc_std_workspace_std-f7c6211f96295e24.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_std_workspace_std-f7c6211f96295e24.rlib
                ├── libstd-91f0e69a7ebd6e00.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libstd-91f0e69a7ebd6e00.dylib
                ├── libstd-91f0e69a7ebd6e00.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libstd-91f0e69a7ebd6e00.rlib
                ├── libstd_detect-16528d37a8f3eab3.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libstd_detect-16528d37a8f3eab3.rlib
                ├── libtest-cf5ce37eca3c5b4f.dylib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libtest-cf5ce37eca3c5b4f.dylib
                ├── libtest-cf5ce37eca3c5b4f.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libtest-cf5ce37eca3c5b4f.rlib
                ├── libunicode_width-5e001a9dfe8e0c3c.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libunicode_width-5e001a9dfe8e0c3c.rlib
                └── libunwind-95847f4481c1d402.rlib -> /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libunwind-95847f4481c1d402.rlib
```

UebelAndre: > I hope https://github.com/bazelbuild/bazel-bench works already :)

I'm having issues getting it to run. Funny enough it's in the `pip install` step. The maintainers should use `rules_python` repository rules and make whatever this is doing just a Bazel command.

UebelAndre: @hlopko is there nothing available in Bazel CI that would show some metrics from the builds?

hlopko: Ok that looks acceptable, I don't worry anymore. Just to be extra sure, how much disk space do those files occupy?

Re bazel-bench, yeah it says WIP, but the equivalent script we have internally is super useful. If there is a concrete issue you're having feel free to file it on the repo, owners of it are all wonderful people, and I'm happy to ping them :)

hlopko (approved):

hlopko:
Thank you! :)

rust/private/toolchain_utils.bzl
UebelAndre: > Ok that looks acceptable, I don't worry anymore. Just to be extra sure, how much disk space do those files occupy?

```
rust_darwin_x86_64 % du -Lh -d0 toolchain_for_x86_64-apple-darwin_impl
386M	toolchain_for_x86_64-apple-darwin_impl
```

Seems reasonable(?)

> Re bazel-bench, yeah it says WIP, but the equivalent script we have internally is super useful. If there is a concrete issue you're having feel free to file it on the repo, owners of it are all wonderful people, and I'm happy to ping them :)

I definitely think it'd be worth pinging the group to do a sanity check and make sure it works outside of Google (don't know if it currently does or doesn't. I'm happy to find out the issue is only on my machine). That definitely seems like some useful functionality and would be great to run something like that on a regular basis 😄

hlopko: Yup I think that is reasonable (both things :))

PR #1123, opened 2022-02-04
propagate buildscript link search paths to parent crate

### LINKED ISSUE #859: rustc-link-search parameters are not propagated to dependent crates
I'm trying to use the https://crates.io/crates/sentry-contrib-native crate which in turn depends on https://crates.io/crates/sentry-contrib-native-sys which has a build.rs that adds a rustc-link-search param. This parameter is correctly present in `bazel-out/k8-fastbuild/bin/external/raze__sentry_contrib_native_sys__0_3_1/sentry_contrib_native_sys_build_script.linkflags` but compilation eventually fails because the `rust_binary` that transitively depends on `sentry-contrib-native-sys` does not pick up the flag `-L=${pwd}/bazel-out/k8-fastbuild/bin/external/raze__sentry_contrib_native_sys__0_3_1/sentry_contrib_native_sys_build_script.out_dir/lib`, but it does seem to pick up the `-lsentry`, `-lcrashpad_client`, etc which can't be founds with out the `-L` flag.

It seems like this flag needs to be propagated up through transitive dependencies to the target that gets linked.

iRebbok:
Adds a new `link_search_paths: File` field to `BuildInfo` provider.

Resolves #859

I use native Windows bindings through [windows-rs](https://github.com/microsoft/windows-rs) crate and it fails to link without propagating the link search path to a platform-dependent native library. The bindings call for a native library via `#[link(name = "windows")]` and propagate search path by a buildscript from a dependency (i.e. a transitive dependency for a root crate).

Propagating link search paths seems to be what is cargo doing.


rust/private/rustc.bzl
```diff
@@ -993,6 +993,11 @@ def _create_extra_input_args(ctx, file, build_info, dep_info):
         input_files.append(build_info.out_dir)
         input_files.append(build_info.link_flags)
 
+    # Populate with transitive link search paths
+    for dep_build_info in dep_info.transitive_build_infos.to_list():
```
hlopko, line 997: So this code here is quadratic in extreme situations (and there is pre-existing code that has the same problem, I'm just trying to improve things incrementally). I'd prefer if there was a depset of files field in DepInfo directly, and the resulting code here would look like:

```
build_flags_files.append(dep_info.link_search_path_files)
input_files.append(dep_info.link_search_path_files)
```

hlopko:
In general looks good to me, but I'd prefer if @illicitonion took a look at the runner code. Thank you for working on this!!

illicitonion (approved):
LGTM - thanks!

rust/private/rustc.bzl
iRebbok: I took your advice in the recent commit. I added a new `link_search_path_files` field to DepInfo and now flatten them to propagate.

iRebbok:
Messed a bit, but atleast it passes all the tests now!

illicitonion (approved):
Thanks!

PR #1127, opened 2022-02-07
Add experimental `import` proc macro

cfredric:
This PR implements the experimental `import` macro discussed in #1008 and #1013. This macro is the compiler-side of the changes required to do this renaming, so that the compiler looks for the crate using the correct name.

The macro reads two environment variables, in order to be aware of whether renaming is turned on for this repository, and if so, which directory is considered "third-party" (and therefore should not be renamed).

This PR also contains a bugfix or two for the Bazel side of this flag - I made the renaming itself slightly more efficient, fixed a backward boolean, and made `rust_proto_library` participate in renaming as well.

All tests in @rules_rust pass with and without renaming enabled.

I also updated the `examples` workspace to contain some test cases for renaming, and modified its other packages so that they build/pass with and without renaming enabled. **The exception to this is the two rust_doc[_test] targets**. [rust_doc](https://bazelbuild.github.io/rules_rust/flatten.html#rust_doc) and [rust_doc_test](https://bazelbuild.github.io/rules_rust/flatten.html#rust_doc_test) do not currently support a `proc_macro_deps` attribute (nor a `deps` attribute), so it is currently not possible to use the macro in a doc comment. So, these targets break if renaming is enabled, at the moment.


hlopko (commented):
So this is a large PR, and I have some ideas that I didn't have before (I'm sorry for all the back and forth). Would you mind doing the following?

1. I think the size of this PR warrants splitting code and the tests into separate PRs:
	1. The first PR would only contain macro code dump, no integration, no examples
	1. The second PR would only contain rules integration, no examples
	1. The third PR would only contain examples
1. That said, I think modifying all examples to use the macro may be too confusing for users, they may think this is how they should write their code, and in reality the renaming is experimental. So I'm afraid I'll suggest to implement a way of opting-in to renaming for individual targets. So between 1.2. and 1.3. I'd add:
	1. A PR that adds an attribute to all Rust rules named 'crate_name_format' with 3 possible values: 'target_name', 'target_label', 'default'. Default is the default, and it means whatever the build_setting sets. This is distantly similar to other 3-state flags in Bazel like [dynamic_mode](https://docs.bazel.build/versions/main/command-line-reference.html#flag--dynamic_mode). This PR would also change the code in rustc.bzl to override what build_setting does when target sets the attribute to something else than default.
	If the user uses this attribute on a target, they will have to make sure all targets that depend on it also set the attribute, otherwise the build will obviously fail. Alternatively we could detect this in rules implementation, but that would be a separate PR.
1. We may want to later address the need to add explicit dependency on the macro in `proc_macro_deps`, but I think this is fine for now.

Does this make any sense?

cfredric:
(I think I'm holding GitHub wrong, and can't figure out how to reply to your comment in a threaded way. But I'll reply here.)

> So this is a large PR, and I have some ideas that I didn't have before (I'm sorry for all the back and forth). Would you mind doing the following?
> 
> 1. I think the size of this PR warrants splitting code and the tests into separate PRs:
  i. The first PR would only contain macro code dump, no integration, no examples
  ii. The second PR would only contain rules integration, no examples
  iii. The third PR would only contain examples

Sounds reasonable - this PR did get bigger than I planned. I'll see if I can figure out how to wrangle my branch into several new ones, to submit PRs for the proc_macro itself, tweaking of the utils.bzl implementation, implementation for proto_library rules, and examples, all separately. (I may wait on some of those until their predecessor PRs have landed.)

> 2. That said, I think modifying all examples to use the macro may be too confusing for users, they may think this is how they should write their code, and in reality the renaming is experimental. 

Agreed, the examples probably don't need to be updated unless we want them to build with renaming turned on - but since most users probably won't use that flag, that's probably unnecessary.

> So I'm afraid I'll suggest to implement a way of opting-in to renaming for individual targets. So between 1.2. and 1.3. I'd add:
  i. A PR that adds an attribute to all Rust rules named 'crate_name_format' with 3 possible values: 'target_name', 'target_label', 'default'. Default is the default, and it means whatever the build_setting sets. This is distantly similar to other 3-state flags in Bazel like [dynamic_mode](https://docs.bazel.build/versions/main/command-line-reference.html#flag--dynamic_mode). This PR would also change the code in rustc.bzl to override what build_setting does when target sets the attribute to something else than default.
If the user uses this attribute on a target, they will have to make sure all targets that depend on it also set the attribute, otherwise the build will obviously fail. Alternatively we could detect this in rules implementation, but that would be a separate PR.

I'm not sure I follow the motivation for this attribute. Can you explain? Why would a user need to rename only certain first-party targets? (This isn't necessary during rollout, since the macro is a no-op if the build setting is disabled. So an easy rollout would be to update all targets to use the macro, incrementally, and then flip the build setting.)

> 3. We may want to later address the need to add explicit dependency on the macro in proc_macro_deps, but I think this is fine for now.

Agreed. Maybe as an implicit dependency of every target, similar to how cargo_build_script_runner works? Though that is an unnecessary dependency for targets that don't use the macro. Maybe we need to add another flag, to control whether every target implicitly depends on the proc_macro? Or maybe there are better solutions to this.

cfredric:
Closing this now, as these changes were split apart and merged as #1142, #1143, #1145, and #1151.

Not included in those PRs was:
* a test case in `examples/` (can't be created without having all examples depend on the macro, which currently breaks for wasm targets)
* support for `rust_doc[_test]` rules (can't be created until those rules have a `proc_macro_deps` attribute)

PR #1129, opened 2022-02-07
Fix `rust_doc_test` using execpath paths instead of rootpath

UebelAndre:
This change was split from https://github.com/bazelbuild/rules_rust/pull/1119#discussion_r800647996

### DIFF
--- a/rust/private/rustdoc.bzl
+++ b/rust/private/rustdoc.bzl
@@ -48,7 +48,8 @@ def rustdoc_compile_action(
         toolchain,
         crate_info,
         output = None,
-        rustdoc_flags = []):
+        rustdoc_flags = [],
+        is_test = False):
     """Create a struct of information needed for a `rustdoc` compile action based on crate passed to the rustdoc rule.
 
     Args:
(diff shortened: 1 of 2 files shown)


krasimirgg (approved):
Looks good!

Just to clarify, does this need to be submitted after https://github.com/bazelbuild/rules_rust/pull/1119?

illicitonion (approved):
I guess #1119 will act as a test for this?

UebelAndre:
> Just to clarify, does this need to be submitted after #1119?

No, these changes are effectively a no-op since the execpath and rootpath of rustdoc is the same today. My thought in opening this was to reduce the size of that PR.

UebelAndre:
> I guess #1119 will act as a test for this?

That's where these changes would begin to take effect. Technically there's already coverage since there's already [tests for rust_doc_test](https://github.com/bazelbuild/rules_rust/blob/48f97b33c6750f479b2d256f3461426406942df5/test/unit/rustdoc/rustdoc_unit_test.bzl#L93-L96). I don't think additional testing is needed in this change.

PR #1131, opened 2022-02-09
pin buildifier

UebelAndre:
https://github.com/bazelbuild/buildtools/issues/1044 tracks what I feel is more of a bug in the new [unused-variable](https://github.com/bazelbuild/buildtools/blob/5.0.0/WARNINGS.md#unused-variable) defect. I would like to avoid stamping things with `@unused` and creating cases where dead code can be forgotten about (see the issue for more details).


UebelAndre:
Correction, the defect is not "new" but seems to have just started working in `5.0.0`

illicitonion (approved):

PR #1132, opened 2022-02-09
Addressed some new buildifier defects

UebelAndre:
Some things I think should be merged from running `buildivier 5.0.0`

### DIFF
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -62,14 +62,13 @@ def relativize(path, start):
     src_parts = _path_parts(start)
     dest_parts = _path_parts(path)
     n = 0
-    done = False
     for src_part, dest_part in zip(src_parts, dest_parts):
         if src_part != dest_part:
             break
         n += 1
 
     relative_path = ""
-    for i in range(n, len(src_parts)):
+    for _ in range(n, len(src_parts)):
         relative_path += "../"
     relative_path += "/".join(dest_parts[n:])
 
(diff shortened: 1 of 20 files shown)


rust/private/utils.bzl
```diff
@@ -513,11 +505,18 @@ def _replace_all(string, substitutions):
...
+        substitutions: the list of patterns and replacements to apply.
+
+    Returns:
+        A string with the appropriate substitutions performed.
     """
 
     # Find the highest-priority pattern matches.
     plan = {}
-    for subst_index, (pattern, replacement) in enumerate(substitutions):
+    for _subst_index, (pattern, replacement) in enumerate(substitutions):
```
illicitonion, line 519:
```suggestion
    for pattern, replacement in substitutions:
```

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -19,17 +19,6 @@ def _native_dep_lib_name(ctx):
     else:
         return "libnative_dep.a"
 
-def _lib_has_no_native_libs_test_impl(ctx):
```
illicitonion, line 22: What's the deal with this removal? Feels like the error here was probably not adding a test for it, rather than adding a superfluous impl?

illicitonion:
Thanks! One question then I think this is good to go!

test/unit/native_deps/native_deps_test.bzl
UebelAndre: we no loner use `lib` as a crate type. It's always explicitly `rlib`. To use this test would require adding a new code path to re-implement `lib` which I don't think would be that valuable.

illicitonion (approved):
Thanks!

test/unit/native_deps/native_deps_test.bzl
UebelAndre: https://github.com/bazelbuild/rules_rust/blob/449935861615095c7cc3c1ef477babe7ab48baf4/rust/private/rust.bzl#L170-L183

PR #1134, opened 2022-02-11
Use PIC for rust dylib and cdylib

makroiss:


illicitonion:
Thanks for the contribution! This looks good to me, but as we support versions of Bazel before `lambda` was supported, could you pull out a top-level `def`, or pass a boolean, rather than making and passing lambda functions?

Also, it'd be great to add a test for this behaviour - I imagine it'd be pretty easy to add a unit test to https://github.com/bazelbuild/rules_rust/tree/main/test/unit that asserts on the link flags.

After that, we can get this quickly merged!

hlopko:
I'll work on a integration test for this.

For unittest, I'll also update this thread with how to best write it, there are some questions I need to find answers first :)

hlopko:
The failing test is merged, and I verified that cdylibs fail to link without this PR, and don't fail to link with this PR. This whole problem is very Linux specific so I won't discuss Mac or Windows as they don't suffer from this problem. 

One important detail is that Bazel produces PIC objects in `-c fastbuild` (the default) and in `-c debug`. Only in `-c opt` Bazel creates both PIC and NOPIC objects, and uses PIC for shared libraries, and NOPIC for executables. PIC objects and PIC static libraries have a different name: instead of `libfoo.a` they are named `libfoo.pic.a` (and I failed to realize that in our offline discussion with @makroiss, I apologize for that).

To finish the test side of this PR we need to update `native_deps_test.bzl` and `native_action_inputs_test.bzl`, both to be `-c` (`--compilation_mode`) independent. I think that's relatively simple and "just a matter of programming". You can use `ctx.var["COMPILATION_MODE"]` to detect if the current compilation mode is `opt`, and if it is, expect `.pic.a` files instead of `.a`.

I agree that we could rewrite this PR to use top-level functions instead of lambdas, and I would normally be advocating for that. However I know that @krasimirgg is working on a fix for https://github.com/bazelbuild/rules_rust/pull/841  (actually, the plan is to rollback the PR and add a workaround until upstream rustc is fixed in https://github.com/rust-lang/rust/issues/81490), and the fix involves lambdas again, in a way that I don't see how to rewrite using static functions. So I'd at least consider updating min Bazel version to 5.0:

* the policy says we will support the current Bazel LTS (currently 5.0)
* the policy also says we will support the previous LTS for at least 3 months after new LTS is released. That would be mid-April I think.
* the policy talks about releases of rules_rust, not about HEAD of rules_rust. We don't yet have releases :)

That said, I hate breaking people unexpectedly, so I stand aside in the vote. @illicitonion, @UebelAndre if both of you say let's bump min Bazel version, let's do it.

If not, then the alternative is (besides waiting till April with no action) to:

1. create a Starlark repository rule that checks current Bazel version (https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl#L20). 
2. If the current Bazel version is >= 5.0, symlink a version of a .bzl file that uses lambdas
3. If the current Bazel version is < 5.0, symlink a version of a .bzl file with current logic
4. This PR and @krasimirgg's workaround will only add the fix to the .bzl file for 5.0
5. We remove this lovely code generation hack in April

krasimirgg:
We chatted with @hlopko and reached the conclusion that it will be OK to rewrite this to not use lambdas, in order to support bazel <5 for now (also for my planned follow-up fix for https://github.com/bazelbuild/rules_rust/pull/841).
More precisely (since this is related to the follow-up fix), instead of using `allow_closure = True`:
`args.add_all(dep_info.transitive_noncrates, map_each = make_link_flags, allow_closure = True)`
we update make_link_flags, etc., to accept a custom struct argument (e.g., `(linker_input, use_pic)`),
and we manually transform `dep_info.transitive_noncrates` into a list of those struct-s, which we then feed to `args.add_all(list, ...)`.

The downside of this compared to using closures is that we'll have additional analysis-time struct construction overhead, but it seems OK since we already do something like that to collect the library deps. (We still keep the args construction lazy.)
We could leave a comment to try and switch to using lambdas once we're no longer supporting bazel <5.

makroiss:
Thanks for review Daniel, Marcel and Krasimir. I changed the PR according to Krasimir and Marcel's suggestion.

hlopko (approved):

UebelAndre:
I think I'm a little lost. Why do we need lambdas? Isn't that just syntactic sugar for inlined functions? Why doesn't the suggestion for a normal function solve for the same thing?

hlopko:
Hmm I added a TODO and now buildifier is sad, and I don't know why.. @makroiss, could you please reformat and reupload?

hlopko:
For the change that @krasimirgg has in the baking we need the lambda to capture a variable from the scope. We can't do that easily using functions. I don't think @krasimirgg uploaded the PR yet, but it's coming in the next few days.

makroiss:
I reran buildifier, should be good to go. Thank you for all the help in getting this landed.

PR #1141, opened 2022-02-16
Make native_deps test fail with pie

hlopko:
This PR changes the native_deps test so it starts failing on linux with `-c opt`. The goal is to expose a problem that https://github.com/bazelbuild/rules_rust/pull/1134 is solving.


hlopko:
After this PR is merged and the problem fixed, we should add a CI run with `-c opt`.

illicitonion (approved):

PR #1142, opened 2022-02-17
Add new `import` macro prototype.

cfredric:
This PR implements the experimental import macro discussed in https://github.com/bazelbuild/rules_rust/discussions/1008 and https://github.com/bazelbuild/rules_rust/issues/1013. This macro is the compiler-side of the changes required to do this renaming, so that the compiler looks for the crate using the correct name.

The macro reads two environment variables, in order to be aware of whether renaming is turned on for this repository, and if so, which directory is considered "third-party" (and therefore should not be renamed).


rust/private/rustc.bzl
```diff
@@ -660,6 +660,9 @@ def construct_arguments(
     # Ensure the sysroot is set for the target platform
     env["SYSROOT"] = toolchain.sysroot
 
+    env["RULES_RUST_RENAME_FIRST_PARTY_CRATES"] = str(toolchain._rename_first_party_crates).lower()
```
UebelAndre, line 663: Would it be possible to avoid always adding this to the action?

cfredric: I suppose so - in principle we only need an `Option<String>` in the rust macro. I'll make that change.

In terms of best practices for a build system, is it ok to only set the env var some of the time? I'm imagining things breaking if a user has a `RULES_RUST_THIRD_PARTY_DIR` env var set, but the build setting to rename things wasn't enabled. I guess that's probably unlikely enough that we're ok with that risk?

UebelAndre: I don't think there's a risk but it would invalidate all targets which I don't think is entirely necessary. If this becomes core to how all targets are built then I think it'd be fine to always include it but for now my understanding is that this is opt-in functionality and I think it's better to not unnecessarily include related data into all builds even when unused.

cfredric: Ah, of course! That didn't occur to me, but makes total sense.

hlopko (approved):
Thank you!!

PR #1146, opened 2022-02-18
cdylib_has_native_dep_and_alwayslink_test: fix for when not using gold

krasimirgg:
fixes the test when using non-gold linker, see
https://github.com/bazelbuild/rules_rust/commit/f427b899f3900a58346027e9c037d5e5da570e81#commitcomment-66969879.


illicitonion (approved):
Thanks!

PR #1148, opened 2022-02-21
create symlinks to ambiguous native dependencies

krasimirgg:
Depends on https://github.com/bazelbuild/rules_rust/pull/1147.

This addresses https://github.com/bazelbuild/rules_rust/issues/840 by creating symlinks to ambiguous native dependencies of a rust_binary that would otherwise cause the link to fail, as in the `example/`s added.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -23,6 +23,7 @@ load("//rust/private:providers.bzl", _BuildInfo = "BuildInfo")
 load("//rust/private:stamp.bzl", "is_stamping_enabled")
 load(
     "//rust/private:utils.bzl",
+    "abs",
     "expand_dict_value_locations",
     "expand_list_element_locations",
     "find_cc_toolchain",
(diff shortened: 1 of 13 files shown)


rust/private/rustc.bzl
```diff
@@ -328,6 +329,112 @@ def _process_build_scripts(
...
+    # A dictionary from file paths of ambiguous libraries to the corresponding
+    # symlink.
+    ambiguous_libs = {}
+
+    # A dictionary maintaining a mapping from preferred library name to the
+    # last visited artifact with that name.
+    visited_libs = {}
+    for link_input in dep_info.transitive_noncrates.to_list():
+        for lib in link_input.libraries:
+            # Focus on static libs; dynamic libs use a different location
```
hlopko, line 399: Do they? I believe shared libs still use `-L` for the locations where to search for library at link time, plus they use `-rpath` for search locations at runtime. Does anything break when we remove this `if`?

rust/private/rustc.bzl
```diff
@@ -328,6 +329,112 @@ def _process_build_scripts(
...
+                continue
+            artifact = get_preferred_artifact(lib, use_pic)
+            name = get_lib_name(artifact)
+
+            # On Linux-like platforms, normally library base names start with
+            # `lib`, following the pattern `lib[name].(a|lo)` and we pass
+            # -lstatic=name.
+            # On Windows, the base name looks like `name.lib` and we pass
+            # -lstatic=name.
+            # FIXME: Under the native-link-modifiers unstable rustc feature,
```
hlopko, line 411: Could you add a TODO/FIXMED into comments for both new functions in this file (something like "remove once min supported bazel version is >=5"? Extra points for referencing a github issue.

rust/private/rustc.bzl
```diff
@@ -328,6 +329,112 @@ def _process_build_scripts(
...
+            # we could use -lstatic:+verbatim instead.
+            name_is_ambiguous = (
```
hlopko, line 413: Nit: `needs_symlink_to_standardize_name`? Or something in that spirit. The name without `lib` is not necessarily ambiguous

rust/private/rustc.bzl
```diff
@@ -1146,16 +1262,21 @@ def _make_link_flags_default(linker_input_and_use_pic):
...
             ])
         else:
-            ret.extend(_portable_link_flags(lib, use_pic))
+            ret.extend(_portable_link_flags(lib, use_pic, ambiguous_libs))
     return ret
 
-def _add_native_link_flags(args, dep_info, linkstamp_outs, crate_type, toolchain, cc_toolchain, feature_configuration):
+def _libraries_dirnames(linker_input_and_use_pic_and_ambiguous_libs):
+    link_input, use_pic, _ = linker_input_and_use_pic_and_ambiguous_libs
+    return [get_preferred_artifact(lib, use_pic).dirname for lib in link_input.libraries]
```
hlopko, line 1270: Nit: we could deduplicate, the current code will list `_ambiguous_libs` once per ambiguous library. Not sure this is at all necessary, it should be just a temporary solution, but..

hlopko (approved):

rust/private/rustc.bzl
krasimirgg: Removing this `if` causes `//test/unit/versioned_libs` tests to fail. I believe it includes multiple versions of dynamic libraries that this marks as ambiguous, but which don't really seem problematic AFAICT.

rust/private/rustc.bzl
krasimirgg: I don't follow how we'll be able to remove these functions once min supported bazel version is >=5. AFAIU, this will (only) enable us to use closures in args.add_all, and we already have a TODO there.

Added FIXMEs to both new function referencing the linking modifiers rustc RFC stabilization.

rust/private/rustc.bzl
krasimirgg: done

rust/private/rustc.bzl
hlopko: Ah, it fails because it exercises shared library with a nonstandard name (with version). Ok then let's add a TODO saying that shared libs are not deduplicated right now, and that we hope link modifiers are stabilized soon so we don't have to make it work.

rust/private/rustc.bzl
hlopko: Yeah, waiting for the stabilization is what I wanted to say, thank you :)

rust/private/rustc.bzl
krasimirgg: done

Issue #966, opened 2021-10-12
incompatible_make_rust_providers_target_independent


scentini:
### Motivation

Currently `CrateInfo.deps` and `CrateInfo.proc_macro_deps` have the type `depset[Target]`. In `rustc_compile_action` the Targets are inspected when creating the target’s `DepInfo`.
This is inconvenient if we want to compile a Rust source file from an aspect (that depends on other Rust source files compiled by the aspect), as there is no concrete target that we could put into the `CrateInfo.{deps|proc_macro_deps}` fields. The aspect will instead return a custom provider that will contain a `CrateInfo` and `DepInfo.`

### Incompatible changes

Instead of `depset[Target],` we’ll make `CrateInfo.deps` and `CrateInfo.proc_macro_deps` be of type `depset[DepVariantInfo]`. `DepVariantInfo` is a new provider that wraps the providers a crate’s dependency may have. We’ll also need an additional field in `CrateInfo`: `CrateInfo.owner`, which will contain the label that the `CrateInfo` was created from. This is because we need to know the label so we can memorize the alias for the crate.

The `//rustc/settings:incompatible_make_rust_providers_target_indepent` guards this incompatible change. Once flipped, `rustc_compile_action` will check that the `CrateInfo.owner` field is present, and also that `CrateInfo.deps` and `CrateInfo.proc_macro_deps` have `depset[DepVariantInfo]` type.

### Migration steps
One should pass a `depset[DepVariantInfo]` to `CrateInfo.deps` and `CrateInfo.proc_macro_deps`, as well as the owner label:

```
...

toolchain = ctx.toolchains[Label("//rust:toolchain")]
make_rust_providers_target_independent = toolchain._incompatible_make_rust_providers_target_independent

deps = [DepVariantInfo(
    crate_info = dep[CrateInfo] if CrateInfo in dep else None,
    dep_info = dep[DepInfo] if DepInfo in dep else None,
    build_info = dep[BuildInfo] if BuildInfo in dep else None,
    cc_info = dep[CcInfo] if CcInfo in dep else None,
) for dep in ctx.attr.deps] if make_rust_providers_target_independent else ctx.attr.deps

proc_macro_deps = [DepVariantInfo(
    crate_info = dep[CrateInfo] if CrateInfo in dep else None,
    dep_info = dep[DepInfo] if DepInfo in dep else None,
    build_info = dep[BuildInfo] if BuildInfo in dep else None,
[...]

PR #1150, opened 2022-02-21
Clean up outdated comments in `transform_deps`.

cfredric:
These comments are related to the migration described in #966. This issue is closed, so it looks like the migration is finished - so these comments are out of date.


UebelAndre (approved):
Thanks!

PR #1151, opened 2022-02-22
Add a conditional implicit dependency on the import macro.

cfredric:
This is related to #1008 and #1013. This PR aims to improve the UX of the import macro by not requiring every target to explicitly depend on the macro.

This PR adds an implicit attribute to every `rust_library` and `rust_binary` rule, which (if the `use_real_import_macro` flag is enabled) adds the `import` macro crate to the target's `proc_macro_deps`. If the `use_real_import_macro` flag is disabled (which is the default), the implicit attribute points to a no-op shell script, and nothing is added to the target's dependencies.

### DIFF
--- a/rust/private/transitions.bzl
+++ b/rust/private/transitions.bzl
@@ -48,13 +48,7 @@ import_macro_dep_bootstrap_transition = transition(
 
 def _with_import_macro_bootstrapping_mode_impl(ctx):
     target = ctx.attr.target[0]
-    providers = [target[rust_common.crate_info], target[rust_common.dep_info]]
-    if hasattr(ctx, "executable") and hasattr(ctx.executable, "target"):
-        symlink = ctx.actions.declare_file(target.label.name + "_symlink")
-        ctx.actions.symlink(output = symlink, target_file = ctx.executable.target, is_executable = True)
-        providers.append(DefaultInfo(executable = symlink))
-
-    return providers
+    return [target[rust_common.crate_info], target[rust_common.dep_info]]
 
 with_import_macro_bootstrapping_mode = rule(
     implementation = _with_import_macro_bootstrapping_mode_impl,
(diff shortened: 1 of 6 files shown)


util/import/BUILD.bazel
```diff
@@ -1,11 +1,17 @@
 load("//rust:defs.bzl", "rust_library", "rust_proc_macro", "rust_test")
+load("//rust/private:transitions.bzl", "with_import_macro_bootstrapping_mode")
```
hlopko, line 2: You can use `# buildifier: disable=bzl-visibility` to disable buildifier warning

rust/private/transitions.bzl
```diff
@@ -18,3 +20,53 @@ wasm_bindgen_transition = transition(
...
+import_macro_dep_bootstrap_transition = transition(
+    implementation = _import_macro_dep_bootstrap_transition,
+    inputs = [],
+    outputs = ["@rules_rust//rust/settings:use_real_import_macro"],
+)
+
+def _with_import_macro_bootstrapping_mode_impl(ctx):
+    target = ctx.attr.target[0]
+    providers = [target[rust_common.crate_info], target[rust_common.dep_info]]
+    if hasattr(ctx, "executable") and hasattr(ctx.executable, "target"):
```
hlopko, line 52: Could you remind me why is this necessary?

util/import/BUILD.bazel
cfredric: Done

rust/private/utils.bzl
```diff
@@ -399,6 +399,21 @@ def transform_deps(deps):
         cc_info = dep[CcInfo] if CcInfo in dep else None,
     ) for dep in deps]
 
+def create_import_macro_dep(ctx):
```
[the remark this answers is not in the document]
cfredric: Done

hlopko (approved):

PR #1154, opened 2022-02-25
Remove unnecessary files

hlopko:
Bazel automatically handles symlink targets, there is no need to add
both the symlink and the target to actions.


UebelAndre (approved):
Oh awesome! Really good to know, thanks!

PR #1158, opened 2022-02-27
Replaced `crates_universe` rule with `crates_repository`

UebelAndre:
This PR replaces the existing `crate_universe` rules and binaries with https://github.com/abrisco/cargo-bazel which is an equivalent tool that generally solves for the same desires and is more self-contained.


UebelAndre:
@illicitonion would likely be the best reviewer since, to my knowledge, they're avid consumers of `crates_universe`.

I tried to refactor this PR into small sections to make this easier to review. Each commit should be reviewed individually and anything that refers to "Updating lockfiles" is a commit containing generated content and is probably not worth digging into after confirming the code changes are in a good state.

illicitonion (approved):
LGTM!

shikhar:
Curious, what is the relationship between `crates_universe` and `cargo_bazel`? Looks like there is some shared history?

illicitonion:
> Curious, what is the relationship between `crates_universe` and `cargo-bazel`? I see similar APIs, so looks like there is some shared history?

My team and I put together `crates_universe` as an experimental prototype, and after a few folks including @UebelAndre started using it, @UebelAndre was inspired to rewrite it in a much cleaner way. They have some shared history, but cargo-bazel is a much more principled, well-thought-out implementation, learning from the lessons we learnt in `crates_universe`, rather than trying to do whatever expedient things we could in the original prototype to get something working fast :)

PR #1159, opened 2022-02-28
process_wrapper: replace C++ implementation with rust.

gigaroby:
The main advantages of this work are that it eliminates a bunch of
platform-specific C++ code in favor of the Rust standard library
(mainly std::process::Command and std::fs::File) and it simplifies
future improvements in the area of pipelining and incremental
compilation. Building a Rust process_wrapper requires a way to use
rust_binary without support from process_wrapper itself. This is
achieved through a transition in //util/process_wrapper/transitions.bzl.


util/process_wrapper/transitions.bzl
```diff
@@ -0,0 +1,46 @@
+# buildifier: disable=module-docstring
+def _without_process_wrapper_transition_impl(_settings, _attr):
+    """This transition allows rust_* rules to invoke rustc without process_wrapper."""
+    return {"//util/process_wrapper:use_process_wrapper": False}
```
hlopko, line 4: Please add `"//command_line_option:compilation_mode": "opt"` into the transition. With that we will ensure that process wrapper is optimized even when the actual project is built in `fastbuild` or `debug`.

gigaroby: Done.

util/process_wrapper/util.rs
```diff
@@ -0,0 +1,102 @@
...
+        // a \ at the end of a line allows us to escape the new line break,
+        // \\ yields a single \, so \\\ translates to a single \ and a new line
+        // escape
+        let end_backslash_count = line.chars().rev().take_while(|&c| c == '\\').count();
+        // a 0 or even number of backslashes do not lead to a new line escape
+        let escape = !(end_backslash_count == 0 || end_backslash_count % 2 == 0);
+        //  remove backslashes and add back two for every one
+        let l = line.trim_end_matches('\\');
+        escaped_line.push_str(l);
+        escaped_line.push_str("\\".repeat(end_backslash_count / 2).as_str());
```
hlopko, line 37: Am I optimizing prematurely again? Why not preallocate the string, and append to it in a for loop, so we don't end up creating intermediate strings?

util/process_wrapper/util.rs
```diff
@@ -0,0 +1,102 @@
...
+pub(crate) fn read_file_to_array(path: String) -> Result<Vec<String>, String> {
+    let file = File::open(path).map_err(|e| e.to_string())?;
+    read_to_array(file)
+}
+
+fn stamp_status_to_array(reader: impl Read) -> Result<Vec<(String, String)>, String> {
+    let escaped_lines = read_to_array(reader)?;
+    escaped_lines
+        .into_iter()
+        .map(|l| -> Result<(String, String), String> {
```
hlopko, line 58: I think the type in the closure will be inferred, we can delete it

util/process_wrapper/flags.rs
```diff
@@ -0,0 +1,304 @@
...
+use std::collections::{HashMap, HashSet};
+use std::error::Error;
+use std::fmt;
+use std::fmt::Write;
+use std::mem::{replace, take};
+
+#[derive(Debug, Clone)]
+pub(crate) enum FlagParseError {
+    UnkownFlag(String),
+    RequiredArgumentMissing(String),
```
hlopko, line 24: It seems we don't have required arguments, should we delete this and the code related to required args?

util/process_wrapper/flags.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+impl<'a, T> fmt::Display for FlagDef<'a, T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let required = if self.required { "(required)" } else { "" };
+        write!(f, "{} {}\t{}", self.name, required, self.help)
+    }
+}
+
+impl<'a, T> fmt::Debug for FlagDef<'a, T> {
```
hlopko, line 59: Why not `#[derive(Debug)]`?

util/process_wrapper/main.rs
```diff
@@ -0,0 +1,219 @@
...
+            *arg = new;
+        }
+    }
+    args
+}
+
+fn main() {
+    // Process argument list until -- is encountered.
+    // Everything after is sent to the child process.
+    let mut subst_mapping_raw = vec![];
```
hlopko, line 69: So, WDYT about moving all of this into the flags module, and just having one function that returns a struct owning all these fields? Even the flags definition

util/process_wrapper/BUILD.bazel
```diff
@@ -1,26 +1,46 @@
...
+config_setting(
+    name = "no_process_wrapper",
+    flag_values = {
+        ":use_process_wrapper": "False",
+    },
+)
+
+# A flag that enables/disables the use of process_wrapper when invoking rustc.
+# This is useful for building process_wrapper itself without causing dependency cycles.
+bool_flag(
```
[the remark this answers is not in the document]
gigaroby: Done.

util/process_wrapper/BUILD.bazel
```diff
@@ -1,26 +1,46 @@
...
+    name = "use_process_wrapper",
+    build_setting_default = True,
+)
+
+# Changing the name of this rule requires a corresponding
+# change in //rust/private/rustc.bzl:925
+without_process_wrapper(
```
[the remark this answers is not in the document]
gigaroby: Done.

util/process_wrapper/BUILD.bazel
```diff
@@ -1,26 +1,46 @@
...
+    actual = select({
+        # This will never get used, it's only here to break the circular dependency to allow building process_wrapper
+        ":no_process_wrapper": ":process_wrapper_fake",
+        "//conditions:default": ":process_wrapper_impl",
     }),
     visibility = ["//visibility:public"],
 )
+
+config_setting(
+    name = "no_process_wrapper",
```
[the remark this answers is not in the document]
gigaroby: Done.

util/process_wrapper/transitions.bzl
```diff
@@ -0,0 +1,46 @@
...
+
+without_process_wrapper_transition = transition(
+    implementation = _without_process_wrapper_transition_impl,
+    inputs = [],
+    outputs = ["//util/process_wrapper:use_process_wrapper"],
+)
+
+def _without_process_wrapper_impl(ctx):
+    target = ctx.attr.target[0]
+    executable = target[DefaultInfo].files_to_run.executable
```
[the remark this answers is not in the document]
gigaroby: Yep, done.

util/process_wrapper/transitions.bzl
```diff
@@ -0,0 +1,46 @@
...
+    basename = executable.basename
+    if basename.endswith(".exe"):
```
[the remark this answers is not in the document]
gigaroby: Done.

util/process_wrapper/flags.rs
gigaroby: Done.

util/process_wrapper/util.rs
gigaroby: That's what escaped_line is, am I misunderstanding?

util/process_wrapper/flags.rs
gigaroby: Because I don't want to print the reference'd data.

util/process_wrapper/main.rs
gigaroby: Done.

util/process_wrapper/BUILD.bazel
```diff
@@ -1,26 +1,46 @@
...
+# This is useful for building process_wrapper itself without causing dependency cycles.
+bool_flag(
+    name = "use_process_wrapper",
+    build_setting_default = True,
+)
+
+# Changing the name of this rule requires a corresponding
+# change in //rust/private/rustc.bzl:925
+without_process_wrapper(
+    name = "process_wrapper_impl",
```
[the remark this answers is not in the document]
gigaroby: The reason why this is not "process_wrapper" is that the select rule has that name to maintain compatibility will all the rules that reference it. If we change it, we also need to go change all the labels pointing to it. Is that worth it?

util/process_wrapper/BUILD.bazel
```diff
@@ -1,26 +1,40 @@
...
-    }),
-    defines = [] + select({
-        "@bazel_tools//src/conditions:host_windows": [
-            "UNICODE",
-            "_UNICODE",
-        ],
-        "//conditions:default": [],
+    actual = select({
+        # This will never get used, it's only here to break the circular dependency to allow building process_wrapper
+        ":use_fake_process_wrapper": ":process_wrapper_fake",
```
UebelAndre, line 10: Am I blind? Where is `process_wrapper_fake` being defined?

rust/private/transitions.bzl
```diff
@@ -64,3 +64,49 @@ with_import_macro_bootstrapping_mode = rule(
         ),
     },
 )
+
+def _without_process_wrapper_transition_impl(_settings, _attr):
+    """This transition allows rust_* rules to invoke rustc without process_wrapper."""
+    return {
+        "//command_line_option:compilation_mode": "opt",
```
hlopko, line 71: I'm sorry for that, this was a wrong suggestion - there is an 'exec' transition already happening when building the process wrapper, and that's where compilation mode is set. And since there is `--host_compilation_mode`, the process wrapper should respect that. I think I can edit the file so you don't have to.

util/process_wrapper/BUILD.bazel
hlopko: What a nerd-snipe @gigaroby :) You are not blind, this is a funny tricky edge case in Bazel. I'm not going to admit how much time I've just spent trying to understand this :) `process_wrapper_fake` is only accessed through `alias`. Bazel checks if the target of the alias is a generated file, and if it is not, Bazel assumes it's a source file. Bazel actually doesn't check if this source file exists and happily goes ahead, probably assuming the build will fail in execution. Since we don't use the file in execution, there is nothing to fail. Beautiful.

@gigaroby could you please add an empty target for this anyway? As it stands now this is super confusing.

@comius just in case you were not aware, or this is actually a bug (I think this may have been intentional "optimization" but I can be wrong). Bazel seems to be not checking if a target of an alias exists.

util/process_wrapper/flags.rs
```diff
@@ -0,0 +1,270 @@
...
+            if flag == "--help" {
+                return Ok(ParseOutcome::Help(self.help(program_name)));
+            }
+            if flag == "--" {
+                return Ok(ParseOutcome::Parsed(args));
+            }
+            if args.is_empty() {
+                return Err(FlagParseError::ValueMissing(flag.clone()));
+            }
+            if let Some(e) = self.single.get_mut(&flag) {
```
hlopko, line 149: Nit: e -> flag_def? Also below.

util/process_wrapper/flags.rs
```diff
@@ -0,0 +1,270 @@
...
+                e.data.get_or_insert_with(Vec::new).append(&mut args);
+                continue;
+            }
+            return Err(FlagParseError::UnkownFlag(flag));
+        }
+        Ok(ParseOutcome::Parsed(vec![]))
+    }
+}
+
+// Returns an iterator over Vecs [--foo, arg1, arg2, ... argN].
```
hlopko, line 168: I'm not sure I understand the comment, is it possible it's not valid anymore?

util/process_wrapper/BUILD.bazel
gigaroby: Done.

util/process_wrapper/flags.rs
```diff
@@ -0,0 +1,270 @@
...
+fn iter_flags(
```
hlopko, line 169: Github just decided to throw away my comment because the diff was not valid anymore, and let me send my other comments. Yay I get to write this from scratch again. -.-

So, I find it a bit hard to follow the code and I was wondering if code like this would be more readable:

```
let mut argv_iter = argv.into_iter().peekable();

while let Some(_) = argv_iter.peek() {
	let (flag, values, rest) = consume_next_flag(argv_iter);
	...
	argv_iter = rest;
}
```

WDYT?

util/process_wrapper/util.rs
```diff
@@ -0,0 +1,101 @@
...
+        //  remove backslashes and add back two for every one
+        let l = line.trim_end_matches('\\');
+        escaped_line.push_str(l);
+        escaped_line.push_str("\\".repeat(end_backslash_count / 2).as_str());
+        if escape {
+            // we add a newline as we expect a line after this
+            escaped_line.push('\n');
+        } else {
+            ret.push(escaped_line);
+            escaped_line = String::new();
```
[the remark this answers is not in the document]
gigaroby: Vec.push gives away ownership so I need to allocate a new String.

util/process_wrapper/flags.rs
gigaroby: I simplified it a bit further. WDYT?

util/process_wrapper/flags.rs
```diff
@@ -0,0 +1,276 @@
...
+fn consume_args<I: Iterator<Item = String>>(
+    flag: &str,
+    argv_iter: &mut Peekable<I>,
+) -> Vec<String> {
+    if flag == "--" {
+        // If we have found --, the rest of the iterator is just returned as-is.
+        argv_iter.collect()
+    } else {
+        let mut args = vec![];
+        while let Some(arg) = argv_iter.next_if(|s| !s.starts_with("--")) {
```
hlopko, line 198: can we somehow use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while to simplify?

util/process_wrapper/options.rs
```diff
@@ -0,0 +1,229 @@
...
+        .map_err(OptionError::FlagError)?
+    {
+        ParseOutcome::Help(help) => {
+            eprintln!("{}", help);
+            exit(0);
+        }
+        ParseOutcome::Parsed(p) => p,
+    };
+    let current_dir = std::env::current_dir()
+        .unwrap()
```
hlopko, line 95: Why these unwraps? We could just `?`

util/process_wrapper/flags.rs
hlopko: Nope we cannot :)

hlopko (approved):
What a journey!!! Thank you!

UebelAndre:
This change may have introduced a regression. https://github.com/bazelbuild/rules_rust/commit/d9ba374846831d71edc4876d65d6b49c19f8c634#commitcomment-68356750

@gigaroby @hlopko would either of you be able to take a look?

ddeville:
fwiw I've been trying out the PR on Windows and I'm not having issues.

dfreese:
Bisected to this commit, when trying to upgrade rules_rust.  This triggered the following error for me:
```
    //proto/protobuf:protobuf_prost_generated (782077f4a782b271e6081e5d253d233307fbb065ebd24cbeda52275882be7e2c)
    //tools/prostgen:prostgen (ad7becf1f627c98f4c44ac49f5991ab41aa98662628ab4c26b029d7fb1035392)
.-> @rules_rust//util/process_wrapper:process_wrapper (ad7becf1f627c98f4c44ac49f5991ab41aa98662628ab4c26b029d7fb1035392)
|   @rules_rust//util/process_wrapper:process_wrapper_impl (ad7becf1f627c98f4c44ac49f5991ab41aa98662628ab4c26b029d7fb1035392)
|   @rules_rust//util/process_wrapper:process_wrapper_bin (ad7becf1f627c98f4c44ac49f5991ab41aa98662628ab4c26b029d7fb1035392)
`-- @rules_rust//util/process_wrapper:process_wrapper (ad7becf1f627c98f4c44ac49f5991ab41aa98662628ab4c26b029d7fb1035392)
```
Where the protobuf:protobuf_prost_generated target depends on the prostgen tool.  This was in the `host` configuration.  If I switched that to the `exec` configuration, the error went away, but it does seem like there's potentially a problem for rules that expected to be run on `host`.

hlopko:
@dfreese - FYI from https://docs.bazel.build/versions/main/skylark/rules.html#configurations:

> Historically, Bazel didn’t have the concept of execution platforms, and instead all build actions were considered to run on the host machine. Because of this, there is a single “host” configuration, and a “host” transition that can be used to build a dependency in the host configuration. Many rules still use the “host” transition for their tools, but this is currently deprecated and being migrated to use “exec” transitions where possible.

I think it's reasonable to proactively swift from using `host` to `exec`. Resiliency on transitions is one good reason (`host` doesn't work well with transitions - you can't transition from `host`).

PR #1163, opened 2022-03-03
Enable `experimental_individual_tool_repositories`

UebelAndre:
This is WIP to test on all platforms.


wt:
This does not appear to work with .xz archives.

UebelAndre:
> This does not appear to work with .xz archives.

Would you be willing to send a diff or something to add support for that? One of the goals is to eliminate custom repository rules and instead pull everything through simple `http_archive` definitions. Does that rule support `xz` archives?

wt:
Http_archive does support .tar.xz archives. I wonder if moving the instances where we download and extract separately to download and extract in one step might help with this.

FWIW, I am trying to figure out where the issue is. It is very weird. I have both gz and xz files mirrored on my internal web server. The problem appears to be that it is downloading the gz file and naming it with the xz name. I will try to figure out what is going on, but I think something is assuming gz files, though I haven't yet found it.

Another thing is that the STATIC_RUST_URL env var was removed. You should take that out of this change.

UebelAndre:
@hlopko @illicitonion do either of you have a moment to check out the build error on this PR and see if something jumps out at you? It's been the thing that's halted progress here.

UebelAndre:
https://github.com/bazelbuild/rules_rust/pull/1325 solved for the issue I was seeing on linux [here](https://buildkite.com/bazel/rules-rust-rustlang/builds/6209#749cfe0e-d0a4-4465-9e8b-27e2e68ff929)

PR #1164, opened 2022-03-03
Treat rustfmt as optional

illicitonion:
All the code above this call treats rustfmt as optional, but then when
we make the BUILD for the rust toolchain, we unconditionally point at
the presumed rustfmt label.

At
https://github.com/bazelbuild/rules_rust/blob/59fab4e79f62bfa13551ac851a40696c24c0c3a4/rust/toolchain.bzl#L331-L335
we conditionally create a symlink based on whether the attribute is set,
but because we unconditionally set the attribute, this causes a hard
error failing to symlink to a non-existent file.

Instead, treat rustfmt as optional in this function too, so that the
code above and below properly convey the optionality to each other.

### DIFF
--- a/rust/private/repository_utils.bzl
+++ b/rust/private/repository_utils.bzl
@@ -186,7 +186,7 @@ rust_toolchain(
     rust_doc = "@{workspace_name}//:rustdoc",
     rust_std = "@{workspace_name}//:rust_std-{target_triple}",
     rustc = "@{workspace_name}//:rustc",
-    rustfmt = "@{workspace_name}//:rustfmt_bin",
+    rustfmt = {rustfmt_label},
     cargo = "@{workspace_name}//:cargo",
     clippy_driver = "@{workspace_name}//:clippy_driver_bin",
     rustc_lib = "@{workspace_name}//:rustc_lib",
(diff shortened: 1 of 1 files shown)


rust/private/repository_utils.bzl
```diff
@@ -235,6 +236,9 @@ def BUILD_for_rust_toolchain(
     rustc_srcs = "None"
     if include_rustc_srcs:
         rustc_srcs = "\"@{workspace_name}//lib/rustlib/src:rustc_srcs\"".format(workspace_name = workspace_name)
+    rustfmt_label = "None"
+    if rustfmt_version:
+        rustfmt_label = "@{workspace_name}//:rustfmt_bin".format(workspace_name = workspace_name)
```
UebelAndre, line 241: You'll need quotes here if you want to be able to pass `None` in the template in this way.
```suggestion
        rustfmt_label = "\"@{workspace_name}//:rustfmt_bin\"".format(workspace_name = workspace_name)
```

UebelAndre (changes requested):

rust/private/repository_utils.bzl
```diff
@@ -210,6 +210,7 @@ def BUILD_for_rust_toolchain(
         target_triple,
         include_rustc_srcs,
         default_edition,
+        rustfmt_version,
```
UebelAndre, line 213: nit: since the version is not actually used, can this be something like `include_rustfmt`?

UebelAndre (changes requested):
Looks good! Just one request 😄

rust/private/repository_utils.bzl
illicitonion: Changed to `has_rustfmt` :)

PR #1165, opened 2022-03-03
Allow `cargo_build_script` to forward rustc args to build scripts

UebelAndre:
https://github.com/rust-lang/cargo/issues/9600 lead to the introduction of `CARGO_ENCODED_RUSTFLAGS` which was introduced to allow build scripts to access rust flags. This PR add this flag and sets `--sysroot` so crates like `autocfg` (which makes use of this flag [here](https://github.com/cuviper/autocfg/blob/1.1.0/src/lib.rs#L420-L426)) is correctly able to make use of the generated `rust_toolchain`. This avoids issues like the following for users who create toolchains from separate `rustc` and `rust_std` artifacts.
```
error[E0107]: this struct takes 3 generic arguments but 2 generic arguments were supplied
  --> external/cargo_aliases__clap-3.0.0-beta.2/src/parse/matches/arg_matches.rs:77:22
   |
77 |     pub(crate) args: IndexMap<Id, MatchedArg>,
   |                      ^^^^^^^^ --  ---------- supplied 2 generic arguments
   |                      |
   |                      expected 3 generic arguments
   |
help: add missing generic argument
   |
77 |     pub(crate) args: IndexMap<Id, MatchedArg, S>,
   |                                             +++
 
```

### DIFF
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -119,7 +119,7 @@ def _build_script_impl(ctx):
 
     # In form build scripts of rustc flags
     # https://github.com/rust-lang/cargo/issues/9600
-    env["CARGO_ENCODED_RUSTFLAGS"] = " ".join([
+    env["CARGO_ENCODED_RUSTFLAGS"] = "\x1f".join([
         # Allow build scripts to locate the generated sysroot
         "--sysroot=${{pwd}}/{}".format(toolchain.sysroot),
     ])
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -117,7 +117,7 @@ def _build_script_impl(ctx):
         if cc_toolchain.sysroot:
             env["SYSROOT"] = cc_toolchain.sysroot
 
-    # In form build scripts of rustc flags
+    # Inform build scripts of rustc flags
     # https://github.com/rust-lang/cargo/issues/9600
     env["CARGO_ENCODED_RUSTFLAGS"] = "\x1f".join([
         # Allow build scripts to locate the generated sysroot
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -7,7 +7,7 @@ load("//rust:defs.bzl", "rust_binary", "rust_common")
 load("//rust/private:rustc.bzl", "BuildInfo", "get_compilation_mode_opts", "get_linker_and_args")
 
 # buildifier: disable=bzl-visibility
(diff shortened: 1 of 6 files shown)


cargo/cargo_build_script.bzl
```diff
@@ -121,6 +117,13 @@ def _build_script_impl(ctx):
         if cc_toolchain.sysroot:
             env["SYSROOT"] = cc_toolchain.sysroot
 
+    # In form build scripts of rustc flags
+    # https://github.com/rust-lang/cargo/issues/9600
+    env["CARGO_ENCODED_RUSTFLAGS"] = " ".join([
```
illicitonion, line 122: Should this also include from `rustc_flags`?

illicitonion, line 122: As per the docs at https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
```suggestion
    env["CARGO_ENCODED_RUSTFLAGS"] = "\x1f".join([
```

cargo/cargo_build_script.bzl
```diff
@@ -121,6 +117,13 @@ def _build_script_impl(ctx):
         if cc_toolchain.sysroot:
             env["SYSROOT"] = cc_toolchain.sysroot
 
+    # In form build scripts of rustc flags
```
illicitonion, line 120:
```suggestion
    # Inform build scripts of rustc flags
```

cargo/cargo_build_script.bzl
UebelAndre: great idea! Done 😄

illicitonion (commented):
LGTM - could you add a test?

UebelAndre:
> LGTM - could you add a test?

Fixed a bug and added a test!

test/cargo_build_script/build.rs
```diff
@@ -1,4 +1,24 @@
...
+    let encoded_rustflags = std::env::var("CARGO_ENCODED_RUSTFLAGS").unwrap();
+
+    let flags: Vec<String> = encoded_rustflags
+        .split('\x1f')
+        .map(str::to_string)
+        .collect();
+    assert_eq!(flags.len(), 2);
+
+    assert!(flags[0].starts_with("--sysroot"));
+    assert!(flags[1].starts_with("--verbose"));
```
illicitonion, line 14:
```suggestion
    assert_eq!(flags[1], "--verbose");
```

illicitonion (approved):
LGTM!

PR #1173, opened 2022-03-06
Add ARM gnueabihf triples

bsilver8192:
These are aarch32 with floating point values passed in FPU registers.

### DIFF
--- a/rust/platform/platform.bzl
+++ b/rust/platform/platform.bzl
@@ -12,6 +12,7 @@ load(
 _SUPPORTED_CPU_ARCH = [
     "aarch64",
     "arm",
+    "armv7",
     "i686",
     "powerpc",
     "s390x",
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -24,6 +24,8 @@ SUPPORTED_T2_PLATFORM_TRIPLES = [
     "aarch64-unknown-linux-gnu",
     "arm-unknown-linux-gnueabi",
     "armv7-unknown-linux-gnueabi",
+    "arm-unknown-linux-gnueabihf",
+    "armv7-unknown-linux-gnueabihf",
     "i686-linux-android",
     "i686-unknown-freebsd",
     "powerpc-unknown-linux-gnu",


UebelAndre:
@hlopko Wondering if you have an insights into if [@platforms](https://github.com/bazelbuild/platforms) might be willing to add definitions for ABI. I'm otherwise torn on this PR since I think it adds platforms that require custom user definitions.

bsilver8192:
FWIW I use `@platforms//cpu:armv7` with both of these triples. I look at the Bazel platform as a place to put the differences that matter to Bazel rules, and a toolchain triple as a place to put the differences that matter to the toolchain. For example, I have multiple Bazel platforms that use the same triple but different sysroots.

In what way do you see that these platforms require more custom user definitions than any others @UebelAndre ?

UebelAndre:
> In what way do you see that these platforms require more custom user definitions than any others @UebelAndre ?

How do you distinguish `arm-unknown-linux-gnueabi` from `arm-unknown-linux-gnueabihf`?

hlopko:
@UebelAndre I honestly don't know, you'll have to ask them :) I don't immediately see why that discussion should block merging of this PR though?

On the topic, long long time ago I wrote https://docs.google.com/document/d/1CgU-GKocMAfsUSI3bbGZ0YRkOWczitIoKs29x3zR914/edit#heading=h.3fbh1otqm5sz. There I tried to argue that many things that people want to use constraints for should actually be build settings (for example '--compiler' in the C++ land). I'm not sure what to do about abi, but without thinking too deeply I guess it could be a build setting as well.. \shrug

@bsilver8192 could you split updates of versions in examples from interesting changes into separate PRs please?

UebelAndre:
> @bsilver8192 could you split updates of versions in examples from interesting changes into separate PRs please?

that would require maintaining two separate lists of platform triples, one for "supported by `rules_rust`" and another for "accounted for by `crate_universe`". I would prefer to maintain just one list so I think it's appropriate to have those changes in this PR/

UebelAndre:
> @UebelAndre I honestly don't know, you'll have to ask them :) I don't immediately see why that discussion should block merging of this PR though?

I do not think we should consider a platform triple "supported" if users are gonna have to make custom constraints for it. Because Bazel largely has no concept of `ABI`, it's not possible to distinguish `arm-unknown-linux-gnueabi` from `arm-unknown-linux-gnueabihf` since there's no constraint representing `gnueabi` or `gnueabihf`. I would say if we're going to accept this change that abi constraints are added to `@rules_rust//rust/platform/abi` (or potentially `@platforms//abi`) and all Rust T1-T3 platforms get added in one fell swoop.

bsilver8192:
> > @UebelAndre I honestly don't know, you'll have to ask them :) I don't immediately see why that discussion should block merging of this PR though?
> 
> I do not think we should consider a platform triple "supported" if users are gonna have to make custom constraints for it. Because Bazel largely has no concept of `ABI`, it's not possible to distinguish `arm-unknown-linux-gnueabi` from `arm-unknown-linux-gnueabihf` since there's no constraint representing `gnueabi` or `gnueabihf`. I would say if we're going to accept this change that abi constraints are added to `@rules_rust//rust/platform/abi` (or potentially `@platforms//abi`) and all Rust T1-T3 platforms get added in one fell swoop.

Using either arm-unknown-linux-gnueabi or arm-unknown-linux-gnueabihf works fine without adding anything. AFAIK Rust `cfg` doesn't distinguish between them, so rules_rust can treat them the same (by passing the triple through to the toolchain and the paths to download stdlib etc).

Setting up multiple platforms, some of which use arm-unknown-linux-gnueabi and others which use arm-unknown-linux-gnueabihf, is a bit trickier, but people with builds like that already need to do it some way for toolchains for other languages. rules_rust already supports that using `rust_toolchain`, which I think is good. Trying to automatically support every imaginable set of platforms together is never going to work.

I don't think that `@rules_rust//rust/platform/abi` will make the common case easier. Anybody with a gnueabihf platform will have to know to add that to `constraint_values` for their platform, vs only requiring people with complicated setups to do something extra.

Also, I suspect that gnueabihf is more common than gnueabi. Most modern armv7 chips have an FPU, and most of the world is moving to a matching userspace. In particular Raspbian is armhf aka gnueabihf.

Ultimately, it's up to you. I think I see how to add `@rules_rust//rust/platforms/abi` to automatically distinguish these toolchains, although I could use some guidance on where to document the additional constraints to be added to platforms in order for the automatic rust toolchains to match them for some platforms.

bsilver8192:
> > @bsilver8192 could you split updates of versions in examples from interesting changes into separate PRs please?
> 
> that would require maintaining two separate lists of platform triples, one for "supported by `rules_rust`" and another for "accounted for by `crate_universe`". I would prefer to maintain just one list so I think it's appropriate to have those changes in this PR/

I ran the command to bump the lockfiles to get the new triples so CI passes, and it's unclear to me how to properly update the checksum to reflect new triples without also upgrading the versions. I agree that not updating the lockfiles doesn't make sense, so I think that leaves updating all the versions. Also, looks like `main` bumped most of the versions, so the diff there is smaller now.

hlopko:
> > @bsilver8192 could you split updates of versions in examples from interesting changes into separate PRs please?
> 
> that would require maintaining two separate lists of platform triples, one for "supported by `rules_rust`" and another for "accounted for by `crate_universe`". I would prefer to maintain just one list so I think it's appropriate to have those changes in this PR/

I'm not sure I follow, but I don't necessarily have to follow :) What I was suggesting was to not update the version of anyhow in this PR. But if it is hard for some reason, I'm fine with a single PR.

hlopko:
If I was a superhero, my superpower would be commenting on stuff that is explained right in the next paragraph. Seeing now your https://github.com/bazelbuild/rules_rust/pull/1173#issuecomment-1061265940, I understand, and am fine with a single PR.

hlopko:
> I do not think we should consider a platform triple "supported" if users are gonna have to make custom constraints for it. Because Bazel largely has no concept of ABI, it's not possible to distinguish arm-unknown-linux-gnueabi from arm-unknown-linux-gnueabihf since there's no constraint representing gnueabi or gnueabihf. I would say if we're going to accept this change that abi constraints are added to @rules_rust//rust/platform/abi (or potentially @platforms//abi) and all Rust T1-T3 platforms get added in one fell swoop.

Platforms and constraints are tricky. Now that I was forced to think about this :) I think `abi` should not be a constraint, unless we have a broad Bazel ecosystem buy-in, and the constraint lives in `@platforms`. I'll try to explain why having the `rules_rust`-specific constraint is problematic. Imagine a multilanguage workspace, having Rust, Java, and Python together. You have a build involving all 3 languages, so you need one `platform` to select toolchains in all languages. For rust, you want to pick a specific `abi`, so you put the `constraint_value` into the platform. Your Java and Python toolchains, maintained by `rules_java` and `rules_python`, don't know about the `abi` constraint in `rules_rust`, and they don't use it in their toolchains. Therefore, you are stuck with either:

* not using `abi` from rules_rust in your platform
* modifying all toolchains of all other languages to specify the right `abi` value, even when they don't do anything about it. In cases where you want multiple platforms so rules_rust pick a different toolchain (for example `gnueabihf` and `gnueabi`), you will have to register multiple identical toolchains in Java and Python.

That's why I was suggesting to not use constraint setting for `abi`, but instead use build_setting that is controlled separately. The user, when building your multilanguage project, will have to specify both `--platforms` AND `--rules_rust_abi_flag`, and they will have to make sure they are in sync. Your rust toolchains will have to use selects on `--rules_rust_abi_flag` to pick the right libraries and flags. This is what https://docs.google.com/document/d/1CgU-GKocMAfsUSI3bbGZ0YRkOWczitIoKs29x3zR914/edit#heading=h.3fbh1otqm5sz talks about in more depth.

Years ago I also wrote a "cookbook": https://docs.google.com/document/d/1UZaVcL08wePB41ATZHcxQV4Pu1YfA1RvvWm8FbZHuW8/edit#heading=h.r33iyda45sjm. Not sure how up-to-date it is now.

hlopko:
So, what I wanted to say is handling `abi` and discussions about constraints and settings and toolchains are hard and will take a lot of time and energy. If there is a way to unblock @bsilver8192 without introducing debt, I'd vote for it. If the user doesn't use both `gnueabihf` and `gnueabi`, they don't have a problem. If they do use both, they will have to juggle with `--extra_toolchains` to pick the right one at a right time. When life gives you lemons...

Coordinated plan towards enabling all T1-T3 platforms sounds like a great long term goal that will make this "a solved problem". And it's good idea.

Lastly, I'm merely sharing my opinion, I don't actively use these toolchains, and @UebelAndre is the one who has historically spent a lot of time and energy in maintaining them, so you @UebelAndre are the decider here.

bsilver8192:
> If the user doesn't use both `gnueabihf` and `gnueabi`, they don't have a problem. If they do use both, they will have to juggle with `--extra_toolchains` to pick the right one at a right time. When life gives you lemons...

Actually, you can use them both without going quite that far. rules_rust makes it easy to call `rust_toolchain_repository` for each of the triples, and then have customized `toolchain`s with appropriate `target_compatible_with` so they can all be registered all the time with `register_toolchains`. I've already got appropriate constraints set up for my C++ toolchains, so applying them to the Rust toolchains too makes everything work nicely. It's set up pretty well IMHO.

UebelAndre:
I've opened https://github.com/bazelbuild/platforms/issues/38 to track common constraint values for ABI.

UebelAndre:
Is there no way to introduce support for the new platform without this change? I feel this would introduce undesirable side effects in folks who use [SUPPORTED_PLATFORM_TRIPLES](https://github.com/bazelbuild/rules_rust/blob/1b1f21ac5c1b188bf658cb408afdb58f1668baef/rust%2Fplatform%2Ftriple_mappings.bzl#L38) for custom toolchain definitions and select statements (notably impacting `crate_universe` which you can see in this PR).

bsilver8192:
Oh right, I forgot about the infrastructure to create toolchains for all the triples, and that other people use `crate_universe`... I'm not doing either of those, so issues there don't affect me. I see why those are problematic though.

> Is there no way to introduce support for the new platform without this change?

The only part which didn't work for me before this change is triggering the correct `cfg`. I think I need all of these changes to make that work. If I only change `_SUPPORTED_CPU_ARCH`, then I get failures like this:
```
ERROR: /bazel-cache/brian/bazel/_bazel_brian/227d44ac296d5d23388913ec67ab1659/external/rules_rust/rust/platform/BUILD.bazel: no such target '@rules_rust//rust/platform:armv7-unknown-linux-gnueabihf': target 'armv7-unknown-linux-gnueabihf' not declared in package 'rust/platform' (did you mean 'armv7-unknown-linux-gnueabi'?) defined by /bazel-cache/brian/bazel/_bazel_brian/227d44ac296d5d23388913ec67ab1659/external/rules_rust/rust/platform/BUILD.bazel
```

WDYT about creating a `NONDEFAULT_SUPPORTED_PLATFORM_TRIPLES` list which is used to generate the various build rules, but doesn't feed into `rust_register_toolchains` or `crate_universe`? I'd be fine leaving it undocumented too.

bsilver8192:
Also, I'm fine carrying the 3 lines that matter to me as a local patch if there isn't a good way to merge this. I've had far bigger ones adding support for my platforms in other places :)

Issue #1168, opened 2022-03-05
`crates_repository`: The `default_features` field in `crate.spec` doesn't seem to work


SlyMarbo:
Hi there. I'm currently updating from `crate_universe` to the new `crates_repository` functionality introduced in #1158. The introduction of support for `default_features = False` is exciting, as I'm building a lot of `#![no_std]` code, so I often need to disable the default features to turn off the `std` feature.

However, setting `default_features = False` in a `crate.spec` doesn't seem to have any effect. I've had a quick look through the code and `crate.spec` propagates the field to the JSON it produces, but I can't find anything pulling it out of the JSON and crates still seem to be using their default features. Have I missed something? Thanks for your help.

SlyMarbo:
I should add that with `crate_universe`, I worked around the inability to disable default features by using `crate.override`'s `features_to_remove` field. There doesn't seem to be anything similar in the new `crate.annotate`, so I'm not sure how to work around the issue.

UebelAndre:
@SlyMarbo can you check if https://github.com/bazelbuild/rules_rust/pull/1174 solved the issue you were seeing with `default_features = False` having no effect?

Additionally, could you file a separate issue for [your last comment](https://github.com/bazelbuild/rules_rust/issues/1168#issuecomment-1059806271)?

SlyMarbo:
@UebelAndre sadly #1174 doesn't seem to have helped. If it helps, I tried it by using commit 261f4e0. That commit also seems to have no reference to `default_features`, beside the field in `crate.spec`. I've opened #1176 for my second comment. Thanks :)

UebelAndre:
@SlyMarbo Can you provide a repro case to test with?

SlyMarbo:
I'm working on a smaller repro, but that doesn't seem to be exhibiting the issue so I'll keep debugging my main case ([ProjectSerenity/firefly](https://github.com/ProjectSerenity/firefly)) and let you know how I get on.

SlyMarbo:
Here's my failed reproduction, just in case that helps: [SlyMarbo/sample-rules-rust-std](https://github.com/SlyMarbo/sample-rules-rust-std).

SlyMarbo:
I have now been able to reproduce the issue in the smaller repro. For some reason, `default_features = False` worked correctly with just the `managed` crate, but not `smoltcp`. I can make a branch using the previous `crates_universe` functionality if that would help.

UebelAndre:
[...]

PR #1174, opened 2022-03-07
Fixed `crate.spec` being ignored in `crate_universe` rules.

UebelAndre:
[cargo_toml::DependencyDetail](https://gitlab.com/crates.rs/cargo_toml/-/blob/v0.11.3/src/cargo_toml.rs#L504) currently looks for kebab-case variables to deserialize since that's the expectation from Cargo manifests. This change fixes splicing manifests to correctly contain `kebab-case` key values for crate specs.

This addresses part of https://github.com/bazelbuild/rules_rust/issues/1168

### DIFF
--- a/crate_universe/private/crates_vendor.bzl
+++ b/crate_universe/private/crates_vendor.bzl
@@ -1,7 +1,7 @@
 """Rules for vendoring Bazel targets into existing workspaces"""
 
 load("//crate_universe/private:generate_utils.bzl", "collect_crate_annotations", "render_config")
-load("//crate_universe/private:splicing_utils.bzl", "splicing_config")
+load("//crate_universe/private:splicing_utils.bzl", "kebab_case_keys", "splicing_config")
 load("//crate_universe/private:urls.bzl", "CARGO_BAZEL_LABEL")
 load("//rust/platform:triple_mappings.bzl", "SUPPORTED_PLATFORM_TRIPLES")
 
@@ -58,7 +58,8 @@ def _write_data_file(ctx, name, data):
 def _write_splicing_manifest(ctx):
     # Deserialize information about direct packges
     direct_packages_info = {
-        pkg: json.decode(data)
+        # Ensure the data is using kebab-case as that's what `cargo_toml::DependencyDetail` expects.
+        pkg: kebab_case_keys(dict(json.decode(data)))
         for (pkg, data) in ctx.attr.packages.items()
     }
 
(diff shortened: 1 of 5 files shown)


illicitonion (approved):
Is there some kind of end-to-end test/example we could have that shows that the config is actually picked up?

UebelAndre:
> Is there some kind of end-to-end test/example we could have that shows that the config is actually picked up?

There certainly is but I couldn't come up with one that didn't require spelunking through crates.io for the perfect crate to test with. Which I think would have been something that toggles something that I can assert on in a test if `default-features = False`... Would you happen to know of any or can you think of a nice regression test?

illicitonion:
> > Is there some kind of end-to-end test/example we could have that shows that the config is actually picked up?
> 
> There certainly is but I couldn't come up with one that didn't require spelunking through crates.io for the perfect crate to test with. Which I think would have been something that toggles something that I can assert on in a test if `default-features = False`... Would you happen to know of any or can you think of a nice regression test?

I think the most common use-case of `default-features = False` is to disable std... I don't know how easy it is to build against the target `thumbv6m-none-eabi` (which doesn't have std) with Bazel, but if it's easy to do, any number of crates with optional `std` can be used as a test-bed there... I happen to know https://github.com/illicitonion/num_enum is a crate which verifies its own `no-std`-compat that way, but I'm sure there are many others too.

That said, I don't know whether building against `thumbv6m-none-eabi` is easy on our CI :D

UebelAndre:
> > > Is there some kind of end-to-end test/example we could have that shows that the config is actually picked up?
> > 
> > 
> > There certainly is but I couldn't come up with one that didn't require spelunking through crates.io for the perfect crate to test with. Which I think would have been something that toggles something that I can assert on in a test if `default-features = False`... Would you happen to know of any or can you think of a nice regression test?
> 
> I think the most common use-case of `default-features = False` is to disable std... I don't know how easy it is to build against the target `thumbv6m-none-eabi` (which doesn't have std) with Bazel, but if it's easy to do, any number of crates with optional `std` can be used as a test-bed there... I happen to know https://github.com/illicitonion/num_enum is a crate which verifies its own `no-std`-compat that way, but I'm sure there are many others too.
> 
> That said, I don't know whether building against `thumbv6m-none-eabi` is easy on our CI :D

I've opened an issue at https://github.com/bazelbuild/rules_rust/issues/1175 to track writing tests so that this bugfix is not blocked on that. There's likely a better way to write the starlark macros to make this more testable without requiring a full example. I'll look into that when I can find the time.

Issue #1166, opened 2022-03-05
regressive commit


wt:
https://github.com/bazelbuild/rules_rust/commit/3db49e80c14b4c5dc9bd6a7c13a848ef71891c35 is a regressive commit.

This commit breaks a very simple query for me.

I have a simple hello world rust target here: `//rust/example/hello_world:hello_world`.

I run the following query: `bazel query 'deps(//rust/example/hello_world:hello_world)'`

Before 3db49e80c14b4c5dc9bd6a7c13a848ef71891c35 (specifically 5591596ec36fc156fbeae7f055e612f4bdca1d35), it works. After that commit, I get the following:
```
➜  source git:(new_rules_rust) bazel query 'deps(//rust/example/hello_world:hello_world)'
INFO: Invocation ID: 983a9a5a-480d-4f0a-8285-715f2703a72d
ERROR: /private/var/tmp/_bazel_wturkal/98aac8fab9f210d0477cc94a59a95ebe/external/rules_rust/util/import/raze/BUILD.bazel:24:6: no such package '@rules_rust_util_import__lazy_static__1_4_0//': The repository '@rules_rust_util_import__lazy_static__1_4_0' could not be resolved: Repository '@rules_rust_util_import__lazy_static__1_4_0' is not defined and referenced by '@rules_rust//util/import/raze:lazy_static'
ERROR: /private/var/tmp/_bazel_wturkal/98aac8fab9f210d0477cc94a59a95ebe/external/rules_rust/util/import/raze/BUILD.bazel:24:6: no such package '@rules_rust_util_import__lazy_static__1_4_0//': The repository '@rules_rust_util_import__lazy_static__1_4_0' could not be resolved: Repository '@rules_rust_util_import__lazy_static__1_4_0' is not defined and referenced by '@rules_rust//util/import/raze:lazy_static'
ERROR: /private/var/tmp/_bazel_wturkal/98aac8fab9f210d0477cc94a59a95ebe/external/rules_rust/util/import/raze/BUILD.bazel:15:6: no such package '@rules_rust_util_import__aho_corasick__0_7_15//': The repository '@rules_rust_util_import__aho_corasick__0_7_15' could not be resolved: Repository '@rules_rust_util_import__aho_corasick__0_7_15' is not defined and referenced by '@rules_rust//util/import/raze:aho_corasick'
[...]

PR #1179, opened 2022-03-07
Tweak import macro bootstrap to trick rust analyzer aspect

hlopko:
In https://github.com/bazelbuild/rules_rust/issues/1166 we discovered that rust analyzer aspect doesn't understand the `_with_import_macro_bootstrapping_mode` rule. This PR changes the rule to pretend to be more like a normal Rust rule.

CC @cfredric


util/import/BUILD.bazel
```diff
@@ -1,11 +1,11 @@
...
 
 # buildifier: disable=bzl-visibility
-load("//rust/private:transitions.bzl", "with_import_macro_bootstrapping_mode")
+load("//rust/private:transitions.bzl", "alias_with_import_macro_bootstrapping_mode")
 
-with_import_macro_bootstrapping_mode(
+alias_with_import_macro_bootstrapping_mode(
     name = "import_macro",
-    target = "import_macro_impl",
+    deps = ["import_macro_impl"],
```
UebelAndre, line 8: Could this not be `default = [Label("@rules_rust//uril/import:import_macro_impl")]` on the attribute?

UebelAndre (changes requested):

util/import/BUILD.bazel
hlopko: It could, but I think I prefer the current setup - scanning through the BUILD file gives you the idea what's going on. With the `default` trick you have to go to the bzl file and back to follow the breadcrumbs... Are there advantages that I don't see?

wt: I think the default is probably a better way to handle this in this case. If it's logically a default, that's how it should be modeled.

UebelAndre: If `alias_with_import_macro_bootstrapping_mode` is expected to be used anywhere else, I'd say `default` should be used so other users don't have to think about it. Not sure if it's directly relevant to this change but it'd be nice to see all the doc fields filled out on the rule to explain what it's purpose was. Can you elaborate on it a bit here though?

UebelAndre (changes requested):

util/import/BUILD.bazel
hlopko: I don't see `alias_with_import_macro_bootstrapping_mode` as being logically tied to the import_macro. It's a rule that changes the configuration for its `deps`/`target`. And I really value the readability and explicitness of the BUILD file. To be honest I haven't seen a non-implicit label attribute with a default yet, I think they are rare and surprising.

I think the optimal solution would have been to create a `rust_binary`-like rule that does everything the standard `rust_binary` rule does plus it does the transition. That is sadly a bit too complicated and full of boilerplate right now.

hlopko:
The `alias_with_import_macro_bootstrapping_mode` is not expected to be used anywhere else. I see it as a private rule, and implementation detail. Vast majority of users should not have to know about it, and if somebody is looking there, it's likely because they are investigating something and they are trying to understand what's going on. 

We intentionally omitted the docs attributes so this rule doesn't pollute our docs pages. But now that I'm thinking about it it's not the presence of docs, but explicit configuration in bzl files in //docs that matters for what is shown on docs pages. Therefore I'll add docs in the morning :)

util/import/BUILD.bazel
UebelAndre: I don't care too much about it being a default. Had I written this, I would generally opt to keep private/implementation-detail rules as concise as possible by implementing the majority of the funcitonality in the `.bzl` file to have a single source of truth and only ever specifying `my_rule(name = "my_rule")` in the BUILD file. I find this makes things far more readable and easier to debug.

UebelAndre:
Sweet! I'll await docs

hlopko:
That, tests, and also I think I'll re-do the PR to add support for aliases into rust analyzer tooling (currently alias is not traversed), and then I'm change the transitioning rules to mimic alias (name of the attribute will be `actual`).

hlopko:
This is now ready for another review, please take a look :)

rust/private/transitions.bzl
```diff
@@ -46,18 +46,20 @@ import_macro_dep_bootstrap_transition = transition(
     outputs = ["@rules_rust//rust/settings:use_real_import_macro"],
 )
 
-def _with_import_macro_bootstrapping_mode_impl(ctx):
-    target = ctx.attr.target[0]
+def _alias_with_import_macro_bootstrapping_mode_impl(ctx):
+    actual = ctx.attr.actual
+    if len(actual) != 1:
```
UebelAndre, line 51: Is this check still needed?

rust/private/transitions.bzl
```diff
@@ -46,18 +46,20 @@ import_macro_dep_bootstrap_transition = transition(
...
+        fail("`deps` attribute has to have exactly one element.")
+    target = actual[0]
```
UebelAndre, line 53: Perhaps this is needed for the transition? But the `fail` seems dated?

rust/private/transitions.bzl
```diff
@@ -46,18 +46,20 @@ import_macro_dep_bootstrap_transition = transition(
...
     return [target[rust_common.crate_info], target[rust_common.dep_info]]
 
-with_import_macro_bootstrapping_mode = rule(
-    implementation = _with_import_macro_bootstrapping_mode_impl,
+alias_with_import_macro_bootstrapping_mode = rule(
+    implementation = _alias_with_import_macro_bootstrapping_mode_impl,
     attrs = {
-        "target": attr.label(
+        # Using `actual` so tooling such as rust analyzer aspect traverses the target.
+        "actual": attr.label(
```
UebelAndre, line 60: Small nit
```suggestion
        "actual": attr.label(
            doc = "The target this alias refers to.",
```

rust/private/transitions.bzl
```diff
@@ -46,18 +46,20 @@ import_macro_dep_bootstrap_transition = transition(
...
+    actual = ctx.attr.actual
+    if len(actual) != 1:
+        fail("`deps` attribute has to have exactly one element.")
+    target = actual[0]
     return [target[rust_common.crate_info], target[rust_common.dep_info]]
 
-with_import_macro_bootstrapping_mode = rule(
-    implementation = _with_import_macro_bootstrapping_mode_impl,
+alias_with_import_macro_bootstrapping_mode = rule(
+    implementation = _alias_with_import_macro_bootstrapping_mode_impl,
```
UebelAndre, line 57: Can you fill in docs for this rule?
```suggestion
    implementation = _alias_with_import_macro_bootstrapping_mode_impl,
    doc = "TODO",
```

UebelAndre (dismissed):
Looking good! Just some nits 😅

UebelAndre:
I'm not entirely sure what happend here: https://buildkite.com/bazel/rules-rust-rustlang/builds/5705#1fa56c67-7d6d-407a-b50c-aa78c5327cf1 
There was a timeout but as far as I can tell the timeout should be 15 minutes and the build only ran for 11... I'll dig more into this but I highly suspect a rebuild will fix it (since that's what happened to me).

rust/private/transitions.bzl
hlopko: Removed.

rust/private/transitions.bzl
hlopko: Yup the transition makes the `attr.label` be a list.

UebelAndre (approved):
Awesome! Thank you!

Issue #1167, opened 2022-03-05
Cross-compiling for Android doesn't seem to propagate architecture correctly


korDen:
I have a basic method that I want to invoke through JNI. Think hello world. I have a similar one implemented in C. It works. Rust one fails to link. Let's start with a basic repro.

```
rust_static_library(
    name = "android-rs",
    srcs = ["src/lib.rs"], // #[no_mangle] pub fn hello_from_jni(...) { ... }
)

// Exact same think but in C++, just as a sanity check.
cc_library(
    name = "android-cc",
    srcs = ["sanity_check.cc"], // extern "C" void hello_from_jni(...) { ... }
)

android_binary(
    name = "android",
    custom_package = "rust.on.android",
    manifest = "AndroidManifest.xml",
    deps = [
        ":android-rs", // fails
        ":android-cc", // succeeds
    ],
)
```

```bazel build :android```

Command that fails:
```external/androidndk/ndk/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang -shared -o bazel-out/android-armeabi-v7a-fastbuild/bin/_nativedeps/1_3_0_0_3_1603a52a129c5902439281827f3522ba084d3ca86608ba9c80599aee4f3a27a0.so bazel-out/android-armeabi-v7a-fastbuild/bin/platforms/android/libandroid_rs-3993832558.a external/androidndk/ndk/sources/cxx-stl/llvm-libc++/libs/armeabi-v7a/libandroid_support.a external/androidndk/ndk/sources/cxx-stl/llvm-libc++/libs/armeabi-v7a/libc++_static.a external/androidndk/ndk/sources/cxx-stl/llvm-libc++/libs/armeabi-v7a/libc++abi.a external/androidndk/ndk/sources/cxx-stl/llvm-libc++/libs/armeabi-v7a/libunwind.a '-Wl,-soname=libandroid' -lSystem -lresolv -static-libgcc -Wl,-S -target armv7-none-linux-androideabi -gcc-toolchain external/androidndk/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64 -Lexternal/androidndk/ndk/sources/cxx-stl/llvm-libc++/libs/armeabi-v7a -no-canonical-prefixes -Wl,-z,relro -Wl,--gc-sections '--sysroot=external/androidndk/ndk/platforms/android-30/arch-arm'```

Errors:
```
[...]

PR #1181, opened 2022-03-09
Allow specifying additional target triples in rust_register_toolchains() and fix androideabi system

korDen:
In addition to host target, "wasm32-unknown-unknown" and "wasm32-wasi" targets are being installed, which includes downloading corresponding rust-std libraries. This CL allow specifying custom additional target triples by introducing a new parameter to rust_register_toolchains(), which defaults to ["wasm32-unknown-unknown" and "wasm32-wasi"] for compatibility. As an example, it is now possible to compile for Android if ["armv7-linux-androideabi"] is specified as an extra_target_triples.

Specifically for Android, most of the required infrastructure was already in place. For example, @rules_rust//rust/platform:triple_mappings.bzl already has all the necessary information to compile Android libraries (e.g. _SYSTEM_TO_STDLIB_LINKFLAGS). However, "armv7-linux-androideabi" is not a 'proper' target tripple because 'androideabi' is not a proper system. As such, special handling is required to map 'androideabi' back to 'android'.

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -212,7 +212,10 @@ def triple_to_system(triple):
     if len(component_parts) < 3:
         fail("Expected target triple to contain at least three sections separated by '-'")
 
-    return component_parts[2]
+    system = component_parts[2]
+    if (system == "androideabi"):
+        system = "android"
+    return system
 
 def triple_to_arch(triple):
     """Returns a system architecture name for a given platform triple
@@ -289,6 +292,10 @@ def triple_to_constraint_set(triple):
     vendor = component_parts[1]
     system = component_parts[2]
     abi = None
+
+    if system == "androideabi":
+        system = "android"
+        cpu_arch = "arm"
 
     if len(component_parts) == 4:
         abi = component_parts[3]
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -196,6 +196,36 @@ def abi_to_constraints(abi):
         # figure out how they're doing this
         return []
 
+def triple_to_struct(triple):
+    if triple == "wasm32-wasi":
+        return struct(
+            cpu_arch = "wasi",
+            vendor = None,
+            system = "wasi",
+            abi = None,
+        )
+
+    component_parts = triple.split("-")
+    if len(component_parts) < 3:
+        fail("Expected target triple to contain at least three sections separated by '-'")
+
+    cpu_arch = component_parts[0]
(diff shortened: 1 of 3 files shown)


korDen:
Fixes #1167

rust/platform/triple_mappings.bzl
```diff
@@ -212,7 +212,10 @@ def triple_to_system(triple):
     if len(component_parts) < 3:
         fail("Expected target triple to contain at least three sections separated by '-'")
 
-    return component_parts[2]
+    system = component_parts[2]
+    if (system == "androideabi"):
```
UebelAndre, line 216: Small nit for consistency's sake, if you don't mind
```suggestion
    if system == "androideabi":
```

UebelAndre (changes requested):

rust/platform/triple_mappings.bzl
korDen: Done is a child commit.

rust/platform/triple_mappings.bzl
```diff
@@ -196,6 +196,36 @@ def abi_to_constraints(abi):
         # figure out how they're doing this
         return []
 
+def triple_to_struct(triple):
```
UebelAndre, line 199: This already exists in `@rules_rust//rust/platform/triple.bzl`. Perhaps that could be used instead?

UebelAndre (changes requested):

rust/platform/triple_mappings.bzl
korDen: Done.

rust/platform/triple_mappings.bzl
```diff
@@ -196,58 +198,16 @@ def abi_to_constraints(abi):
...
-    if triple == "wasm32-wasi":
-        return "wasi"
-
-    component_parts = triple.split("-")
-    if len(component_parts) < 3:
-        fail("Expected target triple to contain at least three sections separated by '-'")
-
-    return component_parts[2]
-
-def triple_to_arch(triple):
```
UebelAndre, line 217: Since this module is public, these are part of our public API. Can you restore the functions and maybe print a warning that `@rules_rust//rust/platforms/triple.bzl` should be used instead?

UebelAndre (changes requested):
One last request from me, thanks!

rust/platform/triple_mappings.bzl
korDen: Done.

UebelAndre (approved):
Thank you!

PR #1184, opened 2022-03-10
Regenerated `cargo-raze` outputs with v0.15.0

UebelAndre:


illicitonion (approved):
Sure I'm guessing this is probably fine...!

PR #1186, opened 2022-03-12
Fix iOS linker arguments

keith:
These were interpreted as a single argument, instead of 2 arguments
separated by a space, this lead to this linker error:

```
clang: error: unknown argument: '-framework Security'
clang: error: unknown argument: '-framework Foundation'
```

Instead we can use the `-Wl` syntax so we don't have to worry about
ordering.


keith:
Alternatively unless something specifically from rust depends on these we could probably drop these and assume user's builds already include them as needed.

illicitonion (approved):
Thanks!

keith:
Thanks!

PR #1187, opened 2022-03-12
Create internal rust_binary rule instead of using transitions

dfreese:
This seems like a cleaner way to break the dependency that rust_binary
has on the process wrapper.  The common args are consumed and modified
to set the process wrapper to None, which is detected and performs the
appropriate switch.  The transitions and settings can then be removed.


ddeville:
I can confirm that this fixes some cycles I was seeing when querying the bazel dependency graph after #1159.

rust/private/rustc.bzl
```diff
@@ -922,8 +922,7 @@ def rustc_compile_action(
             dsym_folder = ctx.actions.declare_directory(crate_info.output.basename + ".dSYM")
             action_outputs.append(dsym_folder)
 
-    # This uses startswith as on windows the basename will be process_wrapper_fake.exe.
-    if not ctx.executable._process_wrapper.basename.startswith("process_wrapper_fake"):
+    if ctx.executable._process_wrapper != None:
```
UebelAndre, line 925: Is this meaningfully different from:
```suggestion
    if ctx.executable._process_wrapper:
```

rust/private/rust.bzl
```diff
@@ -928,6 +928,33 @@ rust_binary = rule(
...
+        executable = True,
+        allow_single_file = True,
+        cfg = "exec",
+    )
+    return new_attr
+
+# Provides an internal rust_binary to use that we can use to build the process
+# wrapper, this breaks the dependency of rust_binary on the process wrapper by
+# setting it to None, which the functions in rustc detect and build accordingly.
+rust_binary_for_process_wrapper = rule(
```
UebelAndre, line 944: small nit
```suggestion
rust_binary_without_process_wrapper = rule(
```

UebelAndre (changes requested):
One question but otherwise looks good!

dfreese:
All of those changes look fine to me.  I'll be able to get back to this later Sunday night, so feel free to update and merge. I think you should have permission.

UebelAndre (approved):
Thanks for working on this!

hlopko:
Thanks! Indeed much simpler!

PR #1188, opened 2022-03-14
Add a test showing that rust_analayzer aspect traverses aliases

hlopko:


hlopko:
Apparently Bazel is handling aliases specially and does this for us for free.

illicitonion (approved):

PR #1189, opened 2022-03-14
Specify root target for rust_analyzer test

hlopko:
Without it `gen_rust_project` uses `//...` and that makes the test always pass using the test package no matter what the aspect visits


illicitonion (approved):

test/rust_analyzer/rust_analyzer_test_runner.sh
```diff
@@ -66,7 +66,7 @@ function rust_analyzer_test() {
     
     pushd "${workspace}" &> /dev/null
     echo "Generating rust-project.json..."
-    bazel run "@rules_rust//tools/rust_analyzer:gen_rust_project"
+    bazel run "@rules_rust//tools/rust_analyzer:gen_rust_project" //:mylib_test
```
UebelAndre, line 69: I'm confused as to why this does anything different. Is this supposed to be an argument the `gen_rust_project` target? Wouldn't that require
```suggestion
    bazel run "@rules_rust//tools/rust_analyzer:gen_rust_project" -- //:mylib_test
```
Additionally, I feel like this should work as it was. Doesn't this reduce the scope of what's being tested here?

UebelAndre (changes requested):

test/rust_analyzer/rust_analyzer_test_runner.sh
hlopko: In this case `--` is optional, but I added it anyway for clarity (to be honest it was an oversight on my side and I only now discovered that it works both ways :)).

If we don't specify the `//:mylib_test` we get `//...`. `aspect_traversal_test` is there to verify that the rust analyzer aspect follows edges to all various dependencies and puts them into the `rust-project.json`. If we use `//...`, the wildcard will match all the targets in the package and put them into the `rust-project.json` even when aspect doesn't do anything. This is what was happening and what this PR fixes.

For `merging_crates_test` the choice of label is less important (shouldn't matter). We could improve the machinery so we pick `//:mylib_test` for the `aspect_traversal_test` and something else for other test, but I think it's fine to stay simple and always use the same entry point.

UebelAndre (approved):

PR #1194, opened 2022-03-14
Added support for producing distribution archives

UebelAndre:
This change introduces the `//distro:rules_rust` target which produces a minimal archive of `rules_rust`. This archive is one of the final stepping stones to closing out https://github.com/bazelbuild/rules_rust/issues/415 which is something I'd like to see done this week.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -18,6 +18,25 @@ alias(
     visibility = ["//visibility:public"],
 )
 
+filegroup(
+    name = "distro",
+    srcs = glob(["*.bzl"]) + [
+        "//bindgen:distro",
+        "//cargo:distro",
+        "//crate_universe:distro",
+        "//proto:distro",
+        "//rust:distro",
+        "//tools:distro",
+        "//util:distro",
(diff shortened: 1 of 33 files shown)


hlopko (approved):
Looks good to me, thank you!

One question though. Right now we release everything in rules_rust, including experimental things and things not covered by the compatibility policy. Alternative approach would be to only release stable stuff and ask users who need more to use HEAD. That would emphasize that they are using something not well supported.

I think I'm fine either way, I just wanted to make sure there was a discussion about this.

illicitonion (approved):
LGTM!

> One question though. Right now we release everything in rules_rust, including experimental things and things not covered by the compatibility policy. Alternative approach would be to only release stable stuff and ask users who need more to use HEAD. That would emphasize that they are using something not well supported.
> 
> I think I'm fine either way, I just wanted to make sure there was a discussion about this.

Also fine either way :)

UebelAndre:
Gonna merge for the time being but am happy to continue the discussion and make changes so everyone's happy.

Issue #228, opened 2019-05-21
Consider integrating compilation pipelining


GregBowyer:
We dont have to do this right away, but as https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199 comes to fruit, we could plug this into bazel pretty easily.

acmcarther:
The underlying notes are most relevant for reimplementing this:
https://github.com/rust-lang/compiler-team/blob/master/working-groups/pipelining/NOTES.md

I skimmed it briefly and it mostly made sense, but one snippet caught me off guard

> An alternative solution proposed by @ehuss is that the compiler could print a message on stdout/stderr to Cargo whenever a file has been produced. Cargo already does this, for example, when invoked with --message-format=json. The compiler already emits errors as JSON blobs with --error-format=json, although the compiler doesn't emit other information via JSON right now.

Contrary to the "pro":

> - Should be somewhat usable by other build systems as it's pretty standard to listen to stderr/stdout from spawned processes.

I suspect that it might be more difficult from Bazel's perspective to initiate subsequent compilation steps based on messages on stdout.

acmcarther:
@damienmg:

It looks like the implementation of compilation pipelining here works by having rustc inform the caller (via stdout) when artifacts are generated during execution. Cargo seems to have been updated to use that additional information to dispatch subsequent invocations to `rustc` while prior invocations are still in progress. My very superficial understanding of the internal model of Bazel tells me that this would be equivalent to an action becoming "partially complete" -- something not well supported by Bazel.

Do you know of any similar prior art where Bazel needed to work with tools which run for long durations and incrementally produce artifacts?

---

More detail:

Based on the document, the two kinds of artifacts being produced are 
1. rmeta: build metadata 
2. rlib: linkable libraries

(1) is required in order to initiate subsequent compilation steps and produce dependent rlibs, while (2) is required for final linking. The optimization at play takes advantage of the fact that (1) is generally much faster to produce than (2), and (2) is only required at the very end (IIUC).

mfarrugi:
The two rustc invocation approach already mostly works, but currently `--emit metadata` by itself is not enough to produce the correct metadata.

There's some converation around that @ https://rust-lang.zulipchat.com/#narrow/stream/195180-t-compiler.2Fwg-pipelining and https://github.com/rust-lang/rust/issues/58465#issuecomment-479032740

mfarrugi:
[...]

PR #1207, opened 2022-03-21
process_wrapper: add support for terminating rustc after it emits rmeta.

gigaroby:
This is a key component of supporting pipelining in rules_rust. The
bazel rules will (in a follow-up PR) configure rustc to run either to
completion for rules whose dependencies require the full rlib files or
until they emit the rmeta files if dependencies only require that.

This is safe to commit as there are no changes to user-visible behavior
until the new flags are used.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -59,6 +59,16 @@ def rules_rust_dependencies():
         url = "https://github.com/bazelbuild/apple_support/releases/download/0.11.0/apple_support.0.11.0.tar.gz",
     )
 
+    # process_wrapper needs a low-dependency way to process json.
+    maybe(
+        http_archive,
+        name = "rules_rust_tinyjson",
+        sha256 = "9c21866c7f051ebcefd028996494a374b7408ef946826cefc9761d58cce0fd36",
+        url = "https://github.com/rhysd/tinyjson/archive/refs/tags/v2.3.0.zip",
+        strip_prefix = "tinyjson-2.3.0",
+        build_file = "@rules_rust//util/process_wrapper:BUILD.tinyjson.bazel",
+    )
+
 # buildifier: disable=unnamed-macro
 def rust_register_toolchains(
         dev_components = False,
--- a/test/process_wrapper/fake_rustc.rs
+++ b/test/process_wrapper/fake_rustc.rs
@@ -1,0 +1,8 @@
+//! This binary mocks the output of rustc when run with `--error-format=json` and `--json=artifacts`.
+
+fn main() {
+    eprintln!(r#"{{"rendered": "should be\nin output"}}"#);
+    eprintln!(r#"{{"emit": "metadata"}}"#);
+    std::thread::sleep(std::time::Duration::from_secs(1));
+    eprintln!(r#"{{"rendered": "should not be in output"}}"#);
+}
--- a/test/process_wrapper/rustc_quit_on_rmeta.rs
+++ b/test/process_wrapper/rustc_quit_on_rmeta.rs
@@ -1,0 +1,94 @@
+#[cfg(test)]
+mod test {
+    use std::path::PathBuf;
+    use std::process::Command;
+    use std::str;
+
+    use runfiles::Runfiles;
+
+    // fake_rustc runs the fake_rustc binary under process_wrapper with the specified
+    // process wrapper arguments. No arguments are passed to fake_rustc itself.
+    fn fake_rustc(process_wrapper_args: &[&'static str]) -> String {
+        let r = Runfiles::create().unwrap();
+        let fake_rustc = r.rlocation(
+            &[
+                "rules_rust",
+                "test",
+                "process_wrapper",
+                if cfg!(unix) {
+                    "fake_rustc"
+                } else {
+                    "fake_rustc.exe"
+                },
+            ]
+            .iter()
+            .collect::<PathBuf>(),
+        );
+
+        let process_wrapper = r.rlocation(
+            &[
+                "rules_rust",
+                "util",
+                "process_wrapper",
+                if cfg!(unix) {
+                    "process_wrapper"
+                } else {
+                    "process_wrapper.exe"
+                },
+            ]
+            .iter()
+            .collect::<PathBuf>(),
+        );
+
+        let output = Command::new(process_wrapper)
+            .args(process_wrapper_args)
+            .arg("--")
+            .arg(fake_rustc)
+            .output()
+            .unwrap();
+
+        assert!(
+            output.status.success(),
+            "unable to run process_wrapper: {} {}",
+            str::from_utf8(&output.stdout).unwrap(),
+            str::from_utf8(&output.stderr).unwrap(),
+        );
+
+        String::from_utf8(output.stderr).unwrap()
+    }
+
+    #[test]
+    fn test_rustc_quit_on_rmeta_quits() {
+        let out_content = fake_rustc(&["--rustc-quit-on-rmeta", "true"]);
+        assert!(
+            !out_content.contains("should not be in output"),
+            "output should not contain 'should not be in output' but did: {}",
+            out_content
+        );
+    }
+
+    #[test]
+    fn test_rustc_quit_on_rmeta_output_json() {
+        let json_content = fake_rustc(&[
+            "--rustc-quit-on-rmeta",
+            "true",
+            "--rustc-output-format",
+            "json",
+        ]);
+        assert_eq!(
+            json_content,
+            concat!(r#"{"rendered": "should be\nin output"}"#, "\n")
+        );
+    }
+
+    #[test]
+    fn test_rustc_quit_on_rmeta_output_rendered() {
+        let rendered_content = fake_rustc(&[
+            "--rustc-quit-on-rmeta",
+            "true",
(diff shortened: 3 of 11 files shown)


util/process_wrapper/BUILD.tinyjson
```diff
@@ -0,0 +1,7 @@
+load("@//rust/private:rust.bzl", "rust_library_without_process_wrapper")
```
UebelAndre, line 1: I know this PR is WIP but just so I don't forget, can you rename this file to `BUILD.tinyjson.bazel` so it's picked up by buildifier?

UebelAndre, line 1: I know this PR is WIP but just so I don't forget, can you rename this file to `BUILD.tinyjson.bazel` so it's picked up by buildifier?

util/process_wrapper/rustc.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub(crate) fn process_message(
+    line: String,
+    output: Output,
+    stop: &SyncSender<()>,
+) -> Option<String> {
+    let parsed: JsonValue = line
+        .parse()
+        .expect("process wrapper error: expected json messages in pipeline mode");
+    if let Some(emit) = get_key(parsed.clone(), "emit") {
```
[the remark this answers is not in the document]
gigaroby: There were two places where parsed might have been moved. Modified get_key so it takes a reference to the json object and instead clones the string value out of it. WDYT?

scentini (approved):

util/process_wrapper/output.rs
```diff
@@ -0,0 +1,27 @@
...
+        let mut reader = io::BufReader::new(read_end);
+        let mut writer = io::LineWriter::new(write_end);
+        loop {
+            let mut line = String::new();
+            let read_bytes = reader.read_line(&mut line)?;
+            if read_bytes == 0 {
+                break;
+            }
+            if let Some(to_write) = process_line(line) {
+                writer.write_all(&to_write.into_bytes())?;
```
hlopko, line 22: Is `&to_write.into_bytes()` the same as `to_write.as_bytes()`?

hlopko (approved):

hlopko:
@UebelAndre do you also want to take a look?

test/process_wrapper/rustc_quit_on_rmeta.rs
```diff
@@ -0,0 +1,89 @@
...
+        output.status.success(),
+        "unable to run process_wrapper: {} {}",
+        str::from_utf8(&output.stdout).unwrap(),
+        str::from_utf8(&output.stderr).unwrap(),
+    );
+
+    String::from_utf8(output.stderr).unwrap()
+}
+
+#[test]
```
UebelAndre, line 56: Can this be moved into a `tests` module?
```suggestion
#[cfg(test)]
mod test {
    #[test]
```

test/process_wrapper/rustc_quit_on_rmeta.rs
```diff
@@ -0,0 +1,89 @@
+use std::path::PathBuf;
+use std::process::Command;
+use std::str;
+
+use runfiles::Runfiles;
+
+fn fake_rustc(process_wrapper_args: &[&'static str]) -> String {
```
UebelAndre, line 7: Can you document this function?

util/process_wrapper/rustc.rs
```diff
@@ -0,0 +1,53 @@
...
+            Some(s.clone())
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
+
+pub(crate) fn stop_on_rmeta_completion(
```
UebelAndre, line 28: nit: Can you add a docstring for this function?

rust/repositories.bzl
```diff
@@ -59,6 +59,16 @@ def rules_rust_dependencies():
         url = "https://github.com/bazelbuild/apple_support/releases/download/0.11.0/apple_support.0.11.0.tar.gz",
     )
 
+    # process_wrapper needs a low-dependency way to process json.
+    maybe(
+        http_archive,
+        name = "tinyjson",
```
UebelAndre, line 65: How do folks feel about scoping the name to avoid it impacting users?
```suggestion
        name = "rules_rust_tinyjson",
```

UebelAndre (changes requested):
I mostly have nits but could you also explain "pipelining" to me? What's the workflow that this unblocks?

gigaroby:
@UebelAndre thanks for review :)

> I mostly have nits but could you also explain "pipelining" to me? What's the workflow that this unblocks?

Sure :) Pipelining (in Cargo terms) allows Cargo to unblock a lib -> lib dependency by asking the compiler to emit a metadata file during compilation. When rustc is asked to provide a metadata file, it will produce two files for any given library a .rmeta (the metadata) which is produced quickly and a .rlib (the full object). As soon as Cargo sees the rmeta file, it unblocks all the library targets that depend on that, increasing parallelism. Bazel is (currently) unable to unblock dependencies mid-action so we need two different invocations of the compiler. The first invocation will be killed as soon as the metadata file is produced while the second will continue to completion. The idea is that for rlib -> rlib dependencies we will only run the lightweight metadata invocation, then unblock the dependencies, then compile the full rlib at link time (or whatever other action depends on the full file). You can also read the original rationale for introducing this in Cargo at https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199.

I hope this answers your question.

test/process_wrapper/fake_rustc.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    eprintln!(r#"{{"rendered": "I am a fake rustc\nvery very fake"}}"#);
```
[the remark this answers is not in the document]
UebelAndre: Can you add a comment about that here? 

Also, tiny nit (and not to be a spoil-sport) but could the message be something a bit more serious sounding? 😅 If you really like it though it's fine to leave as is.

dfreese:
Ignorant question: if pipelining requires two runs of cargo, instead of building up a way to terminate rustc, could `--emit=metadata` be the only emit value passed in the first invocation?  What would be the technical rationale for actually needing to kill the process?

Related, it seems like having that as always being the first step would cause the build to fail fast, probably solving #428, more-or-less.

gigaroby:
@UebelAndre are you happy with this change now?

dfreese:
> Sadly no. As I have found out while working on the bazel side of this (which is still WIP), the CLI invocations of the two compiler instances need to be basically identical for this to work. Changing any of the flags (including stuff unrelated to codegen, like --error-format) causes rustc to not accept the generated .rmeta file once we get to the linking stage where we need the .rlib.

Sad day. Has that been flagged to the rust team?  That seems like something rustc should be able to support.

hlopko:
In addition to command line flags changing the content of the rlib as @gigaroby explained, there's also a second issue/feature :) `--emit=metadata` and `--emit=metadata,link` produces different `.rmeta` files. The first produces what you know as `cargo check` outputs. Those `.rmeta` files cannot be used for downstream compilations producing object files since they don't contain all the necessary information, they can only be used to emit most (but strictly speaking not all) compilation errors. If you want to use the `.rmeta` files for compilation, you have to use what cargo pipelining does, which is `--emit=metadata,link` (other combinations work too, like `--emit=metadata,mir`, but then you're hitting the first issue). Unsurprisingly, `.rmeta` file is generated faster with `--emit=metadata` than with `--emit=metadata,link`+process termination.

So regarding #428, we could use the metadata files from pipelining for an equivalent of `cargo check`, but those actions will still be slower than using plain `--emit=metadata`.

util/process_wrapper/output.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub(crate) fn process_output<F>(
+    read_end: Box<dyn Read + Send>,
+    write_end: Box<dyn Write + Send>,
+    mut process_line: F,
+) -> thread::JoinHandle<io::Result<()>>
+where
+    F: FnMut(String) -> Option<String> + Send + 'static,
+{
+    thread::spawn(move || {
```
UebelAndre, line 12: Is it necessary to perform this work concurrently? I've had frustrating experiences with actions actions not playing nice with others by occupying a ton of resources on the host.

test/process_wrapper/fake_rustc.rs
```diff
@@ -0,0 +1,7 @@
+// This binary mocks the output of rustc when run with --error-format=json and --json=artifacts.
```
UebelAndre, line 1: Oddly, `rustdoc` reads `//!` when it's at the top of a file to describe the module. But one more nit.
```suggestion
//! This binary mocks the output of rustc when run with `--error-format=json` and `--json=artifacts`.

```

UebelAndre (changes requested):
Sorry for the delay. I have one more question and my final set of nits. Generally looks good though, great work and thanks!

UebelAndre (approved):
Thanks for putting up with all my boring requests! Looks good to me 😄

jsgf:
BTW, I've successfully implemented Rust pipelining in Buck rules. The key is to not use `--emit metadata`. More details: https://github.com/bazelbuild/rules_rust/issues/228#issuecomment-1100788353

Issue #1063, opened 2021-12-10
Cannot link rust_static_library in a rust_binary target


boxdot:
The symbols from `rust_static_library` are not found when linking it to the `rust_binary` target via `deps`. In the invocation of the linker the path to the library directory is added via `-L`, however there is not library parameter with `-l`. Also note that the library has a numeric suffix generated from its name.

A workaround is possible via defining a `cc_library` target, linking the `rust_static_library` to it, and then using it in the `rust_binary` target.

Example BUILD file and repository to reproduce:

https://github.com/boxdot/bazel-rust-linking-issue/blob/fc92fb2b6384c78f9a5df95035c5dd882ad11694/BUILD.bazel

UebelAndre:
@hlopko curious what your thoughts are on this interaction.

hlopko:
Hey @boxdot thank you for the repro, it was super helpful. There are indeed various bugs that I think I hacked out in https://github.com/bazelbuild/rules_rust/pull/1219 (however that draft is not ready for submission, it needs tests and potentially polishing - feel free to take over :). See the draft description for details.

On a different level though a question begs to be asked - why are you using `rust_static_library` if you are using `rust_binary` for linking? I believe that rustc doesn't support this scenario. Why not use `rust_library`?

boxdot:
Thank you for providing a possible fix.

>  why are you using `rust_static_library`

As in my example, the dependency is on the C-level. It is not a usual Rust crate dependency. The rust lib implements a C-API. The final rust binary uses the C-API loosely via linking the needed symbols. So, we need to link the library as static library into the Rust binary. C-API is important here, because in a real-world scenario, the rust binary might link different implementations of the C-API. Or, the Rust implementation of the C-API can be linked into different binaries (Rust or not).

hlopko:
I see, then I'm afraid rustc doesn't really support your use case right now. The `staticlib` crate output contains allocators shims generated, and `bin` crate output also generates them. The linking fails with duplicated symbols (on linux, I haven't tested on mac, but even if it passed there it wouldn't be a robust solution). There is nothing we can do on rules_rust side to fix it.

Is it an option to use `cc_binary` instead of `rust_binary`? That is fully supported by rules_rust and by rustc.

scentini:
[...]

PR #1219, opened 2022-03-24
DO_NOT_SUBMIT: Do not provide CrateInfo from staticlib/cdylib

hlopko:
This draft should be split into smaller PRs and needs tests.

This draft addresses https://github.com/bazelbuild/rules_rust/issues/1063.
There are multiple small fixes:

* currently we don't include the output hash for cdylib, but we do for
  staticlib. It's more accurate to not include the output hash for
  staticlib as well.
* currently rust_static_library and rust_shared_library announce they
  provide CrateInfo, DepInfo, and DefaultInfo. They should announce
  providing CcInfo (and maybe DefaultInfo - I thought that is implied by
  default so there's no need to explicitly mention that)
* currently rust_static_library and rust_shared_library actually
  provide CrateInfo and DepInfo. That is incorrect - outputs of these
  rules are not meant for consumption by Rust rules. These rules are
  meant to be used when leaving the world of Rust rules and entering the
  world of other language that wants to depend on native library (and it
  doesn't matter to this user if the library is written in Rust, C, C++,
  or other)

Lastly, this PR makes https://github.com/boxdot/bazel-rust-linking-issue build. Surprising plot twist - it is actually not supported to link rust_static_library into a rust_binary. From rustc docs:

> --crate-type=staticlib, #[crate_type = "staticlib"] - A static system
> library will be produced. This is different from other library outputs
> in that the compiler will never attempt to link to staticlib outputs.
> The purpose of this output type is to create a static library containing
> all of the local crate's code along with all upstream dependencies. This
> output type will create *.a files on Linux, macOS and Windows (MinGW),
> and *.lib files on Windows (MSVC). This format is recommended for use in
> situations such as linking Rust code into an existing non-Rust
> application because it will not have dynamic dependencies on other Rust
> code.

https://doc.rust-lang.org/reference/linkage.html#linkage.

The `bazel-rust-linking-issue` fails at linktime since there are
multiple definitions of allocator stubs. We may want to consider
detecting this situation and failing the build with a nice error message
saying this is not how these rules are intended to be used.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -249,7 +249,7 @@ def _rust_library_common(ctx, crate_type):
     # deterministic name is important since it ends up embedded in the executable. This is problematic when one needs
     # to include the library with a specific filename into a larger application.
     # (see https://github.com/bazelbuild/rules_rust/issues/405#issuecomment-993089889 for more details)
-    if crate_type != "cdylib":
+    if crate_type not in ["cdylib", "staticlib"]:
         output_hash = determine_output_hash(crate_root, ctx.label)
     else:
         output_hash = None
@@ -731,7 +731,7 @@ rust_library = rule(
 
 rust_static_library = rule(
     implementation = _rust_static_library_impl,
-    provides = _common_providers,
+    provides = [CcInfo],
     attrs = dict(_common_attrs.items()),
     fragments = ["cpp"],
     host_fragments = ["cpp"],
@@ -755,7 +755,7 @@ rust_static_library = rule(
 
 rust_shared_library = rule(
     implementation = _rust_shared_library_impl,
-    provides = _common_providers,
+    provides = [CcInfo],
     attrs = dict(_common_attrs.items()),
     fragments = ["cpp"],
     host_fragments = ["cpp"],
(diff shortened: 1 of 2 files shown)


scentini:
Closing this draft PR:

> - currently we don't include the output hash for cdylib, but we do for staticlib. It's more accurate to not include the output hash for staticlib as well

- was addressed by https://github.com/bazelbuild/rules_rust/pull/1222

> - currently rust_static_library and rust_shared_library announce they provide CrateInfo, DepInfo, and DefaultInfo. They should announce providing CcInfo (and maybe DefaultInfo - I thought that is implied by default so there's no need to explicitly mention that)

- was addressed by https://github.com/bazelbuild/rules_rust/pull/1298, except that `rust_static_library` and `rust_shared_library` do not advertise `CcInfo`, as the Rust rules don't provide a `CcInfo` on `wasm32` 

> - currently rust_static_library and rust_shared_library actually provide CrateInfo and DepInfo. That is incorrect - outputs of these rules are not meant for consumption by Rust rules

- was also addressed by https://github.com/bazelbuild/rules_rust/pull/1298

PR #1222, opened 2022-03-24
don't emit --codegen={metadata,extra-filename} for rust_static_library and rust_shared_library

krasimirgg:
Extracted from https://github.com/bazelbuild/rules_rust/pull/1219/files, addressing this part:
* currently we don't include the output hash for cdylib, but we do for staticlib. It's more accurate to not include the output hash for staticlib as well.

Related to, but does not address https://github.com/bazelbuild/rules_rust/issues/1063.

@scentini suggested to not even emit --codegen flags in these cases.

Co-authored-by: Marcel Hlopko <hlopko@google.com>


test/unit/crate_name/crate_name_test.bzl
```diff
@@ -71,6 +71,17 @@ def _slib_library_name_test_impl(ctx):
...
 
+def _no_extra_filename_test_impl(ctx):
+    """Check that no extra filename is used.
+
+    Args:
+      ctx: rule context.
+    """
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    assert_argv_contains(env, tut.actions[0], "--codegen=extra-filename=")
```
scentini, line 82: WDYT about removing --codegen=extra-filename= from the command line altogether?

krasimirgg: good idea, let me try

scentini (approved):

rust/private/rustc.bzl
```diff
@@ -714,12 +714,14 @@ def construct_arguments(
     if hasattr(attr, "_error_format"):
         rustc_flags.add("--error-format=" + attr._error_format[ErrorFormatInfo].error_format)
 
-    # Mangle symbols to disambiguate crates with the same name
-    extra_filename = "-" + output_hash if output_hash else ""
-    rustc_flags.add("--codegen=metadata=" + extra_filename)
+    if output_hash:
+        # Mangle symbols to disambiguate crates with the same name
+        extra_filename = "-" + output_hash
+        rustc_flags.add("--codegen=metadata=" + extra_filename)
```
UebelAndre, line 720: I'm a bit confused by this gating. Can you add a comment about why these two flags are now conditionally added?

UebelAndre (changes requested):
If I understand this right this change seems good. Just one clarifying request from me

rust/private/rustc.bzl
krasimirgg: added some comments

UebelAndre (approved):
Thanks!

PR #1226, opened 2022-03-24
Fix some unit tests to run from another workspace

siddharthab:
Some tests assume that they are running in the rules_rust workspace.
Removing these assumptions allow us to run these tests from other bazel
workspaces.

For example, I help maintain https://github.com/grailbio/bazel-toolchain
and we check if cc toolchain that we configure is compatible with
rules_rust by running the unit tests in this repo.

### DIFF
--- a/test/unit/linkstamps/linkstamps_test.bzl
+++ b/test/unit/linkstamps/linkstamps_test.bzl
@@ -20,7 +20,8 @@ def _supports_linkstamps_test(ctx):
     linkstamp_out = linkstamp_action.outputs.to_list()[0]
     asserts.equals(env, linkstamp_out.basename, "linkstamp.o")
     tut_out = tut.files.to_list()[0]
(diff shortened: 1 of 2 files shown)


UebelAndre (approved):
This looks good to me but I'm not sure if we have a way to regression test something like this so new tests might be added that regress this functionality. Do you know how other rules handle this?

siddharthab:
I don't think it's important to regression test this. I submitted similar changes to rules_go as well. If you do want to regression test it, you can run your test script from a different (dummy) workspace in your git repo.

siddharthab:
Note that the dummy workspace for the regression test won't really be dummy because it will still need to define all the test transitive dependencies (at least until bazel modules are more common).

But I really think it's ok to not have regression tests; it's not common for users to run tests of their dependencies from their own workspace. For the few who do, they can submit fixes as and when these things break.

illicitonion (approved):
LGTM - thanks!

PR #1227, opened 2022-03-25
Fixed `crates_repository` deleting `.cargo/config.toml` files.

UebelAndre:
Uses of [`crates_repository::cargo_config = "//:.cargo/config.toml"`](https://github.com/bazelbuild/rules_rust/blob/0.1.0/crate_universe/private/crates_repository.bzl#L121-L123) would unfortunately cause that file to be deleted. This PR fixes that.

### DIFF
--- a/crate_universe/src/cli/splice.rs
+++ b/crate_universe/src/cli/splice.rs
@@ -2,6 +2,7 @@
 
 use std::path::PathBuf;
 
+use anyhow::Context;
 use clap::Parser;
 
 use crate::cli::Result;
@@ -26,10 +27,14 @@ pub struct SpliceOptions {
     #[clap(long)]
     pub cargo_lockfile: Option<PathBuf>,
 
-    /// The directory in which to build the workspace. A `Cargo.toml` file
-    /// should always be produced within this directory.
+    /// The directory in which to build the workspace. If this argument is not
+    /// passed, a temporary directory will be generated.
     #[clap(long)]
-    pub workspace_dir: PathBuf,
+    pub workspace_dir: Option<PathBuf>,
+
+    /// The location where the results of splicing are written.
+    #[clap(long)]
+    pub output_dir: PathBuf,
 
     /// If true, outputs will be printed instead of written to disk.
     #[clap(long)]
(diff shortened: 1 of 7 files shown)


crate_universe/src/splicing/splicer.rs
```diff
@@ -383,12 +368,34 @@ impl<'a> SplicerKind<'a> {
...
+                                "Failed to delete existing cargo config: {}",
+                                config.display()
+                            )
+                        })?;
                     }
                 }
             }
         }
 
+        // Make sure no other config files exist
```
illicitonion, line 382: I believe cargo will traverse parents all the way up to `/` to find config files - should we maybe detect (but presumably error rather than try to delete) those too?

illicitonion (approved):

crate_universe/src/splicing/splicer.rs
UebelAndre: Oh, I didn't know that. Yeah, the logic here should only be trying to remove unspecified config files from the Cargo workspace so they don't impact them. If there's one at the system root or something I think this should error since I otherwise don't think it'd be obvious why the generated lockfile doesn't match any other host.

UebelAndre: Ok. Updated and added some tests.

crate_universe/src/splicing/splicer.rs
```diff
@@ -383,12 +368,53 @@ impl<'a> SplicerKind<'a> {
...
+        let mut current_parent = workspace_dir.parent();
+        while let Some(parent) = current_parent {
+            let dot_cargo_dir = parent.join(".cargo");
+            for config in vec![
+                dot_cargo_dir.join("config.toml"),
+                dot_cargo_dir.join("config"),
+            ] {
+                if config.exists() {
+                    bail!(
+                        "A Cargo config file was found in a parenting directory to the current workspace. \nWorkspace = {}\nCargo config = {}",
```
illicitonion, line 409:
```suggestion
                        "A Cargo config file was found in a parent directory to the current workspace. This is not allowed because these settings will leak into your Bazel build but will not be reproducible on other machines.\nWorkspace = {}\nCargo config = {}",
```

illicitonion (approved):
LGTM - thanks!

UebelAndre:
@illicitonion I double checked the new code that would error on Cargo config files being detected and realized on my linux machine I have [--output_user_root](https://docs.bazel.build/versions/main/command-line-reference.html#flag--output_user_root) set which meant I conveniently circumvented the fact that my home directory does indeed have a config file. The additional change runs generation in a temporary directory so hosts that might still want to have a normal cargo setup with a config can build using these rules.

crate_universe/src/cli/splice.rs
```diff
@@ -55,12 +60,15 @@ pub fn splice(opt: SpliceOptions) -> Result<()> {
     let extra_manifests_manifest =
         ExtraManifestsManifest::try_from_path(opt.extra_manifests_manifest)?;
 
+    // Determine the splicing workspace
+    let temp_dir = tempfile::tempdir().context("Failed to generate temporary directory")?;
+    let splicing_dir = match &opt.workspace_dir {
+        Some(dir) => dir.clone(),
+        None => temp_dir.as_ref().to_path_buf(),
+    };
```
illicitonion, line 68: We can avoid making the tempdir if we're not going to use it:

```suggestion
    // Determine the splicing workspace
    let temp_dir;
    let splicing_dir = match &opt.workspace_dir {
        Some(dir) => dir.clone(),
        None => {
            temp_dir = tempfile::tempdir().context("Failed to generate temporary directory")?;
            temp_dir.as_ref().to_path_buf()
        },
    };
```

illicitonion (approved):

PR #1230, opened 2022-03-26
Allow isolation and coniguration of the bootstrap cargo runs.

wt:
This will allow the isolation and configuration of the cargo-bazel
bootstrap builds. This allows a custom configuration for cargo to
be used in the bootstrap process. I have successfully used this to
build on an isolated machine with a caching proxy of crates.io.


cargo/cargo_environ.bzl
```diff
@@ -0,0 +1,28 @@
+"""Helper functions for generating the environment for cargo executions."""
+
+load("//cargo/private:cargo_utils.bzl", "cargo_home_path")
+
+CARGO_BAZEL_ISOLATED = "CARGO_BAZEL_ISOLATED"
```
UebelAndre, line 5: This is unique to `crate_universe` and should probably stay within that package.

cargo/private/cargo_utils.bzl
```diff
@@ -151,18 +151,41 @@ def _resolve_repository_template(
...
 
     Returns:
         struct: A struct containing the labels of expected tools
     """
+
+    # This is a bit hidden but to ensure Cargo behaves consistently based
+    # on the user provided config file, the config file is installed into
+    # the `CARGO_HOME` path. This is done so here since fetching tools
+    # is expected to always occur before any subcommands are run.
+    if repository_ctx.attr.isolated and repository_ctx.attr.cargo_config:
```
UebelAndre, line 183: This functionality seems unrelated to the rest of `get_rust_tools`. Could this perhaps go in a separate function?

cargo/cargo_bootstrap.bzl
```diff
@@ -239,6 +241,10 @@ cargo_bootstrap_repository = repository_rule(
             ],
             default = "release",
         ),
+        "cargo_config": attr.label(
+            doc = "The cargo config file.",
+            allow_single_file = [".cfg"],
```
UebelAndre, line 246: Aren't cargo configuration files expected to be named either `config` or `config.toml`? Where does `.cfg` come from? Sorry if I'm overlooking something in https://doc.rust-lang.org/cargo/reference/config.html

UebelAndre (changes requested):
Thanks for putting this together and sorry for the delay here. I saw CI was red but looked closer to find that it was only docs (which is fixed by running `./docs/update_docs.sh` from the root of the repo).  First review pass is now complete 😄

wt:
Does anyone know what the docs test failure is trying to tell me?

cargo/private/cargo_utils.bzl
```diff
@@ -151,18 +151,41 @@ def _resolve_repository_template(
...
+
+def get_rust_tools(cargo_template, rustc_template, host_triple, version, repository_ctx):
     """Retrieve `cargo` and `rustc` labels based on the host triple.
 
     Args:
         cargo_template (str): A template used to identify the label of the host `cargo` binary.
         rustc_template (str): A template used to identify the label of the host `rustc` binary.
         host_triple (struct): The host's triple. See `@rules_rust//rust/platform:triple.bzl`.
         version (str): The version of Cargo+Rustc to use.
+        repository_ctx (repository_ctx): The rule's context object.
```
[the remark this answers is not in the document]
wt: I combine the `get_rust_tools` functions multiple files into this one. One of them required `repository_ctx`. Given that multiple attrs from it are used, I'd prefer to leave that refactor to a future commit since I think it will require non-trivial plumbing. Would that be okay?

cargo/cargo_environ.bzl
wt: It no longer is unique to crate_universe after this change. Would the following work?

* Introduce a new `CARGO_ISOLATED` env var and use it int he cargo directory
* Make `CARGO_BAZEL_ISOLATED` default to the value of `CARGO_ISOLATED`

cargo/private/cargo_utils.bzl
wt: The reason that I left this in is that it's used no where else and would be a function with side effects (creating the symlink). Do you have a suggestion for the name of such a function? I am happy to move it if you feel strongly about it.

Also, this function would take a `repository_ctx`. I so this would mean that would need to continue to be passed to the `get_rust_tools`. This is one of the places where there is fallout from not passing the `repository_ctx` that I mentioned above.

wt (commented):
@UebelAndre Can you please share your thoughts on my replies? Thanks!

UebelAndre (changes requested):

cargo/cargo_bootstrap.bzl
wt: I'm just going to remove the extension restriction. I don't see any good reason for it.

PR #1248, opened 2022-04-07
rust_test_suite: ensure crate names are valid

durin42:
I noticed this while setting up rust_test_suite auto-generation at
work. Seems like a simple enough fix.


illicitonion (approved):
Thanks!

Issue #1233, opened 2022-03-29
`rust_doc_test` failing with "Failed to load argument file"


UebelAndre:
I'm not entirely sure when this started occurring but I'm seeing the following error in some `rust_doc_test` targets
```
==================== Test output for //:rustdoc_test:
error: Failed to load argument file: IO Error: rustdoc_test.rustdoc_test.sh-0.params: No such file or directory (os error 2)

================================================================================
```

Through some bisecting I found the change in behavior to happen between bbccf77928c448f2ed136bb417f8f0affaa18ac8 and c435cf4478fc6e097edc5dba0e71de6608ab77d8 but that delta doesn't make much sense to me since neither of these commits look like they'd impact this behavior. Perhaps the error message is not deterministic. This also no longer seems to happen for hosts other than windows. More info is needed.

tp-woven:
In case it helps, we're also seeing this error, on both Linux and MacOS.

UebelAndre:
> In case it helps, we're also seeing this error, on both Linux and MacOS.

Could you share what commit of `rules_rust` you started seeing this on?

UebelAndre:
I also hit this error in https://github.com/bazelbuild/rules_rust/pull/1252. Since the `//crates_universe:rustdoc_test` target worked before that change on MacOS, it can hopefully reveal the root cause of this issue.

tp-woven:
> > In case it helps, we're also seeing this error, on both Linux and MacOS.
> 
> Could you share what commit of `rules_rust` you started seeing this on?

We're on `ea22478f6dac4604b4ef38fb8e713cd1c0880084`. This is the first time we're trying to use this rule in our repo, so we don't have "good" commits where this is working.

fmorency:
I'm also noticing the issue on the project I'm porting to Bazel

```
error: Failed to load argument file: IO Error: src/many/many_doc_test.rustdoc_test.sh-0.params: No such file or directory (os error 2)
```

I'm using `rules_rust` git e83d5f3c77d2b41a0ad2fb9c5a0fb1dfb5e912a7 on Linux.

PR #1252, opened 2022-04-08
Workaround for issue in linux Cargo binaries

UebelAndre:
This allows `crate_universe` to handle https://github.com/rust-lang/cargo/issues/10547.

Reviewers should only focus on 2e9facb8920327af37e1e5566f7940903138c2ca


illicitonion (approved):

UebelAndre:
After re-pinning i ran into https://github.com/bazelbuild/rules_rust/issues/1253. The solution for this is f19ea439bf47dc4806b087af911e4e3131bc7972.

UebelAndre:
Also after re-pinning it seems `//crate_universe:rustdoc_test` started failing on MacOS. This PR would be a good candidate to investigate the root cause for https://github.com/bazelbuild/rules_rust/issues/1233

Issue #1251, opened 2022-04-08
Targets may build with a different edition when imported into a different workspace


silas-enf:
For a concrete example of the problem, see https://github.com/dtolnay/cxx/issues/1031.

Scenario:
* I set a rust edition default in my WORKSPACE (e.g. `rust_repositories(..., edition="2021", ...)`)
* Another developer sets a different rust edition (e.g. "2018") in their WORKSPACE and uses that default for their rust rules
* I import and use their code in my project

(Naively) expected behavior:
* Targets from the imported repository build with `edition="2018"` where not overridden at the target level
* Targets from my repository build with `edition="2021"` where not overriden at the target level

Actual behavior:
* All targets default to `edition="2021"` regardless.
* Some targets which relied on `edition="2018"` break as a result.

My understanding is that this is a bazel limitation more than anything -- bazel doesn't consult the WORKSPACE of an imported project at all, and from what I can tell doesn't offer much of a middle ground between setting global defaults in the workspace on the one hand and requiring the argument for all targets that use it on the other.

I don't know if there's a satisfying way to fix this, but it'd be good to at least have the docs mention the limitation and mitigation.

PR #1254, opened 2022-04-11
Set edition for process_wrapper and cargo_build_script_runner

dtolnay:
These otherwise get built with the downstream repo's default edition, which is not necessarily correct.

Related to #1251.


illicitonion (approved):
Thanks!

PR #1256, opened 2022-04-11
Remove DEFAULT_RUST_EDITION

dtolnay:
This is 1 commit on top of #1255. It removes `DEFAULT_RUST_EDITION` as the default `edition` argument of `rust_repositories` and various other places. Instead, if the downstream WORKSPACE does not pass an `edition` to `rust_repositories`, the `edition` becomes mandatory in every `rust_*` target in their workspace, rather than using an arbitrary fallback default defined within rules_rust (previously 2018).

This approach was suggested by @dfreese in https://github.com/bazelbuild/rules_rust/pull/1255#issuecomment-1095049266 and +1 by @UebelAndre in https://github.com/bazelbuild/rules_rust/pull/1255#pullrequestreview-938108921. I am ambivalent between them and would be happy with either this or #1255 by itself.


rust/private/rust.bzl
```diff
@@ -106,18 +106,21 @@ def _determine_lib_name(name, crate_type, toolchain, lib_hash = None):
...
         attr (struct): The current rule's attributes
         toolchain (rust_toolchain): The `rust_toolchain` for the current target
+        label (Label): The label of the target being built
 
     Returns:
         str: The target Rust edition
     """
     if getattr(attr, "edition"):
         return attr.edition
+    elif toolchain.default_edition == "":
```
dtolnay, line 122: I attempted to use `toolchain.default_edition == None`, but `None` is not a valid default value for an attribute of type `attr.string` and I didn't see how to make an optional nullable string attribute. Please let me know whether this seems fine.

dtolnay:
Sorry about all the force pushes -- it took a bit of iteration with CI to get all the platform-specific and `"manual"` targets filled in correctly.

I figure we do *not* want to set a default edition in the top-level WORKSPACE.bazel because doing so would mask problems such as #1254, so some new `edition` attrs are required in this PR at the target level. I have extracted that commit to #1257 in case it's easier to review separately and because it's mergeable separately from this PR.

rust/private/rust.bzl
UebelAndre: Could this work?
```suggestion
    elif not toolchain.default_edition:
```

rust/private/common.bzl
```diff
@@ -55,6 +54,5 @@ rust_common = struct(
     crate_info = CrateInfo,
     dep_info = DepInfo,
     stdlib_info = StdLibInfo,
-    default_edition = DEFAULT_RUST_EDITION,
```
UebelAndre, line 58: Maybe it's worth leaving this for a little bit to reduce the impact to public API? I'm torn though.

UebelAndre (approved):
I feel like there's something I'm not thinking about regarding the impact of this change but what I've come up with seems like there's good guidance on how to maintain the current behavior. So this looks good to me, give or take a few nits, thanks!

rust/private/common.bzl
UebelAndre: I don't have any strong answers. Just noticing we're deleting a value from `rust_common` and thought maybe there should be more comms about this. But it can also be a note in the next release.

UebelAndre:
@dtolnay can you rebase this one so we can get it merged, and then https://github.com/bazelbuild/rules_rust/pull/1255?

dtolnay:
- Rebased past #1257

PR #1257, opened 2022-04-11
Outside of examples, fill in all `edition` attrs

dtolnay:
This is extracted from #1256 in case these changes are easier to review separately.

This is effectively a fully exhaustive version of #1254. Due to #1251, specifying `edition` in each target is required in order for downstream workspaces importing `rules_rust` to get a usable edition for all of `rules_rust`'s targets, instead of attempting to use the downstream workspace's default edition which won't necessarily be appropriate.

Note that under `examples`, I have added `edition` to the `rust_register_toolchains` in various WORKSPACE.bazel files rather than to each individual target. This reduces noise in the example projects. I have not done the same for the top-level WORKSPACE because doing so would mask trouble along the lines of #1254.

### DIFF
--- a/cargo/bootstrap/BUILD.bazel
+++ b/cargo/bootstrap/BUILD.bazel
@@ -14,6 +14,7 @@ rust_binary(
     srcs = [
         "bootstrap_installer.rs",
     ],
+    edition = "2018",
     rustc_env = {
         "RULES_RUST_CARGO_BOOTSTRAP_BINARY": "$(rootpath bootstrap_installer.rs)",
     },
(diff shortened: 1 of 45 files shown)


UebelAndre (approved):
Thanks!

UebelAndre:
@dfreese do you have time to take another look at these PRs?

UebelAndre:
In the absence of @dfreese, my understanding is that this generally meets their concerns and I think this and the related changes are ready to merge. @dtolnay can you rebase?

dtolnay:
- Rebased to resolve conflict with #1207, which added some new Rust targets requiring `edition` in *util/process_wrapper* and *test/process_wrapper*

dtolnay:
- Rebased again to resolve conflict with #1286

PR #1259, opened 2022-04-13
Add workaround for arm vs armv7 android issue

keith:
When building for the armv7 android target, we would ideally specify a
cpu:armv7 constraint, but because of the linked bazel issue this is
currently hardcoded as arm in bazel's NDK setup. As far as I can tell
the platform we request for use with rules rust has to match the
constraints in bazel, otherwise it won't discover the NDK CC toolchain
for linking.

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -269,6 +269,13 @@ def triple_to_constraint_set(target_triple):
             "@rules_rust//rust/platform/os:unknown",
         ]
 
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
This seems acceptable to me. But @hlopko could you take a look too? Feel free to merge if you think this is acceptable too.

keith:
Note I also submitted https://github.com/bazelbuild/bazel/pull/15248 but it's unclear to me if that will be a reasonable change, it might break some stuff that I don't know about

keith:
Thanks all!

PR #1266, opened 2022-04-16
crate_universe: propagate build_script_tools

duarten:
Propagate the build_script_tools attribute from the crate annotations to
the generated cargo_build_script rule.


UebelAndre (approved):
Thanks!

UebelAndre (approved):

Issue #1268, opened 2022-04-19
updates needed for native-links-modifiers RFC stabilization


krasimirgg:
Related to https://github.com/bazelbuild/rules_rust/issues/637.
As part of the stabilization of [the native-link-modifiers RFC](https://rust-lang.github.io/rfcs/2951-native-link-modifiers.html), rust-lang commit https://github.com/rust-lang/rust/commit/1004783ef9bdcf006f0ed33badacf83a5934feb2 changes the way rustc lowers `-l` flags to the linker invocation: an invocation like `rustc … -lstatic=lib`:
* before produced `-Wl,--whole-archive -llib -Wl,--no-whole-archive`
* after produces a plain `-llib` linker flag fragment.

After this change, it's now possible to instruct rustc to produce a whole-archive linker invocation fragment by using the new `whole-archive` linker modifier (`rustc … -lstatic:+whole-archive=lib ...`).
This behavior change is stabilized in recent rustc nightlies.

This is a good rustc change since it gives fine-grained control over the exact linker flavor (e.g., as a whole archive) of native dependencies to final artifacts, which will allow us to make progress on https://github.com/bazelbuild/rules_rust/issues/637.

However the new behavior causes the linking of final rules_rust targets to fail in some cases containing hybrid code interfacing between Rust and native libraries, where it was working previously: with the behavior before native libraries were passed as whole-archives to the linker which effectively disabled the linker backward reference diagnostics and made their linker flag fragments largely order-independent. With the new behavior, these turn into hard linker errors sometimes, e.g.,`ld: error: backward reference detected.` when linking with `ld` with `--warn-backrefs`. Internally we've got a few dozed cases, fixing them might take a while and potentially may need additional rules_rust work to adapt our linker flag generation with the new rustc behavior.

To give us some time to diagnose and fix these issues, I'd like to start by introducing a stopgap attribute to allow us to switch back to the old behavior for problematic targets and selectively migrate them to the new behavior one by one. Specifically, I'd like to add a new `experimental_use_whole_archive_for_native_deps: bool; default: false` attribute to common rust_target attributes. If explicitly set to `true` for a final target, its effect would be to use the `whole-archive` modifier for native dependencies (e.g., `-lstatic:+whole-archive=lib` instead of the standard `-lstatic=lib`).

I'd like to use this issue to track whatever additional work needs to be done to adapt rules_rust for the new rustc behavior and the eventual removal of the stopgap attribute.

PR #1269, opened 2022-04-19
add a stopgap experimental_use_whole_archive_for_native_deps attribute

krasimirgg:
This adds a new stopgap attribute enabling us to fall back to the old rustc behavior wrt whole-archive handling, see https://github.com/bazelbuild/rules_rust/issues/1268.


rust/private/rustc.bzl
```diff
@@ -780,7 +780,7 @@ def construct_arguments(
             rustc_flags.add("--codegen=linker=" + ld)
             rustc_flags.add_joined("--codegen", link_args, join_with = " ", format_joined = "link-args=%s")
 
-        _add_native_link_flags(rustc_flags, dep_info, linkstamp_outs, ambiguous_libs, crate_info.type, toolchain, cc_toolchain, feature_configuration)
+        _add_native_link_flags(rustc_flags, dep_info, linkstamp_outs, ambiguous_libs, crate_info.type, toolchain, cc_toolchain, feature_configuration, ctx.attr.experimental_use_whole_archive_for_native_deps)
```
scentini, line 783:
```suggestion
        _add_native_link_flags(rustc_flags, dep_info, linkstamp_outs, ambiguous_libs, crate_info.type, toolchain, cc_toolchain, feature_configuration, attr.experimental_use_whole_archive_for_native_deps)
```

We don't want to access attributes directly from `ctx` because that causes failures when we reach the code via an aspect -> in such cases one needs `ctx.rule.attr` instead of `ctx.attr`.

rust/private/rustc.bzl
```diff
@@ -1347,15 +1350,15 @@ def _add_native_link_flags(args, dep_info, linkstamp_outs, ambiguous_libs, crate
         make_link_flags = _make_link_flags_default
 
     # TODO(hlopko): Remove depset flattening by using lambdas once we are on >=Bazel 5.0
-    args_and_pic_and_ambiguous_libs = [(arg, use_pic, ambiguous_libs) for arg in dep_info.transitive_noncrates.to_list()]
-    args.add_all(args_and_pic_and_ambiguous_libs, map_each = _libraries_dirnames, uniquify = True, format_each = "-Lnative=%s")
+    args_and_pic_and_ambiguous_libs_and_experimental_use_whole_archive_for_native_deps = [(arg, use_pic, ambiguous_libs, experimental_use_whole_archive_for_native_deps) for arg in dep_info.transitive_noncrates.to_list()]
```
scentini, line 1353: This name is getting out of hand :) WDYT about dropping "experimental" from the variable name?

scentini (changes requested):
I guess an integration test is not easy to come up with, but could we add a unit test that tests that the commandline is what we expect when the attribute's value is `True`?

rust/private/rustc.bzl
krasimirgg: Thanks! Good to know!

krasimirgg:
After chatting 1:1, it appears currently there's no good way to test this (the new feature requires a recent nightly, we don't have a nightly build bot setup yet, and we don't have a fully unit-test style test framework that doesn't attempt to always analyse the target_under_test-s). Will work on improvements to the test infra have some form of rustc nightly bot coverage in the future.

scentini (approved):

Issue #1277, opened 2022-04-20
crate_universe: additive_build_file annotations merged into wrong files (stacked)


uhthomas:
**rules_rust version:** b778fca0ac4dfcf3c9e24f1517e73d39d739c730

```starlark
load("@rules_rust//crate_universe:defs.bzl", "crate", "crates_repository")

crates_repository(
    name = "crate_index",
    annotations = {
        "rand": [crate.annotation(additive_build_file_content = "# hello rand")],
        "serde": [crate.annotation(additive_build_file_content = "# hello serde")],
    },
    generator = "@cargo_bazel_bootstrap//:cargo-bazel",
    lockfile = "//:Cargo.Bazel.lock",
    packages = {
        "rand": crate.spec(version = "0.8.5"),
        "serde": crate.spec(version = "1.0.136"),
    },
)

load("@crate_index//:defs.bzl", "crate_repositories")

crate_repositories()
```

The resulting [Cargo.Bazel.lock](https://gist.github.com/uhthomas/5f89865499d5b5dea05ec5cbc2fc00cf).

Then, the generated build files:

```
❯ bazel query --output=location @crate_index__rand-0.8.5//...
❯ bazel query --output=location @crate_index__serde-1.0.136//...
```

<img width="871" alt="Screenshot 2022-04-20 at 16 55 33" src="https://user-images.githubusercontent.com/9749173/164272469-faa325e4-f91b-4b43-bee1-f10946c82da0.png">

<img width="897" alt="Screenshot 2022-04-20 at 16 55 54" src="https://user-images.githubusercontent.com/9749173/164272541-d36d8739-0513-4f01-8b18-b5dd54fe9f24.png">

uhthomas:
It looks like this is actually included within the lockfile ([here](https://gist.github.com/uhthomas/5f89865499d5b5dea05ec5cbc2fc00cf#file-cargo-bazel-lock-L453)), so it's a bug within the generator.

Likely the starlark repository rule `generate_config`.

https://github.com/bazelbuild/rules_rust/blob/b778fca0ac4dfcf3c9e24f1517e73d39d739c730/crate_universe/private/generate_utils.bzl#L196

UebelAndre:
I think the issue is just that `contents` needs to be within the for-loop:
https://github.com/bazelbuild/rules_rust/blob/0.2.1/crate_universe/private/generate_utils.bzl#L208

[...]

PR #1278, opened 2022-04-20
Don't leak additive build file content across crates

uhthomas:
Fixes https://github.com/bazelbuild/rules_rust/issues/1277

### DIFF
--- a/crate_universe/private/generate_utils.bzl
+++ b/crate_universe/private/generate_utils.bzl
@@ -205,16 +205,14 @@ def generate_config(repository_ctx):
     annotations = collect_crate_annotations(repository_ctx.attr.annotations, repository_ctx.name)
 
     # Load additive build files if any have been provided.
-    content = list()
     for data in annotations.values():
-        additive_build_file_content = data.pop("additive_build_file_content", None)
-        if additive_build_file_content:
-            content.append(additive_build_file_content)
-        additive_build_file = data.pop("additive_build_file", None)
-        if additive_build_file:
-            file_path = repository_ctx.path(Label(additive_build_file))
-            content.append(repository_ctx.read(file_path))
-        data.update({"additive_build_file_content": "\n".join(content) if content else None})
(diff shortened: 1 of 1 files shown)


crate_universe/private/generate_utils.bzl
```diff
@@ -205,16 +205,14 @@ def generate_config(repository_ctx):
...
-        additive_build_file = data.pop("additive_build_file", None)
-        if additive_build_file:
-            file_path = repository_ctx.path(Label(additive_build_file))
-            content.append(repository_ctx.read(file_path))
-        data.update({"additive_build_file_content": "\n".join(content) if content else None})
+        f = data.pop("additive_build_file", None)
+        content = "\n".join([
+            data.pop("additive_build_file_content", None) or "",
+            repository_ctx.read(Label(f)) if f else "",
+        ]).strip()
```
UebelAndre, line 213: Since this content is generally rendered at the bottom of a file, could we avoid the `strip` so we don't strip the final newline?

UebelAndre (changes requested):
One nit but otherwise looks good! Thanks!

crate_universe/private/generate_utils.bzl
uhthomas: Hmm, good point. The strip was there to prevent the content from being just `\n`, as well as keeping the source a bit cleaner. Would it make sense to explicitly add a newline in addition to stripping? Since there is no guarantee of a newline from `repository_ctx.read` anyway.

UebelAndre: I lean toward the more hands off approach where we just render what users told us. That way folks who care about newlines have the power to fix it

uhthomas: I agree, that makes sense. Would you mind taking another look?

UebelAndre (approved):
Looks good to me, Thank you!

PR #1282, opened 2022-04-20
Fix crate annotation anchor

uhthomas:
The docs currently link to https://bazelbuild.github.io/rules_rust/crate_universe.html#crateannotations, which isn't valid. The correct URL is https://bazelbuild.github.io/rules_rust/crate_universe.html#crateannotation.


UebelAndre (approved):
Thanks!

Issue #1262, opened 2022-04-14
crates_repository: cannot build https://github.com/rust-lang/regex


uhthomas:
rules_rust: 0.2.1

Given

```starlark
load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")

rules_rust_dependencies()

rust_register_toolchains()

load("@rules_rust//crate_universe:repositories.bzl", "crate_universe_dependencies")

crate_universe_dependencies()

load("@rules_rust//crate_universe:defs.bzl", "crates_repository", "crate")

crates_repository(
    name = "crate_index",
    lockfile = "//:Cargo.Bazel.lock",
    packages = {
        "regex": crate.spec(
            git = "https://github.com/rust-lang/regex",
            rev = "1.5.5",
        )
    },
)

load("@crate_index//:defs.bzl", "crate_repositories")

crate_repositories()
```

The crate will fail to build.

```sh
❯ bazel build @crate_index//:regex
(18:01:03) INFO: Current date is 2022-04-14
(18:01:05) INFO: Repository crate_index__regex-1.5.5 instantiated at:
  /Users/thomas/code/<redacted>/WORKSPACE:166:19: in <toplevel>
  /private/var/tmp/_bazel_thomas/796dd52d4d981101108548a9446ddca6/external/crate_index/defs.bzl:392:10: in crate_repositories
  /private/var/tmp/_bazel_thomas/796dd52d4d981101108548a9446ddca6/external/bazel_tools/tools/build_defs/repo/utils.bzl:233:18: in maybe
Repository rule new_git_repository defined at:
  /private/var/tmp/_bazel_thomas/796dd52d4d981101108548a9446ddca6/external/bazel_tools/tools/build_defs/repo/git.bzl:186:37: in <toplevel>
(18:01:05) ERROR: An error occurred during the fetch of repository 'crate_index__regex-1.5.5':
   Traceback (most recent call last):
        File "/private/var/tmp/_bazel_thomas/796dd52d4d981101108548a9446ddca6/external/bazel_tools/tools/build_defs/repo/git.bzl", line 174, column 30, in _new_git_repository_implementation
                update = _clone_or_update(ctx)
        File "/private/var/tmp/_bazel_thomas/796dd52d4d981101108548a9446ddca6/external/bazel_tools/tools/build_defs/repo/git.bzl", line 36, column 20, in _clone_or_update
                git_ = git_repo(ctx, directory)
[...]

PR #1283, opened 2022-04-20
crate_universe: Support fetching crates with git branch, tag or rev

uhthomas:
Fixes https://github.com/bazelbuild/rules_rust/issues/1262.

Fortunately the code to support branches and tags already exists, and so it's as simple as adding them to the json blob.

Not only that, but code already exists which enforces the exclusivity of `branch`, `tag` and `rev`.

```sh
error: failed to parse manifest at `/var/folders/ws/lyfqx2sn72v223fhz5kb77m40000gp/T/.tmpIq5bdq/Cargo.toml`

Caused by:
  dependency (<redacted>) specification is ambiguous. Only one of `branch`, `tag` or `rev` is allowed.
```

Would it make sense to enforce this even earlier on though? It might just be a waste of cycles.


UebelAndre (approved):
This seems good to me but I'm wondering if there are some tests that can be added to regression test this...

uhthomas:
I think some tests would be a good idea. Would you be able to offer some guidance on where it would be appropriate to add said tests?

UebelAndre:
> I think some tests would be a good idea. Would you be able to offer some guidance on where it would be appropriate to add said tests?

I think this is gonna require some refactoring, but for any macro that in some way gets written to a file, we should make a target that does explicitly that, and then have a unittest de-serialize it and check that certain values are properly set. This way when new fields are added to the Rust structs, the test will require people to make sure the macros are up to date too. Does that make sense?

UebelAndre:
@uhthomas I think adding onto https://github.com/bazelbuild/rules_rust/pull/1291 would be sufficient testing. Deserialize the struct and make sure the values you set are correctly represented in rust. Passed that I'd say maybe there should be some tests for [rendering](https://github.com/bazelbuild/rules_rust/blob/0.2.1/crate_universe/src/rendering.rs) to ensure `tag` and `branch` will actually get rendered. After my PR I think there should be enough prior art to make writing the tests easy. Does that sound reasonable to you? (don't want to hold the change hostage to testing so trying to make adding tests easy 😅)

ted-logan:
I've opened a pull request expanding the change here with test data to verify the serialization and deserialization of crate.spec(): https://github.com/bazelbuild/rules_rust/pull/1846

UebelAndre:
Closing this out as the functionality was merged in https://github.com/bazelbuild/rules_rust/pull/1846

Issue #871, opened 2021-07-31
Add chksums for tar.xz files


wt:
I would like to see some support for tar.xz files since they compress way better. The known file chksums currently only chksum the tar.gz files. This will require some refactoring I guess.

I think we have a couple options to get this in place.

1. Make the FILE_KEY_TO_SHA dict keys include the extensions and make the lookups use the proper keys that include the extensions.
2. Not support tar.xz files, which would be a bummer.

I would love to know if there are ways to plumb this through.

Thanks.

illicitonion:
This sounds like a good improvement, and like a reasonable approach. I think the steps I'd take to go about this are:
1. Add a new attribute to the `rust_toolchain_repository` `repository_rule`, called `type` to mirror `https://docs.bazel.build/versions/main/repo/http.html#http_archive-type` - for now I guess let's let it take values `tar.gz` or `tar.xz`, and default to `tar.gz`.
2. Instead of a single `DEFAULT_STATIC_RUST_URL_TEMPLATES`, derive that value based on the `type` attribute.
3. Assume that if someone is explicitly overriding their template, their value doesn't need any modifying - if they want to switch to `tar.xz`, they should modify their template, as well as setting the `type` attr.
4. Make `FILE_KEY_TO_SHA` either have a tuple as a key, or be a two-level dict, and modify `util/fetch_shas.sh` to generate the dict
5. Update all of the call sites accordingly

Feel free to reach out here or [on slack](https://bazelbuild.slack.com/archives/CSV56UT0F) if you need a hand!

PR #1284, opened 2022-04-21
Enable xz archives

wt:
XZ archives are about half the size of the gz equivalents. This
improves the cold repo bootstrap case a fair amount.

Fixes: #871


wt:
The static rust environment variable is interfering with this since the url it adds ends up on the top of the list. I have removed it. It's not clear to me that the functionality is super useful since the default location for urls matches it. If it's needed, I would love some suggestions for how to deal with it.

uhthomas:
Is it possible to use `tar.zst` instead?

Arch Linux switched from xz to zst because it decompresses significantly faster (~1300%), at the cost of slightly larger archives (~0.8%).

wt:
Are zstd archives supported by bazel? If so, the diff from here to there would be pretty small. There doesn't appear to be a programatic list of the archive types supported, so each much be added manually.

Assuming that is done, at the very least you can supply the sha256s manually to the register_rust_toolchains function.

Having said that, tar.zstd are not available from the rust distribution that I could find. Nothing would stop you from downloading an xz file and converting it to use zstd compression and distributing them from a web site.

uhthomas:
>Are zstd archives supported by bazel?

Yep! :) The list of supported archives can be found with the [http_archive](https://bazel.build/rules/lib/repo/http#http_archive) documentation.

>It supports the following file extensions: "zip", "jar", "war", "aar", "tar", "tar.gz", "tgz", "tar.xz", "txz", "tar.zst", "tzst", tar.bz2, "ar", or "deb".

>Having said that, tar.zstd are not available from the rust distribution that I could find. 

That is a shame, though I thought this might be the case. Thank you for considering my suggestion.

wt:
Find the files for download and it would be easy to add them to the fold. BTW, those archive formats are buried in many comments around bazel. It would be truly awesome if they were available programatically for extensions. If you wanted to create an issue to ask for that on the main bazel repo, feel free to tag me.

Also, if you can find the location of the .tar.zst files. The additional diff on top of this one would be pretty small.

illicitonion (approved):
Thanks!

UebelAndre:
Cross posting from https://github.com/bazelbuild/rules_rust/commit/841fc6fb82141576d91aecb1d3f2656d58b0ab71#commitcomment-72310891 to try and consolidate info on the PR.

cc @krasimirgg @wt

PR #1286, opened 2022-04-22
feat: support extra deps in rust_doc_test

roman-kashitsyn:
Currently, `rust_doc_test` (unlike `rust_test`) doesn't support adding
extra test-only dependencies, which makes it unnecessarily hard to
migrate doc tests from Cargo to rules_rust.

This change extends the `rust_doc_test` rust to support extra
test-only dependencies.

### DIFF
--- a/examples/hello_lib/BUILD.bazel
+++ b/examples/hello_lib/BUILD.bazel
@@ -28,13 +28,6 @@ rust_static_library(
     ],
 )
 
-# A helper library used only in doc tests.
-# We use it to check that the doc tests can use test-only dependencies.
-rust_library(
-    name = "frobnicate",
-    srcs = ["src/frobnicate.rs"],
-)
-
 # Regression test for #368: static lib with dependencies fail.
 rust_static_library(
     name = "hello_test_staticlib",
@@ -72,7 +65,6 @@ rust_doc(
 rust_doc_test(
     name = "hello_lib_doc_test",
     crate = ":hello_lib",
-    deps = [":frobnicate"],
 )
 
 rust_analyzer(
--- a/examples/hello_lib/BUILD.bazel
+++ b/examples/hello_lib/BUILD.bazel
@@ -28,6 +28,13 @@ rust_static_library(
     ],
 )
 
+# A helper library used only in doc tests.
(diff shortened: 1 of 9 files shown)


examples/hello_lib/BUILD.bazel
```diff
@@ -28,6 +28,13 @@ rust_static_library(
     ],
 )
 
+# A helper library used only in doc tests.
```
UebelAndre, line 31: There's actually a unit test section for rustdoc, can you add coverage there instead?
https://github.com/bazelbuild/rules_rust/tree/0.2.1/test/unit/rustdoc

rust/private/rustdoc_test.bzl
```diff
@@ -154,6 +171,14 @@ rust_doc_test = rule(
             providers = [rust_common.crate_info],
             mandatory = True,
         ),
+        "deps": attr.label_list(
+            doc = dedent("""\
+                List of other libraries to be linked to this library target.
+
+                These can be either other `rust_library` targets or `cc_library` targets if
+                linking a native library.
+            """),
```
UebelAndre, line 180: I'm surprised the other attributes don't have a provider requirement for deps, can you add that here though since I don't think `rust_doc_test` would need any more?
```suggestion
            """),
            providers = [CrateInfo, CcInfo],
```
I think that syntax is correct but https://bazel.build/rules/lib/attr#label would have more info.

UebelAndre (changes requested):

rust/private/rustdoc_test.bzl
roman-kashitsyn: Done.

examples/hello_lib/BUILD.bazel
roman-kashitsyn: Done.

UebelAndre (approved):
Looks good to me! Seems that buildifier needs to be run though.

roman-kashitsyn:
> Looks good to me! Seems that buildifier needs to be run though.

Thank you! I fixed the formatting. I'll configure my editor to run buildifier with the right flags.

PR #1289, opened 2022-04-24
Make get_host_triple public to get a triple from Bazel's repository_ctx

dtolnay:
I found half-baked duplicate implementations of this logic in cargo-raze ([*examples_repository.bzl*](https://github.com/google/cargo-raze/blob/8ef4e9291eefbb0424036492a0f7066742753b08/tools/examples_repository.bzl#L37-L42)), in cxx ([*vendor.bzl*](https://github.com/dtolnay/cxx/blob/95ff4822aa766613a6e47523a3ead86cd5f24bbc/tools/bazel/vendor.bzl#L24-L37)), and in rules_rust's test suite of the load_arbitrary_tool function ([*load_arbitrary_tool_test.bzl*](https://github.com/bazelbuild/rules_rust/blob/207ee4fbcf4bac4b4c1da24097b453bb7cfcd011/test/load_arbitrary_tool/load_arbitrary_tool_test.bzl#L5-L13)). It seems such logic is necessary for correctly using load_arbitrary_tool from the implementation of a repository_rule.

# Documentation

### Example:

```python
load("@rules_rust//rust:repositories.bzl", "load_arbitrary_tool")
load("@rules_rust//rust/platform:triple.bzl", "get_host_triple")
 
def _impl(repository_ctx):
    load_arbitrary_tool(
        ctx = repository_ctx,
        tool_name = "cargo",
        tool_subdirectories = ["cargo"],
        target_triple = get_host_triple(repository_ctx).str,
    )
 
example = repository_rule(implementation = _impl)
```
 
### Args:

- `repository_ctx` (repository_ctx): The repository_rule's context object
- `abi` (str): Since there's no consistent way to check for ABI, this info may be explicitly provided

### Returns:

- `struct`: A triple struct; see the `triple` function in this module


hlopko (approved):

UebelAndre (changes requested):

rust/private/repository_utils.bzl
```diff
@@ -512,6 +512,45 @@ def produce_tool_path(tool_name, target_triple, version):
         fail("No tool version was provided")
     return "-".join([e for e in [tool_name, version, target_triple] if e])
 
+def target_triple_for_os(repository_os):
```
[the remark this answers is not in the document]
dtolnay: Ah, good find. I hadn't come across that. This logic isn't Cargo-specific at all so I didn't think to look in the `cargo` dir.

I've updated the PR to consolidate both that and the one I originally added into a `get_host_triple` in rust/platform/triple.bzl.

test/load_arbitrary_tool/load_arbitrary_tool_test.bzl
```diff
@@ -19,7 +14,7 @@ def _load_arbitrary_tool_test_impl(repository_ctx):
         tool_subdirectories = ["cargo"],
         version = "1.53.0",
         iso_date = None,
-        target_triple = target_triple,
+        target_triple = host_triple.str,
```
dtolnay, line 17: I added a `str` field to the struct returned by `triple("...")`, because `triple.triple` didn't strike me as evocative enough. `str` makes sense to me as the string representation of the triple, which is what `target_triple` wants.

Alternatively we could look into making `target_triple` accept the triple struct directly, or union of triple struct and string so that either way works.

UebelAndre: > I added a str field to the struct returned by triple("..."), because triple.triple didn't strike me as evocative enough. str makes sense to me as the string representation of the triple, which is what target_triple wants.

I'm fine with the use of `str` but I'd prefer there to be either `str` or `triple`. Could you maybe update the docstring to indicate which of the two is the correct to use?

> I added a `str` field to the struct returned by `triple("...")`, because `triple.triple` didn't strike me as evocative enough. `str` makes sense to me as the string representation of the triple, which is what `target_triple` wants.
> 
> Alternatively we could look into making `target_triple` accept the triple struct directly, or union of triple struct and string so that either way works.
> 
> Or we could make `get_host_triple` return the string and not the struct, forcing `triple(get_host_triple(repository_ctx))` when someone needs the struct.

I think I would prefer to continue to pass the `triple` struct until we find a performance reason to do otherwise. We have a lot of places where we parse the same string over and over. I expect passing a struct to be slightly better.

UebelAndre (approved):
Looks good to me! Thanks!

If you could update one comment and rebase we can get this in 😄

dtolnay:
Rebased and updated the documentation of the return type of `triple("...")`:

```diff
         - str (str): Original string representation of the triple
-        - triple (str): Same as str
+        - triple (str): Deprecated; same as str
```

I'll open another PR after this to remove `triple` from this struct so that we can discuss that independently of this change.

PR #1297, opened 2022-04-26
Remove `triple` field from triple struct in favor of `str`

dtolnay:
If I have:

```python
triple = get_host_triple(repository_ctx)

f(triple.triple)
```

then `triple.triple` does not do a good job conveying what is going on; how the two things called `triple` (actually 3 if you count the function named `triple` as well) are different from one another.

To improve this in #1289 I added a `str` field to the triple struct, just duplicating the contents of the `triple` field but with a clearer name. In contrast to `triple.triple`, writing `f(triple.str)` is clearer to me and conveys "oh yeah `f` takes a string argument (as opposed to a struct) and `triple.str` is the string representation of the triple".

This PR removes the field `triple.triple` in favor of `triple.str` since there is no longer a reason to use the former.

### DIFF
--- a/cargo/cargo_bootstrap.bzl
+++ b/cargo/cargo_bootstrap.bzl
@@ -204,7 +204,7 @@ def _cargo_bootstrap_repository_impl(repository_ctx):
 
     # In addition to platform specific environment variables, a common set (indicated by `*`) will always
     # be gathered.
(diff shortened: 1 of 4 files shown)


UebelAndre (approved):
Thanks for the follow-up! Looks good to me 😄

PR #1298, opened 2022-04-26
Don't emit `CrateInfo` from `rust_static_library` and `rust_shared_library`

scentini:
These rules emit a `staticlib` and `cdylib` which are not meant for consumption by other rust targets.

However, one should still be able to write a `rust_test` for `rust_shared_library` and `rust_static_library` targets, which needs a `CrateInfo`. For this purpose we provide a `CrateInfo` wrapped in a `TestCrateInfo` provider, that `rust_test` understands.


hlopko (approved):

rust/private/rustc.bzl
```diff
@@ -986,15 +986,23 @@ def rustc_compile_action(
...
-        dep_info,
         DefaultInfo(
             # nb. This field is required for cc_library to depend on our output.
             files = depset(outputs),
             runfiles = runfiles,
             executable = crate_info.output if crate_info.type == "bin" or crate_info.is_test or out_binary else None,
         ),
     ]
+
+    if crate_info.type in ["staticlib", "cdylib"]:
```
PiotrSikora, line 997: This confuses rule type with `crate_info.type`, and breaks `rust_binary(crate_type="cdylib")` which we use when building WebAssembly targets due to [a missing support for `rust_wasm_binary()`](https://github.com/bazelbuild/rules_rust/issues/771):

```
[...]
ERROR: /tmp/proxy-wasm-cpp-host/test/test_data/BUILD:45:17: in rust_binary rule //test/test_data:_wasm_env_wasm: 
.../external/rules_rust/rust/private/rust.bzl:295:5: rule advertised the 'CrateInfo' provider, but this provider was not among those returned
ERROR: /tmp/proxy-wasm-cpp-host/test/test_data/BUILD:45:17: Analysis of target '//test/test_data:_wasm_env_wasm' failed
ERROR: Analysis of target '//test/test_data:env.wasm' failed; build aborted: 
[...]
```

UebelAndre: @PiotrSikora Sorry about the breakage :( 

can you create a separate issue for this? It'd then be good if you could provide a local repro case or maybe just open a PR if you've already got a solution in mind.

PiotrSikora: I have a fix (https://github.com/bazelbuild/rules_rust/pull/1315), but local repro is tricky due to a use of custom rule and transition to force build for Wasm.

PR #1299, opened 2022-04-26
Don't propagate non-shared transitive linker inputs from `rust_static|shared_library`

scentini:
A Rust `staticlib` or `cdylib` doesn't need to propagate linker inputs of its dependencies, except for shared libraries.

- The tests are currently disabled on osx because the CI [Toolchain does not support dynamic linking]( https://buildkite.com/bazel/rules-rust-rustlang/builds/6126#a83dbb56-50b0-4a95-bb39-09e3a78ed0d0).
- I also had to link to `Bcrypt.lib` => https://github.com/rust-lang/rust/issues/91974


scentini:
cc @krasimirgg

PR #1300, opened 2022-04-26
rustc: fix a conditional

gburgessiv:
`("bar" or "baz")` always evaluates to "bar", so this conditional ignores `"dylib"`. This seems unintended.

I don't have a particular test-case in mind for this; just noticed the oddity as I was browsing through the code.

cc @krasimirgg, who seems to have tweaked things around here recentlyish?

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -119,7 +119,7 @@ def _are_linkstamps_supported(feature_configuration, has_grep_includes):
             has_grep_includes)
 
 def _should_use_pic(cc_toolchain, feature_configuration, crate_type):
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
This looks good to me but would now be a change in behavior. @hlopko @scentini I'm wondering what your thoughts are here too.

scentini:
IIUC we currently don't have a use case for `dylib`, so this fix comes at the right time. Thanks!

scentini:
IIUC we currently don't have a use case for `dylib`, so this fix comes at the right time. Thanks!

scentini (approved):

gburgessiv:
(Thank you all! :) )

PR #1304, opened 2022-04-29
Add support for aarch64-apple-ios-sim

keith:
This adds / improves support for for the arm64 iOS simulator that's used
on M1 macs. Previously there was ambiguity in the constraints used for
this platform because bazel does not pick the best match, so this adds
the device constraint in this case to disambiguate. Ideally we could
rely on not having the `sim` suffix to be enough to add the device
constraint, but older triples like `x86_64-apple-ios` break this
pattern. I broke this logic out into a new function since it relies on
more info from the triple than just the ABI now.

https://github.com/bazelbuild/bazel/issues/11454


keith:
What does your platform_mappings configuration look like? It might be an ordering problem that wouldn't be an issue if you define them in a way that the simulator one is first?

dae:
I don't have one - I use a wrapper script that invokes bazel with the appropriate --platforms argument based on things like ARCH and PLATFORM_NAME that xcode has set.

keith:
Ah I bet you're right with your original comment:

> I list the sim target above the non-sim target in Rust's extra_target_triples

that this is the difference, can you try flipping them?

dae:
I wouldn't expect the flipped approach to work, since my understanding is that Bazel will use the first registered toolchain that satisfies the active constraints.

keith:
Yea makes sense. I think doing something like this PR might be preferred to that since it feels a bit fragile to rely on that ordering

dae:
The fact that the :device target already exists in apple_support seems to imply you're not the only one to feel that way.

Just out of curiosity, what were the changes on https://github.com/google/cargo-raze/pull/484 needed for? Did you encounter crates in the wild that were specifically targeting the -sim variant?

keith:
Without that it seems that some of the conditionals in our dependencies didn't correctly `select()` for this version.

snowp:
The raze issue is that raze will expand cfg(unix) to all unix families listed in the Cargo.toml configuration, but only for allowed target triples. This would typically manifest as crates not depending on the libc crate when targeting the sim, despite the sim being unix-like

UebelAndre:
Feel free to ping me when this is ready for review.

rust/platform/triple_mappings.bzl
```diff
@@ -276,5 +293,6 @@ def triple_to_constraint_set(target_triple):
     constraint_set += vendor_to_constraints(triple_struct.vendor)
     constraint_set += system_to_constraints(triple_struct.system)
     constraint_set += abi_to_constraints(triple_struct.abi)
+    constraint_set += extra_ios_constriants(triple_struct)
```
UebelAndre, line 296: Do you know if there's a downside to having duplicate constraints? The only thing that bothers me about the change is that it seems there's overlapping checks here that may one day result in an invalid configuration. Should `ios` hosts separately evaluate constraints? I.e.:
```python
if triple_struct.system == "ios":
    constraint_set = ios_constraints(triple_struct)
else:
    # what's already here.
```

UebelAndre (commented):
Thanks! I think this looks good but have one question that I could be talked down from.

rust/platform/triple_mappings.bzl
keith: I don't think there's any downside for duplications, but I haven't verified. But in this current setup there shouldn't be any duplicates since this new function doesn't add the CPU / system ones, and no other functions add the simulator / device ones (anymore)

UebelAndre (approved):
Looks good to me, thanks!

keith:
Thanks all!

Issue #1307, opened 2022-04-29
Sha256s in the fake_toolchain_for_test_of_sha256 example test repo don't matter


wt:
The sha256s are additive. Since the version of rust setup in that example repo is real, the real shas will be used if the shas in the call to rust_repository_set at [here](https://github.com/bazelbuild/rules_rust/blob/main/examples/WORKSPACE.bazel#L48-L60) don't contain the keys that will actually be used. Now that the keys include the file extensions (e.g. `rust-1.46.0-x86_64-unknown-linux-gnu.tar.gz` instead of `rust-1.46.0-x86_64-unknown-linux-gnu`), the shas defined in the repo aren't even used and can be removed. It might be better to have a fake rust, rust-std, and rustfmt .tar.gzs and tar.xzs with shas that don't exist in the known shas list to make sure this is really working.

For now, I think we should remove the shas from the call as they don't really matter. You can actually change them to almost anything (including nothing), and you will still be able to run `bazel query @fake_toolchain_for_test_of_sha256//...` from the examples directory.

If you want to force a failure you can add tar.gz to the end of any of the keys and botch the sha for the key. That will cause an incorrect sha to be resolved, and will result in the download failing to match the sha for that file.

PR #1308, opened 2022-04-29
Define default/gz/xz fake tool repos explicitly for testing.

wt:
Related to #1307


UebelAndre (changes requested):
So sorry for the delay here!

I may just be totally out of it right now but can you show some error this would catch? Do these explicitly test `sha256` values? My understanding is that sha256 values are totally optional and to boot the rules won't report when one isn't used (like you'd see with `http_archive`). Is that correct?

PR #1309, opened 2022-04-29
Use tinyjson from crates.io instead of github.com.

### LINKED ISSUE #1273: Can we please depend on the crates.io version of tinyjson instead of the version in github?
I wondering if we should bias toward depending on crates from crates.io instead of from a github repo release page.

34fd46756ab0a209d019d0475178bccf1738c4a1 introduced a new dep on tinyjson that pulls from a github release zip instead of pulling from a release at crates.io. This seems less than ideal.

The url of the crates.io release is at `https://crates.io/api/v1/crates/tinyjson/2.3.0/download` with a sha256 of `1a8304da9f9370f6a6f9020b7903b044aa9ce3470f300a1fba5bc77c78145a16`

As a bonus, that release archive is smaller than the one in the github release.

wt:
We should be pulling in deps from crates.io as much as possible
since the downloads will usually be smaller, and we can avoid
a proliferation of remote locations for resources.

Fixes #1273

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -63,9 +63,10 @@ def rules_rust_dependencies():
     maybe(
         http_archive,
         name = "rules_rust_tinyjson",
-        sha256 = "9c21866c7f051ebcefd028996494a374b7408ef946826cefc9761d58cce0fd36",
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thank you!

krasimirgg:
Looks reasonable!cc @gigaroby

PR #1310, opened 2022-04-30
Move the rust source to its own repo.

wt:
This repo will only be fetched when it is referenced, so we no
longer need to explicitly handle the download of the repo.

This is a subset of the change here: #1163

I have not yet been able to figure out what is wrong with that one, so I extracted out just the repo for the srcs. I think if we do these in smaller chunks it might be easier to handle.

Using an repo like this allows bazel to handle when then repo needs to be downloaded.

### DIFF
--- a/rust/private/repository_utils.bzl
+++ b/rust/private/repository_utils.bzl
@@ -1,5 +1,7 @@
 """Utility macros for use in rules_rust repository rules"""
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load("//rust:known_shas.bzl", "FILE_KEY_TO_SHA")
 load(
     "//rust/platform:triple_mappings.bzl",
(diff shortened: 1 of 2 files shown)


wt:
It might also be worth noting that this fixes an error where a maybe was calling a macro instead of a repo rule as well.

wt:
It's weird that buildifier is complaining about an undocumented arg that is documented. Is that a bug, or am I not seeing something?

rust/private/repository_utils.bzl
```diff
@@ -396,13 +405,17 @@ def load_rust_stdlib(ctx, target_triple):
     if "BAZEL_RUST_STDLIB_LINKFLAGS" in ctx.os.environ:
         stdlib_linkflags = ctx.os.environ["BAZEL_RUST_STDLIB_LINKFLAGS"].split(":")
 
+    rustc_srcs = None
+    if should_include_rustc_srcs(ctx):
+        rustc_srcs = "\"@rust-src_{}//:rustc_srcs\"".format(ctx.attr.version)
```
illicitonion, line 410: Hard-coding the name here seems a little strange - would it make more sense to have a new attr for the name of the `rust-src` repository (or full target) which may be `None`, and set `rustc_srcs` according to it, and having `rust_repository_set` set that attribute as it expects, rather than hard-coding what a wrapper happens to set the name to?

illicitonion (commented):
Generally looks reasonable - thanks!

UebelAndre:
Sorry for the big delay here. This work was done in https://github.com/bazelbuild/rules_rust/pull/1455 and is available in [release 0.8.1](https://github.com/bazelbuild/rules_rust/releases/tag/0.8.1). Thanks for providing the inspiration!

PR #1315, opened 2022-05-03
Emit CrateInfo when building Wasm libraries as executables.

### LINKED ISSUE #1330: rust_binary with crate_type = cdylib doesn't produce expected providers
We have a rule like this:

```bzl
    rust_binary(
        name = ...,
        edition = "2018",
        crate_type = "cdylib",
        out_binary = True,
...
    )
```

which since https://github.com/bazelbuild/rules_rust/commit/5abeb93333db6ca1c44dbb55860987c408553a11 (in release 0.4.0) fails with:

```
ERROR: /Users/ksmiley/dev/envoy/test/extensions/common/wasm/test_data/BUILD:12:17: in rust_binary rule //test/extensions/common/wasm/test_data:_wasm_test_rust_wasm:
/private/var/tmp/_bazel_ksmiley/4096f01a2eae006dd7eb0d708a9f3935/external/rules_rust/rust/private/rust.bzl:295:5: rule advertised the 'CrateInfo' provider, but this provider was not among those returned
ERROR: /Users/ksmiley/dev/envoy/test/extensions/common/wasm/test_data/BUILD:12:17: Analysis of target '//test/extensions/common/wasm/test_data:_wasm_test_rust_wasm' failed
ERROR: Analysis of target '//test/extensions/common/wasm:wasm_vm_test' failed; build aborted:
```

It looks like the issue is that `rust_binary` still states that it will create `CrateInfo` yet it does not in this case. Our use case doesn't seem to care about this, so just removing that requirement is enough. This is reproducible on this branch https://github.com/envoyproxy/envoy/pull/21208 with `bazel build --nobuild //test/extensions/common/wasm:wasm_vm_test --define=wasm=wamr`.

PiotrSikora:
Broken in #1298.

Fixes #1330.


PiotrSikora:
cc @UebelAndre @scentini

rust/private/rustc.bzl
```diff
@@ -994,7 +994,7 @@ def rustc_compile_action(
         ),
     ]
 
-    if crate_info.type in ["staticlib", "cdylib"]:
+    if crate_info.type in ["staticlib", "cdylib"] and not out_binary:
```
UebelAndre, line 997: I'm really not a fan of this attribute. Isn't https://github.com/bazelbuild/rules_rust/issues/771 something that would solve for this? Perhaps I'm still missing some context but the attribute feels like a toggle for some super niche functionality that alone adds unaccounted edge cases but is not currently an issue because there's a small group of users who use this and all do the same thing. I'd love to either have my understanding of this problem space be updated or delete the thing that seems like a hole we punched in the rules to support a few things.

UebelAndre (changes requested):

rust/private/rustc.bzl
PiotrSikora: AFAIK, the only use case for `out_binary` is building Wasm modules, so #771 would definitely fix it, but I don't have bandwidth to implement `rust_wasm_{binary,library}` myself, so a quick fix it is.

scentini: I just learned about this attribute and I too am unhappy about it. @UebelAndre do we have a maintainer who supports `wasm` builds? While I don't want to block this PR on it, I think we should either have maintainers/contributors that can pick up https://github.com/bazelbuild/rules_rust/issues/771 and clean up the codebase or discuss whether this is something that we want to support.

scentini (commented):
Could you please add tests for this?

rust/private/rustc.bzl
UebelAndre: To my knowledge there isn't a maintainer who supports `wasm` builds. Unless we can find a consumer of this functionality that's willing to cleanly implement the rules and add test then I don't think we can officially say we support this. I'd be reluctantly ok with letting this one through but can't say I'm currently capable of understanding whether or not future changes will cause regressions.

PiotrSikora: Could we get this in? It looks that v0.4.0 was shipped without the fix.

I agree that we should have `rust_wasm_{binary,library}` (and I advocated for this on the PR that added `out_binary`), at which point we could remove `out_binary`, but since it's been added and worked for the past few years, breaking it when the fix is so trivial doesn't seem very productive.

scentini: As I said, I'm willing to approve this, but I am still waiting for a test.

PiotrSikora: Sorry, I've missed that comment.

I've spend some time trying add tests before opening this PR and again today, but I have trouble getting it to work, since even something as simple as:
```
rust_binary(
        name = "wasm_intermediate",
        srcs = ["lib.rs"],
        crate_type = "cdylib",
        out_binary = True,
        edition = "2018",
)
```
fails to build in tests, because Bazel complains about missing C++ toolchain, and I cannot figure out why.

Any ideas?

However, I verified that this PR fixes the the build in Envoy, which is an existing consumer of the broken feature.

scentini: Could you provide the exact error? Does it happen when you create other targets, eg `rust_shared_library`?

PiotrSikora:
```
ERROR: .../rules_rust/test/test_env/BUILD.bazel:4:12: While resolving toolchains for target //test/test_env:hello-world: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --
incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
```

It happens only for `rust_binary(crate_type = "cdylib", out_binary = True)`.

PiotrSikora: @UebelAndre @scentini friendly ping, either for this or #1339, since it blocks Rust updates in Envoy and Proxy-Wasm.

scentini: Hey @PiotrSikora, I don't believe I'll have time in the near future to explore how rust+wasm builds work. Your best bet would be to either explore how the envoy rules work so that you can provide testing for this use case, or find someone willing to do so. In the meantime I consider this use case unsupported.

Re #1339 - did you confirm that it fixes your issue? That one I can work on soon.

PR #1320, opened 2022-05-05
Prevents running of clippy on bindgen targets

### LINKED ISSUE #1048: Clippy aspect prevents customization of lints for bindgen targets
The Clippy aspect attempts to prevent some errors in bindgen targets by limiting the lints that are denied. See:
https://github.com/bazelbuild/rules_rust/blob/dc66c1612b7a3e96531eff22136570124c8eec81/rust/private/clippy.bzl#L129

However, because these clippy flags are being appended to the rustc flags, clippy lint customizations in the rustc flags (e.g. ["-A," "clippy::assertions_on_constants"]) will be overridden.

csmulhern:
Fixes #1048.

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -145,14 +145,6 @@ See https://github.com/bazelbuild/rules_rust/pull/1264#discussion_r853241339 for
 
         if clippy_flags:
             args.rustc_flags.extend(clippy_flags)
-
-        elif "__bindgen" in ctx.rule.attr.tags:
-            # bindgen-generated content is likely to trigger warnings, so
-            # only fail on clippy warnings
-            args.rustc_flags.add("-Dclippy::style")
-            args.rustc_flags.add("-Dclippy::correctness")
-            args.rustc_flags.add("-Dclippy::complexity")
-            args.rustc_flags.add("-Dclippy::perf")
         else:
             # The user didn't provide any clippy flags explicitly so we apply conservative defaults.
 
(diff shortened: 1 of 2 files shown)


rust/private/clippy.bzl
```diff
@@ -72,7 +72,7 @@ def _get_clippy_ready_crate_info(target, aspect_ctx):
 
 def _clippy_aspect_impl(target, ctx):
     crate_info = _get_clippy_ready_crate_info(target, ctx)
-    if not crate_info:
+    if not crate_info or "__bindgen" in ctx.rule.attr.tags:
```
UebelAndre, line 75: instead of checking this here, could we instead add `noclippy` tags to the `bindgen` targets? This way the clippy rules don't need to be aware of bindgen specifically.

UebelAndre (changes requested):

UebelAndre:
@scentini I forget, but do you use the `bindgen` rules?

rust/private/clippy.bzl
csmulhern: Done.

UebelAndre (approved):
Looks good to me! I'll wait till @scentini to be able to take a look too.

One final thought is it might also make sense to prevent clippy from running on generated files like [rustfmt](https://github.com/bazelbuild/rules_rust/blob/main/rust/private/rustfmt.bzl#L29-L30) does.

scentini (approved):
Looks good, thanks!

PR #1325, opened 2022-05-08
Explicitly set `LD_LIBRARY_PATH` in `Rustc` actions.

UebelAndre:
When investigating ways to reduce the time needed to download toolchains (https://github.com/bazelbuild/rules_rust/pull/1163) I found in a couple of cases where `rustc` was either not using the correct libraries (occurs when using `cc_toolchain`s downloaded by Bazel) or not able to find the ones it needs (occurs when the `rust_toolchain` is represented by multiple external repositories). This change explicitly sets `LD_LIBRARY_PATH` to guide `rustc` to using paths in registered toolchains for finding the necessary files.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -584,6 +584,41 @@ def collect_inputs(
 
     return compile_inputs, out_dir, build_env_files, build_flags_files, linkstamp_outs, ambiguous_libs
 
+def _collect_ld_library_path(env, rust_toolchain, cc_toolchain):
+    """Construct the value of `LD_LIBRARY_PATH` based on the existing environment and registered toolchains.
+
+    Args:
+        env (dict): A dict of existing environment variables
+        rust_toolchain (rust_toolchain): The registered rust_toolchain
+        cc_toolchain (CcToolchainInfo): A cc toolchain provider to be used.
+
+    Returns:
+        str: The value of `LD_LIBRARY_PATH`.
+    """
+    sep = ";"
+    ld_paths = []
+
+    # Maintain existing values
+    if "LD_LIBRARY_PATH" in env:
+        ld_paths.extend(env["LD_LIBRARY_PATH"].split(sep))
+    if "DYLD_LIBRARY_PATH" in env:
+        ld_paths.extend(env["DYLD_LIBRARY_PATH"].split(sep))
+
+    # Add the rustc sysroot
+    ld_paths.append("{}/lib".format(rust_toolchain.sysroot))
+
+    # When possible add the cc sysroot
+    if cc_toolchain.sysroot:
+        ld_paths.extend([
+            "{}/{}".format(cc_toolchain.sysroot, path)
+            for path in [
+                "lib64",
+                "lib",
+            ]
+        ])
+
+    return sep.join(depset(ld_paths).to_list())
+
 def construct_arguments(
         ctx,
         attr,
@@ -809,6 +844,15 @@ def construct_arguments(
 
     # Ensure the sysroot is set for the target platform
     env["SYSROOT"] = toolchain.sysroot
+
+    # Explicitly set the `LD_LIBRARY_PATH` variable to make `rustc` more hermetic.
+    ld_library_path = _collect_ld_library_path(
+        env = env,
+        rust_toolchain = toolchain,
+        cc_toolchain = cc_toolchain,
+    )
+    env["LD_LIBRARY_PATH"] = ld_library_path
+    env["DYLD_LIBRARY_PATH"] = ld_library_path
 
     if toolchain._rename_first_party_crates:
         env["RULES_RUST_THIRD_PARTY_DIR"] = toolchain._third_party_dir


hlopko:
Could you explain the concrete cases where rustc used incorrect libraries (which C++ toolchain, which library was misused, which library should have been used instead) and which libraries were not found when rust_toolchain was represented in multiple repositories?

Using `LD_LIBRARY_PATH` in Bazel is usually not needed and often leads to nonhermetic builds, so I'd like to understand the scenarios involved. Ideally we should be able to tell the linker where to look for libraries using command line flags.

hlopko:
> I have a collection of custom C++ toolchains I use for one of my projects. These are designed to allow me to do cross compilation to a couple of platforms. Before these changes, I was getting build errors when I tried to build for musl platforms and found this was caused by rustc loading libraries (like libc.so.6) from /lib/x86-64-linux-gnu instead of my toolchain.

That makes me think that the fix could be on the C++ toolchain side. If you have a custom toolchain you can tell it to set `LD_LIBRARY_PATH`. We fetch env variables from the C++ toolchain and set it for Rustc here: https://cs.opensource.google/bazel/rules_rust/+/main:rust/private/rustc.bzl;l=309?q=rustc.bzl.

If I were to fix it, I'd probably not use `LD_LIBRARY_PATH`, but instead check if all `-L/-rpath` flags in the C++ toolchain are correct. I suspect there are some flags missing. Which linker are you using?

> The use of LD_LIBRARY_PATH here is intended to only point to paths provided by currently registered toolchains, which I think is not going to be an issue for hermetic builds. My expectation is that the use of LD_LIBRARY_PATH on the host (set to something in your environment before you call bazel) will cause problems.

Then IMHO it's the job of the toolchain to prepare the correct environment variables, not rustc.bzl. I'll comment on the code to be more precise in a second.

rust/private/rustc.bzl
```diff
@@ -584,6 +584,41 @@ def collect_inputs(
...
+
+    Returns:
+        str: The value of `LD_LIBRARY_PATH`.
+    """
+    sep = ";"
+    ld_paths = []
+
+    # Maintain existing values
+    if "LD_LIBRARY_PATH" in env:
+        ld_paths.extend(env["LD_LIBRARY_PATH"].split(sep))
```
hlopko, line 603: In addition to my impression that this env var (or equivalent linker flags) should be set by the C++ toolchain, maybe you could use https://docs.bazel.build/versions/main/command-line-reference.html#flag--action_env on the Bazel command line to achieve the same effect?

rust/private/rustc.bzl
```diff
@@ -584,6 +584,41 @@ def collect_inputs(
...
+    if "DYLD_LIBRARY_PATH" in env:
```
hlopko, line 604: This is mac-only, right? I guess we shouldn't set the variable on Windows or Linux.

rust/private/rustc.bzl
```diff
@@ -584,6 +584,41 @@ def collect_inputs(
...
+        ld_paths.extend(env["DYLD_LIBRARY_PATH"].split(sep))
+
+    # Add the rustc sysroot
+    ld_paths.append("{}/lib".format(rust_toolchain.sysroot))
```
hlopko, line 608: I'd have guessed that rustc does this by default when it is constructing the linker invocation internally. Is it an option to copypaste linker invocation flags before and after this PR? Which sysroot library/libraries are not loaded properly without this line?

rust/private/rustc.bzl
```diff
@@ -584,6 +584,41 @@ def collect_inputs(
...
+
+    # When possible add the cc sysroot
+    if cc_toolchain.sysroot:
```
hlopko, line 611: This part seems like is a part of the C++ toolchain configuration. I know that sysroot handling in the C++ toolchain is not great, do you use https://docs.bazel.build/versions/main/command-line-reference.html#flag--grte_top? If not, just hardcode all the flags in the CcToolchainConfigInfo and ignore there is a builtin support for sysroot in Bazel :)

rust/private/rustc.bzl
UebelAndre: > Then IMHO it's the job of the toolchain to prepare the correct environment variables, not rustc.bzl. I'll comment on the code to be more precise in a second.

I think I agree. Would it make sense to allow `rust_toolchain` to set environment variables? Or how could I have the toolchain define this information?

UebelAndre:
> That makes me think that the fix could be on the C++ toolchain side. If you have a custom toolchain you can tell it to set `LD_LIBRARY_PATH`. We fetch env variables from the C++ toolchain and set it for Rustc here: https://cs.opensource.google/bazel/rules_rust/+/main:rust/private/rustc.bzl;l=309?q=rustc.bzl.
> 
> If I were to fix it, I'd probably not use `LD_LIBRARY_PATH`, but instead check if all `-L/-rpath` flags in the C++ toolchain are correct. I suspect there are some flags missing. Which linker are you using?

Ah cool that sounds like a better approach. I can remove the `cc_toolchain` stuff from this PR.

> Then IMHO it's the job of the toolchain to prepare the correct environment variables, not rustc.bzl. I'll comment on the code to be more precise in a second.

I think I agree. Would it make sense to allow `rust_toolchain` to set environment variables? Or how could I have the toolchain define this information?

rust/private/rustc.bzl
UebelAndre: `rustc` (and there rest of the tooling) does so relative to where the binaries are. The issue is if your `rust_toolchain` is not backed by a single repository that contains all files in the same directories, then some things go missing. My aim for https://github.com/bazelbuild/rules_rust/pull/1119 was to allow `rust_toolchain` to be built form arbitrary targets and still work. In https://github.com/bazelbuild/rules_rust/pull/1163 I found that unfortunately `rustc` isn't able to find libraries in toolchains built like this.

Can you expand on 
>  Is it an option to copypaste linker invocation flags before and after this PR

Is this Bazel output (subcommand?) or something else?

rust/private/rustc.bzl
UebelAndre: Yeah, unfortunately we don't really have a consistent way to detect the exec environment since toolchains are not required to provide a triple.

rust/private/rustc.bzl
UebelAndre: I feel strongly that `--action_env` should not be used for `LD_LIBRARY_PATH`. This indicates a non hermetic aspect to your builds. That said though, this line attempts to handle cases where users set `--action_env`.

UebelAndre:
friendly ping @hlopko 😅

rust/private/rustc.bzl
hlopko: Ok I think I understand, I hope. What I'm hearing is that if the user uses `--action_env`, we want to append to that, not replace it. So this thread is blocked by the resolution of the threads below.

rust/private/rustc.bzl
hlopko: And actually this is setting the variable only when it's already present. I think that it's OK in that case, sorry for not realizing this earlier.

rust/private/rustc.bzl
hlopko: > My aim for #1119 was to allow rust_toolchain to be built form arbitrary targets and still work. In #1163 I found that unfortunately rustc isn't able to find libraries in toolchains built like this.

Do you know what exactly is the issue? It sounds like a reasonable expectation that a Rust sysroot that contains symlinks to random places should still work ? Do you have a reproduction case? Of course on Windows is hard, symlinks are complicated and paths can only bed 260 chars short and files can't be opened by multiple processes. I'd like to understand what exactly was the problem before accepting a workaround. It'd be doubly useful to know because internally we also create a sysroot full of symlinks, and I don't recall having any issues (CC @krasimirgg).

> Is this Bazel output (subcommand?) or something else?
Repro workplace would be the best, obviously :) But linker invocatoin (flags that rustc passes to the linker) or even the Bazel output would be useful.

rust/private/rustc.bzl
hlopko: Was this comment understandable? Does it make sense?

UebelAndre: (sorry, missed this. 😅)

I do not use `--grte_top` but based on your earlier comments about toolchains defining sufficient flags and variables for building, I would say this should be removed and will do so.

rust/private/rustc.bzl
UebelAndre: I think https://github.com/bazelbuild/rules_rust/pull/1163 is the repro case you're asking for. I had split `rust_repository_set` into simple `http_archive` repositories for each Rust asset (`rustc`, `llvm-tools`, `rust-std`) which creates the scenario where the paths of things like `rust-lld` do not have certain files physically next to it. When the `Rustc` action then gets to linking, it can't find the necessary files.

UebelAndre:
Another ping @hlopko, I think the direction this change is going is `rust_toolchain` should support setting environment variables?

hlopko:
I'm sorry for the delay, I still failed to find time to try to reproduce and wrap my head around rustc sysroot issue. Maybe @krasimirgg will be faster :)

hlopko:
>  support setting environment variables?

> Another ping @hlopko, I think the direction this change is going is `rust_toolchain` should support setting environment variables?

What exactly do you mean? An attribute with env vars that rustc.bzl then reads from the Rust toolchain info and passes them to each actions? That doesn't sound controversial. Do you also want to "parse" `LD_LIBRARY_PATH` and add things to it automatically?

UebelAndre:
> What exactly do you mean? An attribute with env vars that rustc.bzl then reads from the Rust toolchain info and passes them to each actions? That doesn't sound controversial. Do you also want to "parse" `LD_LIBRARY_PATH` and add things to it automatically?

Yeah, that's exactly what I mean. And yeah, I also want the ability to have the toolchain automatically set the generated sysroot path on a `LD_LIBRARY_PATH` environment variable it would have in it's `env` attribute.

UebelAndre:
@hlopko @krasimirgg I'm trying to figure out how I can update `rust_toolchain` to conditionally set `LD_LIBRARY_PATH` to the rust sysroot based on some flag but I struggle to come up with something that does not require a change to the `Rustc` action. In order to get `link_env` (https://github.com/bazelbuild/rules_rust/blob/0.5.0/rust/private/rustc.bzl#L311-L315) don't I need a ton of context from a target that's about to be built? The thought there is that if the `cc_toolchain` sets these variables too, that `rust_toolchain` would know to either append or prepend it's own flags. Any thoughts there? Perhaps that's unnecessary simply having a boolean toggle is sufficient for setting it or not?

hlopko:
(I apologize for not replying on this, I was on extended leave not checking my emails much)

PR #1331, opened 2022-05-09
Remove CrateInfo expectation from rust_binary

keith:
Fixes https://github.com/bazelbuild/rules_rust/issues/1330


keith:
not sure if this is valid, want to see what CI says

PR #1332, opened 2022-05-10
Build `rust_test` targets with `crate` using the same crate name as the underlying library target.

cfredric:
This PR changes `rust_test` to use the same crate name as the underlying target, in cases where the `crate` attribute is used. This is consistent with `cargo`, which builds the crate using the name of the library when running `cargo test`.

The behavior of `rust_test` targets that don't use the `crate` attribute is unchanged.


cfredric:
Looks like there's an issue with compile_data, but I don't immediately see why. I'd have expected things to work fine, unless Bazel is overwriting the output of one target with the output of another, due to the test using the same crate name as the library?

cfredric:
FWIW, I believe the failure to link on some windows CI bots is some transient issue - it's not a consistent failure.

cfredric:
Friendly ping, could someone please take a look at this?

UebelAndre:
@scentini were you gonna take a look?

test/rust/src/lib.rs
```diff
@@ -12,4 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![crate_name = "hello_lib"]
```
scentini, line 15: Why do we need this change?

scentini (commented):
Except for the one question I have, LGTM.

scentini:
Sorry for the delay, I thought I'd already reviewed this PR.

PR #1334, opened 2022-05-11
Replace slashes with underscores in default crate names.

cfredric:
This change is consistent with the existing documentation, which says that all illegal characters will be converted to underscores (but in reality, only hyphens were converted).

This change has no risk of breakage, since it turns broken builds into passing ones, but not vice versa.

### DIFF
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -272,7 +272,9 @@ def name_to_crate_name(name):
     Returns:
         str: The name of the crate for this target.
     """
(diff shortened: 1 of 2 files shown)


illicitonion (commented):
LGTM, but looks from CI like something on Windows isn't loving the change?

cfredric:
Fixed, PTAL!

illicitonion (approved):

password633:
****

PR #1336, opened 2022-05-13
Replace slashes with underscores in default crate names.

cfredric:
This is another attempt at #1334, which sought to make "/" characters legal in Rust target names.

This PR differs from #1334 in how it handles build artifacts that are based on `crate_info.output`. There are a few places that currently assume that calling `ctx.actions.declare_file(...)` will always declare a file in the same directory as `crate_info.output`, i.e. as `ctx.actions.declare_file(ctx.label.name + ...)`. In a world where label names may include slashes (which is the case today!), this is not a safe assumption, since a slash in the target name causes bazel to create a subdirectory for the output.

Specifically, this PR fixes incorrect locations for the following items (when the target name includes a slash):
* PDB files
* dSYM files
* .lib files

This leaves a few things that look like they make the same assumption, but I'm not sure how they're used or if they should be updated:
* `"_ambiguous_libs/" + crate_info.output.basename + "/" + symlink_name` for ambiguous libraries
* The linkstamp output path (`"_objs/" + crate_info.output.basename + "/" + ...`)

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -923,7 +923,7 @@ def rustc_compile_action(
     if toolchain.os == "windows" and crate_info.type == "cdylib":
         # Rustc generates the import library with a `.dll.lib` extension rather than the usual `.lib` one that msvc
         # expects (see https://github.com/rust-lang/rust/pull/29520 for more context).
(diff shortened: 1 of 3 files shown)


scentini (approved):
Thanks!

PR #1340, opened 2022-05-16
Support crates that have mix of generated and nongenerated inputs

scentini:
Rustc expects the crate's sources to exist next to the crate's root.
When a rust target specifies sources that are a mixture of generated and non-generated files, some of them will be located in the source directory, while others under `bazel-out`, which violates the above expectation.

This PR adds support for specifying both generated and non-generated sources in the same target by creating symlinks for the non-generated sources.

### DIFF
--- a/test/generated_inputs/BUILD.bazel
+++ b/test/generated_inputs/BUILD.bazel
@@ -6,7 +6,7 @@ load(":input_from_different_cfg.bzl", "input_from_different_cfg")
 write_file(
     name = "generate_src",
     out = "src.rs",
-    content = ["""
+    content = """
 #[cfg(not(generated_file_as_root))]
 pub fn forty_two() -> i32 { 42 }
 
@@ -17,7 +17,7 @@ mod lib_for_src;
 pub fn get_forty_two_as_string() -> String {
     format!("{}", lib_for_src::forty_two())
 }
-""".splitlines()],
+""".splitlines(),
     newline = "unix",
 )
 


rust/private/rust.bzl
```diff
@@ -124,11 +124,63 @@ def get_edition(attr, toolchain, label):
...
+
+    generated_root = crate_root
+    if crate_root and crate_root.is_source:
+        generated_root = ctx.actions.declare_file(crate_root.basename)
+        ctx.actions.symlink(
+            output = generated_root,
+            target_file = crate_root,
+            progress_message = "Creating symlink to source file: {}".format(crate_root.path),
+        )
+    if generated_root:
```
UebelAndre, line 160: What is this guarding against? Will `generated_root` ever be a falsey value?

rust/private/rust.bzl
```diff
@@ -124,11 +124,63 @@ def get_edition(attr, toolchain, label):
...
+        if src.is_source:
+            src_symlink = ctx.actions.declare_file(src.basename)
+            ctx.actions.symlink(
+                output = src_symlink,
+                target_file = src,
+                progress_message = "Creating symlink to source file: {}".format(src.path),
+            )
+            generated_sources.append(src_symlink)
+        else:
+            generated_sources.append(src)
```
UebelAndre, line 175: What happens if some source files come from different configurations? It seems like if you have any generated source files, that you'd then want to explicitly make symlinks for all files since they wouldn't all be guaranteed to be under the same path in `bazel-out`, right?

test/generated_inputs/BUILD.bazel
```diff
@@ -0,0 +1,76 @@
+load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_library", "rust_test")
+
+genrule(
```
UebelAndre, line 3: Can you use [@bazel_skylib//rules:write_file.bzl](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/write_file_doc.md) instead of `genrule` here?

UebelAndre (changes requested):
Thanks for working on this! Just a few questions 😄

UebelAndre:
Hey @scentini any updates here? The changes seem almost good to go and I just came across a case where this would be useful (and this is not the first time 😅).

scentini:
Hey @UebelAndre, other things took priority, but as I now know that it's not only my use case, I will prioritize accordingly!

rust/private/rust.bzl
scentini: Yes, in the case where `crate_root` was not specified on the target.

rust/private/rust.bzl
scentini: I modified the code to address generated files coming from different configurations, and added a test. PTAL!

test/generated_inputs/BUILD.bazel
scentini: Done.

PR #1343, opened 2022-05-19
Include generated snippets from wasm_bindgen

rharink:
When generating wasm bindings for targets `web` or `bundler` a `snippets` directory is created containing imports related to the generated bindings. 

This fix includes the snippets directory in the Bazel output, so it can be used by other targets.


wasm_bindgen/wasm_bindgen.bzl
```diff
@@ -104,6 +104,9 @@ def _rust_wasm_bindgen_impl(ctx):
     js_out = [ctx.actions.declare_file(ctx.attr.name + ".js")]
     ts_out = [ctx.actions.declare_file(ctx.attr.name + ".d.ts")]
 
+    if ctx.attr.target == "web" or ctx.attr.target == "bundler":
+        js_out.append(ctx.actions.declare_directory("snippets"))
```
UebelAndre, line 108: Should this be somehow scoped using `name`?
```suggestion
        js_out.append(ctx.actions.declare_directory(ctx.attr.name + "/snippets"))
```

UebelAndre (changes requested):
Thanks for opening this PR and sorry for the delay here!

I'm wondering if snippets needs to be explicitly wired up using a unique arg?
https://github.com/bazelbuild/rules_rust/blob/0.4.0/wasm_bindgen/wasm_bindgen.bzl#L113-L118

I otherwise don't think the path is guaranteed to match the declared directory.

wasm_bindgen/wasm_bindgen.bzl
rharink: Hi, no worries :)

The directory generated by wasm-bindgen is always named `snippets` and is generated next to the javascript files in the root (--out-dir) directory. Scoping the `snippets` directory would result in a subdirectory `<name>/snippets` which would be unresolvable. Or am I missing something 😅.

Maybe we could use a unique arg to include the snippet directory or not, `include_snippets = True/False`? 

I've made an example repo that builds a wasm target and uses wasm-bindgen to generate a minimal example with a `snippets` directory: https://github.com/rharink/test_wasm_bindgen_snippets.

UebelAndre: My thought is if you have multiple targets in the same package that the `snippet` output would conflict.

rharink: We can't influence where `wasm-bindgen` creates the `snippets` directory, or it's name. Any ideas on how we could solve this when multiple targets use snippets?

UebelAndre: you'll probably have to trick `wasm-bindgen` into rendering into a named directory. Which you could probably do by changing:
```starlark
js_out = [ctx.actions.declare_file(ctx.attr.name + ".js")]
ts_out = [ctx.actions.declare_file(ctx.attr.name + ".d.ts")]
```
to
```starlark
js_out = [ctx.actions.declare_file("{0}/{0}.js".format(ctx.attr.name)]
ts_out = [ctx.actions.declare_file("{0}/{0}.d.ts".format(ctx.attr.name")]
```
The the suggestion above would probably match. But I don't know what impact this would have downstream.

rickvanprim:
@rharink are you still interested in trying to get this in?  Based on my very limited usage of `wasm_bindgen` and hitting this issue, I'm hoping this can get merged (and I can remove the local `patch`) 🙂

rharink:
Hi, yeah I'm also working with a local patch 😅. Got lost in work. Ill try and pick this up asap.

wasm_bindgen/wasm_bindgen.bzl
rharink: I've added a scoped directory so multiple targets can use the snippets directory. 
One concern is the output is now always generated in the scoped directory, is this a breaking change?

UebelAndre:
Hello, is this still a desired change? 😄

rickvanprim:
> Hello, is this still a desired change? 😄

I would still find this useful 🙂  One less patch 😅

PR #1344, opened 2022-05-19
feat: build script toolchains annotations

roman-kashitsyn:
This change adds a new attribute to crate.annotation,
"build_script_toolchains". This attribute specifies the value of
"toolchains" that cargo-bazel will add to the generated
`cargo_build_script` target.

I've tested the change by annotating a few targets with the new
attribute and inspecting the generated build files.


UebelAndre (changes requested):
The unittests have some hashes in them to indicate when changes impact the need to regenerate dependencies. You'll need to update these: https://github.com/bazelbuild/rules_rust/blob/0.4.0/crate_universe/src/lockfile.rs#L240-L382

This also means examples will need to be regenerated.

roman-kashitsyn:
> This also means examples will need to be regenerated.

What's the correct way of regenerating examples? I tried `CARGO_BAZEL_REPIN=1 bazel test //..." in `examples/crate_universe`, but `vendor_*` examples do not compile:

```
error: couldn't read vendor_local_pkgs/crates/log-0.4.16/build.rs: No such file or directory (os error 2)
```

UebelAndre:
> @UebelAndre , what's the correct way of regenerating examples? I tried `CARGO_BAZEL_REPIN=1 bazel test //...` in `examples/crate_universe`, but `vendor_*` examples do not compile:
> 
> ```
> error: couldn't read vendor_local_pkgs/crates/log-0.4.16/build.rs: No such file or directory (os error 2)
> ```

I've added some notes on that here: https://github.com/bazelbuild/rules_rust/pull/1351

TLDR: Try pushing the lockfiles in a new commit. CI will do the bootstrapping needed to fix the error you're seeing.

UebelAndre (approved):
Thanks!

PR #1346, opened 2022-05-19
Update apple_support

keith:
There aren't any changes to what rules_rust uses in this update, but if
folks need this for other reasons this reduces transitive version
issues.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -65,8 +65,8 @@ def rules_rust_dependencies():
     maybe(
         http_archive,
         name = "build_bazel_apple_support",
-        sha256 = "76df040ade90836ff5543888d64616e7ba6c3a7b33b916aa3a4b68f342d1b447",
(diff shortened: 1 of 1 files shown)


keith:
The windows failure here appears to be unrelated 🤔

UebelAndre:
> The windows failure here appears to be unrelated 🤔

https://buildkite.com/bazel/rules-rust-rustlang/builds/6283#6cb312ce-1688-44f0-b707-97f4c8ba82cb
```
  = note: LINK : fatal error LNK1104: cannot open file 'bazel-out\x64_windows-fastbuild\bin\vendor_remote_pkgs\vendor_remote.vendor_remote.942c26a8-cgu.0.rcgu.o'
```

It does indeed. @hlopko I'm wondering if there's something up with windows workers. Do you know who'd be able to help look into that?

UebelAndre (approved):
Thanks!

PR #1347, opened 2022-05-20
Added Rust 1.61.0

UebelAndre:
https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html


illicitonion (approved):

UebelAndre:
@wt Hey, do you know why the `known_shas.sh` looks so pruned after running `bazel run util:fetch_shas`?

util/fetch_shas.sh
```diff
@@ -54,54 +54,47 @@ enumerate_keys() {
   done
 }
 
-emit_bzl_file_contents() {
-  if which parallel > /dev/null; then
```
UebelAndre, line 58: This code path generated different content than the other path. It was removed because it's slower and there should be only one way to update known_shas.bzl

util/fetch_shas.sh
```diff
@@ -54,54 +54,47 @@ enumerate_keys() {
...
-        > "${TMPDIR}"/shas.txt
-    popd >/dev/null
-  fi
+  pushd "${out_dir}" > /dev/null
+  find . -type f -print | \
+  awk '{
+      file_key=substr($1, 3);
+      getline <$1;
+      printf("%s %s\n", file_key, $1);
+      if (match(file_key, /\.tar\.\(gz|xz\)$/)) {
```
UebelAndre, line 79: This line now correctly handles xz extensions and avoids a regression.

util/fetch_shas.sh
UebelAndre: > @wt Hey, do you know why the `known_shas.sh` looks so pruned after running `bazel run //util:fetch_shas`?

Found the issue and resolved it in d1620614353bbd6d65101f8cbe026b6a78ea9306. @illicitonion can you take a look at that commit and let me know what you think? I can move that into it's own PR if you'd prefer.

illicitonion (approved):

util/fetch_shas.sh
UebelAndre: This was resolved by https://github.com/bazelbuild/rules_rust/pull/1355

wt:
> @wt Hey, do you know why the `known_shas.sh` looks so pruned after running `bazel run //util:fetch_shas`?
> 
> edit: This was resolved by #1355

This was clearly a bug from me trying to avoid messing with the implementation that used the gnu parallel command if it existed. Thanks for fixing. Sorry about that. I saw that the fix also removed the keys without extensions. I hope that was tested well.

PR #1349, opened 2022-05-20
Ensure memchr is linked after libobject

bsilver8192:
Some other changes (coming in a separate PR) triggered these precompiled .rlibs to be linked in the wrong order. On amd64 at least, libobject has a dependency on memchr. It's an SSE-specific symbol, so it might be different on other platforms.

Not sure how to write a test for this unfortunately. depset flattening order is unspecified (within the constraints of the topological sort) but deterministic for a given configuration.


bsilver8192:
I think the CI failure is unrelated. If it's something I broke, please let me know.

krasimirgg (approved):
This PR looks good. Thank you!

hlopko:
@UebelAndre what do you think about creating a new rules_rust release with this commit? Seems to me this blocks all users on latest Rust.

bsilver8192:
> @UebelAndre what do you think about creating a new rules_rust release with this commit? Seems to me this blocks all users on latest Rust.

Did you get it to fail before this change with a specific Rust version? I observed it working on the Rust version from rules_rust WORKSPACE, then my tweak to where allocator_library falls in the link order in #1350 changed Bazel's choice of ways to flatten the depset and broke it.

hlopko:
For the record, using `lld` as the linker works around the issue.

Issue #1238, opened 2022-03-30
Recommendation for allocators when linking with C* binary


keith:
When using rules_rust to link a rust library into a final C / C++ binary, you end up hitting some linker issues around missing allocator symbols:

```
Undefined symbols for architecture arm64:
  "___rust_alloc", referenced from:
      core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h2255888a02196d18 in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h8c8e34d5191bcf2c in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      alloc::raw_vec::finish_grow::hf9c3a7b5d277e764 in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      addr2line::Context$LT$R$GT$::find_frames::h2faede176ceedb5a in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      addr2line::ResDwarf$LT$R$GT$::parse::hde4e81cf26db861b in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      addr2line::ResUnit$LT$R$GT$::parse_lines::h989d285cff418912 in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      addr2line::ResUnit$LT$R$GT$::render_file::h057f5d698fb8b6fa in libstd-eb660d415c354e23.a(std-eb660d415c354e23.std.dfe53c44-cgu.0.rcgu.o)
      ...
  "___rust_alloc_error_handler", referenced from:
      alloc::alloc::handle_alloc_error::rt_error::h0566a1d251bd0a5b in liballoc-7e50779556d46264.a(alloc-7e50779556d46264.alloc.6069a24a-cgu.0.rcgu.o)
  "___rust_alloc_zeroed", referenced from:
[...]

PR #1350, opened 2022-05-20
Add an example of linking a rust_library into a cc_binary

bsilver8192:
allocator_library is now exposed at a higher level to support this.

This required swapping the order of libstd and allocator_library in the link. allocator_library is moving earlier, so this should enable it to depend on strictly more libraries than anything which worked before.


krasimirgg (approved):
This looks good. Thank you!

examples/cc_binary/forward_allocator.c
```diff
@@ -0,0 +1,46 @@
+#include <stdint.h>
```
keith, line 1: should we move this file to a shared location where consumers of the rules can use it as well?

bsilver8192: Do you want to maintain it as a supported library here, or ask each project that uses it to maintain a version compatible with their rust toolchains? As you can see from the change I made yesterday, it does require updates for new Rust toolchains sometimes, because this is no stable interface to this part of the Rust ABI.

keith: Yea I can't say for sure since I'm not a maintainer on this, but as a user I would love it to be maintained here for everyone as opposed to me having to copy paste 🙃

krasimirgg: I think this is good as an example, but at least right now IMO it doesn't seem like something to add to a shared location and to commit rules_rust to maintain a specific version of this. My main concern is that this is subtle enough that in some non-trivial environments it would require some workspace-specific tweaks that will be hard to maintain and not be universal for all users. For example, we've got an environment where we have to weaken some of the `__rust*` symbols, otherwise we may run into duplicate symbol errors in some configurations when the dependency graph is sufficiently complicated.
I think it's worthwhile to follow-up the conversation about this over at https://github.com/bazelbuild/rules_rust/issues/1238.

@hlopko also may have thoughts about this.

bsilver8192:
I'm happy to merge main to update this if the maintainers want to merge it, let me know. I definitely understand not wanting to maintain these hacks though.

krasimirgg:
> I'm happy to merge main to update this if the maintainers want to merge it, let me know. I definitely understand not wanting to maintain these hacks though.

This PR looks good to me, ready to merge after updating on top of the main branch.

rust/toolchain.bzl
```diff
@@ -196,20 +196,20 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_std, allocator_library):
...
+        allocator_inputs = []
         if allocator_library:
-            allocator_inputs = [allocator_library[CcInfo].linking_context.linker_inputs]
+            allocator_inputs = allocator_library[CcInfo].linking_context.linker_inputs.to_list()
 
         cc_infos.append(CcInfo(
             linking_context = cc_common.create_linking_context(
                 linker_inputs = depset(
-                    [link_inputs],
-                    transitive = allocator_inputs,
```
krasimirgg, line 212: Wouldn't it be more precise instead of switching the order here (effectively declaring allocator_inputs to depend on all of rust stdlib), to have allocator inputs depend just on alloc_inputs?

examples/cc_binary/forward_allocator.c
```diff
@@ -0,0 +1,52 @@
...
+//
+// This file strongly assumes that the default allocator is used. It will
+// not work with any other allocated switched in via `#[global_allocator]`.
+
+// New feature as of https://github.com/rust-lang/rust/pull/88098.
+__attribute__((weak)) uint8_t __rust_alloc_error_handler_should_panic = 0;
+
+uint8_t *__rdl_alloc(uintptr_t size, uintptr_t align);
+__attribute__((weak))
+uint8_t *__rust_alloc(uintptr_t size, uintptr_t align) {
```
krasimirgg, line 28: could you add `extern "C"` before `__attribute__((weak))` of the __rust_* definitions? Without it, there is risk that these names get mangled in some configurations.

bsilver8192:
#1490 addresses my use case perfectly. I don't think there's anything useful this PR adds on top of it.

PR #1353, opened 2022-05-20
Re-enable disabled windows jobs in CI

UebelAndre:


hlopko (approved):
Seems extra_workspace_members is still failing? We can submit this without that one if you want.

hlopko:
Hmm or is windows CI not healthy right now? https://buildkite.com/bazel/rules-rust-rustlang/builds/6292#0e035c11-ab76-45a0-91da-309a78e6d7ce seems unrelated to https://github.com/bazelbuild/rules_rust/pull/1349 and the error message looks similar to this one.

UebelAndre:
> Hmm or is windows CI not healthy right now? https://buildkite.com/bazel/rules-rust-rustlang/builds/6292#0e035c11-ab76-45a0-91da-309a78e6d7ce seems unrelated to #1349 and the error message looks similar to this one.

I've yet to take a closer look at this but yeah, there seems to be something systemic happening. Either a crate was updated that does not work with the way CI is setup today, or CI infra was updated and is no longer compatible with `rules_rust`.

PR #1356, opened 2022-05-23
Ensure crate_features to be included correctly in rust_project.json

mypmc:
Fixed that the attribute `crate_features` is not included in rust_project.json in some cases.

I'm not sure what caused the problem, but I've confirmed that this fixes my issue.


tools/rust_analyzer/aquery.rs
```diff
@@ -48,7 +48,7 @@ pub struct CrateSpec {
     pub deps: BTreeSet<String>,
     pub proc_macro_dylib_path: Option<String>,
     pub source: Option<CrateSpecSource>,
-    pub cfg: Vec<String>,
+    pub cfg: BTreeSet<String>,
```
UebelAndre, line 51: I don't think this should be `BTreeSet`. Wouldn't we want this to maintain the original sorting? At least looking at https://rust-analyzer.github.io/manual.html#non-cargo-based-projects `cfg` seems to be represented by an unsorted `string[]`. Does your fix still apply if this is `Vec<String>`?

mypmc: `Vec<String>` should be okay here.
But I'm not sure whether if we need to `dedup` or not.

Using `sort` to dedup `cfg`s.

```rust
existing.cfg.extend(spec.cfg);
existing.cfg.sort();
existing.cfg.dedup();
```

What do you think?

FYI: Both of `rustc` and `rust-analyzer` seem to parse `cfg`s into `HashSet`.

UebelAndre: Do `rustc` and `rust-analyzer` also do sorting?

UebelAndre (approved):
I think this seems good, thanks! I'd like @hlopko to take a look as well and had one follow up question on a previous thread, if you don't mind.

tools/rust_analyzer/aquery.rs
mypmc: I think either of them don't sort cfgs but not sure.
- https://github.com/rust-lang/rust/blob/master/compiler/rustc_interface/src/interface.rs#L83-L141
- https://github.com/rust-lang/rust-analyzer/blob/5c88d9344c5b32988bfbfc090f50aba5de1db062/crates/project-model/src/workspace.rs#L475-L476

PR #1359, opened 2022-05-24
Don't leak native dependencies of proc_macro

scentini:
Native dependencies of a `rust_proc_macro` shouldn't propagate to the libraries depending on the `rust_proc_macro` target.


UebelAndre (approved):
Looks great! Thank you!

PR #1362, opened 2022-05-26
Add armv7-linux-androideabi tier 2 triple

keith:
From https://doc.rust-lang.org/rustc/platform-support.html, required for
use with raze.

Related: https://github.com/google/cargo-raze/pull/489


keith:
https://github.com/google/cargo-raze/pull/489

UebelAndre (approved):
Thanks! The secret to passing CI is to do the following:
- run `./docs/update_docs.sh`
- run `cd examples/crate_universe && REPIN=1 bazel test //...`

If you commit and push the results from that I think CI will be happy.

keith:
Thanks!

PR #1363, opened 2022-05-27
Add `env` attribute to `rust_toolchain`.

UebelAndre:
This change allows users to set unique environment variables on `rust_toolchain` and applies them to `Rustc`  and `CargoBuildScriptRun` actions.


hlopko (approved):
It's a pity we can't easily test these changes..

UebelAndre:
> It's a pity we can't easily test these changes..

I was thinking the same thing. Do you know of any repo that has good tests for their toolchains?

PR #1364, opened 2022-05-27
Update rules_perl in examples

keith:
Without this update the perl toolchain wouldn't be found on my M1 mac


UebelAndre (approved):
Thanks!

PR #1365, opened 2022-05-27
Use a more compact encoding in the `import` macro.

cfredric:
This PR just tweaks the encoding scheme used by the `import` macro to be more compact in most cases; and adds more unit test coverage for the encoding/decoding (on both the Rust side and the bzl side).

This only has an effect when the experimental crate renaming flag is enabled, so it will be a no-op for most users.


scentini (approved):
Thanks!

PR #1366, opened 2022-05-28
Use target instead of rule in rust_register_toolchains edition docs

dfreese:
It's a really pedantic change, but is more consistent with bazel terminology.


UebelAndre (approved):
Thanks!

PR #1368, opened 2022-05-29
Fix incorrect assertion in test_env_launcher test

dfreese:
This appears to be a copy-paste error in #1079, since the assertion is
already is currently duplicated on line 26.


UebelAndre (approved):
Good catch, thanks!

PR #1370, opened 2022-05-31
Avoid using common substrings as encodings in `import` macro.

cfredric:
This tweaks the encoding used by the `import` macro to avoid using encodings that could be common substrings in target names.

E.g., `foo_c_api` is a perfectly reasonable target name, but since it contains the substring `_c_`, it must be mangled by the macro. It would be better if the encoding avoided such cases.


scentini (approved):
Thanks!

PR #1371, opened 2022-06-01
cargo: Fix handling of relative sysroots

kalcutter:
Currently, `cc_toolchain.sysroot` is appended as an argument to "CC" and
"CXX". This argument, however, will typically be overridden by any
--sysroot flag already present in "CFLAGS" and "CXXFLAGS". When sysroot
is relative, this fails because cargo_build_script_runner is not
executed from the execroot.

Do not append `--sysroot` to "CC" and "CXX". Instead, modify any
--sysroot arguments with relative paths to be absolute.


illicitonion:
Thanks for the contribution! This definitely looks nicer, and I _think_ should be fine because we now properly populate CFLAGS and CXXFLAGS (as of https://github.com/bazelbuild/rules_rust/pull/1081), which should contain a `--sysroot=` flag if it needs to be set, but I'm also pretty sure we have 0 test coverage for needing a custom sysroot...

@sayrer do you still have a handy repro environment from https://github.com/bazelbuild/rules_rust/issues/659 to make sure it's now safe to effectively revert https://github.com/bazelbuild/rules_rust/pull/664, which this PR does?

kalcutter:
@illicitonion I would have included a test but I didn't see an easy way. Would you be OK to merge this without a test for 0.6.0 since it fixes real bugs? We can create an issue to improve the testing of cross-compilation?

illicitonion (approved):
Makes sense - let's 🤞 and run with it :) Thanks again!

kalcutter:
Thanks! 🎉

PR #1375, opened 2022-06-04
Refactored the Rustfmt tool to work cross-platform

UebelAndre:
This change updates the `@rules_rust//tools/rustfmt` tool to no longer rely on building and reading manifest files to determine what sources to query. `bazel query` is expressive enough to request the list of source files that should be formatted.


UebelAndre:
closes https://github.com/bazelbuild/rules_rust/pull/872

UebelAndre:
This change is blocked by:
- https://github.com/bazelbuild/rules_rust/pull/1377

tools/rustfmt/srcs/lib.rs
```diff
@@ -67,8 +70,33 @@ pub fn parse_rustfmt_manifest(manifest: &Path) -> RustfmtManifest {
...
 }
+
+#[cfg(target_family = "windows")]
+const PATH_ENV_SEP: &str = ";";
+
+#[cfg(target_family = "unix")]
+const PATH_ENV_SEP: &str = ":";
+
+/// Parse the runfiles of the current executable for manifests generated
+/// but the `rustfmt_aspect` aspect.
```
illicitonion, line 91:
```suggestion
/// by the `rustfmt_aspect` aspect.
```

tools/rustfmt/srcs/main.rs
```diff
@@ -14,52 +18,46 @@ fn main() {
...
-            // Early return if a list of targets and not packages were provided
-            if is_all_targets {
-                return options.packages.clone();
+/// The edition to use in cases where the default edition is unspecified by Bazel
+const FALLBACK_EDITION: &str = "2018";
+
+/// Determine the Rust edition to use in cases where a target has not explicitly
+/// specified the edition via an `edition` attribute.
+fn get_default_edition() -> &'static str {
+    // TODO: THe `RUST_DEFAULT_EDITION` should be set by the `rust_toolchain`
```
illicitonion, line 27: When isn't this? What would break if we just did `env!("RUST_DEFAULT_EDITION")` rather than the option/fallback stuff? It seems like there's a _lot_ of code here to handle this case?

tools/rustfmt/srcs/main.rs
```diff
@@ -71,74 +69,111 @@ fn query_rustfmt_targets(options: &Config) -> Vec<String> {
...
+        // Filter all local targets ending in `*.rs`.
+        //     Get all source files.
+        //         Get direct dependencies.
+        //             Get all targets with an empty `edition` attribute.
+        //             Except for targets tagged with `norustfmt`.
+        //             And except for targets with a populated `crate` attribute since `crate` defines edition for this target
         format!(
-            "--aspects={}//rust:defs.bzl%rustfmt_aspect",
-            env!("ASPECT_REPOSITORY")
+            r#"let scope = set({scope}) in filter("^//.*\.rs$", kind("source file", deps(attr(edition, "{edition}", $scope) except attr(tags, "(^\[|, )norustfmt(, |\]$)", $scope) + attr(crate, ".*", $scope), 1)))"#,
```
illicitonion, line 85: It feels worth factoring out this query from the two places it's present, because it's pretty intricate and I imagine coordinating changes across the two copies will be hard?

illicitonion (approved):

tools/rustfmt/srcs/main.rs
UebelAndre: Updated, thanks for calling this out!

test/generated_inputs/BUILD.bazel
```diff
@@ -28,6 +28,7 @@ rust_library(
         ":src.rs",
     ],
     edition = "2018",
+    tags = ["norustfmt"],
```
UebelAndre, line 31: All the tagging here is to work around the following error when a target has a mix of local and generated source files.
```
Error writing files: failed to resolve mod `src`: /Users/user/Code/rules_rust/test/generated_inputs/src.rs does not exist
Error writing files: failed to resolve mod `src`: /Users/user/Code/rules_rust/test/generated_inputs/src.rs does not exist
Error writing files: failed to resolve mod `generated_in_different_cfg`: /Users/user/Code/rules_rust/test/generated_inputs/generated_in_different_cfg.rs does not exist
```

This error comes from `rustfmt` but given that the functionality for supporting generated sources in `rules_rust` is new, I'd rather try to account for this in a separate PR.

UebelAndre:
@illicitonion would you be able to take one more look at this PR?

illicitonion (approved):
:shipit:

UebelAndre:
The query introduced here is susceptible to what I believe is a bug in how certain files are represented:
- https://github.com/bazelbuild/bazel/issues/15850

PR #1376, opened 2022-06-04
Renamed toolchain files targets

UebelAndre:
The configuration of the `toolchain_tool` targets is up to however they're consumed. Claiming they're `exec` is misleading.


scentini (approved):
Thanks!

PR #1378, opened 2022-06-04
Added comments to internal function

UebelAndre:
Minor commenting while I was learning more about [pic](https://en.wikipedia.org/wiki/Position-independent_code).


scentini (approved):
Thanks!

PR #1381, opened 2022-06-06
Added support for vendoring external manifests

UebelAndre:
This allows users to take Cargo manifests from external repositories and vendor dependencies for them. The most common application for this feature is for Bazel rules which are powered by a binary-only crate (a crate that offers no rlib) (e.g. `wasm_bindgen`).


illicitonion (approved):
LGTM - thanks!

PR #1391, opened 2022-06-07
Use  `target_compatible_with` to make `macos` with `Xcode` happy

scentini:
Bazel's default toolchain configuration does not support dynamic linking when the Xcode toolchain is used on `macos`. Some tests that target dynamic linking caused `bazel build //...` to fail for `rules_rust` developers that have Xcode installed.
This PR adds `target_compatible_with` to the relevant targets so that they are skipped on `macos`.


UebelAndre (approved):
Thanks!

PR #1392, opened 2022-06-07
Fixed Clippy bug with `--@rules_rust//:clippy_flags`.

UebelAndre:
This change fixes the following error when using the `--@rules_rust//:clippy_flags` config setting
```
ERROR: /Users/user/Code/rules_rust/test/clippy/BUILD.bazel:19:10: in //rust/private:clippy.bzl%rust_clippy_aspect aspect on rust_test rule //test/clippy:ok_test:
	File "/Users/user/Code/rules_rust/rust/private/clippy.bzl", line 147, column 29, in _clippy_aspect_impl
		args.rustc_flags.extend(clippy_flags)
Error: 'Args' value has no field or method 'extend'
WARNING: errors encountered while analyzing target '//test/unit/clippy:library_clippy_aspect_with_extra_flags_test': it will not be built
```

Regression testing was also added to catch future breakages.


scentini (approved):
Thanks!

PR #1396, opened 2022-06-09
Ran clang-format on all C++ sources

UebelAndre:
Minor cleanup. The following config was used:
```yaml
---
BasedOnStyle: Google
...
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+---
+BasedOnStyle: Google
+...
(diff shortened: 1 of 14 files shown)


UebelAndre:
> I'm not sure this is worthwhile... If we're going to apply formatting, I think I'd want to have CI enforce that we stay formatted, but given we only really have C/C++ in tests and examples, I'm not sure it's worth gating CI on test formatting... Is there a strong motivation here?

I was testing some other rules I wrote to run clang-format and thought I'd just push the changes to have cleaner code. This PR can be rejected since it doesn't add enforcement but figured the cleanup wouldn't hurt.

illicitonion (approved):

Issue #1374, opened 2022-06-02
Possible regression causing linking of proc_macro


snowp:
We updated rules_rust from 0.4.0 to 28ac6b133dfce65f9e9e01a5de2b4e034410e080 and ran into an issue where proc_macro is now statically linked into a test binary, causing 

```
/home/runner/.cache/bazel/_bazel_runner/374864cd5aeb008ef6a614637fda772e/sandbox/linux-sandbox/630/execroot/__main__/bazel-out/k8-fastbuild/bin/src/wrapper/buffer_test.runfiles/__main__/src/wrapper/_buffer_test: error while loading shared libraries: libstd-05b39ac0cb4c5688.so: cannot open shared object file: No such file or directory
```

21eed19188c0359d72f9a508a0c0e7040ff20070...28ac6b133dfce65f9e9e01a5de2b4e034410e080

I made sure that this was not coming from the Rust version upgrade (tried bad sha + 1.60). 

I don't have a simple reproducer, though I suspect it's coming from our usage of cxx, though we also make use of other proc_macro libraries such as tokio, thiserror, prost, etc.

scentini:
I managed to create a [simple repro](https://github.com/scentini/repro_1374). This is not related to `proc_macros`; the culprit is https://github.com/bazelbuild/rules_rust/pull/1333 and @krasimirgg and I are discussing the best way forward.

scentini:
@snowp could you check whether https://github.com/bazelbuild/rules_rust/pull/1394 solves the problem for you?

scentini:
Thanks! That's the same error, just on macos this time, which is a reminder that we should apply the solution to all platforms.

Reflejo:
@scentini Were you able to make any progress with this issue? Thanks!

scentini:
Hey folks, I have https://github.com/bazelbuild/rules_rust/pull/1500 up for review.

snowp:
After applying this fix I'm still seeing
```
dyld[51630]: Library not loaded: '@rpath/libtest-e6734f05acc07fba.dylib'
  Referenced from: '/private/var/tmp/_bazel_snowp/c1591b76b8a2fcd15b6ff3cd34fdd7da/execroot/__main__/bazel-out/darwin_arm64-fastbuild/bin/test/client/logger_runner'
[...]

PR #1397, opened 2022-06-09
Updated `rules_rust` version to `0.6.0`.

UebelAndre:


UebelAndre:
If possible I'd like this to include:
- https://github.com/bazelbuild/rules_rust/pull/1381
- https://github.com/bazelbuild/rules_rust/pull/1375

scentini:
Could we please also wait for a solution for https://github.com/bazelbuild/rules_rust/issues/1374, which should come today/early next week?

kalcutter:
Please also include #1371 if its possible to get that merged in a reasonable amount of time.

kalcutter:
#1371 has been merged :tada:

UebelAndre:
> Could we please also wait for a solution for #1374, which should come today/early next week?

@scentini hey, any updates here? I'd love to be able to get a release out and think it'd be worth while to do a patch release whenever this is fixed.

scentini:
If a patch release is on the table let's not block 0.6.0. I tried multiple approaches yesterday, but there are some issues with macos behavior (and I haven't gotten yet to checking Windows), and while I'm actively investigating, I don't have a timeline for the fix just yet.

UebelAndre:
> If a patch release is on the table let's not block 0.6.0. I tried multiple approaches yesterday, but there are some issues with macos behavior (and I haven't gotten yet to checking Windows), and while I'm actively investigating, I don't have a timeline for the fix just yet.

I'd like releases to be an easy and light weight thing to do. So I would consider a patch release (or even a new release) something always on the table. Anyone can open a PR updating the version and the automated deploy action takes care of the rest.

PR #1400, opened 2022-06-10
Added more "ignore" tags to rustfmt and clippy rules.

UebelAndre:
The advertised tags in [Bazel docs](https://bazel.build/reference/be/common-definitions#common-attributes) are all hyphen delimited. This PR updates the `rustfmt` and `clippy` rules to support this style and avoid confusion in BUILD files.

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -60,9 +60,16 @@ def _get_clippy_ready_crate_info(target, aspect_ctx):
     if target.label.workspace_root.startswith("external"):
         return None
 
-    # Targets annotated with `noclippy` will not be formatted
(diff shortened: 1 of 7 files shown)


rust/private/clippy.bzl
```diff
@@ -60,9 +60,16 @@ def _get_clippy_ready_crate_info(target, aspect_ctx):
     if target.label.workspace_root.startswith("external"):
         return None
 
-    # Targets annotated with `noclippy` will not be formatted
-    if aspect_ctx and "noclippy" in aspect_ctx.rule.attr.tags:
-        return None
+    # Targets with specific will not be formatted
```
illicitonion, line 63:
```suggestion
    # Targets with specific tags will not be formatted
```

illicitonion (approved):

PR #1404, opened 2022-06-11
Use a vec, not set for rustc_flags for crate_universe annotations

UebelAndre:
This solves for issues where users might be using the following [crate annotation](https://bazelbuild.github.io/rules_rust/crate_universe.html#crateannotation)
```starlark
"openssl-sys": [crate.annotation(
    rustc_flags = [
        "-l",
        "dylib=ssl",
        "-l",
        "dylib=crypto",
    ],
)],
```

Before this would render to a list that omitted the second `-l` argument and result in an invalid build command.
```starlark
rustc_flags = [
    "-l",
    "dylib=ssl",
    "dylib=crypto",
],
```
This is now fixed and the list renders correctly
```starlark
rustc_flags = [
    "-l",
    "dylib=ssl",
    "-l",
    "dylib=crypto",
],
```

### DIFF
--- a/crate_universe/src/context/crate_context.rs
+++ b/crate_universe/src/context/crate_context.rs
@@ -421,8 +421,6 @@ impl CrateContext {
             }
 
             // Rustc flags
-            // TODO: SelectList is currently backed by `BTreeSet` which is generally incorrect
-            // for rustc flags. Should SelectList be refactored?
             if let Some(extra) = &crate_extra.rustc_flags {
                 self.common_attrs.rustc_flags.append(&mut extra.clone());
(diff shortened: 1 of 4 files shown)


crate_universe/src/context/crate_context.rs
```diff
@@ -424,9 +424,7 @@ impl CrateContext {
             // TODO: SelectList is currently backed by `BTreeSet` which is generally incorrect
```
illicitonion, line 424: This TODO is now obsolete?

illicitonion (approved):

PR #1409, opened 2022-06-13
Updated crate_universe version (to `0.3.0`) and dependencies

UebelAndre:


UebelAndre:
https://github.com/rustsec/rustsec/issues/587 forces the `cargo-lock` pin to be `<8.0.0`

PR #1410, opened 2022-06-14
Replace `aspect_repository` with rustc_env for exposing `workspace_name`

UebelAndre:
This is minor cleanup I've found to be easier to understand and work with in comparison to `aspect_repository`.


hlopko:
This change will break us internally (for a surprising reason - we don't support external repositories, so using `@` labels like `@rules_rust` doesn't work). Since you describe it as a minor cleanup, and since the aspect_repository only exists temporarily because of https://github.com/bazelbuild/bazel/issues/11734, I'm curious if we could not submit this PR and wait a bit longer for the Bazel issue to be fixed. Then we can delete the workaround.

I've pinged the Bazel issue again now.

CC @krasimirgg

UebelAndre:
> This change will break us internally (for a surprising reason - we don't support external repositories, so using `@` labels like `@rules_rust` doesn't work). Since you describe it as a minor cleanup, and since the aspect_repository only exists temporarily because of [bazelbuild/bazel#11734](https://github.com/bazelbuild/bazel/issues/11734), I'm curious if we could not submit this PR and wait a bit longer for the Bazel issue to be fixed. Then we can delete the workaround.
> 
> I've pinged the Bazel issue again now.
> 
> CC @krasimirgg

That issue is only relevant to the `rust_analyzer` target. `rustfmt` still needs to know the workspace name and had actually stopped using consuming the variable set to `aspect_repository`. So there it'd be a no-op cleanup but figured the same mechanism could be used to solve for this issue and reduce the footprint introduced by https://github.com/bazelbuild/bazel/issues/11734 by making another rule slightly more generic and moving comments over to the only logic impacted by that. At the very least I'd like to keep the cleanup for `rustfmt` but would want to do some tighter colocation of this code in the mean time. I don't expect any fix for an open Bazel issue to make it's way into rules_rust any time soon.

UebelAndre:
On second thought, it current feels like negative value to generalize the workspace name stuff. I'll decline this PR for now but it's quite a useful trick for finding runfiles. Maybe it should be built into the [runfiles library](https://github.com/bazelbuild/rules_rust/tree/0.5.0/tools/runfiles)?

PR #1377, opened 2022-06-04
Added TemplateVariableInfo to rust_toolchain

UebelAndre:
See https://github.com/bazelbuild/bazel/issues/14009#issuecomment-921960766

PR #1412, opened 2022-06-14
Return TemplateVariableInfo

uhthomas:
Fixes https://github.com/bazelbuild/rules_rust/pull/1377#issuecomment-1154989592

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -518,7 +518,7 @@ def _rust_toolchain_impl(ctx):
         _rename_first_party_crates = rename_first_party_crates,
         _third_party_dir = third_party_dir,
     )
-    return [toolchain]
+    return [toolchain, toolchain.make_variables]
 
 rust_toolchain = rule(
     implementation = _rust_toolchain_impl,
@@ -648,8 +648,10 @@ rust_toolchain = rule(
             default = "@rules_rust//rust/settings:third_party_dir",
         ),
     },
-    toolchains = [
-        "@bazel_tools//tools/cpp:toolchain_type",
+    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
+    provides = [
+        platform_common.ToolchainInfo,
+        platform_common.TemplateVariableInfo,
     ],
     incompatible_use_toolchain_transition = True,
     doc = """Declares a Rust toolchain for use.
--- a/test/rust_toolchain_template_variable_info/defs.bzl
+++ b/test/rust_toolchain_template_variable_info/defs.bzl
@@ -1,0 +1,26 @@
+"""A module which defines rules for testing TemplateVariableInfo"""
+
+def _rust_toolchain_template_variable_info_test_impl(ctx):
+    exe = ctx.actions.declare_file(ctx.label.name + ".exe")
+    ctx.actions.symlink(
+        output = exe,
+        target_file = ctx.executable._bin,
+        is_executable = True,
+    )
+    return [DefaultInfo(
+        executable = exe,
+        runfiles = ctx.attr._bin[DefaultInfo].default_runfiles,
+    )]
+
+rust_toolchain_template_variable_info_test = rule(
+    implementation = _rust_toolchain_template_variable_info_test_impl,
+    attrs = {
+        "_bin": attr.label(
+            default = "//test/rust_toolchain_template_variable_info:bin",
+            executable = True,
+            cfg = "exec",
+        ),
+    },
+    toolchains = ["//rust:toolchain"],
+    test = True,
+)
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -222,20 +222,20 @@ rust_bindgen = rule(
 )
 
 def _rust_bindgen_toolchain_impl(ctx):
-    return platform_common.ToolchainInfo(
+    return [platform_common.ToolchainInfo(
         bindgen = ctx.executable.bindgen,
         clang = ctx.executable.clang,
         libclang = ctx.attr.libclang,
         libstdcxx = ctx.attr.libstdcxx,
         rustfmt = ctx.executable.rustfmt,
-    )
+    )]
 
 rust_bindgen_toolchain = rule(
     _rust_bindgen_toolchain_impl,
     doc = """\
 The tools required for the `rust_bindgen` rule.
 
-This rule depends on the [`bindgen`](https://crates.io/crates/bindgen) binary crate, and it 
+This rule depends on the [`bindgen`](https://crates.io/crates/bindgen) binary crate, and it
 in turn depends on both a clang binary and the clang library. To obtain these dependencies,
 `rust_bindgen_dependencies` imports bindgen and its dependencies.
 
--- a/cargo/cargo_build_script.bzl
+++ b/cargo/cargo_build_script.bzl
@@ -265,7 +265,7 @@ _build_script_run = rule(
                 List of compiler flags passed to `rustc`.
 
                 These strings are subject to Make variable expansion for predefined
-                source/output path variables like `$location`, `$execpath`, and 
+                source/output path variables like `$location`, `$execpath`, and
                 `$rootpath`. This expansion is useful if you wish to pass a generated
                 file of arguments to rustc: `@$(location //package:target)`.
             """),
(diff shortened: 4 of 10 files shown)


UebelAndre (changes requested):
Can you add a test, demonstrating how the TemplateVariableInfo would be consumed?

Additionally, would you be able to move the `str(Label())` stuff into a separate PR so the required changes to the toolchain for the new functionality is clearer?

uhthomas:
> Can you add a test, demonstrating how the TemplateVariableInfo would be consumed?
> 
> Additionally, would you be able to move the `str(Label())` stuff into a separate PR so the required changes to the toolchain for the new functionality is clearer?

Happy to do those things, will update shortly.

uhthomas:
@UebelAndre I've removed the `str(Label())` stuff and added a test. Would you be able to take a look?

.bazelci/presubmit.yml
```diff
@@ -191,6 +191,11 @@ tasks:
     platform: ubuntu2004
     run_targets:
       - "//test/rust_analyzer:rust_analyzer_test"
+  rust_toolchain_template_info_tests:
+    name: Rust TemplateVariableInfo Tests
+    platform: ubuntu2004
+    test_targets:
+      - "//test/rust_toolchain_template_variable_info:rust_toolchain_template_info_test"
```
UebelAndre, line 198: Wouldn't this test be picked up by existing jobs?

test/rust_toolchain_template_variable_info/defs.bzl
```diff
@@ -0,0 +1,26 @@
+"""A module which defines rules for testing TemplateVariableInfo"""
```
UebelAndre, line 1: There's already similar tests in [//test/unit/toolchain_make_variables/toolchain_make_variables_test.bzl](https://github.com/bazelbuild/rules_rust/blob/0.6.0/test/unit/toolchain_make_variables/toolchain_make_variables_test.bzl), could you instead expand the pattern there to cover the new cases?

rust/toolchain.bzl
```diff
@@ -518,7 +518,7 @@ def _rust_toolchain_impl(ctx):
         _rename_first_party_crates = rename_first_party_crates,
         _third_party_dir = third_party_dir,
     )
-    return [toolchain]
+    return [toolchain, toolchain.make_variables]
```
UebelAndre, line 521: So this just means any rule that depends on a `rust_toolchain` gets the make variables?

UebelAndre (changes requested):
I love all the cleanup but could the provider defnitions (with the exception of `rust_toolchain`) also be moved to a separate PR? To keep the implementation of `TemplateVariableInfo` as concise as possible?

rust/toolchain.bzl
uhthomas: Yep! As seen in the test.

test/rust_toolchain_template_variable_info/defs.bzl
uhthomas: Ah, yes. Will merge them. Thanks!

uhthomas: Only trouble is https://github.com/bazelbuild/bazel-skylib/issues/375 sort of blocks it.

UebelAndre: Do you need to write a rule that consumes the toolchain? Is the target you wrote for the test the thing you want to do? Have some target with:

```starlark
env = {
    "CARGO": "$(CARGO)",
    "RUSTC": "$(RUSTC)",
}
```

That should be testable in a similar way to the current tests. You're not transitioning these targets so couldn't you also use `current_rust_toolchain` to find values to match against if you wanted to test environment variables for the action?

UebelAndre (changes requested):

UebelAndre:
@illicitonion may have some good insight here.

test/rust_toolchain_template_variable_info/defs.bzl
UebelAndre: What do you two think of the testing on https://github.com/bazelbuild/rules_rust/pull/1416 ?

illicitonion: LGTM!

uhthomas: Looks okay, though there should be a test for the "toolchains" attribute directly on the rule. This should also enable variable substitution.

uhthomas: Yes, specifically it need to be part of the rule definition itself, not a target. Such:

```starlark
_ = rule(
    implementation = ...,
    toolchains = ["//rust/toolchain"],
)
```

UebelAndre: Ah, isn’t that currently covered by the existing rust rules? Either way, it would be easy to write an new rule to cover that.

uhthomas: >isn’t that currently covered by the existing rust rules?

I don't think so. We're specifically testing for the usage of template variables when using the toolchain.

>Either way, it would be easy to write a new rule to cover that.

I think so too :)

uhthomas: Even better, some usages of `"$(rootpath @rules_rust//rust/toolchain:current_cargo_files)"` should be replaced with the template variable info.

https://cs.github.com/bazelbuild/rules_rust?q=%22%24%28rootpath+%40rules_rust%2F%2Frust%2Ftoolchain%3Acurrent_cargo_files%29%22

https://github.com/bazelbuild/rules_rust/blob/16175c881cc1070c13a57c8db9d73ca2cab39acb/crate_universe/BUILD.bazel#L105

https://github.com/bazelbuild/rules_rust/blob/16175c881cc1070c13a57c8db9d73ca2cab39acb/crate_universe/tools/cross_installer/BUILD.bazel#L22

https://github.com/bazelbuild/rules_rust/blob/16175c881cc1070c13a57c8db9d73ca2cab39acb/crate_universe/test_data/private/BUILD.bazel#L11

UebelAndre: https://github.com/bazelbuild/rules_rust/pull/1412#discussion_r900284602 I think this should be a separate PR but sounds reasonable.

@uhthomas Can you take another look? I've updated the tests.

UebelAndre: Seems like the functionality you want is only supported in Bazel >5?
https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/0181729b-eb6e-47cf-8dc8-e291aac72a12/test/unit/toolchain_make_variables/rust_toolchain_make_variable_expansion_test/test.log

uhthomas: No? It outputs a path just fine, though the path seems to be different?

```
In test _rust_toolchain_make_variable_expansion_test from //test/unit/toolchain_make_variables:toolchain_make_variables_test.bzl: Expected env[ENV_VAR_CARGO] to be equal to 'bazel-out/k8-fastbuild/bin/external/rust_linux_x86_64/toolchain_for_x86_64-unknown-linux-gnu_impl/bin/cargo', got 'bazel-out/host/bin/external/rust_linux_x86_64/toolchain_for_x86_64-unknown-linux-gnu_impl/bin/cargo'
```

```diff
-bazel-out/k8-fastbuild/bin/external/rust_linux_x86_64/toolchain_for_x86_64-unknown-linux-gnu_impl/bin/cargo|
+bazel-out/host/bin/external/rust_linux_x86_64/toolchain_for_x86_64-unknown-linux-gnu_impl/bin/cargo
````

UebelAndre: Is this something you'd be able to pickup and find a solution for? It's also odd to me that the tests passed on all other jobs but not the Min Bazel Version ones

UebelAndre:
I think all the implementation here was merged in https://github.com/bazelbuild/rules_rust/pull/1416. If there's anything I missed on that PR I'm happy to review another, but hopefully you have the features you were hoping to get now 😄

uhthomas:
Amazing! Thanks

PR #1413, opened 2022-06-15
Add extra_rustc_flag and extra_exec_rustc_flag

mypmc:
Retake of #1407.
Each new flags are singular, and appended after the flags passed to plural ones.

Use case:

```
build --@rules_rust//:extra_rustc_flags="-C...,-C...,-Z..."

build:dev --@rules_rust//:extra_rustc_flag="-C..."
build:dev --@rules_rust//:extra_rustc_flag="-C..."

build:test --config=dev
build:test --@rules_rust//:extra_rustc_flag="-C..."

build:fuzz --config=test
build:fuzz --@rules_rust//:extra_rustc_flag="-C..."
build:fuzz --@rules_rust//:extra_rustc_flag="-C..."
build:fuzz --@rules_rust//:extra_rustc_flag="-Z..."
```

PR #1407, opened 2022-06-12
Allow multiple uses of extra_rustc_flags

mypmc:
I would like to support the following use cases.

In `.bazelrc`.
```
build:dev --@rules_rust//:extra_rustc_flags="-C..."
build:test --config=dev
build:test --@rules_rust//:extra_rustc_flags="-C..." # add additional flags
build:fuzz --config=test
build:fuzz --@rules_rust//:extra_rustc_flags="-Cpasses=sancov,-Cllvm-args=-sanitizer-coverage-level=3,-Cllvm-args=-sanitizer-coverage-inline-8bit-counters,-Znew-llvm-pass-manager=no"
```

PR #1414, opened 2022-06-15
Bump the min supported bazel version

mypmc:
To use `config.string` with `allow_multiple`.

See also #1413


krasimirgg:
This looks good to me following https://github.com/bazelbuild/rules_rust/pull/1407#issuecomment-1154772168.
Since this is a "major version" flip, adding @UebelAndre and @scentini to take a look.

scentini:
I think we can move past 4.0 at this point. Our policy states that `rules_rust` support the current Bazel LTS: https://github.com/bazelbuild/rules_rust/blob/main/COMPATIBILITY.md#what-bazel-versions-are-supported

scentini (approved):

PR #1416, opened 2022-06-17
Added `TemplateVariableInfo` to `rust_toolchain`.

UebelAndre:
This change allows `rust_toolchain` to directly provide expandable make variables instead of relying on `current_rust_toolchain`.

closes https://github.com/bazelbuild/rules_rust/pull/1412


illicitonion (approved):
Thanks!

PR #1418, opened 2022-06-23
fix: `rust_doc_test` failure to find params file

### LINKED ISSUE #1233: `rust_doc_test` failing with "Failed to load argument file"

ParkMyCar:
**Context**
Today `rust_doc_test` fails when Bazel [Args](https://docs.bazel.build/versions/main/skylark/lib/Args.html) "spill" over to a file. To a user this failure is seemingly random because Bazel will auto-magically spill `Args` onto a file when there are too many args for the command line, or when it can improve performance. The generated test script, e.g. `hellolib.rustdoc_test.sh`, is then run from runfiles, which is separate from the Arg file Bazel created.

It's especially tricky because the amount of command line args Windows supports is < macOS < Linux, so there's a "silent" OS dependency here too.

**Solution**
This PR fixes the issue by manually declaring a params file, e.g. `hellolib.rustdoc_opt_params`, that is a sibling file to our test runner, `hellolib.rustdoc_test.sh`. We then pass the path of this optional params file to our test writer. The test writer checks for the presence of the Bazel Args file, and if it finds one, copies the content into our manually declared params file. 

Our manually declared params file then gets moved into runfiles with our test script. The test script can then find the file, and `rustdoc` picks up the args.

Note: Today we detect the params file by matching a prefix of `@` and suffix of `.rustdoc_test.sh-0.params`. I'm not sure if this is accurate for all systems, or all versions of bazel. I tried the [`use_param_file`](https://docs.bazel.build/versions/main/skylark/lib/Args.html#use_param_file) option on `Args` to make it easier to detect a params file, but that seem to overwrite (or just generally effect) the other arguments we'd pass to `rustdoc_test_writer.rs`.

Fixes #1233


crate_universe/src/cli.rs
```diff
@@ -19,7 +19,11 @@ pub use splice::splice;
 pub use vendor::vendor;
 
 #[derive(Parser, Debug)]
-#[clap(name = "cargo-bazel", about, version)]
+#[clap(
+    name = "cargo-bazel",
+    about = "crate_universe` is a collection of tools which use Cargo to generate build targets for Bazel.",
+    version
+)]
```
ParkMyCar, line 26: Note: this change, the the other changes to `#[clap(author ...` are not relevant to the underlying fix, but the changes were necessary to get the rust_doc_test test passing on Windows

ParkMyCar:
Turns out this fix isn't perfect, we still hit character limits on Windows. Specifically if the crate you're adding a doc test for has a large dependency tree you can hit `command is longer than CreateProcessW's limit (32767 characters)`. I thought maybe separating the args in the `.bat` or `.sh` file by newlines would help but I don't think it did

illicitonion:

It may be worth taking a look at https://github.com/bazelbuild/rules_rust/pull/1076 for inspiration in terms of the limits - that PR ended up working out how to _always_ use param files and I think mostly get things working out including for very long arg lists.

/cc @ddeville

ParkMyCar:
Bump on this @illicitonion, I updated the PR if you have a chance to review :)

tools/rustdoc/rustdoc_test_writer.rs
```diff
@@ -19,6 +20,9 @@ struct Options {
     /// The path where the script should be written.
     output: PathBuf,
 
+    /// The path where we can copy the params file Bazel might generate
+    optional_params_file: PathBuf,
```
illicitonion, line 24:
```suggestion
    /// If Bazel generated a params file, we may need to strip roots from it.
    /// This is the path where we will output our stripped params file.
    optional_output_params_file: PathBuf,
```

illicitonion (commented):
LGTM - thanks for the contribution!

Just one naming/commenting comment :)

illicitonion (approved):
Thanks!

ParkMyCar:
Thanks for merging!

PR #1419, opened 2022-06-23
Allow tree artifacts as srcs inputs to rustc compilation

chancila:


bsilver8192:
I noticed this in the open PRs list, and for reference I've integrated protobuf by running build.rs scripts that use prost-build and tonic-build directly via `cargo_build_script`. This packages the outputs from `$OUT_DIR` up into a directory artifact already. I don't see anything wrong with adding the flexibility from this PR, just figured I'd add an alternative approach to using the same packages for anybody who sees this.

Issue #1120, opened 2022-02-03
Systematically handle MAX_PATH issues


UebelAndre:
I feel like I run into [Windows MAX_PATH](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd) issues more often in `rules_rust` than I do for other rules. I'm wondering if there's a way to address this in a systematic way using the [process_wrapper](https://github.com/bazelbuild/rules_rust/tree/5225e23c9f545c2e990691afcdb6744927a49eeb/util/process_wrapper) or something. But it'd be nice to not need to worry about this since there's apparently solutions (`\\?\` prefixes) that might be able to help.

hlopko:
So if I understand right - the issue is that rules_rust produce paths that are too long on Windows. We have multiple solutions:

1) decreasing the chance of this being a problem by using shorter names (for example by using `crate_universe_examples` instead of `rules_rust_examples_crate_universe`)
2) implementing support for `\\?\` in the process wrapper
3) implementing support for `\\?\` in starlark implementation of rules

Let's use this issue to collect more data first. Which paths are problematic? Tools? Files?

UebelAndre:
https://github.com/bazelbuild/rules_rust/pull/1006 was another example of this issue

UebelAndre:
https://github.com/bazelbuild/rules_rust/issues/423 is also an example of this issue.

scentini:
https://github.com/bazelbuild/rules_rust/pull/1434 shortened a target name for the same reason too.

So far, whenever I've ran into this issue, the culprit turned out to be `MSVC`, especially the linker, and I don't believe we can do anything reasonable about that (maybe in process_wrapper we could copy all the files to a temporary location and copy the output file to the desired location afterwards, but that would be expensive). Bazel's [best practices](https://docs.bazel.build/versions/main/windows.html#avoid-long-path-issues) suggest using a short `output_user_root` but we already do that on CI and are still close to the limit even for some toolchain files.

I believe one of the reasons we are hitting this issue sooner than other rulesets due to the temporary files that `rustc` produces, eg
```
bazel build --spawn_strategy=local  --//:extra_rustc_flags=-Csave-temps test/unit/native_deps:proc_macro_has_native_dep
```
[...]

PR #1420, opened 2022-06-24
[WIP] Proc macro dep

krasimirgg:


UebelAndre:
> But I realised I don't know enough about the crate_universe infrastructure to try this out, e.g., the .bzl files under https://github.com/bazelbuild/rules_rust/tree/main/crate_universe/3rdparty/crates are auto-generated and contain references to `crate_universe_crate_index`, and I don't know how to re-generate them.

There's a `crates_vendor` rule that generates the target. I can update the comment at the top of each file to reflect this. But otherwise you can run:

```shell
bazel run //crate_universe/3rdparty:crates_vendor
```

This should update the generated files without changing any of the pins.

krasimirgg:
> > But I realised I don't know enough about the crate_universe infrastructure to try this out, e.g., the .bzl files under https://github.com/bazelbuild/rules_rust/tree/main/crate_universe/3rdparty/crates are auto-generated and contain references to `crate_universe_crate_index`, and I don't know how to re-generate them.
> 
> There's a `crates_vendor` rule that generates the target. I can update the comment at the top of each file to reflect this. But otherwise you can run:
> 
> ```shell
> bazel run //crate_universe/3rdparty:crates_vendor
> ```
> 
> This should update the generated files without changing any of the pins.
> 
> I filed:
> 
> * [Updated comments in crate_universe generated files to describe how to regenrate them. #1446](https://github.com/bazelbuild/rules_rust/issues/1446)

This is awesome! Thank you very much!

krasimirgg:
> re: [#1420 (comment)](https://github.com/bazelbuild/rules_rust/pull/1420#issuecomment-1180371623)
> 
> I think that'd be acceptable but unfortunate. The fundamental problem (#1120) is not solved in the rename and could definitely come back if an external crate just happens to have a long path somewhere. Is there anyone on or around your team that has any amount of windows experience to take a look at the root issues. I know there are ways to tell windows to ignore `MAX_PATH` constraints using `\\?\` but I'm not sure what all is needed to get the system to correctly understand the intention and not just see a funny looking path.

We will follow-up on #1120. It seems like there are multiple separate issues that contribute the problems with long paths on Windows and addressing all of them doesn't seem likely in the near future.

BUILD.bazel
```diff
@@ -45,6 +45,15 @@ clippy_flags(
     visibility = ["//visibility:public"],
 )
 
+# This setting may be used to identify dependencies of proc-macro-s.
+# Its value controls the BAZEL_RULES_RUST_IS_PROC_MACRO_DEP environment variable
+# made available to the rustc invocation.
+is_proc_macro_dep(
```
scentini, line 51: I'd add a comment that one should not depend on this flag or the environment variable, in case having the transition as a default proves to be too expensive and we need to add an additional `is_bootstrap` flag that will control whether this flag is relevant or not.

scentini (approved):

rust/private/rust.bzl
```diff
@@ -866,10 +867,31 @@ rust_shared_library = rule(
...
+    implementation = _proc_macro_dep_transition_impl,
+)
+
 rust_proc_macro = rule(
     implementation = _rust_proc_macro_impl,
     provides = _common_providers,
-    attrs = dict(_common_attrs.items()),
+    attrs = dict(
+        _common_attrs.items(),
+        _allowlist_function_transition = attr.label(default = Label("//tools/allowlists/function_transition_allowlist")),
```
UebelAndre, line 884: curious how this works, but is good to know about

rust/private/rust.bzl
```diff
@@ -685,6 +685,7 @@ _common_attrs = {
         default = Label("//util/import"),
         cfg = "exec",
     ),
+    "_is_proc_macro_dep": attr.label(default = Label("//:is_proc_macro_dep")),
```
UebelAndre, line 688: nit: make this consistent with the lines above

UebelAndre (approved):
I think the change looks fine but am over-all nervous about introducing a transition here. I don't have any concrete issue from this though so fingers crossed, and thanks!

rust/private/rust.bzl
krasimirgg: Thanks! Updated.

rust/private/rust.bzl
krasimirgg: This is doing a couple of things and deserved a comment; added a comment about it.

BUILD.bazel
krasimirgg: After chatting f2f, went ahead and added the new `is_proc_macro_dep_enabled` knob to control whether this feature is enabled (resulting in a no-op transition if is_proc_macro_dep_enabled is false, which is the default.

scentini:
Thanks! With the latest update my concern about building some targets 3 instead of 2 times goes away and I think we're ready to land this.

PR #1424, opened 2022-06-28
Do not add proc-macro deps to transitive_link_search_paths

krasimirgg:
While working on https://github.com/bazelbuild/rules_rust/pull/1420, which added a user-defined transition on the `deps` attribute of `rust_proc_macro`, I was hitting a strange `//crate_universe:rustdoc_test` failure:
https://buildkite.com/bazel/rules-rust-rustlang/builds/6631#01819625-365a-42fd-bb11-f3a95db9f999/280-523
```
Executing tests from //crate_universe:rustdoc_test
thread 'main' panicked at 'process wrapper error: No such file or directory (os error 2)', util/process_wrapper/main.rs:54:21
```

It took me a while to figure out exactly what was going on, since the process_wrapper error doesn't mention the path to the missing file nor the context in which the failure had occurred.

The root cause of the issue was that link_search_path's for (transitive) dependencies of proc-macro's were being added to the process_wrapper invocation. When I changed the `cfg` of `rust_proc_macro.deps`, these were pointing to a configuration-specific relative path starting with blaze-out/, like 
`bazel-out/k8-opt-exec-2B5CBBC6-ST-fa1bc2ca6f16/bin/external/crate_universe_crate_index__proc-macro2-1.0.39/proc-macro2_build_script.linksearchpaths`, which doesn't exist at test time (the "standard" deps link_search_path's are passed relative to `bin/`, e.g., `external/crate_universe_crate_index__blah-1.0.39/blah_build_script.linksearchpaths`.

I updated rules_rust to not add the link_search_paths for transitive deps of proc_macros and added some error logging to process_wrapper so errors like these are more evident. I believe this is reasonable since proc_macro's and their deps are not attempted to be linked into their clients, instead rustc dynamically loads them while compiling clients.

Required by https://github.com/bazelbuild/rules_rust/pull/1420. Without this change, that PR fails tests, so that will act as a test for this in practice.


UebelAndre:
Can you expand on the need for the transition? Why can't this be solved by some filtering exclusively within starlark-land?

scentini (approved):
Thanks for adding some tests around link_search_path_files !!

PR #1425, opened 2022-06-30
crate_universe: Render manifests with starlark

### LINKED ISSUE #1287: crate_universe: Rewrite generator in starlark
This is just an idea, I'd really appreciate some feedback and ideas.

Is it conceivably possible to rewrite the generator in pure starlark? This would remove the bootstrapping problem of distributing generator binaries or hackily building the generator with the native cargo toolchain.

https://github.com/bazelbuild/rules_rust/blob/59fab4e79f62bfa13551ac851a40696c24c0c3a4/crate_universe/deps_bootstrap.bzl#L9

We're using a custom generator to work around some issues, but this makes the analysis phase almost a minute slower as repository rules are not remotely cacheable.

```
Finished release [optimized] target(s) in 38.77s
```

Failing a full rewrite, some more ideas:
- Rewrite only some functionality in starlark.
- Move some functionality out of the repository rule and into just a normal rule. This could be achieved with a macro. The normal rule could then check if repinning is required, whilst allowing the repository rule to generate the build files it needs.

Thanks for taking the time to read this :) Please don't hesitate to ask any questions.

uhthomas:
Fixes #1287


UebelAndre (changes requested):
Hey, I haven't done a real pass through this code yet but did see a ton of deletion I didn't expect. Why wouldn't this change initally be a new `crates_vendor` mode that writes the "Bazel lock files" and uses the new functionality here to render it into repository rules? It seems far less invasive and give folks a chance to easily transition their setups.

uhthomas:
> Hey, I haven't done a real pass through this code yet but did see a ton of deletion I didn't expect. Why wouldn't this change initally be a new `crates_vendor` mode that writes the "Bazel lock files" and uses the new functionality here to render it into repository rules? It seems far less invasive and give folks a chance to easily transition their setups.

Hmm, I suppose it makes sense to keep the existing functionally in addition to this?

The approach taken in this change differs dramatically as the build files aren't rendered, but instead interpreted which makes it fundamentally incompatible with the existing approach. Maintaining both would be tedious and strange. 

I'm not at a computer at the minute so I can't give an example; the build files for the repository rules are almost all identical. They contain a snippet of the Cargo Bazel manifest and call a macro to interpret said snippet. **Vendoring** these would be pointless.

I mentioned this in the original issue: it is possible to see the result of the macros with `bazel query --output=build @crate_universe_crate_index//... or bazel query --output=build @crate_universe_crate_index__toml-0.5.9//...`. Might be helpful?

Happy to discuss and make changes where necessary :) Some work still needs to be done for feature parity.

UebelAndre:
> The approach taken in this change differs dramatically as the build files aren't rendered, but instead interpreted which makes it fundamentally incompatible with the existing approach. Maintaining both would be tedious and strange.

Perhaps I don't fully understand the changes being made. My current understanding of the fundamental change is this is a new set of functionality that *could* be added along side the existing functionality and things be deleted in a second step (making the initial implementation is less noisy). Could you link some updated docs that demonstrate the new workflow?

UebelAndre:
> I'm on holiday this week unfortunately

Nothing unfortunate about being on holiday, enjoy the time off 😄.

I think this change needs more design discussion but understand the desire to remove the need for the `cargo-bazel` binary in the analysis phase. When you're back we can continue the conversation.

mering:
@uhthomas @UebelAndre Has there been any progress outside of this PR in the past two years?

PR #1430, opened 2022-07-04
Fix rustdoc_test

scentini:


scentini:
Actual fix: https://github.com/bazelbuild/rules_rust/pull/1434

PR #1440, opened 2022-07-08
Update `//bindgen` to use `crate_universe`

UebelAndre:


illicitonion:
LGTM! Should this delete `bindgen/raze`?

UebelAndre:
> LGTM! Should this delete `bindgen/raze`?

Good catch, thanks!

PR #1444, opened 2022-07-10
Upgrade stardoc

bsilver8192:
I am eagerly awaiting https://github.com/bazelbuild/stardoc/pull/124 so
the anchor links work.

Looks like that's the only change to the generated files.


UebelAndre (approved):
Thanks!

PR #1452, opened 2022-07-12
Stamp only binaries by default

scentini:
Currently we pass `bazel-out/volatile-status.txt` to all the `Rustc` actions, which is bad for remote caching with `--stamp`enabled.

This PR makes only `rust_binary` be affected by the `--stamp` flag by default. The rest of the rust rules have `stamp = 0` as a default. Should you want to force stamp you can set `stamp = 1`. If you want to make your target stamp behavior depend on the `--stamp` command line flag, use `stamp = -1`.

Additionally, this PR makes `rules_rust` also use the stable keys for stamping, by passing `bazel-out/stable-status.txt` to the relevant `Rustc` actions when stamping is enabled. This makes `rules_rust` adhere to the contract described here: https://docs.bazel.build/versions/main/user-manual.html#flag--workspace_status_command


UebelAndre:
> Currently we pass `bazel-out/volatile-status.txt` to all the `Rustc` actions, which is bad for remote caching with `--stamp`enabled.

Isn't stamping in general bad for remote caching? In places I've used stamping I've explicitly enabled the remote cache by making the following config:

```
build:stamp --stamp
build:stamp --noremote_upload_local_results
```

Is that bad? Should I be trying to cache stamped builds? This always seemed like a losing battle to me.

All that aside, would this change continue to catch cases where I have a `rust_library` that includes some code I want to embed a commit hash into? Taking [//crate_universe:cargo_bazel_bin](https://github.com/bazelbuild/rules_rust/blob/0.7.0/crate_universe/src/main.rs) as an example, the code that consumes the commit hash is in a `rust_library` dependency and all the `rust_binary` is is a small entrypoint for the library. How would I embed the commit hash from workspace status there?

scentini:
`Isn't stamping in general bad for remote caching? `
I guess that depends on the stamping mechanism as well as the language compilation model.

C++ rules for example have a [linkstamp](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library.linkstamp) attribute to signify that the `.cc` file in question is using `stamp` data, which IIUC results in only that `.o` file being affected by changing values and the final linking action. The rest of the C++ actions will get cache hits with `--stamp`.

For Rust it is a bit more complex because the `.rlib`s depend on each other. You can mark the `rust_library` to be `stamp=1` and it will be affected by the changing values in `volatile-status.txt`. But then all the `rustc` compile actions from the reverse dependencies will also be reexecuted, as the input `.rlib`s will have changed. I think this PR still brings us to a better state than what we had before, because actions that don't transitively depend on the `rust_library.stamp=1` target will get cache hits.

For your last point -> I wanted to suggest that the `rust_library` that consumes the commit hash sets `stamp=1`. That however will override --nostamp so one will never get a cache hit on anything that depends on that library. A better solution seems to be a `config_setting` that checks whether we are in `--stamp` mode and sets `stamp=1` only when the answer is yes.
I believe the best course of action is to make that `rust_library` a `cc_library` instead: we do not pass `.o` files as inputs to the `rlib` compilation actions, only to the final linking action. That means even with `stamp=1` you'd get cache hits for most of the `Rustc` actions.

UebelAndre:
> For your last point -> I wanted to suggest that the `rust_library` that consumes the commit hash sets `stamp=1`. That however will override --nostamp so one will never get a cache hit on anything that depends on that library. A better solution seems to be a `config_setting` that checks whether we are in `--stamp` mode and sets `stamp=1` only when the answer is yes.
> I believe the best course of action is to make that `rust_library` a `cc_library` instead: we do not pass `.o` files as inputs to the `rlib` compilation actions, only to the final linking action. That means even with `stamp=1` you'd get cache hits for most of the `Rustc` actions.

I'm not sure I understand. I want the benefit of caching up until I recognize I want a release artifact that I'll distribute from Bazel. At that point I'll use stamping. But you're saying I should instead always force stamping? So when I ship my artifact I can benefit from stamping? Isn't that at the potential cost of cache hits if I were to use any stable status flags or at the cost of deterministic outputs if I have other unrelated commits since I last built?

UebelAndre:
re: https://github.com/bazelbuild/rules_rust/pull/1452#issuecomment-1184434508

I think adding stable status flags to `--stamp` functionality, setting the `stamp = 0` default on `rust_library` and `stamp = -1` on `rust_binary` is a great solution. Is that something you think could be done in this change?

UebelAndre:
> At Google we have multiple uses of stamping, not just for building release artifacts, which is why it's important for us to use the remote cache whenever it makes sense.

Curious to know what the uses are. Stamping, for me, has been more of a hindrance than a help in all cases except release artifacts.

UebelAndre (approved):
Awesome! Thank you so much and great work!

UebelAndre:
> As you pointed to [//crate_universe:cargo_bazel](https://github.com/bazelbuild/rules_rust/blob/0.7.0/crate_universe/BUILD.bazel#L57): Should I set `stamp=-1` over there in this PR?

Yeah, it can either be this PR or a followup if you don't think it makes sense here.

scentini:
I did it in this PR, so I don't forget about it :)

PR #1455, opened 2022-07-12
Added `rust_analyzer_toolchain` and repository rules for creating one

UebelAndre:
This change introduces a `rust_analyzer_toolchain` and additional repository rules for creating it from remote `rust-src` artifacts for use with [rust-analyzer rules](https://github.com/bazelbuild/rules_rust/blob/0.7.0/docs/rust_analyzer.md). By creating this separate toolchain, it becomes easier for users to avoid unknowingly impacting CI in a negative way when attempting to improve their developer workflow. 

A followup change will be made to remove the `rustc_srcs` attribute from `rust_toolchain`.


UebelAndre:
@hlopko @krasimirgg @scentini does this negatively impact any of your workflows?

hlopko:
Do I understand the problem correctly? Users want to use rust_analyzer locally, so they `rust_register_toolchains` in the WORKSPACE file to download Rust sources. However by doing so they also force downloading of Rust sources on the CI runs, where it is unnecessary.

I think the current recommended way of using rust_analyzer with rules_rust is to either put `--repo_env=RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS=true` into users bazelrc, or to put `--repo_env=RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS=false` into CI's bazelrc. This PR makes this unnecessary, and rust_analyzer works without further configuration, assuming that all rust_toolchains have a corresponding rust_analyzer_toolchain in the workspace (and this is where this PR makes a bit of churn for us - we don't use repository rules to get rust toolchains, we write `rust_toolchain` targets by hand and we maintain them for all projects centrally). 

So assuming you knew about `--repo_env=RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS=true/false` and it doesn't solve the problem you're having, I think this PR makes sense and can go in. I'll leave it to @krasimirgg to double check the implementation and manage the internal migration.

rust/defs.bzl
```diff
@@ -130,6 +131,9 @@ rust_common = _rust_common
 rust_analyzer_aspect = _rust_analyzer_aspect
 # See @rules_rust//rust/private:rust_analyzer.bzl for a complete description.
 
+rust_analyzer_toolchain = _rust_analyzer_toolchain
```
krasimirgg, line 134: I'm just curious why export `rust_analyzer_toolchain` here instead of via `rust/toolchain.bzl` which exports `rust_toolchain`?

rust/private/rust_analyzer.bzl
```diff
@@ -214,7 +235,10 @@ def _rust_analyzer_detect_sysroot_impl(ctx):
 
 rust_analyzer_detect_sysroot = rule(
     implementation = _rust_analyzer_detect_sysroot_impl,
-    toolchains = ["@rules_rust//rust:toolchain"],
+    toolchains = [
+        "@rules_rust//rust:toolchain",
+        "@rules_rust//rust/rust_analyzer:toolchain_type",
```
krasimirgg, line 240: why not name this `"@rules_rust//rust/rust_analyzer:toolchain_type"` for consistency with `"@rules_rust//rust:toolchain"`?

rust/private/toolchain_utils.bzl
```diff
@@ -42,8 +42,6 @@ def _toolchain_files_impl(ctx):
         )
     elif ctx.attr.tool == "rustc_lib":
         files = toolchain.rustc_lib
-    elif ctx.attr.tool == "rustc_srcs":
```
krasimirgg, line 45: just to clarify the effect of this PR: already after this gets submitted, in case a user has a `rust_toolchain` with `rustc_srcs`, they would be ignored by rust_analyzer (even before we drop rust_toolchain.rustc_srcs in a follow-up PR)?

krasimirgg (commented):
Overall looks good. I've got a few questions.

rust/defs.bzl
UebelAndre: I thought everything goes here. Do you think `rust_analyzer_toolchain` should be removed or `rust_toolchain` added?

rust/private/rust_analyzer.bzl
UebelAndre: I think `@rules_rust//rust:toolchain` is inconsistent with other Bazel rules ([rules_cc](https://github.com/bazelbuild/rules_cc/blob/0.0.2/cc/BUILD#L33-L36), [rules_python](https://github.com/bazelbuild/rules_python/blob/0.10.2/python/BUILD)). I care more that the community is consistent more than whatever the value in this PR. What would you recommend happen here?

rust/private/toolchain_utils.bzl
UebelAndre: This code is actually not quite correct. It does not handle cases where the `rust_toolchain` did not get `rustc_srcs`. Since code here is broken in what I perceive to be the common case I opted to remove it. The toolchain should still support fetching and consuming `rust-src` artifacts but after this change, those sources will not be used by rust-analyzer.

UebelAndre:
> Do I understand the problem correctly? Users want to use rust_analyzer locally, so they `rust_register_toolchains` in the WORKSPACE file to download Rust sources. However by doing so they also force downloading of Rust sources on the CI runs, where it is unnecessary.
> 
> I think the current recommended way of using rust_analyzer with rules_rust is to either put `--repo_env=RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS=true` into users bazelrc, or to put `--repo_env=RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS=false` into CI's bazelrc. This PR makes this unnecessary, and rust_analyzer works without further configuration, assuming that all rust_toolchains have a corresponding rust_analyzer_toolchain in the workspace (and this is where this PR makes a bit of churn for us - we don't use repository rules to get rust toolchains, we write `rust_toolchain` targets by hand and we maintain them for all projects centrally).
> 
> So assuming you knew about `--repo_env=RULES_RUST_TOOLCHAIN_INCLUDE_RUSTC_SRCS=true/false` and it doesn't solve the problem you're having, I think this PR makes sense and can go in. I'll leave it to @krasimirgg to double check the implementation and manage the internal migration.

The changes here are attempting to reduce the burden of knowledge on users and make integrations with `rust-analyzer` not require any understanding of where flags need to surgically enable this functionality for developers in a large repo where rust is a small subset of the code there. A consequence of messing up this surgery is everyone ends up having to download `rust-src` artifacts when they will never be used. Ideally, once a developer sets up their IDE to expect a `rust-project.json` then `rules_rust` should do all the heavy lifting for them "for free" at no cost to build times. Developers are free to run `@rules_rust//tools/rust_analyzer:gen_rust_project` and Bazel will fetch everything that's needed while in CI, the large `rust-src` artifact will not be downloaded as nothing depends on it.

UebelAndre:
Didn't mean to close this :(

rust/private/toolchain_utils.bzl
krasimirgg: thanks

rust/private/rust_analyzer.bzl
krasimirgg: Thanks! Then should we consider renaming `@rules_rust//rust:toolchain` to `@rules_rust//rust:toolchain_type` (as a follow-up)?

rust/defs.bzl
krasimirgg: I suggest removing `rust_analyzer_toolchain` from https://github.com/bazelbuild/rules_rust/blob/main/rust/defs.bzl and adding it to https://github.com/bazelbuild/rules_rust/blob/main/rust/toolchain.bzl.

I can't find a `rust_toolchain` in https://github.com/bazelbuild/rules_rust/blob/main/rust/defs.bzl. Instead, it's defined in https://github.com/bazelbuild/rules_rust/blob/main/rust/toolchain.bzl. Since the `rust_analyzer_toolchain` is a toolchain, it feels more consistent to add it alongside of the the `rust_toolchain` definition.

UebelAndre: Moved this to `rust/toolchain.bzl`

rust/private/rust_analyzer.bzl
UebelAndre: Created https://github.com/bazelbuild/rules_rust/pull/1479

krasimirgg (approved):
Looks good to me! Thank you!

Issue #1456, opened 2022-07-12
cargo-bazel will not build with newer versions of GCC


uhthomas:
There are many lines of

```
/root/.cache/bazel/_bazel_root/1e0bb3bee2d09d2e4ad3523530d3b40c/external/zig_sdk/lib/libc/include/generic-glibc/regexp.h:31:2: error: "Please update your code to use <regex.h> instead (no trailing 'p')."
17:34:04       #error "Please update your code to use <regex.h> instead (no trailing 'p')."
17:34:04        ^
17:34:04       In file included from external/libgit2/src/merge_file.c:14:
17:34:04       In file included from external/libgit2/src/diff_xdiff.h:13:
17:34:04       In file included from external/libgit2/src/xdiff/xdiff.h:26:
17:34:04       external/libgit2/src/xdiff/git-xdiff.h:40:8: error: unknown type name 'git_regexp'
17:34:04               const xdl_regex_t *preg, const char *buf, size_t size,
17:34:04                     ^
```

This looks to be because `@libgit2` is not correctly linking headers. A patch file resolves this:

```
diff --git a/crate_universe/3rdparty/BUILD.libgit2.bazel b/crate_universe/3rdparty/BUILD.libgit2.bazel
index 031ee78..38124c5 100644
--- a/crate_universe/3rdparty/BUILD.libgit2.bazel
+++ b/crate_universe/3rdparty/BUILD.libgit2.bazel
@@ -178,19 +178,28 @@ write_file(
     ],
 )
 
+cc_library(
+    name = "include",
+    hdrs = [":configure_features"] + glob(["include/**/*.h"]),
+    strip_include_prefix = "include",
+)
+
 cc_library(
     name = "git2",
-    srcs = glob(
+    srcs = [
+        "src/allocators/failalloc.c",
+        "src/allocators/stdalloc.c",
+        "src/allocators/win32_leakcheck.c",
+        # Use the CollisionDetection SHA1 implementation.
+        "src/hash/sha1/collisiondetect.c",
+        "src/hash/sha1/sha1dc/sha1.c",
+        "src/hash/sha1/sha1dc/ubc_check.c",
+    ] + glob(
         include = [
             "src/*.c",
-            "src/*.h",
-            "src/xdiff/*.h",
             "src/xdiff/*.c",
-            "src/transports/*.h",
             "src/transports/*.c",
-            "src/streams/*.h",
             "src/streams/*.c",
-            "src/hash/*.h",
[...]

PR #1457, opened 2022-07-12
Fix @libgit2

uhthomas:
The include prefix for "src/" was not stripped correctly.


uhthomas:
Fixed #1456

uhthomas:
Not sure. It could be some configuration in our .bazelrc which makes things worse.

It just seemed like the headers weren't linked and so would fall back to system headers. Specifically in the case of regexp.h it will error unconditionally.

https://sourceware.org/bugzilla/show_bug.cgi?id=18681

UebelAndre:

uhthomas:
Happy to help.

Issue #1466, opened 2022-07-14
Coverage includes reference to non-project file when using `thread_local!`


fmorency:
Minimal sample repository in https://github.com/fmorency/bazelcov

# Problem

Code coverage on a project using `thread_local!` will contain a reference to 
```
/rustc/[SOME_HASH]/library/std/src/thread
```

which results in an error when trying to generate the coverage report using `genhtml`

```
genhtml: ERROR: cannot read /rustc/[SOME_HASH]/library/std/src/thread/local.rs
```

# Proposed solution

Add an additional `-ignore-filename-regex` to the `llvm-cov` call in `collect_coverage.sh`, e.g.,
```
-ignore-filename-regex='/rustc/.+'
```

UebelAndre:
Cc @krasimirgg @hlopko @keith who might have better intuition here

keith:
Nice repro case, turns out this is also an issue without bazel with rust. I repro'd by copying your file into a fresh cargo project:

```
% /Users/ksmiley/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin/llvm-cov show target/debug/deps/threadtest-e0b0e269f87e678c -instr-profile=foo.profdata -show-line-counts-or-regions
/private/tmp/threadtest/src/main.rs:
    1|      1|use std::cell::RefCell;use std::cell::RefCell;
    2|       |
    3|      1|thread_local!{
    4|      1|    static BAR: RefCell<f32> = RefCell::new(1.0);
    5|      1|}
    6|       |
    7|      0|fn main() {
    8|      0|}
    9|       |
   10|      1|#[test]
   11|      1|fn test_p1() {
   12|      1|   BAR.with(|f| {
   13|      1|       assert_eq!(*f.borrow(), 1.0);
   14|      1|   })
   15|      1|}

error: /rustc/52ca603da73ae9eaddf96f77953b33ad8c47cc8e/library/std/src/thread/local.rs: No such file or directory
warning: The file '/rustc/52ca603da73ae9eaddf96f77953b33ad8c47cc8e/library/std/src/thread/local.rs' isn't covered.
```

I suppose if we could determine the sha, which appears to be the sha of rust the build was from, we could do another path-equivalence argument to normalize it, but I'm not sure to what, since the file not existing would still be a problem. The thing that surprises me is that the lcov file has symbols listed that are clearly the ones from the source file:

```
[...]

PR #1467, opened 2022-07-14
Limit coverage to requested files

keith:
Bazel provides COVERAGE_MANIFEST which is a file we can pass through to
llvm-cov to limit the files that we produce coverage for. Using this
revealed the path-equivalence was backwards because we're using the
coverage prefix map to disable absolute paths in code coverage info,
but llvm-cov operates on absolute paths, so we have to remap the `.` in
the coverage data to the full path so that llvm-cov understands the
relative paths (which it absolutizes internal) from the manifest file.
This feels a bit weird and ideally it could operate entirely on these
relative paths instead.

This has the side effect of excluding files that aren't part of bazel
sources, which fixes https://github.com/bazelbuild/rules_rust/issues/1466

This should also improve interaction with `--instrumentation_filter`.

### DIFF
--- a/util/collect_coverage.sh
+++ b/util/collect_coverage.sh
@@ -19,6 +19,7 @@ readonly profdata_file=$COVERAGE_DIR/coverage.profdata
   -instr-profile "$profdata_file" \
   -ignore-filename-regex='.*external/.+' \
   -ignore-filename-regex='/tmp/.+' \
-  -path-equivalence="$ROOT",. \
+  -path-equivalence=.,"$ROOT" \
   "$RUNFILES_DIR/$TEST_WORKSPACE/$TEST_BINARY" \
+  @"$COVERAGE_MANIFEST" \
   | sed 's#/proc/self/cwd/##' > "$COVERAGE_OUTPUT_FILE"


UebelAndre:
This file feels a bit cryptic to me. I'm happy to hear it's working but don't know enough about what it's doing to think about how to test it. Could this potentially be something that's replaced with a `rust_binary`? I would imagine the use of bash here adds an unnecessary requirement to windows hosts and also makes testing just a bit harder to author.

keith:
Currently with bazel's infra that reads this file I don't think we have a choice, unless we add a custom test runner which we talked about briefly on the other PR I think. It's also bash so realistically I guess it won't work on windows either way until bazel becomes more flexible for coverage.

I wonder if we could write an integration test that could transition on coverage and depend on a rust_test and validate the coverage file contents 🤔

keith:
Note as far as this change goes the important part is limiting the files to the ones bazel tells us. The order argument was wrong in the first place it was just being ignored until we started limiting the file.

scentini:
> I wonder if we could write an integration test that could transition on coverage and depend on a rust_test and validate the coverage file contents

This should be easily doable.

I would very much like to 
a) have tests for coverage in general before declaring it supported, as it is hard to see what is being fixed here; or
b) make it guarded by an experimental flag as long as a) is not the case.

krasimirgg (approved):
The fixes look good, +1 to coming up with coverage tests robust enough to be able to verify fixes like these (possibly as follow-ups).

UebelAndre (approved):

fmorency:
I tested my project, and everything worked as expected. Thanks a lot 👍🏻

keith:
I tried to spike out a quick test, it doesn't seem like there is any way to access the coverage.dat file from dependencies of a test, I was envisioning something like:

```bzl
rust_test(
    name = "foo",
...
)

coverage_validation_test(
     name = "foo.coverage_validation",
     target = "foo",
)
```

but in this test's implementation I don't see a way to get my hands on that file to do something about it.

PR #1474, opened 2022-07-20
Update rules to use a helper to declare the cpp toolchain dependency.

katre:


katre:
The `use_cpp_toolchain` function was added in Bazel 5.1, it looks like the min version test is at Bazel 5.0.

rust/private/utils.bzl
```diff
@@ -55,6 +55,17 @@ def find_cc_toolchain(ctx):
...
+def use_cc_toolchain(mandatory = False):
+    """Helper to depend on the cc toolchain.
+
+    Args:
+        mandatory: Whether or not it should be an error if the toolchain cannot be resolved.
+
+    Returns:
+        A list that can be used as the value for `rule.toolchains`.
+    """
+    return use_rules_cc_toolchain(mandatory)
```
UebelAndre, line 67: What is the benefit in using this function? It feels like this change would be better to hold until `6.0` is the min tested version.

UebelAndre (changes requested):

rust/private/utils.bzl
katre: This is part of a project to make the CC toolchain optional, since many users cannot currently use toolchain resolution. It's currently causing us problems on the internal side and so I'd like to get this through, but if the rules_rust maintainers want to hold off I'll look into workarounds.

UebelAndre: I would prefer that `5.0` remain the min tested version as it's a current LTS release but if this is a major issue then I can concede it's not a huge deal. Though, work to support something outside of toolchain resolution feels like negative work and a burden to maintainers.

katre: Switching to `use_cpp_toolchain` is the currently recommended way to declare a dependency on the CC toolchain, but you're right that it's not yet in an LTS release. I'm fine with closing this PR and figuring out a workaround, but this will be a change that should be made after the next LTS release.

katre:
This is part of a project to make the CC toolchain optional, since many
users cannot currently use toolchain resolution.

UebelAndre:
@katre If this change is still useful, I think now would be an appropriate time to adopt it.

PR #1475, opened 2022-07-20
Give useful error on missing workspace manifest

illicitonion:
Previously if the workspace root manifest was missing, a panic would be
triggered due to unwrapping a `None` option.

Now, we suggest the missing manifest if possible, or direct you to work
it out yourself if we can't.


UebelAndre (approved):
One nit but otherwise looks good to me, thanks!

PR #1478, opened 2022-07-21
Fix typo in function name

gferon:
Just found a small typo while exploring the insides of `rules_rust`.


illicitonion (approved):
Thanks for the fix!

PR #1479, opened 2022-07-21
Updated all toolchain_type definitions to be named `toolchain_type`.

UebelAndre:
Upstream Bazel rules seem to consistently define [toolchain_type](https://docs.bazel.build/versions/main/toolchains.html#writing-rules-that-use-toolchains) names as `toolchain_type` ([rules_cc](https://github.com/bazelbuild/rules_cc/blob/0.0.2/cc/BUILD#L33-L36), [rules_python](https://github.com/bazelbuild/rules_python/blob/0.10.2/python/BUILD)). The changes here move `rules_rust` to be consistent with upstream Bazel.

This will be a breaking change for any users who index toolchains via the string of the label of the toolchain type in their rules.

### DIFF
--- a/bindgen/BUILD.bazel
+++ b/bindgen/BUILD.bazel
@@ -10,6 +10,7 @@ toolchain_type(
 alias(
     name = "bindgen_toolchain",
     actual = "toolchain_type",
+    deprecation = "instead use `@rules_rust//bindgen:toolchain_type`",
     tags = ["manual"],
 )
 
--- a/bindgen/BUILD.bazel
+++ b/bindgen/BUILD.bazel
@@ -3,7 +3,15 @@ load("//bindgen:bindgen.bzl", "rust_bindgen_toolchain")
 
 package(default_visibility = ["//visibility:public"])
 
-toolchain_type(name = "bindgen_toolchain")
+toolchain_type(
+    name = "toolchain_type",
+)
+
+alias(
+    name = "bindgen_toolchain",
+    actual = "toolchain_type",
+    tags = ["manual"],
+)
 
 bzl_library(
     name = "bzl_lib",
@@ -43,5 +51,5 @@ rust_bindgen_toolchain(
 toolchain(
     name = "default_bindgen_toolchain",
     toolchain = "default_bindgen_toolchain_impl",
-    toolchain_type = "//bindgen:bindgen_toolchain",
+    toolchain_type = "//bindgen:toolchain_type",
 )
--- a/bindgen/bindgen.bzl
+++ b/bindgen/bindgen.bzl
@@ -92,7 +92,7 @@ def _rust_bindgen_impl(ctx):
     if header not in cc_header_list:
         fail("Header {} is not in {}'s transitive headers.".format(ctx.attr.header, cc_lib), "header")
 
-    toolchain = ctx.toolchains[Label("//bindgen:bindgen_toolchain")]
+    toolchain = ctx.toolchains[Label("//bindgen:toolchain_type")]
     bindgen_bin = toolchain.bindgen
     rustfmt_bin = toolchain.rustfmt or rust_toolchain.rustfmt
     clang_bin = toolchain.clang
@@ -214,8 +214,8 @@ rust_bindgen = rule(
     outputs = {"out": "%{name}.rs"},
     fragments = ["cpp"],
     toolchains = [
-        str(Label("//bindgen:bindgen_toolchain")),
-        str(Label("//rust:toolchain")),
+        str(Label("//bindgen:toolchain_type")),
+        str(Label("//rust:toolchain_type")),
         "@bazel_tools//tools/cpp:toolchain_type",
     ],
     incompatible_use_toolchain_transition = True,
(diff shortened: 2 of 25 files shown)


krasimirgg (approved):
This looks good, thank you!

Not sure about the policy around such breaking changes, whether we should bump rules_rust version...

UebelAndre:
> This looks good, thank you!
> 
> Not sure about the policy around such breaking changes, whether we should bump rules_rust version...

we definitely will bump the version but I'd expect that to be a separate commit.

@scentini does this sound good to you too? Currently [rules_go](https://github.com/bazelbuild/rules_go/blob/v0.34.0/go/BUILD.bazel#L26-L29) and [rules_kotlin](https://github.com/bazelbuild/rules_kotlin/blob/v1.6.0/kotlin/internal/toolchains.bzl#L365-L368) define different names. Is `toolchain_type` something we should try to standardize and make it consistent with the built-in bazel rules? I'm definitely a fan of consistency and don't really care what the toolchain types are called as long as they're distinguishable from each other.

scentini:
Bazel does seem to indicate that `toolchain_type` is the convention: https://docs.bazel.build/versions/main/toolchains.html#writing-rules-that-use-toolchains

>  By convention, toolchain_type targets are named "toolchain_type" and
    distinguished by their package path. So the full path for this would be
   `//bar_tools:toolchain_type.`

Given that you also create `alias`es to the old names, is this really a breaking change? We could also add a deprecation message to those so that the users are informed about the upcoming breaking change of removing them.

scentini:
Also, maybe we should consider providing a `use_rust_toolchain()` function similar to katre's PR https://github.com/bazelbuild/rules_rust/pull/1474. If we require Starlark rule writers to use such a function when defining rules/aspect, changes like this one should not affect them.

UebelAndre:
> Given that you also create `alias`es to the old names, is this really a breaking change? We could also add a deprecation message to those so that the users are informed about the upcoming breaking change of removing them.

The alias works to support changes like this where users defined rules with the alias toolchain type:
```diff
diff --git a/rust/private/clippy.bzl b/rust/private/clippy.bzl
index 21f39656..ec94ae76 100644
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -226,7 +226,7 @@ rust_clippy_aspect = aspect(
     },
     provides = [ClippyInfo],
     toolchains = [
-        str(Label("//rust:toolchain_type")),
+        str(Label("//rust:toolchain")),
         "@bazel_tools//tools/cpp:toolchain_type",
     ],
     incompatible_use_toolchain_transition = True,
```

but does not help changes like this where users are indexing toolchains
```diff
diff --git a/rust/private/utils.bzl b/rust/private/utils.bzl
index 4ae5b5fc..b3e8e809 100644
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -34,7 +34,7 @@ def find_toolchain(ctx):
     Returns:
         rust_toolchain: A Rust toolchain context.
     """
-    return ctx.toolchains[Label("//rust:toolchain_type")]
+    return ctx.toolchains[Label("//rust:toolchain")]

 def find_cc_toolchain(ctx):
     """Extracts a CcToolchain from the current target's context
```

```
INFO: Invocation ID: cda0dfd4-f752-4759-bede-89db058aa412
ERROR: /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rules_rust_tinyjson/BUILD.bazel:4:37: in rust_library_without_process_wrapper rule @rules_rust_tinyjson//:tinyjson:
	File "/Users/user/Code/rules_rust/rust/private/rust.bzl", line 236, column 32, in _rust_library_impl
		return _rust_library_common(ctx, "rlib")
	File "/Users/user/Code/rules_rust/rust/private/rust.bzl", line 298, column 31, in _rust_library_common
	File "/Users/user/Code/rules_rust/rust/private/utils.bzl", line 37, column 26, in find_toolchain
		return ctx.toolchains[Label("//rust:toolchain")]
Error: In rust_library_without_process_wrapper rule @rules_rust_tinyjson//:tinyjson, toolchain type //rust:toolchain was requested but only types [//rust:toolchain_type, @bazel_tools//tools/cpp:toolchain_type] are configured
ERROR: /private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/external/rules_rust_tinyjson/BUILD.bazel:4:37: Analysis of target '@rules_rust_tinyjson//:tinyjson' failed
ERROR: Analysis of aspects '[//rust:defs.bzl%rust_clippy_aspect] on //crate_universe/private:bzl_srcs' failed; build aborted:
INFO: Elapsed time: 4.092s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (146 packages loaded, 4459 targets\
FAILED: Build did NOT complete successfully (146 packages loaded, 4459 targets\
 configured)
```

scentini:
Ah, I missed the indexed toolchains part, thanks for explaining.

> I'm not sure I fully understand the benefit of that change. Is that something that can be deferred to a different PR?
Definitely! I don't think we can make this PR backwards compatible, so I'm fine with it being merged as is.

My idea (and what I believe katre@ is doing the other PR) is that if we require users to never spell out the label, one would be able to make changes such as this PR, or changing the location of the toolchains for example without breaking the users.
We already have a `find_toolchain` function, which we should probably advertise as **the** way to obtain the toolchain. Had we required that the rule/aspect writers use
```
toolchains = [
    use_rust_toolchain(),
],
```
instead of the current

```
toolchains = [
    str(Label("//rust:toolchain_type")),
],
```
this PR would not be backwards incompatible.

scentini (approved):

Issue #1480, opened 2022-07-24
`rust_analyzer` is not downloaded/installed for the workspace


mingyc:
I have a `WORKSPACE` file with the following settings:

```Starlark
# Ignored setup ...

# https://bazelbuild.github.io/rules_rust/#setup
load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_analyzer_toolchain_repository", "rust_register_toolchains")

rules_rust_dependencies()

rust_register_toolchains(
    edition = "2021",
    version = "1.58.0",
)

# https://bazelbuild.github.io/rules_rust/rust_analyzer.html
load("@rules_rust//tools/rust_analyzer:deps.bzl", "rust_analyzer_dependencies")

rust_analyzer_dependencies()
```

After I ran the command `bazel run @rules_rust//tools/rust_analyzer:gen_rust_projec`, it produces a `rust-project.json` with the following `sysroot_src`:

```
"/home/ming/.cache/bazel/_bazel_ming/9c28e31a82d7db73ccf36bd435554ded/execroot/__main__/external/rust_analyzer_1.58.0_srcs/lib/rustlib/src/library"
```

However, `rust-analyzer` complains that it fails to load workspace with

```
rust-analyzer failed to load workspace: could not find libcore in sysroot path
```

and when I looked into `execroot/__main__/external` it doesn't contain rust-analyzer but only

```
bazel_tools  local_config_cc  rules_rust  rules_rust_tinyjson  rust_linux_x86_64__x86_64-unknown-linux-gnu_tools
``` 

I suspect that `rust_register_toolchains` doesn't really download rust-analyzer. I also tried to look into its implementation at [lines](https://github.com/bazelbuild/rules_rust/blob/81a77acde2bf0d9a3d248806baf7ba19641135b7/rust/repositories.bzl#L139-L153). Looks like it tries to download a binary of `rust_analyzer_1.58.0_srcs` using `rust_analyzer_toolchain_repository` and was implemented recently in #1455. Is there a bug?

UebelAndre:
The rules have never downloaded a [rust-analyzer](https://github.com/rust-lang/rust-analyzer) release before. Do you have cargo installed on your host? I'm wondering if I'm not seeing this issue because rust-analyzer is still able to find parts of the stdlib.

omerbenamram:
I think this is indeed a bug.

[...]

PR #1483, opened 2022-07-25
Fix rust-analyzer being unable to find rust sysroot sources.

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/1480


tools/rust_analyzer/main.rs
```diff
@@ -26,6 +26,11 @@ fn main() -> anyhow::Result<()> {
         .as_ref()
         .expect("failed to find execution root, is --execution-root set correctly?");
 
+    let output_base = config
+        .output_base
+        .as_ref()
+        .expect("failed to find execution root, is -output-base set correctly?");
```
illicitonion, line 32:
```suggestion
        .expect("failed to find output base, is -output-base set correctly?");
```

illicitonion (approved):

PR #1485, opened 2022-07-26
feat: Add `capture_output` attr to `rust_doc` rule

ParkMyCar:
This PR adds a `capture_output` attr to the `rust_doc` rule, which if set to `True` will redirect the output of `stderr` into a file with the suffix of `.rustdoc.out`.

Adding this unlocks the ability for users to define a bazel test rule that asserts their docs build without warnings. This is useful because it allows you to semi-automate the process of making sure your docs are up to date.

### DIFF
--- a/docs/flatten.md
+++ b/docs/flatten.md
@@ -451,8 +451,8 @@ rust_clippy(
 ## rust_doc
 
 <pre>
-rust_doc(<a href="#rust_doc-name">name</a>, <a href="#rust_doc-crate">crate</a>, <a href="#rust_doc-html_after_content">html_after_content</a>, <a href="#rust_doc-html_before_content">html_before_content</a>, <a href="#rust_doc-html_in_header">html_in_header</a>, <a href="#rust_doc-markdown_css">markdown_css</a>,
-         <a href="#rust_doc-rustc_flags">rustc_flags</a>)
+rust_doc(<a href="#rust_doc-name">name</a>, <a href="#rust_doc-capture_output">capture_output</a>, <a href="#rust_doc-crate">crate</a>, <a href="#rust_doc-html_after_content">html_after_content</a>, <a href="#rust_doc-html_before_content">html_before_content</a>, <a href="#rust_doc-html_in_header">html_in_header</a>,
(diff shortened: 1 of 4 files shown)


ParkMyCar:
Hey @illicitonion, any chance you could review this PR too? No rush though :)

P.S. I tried to assign a reviewer but I don't seem to be able to

scentini (commented):
Could we add a test please?

UebelAndre:
> Adding this unlocks the ability for users to define a bazel test rule that asserts their docs build without warnings. This is useful because it allows you to semi-automate the process of making sure your docs are up to date.

Does `rustdoc` really not have a flag to promote warnings to errors? It sounds like that's what you'd want instead of needing to parse a bunch of text in a bespoke action.

ParkMyCar:
> Does `rustdoc` really not have a flag to promote warnings to errors? It sounds like that's what you'd want instead of needing to parse a bunch of text in a bespoke action.

It does, you can pass `-Dwarnings` to deny all warnings. The issue though is the `rust_doc` rule is not declared as a test rule, so even if your rustdoc build fails, running `bazel test //rustdoc_target` still passes.

I would love if there was a way to make a bazel test whose predicate is "test whether or not this other build target succeeds", but AFAICT there isn't.

UebelAndre:
> It does, you can pass `-Dwarnings` to deny all warnings. The issue though is the `rust_doc` rule is not declared as a test rule, so even if your rustdoc build fails, running `bazel test //rustdoc_target` still passes.
> 
> I would love if there was a way to make a bazel test whose predicate is "test whether or not this other build target succeeds", but AFAICT there isn't.

Don't build targets get built under `bazel test //...` unless you're using [--build_tests_only](https://bazel.build/reference/command-line-reference#flag--build_tests_only)? Also, if you want a build target to be guaranteed to run in a test, couldn't you use [@bazel_skylib//rules:build_test.bzl](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/build_test_doc.md)?

ParkMyCar:
Stale

PR #1486, opened 2022-07-27
toolchain files: ensure test depends on std

krasimirgg:

Among the standard rust libs, libtest depends on libstd:
https://github.com/rust-lang/rust/blob/master/library/test/Cargo.toml#L12

The current implementation was adding it to
`between_core_and_std_files`. This may cause backward reference linker
errors in some configurations where the linker is configured to check
for such backward references and the toolchain uses the static version
of libtest.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -14,6 +14,7 @@ def _rust_stdlib_filegroup_impl(ctx):
     core_files = []
     between_core_and_std_files = []
     std_files = []
+    test_files = []
     memchr_files = []
     alloc_files = []
     self_contained_files = [
@@ -24,7 +25,8 @@ def _rust_stdlib_filegroup_impl(ctx):
 
     std_rlibs = [f for f in rust_std if f.basename.endswith(".rlib")]
     if std_rlibs:
-        # std depends on everything
+        # test depends on std
+        # std depends on everything except test
         #
         # core only depends on alloc, but we poke adler in there
         # because that needs to be before miniz_oxide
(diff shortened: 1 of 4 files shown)


rust/toolchain.bzl
```diff
@@ -198,10 +202,18 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_std, allocator_library):
...
+                for f in rust_stdlib_info.test_files
+            ],
+            transitive = [std_inputs],
+            order = "topological",
+        )
 
         link_inputs = cc_common.create_linker_input(
             owner = rust_std.label,
-            libraries = std_inputs,
+            libraries = test_inputs,
```
UebelAndre, line 216: Feels minor but should we instead track `link_inputs` and `test_link_inputs` separately?

UebelAndre (approved):
Seems fine to me, thanks! I just had one kinda general question about the change.

rust/toolchain.bzl
krasimirgg: I considered this a bit, but I think the way it is now is more consistent with the treatment of other groups of inputs above (e.g., `between_code_and_std_inputs` is immediately folded as a transitive dep of `std_inputs` and not mentioned afterwards). Also,  if we separate them here, we'd have to merge them at some later point, and then the merged result will effectively be the same as `test_inputs` since `std_inputs` is already a transitive dep of `test_inputs`.

UebelAndre: Sounds reasonable enough, we can split them later if it becomes an issue I suppose. Thanks again!

PR #1487, opened 2022-07-27
Merge runfiles from transitive dependencies of rust_test crate attr

neilisaac:
Hopefully resolves https://github.com/bazelbuild/rules_rust/issues/266

I'm open to input from maintainers on how to do this.


rust/private/rustc.bzl
```diff
@@ -1046,6 +1046,9 @@ def rustc_compile_action(
         files = getattr(ctx.files, "data", []) + coverage_runfiles,
         collect_data = True,
```
neilisaac, line 1047: https://docs.bazel.build/versions/main/skylark/lib/ctx.html#runfiles and https://docs.bazel.build/versions/main/skylark/rules.html#runfiles-features-to-avoid discourage use of `collect_data`

UebelAndre: Do you want to attempt to remove it in this change? Or an isolated change?

scentini (commented):
Thanks! Could we also get some tests?

rust/private/rustc.bzl
neilisaac: I'm not really sure I understand the implications of using "collect_data" here, so I left it alone.

neilisaac:
@scentini this is the common case that was broken for me: https://github.com/bazelbuild/rules_rust/pull/1487/commits/17a00a30fd8066d415085bcd167993a615147edc

test/rust/src/greeter.rs
```diff
@@ -72,4 +87,10 @@ mod test {
         let hello = Greeter::new("Hi");
         assert_eq!("Hi Rust", hello.greeting("Rust"));
     }
+
+    #[test]
+    fn test_greeting_from_txt_file() {
+        let welcome = Greeter::from_txt_file().unwrap();
+        assert_eq!("Welcome Rust", welcome.greeting("Rust"));
```
neilisaac, line 94: backporting this commit to main, I see:
```
thread 'greeter::test::test_greeting_from_txt_file' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', test/rust/src/greeter.rs:93:48

```

scentini:
Hey @neilisaac, this PR looks good to me, however looks like CI is complaining about `.rs` file formatting, could you please fix it?

neilisaac:
@scentini checks passed!

scentini (approved):

scentini:
Thanks!

PR #1489, opened 2022-07-28
feat: Add `crate_features` attr to `rust_doc` rule

ParkMyCar:
This PR adds the `crate_features` attr to the `rust_doc` rule, which allows you to specify what features should be enabled when building docs. Currently `rust_doc` doesn't enable any features (as far as I can tell, not even default features) which makes it impossible to build docs when a feature isn't available on the platform you're currently running on.

### DIFF
--- a/test/unit/rustdoc/rustdoc_unit_test.bzl
+++ b/test/unit/rustdoc/rustdoc_unit_test.bzl
@@ -90,6 +90,7 @@ def _target_maker(rule_fn, name, rustdoc_deps = [], **kwargs):
     rust_doc(
         name = "{}_doc".format(name),
         crate = ":{}".format(name),
+        crate_features = kwargs.get("crate_features", []),
     )
 
     rust_doc_test(
--- a/docs/flatten.md
+++ b/docs/flatten.md
@@ -451,8 +451,8 @@ rust_clippy(
 ## rust_doc
 
 <pre>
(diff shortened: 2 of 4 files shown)


ParkMyCar:
Hey @illicitonion, any chance you could review this PR too? No rush though :)

P.S. I tried to assign a reviewer but I don't seem to be able to

scentini (commented):
Could we add a test please?

ParkMyCar:
> Could we add a test please?

Great idea! I added a unit test, please let me know if there is a different test or an example I should add

test/unit/rustdoc/rustdoc_unit_test.bzl
```diff
@@ -90,6 +90,7 @@ def _target_maker(rule_fn, name, rustdoc_deps = [], **kwargs):
     rust_doc(
         name = "{}_doc".format(name),
         crate = ":{}".format(name),
+        crate_features = kwargs.get("crate_features", []),
```
scentini, line 93: I'd also like to confirm that the crate features value ends up on the command line. I believe we can do that by putting a specific feature here and checking that it is contained in the command line here:https://github.com/bazelbuild/rules_rust/blob/7bbe005ca8e4747d6c48cd73985ae47424eb8a2e/test/unit/rustdoc/rustdoc_unit_test.bzl#L12

ParkMyCar:
Stale

Issue #1488, opened 2022-07-28
FR: support for linking via cc_common.link


krasimirgg:
We would like to add support to rules_rust for linking rust binaries via the [cc_common.link](https://docs.bazel.build/versions/main/skylark/lib/cc_common.html) Bazel mechanism. This feature would allow for better cross-language interoperability and make it easier to provide cross-language integration of features like LTO and sanitizer support by adapting the rust compilation pipeline to produce actions and artifacts that are more uniform with other languages' rules implementation from Bazel point of view.

This feature should co-exist with the existing rules_rust mechanism that uses rustc as the linker driver, since there are use cases where rustc as the driver makes sense. For projects that are mostly written in rust using rustc-as-a-driver is simpler to set up and more flexible with respect to features such as custom rust allocator support. The use of cc_common.link would require a fully configured C++ toolchain and custom allocator library for example.

Initially we'd like to experiment with this and focus on rust_binary and rust_test, possibly extending the support to other targets that require linking in the future.

As a rough design, we'd like to be able to control the default linking mechanism on a rust_toolchain level, and additionally be able to selectively override the behavior on a per-target basis until the feature is complete.
We'd add
* a new boolean build setting `--//rules_rust:experimental_use_cc_common_link`, defaulting to false,
* a new `rust_toolchain` label attribute `experimental_use_cc_common_link`, defaulting to the build setting above.
* a new temporary int rust_{binary|test|shared_library}.experimental_use_cc_common_link attribute with possible values `[-1, 0, 1]` to rust_binary and other crate types that require linking, defaulting to -1. This is to support gradually moving targets t/from using the new feature; after the feature is stable, we'll remove this attribute.

Given rust `rust_{binary|...}(name = bin)`, and a `rust_toolchain`, common_cc.link is enabled if:
* `bin.experimental_use_cc_common_link` is 1, or
* `bin.experimental_use_cc_common_link` is -1 and `rust_toolchain.experimental_use_cc_common_link` points to a label whose value is true.

Otherwise common_cc.link is disabled.

When enabled, instead of emitting one rustc_compile_action that does both compilation and linking using `rustc`, we will emit 2 actions:
* A rustc one with [`–emit=obj`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--emit-specifies-the-types-of-output-files-to-generate) to produce an .o file
* A `cc_common.link` one to produce the final binary (or appropriate binary artefacts for other crates that require linking).

[...]

PR #1490, opened 2022-08-01
add cc_common.link support for rust_library and rust_test

krasimirgg:
This PR adds support for using cc_common.link to link rust_binary-es and rust_test-s, following https://github.com/bazelbuild/rules_rust/issues/1488:
* a new boolean build setting `--//rules_rust/rust/settings:experimental_use_cc_common_link`, defaulting to false,
* a new rust_toolchain label attribute `experimental_use_cc_common_link`, defaulting to the build setting above.
* a new temporary int `rust_{binary|test|shared_library}.experimental_use_cc_common_link` attribute with possible values `[-1, 0, 1]` defaulting to -1. This is to support gradually moving targets to/from using the new feature; after the feature is stable, we'll remove this attribute.

Given rust `rust_{binary|test}(name = bin)`, and a `rust_toolchain`, `common_cc.link` is enabled if:

* `bin.experimental_use_cc_common_link` is 1, or
* `bin.experimental_use_cc_common_link` is -1 and `rust_toolchain.experimental_use_cc_common_link` points to a label whose value is true.
Otherwise `common_cc.link` is disabled.

When enabled, instead of emitting one `rustc_compile_action` that does both compilation and linking using rustc, we emit 2 actions:
* A rustc one with [–emit=obj](https://doc.rust-lang.org/rustc/command-line-arguments.html#--emit-specifies-the-types-of-output-files-to-generate) to produce an .o file
* A cc_common.link one to produce the final binary (or appropriate binary artefacts for other crates that require linking).

To test this, I've set up a new `test/cc_common_link` directory and [a build bot](https://buildkite.com/bazel/rules-rust-rustlang/builds/6983#018263a7-840d-4894-8a3b-16024ac202f7) to build and test everything there with `---//rules_rust/rust/settings:experimental_use_cc_common_link` set to true. This also acts as an example directory on how to set up a workspace where this is enabled by default.

This feature requires on an appropriate `rust_toolchain.allocator_library`, similarly to how we need one to be able to build a cc_binary with a rust dependency. So I picked up the shared allocator_library repository updates from @bsilver8192's https://github.com/bazelbuild/rules_rust/pull/1350 to set this up.


rust/private/rustc.bzl
```diff
@@ -1179,6 +1214,52 @@ def rustc_compile_action(
...
+
+        # The path to the package dir, including a trailing "/".
+        package_dir = ctx.bin_dir.path + "/"
+        if ctx.label.workspace_root:
+            package_dir = package_dir + ctx.label.workspace_root + "/"
+        if ctx.label.package:
+            package_dir = package_dir + ctx.label.package + "/"
+
+        if not crate_info.output.path.startswith(package_dir):
+            fail("The package dir path {} should be a suffix of the crate_info.output.path {}", package_dir, crate_info.output.path)
```
scentini, line 1246:
```suggestion
            fail("The package dir path {} should be a prefix of the crate_info.output.path {}", package_dir, crate_info.output.path)
```

rust/private/rustc.bzl
```diff
@@ -1179,6 +1214,52 @@ def rustc_compile_action(
...
+        linking_contexts = [toolchain.libstd_and_allocator_ccinfo.linking_context, toolchain.stdlib_linkflags.linking_context]
+
+        for dep in crate_info.deps.to_list():
+            if dep.cc_info:
+                linking_contexts.append(dep.cc_info.linking_context)
+
+        # In the cc_common.link action we need to pass the name of the final
+        # binary (output) relative to the package of this target.
+        # We compute it by stripping the path to the package directory,
+        # which is a suffix of the path of `crate_info.output`.
```
scentini, line 1236:
```suggestion
        # which is a prefix of the path of `crate_info.output`.
```

scentini (approved):

bsilver8192:
Looking at the CI task added here (as part of adding another one), it looks like it runs the tests without the flag enabled. Is that intentional?

bsilver8192:
From looking at the rest of that file, I'm pretty sure you have to duplicate it under `test_flags` to also apply it to the tests.

PR #1499, opened 2022-08-05
pipelining: add the ability to disable pipelining for a single rule.

gigaroby:
While testing internally we found some non-hermetic libraries have
issues with pipelining enabled, adding a per-rule switch allows us to
keep the feature generally enabled while disabling problematic targets.


scentini (approved):
Thanks!

UebelAndre:
> While testing internally we found some non-hermetic libraries have
> issues with pipelining enabled, adding a per-rule switch allows us to
> keep the feature generally enabled while disabling problematic targets.

@gigaroby Would you be able to share an example of an issue you saw?

gigaroby:
@UebelAndre https://github.com/PyO3/pyo3/blob/8bc86c39ebbebb86dd2db0839691748e06e76cd0/pyo3-build-config/src/lib.rs#L104

UebelAndre:
Why does pipelining break this?

TimNN:
It won't necessarily break in all setups / environments. But `env!("OUT_DIR")` ends up being an absolute path, which gets embedded in the library / metadata (hash). 

So if the build directory is different for the `rlib` and `rmeta` compilations, the metadata hashes will be different, making `rustc` very unhappy.

(Embedding the value of `$OUT_DIR` in the compiled binary seems to be a fragile thing to do in general).

PR #1500, opened 2022-08-05
Do not pass `--Clink-arg=-l` for libstd and libtest

### LINKED ISSUE #1374: Possible regression causing linking of proc_macro

scentini:
Fixes #1374

We skip adding `-Clink-arg=-l...` for libstd and libtest from the standard library, as these two libraries are present both as an `.rlib` and a `.so` format. On linux, Rustc adds a -Bdynamic to the linker command line before the libraries specified with `-Clink-arg`, which leads to us linking against the `.so`s but not putting the corresponding value to the runtime library search paths, which results in a "cannot open shared object file: No such file or directory" error at exectuion time. We can fix this by adding a `-Clink-arg=-Bstatic` on linux, but we don't have that option for macos. The proper solution for this issue would be to remove `libtest-{hash}.so` and `libstd-{hash}.so` from the toolchain. However, it is not enough to change the toolchain's `rust_std_{...}` filegroups here: https://github.com/bazelbuild/rules_rust/blob/a9d5d894ad801002d007b858efd154e503796b9f/rust/private/repository_utils.bzl#L144 because rustc manages to escape the sandbox and still finds them at linking time. We need to modify the repository rules to erase those files completely. This PR should be a good workaround until we get do to the proper thing though.

This PR also fixes the following issues for Windows:
* get_lib_name() didn't work properly on windows for `libc`
* `-Clink-arg` should point to the library name with extension included for windows.


krasimirgg (approved):
Looks good! Thank you!

Issue #1294, opened 2022-04-26
Add integration tests for release artifacts


UebelAndre:
There have been quite a few bugs around the release artifact for `rules_rust` missing some critical things:
- https://github.com/bazelbuild/rules_rust/issues/1272
- https://github.com/bazelbuild/rules_rust/issues/1246
- https://github.com/bazelbuild/rules_rust/issues/1242

Either the release artifact should be updated to not be built as a Bazel target, or an integration test should be added to test the completeness of the artifact.

PR #1501, opened 2022-08-05
Implement rules archive release artifact in github action.

UebelAndre:
Maintaining a representation of the release artifact for the rules has proven problematic. Instead of working to setup necessary checks and tests to make sure it works, it would be simpler to use an approach outside of Bazel.

closes https://github.com/bazelbuild/rules_rust/issues/1294 since it's no longer relevant


UebelAndre:
Using the changes in this PR, I was able to successfully produce a release
Screen Shot 2022-08-05 at 10 22 31 AM
Screen Shot 2022-08-05 at 10 23 04 AM

illicitonion (approved):
Works for me!

PR #1508, opened 2022-08-09
Also apply get_lib_name correctly to the C++ runtime libraries

scentini:


krasimirgg (approved):
LGTM

scentini:
hlopko@ pointed to me why this didn't work internally, when it does work externally:
Internally we populate the `cc_toolchain.{static|dynamic}_runtime_lib` fields, while externally they are not populated in the default C++ toolchain configuration.

As for why this turned out to be an issue on Linux when `get_lib_name` had `for_windows = False` by default:
From [`args.add_all`](https://bazel.build/rules/lib/Args#add_all) documentation:

> **map_each**: callable; or [None](https://bazel.build/rules/lib/globals#None); default = None
A function that converts each item to zero or more strings, which may be further processed before appending. If this param is not provided, the standard conversion is used.
The function is passed either one or two positional arguments: the item to convert, followed by an optional [DirectoryExpander](https://bazel.build/rules/lib/DirectoryExpander). The second argument will be passed only if the supplied function is user-defined (not built-in) and declares more than one parameter.

So we end up passing a `DirectoryExpander` as the second argument, and as it is not `None`, `for_windows` will evaluate to `True`. -.-

PR #1511, opened 2022-08-11
Skip supplying rpaths on Fuchsia

tmandry:
Fuchsia rules assemble shared libraries during packaging, so rpaths are not needed.


illicitonion (approved):
Thanks!

PR #1512, opened 2022-08-12
Added Rust 1.63.0

UebelAndre:
https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html


illicitonion (approved):
There's a surprising number of .rs files changed in this PR? But both set of changes LGTM!

kalcutter:
Please make a new release with this change :tada:

UebelAndre:
> There's a surprising number of .rs files changed in this PR? But both set of changes LGTM!

Clippy! If ever I make a change like that that’s generally why. Clippy or Rustfmt will have some change that doesn’t like some code.

PR #1518, opened 2022-08-15
Updates rust_test to use main.rs as the root when use_libtest_harness is false

### LINKED ISSUE #1516: `rust_test` should look for main.rs instead of lib.rs when use_libtest_harness is false

csmulhern:
Fixes #1516.


UebelAndre (changes requested):
Seems like a test needs to be updated to accompany this change? Otherwise seems good to me 😄 

(also, sorry for the delay!)

csmulhern:
> Seems like a test needs to be updated to accompany this change? Otherwise seems good to me 😄
> 
> (also, sorry for the delay!)

Fixed!

(and no worries)

test/unit/use_libtest_harness/use_libtest_harness_test.bzl
```diff
@@ -36,7 +36,10 @@ def _use_libtest_harness_test():
 
     rust_test(
         name = "mytest_noharness",
-        srcs = ["mytest_noharness.rs"],
+        srcs = [
```
UebelAndre, line 39: Is this PR a breaking change? Do test targets with a single source file not using the test harness now require a `main.rs` or will they be able to function just as they did before?

UebelAndre (approved):
Looks good to me but one clarifying question about the impact of this change.

csmulhern:
Test targets with a single source file not using the test harness function as they did before.

Test targets with multiple source files not using the test harness will break if they had not specified `crate_root` manually (i.e. they relied on the implicit use of `lib.rs` as the crate root, and defined a custom main function in their crate).

UebelAndre:
> Test targets with a single source file not using the test harness function as they did before.
> 
> Test targets with multiple source files not using the test harness will break if they had not specified `crate_root` manually (i.e. they relied on the implicit use of `lib.rs` as the crate root, and defined a custom main function in their crate).

Would you be willing to add a test case for that? Just a copy of `mytest_noharness` that continues to use a single source file? 😅

csmulhern:
Done!

UebelAndre (approved):
Thank you so much!

PR #1526, opened 2022-08-21
Preserve directory structure of source files when some are generated

### LINKED ISSUE #1510: Broken rust_library with modules and generated sources
The commit `c7cb5bd7a85496b31ab3e20216e7b35677572d09` wrongly flattens the hierarchy of rust crates.
This results in the any crate using modules not compiling anymore as long as it uses generated sources as well.

```
src/lib.rs
src/my_module/mod.rs
src/my_module/stuff.rs
```

```
rust_library(
   name = "my_crate",
   srcs = [
       "src/my_module/mod.rs",
       "src/my_module/stuff.rs",
       "src/lib.rs",
       ":generated_file",
   ]
)
```

The error comes from the use of `src.basename` in `transform_sources`, instead of the relative file path as it should.

cameron-martin:
When some input files are generated, the source files get symlinked to `bazel-out` so the generated and source files are in the same directory. This was implemented in #1340. However, the directory structure would get flattened since only the basename was used for the symlink location. Now the package-relative path is used.

Fixes #1510.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -151,9 +151,10 @@ def _transform_sources(ctx, srcs, crate_root):
     generated_sources = []
 
     generated_root = crate_root
+    package_root = paths.dirname(ctx.build_file_path)
 
     if crate_root and (crate_root.is_source or crate_root.root.path != ctx.bin_dir.path):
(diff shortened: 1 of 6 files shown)


illicitonion (approved):
Thanks so much!

Issue #1493, opened 2022-08-01
Add support for bzlmod


UebelAndre:
It would be nice if `rules_rust` supported [bzlmod](https://docs.bazel.build/versions/main/bzlmod.html) workflows. Hopefully someone familiar with that can enable that for this project.

cameron-martin:
I don't know that much about bzlmod, but have started having a look at this. Initially I would like to simply replace `rules_rust_dependencies`. However, I'm having some issues. Details in #1528.

hobofan:
I think it makes sense to reopen this? While the work that has been merge, from what I can tell there is still quite a bit of work that needs to happen to "support bzlmod" on a level that it can actually be adopted.

katre:
Are loading crates via bzlmod supported in the current effort?

opicaud:
Hey, 
Please check #2021

csmulhern:
Bazel 7.0 is releasing later this year, with bzlmod becoming on by default. Is anyone working to prioritize this work to align with the release of Bazel 7.0? Looks like there has been some work done in #2021.

UebelAndre:
It unfortunately seems no maintainer has any experience with bzlmod at all. I've asked other members of the Bazel org to help do reviews but I unfortunately haven't seen any support. The Bzlmod changes could probably be merged as long as there's no change to existing code. For anything that needs to change, the more granular the pull-reqeusts can be the easier the rollout will be. My primary concern is that bzlmod just increases the maintenance burden by introducing a second way to to fetch dependencies. So it'd also be helpful if the changes were setup to maintain a single source of truth.

scentini:
I have scheduled some time with Bzlmod experts later this week to peer review this PR with the hope that we can move this PR forward and gain some expertise in the process.

csmulhern:
Makes sense. Thanks for the update.

I would hope long term, Bzlmod becomes the primarily supported way to fetch dependencies, though that's probably quite far in the future.

Shot in the dark, but I have tried to solicit help in this discussion thread: https://github.com/bazelbuild/bazel/discussions/18329.

csmulhern:
> I have scheduled some time with Bzlmod experts later this week to peer review this PR with the hope that we can move this PR forward and gain some expertise in the process.

That sounds great, thanks @scentini. I am happy to dedicate some bandwidth in the coming month to help with contributions (though I also have no prior expertise).

UebelAndre:
[...]

PR #1528, opened 2022-08-25
Support bzlmod

cameron-martin:
This makes rules_rust load dependencies via bzlmod. Currently only basic functionality is completed, such as registering rustc toolchains and compiling crates. Note that it cannot interact with cargo, wasm or load any other rust toolchains such as rustfmt.
    
There is one new module, `examples/bzlmod/hello_world`, that depends on the root `rules_rust` module. This example can be built and run using:
    
```
cd examples/bzlmod/hello_world
bazel run //:hello_world
```

To register toolchains in an ergonomic way, it defines a new "hub" repository that contains all the toolchain proxies, so they can all be registered like so:

```
```

closes https://github.com/bazelbuild/rules_rust/issues/1493


UebelAndre:
re https://github.com/bazelbuild/rules_rust/issues/1493#issuecomment-1227590691

@scentini do you or anyone you have contact with at Google know anything about bzlmod?

scentini:
I'm not familiar with bzlmod, in case we're doing something silly here it's not obvious to me. @Wyverald could you PTAL?

scentini:
cc @meteorcloudy

meteorcloudy:
Bzlmod may have some bugs in Bazel 5.3.0, it's recommended to test with Bazel@HEAD, you can use Bazelisk:
```
export USE_BAZEL_VERSION=last_green
bazelisk build //:hello_world
```
I got the following toolchain error instead of the visibility error:
```
ERROR: /Users/pcloudy/workspace/rules_rust/examples/bzlmod/hello_world/BUILD.bazel:5:12: While resolving toolchains for target //:hello_world: No matching toolchains found for types @rules_rust~override//rust:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@rules_rust~override//rust:toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
```

rust/private/repository_utils.bzl
```diff
@@ -205,15 +205,15 @@ load("@rules_rust//rust:toolchain.bzl", "rust_toolchain")
 
 rust_toolchain(
     name = "{toolchain_name}",
-    rust_doc = "@{workspace_name}//:rustdoc",
-    rust_std = "@{workspace_name}//:rust_std-{target_triple}",
-    rustc = "@{workspace_name}//:rustc",
+    rust_doc = "//:rustdoc",
```
cameron-martin, line 208: If I try to use the repository name in this, I get the following error:

```
ERROR: /home/cameron/.cache/bazel/_bazel_cameron/538574de729809a828a75cd1284ac81c/external/rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools/BUILD.bazel:89:15: no such package '@[unknown repo 'rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools' requested from @rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools]//': The repository '@[unknown repo 'rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools' requested from @rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools]' could not be resolved: Repository '@rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools' is not visible from repository '@rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools' and referenced by '@rules_rust~override~toolchains~rust_linux_x86_64__x86_64-unknown-linux-gnu_tools//:rust_toolchain'
```

rust/repositories.bzl
```diff
@@ -610,20 +610,20 @@ def rust_repository_set(
...
-
-        if "io_bazel_rules_rust" in native.existing_rules():
-            message += "\n\n" + (
-                "Note that the previous name of \"@io_bazel_rules_rust\" is deprecated.\n" +
-                "See https://github.com/bazelbuild/rules_rust/issues/499 for context."
-            )
-
-        message += "\n" + ("=" * 79)
-        fail(message)
+    # # Inform users that they should be using the canonical name if it's not detected
```
cameron-martin, line 613: This errors, because the name of the repository is something like `@rules_rust~override`

cameron-martin:
I've updated the PR with a rough proof of concept that can compile and run a rust binary.

rust/repositories.bzl
meteorcloudy: The `native.existing_rules()` API no longer works in Bzlmod. Since we have strict deps in Bzlmod, it's already impossible for the user to depend on something they didn't declare. Maybe we don't need this check anymore?

rust/private/repository_utils.bzl
meteorcloudy: If you want to use the canonical repo name in a label string, you'll need to use `@@canonical_repo_name//foo:bar`, see https://github.com/bazelbuild/bazel/issues/15593

But I guess it's fine to use the label without repo name here since it's referring to targets in the same repo as the BUILD file resides?

meteorcloudy: /cc @Wyverald

meteorcloudy:
Nice, can you also fix the buildifier errors?

examples/bzlmod/hello_world/MODULE.bazel
```diff
@@ -0,0 +1,18 @@
+module(
+    name = "hello_world",
+    toolchains_to_register = ["@rust_linux_x86_64__x86_64-unknown-linux-gnu//:toolchain"],
```
meteorcloudy, line 3: Please use the `register_toolchain` directive instead of the `toolchains_to_register` attribute. Then you can group it together with the `use_rep` directive.

meteorcloudy: See https://github.com/bazelbuild/bazel/issues/15829

examples/bzlmod/hello_world/MODULE.bazel
```diff
@@ -0,0 +1,18 @@
...
+)
+
+bazel_dep(name = "rules_rust", version = "0.9.0")
+
+local_path_override(
+    module_name = "rules_rust",
+    path = "../../..",
+)
+
+rust_toolchains = use_extension("@rules_rust//:extensions.bzl", "toolchains")
```
meteorcloudy, line 14: I think the toolchain registration can be done in rules_rust's MODULE.bazel file, so that the example module can just depend on rules_rust and get the toolchain for free?

cameron-martin: Which toolchains are available depends on what parameters the user passes to `rust_toolchains.toolchain`. For example, passing `extra_target_triples` generates additional toolchains.

At least, this would be the case if I were to add all the parameters to the tag class that can be forwarded to `rust_register_toolchains`.

rust/private/repository_utils.bzl
cameron-martin: Omitting the repo name looks fine to me in this context.

rust/repositories.bzl
cameron-martin: Do we still need this for non-bzlmod usages? Maybe @UebelAndre would know?

If so, I could add a parameter to ignore this check that only is intended to be used by the bzlmod module extension.

examples/bzlmod/hello_world/MODULE.bazel
cameron-martin: This does seem to be a significant downside to using bzlmod here. Where previously the user of rules_rust didn't have to know which toolchains were registered, they now have to at the least have the following boilerplate in their MODULE.bazel, as well as know how the options passed to `rust_toolchains.toolchain` map to generated repositories:

```
use_repo(
    rust_toolchains,
    "rust_linux_x86_64__x86_64-unknown-linux-gnu",
    "rust_darwin_aarch64__aarch64-apple-darwin",
    "rust_linux_aarch64__aarch64-unknown-linux-gnu",
    "rust_darwin_x86_64__x86_64-apple-darwin",
    "rust_windows_x86_64__x86_64-pc-windows-msvc",
    "rust_freebsd_x86_64__x86_64-unknown-freebsd",
)

register_toolchains(
    "@rust_linux_x86_64__x86_64-unknown-linux-gnu//:toolchain",
    "@rust_darwin_aarch64__aarch64-apple-darwin//:toolchain",
    "@rust_linux_aarch64__aarch64-unknown-linux-gnu//:toolchain",
    "@rust_darwin_x86_64__x86_64-apple-darwin//:toolchain",
    "@rust_windows_x86_64__x86_64-pc-windows-msvc//:toolchain",
    "@rust_freebsd_x86_64__x86_64-unknown-freebsd//:toolchain",
)
```

I can't think of a way to avoid this currently.

meteorcloudy: It seems we lack of a way to dynamically register generated toolchains. @Wyverald Is it possible to support `register_toolchain` in module extension implementation?

Wyverald: it's possible but we decided not to. I think it's a bit late to introduce it at this point for 6.0. Worst comes to worst, we could have a module called "rules_rust_toolchains" that contains nothing but the above in its MODULE.bazel file.

illicitonion: This may be a pretty significant performance regression - each of these toolchains is in a tar which is hundreds of megabytes, so forcing the user to download 6 of these just to get a toolchain definitions, 5 of which they won't use, is pretty bad.

We could re-structure so that the toolchain definitions live in separate repositories from the actual toolchains, to avoid the proactive downloads, but I wanted to highlight that this is a significant API difference from Bazel.

Wyverald: That's very true. The setup you mentioned is the same as the Java toolchains, so at least there's precedent.

I can think of another potential solution to this. We can make the `toolchains` extension generate a "hub" repo (let's say called "toolchains_hub"), which has nothing but a list of aliases pointing to toolchains that the user requested. And then, we could make rules_rust always register everything in `@toolchains_hub//:all`. That would mean whatever the user requests via tags gets registered, without the user having to do anything. Something like the following:

```bzl
# rules_rust/MODULE.bazel
toolchains = use_extension("//:extensions.bzl", "toolchains")
use_repo(toolchains, "toolchains_hub")
register_toolchains("@toolchains_hub//:all")

# rules_rust/extensions.bzl
toolchains = module_extension(_toolchains_impl, tag_classes = ...)
def _toolchains_impl(ctx):
  toolchain_repos = _somehow_get_repos_from_tags(ctx)
  for toolchain_repo in toolchain_repos: rust_toolchain_repo(**toolchain_repo)
  toolchain_hub_repo(name="toolchains_hub", repos=[tr["name"] for tr in toolchain_repos])
rust_toolchain_repo = repository_rule(...)
toolchain_hub_repo = repository_rule(_toolchain_hub_repo_impl, attrs={"repos": attr.string_list()})
def _toolchain_hub_repo_impl(ctx):
  ctx.file("WORKSPACE")
  ctx.file("BUILD", "\n".join(["alias(name='%s', actual='@%s//:toolchain')" % (repo,repo) for repo in ctx.attr.repos]))

# my_project/MODULE.bazel
bazel_dep(name="rules_rust", version="...")
rust_toolchains = use_extension("//:extensions.bzl", "toolchains")
rust_toolchains.toolchain(edition="2021")
```

cameron-martin: This would work well, as long as using `:all` works correctly for toolchain registration. I'll try it out later.

cameron-martin: Regarding the performance regression, is this any different from the current situation? A lot of unused toolchains get registered currently.

meteorcloudy: Even without Bzlmod, all toolchains get registered in WORKSPACE will be fetched for toolchain resolution, so I think this is the same as the current situation.

illicitonion: > Regarding the performance regression, is this any different from the current situation? A lot of unused toolchains for different platforms get registered currently.

Aah good point, we already split out the meta-toolchain from the expensive repo: https://github.com/bazelbuild/rules_rust/blob/cb9ca1b817febcaf8edc28d160866be759825c31/rust/repositories.bzl#L495-L510

Sorry for the noise!

cameron-martin: It looks like toolchain registration doesn't work with `:all`, since I get the following error when I try to use it:

```
ERROR: /home/cameron/Repos/rules_rust/examples/bzlmod/hello_world/BUILD.bazel:5:12: While resolving toolchains for target //:hello_world: No matching toolchains found for types @rules_rust~override//rust:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@rules_rust~override//rust:toolchain_type'
```

I've updated the PR to use a hub repo anyway, since it does cut down on the boilerplate.

Wyverald: If registering with `:all` isn't working, it's a bug. We have tests (such as https://cs.opensource.google/bazel/bazel/+/master:src/test/java/com/google/devtools/build/lib/skyframe/ToolchainsForTargetsTest.java;l=219;drc=16c4076cd39f43f3ccabbdeb6ab407c2063b038c) ensuring this behavior.

Do you have more information to help us debug? Check the BUILD file you generated to see whether it actually contains toolchains, etc.?

cameron-martin: The BUILD file contains a bunch of aliases to toolchains:

```
alias(name='rust_darwin_aarch64__aarch64-apple-darwin', actual='@rust_darwin_aarch64__aarch64-apple-darwin//:toolchain', visibility = ["//visibility:public"])
alias(name='rust_linux_aarch64__aarch64-unknown-linux-gnu', actual='@rust_linux_aarch64__aarch64-unknown-linux-gnu//:toolchain', visibility = ["//visibility:public"])
alias(name='rust_darwin_x86_64__x86_64-apple-darwin', actual='@rust_darwin_x86_64__x86_64-apple-darwin//:toolchain', visibility = ["//visibility:public"])
alias(name='rust_windows_x86_64__x86_64-pc-windows-msvc', actual='@rust_windows_x86_64__x86_64-pc-windows-msvc//:toolchain', visibility = ["//visibility:public"])
alias(name='rust_freebsd_x86_64__x86_64-unknown-freebsd', actual='@rust_freebsd_x86_64__x86_64-unknown-freebsd//:toolchain', visibility = ["//visibility:public"])
alias(name='rust_linux_x86_64__x86_64-unknown-linux-gnu', actual='@rust_linux_x86_64__x86_64-unknown-linux-gnu//:toolchain', visibility = ["//visibility:public"])
```

The issue seems to be with the interaction between `:all` and aliases. I've filed a bug report at https://github.com/bazelbuild/bazel/issues/16298.

cameron-martin: I've updated this to put all the toolchain proxies into one repository. This does change the public interface, currently does break non-bzlmod usage and does not include rust-analyzer support, but the idea is there.

rust/repositories.bzl
```diff
@@ -597,7 +608,9 @@ def rust_repository_set(
             sha256s = sha256s,
             urls = urls,
             auth = auth,
-        ))
+        )
+
+        all_toolchain_names.append("@{name}//:toolchain".format(name = name))
 
     # This repository exists to allow `rust_repository_set` to work with the `maybe` wrapper.
```
cameron-martin, line 615: I can't say that I really understand the point of this repository. `ALL_TOOLCHAINS` from the repository does not seem to be used anywhere. Can anyone explain this to me?

UebelAndre: To my knowledge, the content of this repository is not used. As the comment states, it's only there to allow `rust_repository_set` to satisfy the `maybe` interface.

cameron-martin: Makes sense.

rust/repositories.bzl
UebelAndre: Sorry, missed this one. The code here can be deleted. I opened https://github.com/bazelbuild/rules_rust/pull/1577 to do so.

cameron-martin: Thanks!

benbrittain:
Is there any work I can take on here to help this get landed?

extensions.bzl
```diff
@@ -0,0 +1,45 @@
...
+                name = "raw_cargo_{}_{}".format(crate.name, crate.version.replace(".", "_")),
+                sha256 = crate.sha256,
+                url = "https://crates.io/api/v1/crates/{}/{}/download".format(crate.name, crate.version),
+                strip_prefix = "{}-{}".format(crate.name, crate.version),
+                type = "tar.gz",
+                build_file_content = _create_build_file_content(crate.name),
+            )
+
+raw_cargo_crate = tag_class(attrs = {"name": attr.string(), "version": attr.string(), "sha256": attr.string()})
+raw_cargo = module_extension(
```
jfirebaugh, line 42: I see `raw_cargo` is used to establish internal dependency on the tinyjson crate. Probably not something intended to be a public API. Perhaps move to `rust/private/extensions.bzl`?

UebelAndre:
@cameron-martin hey, any updates?

cameron-martin:
> @cameron-martin hey, any updates? smile

Sorry I've been busy with other things. I may be able to get back to this in the next week or so.

extensions.bzl
cameron-martin: You're right.

opicaud:
hello @cameron-martin, do you have any update ?
Thanks!

matts1:
I've got a prototype working for a module extension for bzlmod. It's still very much a WIP, and needs a lot of cleaning up, but you can at least build examples/bzlmod/crates_repository.

https://github.com/matts1/rules_rust/tree/module_extension_prototype

TLDR on how to use:
```
crates_vendor = use_extension("@rules_rust//:extensions.bzl", "crates_vendor")
crates_vendor.crates_vendor(
    lockfile = "//:cargo-bazel-lock.json",
)
use_repo(
    crates_vendor,
    crates = "crates_crates_repository",
)
```

cameron-martin:
I've rebased this PR. Here are the main things that need finishing:

- [ ] Support other types of toolchains (rust analyzer, rustfmt, some wasm stuff that I don't fully understand).
- [ ] Support all options when registering toolchains via bzlmod.
- [ ] Work out how to run the examples without having to cd into examples/bzlmod/hello_world

examples/bzlmod/hello_world/MODULE.bazel
cameron-martin: This hub repository now duplicates the toolchain proxies instead of replaces them, so the public API does not change.

cameron-martin:
We also have to consider the API for the module extensions. How many module extensions do we want? Just a single one, called "rust" for everything including crates_universe, etc? One per top-level package, e.g. rust, crates_universe, cargo?

MODULE.bazel
```diff
@@ -0,0 +1,23 @@
+module(
+    name = "rules_rust",
+    version = "0.9.0",
+)
```
matts1, line 4: rules_go put the following in their MODULE.bazel. I think we should add something similar to rules_rust:

`print("WARNING: The rules_go Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.")`

https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_go/0.33.0/MODULE.bazel#L7

matts1:
IMO, we should seperate them into areas of concern, similar to how [other bzlmod modules](https://github.com/bazelbuild/rules_python/blob/main/examples/bzlmod/MODULE.bazel) do it.

examples/bzlmod/hello_world/MODULE.bazel
```diff
@@ -0,0 +1,19 @@
...
+    version = "1.0",
+)
+
+bazel_dep(name = "rules_rust", version = "0.9.0")
+local_path_override(
+    module_name = "rules_rust",
+    path = "../../..",
+)
+
+rust_toolchains = use_extension("@rules_rust//rust:extensions.bzl", "toolchains")
```
cameron-martin, line 12: This module extension should probably be called "rust".

MODULE.bazel
```diff
@@ -0,0 +1,21 @@
...
+bazel_dep(name = "platforms", version = "0.0.5")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "bazel_skylib", version = "1.2.0")
+bazel_dep(name = "apple_support", version = "0.13.0")
+
+# Load tinyjson without using the process wrapper
+raw_cargo = use_extension("//rust/private:extensions.bzl", "raw_cargo")
+raw_cargo.crate(
+    name = "tinyjson",
+    sha256 = "1a8304da9f9370f6a6f9020b7903b044aa9ce3470f300a1fba5bc77c78145a16",
```
cameron-martin, line 15: These values should probably be shared with the non-bzlmod implementation

matts1: PTAL at the PR I sent you. Amongst other things, it shares the values.

https://github.com/cameron-martin/rules_rust/pull/1/commits/0760cf7516a75eae59dbcf7f20316b6bb55642c6

cameron-martin:
This should now be ready to review.

rust/extensions.bzl
```diff
@@ -0,0 +1,42 @@
...
+load(
+    "//rust/private:repository_utils.bzl",
+    "DEFAULT_EXTRA_TARGET_TRIPLES",
+    "DEFAULT_NIGHTLY_VERSION",
+    "DEFAULT_STATIC_RUST_URL_TEMPLATES",
+)
+load(":repositories.bzl", "rust_register_toolchains")
+
+def _rust_impl(ctx):
+    mod = ctx.modules[0]
```
matts1, line 12: Nit: Don't think this is guaranteed to be the root module.

`mod = [m for m in ctx.modules if m.is_root][0]`

UebelAndre:
cc @scentini @krasimirgg Hoping one of you two know anything about `bzlmod` and would be able to review this 😅

scentini:
I read up a bit on `bzlmod` and unless someone with in-depth knowledge offers to do the review, I'll claim the 'expert' title. This PR LGTM, and aside from rebasing and updating the version in `MODULE.bazel`, I'd try merging.

MODULE.bazel
```diff
@@ -0,0 +1,18 @@
...
+
+print("WARNING: The rules_rust Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.")  # buildifier: disable=print
+
+bazel_dep(name = "platforms", version = "0.0.5")
+bazel_dep(name = "rules_cc", version = "0.0.1")
+bazel_dep(name = "bazel_skylib", version = "1.2.0")
+bazel_dep(name = "apple_support", version = "1.3.1")
+
+non_bzlmod_deps = use_extension("//rust/private:extensions.bzl", "non_bzlmod_deps")
+non_bzlmod_deps.non_bzlmod_deps()
```
Wyverald, line 14: this line is unnecessary. the `use_extension` clause is enough to triger the extension

Wyverald (commented):
had a brief pass, mostly looks good! I didn't read the actual implementation very deeply.

MODULE.bazel
```diff
@@ -0,0 +1,18 @@
+module(
+    name = "rules_rust",
+    version = "0.9.0",
```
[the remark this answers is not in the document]
cameron-martin: I've updated this to the same version as in `version.bzl`

rust/private/extensions.bzl
```diff
@@ -0,0 +1,16 @@
...
+
+def _non_bzlmod_deps_impl(_module_ctx):
+    http_archive(**TINYJSON_KWARGS)
+
+non_bzlmod_deps_non_bzlmod_deps = tag_class(attrs = {})
+non_bzlmod_deps = module_extension(
+    doc = "Dependencies for rules_rust",
+    implementation = _non_bzlmod_deps_impl,
+    tag_classes = dict(
+        non_bzlmod_deps = non_bzlmod_deps_non_bzlmod_deps,
```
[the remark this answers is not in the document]
cameron-martin: Renamed to `internal_deps`

MODULE.bazel
cameron-martin: Fixed

scentini:
Thank you @Wyverald !

scentini (approved):

meteorcloudy:
Nice! Will there be the next step to integrate with cargo to actually fetch rust dependencies?

examples/bzlmod/hello_world/MODULE.bazel
```diff
@@ -0,0 +1,19 @@
...
+    version = "1.0",
+)
+
+bazel_dep(name = "rules_rust", version = "0.9.0")
+local_path_override(
+    module_name = "rules_rust",
+    path = "../../..",
+)
+
+rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
```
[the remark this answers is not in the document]
cameron-martin: Currently only the toolchains declared in the top level module are registered. There was an unresolved discussion on whether this was the correct thing to do [here](https://github.com/bazelbuild/rules_rust/pull/1842#discussion_r1116293025), but currently the toolchain that rules_rust defines would not get registered.

cameron-martin:
> Nice! Will there be the next step to integrate with cargo to actually fetch rust dependencies?

@matts1 currently has a PR for this here: https://github.com/bazelbuild/rules_rust/pull/1910

examples/bzlmod/hello_world/MODULE.bazel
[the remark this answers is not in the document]
meteorcloudy: Got it, thanks!

PR #1537, opened 2022-09-01
build_script: move to separate dep line from deps

durin42:
Much like proc_macro_deps, build_script needs to be built for `cfg = "exec"`
rather than as part of what's built for the deployment target. Since there's
no way to mark individual dependencies as `cfg = "exec"` we break the
build_script dependency out onto its own line. Only one build script was
ever allowed anyway, so once we deprecate putting build scripts in deps we'll
get the bonus of making a specific kind of build misconfiguration impossible!

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -328,6 +328,10 @@ def _rust_library_common(ctx, crate_type):
     deps = transform_deps(ctx.attr.deps)
     proc_macro_deps = transform_deps(ctx.attr.proc_macro_deps + get_import_macro_deps(ctx))
 
+    build_script = ctx.attr.build_script
+    if build_script != None:
+        build_script = transform_deps([ctx.attr.build_script])[0]
+
     return rustc_compile_action(
         ctx = ctx,
         attr = ctx.attr,
@@ -338,6 +342,7 @@ def _rust_library_common(ctx, crate_type):
             root = crate_root,
             srcs = depset(srcs),
             deps = depset(deps),
+            build_script = build_script,
             proc_macro_deps = depset(proc_macro_deps),
             aliases = ctx.attr.aliases,
             output = rust_lib,
@@ -369,6 +374,9 @@ def _rust_binary_impl(ctx):
 
     deps = transform_deps(ctx.attr.deps)
     proc_macro_deps = transform_deps(ctx.attr.proc_macro_deps + get_import_macro_deps(ctx))
+    build_script = ctx.attr.build_script
+    if build_script != None:
+        build_script = transform_deps([ctx.attr.build_script])[0]
 
     srcs, crate_root = _transform_sources(ctx, ctx.files.srcs, getattr(ctx.file, "crate_root", None))
     if not crate_root:
@@ -384,6 +392,7 @@ def _rust_binary_impl(ctx):
             root = crate_root,
             srcs = depset(srcs),
             deps = depset(deps),
+            build_script = build_script,
             proc_macro_deps = depset(proc_macro_deps),
             aliases = ctx.attr.aliases,
             output = output,
@@ -415,10 +424,15 @@ def _rust_test_impl(ctx):
 
     deps = transform_deps(ctx.attr.deps)
     proc_macro_deps = transform_deps(ctx.attr.proc_macro_deps + get_import_macro_deps(ctx))
+    build_script = ctx.attr.build_script
+    if build_script != None:
+        build_script = transform_deps([ctx.attr.build_script])[0]
 
     if ctx.attr.crate:
         # Target is building the crate in `test` config
         crate = ctx.attr.crate[rust_common.crate_info] if rust_common.crate_info in ctx.attr.crate else ctx.attr.crate[rust_common.test_crate_info].crate
+        if build_script == None:
+            build_script = crate.build_script
 
         output_hash = determine_output_hash(crate.root, ctx.label)
         output = ctx.actions.declare_file(
@@ -445,6 +459,7 @@ def _rust_test_impl(ctx):
             root = crate.root,
             srcs = depset(srcs, transitive = [crate.srcs]),
             deps = depset(deps, transitive = [crate.deps]),
+            build_script = build_script,
             proc_macro_deps = depset(proc_macro_deps, transitive = [crate.proc_macro_deps]),
             aliases = ctx.attr.aliases,
             output = output,
@@ -477,6 +492,7 @@ def _rust_test_impl(ctx):
             root = crate_root,
             srcs = depset(srcs),
             deps = depset(deps),
+            build_script = build_script,
             proc_macro_deps = depset(proc_macro_deps),
             aliases = ctx.attr.aliases,
             output = output,
@@ -545,6 +561,15 @@ _common_attrs = {
 
             These are other `rust_library` targets and will be presented as the new name given.
         """),
+    ),
+    # Previously `build_script` was part of `deps`, but much like proc_macro_deps we need to tell
+    # bazel that this is `cfg = "exec"` and making it its own dep line solves the problem.
+    "build_script": attr.label(
+        doc = dedent("""\
+            The `cargo_build_script` dependency, if any, to run as part of building this target.
+        """),
+        cfg = "exec",
+        providers = [rust_common.build_info],
     ),
     "compile_data": attr.label_list(
         doc = dedent("""\
--- a/docs/defs.md
+++ b/docs/defs.md
@@ -90,9 +90,10 @@ Add additional rustc_flags from the command line with `--@rules_rust//:extra_rus
 ## rust_binary
 
 <pre>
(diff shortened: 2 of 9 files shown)


UebelAndre:
Can you provide an example of the issue you saw without this change? To my knowledge cargo_build_script is running in the correct configuration. The rule is run in “target” but it takes a “script” attribute in exec which is what’s run to produce outputs

scentini:
We ran into issues using Bazel's [`compatible_with`](https://bazel.build/reference/be/common-definitions#common.compatible_with); we currently need to set many `compatible_with` to `cargo_build_script` targets although they are not supposed to be built in target configuration.

We don't actually build everything in the right configuration - [`deps`](https://github.com/bazelbuild/rules_rust/blob/main/cargo/cargo_build_script.bzl#L259) seems wrong here.

In my ideal world, we would be able to set `cfg="exec"` to the rule definition itself on both `cargo_build_script` and `proc_macro_deps`. But the Bazel configurability team tells me that is not possible.

I've been ponderring on wrapping the `proc_macro_deps` rule in a macro that does the "exec" transition so that we don't have to have the separation of `deps` and `proc_macro_deps`. We could do the same for `cargo_build_script`, but it seems different enough that it warrants its own attribute?

rust/private/rust.bzl
```diff
@@ -328,6 +328,10 @@ def _rust_library_common(ctx, crate_type):
     deps = transform_deps(ctx.attr.deps)
     proc_macro_deps = transform_deps(ctx.attr.proc_macro_deps + get_import_macro_deps(ctx))
 
+    build_script = ctx.attr.build_script
+    if build_script != None:
+        build_script = transform_deps([ctx.attr.build_script])[0]
```
scentini, line 333: I think we can create the DepVariantInfo here, given that we know that build_script contains a BuildInfo, so we don't need to call transform_deps. Same below.

scentini:
Hey @UebelAndre, I tried to address your comment above - we are forced to use [`compatible_with`](https://bazel.build/reference/be/common-definitions#common.compatible_with) on many `cargo_build_script` targets although they are not supposed to be relevant to `target` configuration.

UebelAndre:
> Hey @UebelAndre, I tried to address your comment above - we are forced to use [`compatible_with`](https://bazel.build/reference/be/common-definitions#common.compatible_with) on many `cargo_build_script` targets although they are not supposed to be relevant to `target` configuration.

Ah sorry, seemed like kinda an aside to the ticket itself 😅

First, would you have some detailed documentation on `compatible_with`? I don't think I understand it's use in comparison to `target_compatible_with` or `exec_compatible_with`. 

Second, I agree `cargo_build_script.deps` seems wrong. I thought build scripts were built for the execution platform, then used to generate things for the target platform. Doesn't this mean that conceptually, having a `rust_binary` as `cargo_build_script.runner` in the `exec` configuration solve for this? Is `deps` the only thing that needs to be fixed? (that's likely not true and feel learning more about `compatible_with` would help me realize that).

scentini:
Sorry for the delayed response!

Unfortunately I couldn't find much documentation of `compatible_with`, except that the underlying mechanism is not documented: https://github.com/bazelbuild/bazel/issues/10849. Looks like it might be on its way out.

The issue ended up being in only the `deps` being built in `target` mode: The `compatible_with` propagated through them just as `target_compatible_with` would do, so if we have a `cargo_build_script` that has a `compatible_with` values, all the deps need to set `compatible_with` too. I am preparing a PR for `cfg = "exec"` on `deps` over at https://github.com/bazelbuild/rules_rust/pull/1561, which would unblock us.

I believe that this PR still makes sense though; build scripts are a special type of dependency; one can have only one of them and having an attribute for them would make it more explicit. WDYT?

krasimirgg:
> cargo_build_script

krasimirgg:
(sorry accidentally closed the pr when trying to post a comment)

I think this PR makes sense for the reasons @scentini points out -- they are just too different than standard dependencies and separating them out makes this more clear.

illicitonion:
Works for me 👍

UebelAndre:
> I think this PR makes sense for the reasons @scentini points out -- they are just too different than standard dependencies and separating them out makes this more clear.

I'm not sure my question was clear. There have been conversations in the past around whether or not `cargo_build_script` should be a part of the `//rust` package, which to me translates to, "should the Bazel rules for building Rust code need to re-implement this build script feature from Cargo". I had thought the consensus was "no" and that the Bazel rules should only concern itself with wiring up calls to `rustc` to compile code. This change makes changes to the core rust rules, which grow the footprint of `cargo_build_script` which is the opposite direction I expected. My question isn't "do we think the change makes sense" but "does everyone agree that we should be changing the core rust rules to account for this re-implementation of a Cargo feature?"

Is that clear to everyone?

krasimirgg:
Chatted a bit with @scentini:
* the `BuildInfo` feels like part of core rules_rust; it's important at least for integration with third party dependencies / the crate universe / etc
* looks like `BuildInfo` is at an abstraction that can support additional mechanisms beyond cargo that provide addtional build information. It's of course heavily inspired by cargo and right now in rules_rust we only have cargo as example, but we can "keep the door open" for other build info providers via this info and the rest of rules_rust should "just work" with then.
* it would make sense to have the name of the new attribute be more neutral `build_script`
* there's already a rustc check that we only support at most 1 of these here: https://github.com/bazelbuild/rules_rust/blob/f0cdcedc2a68ffdb3048a8510b380e64eb9a1d72/rust/private/rustc.bzl#L269
I think it would make most sense to keep this assumption of at most 1 for these, it seems like a simple model to think about; if we ever have a compelling reason to allow multiple we can relax it then.

rust/private/rust.bzl
```diff
@@ -546,6 +562,15 @@ _common_attrs = {
             These are other `rust_library` targets and will be presented as the new name given.
         """),
     ),
+    # Previously `build_script` was part of `deps`, but much like proc_macro_deps we need to tell
+    # bazel that this is `cfg = "exec"` and making it its own dep line solves the problem.
+    "build_script": attr.label(
+        doc = dedent("""\
+            The `cargo_build_script` dependency, if any, to run as part of building this target.
+        """),
+        cfg = "exec",
```
UebelAndre, line 571: This feels incorrect to me. The build script will be executed on `exec` but it's outputs need to be for the target platform. I still think I'm confused why the current implementation is not sufficient where the `cargo_build_script` rule takes the build script itself as `cfg = "exec"` and the outputs are consumed as `cfg = "target"`.

UebelAndre (changes requested):

rust/private/rust.bzl
durin42: If you specify it as `cfg = "target"` then it gets interpreted as needing to be _executable_ on the target, which isn't true. AFAIK there's no way to specify "runs on exec but produces outputs for target" in bazel, but @scentini or @krasimirgg should confirm.

UebelAndre: > If you specify it as cfg = "target" then it gets interpreted as needing to be executable on the target

That's not correct. The build script is made `exec` due to https://github.com/bazelbuild/rules_rust/blob/0.10.0/cargo/cargo_build_script.bzl#L278-L284. The outputs of the build script I don't think are intended to be for the execution platform so setting this to `exec` would be incorrect in cases like [openssl-sys](https://github.com/sfackler/rust-openssl/commits/master/openssl-sys/build/main.rs) where the build script is trying to find OpenSSL libraries to link against, right?

krasimirgg: Last two comments both sound correct to me. Let me try to express my viewpoint to hopefully clarify something or else learn why and how it's wrong: there's a distinction between what platform the build script is running on (and built for), and what platform the outputs of the build script are targeting. Suppose we are cross-compiling on an x86 machine and we are aiming to produce final rust libraries and binaries for arm. The build script itself is ran on the x86 machine and produces outputs that are specific to the arm final target. When building the build script itself, all of its build dependencies need to be built in x86 mode in order for them to be runnable on the x86 machine (and the `cfg = exec` on the build script attribute expresses this constraint). The build script is free to output artifacts that contain instructions for any platform -- the `cfg` only controls where it's supposed to run, not what platform its outputs are supposed to be for. So yes, the outputs of the build script will contain the arm-specific definitions that will then be fed up to the rustc cross-compilation invocation (again running on x86) to produce final rlibs and bins that are for arm. In a sense, the outputs of the build script are "target-independent", but they contain target-specific instructions inside.

UebelAndre: I think we're on the same page but I'm still not sure I understand why this attribute should be `cfg = "exec"`. All actions consumed as `cfg = "target"` still have their actions run on some `exec` platform. But the outputs are intended to be run on the `target` platform. Is there a build script that outputs some binary that's loaded or invoked by a call to `rustc`? I otherwise still think `cfg = "target"` is correct.

PR #1539, opened 2022-09-05
Serialize output from cargo not cargo-lockfile

### LINKED ISSUE #1535: crate_universe adds registry+https://github.com/rust-lang/crates.io-index to Cargo.lock
I recently updated from an old crate_universe to the one in the latest rules_rust release. Most things seem to be working well, but I've run into an issue with Cargo.lock being changed.

Each time I run `REPIN=workspace bazel build`, the Cargo.lock file is updated to add explicit references to the default registry, eg:

```diff
@@ -118,7 +118,7 @@ checksum = "7d6083ccb191711e9c2b80b22ee24a8381a18524444914c746d4239e21d1afaf"
 dependencies = [
  "askama_escape",
  "humansize",
- "nom 6.1.2",
+ "nom 6.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits",
  "percent-encoding",
  "proc-macro2",
```

After that, I can build with bazel successfully (though it splices twice - once with REPIN defined, and then the next time I run bazel without REPIN as well).

The problem is that the standard cargo does not include the explicit registry, so if I do some editing in Rust Analyzer, or use something like 'cargo add', the Cargo.lock file is updated to remove the registry entries, leaving a dirty file in git. And if I then run 'bazel build' again, it tries to splice again, and then fails reporting that the lock needs repinning.

Is there some way to get the two to agree on one particular format so they don't end up competing with each other?

Semi-related to https://github.com/bazelbuild/rules_rust/issues/1522

illicitonion:
cargo-lockfile uses different heuristics to decide whether to serialize
the source repository.

Instead of trying to get them to generate the same output, let's just
use cargo's output.

This avoids spurious (but safe!) diffs to Cargo.lock when generating
with cargo vs cargo-bazel.

Fixes #1535.


UebelAndre:
Thanks for working on this! Sorry I've been absent recently things have been pretty busy for me. I'll try to do a full review today but I also feel this is an issue worth raising in https://github.com/RustSec/rustsec/tree/main/cargo-lock if the core issue is that this serializes differently than Cargo does. If it's possible to have a bug fix there then I think we could avoid some extra plumbing.

jfirebaugh:
I'm not sure this is working in all cases. I applied the patches to my local rules_rust, and `REPIN=workspace` commands are still adding `registry+https://github.com/rust-lang/crates.io-index` diffs to Cargo.lock.

I'm using `crates_vendor` if that makes a difference.

illicitonion:
This was fixed upstream in `cargo-lock` via https://github.com/rustsec/rustsec/pull/767 and https://github.com/bazelbuild/rules_rust/pull/1686

PR #1541, opened 2022-09-06
Remove -lgcc from Android builds

keith:
Newer NDKs no longer have this library, if it's still needed for folks
using older NDKs I think they'll have to manually link it.


UebelAndre (approved):
Thanks! Do you think it'd be reasonable to add a small regression test that builds an apk or something to test targeting android? These flags are otherwise dead and I'm not sure any of the other maintainers would be be able to catch regressions for that platform without it :(

PR #1543, opened 2022-09-08
use crate_info.deps in establish_cc_info

krasimirgg:
This avoids the dependence on ctx.attr, allowing custom deps registered via custom transform_deps to be propagated correctly. I believe this should be a non-functional change for rules_rust.
Internally, we have a slightly modified rust_library that has cc_deps attribute which is processed by https://github.com/google/crubit, which posts extra deps via transform_deps. Currently, for obvious reasons, doesn't get processed in establish_cc_info.


UebelAndre (approved):
I think this seems fine. Would you be able to link the [crubit](https://github.com/google/crubit) modified version of the rules so I can better understand this change?

scentini (approved):

PR #1544, opened 2022-09-08
Update lockfile checksums

Wyverald:
After https://github.com/bazelbuild/rules_rust/commit/cb9ca1b817febcaf8edc28d160866be759825c31 went in, I'm getting errors when I run tests in `examples/crate_universe` with [`--incompatible_unambiguous_label_stringification`](https://github.com/bazelbuild/bazel/issues/16196):

```
DEBUG: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/cae3acec4b4a3c8e715b5a1706a73dbf/external/rules_rust/crate_universe/private/generate_utils.bzl:364:14: Digests do not match: Digest("22324143593ee8f6bd791d836d132583b37f45a3d6dd2b2965541a888c421f98") != Digest("1abebc0b516ea71f61b73a4b929d15908034c774c7e364dfb511221aee8c0bbc")
--
  | ERROR: An error occurred during the fetch of repository 'cargo_aliases':
  | File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/cae3acec4b4a3c8e715b5a1706a73dbf/external/rules_rust/crate_universe/private/crates_repository.bzl", line 45, column 28, in _crates_repository_impl
  | repin = determine_repin(
  | File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/cae3acec4b4a3c8e715b5a1706a73dbf/external/rules_rust/crate_universe/private/generate_utils.bzl", line 365, column 13, in determine_repin
  | fail((
  | Error in fail: The current `lockfile` is out of date for 'cargo_aliases'. Please re-run bazel using `CARGO_BAZEL_REPIN=true` if this is expected and the lockfile should be updated.
```

This PR is the result of running the same tests with `CARGO_BAZEL_REPIN=workspace`. However, if I were to submit this PR, running those tests without `--incompatible_unambiguous_label_stringification` would fail. As I'm planning to flip this flag for Bazel 6.0, is there a way to make those tests work with both values of the flag? Particularly, I need some help determining what exactly is causing the digests to be different.

### DIFF
--- a/examples/crate_universe/cargo_aliases/cargo-bazel-lock.json
+++ b/examples/crate_universe/cargo_aliases/cargo-bazel-lock.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "22324143593ee8f6bd791d836d132583b37f45a3d6dd2b2965541a888c421f98",
+  "checksum": "1abebc0b516ea71f61b73a4b929d15908034c774c7e364dfb511221aee8c0bbc",
   "crates": {
     "aho-corasick 0.7.18": {
       "name": "aho-corasick",
--- a/examples/crate_universe/cargo_workspace/cargo-bazel-lock.json
+++ b/examples/crate_universe/cargo_workspace/cargo-bazel-lock.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "7d58a2569ecff202c9bf834d2bba05d02a1270dac40630c0e408216b07b2eb4f",
+  "checksum": "1aa8ae3a3d3c6d658bbc9eae7de85d46e2867654a66e325cb6b972a4d2926c88",
   "crates": {
     "ansi_term 0.12.1": {
       "name": "ansi_term",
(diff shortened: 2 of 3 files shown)


Wyverald:
@illicitonion @UebelAndre any help would be appreciated.

UebelAndre:
The difference appears to be caused by [this line in @rules_rust//crate_universe/private/splicing_utils.bzl](https://github.com/bazelbuild/rules_rust/blob/0.10.0/crate_universe/private/splicing_utils.bzl#L80) returning strings prefixed with `@` when `--incompatible_unambiguous_label_stringification` is enabled and strings without `@` otherwise. This causes a difference in the rendered file which produces a different hash and fails the check you see in your log. The labels there (and may be elsewhere) will need to be sanitized.

Wyverald:
Thanks for the tip. Do you have any more information on how those serialized labels in the splicing manifest are later used (other than being part of the digest computation)? That would help me work out what the best fix here should be.

illicitonion:
I think it's mostly just metadata for error reporting? It should be safe to always serialize those with leading `@`s, something like:

```starlark
def add_leading_at_if_absent(s):
    if not s.startswith("@"):
        return "@" + s
    else:
        return s

manifests = {str(repository_ctx.path(m)): add_leading_at_if_absent(str(m)) for m in repository_ctx.attr.manifests}
```

Wyverald:
Thanks! I sent https://github.com/bazelbuild/rules_rust/pull/1547 instead.

PR #1545, opened 2022-09-08
Add android example

keith:
This is mostly mirrored from https://github.com/keith/bazel-rust-mobile-demo but serves as a test / example to validate changes pushed to this repo. It requires you have an Android NDK / SDK setup to run.


examples/android/WORKSPACE.bazel
```diff
@@ -0,0 +1,50 @@
+workspace(name = "android_examples")
```
UebelAndre, line 1: Does this need to be it's own workspace? I feel we'd get better coverage by having it be a part of `rules_rust_examples`

UebelAndre (changes requested):
Thank you so much! One question but otherwise looks awesome to me 😄

examples/android/WORKSPACE.bazel
keith: the only annoyance is if you want to run those locally, you have to have the android SDK / NDK setup which I imagine most testers won't, I'm happy to merge them if you want though!

illicitonion (approved):
LGTM - thanks!

examples/android/WORKSPACE.bazel
UebelAndre: I think it can be done at a later date but in general, fewer workspaces would be better

UebelAndre (approved):

UebelAndre:
@keith You wouldn't happen to have an iOS example project lying around would you 😄 (Thanks again!)

keith:
Thanks! iOS example here https://github.com/bazelbuild/rules_rust/pull/1546

PR #1546, opened 2022-09-08
Add iOS examples

keith:


keith:
had to do a new CI job here because of https://github.com/bazelbuild/bazel/issues/16099

examples/ios/BUILD.bazel
```diff
@@ -0,0 +1,92 @@
...
+
+rust_library(
+    name = "rust_lib",
+    srcs = ["demo.rs"],
+    edition = "2018",
+    tags = ["manual"],
+    deps = [":allocator_library"],
+)
+
+# This being required is a bug in objc_library, ideally 'main_lib' could depend directly on 'rust_lib'
```
UebelAndre, line 20: Do you know if this is tracked by an upstream issue in Bazel? Or should one be filed for `rules_rust`?

examples/ios/platform_mappings
```diff
@@ -0,0 +1,20 @@
+flags:
```
UebelAndre, line 1: Why are platform mappings needed? Could transitions not be used?

UebelAndre (changes requested):
Thanks! Looks good to me but just had some questions

examples/ios/platform_mappings
keith: the apple rules don't currently support platforms, so the transitions from those rules don't enforce a platform on their dependencies yet

examples/ios/BUILD.bazel
keith: https://github.com/bazelbuild/bazel/pull/15462

UebelAndre: Can we add that the code here? Something like the following maybe?:
```suggestion
# TODO(bazelbuild/bazel#15462): This being required is a bug in objc_library, ideally 'main_lib' could depend directly on 'rust_lib'
```

examples/ios/platform_mappings
UebelAndre: Is there a tracking ticket for this somewhere?

UebelAndre (changes requested):

examples/ios/BUILD.bazel
keith: Adding a TODO to remove, I verified it works without this shim with the latest rolling release

examples/ios/platform_mappings
UebelAndre: Was a ticket made but not linked?

keith: it's on the first line in this file?

UebelAndre: Ah! https://github.com/bazelbuild/rules_apple/issues/1658

Thank you! Sorry for the confusion 😅

UebelAndre (approved):

examples/ios/platform_mappings
keith: npnp <3

PR #1547, opened 2022-09-09
Strip leading '@'s for labels in the splicing manifest

Wyverald:
These labels are only used for error reporting, so it's not necessary for them to be unambiguous.

See https://github.com/bazelbuild/rules_rust/pull/1544 for more background.

Needed to flip https://github.com/bazelbuild/bazel/issues/16196.


UebelAndre (approved):
Thanks!

PR #1549, opened 2022-09-10
Update apple_support

keith:
Nothing in this for rules_rust specifically, but for users using rules_rust + rules_apple, this makes sure they get a version that's compatible with both regardless of WORKSPACE load order.


UebelAndre (approved):
Thanks!

PR #1551, opened 2022-09-14
Handle $(locations) and $(execpaths) expansions

bsilver8192:
Same idea as $(location) and $(execpath), but it's a bit trickier because we don't have the paths until later.

### DIFF
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -238,6 +238,17 @@ def _expand_location(ctx, env, data):
         if directive in env:
             # build script runner will expand pwd to execroot for us
             env = env.replace(directive, "$${pwd}/" + directive)
+    for directive in ("$(execpaths ", "$(locations "):
+        if directive in env:
+            start = env.find(directive)
+            end = env.find(")", start + len(directive))
+            if end == -1:
+                fail("Unterminated %s" % directive)
+
+            # We can't expand these here, because we don't know what they expand
+            # to yet. Instead, mark them for `expand_locations` in
+            # `cargo/cargo_build_script_runner/bin.rs` to deal with.
+            env = env[:start] + "$${multipwd " + env[start:(end + 1)] + "}" + env[(end + 1):]
     return ctx.expand_make_variables(
         env,
         ctx.expand_location(env, data),
--- a/util/process_wrapper/options.rs
+++ b/util/process_wrapper/options.rs
@@ -51,6 +51,7 @@ pub(crate) fn options() -> Result<Options, OptionError> {
     // Process argument list until -- is encountered.
     // Everything after is sent to the child process.
     let mut subst_mapping_raw = None;
+    let mut subst_multipwd = None;
     let mut stable_status_file_raw = None;
     let mut volatile_status_file_raw = None;
     let mut env_file_raw = None;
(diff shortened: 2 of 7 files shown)


bsilver8192:
@UebelAndre are you the right person to review this?

UebelAndre:
> @UebelAndre are you the right person to review this?

I can take a look this weekend. Sorry for the delay! 😅

bsilver8192:
> I can take a look this weekend. Sorry for the delay! sweat_smile

@UebelAndre friendly ping?

UebelAndre:
> > I can take a look this weekend. Sorry for the delay! sweat_smile
> 
> @UebelAndre friendly ping?

You mean to tell me you don't have 11 day weekends?! 😅 (jk)

Thanks for the ping. This one totally slipped my mind. I'll try to do a this week.

util/process_wrapper/options.rs
```diff
@@ -63,6 +64,7 @@ pub(crate) fn options() -> Result<Options, OptionError> {
     let mut rustc_output_format_raw = None;
     let mut flags = Flags::new();
     flags.define_repeated_flag("--subst", "", &mut subst_mapping_raw);
+    flags.define_flag("--subst-multipwd", "", &mut subst_multipwd);
```
UebelAndre, line 67: Can you explain the thought behind the name `multipwd` and why an additional `--subst` flag would not be sufficient?

UebelAndre (changes requested):
One question to try and fully understand this PR

util/process_wrapper/options.rs
bsilver8192: I don't have strong opinions on the name, open to alternatives.

I don't see how to go from `a b c` to `somewhere/a somewhere/b somewhere/c` with a single substitution. It has to operate on each of the paths separately, and Starlark can't see each path separately without fully reimplementing `$(locations)` expansion.

rust/private/utils.bzl
```diff
@@ -238,6 +238,17 @@ def _expand_location(ctx, env, data):
         if directive in env:
             # build script runner will expand pwd to execroot for us
             env = env.replace(directive, "$${pwd}/" + directive)
+    for directive in ("$(execpaths ", "$(locations "):
```
UebelAndre, line 241: What about the singular variant of these?

util/process_wrapper/options.rs
UebelAndre: The idea that there's multiple "present working directories" I think is what makes me defensive of the change. Why not pass multiple `--subst` values?

UebelAndre (changes requested):

rust/private/utils.bzl
bsilver8192: I left the existing handling of those alone, a few lines above this.

util/process_wrapper/options.rs
bsilver8192: Ah, I see the confusion from the name: it's a single working directory, substituted into multiple paths.

bsilver8192:
@UebelAndre I'm happy to rename things if you want, but I think the problem is you don't see why this is necessary yet. I don't see how to make `//some_filegroup` into something that process_wrapper can translate to `somewhere/a somewhere/b somewhere/c` with a single `--subst`. The starlark macro can't know how what paths the label expands to, so it can't leave a simple substitution for process_wrapper.

I'll rebase this once we agree on what it should look like.

bsilver8192:
@UebelAndre any thoughts on this one?

PR #1559, opened 2022-09-20
Set CARGO_MANIFEST_DIR at runtime for tests

bsilver8192:
As documented at
https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates, cargo tests this both when compiling and when running tests.


bsilver8192:
@scentini are you the right person to review this?

illicitonion (approved):
Thanks!

PR #1563, opened 2022-09-23
Remap $PWD to empty string instead of "."

goffrie:
This fixes an inconsistency in file paths (included in e.g. panic info), wherein some paths would have a leading `./` and some would not.

I believe specifically generic code that is instantiated across a crate boundary would end up with an absolute path, then have it remapped to `.`, whereas other code would use the literal path passed on the command line to `rustc`, which is generally a relative path with no `./` prefix.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -708,7 +708,7 @@ def construct_arguments(
         force_all_deps_direct = False,
         force_link = False,
         stamp = False,
-        remap_path_prefix = ".",
+        remap_path_prefix = "",
         use_json_output = False,
         build_metadata = False,
         force_depend_on_objects = False):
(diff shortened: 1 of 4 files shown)


krasimirgg:
Could this yield rewritten paths that look like absolute paths, e.g., if `$pwd` is `/path/to/dir` and a relative path in a message is to `subdir/lib.rs`, replacing `$pwd` with the empty string could yield messages with a path like `/subdir/lib.rs`, which looks like an absolute path and is confusing IMO.

goffrie:
In my testing, rustc is smart and does not do that (it strips the leading
slash).

krasimirgg:
Looks good then! @goffrie could you address this PR's google-cla check?

goffrie:
Done.

UebelAndre:
> In my testing, rustc is smart and does not do that (it strips the leading slash).
> […](#)

Could this be demonstrated by a test?

goffrie:
I added a test and manually verified that it fails without this change (the path has a leading `./`).

krasimirgg (approved):
Thank you!

PR #1564, opened 2022-09-25
Fix build scripts targeting the wrong architecture

dae:
PR #1081 started passing CFLAGS to build scripts, which cargo does too. But unfortunately Bazel's cpp toolchain adds an explicit target to the CFLAGS it provides, and that target uses the host platform instead of the target platform. When building a crate like zstd on a Mac with iOS as the target, this ends up in the crate being built for macOS instead of iOS.


illicitonion (changes requested):
Thanks for the PR! Is this specific to a particular toolchain you're using? It'd be great to add a test for this (as I can see breaking it in the future), but I don't think we currently have any tests using a custom toolchain... We _do_ have an example which builds an iOS app on macos on CI, which is currently passing, though...

dae:
I've added a small check for this to the examples/ios folder - if the first commit is reverted, the test will fail. Note it only affects build scripts that compile C(PP) code, such as blake3/zstd.

illicitonion:
Thanks! Sorry, I meant an end-to-end example, rather than a targeted unit test - could we e.g. add a `zstd` dependency to the example iOS app? Presumably this should fail to build without your change (hence the need for the change)?

dae:
I've spent about 3 hours trying to get an end-to-end example based on crates_universe working, but don't understand why the library is not available to the sh_test(). Any help would be appreciated.

dae:
Ok, finally figured it out - had to use a filegroup due to https://github.com/bazelbuild/bazel/issues/4519.

illicitonion (approved):
Thanks!

duarten:
This PR breaks toolchains that set CFLAGS and do expect it to work. For example, I have a zig toolchain set up, which I'm using for cross compilation. I need to set the target when creating the `cc_toolchain_config_info`, so it correctly flows to the zig binary when invoked by bazel.

illicitonion:
@duarten Sorry for the breakage! Do you have a test-case you could supply to repro the issue?

When to forward what things from the toolchain is a bit of an awkward dance (which is why @dae added a test-case for the iOS case) - it'd be great to get better coverage here to help us work out exactly when we should/shouldn't be forwarding what - particularly, that would help us to highlight an iOS toolchain bug.

duarten:
I pushed a repro [here](https://github.com/duarten/repro-cflags). To repro, run `bazel build //...`.

It contains a rust program that's using ring. Ring attempts to compile a file in arm assembly, which will immediately cause a build failure. I'm using zig for cross compilation, and I need the `-target` in the cflags to inform zig of the target architecture. Since that switch is being omitted, zig is instructed to compile the assembly file for the exec target, which fails.

illicitonion:
@duarten Thanks for the repro, it's been incredibly useful! I've opened https://github.com/bazelbuild/rules_rust/pull/1663 and asked for some feedback from our Rust+iOS expert in working out exactly what else we need to do to resolve this.

duarten:
Thanks @illicitonion! :)

PR #1566, opened 2022-09-27
process_wrapper: print line on error to parse message as json

krasimirgg:

This is helpful to debug errors when pipelining is involved.

Example error I was hitting:
JsonParseError { msg: "Unexpected character 'l' while parsing 'false'", line: 1, col: 3 } With this change the error includes:
thread 'main' panicked at 'process wrapper error: expected json messages
in pipeline mode while parsing: "fatal runtime error: Rust cannot catch foreign exceptions\n"


bsilver8192:
+1 for this being helpful. However, I ran into it in the copy a few lines down in `stop_on_rmeta_completion`, could you change that one too?

krasimirgg:
> +1 for this being helpful. However, I ran into it in the copy a few lines down in `stop_on_rmeta_completion`, could you change that one too?

Thanks! Done.

scentini (approved):

bsilver8192 (approved):

PR #1569, opened 2022-09-29
Keep transitive linkargs when linking inline rust_test.

mickeyh:
I've locally confirmed that this fixes the issue I observed (for a single instance of this bug).


mickeyh:
Hmmm I prefer not to attach an email to my Github profile, so I cannot sign the CLA. Perhaps someone else can pick up this change (I would prefer if it can be upstreamed)

scentini:
Hey @mickeyh, thanks for the contribution! The changes look good to me; I'd also like to see a test for this before approving.

Not signing the CLA however means that we can neither merge your PR, nor make use of your work when fixing the issue at hand, so I'm going to close it for now. I'll try go get to fixing the issue soon, if noone else volunteers.

PR #1575, opened 2022-10-01
Fixes crates_vendor labels in remote mode when used from the root workspace

### LINKED ISSUE #1514: `crates_vendor` in remote mode uses incorrect labels when used from the root workspace
Note: this is a different problem than #1341.

When using crates_vendor in remote mode from the root workspace, incorrect labels are used in the `defs.bzl` and `crates.bzl` generated files. Here are two examples:

defs.bzl:

```
maybe(
    http_archive,
    name = "crates__log-0.4.17",
    sha256 = "abb12e687cfb44aa40f41fc3978ef76448f9b6038cad6aef4259d3c095a2382e",
    type = "tar.gz",
    urls = ["https://crates.io/api/v1/crates/log/0.4.17/download"],
    strip_prefix = "log-0.4.17",
    build_file = Label("@__main__//third_party/crates:BUILD.log-0.4.17.bazel"),
)
```

crates.bzl:

```
load("@__main__//third_party/crates:defs.bzl", _crate_repositories = "crate_repositories")

def crate_repositories():
    maybe(
        crates_vendor_remote_repository,
        name = "crates",
        build_file = Label("@__main__//third_party/crates:BUILD.bazel"),
        defs_module = Label("@__main__//third_party/crates:defs.bzl"),
    )

    _crate_repositories()
```

Something like the following fixes things for me locally:

```diff
diff --git a/crate_universe/private/crates_vendor.bzl b/crate_universe/private/crates_vendor.bzl
index 021ffc3b..f766d92e 100644
--- a/crate_universe/private/crates_vendor.bzl
+++ b/crate_universe/private/crates_vendor.bzl
@@ -130,24 +130,43 @@ def _write_config_file(ctx):
     output_pkg = _get_output_package(ctx)
 
     if ctx.attr.mode == "local":
-        build_file_base_template = "@{}//{}/{{name}}-{{version}}:BUILD.bazel"
+        build_file_template = "@{}//{}/{{name}}-{{version}}:BUILD.bazel".format(
+            ctx.workspace_name,
+            output_pkg
+        )
+
         crate_label_template = "//{}/{{name}}-{{version}}:{{target}}".format(
             output_pkg,
         )
-    else:
-        build_file_base_template = "@{}//{}:BUILD.{{name}}-{{version}}.bazel"
-        crate_label_template = rendering_config["crate_label_template"]
 
-    rendering_config.update({
-        "build_file_template": build_file_base_template.format(
+        crates_module_template = "@{}//{}:{{file}}".format(
             ctx.workspace_name,
             output_pkg,
-        ),
-        "crate_label_template": crate_label_template,
-        "crates_module_template": "@{}//{}:{{file}}".format(
+        )
+    elif ctx.workspace_name != '__main__':
+        build_file_template = "@{}//{}:BUILD.{{name}}-{{version}}.bazel".format(
+            ctx.workspace_name,
+            output_pkg
+        )
+
+        crate_label_template = rendering_config["crate_label_template"]
+
+        crates_module_template = "@{}//{}:{{file}}".format(
             ctx.workspace_name,
             output_pkg,
-        ),
+        )
+    else:
+        build_file_template = "//{}:BUILD.{{name}}-{{version}}.bazel".format(
+            output_pkg
+        )
+
+        crate_label_template = rendering_config["crate_label_template"]
+        crates_module_template = "//{}:{{file}}".format(output_pkg)
+
+    rendering_config.update({
+        "build_file_template": build_file_template,
+        "crate_label_template": crate_label_template,
+        "crates_module_template": crates_module_template,
         "vendor_mode": ctx.attr.mode,
     })
```

csmulhern:
Fixes #1514.


crate_universe/private/crates_vendor.bzl
```diff
@@ -129,23 +129,27 @@ def _write_config_file(ctx):
 
     output_pkg = _get_output_package(ctx)
 
+    label_prefix = ""
+    if ctx.workspace_name != "__main__":
```
UebelAndre, line 133: When would the workspace name ever be `__main__`? Isn't that a user's decision?

UebelAndre (changes requested):

crate_universe/private/crates_vendor.bzl
csmulhern: __main__ is the `workspace_name` value for the “main” repository (i.e. a non-external repository). Labels for the current repository are referred to with `//path/…`. Labels for an external repository are referred to with `@<workspace_name>//path/…`. Labels for the main repository are referred to with `@//path/…`. The issue here is that the existing code can only generate correct labels for targets in an external repository. If the repository where the aliases are being generated is the main repository, the labels should be of the form `@//path/…` or `//path/…`. `__main__` is simply a hardcoded workspace_name value within Bazel that’s used as a placeholder value for the main repository. It cannot be used in labels to refer to targets in the main repository.

UebelAndre: If I print add `print(ctx.workspace_name)` in a rule in this repo, I get:
```
DEBUG: /Users/user/Code/rules_rust/rust/private/rust.bzl:336:10: rules_rust
```

I've never seen `__main__` before as a unique value. Would that be related to [bzlmod](https://bazel.build/build/bzlmod)?

UebelAndre (changes requested):

crate_universe/private/crates_vendor.bzl
csmulhern: No, not related to bzlmod. I believe you see rules_rust rather than the placeholder because you explicitly name your workspace. See: https://github.com/bazelbuild/rules_rust/blob/7ffe0a5556a35f8c11bf9a9ae8bc4020dd44ea13/WORKSPACE.bazel#L1.

csmulhern: This is also the name that is used by bazel for the repo within the output base. External repositories are at `external/@<name>`. The "main" repository is at `execroot/<name>`, or if there is no name, `execroot/__main__`. I believe external repositories must have a name, because the only way to add them to a non-external repository is to use a rule that requires a name to be specified.

UebelAndre: Ah, so this feature is for cases where the workspace is **not** named? Does `@//path` **always** refer to the root/main workspace? For some reason I thought that would expand to `@external_workspace//path` if it was loaded from `external_workspace`.

UebelAndre (changes requested):

crate_universe/private/crates_vendor.bzl
csmulhern: A quick search (https://grep.app/search?q=__main__&filter[repo][0]=bazelbuild/bazel) shows some hits within documentation inside bazel that refer to `__main__` as the "default workspace name".

I suppose the implication here is... when the "main workspace" has been given a name (e.g. "rules_rust"), can you refer to targets within it with labels of the form: "@rules_rust//...". In that case, this change would be sufficient (we can optionally use `@//` labels when the workspace name is `__main__` as well, to be a little more standard). If that's not the case, we need a better check than `workspace_name == "__main__"`.

csmulhern: From some quick testing, it looks like `@rules_rust//...` is acceptable.

> Ah, so this feature is for cases where the workspace is **not** named? Does `@//path` **always** refer to the root/main workspace? For some reason I thought that would expand to `@external_workspace//path` if it was loaded from `external_workspace`.

From my understanding, yes.

See: https://bazel.build/concepts/labels.

> Labels starting with @// are references to the main repository, which will still work even from external repositories.

UebelAndre: I think accounting for the case where a workspace is unnamed is good. I don't think we should replace the use of `@<workspace_name>//` with `@//` since that breaks the ability to vendor crates for use **as** an external repository. Take `rules_rust` as an example, we have the `cargo_bazel` binary which runs in workspaces that consume `rules_rust`. When users add the following to their WORKSPACE file:
```starlark

crate_universe_dependencies()
```

The repositories created there should use build files specifically from `@rules_rust//crate_universe/3rdparty/...` But if the files were rendered as `@//crate_universe/3rdparty/...`, that'd break the external dependencies... Right?

UebelAndre (changes requested):

crate_universe/private/crates_vendor.bzl
csmulhern: I see your point.

In this case, the `workspace_name` would be `rules_rust`, and so the files would continue to be rendered with labels of the form `@rules_rust//crate_universe/3rdparty/...`.

The heart of the problem is that the labels in crates.bzl / defs.bzl in the repository need to work regardless of if they are referenced within the workspace, or as an external repository.

Note: Is it possible that there is a disconnect between the workspace name used to generate these files (e.g. the workspace name specified in WORKSPACE.bazel), and the workspace name when this workspace is imported as an external repository (e.g. when importing rules_rust via `http_archive`, I give it a different name). I'm guessing if this is possible, than this configuration would just be broken across the board and thus not supported.

Cases:

1) The workspace is named (e.g. `rules_rust`), the labels use the workspace name (e.g. `@rules_rust//...`), and the macros are consumed from **within** the workspace.

This is supported today, and will continue to function with this change.

2) The workspace is named (e.g. `rules_rust`), the labels use the workspace name (e.g. `@rules_rust//...`), and the macros are consumed from **outside** the workspace.

This is supported today, and will continue to function with this change.

3) The workspace is unnamed (so has the default name `__main__`), the labels use the workspace name (e.g. `@//...`), and the macros are consumed from **within** the workspace.

This is not the case today. Instead, the labels generated are of the form `@//__main__//...` and bazel complains upon parsing them. This change makes this use case supported.

3) The workspace is unnamed (so has the default name `__main__`), the labels use the workspace name (e.g. `@//...`), and the macros are consumed from **outside** the workspace.

This scenario would not work, as `@//` would refer to the "main" repository, rather than the workspace containing the rules. However, I don't think this scenario is possible. I don't think you could have an unnamed workspace that is also an external repository.

crate_universe/private/crates_vendor.bzl
```diff
@@ -129,6 +129,10 @@ def _write_config_file(ctx):
 
     output_pkg = _get_output_package(ctx)
 
+    workspace_name = ""
```
UebelAndre, line 132: Minor nit: Could this instead be
```suggestion
    workspace_name = ctx.workspace_name
    if workspace_name == "__main__":
        workspace_name = ""
```

crate_universe/private/crates_vendor.bzl
UebelAndre: That all sounds good to me, thanks for breaking everything down for me!

UebelAndre (changes requested):
One minor nit but otherwise looks good to me!

crate_universe/private/crates_vendor.bzl
csmulhern: Done.

UebelAndre (approved):
Thank for working on this!

csmulhern:
No worries. Does CI have good coverage of all the crates_universe configurations? I have done some manual testing, but cannot be sure I've covered all scenarios.

UebelAndre:
> No worries. Does CI have good coverage of all the crates_universe configurations? I have done some manual testing, but cannot be sure I've covered all scenarios.

You would likely need to make an example workspace to test this. If you wanted to add a test you could add a similar CI job to:
https://github.com/bazelbuild/rules_rust/blob/da3d522d59becbec5697dac0b2aa39890656ed4b/.bazelci/presubmit.yml#L318-L327

From there you'd make a new workspace in `examples/crate_universe_unnamed` or something where `crate_vendor` would be run and some rust_test run to demonstrate code could be correctly downloaded, built, and run.

csmulhern:
I've added crate_universe_unnamed. An example unnamed workspace that copies the vendor_remote_manifests and vendor_remote_pkgs examples. Tests of these examples have been added to the GitHub actions presubmit. Let me know your thoughts.

csmulhern:
The duplication is unfortunate.

UebelAndre:
> The duplication is unfortunate.

Unfortunate indeed. I can see what I can do about cleaning this up.

PR #1580, opened 2022-10-03
support rust-analyzer-proc-macro-srv

### LINKED ISSUE #1576: proc_macros do not work with RA on 1.64
Per https://github.com/rust-lang/rust-analyzer/issues/13322 for 1.64 we need to provide sysroot in rust-project.json + set up the symlinks for `rust-analyzer-proc-macro-srv` in order for the proc_macros to work out of the box with rules_rust's RA integration

A temporary workaround for anyone that run into this is specifying the proc macro server in the RA config, e.g. 

```
    "rust-analyzer.procMacro.server": "/Users/snowp/.rustup/toolchains/1.64-aarch64-apple-darwin/libexec/rust-analyzer-proc-macro-srv",

```

I got something that I think works, will get a PR up

snowp:
Starting in 1.64 RA relies on a rust-analyzer-proc-macro-srv binary provided by the toolchain to handle proc macros, as a way to avoid ABI breakages when RA and proc macros are compiled with a different version of Rust. 

Normally (in cargo-land) RA would automatically pick this up from sysroot, but when generating rust-project.json we'll have to now explicitly specify the sysroot there to help RA resolve the new binary.

This PR does two things in order to support this:

1. Ensure that rust-analyzer-proc-macro-srv is symlinked into the sysroot we build (for Rust 1.64 and higher)
2. Writes the sysroot path to rust-project.json

Fixes #1576

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -640,6 +640,11 @@ rust_toolchain = rule(
             doc = "The operating system for the current toolchain",
             mandatory = True,
         ),
+        "rust_analyzer_proc_macro_srv": attr.label(
+            doc = "The location of the rust_analyzer_proc_macro_srv binary.",
+            allow_single_file = True,
+            cfg = "exec",
+        ),
         "rust_doc": attr.label(
             doc = "The location of the `rustdoc` binary. Can be a direct source or a filegroup containing one item.",
             allow_single_file = True,
@@ -657,11 +662,6 @@ rust_toolchain = rule(
             allow_single_file = True,
             cfg = "exec",
             mandatory = True,
-        ),
-        "rust_analyzer_proc_macro_srv": attr.label(
-            doc = "The location of the rust_analyzer_proc_macro_srv binary.",
-            allow_single_file = True,
-            cfg = "exec",
         ),
         "rustc_lib": attr.label(
             doc = "The libraries used by rustc during compilation.",
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -305,7 +305,8 @@ def _generate_sysroot(
         clippy = None,
         llvm_tools = None,
         rust_std = None,
-        rustfmt = None):
+        rustfmt = None,
+        rust_analyzer_proc_macro_srv = None):
     """Generate a rust sysroot from collection of toolchain components
 
     Args:
@@ -318,6 +319,7 @@ def _generate_sysroot(
         llvm_tools (Target, optional): A collection of llvm tools used by `rustc`.
         rust_std (Target, optional): A collection of Files containing Rust standard library components.
         rustfmt (File, optional): The path to a `rustfmt` executable.
+        rust_analyzer_proc_macro_srv (File, optional): The path to a `rust_analyzer_proc_macro_srv`.
 
     Returns:
         struct: A struct of generated files representing the new sysroot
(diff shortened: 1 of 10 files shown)


snowp:
@UebelAndre I think this is in a decent state for review, are you the right person to do so?

UebelAndre:
I will take a look as soon as I can but since I've been short of free time lately, I'm wondering if @hlopko would be able to take a look since in the past I believe they've been pretty up to date on rust-analyzer things.

rust/toolchain.bzl
```diff
@@ -628,6 +640,11 @@ rust_toolchain = rule(
             doc = "The operating system for the current toolchain",
             mandatory = True,
         ),
+        "rust_analyzer_proc_macro_srv": attr.label(
```
UebelAndre, line 643: My two cents about this attribute is it should either be on `rust_analyzer_toolchain` or maybe it doesn't need to exist at all. This change seems to add the ability to track the binary as part of the toolchain but it's ultimately not used by Bazel? Rust-analyzer works outside of Bazel so as long as the sources exist on disk in your output user root, the binary would be usable? Are any of my questions correct here? 😅

UebelAndre (changes requested):

rust/toolchain.bzl
snowp: IIUC the whole idea of this new binary is that it should track the toolchain used to compile the proc_macro dylibs so that RA doesn't break if the proc_macro ABI changes between Rust versions (e.g. you use nightly RA but are building an older stable version).

With that, my thinking was that we want to make sure that when we construct the rust-project.json file the sysroot specified will point to a proc-macro-srv that matches the Rust toolchain used to compile the proc macro dylibs, which lead me to this where we wire up the proc-macro-srv present in the Rust toolchain being used. It's true that we don't need to place the proc-macro-srv in execroot for any of this to work if we had the true sysroot, but it wasn't clear to me how I'd extract the real location of the toolchain sysroot from within Skylark, though maybe there is a way? 

It might be possible to get this working by putting it on the RA toolchain, but to me it seemed like it would then be defined independent of the Rust toolchain and thus not linked to the rustc version used to compile the proc_macro dylibs?

UebelAndre: I think it's fair to point rust_analyzer at a sysroot but unfortunately in Bazel it's not quite the same as for Cargo. The sysroot would likely be a `rust_toolchain_repository` which only contains components used for building on one exec platform to one target platform. The sysroot would change depending on what you were targeting. Would that be an issue for `rust-analyzer-proc-macro-svr`? It kinda seems like it wouldn't be given the binary comes with the `rustc` components which would largely be same across all these sysroots.

tools/rust_analyzer/rust_project.rs
```diff
@@ -89,11 +92,13 @@ pub struct Dependency {
...
 
 pub fn generate_rust_project(
-    sysroot_src: &str,
+    sysroot_path: &str,
+    sysroot_src_path: &str,
     crates: &BTreeSet<CrateSpec>,
 ) -> anyhow::Result<RustProject> {
     let mut project = RustProject {
-        sysroot_src: Some(sysroot_src.into()),
+        sysroot: Some(sysroot_path.into()),
```
UebelAndre, line 100: Also, `sysroot` is a supported flag? I'm not seeing it in https://rust-analyzer.github.io/manual.html#non-cargo-based-projects

UebelAndre (changes requested):

UebelAndre:
Closing this out since https://github.com/bazelbuild/rules_rust/pull/1641 was merged. Thanks so much for working on this @snowp!

PR #1582, opened 2022-10-03
fix: incorrect rustfmt edition query

### LINKED ISSUE #1579: @rules_rust//:rustfmt deduces incorrect edition
When I run
```
bazel run @rules_rust//:rustfmt
``` 
I get a lot of errors of the following kind:

```
error[E0670]: `async fn` is not permitted in Rust 2015
   --> /Users/lifted/Projects/dfinity/rs/transport/src/utils.rs:303:12
    |
303 | pub(crate) async fn connect_to_server(
    |            ^^^^^ to use `async fn`, switch to Rust 2018 or later
    |
    = help: set `edition = "2021"` in `Cargo.toml`
    = note: for more on editions, read https://doc.rust-lang.org/edition-guide
```

Even though I have edition 2021 configured by default and configure a custom `rustfmt.toml` file with the following contents:
```toml
edition = "2021"
```

It seems that the tool is trying to guess editions by running a query like the following one:
```
bazel query 'let scope = set("//...") in filter("^//.*\.rs$", kind("source file", deps(attr(edition, "2015", $scope) except attr(tags, "(^\[|, )(no-format|no-rustfmt|norustfmt)(, |\]$)", $scope) + attr(crate, ".*", $scope), 1)))'
```

There seems to be a bug in this query because it reports sources of packages that don't use edition 2015.

roman-kashitsyn:
This change fixes incorrect query gathering rust targets with an explicitly specified edition.

The comment above the query states the following intention:

> Get all targets with the specified `edition` attribute.
> Except for targets tagged with `norustfmt`.
> And except for targets with a populated `crate` attribute since
> `crate` defines edition for this target

However, the actual query is:

```
let scope = set({scope})
in filter("^//.*\.rs$",
   kind("source file",
         deps(attr(edition, "{edition}", $scope)
         except attr(tags, "...", $scope)
         + attr(crate, ".*", $scope), 1)))
```

According to the query grammar[1], `X except Y + Z` parses as `(X except Y) + Z`, while the author's intention was clearly `X except (Y + Z)`. The fix changes the expression to `X except Y except Z` which implements the original intention correctly.

Fixes #1579.

[1]: https://bazel.build/query/language#algebraic-set-operations


tools/rustfmt/src/main.rs
```diff
@@ -78,7 +78,7 @@ fn edition_query(bazel_bin: &Path, edition: &str, scope: &str, current_dir: &Pat
         //             Except for targets tagged with `norustfmt`.
```
krasimirgg, line 78: nit: could you add no-format and no-rustfmt to this comment

roman-kashitsyn, line 78:
```suggestion
        //             Except for targets tagged with `norustfmt`, `no-rustfmt`, or `no-format`.
```

roman-kashitsyn: Done!

krasimirgg (approved):
Thank you!

PR #1583, opened 2022-10-04
Ignore non-utf8 text in build script output

dae:
When compiling the latest zstd crate on a German locale in non-release mode on Windows, it outputs localized messages that are not UTF8, causing the compile to fail.

### DIFF
--- a/cargo/cargo_build_script_runner/lib.rs
+++ b/cargo/cargo_build_script_runner/lib.rs
@@ -103,12 +103,19 @@ impl BuildScriptOutput {
     /// Converts a [BufReader] into a vector of [BuildScriptOutput] enums.
     fn outputs_from_reader<T: Read>(mut reader: BufReader<T>) -> Vec<BuildScriptOutput> {
         let mut result = Vec::<BuildScriptOutput>::new();
-        let mut line = String::new();
-        while reader.read_line(&mut line).expect("Cannot read line") != 0 {
-            if let Some(bso) = BuildScriptOutput::new(&line) {
-                result.push(bso);
+        let mut buf = Vec::new();
+        while reader
+            .read_until(b'\n', &mut buf)
+            .expect("Cannot read line")
+            != 0
+        {
+            // like cargo, ignore any lines that are not valid utf8
+            if let Ok(line) = String::from_utf8(buf.clone()) {
+                if let Some(bso) = BuildScriptOutput::new(&line) {
+                    result.push(bso);
+                }
             }
-            line.clear();
+            buf.clear();
         }
         result
     }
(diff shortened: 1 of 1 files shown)


cargo/cargo_build_script_runner/lib.rs
```diff
@@ -103,12 +103,17 @@ impl BuildScriptOutput {
     /// Converts a [BufReader] into a vector of [BuildScriptOutput] enums.
     fn outputs_from_reader<T: Read>(mut reader: BufReader<T>) -> Vec<BuildScriptOutput> {
         let mut result = Vec::<BuildScriptOutput>::new();
-        let mut line = String::new();
-        while reader.read_line(&mut line).expect("Cannot read line") != 0 {
-            if let Some(bso) = BuildScriptOutput::new(&line) {
+        let mut buf = Vec::new();
+        while reader
+            .read_until(b'\n', &mut buf)
```
illicitonion, line 108: Could you add a test that if the output doesn't end with a trailing newline, we still parse the last line of output? (i.e. can you remove the trailing newline from the example output in the test?)

Also, could you add a line of invalid utf-8 to the testdata in the test, to verify this change?

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -103,12 +103,17 @@ impl BuildScriptOutput {
...
+            .expect("Cannot read line")
+            != 0
+        {
+            let line = &String::from_utf8_lossy(&buf);
```
illicitonion, line 112: It looks like `cargo` actually completely skips any line which isn't `utf-8`: https://github.com/rust-lang/cargo/blob/0b84a35c2c7d70df4875a03eb19084b0e7a543ef/src/cargo/core/compiler/custom_build.rs#L566-L569

Can you do the same, rather than trying to parse the line?

illicitonion (changes requested):
Wow, this is a fun edge-case :D Thanks for the fix!

cargo/cargo_build_script_runner/lib.rs
dae: Done

cargo/cargo_build_script_runner/lib.rs
dae: Done

illicitonion (approved):
Perfect, thanks!

dae:
Sorry, didn't notice the failed clippy test.

illicitonion:
Sorry, it looks like you need to merge main and then we can get this merged!

dae:
(rebased on main)

Issue #1584, opened 2022-10-04
Pipelined compilation relies on crates building reproducibly


bsilver8192:
I've encountered an issue when enabling `pipelined_compilation`: if a crate doesn't build reproducibly, then the .rmeta and .rlib end up with different hashes, and you get weird errors when trying to use that crate like this:
```
error[E0460]: found possibly newer version of crate `prost_build` which `<internal>` depends on
  --> <internal>/build.rs:14:5
   |
14 |     <internal>::compile_protos(&proto_files, &["../../proto"], |config| {
   |     ^^^^^^^^^^^^^^
   |
   = note: perhaps that crate needs to be recompiled?
   = note: the following crate versions were found:
           crate `prost_build`: /home/brian/.cache/bazel/_bazel_brian/b648ee8a4cf3ea5f702d496242c79348/execroot/com_matician/bazel-out/k8-opt-exec-589A3AC8/bin/external/raze__prost_build__0_10_4/libprost_build-705194240.rlib
           crate `<internal>`: /home/brian/.cache/bazel/_bazel_brian/b648ee8a4cf3ea5f702d496242c79348/execroot/com_matician/bazel-out/k8-opt-exec-589A3AC8/bin/rust/<internal>/lib<internal>-2814067294.rlib
```

This particular case is because `prost-build` embeds absolute paths from environment variables (at https://github.com/tokio-rs/prost/blob/c2d70d962a701131cacdd45da31a8ed5348d43a6/prost-build/src/lib.rs#L1192), but there are other ways to have the same problem.

Setting `disable_pipelining = True` for crates that do this is a usable workaround. I'm not sure if rules_rust can do anything automatically here, or if this just needs to be documented.

(I do plan to fix this so my builds are reproducible eventually, but some people might not care about that, also it was really hard to figure out what the problem was.)

ashi009:
I wonder if we could offer another option for the pipelined build.  Instead of using two actions for both the rmeta and the rib, just emit them together from a single action. As rmeta by itself will help reduce dependency-induced cache invalidation. The current run-twice approach will reduce the max parallel jobs by half, and take more time compared to just waiting for a single action to emit both outputs.

erenon:
This also happens if `env!("CARGO_MANIFEST_DIR")` is in the code, because it expands to an absolute path that contains a random (?) sandbox component. In my case, the error was "can't find crate", especially confusing. Thanks for this ticket!

ParkMyCar:
[...]

PR #1585, opened 2022-10-05
Add an output group for the .rmeta

bsilver8192:
This is similar to the other output groups. It's helpful for debugging. For example, I used it to build the .rmeta multiple times to chase down the root cause of #1584.


krasimirgg (approved):
Thank you!

PR #1586, opened 2022-10-05
Deduplicate CARGO_MANIFEST_DIR logic

illicitonion:
We currently compute this the same in two places - let's have one block of code do so.


illicitonion:
Probably not worth pursuing.

PR #1588, opened 2022-10-05
Expose the output directory from cargo_build_script

bsilver8192:
This is handy for things like I described at
https://github.com/bazelbuild/rules_rust/pull/1571#issuecomment-1264120387, where I needed to depend on the out_dir via other rules.


illicitonion (approved):
Makes sense, thanks!

PR #1594, opened 2022-10-12
Closing parenthesis in `rust_test` example

andrewpollack:


krasimirgg:
This is good! Just run ./docs/update_docs.sh to update the derived markdown docs as well.

Issue #1447, opened 2022-07-11
crate_universe should not modify the manifests provided if they represent a valid cargo workspace


alexeagle:
This rust project
https://github.com/swc-project/swc
has a Cargo.lock file in the root, then a tree of Cargo.toml files for all the crates it builds.

https://gist.github.com/alexeagle/b0a290dd09860cf2db04c6f13025d0aa shows how I've tried to fetch 3p dependencies for this project (I'm trying to build it as an external repository)

I get an error about the lockfile being out-of-date:
```
    Updating crates.io index
error: the lock file /tmp/.tmpDxxX85/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```
but this lockfile is in a 3p project and so I don't want to change it.

On slack @UebelAndre says:
> it seems `crates_repository` doesn't think the manifests you provided are all a part of the same workspace and tries to generate a custom root manifest. You can probably fix this by using your own lockfile but otherwise this is a bug in the splicing phase of the tool.

diceride:
I'm having the same problem when creating a new project from scratch using only the rules_rust ruleset and one single workspace member using rust_library.

Cargo.toml
```toml
[workspace]
members = [
    "foo/bar",
]
```

WORKSPACE
```starlark
crates_repository(
    name = "crate_index",
    cargo_lockfile = "//:Cargo.Bazel.lock",
    manifests = [
        "//:Cargo.toml",
        "//foo/bar:Cargo.toml",
    ],
)

load("@crate_index//:defs.bzl", "crate_repositories")

crate_repositories()
```

```
...
error: the lock file /tmp/.tmpBwan59/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.

Error: Failed to fetch crates for lockfile: exit status: 101
...
```

boxdot:
In my case, the lock has to be updated because the generated `Cargo.toml` contains a special `[package]` section:

```
< name = "direct-cargo-bazel-deps"
< version = "0.0.1"
<
< [[package]]
```

lbruno:
For what it's worth I get the same error message, and I have no diffs between the spliced workspace's Cargo.lock and my top-level Cargo.lock

```
crates_repository(
    name = "crate_index",
    cargo_lockfile = "//:Cargo.lock",
    manifests = [
[...]

PR #1596, opened 2022-10-16
Avoid rendering a mock root package when possible

UebelAndre:
The use of `Cargo.lock` is a relatively new feature (https://github.com/bazelbuild/rules_rust/pull/1426) which did not fully account for cases where `crates_repository` would be consuming `Cargo.lock` files produced by Cargo. This change avoids adding the `direct-cargo-bazel-deps` target to manifests in cases where users did not use [`crates_repository.packages`](https://github.com/bazelbuild/rules_rust/blob/0.11.0/crate_universe/private/crates_repository.bzl#L250-L252), thus allowing Cargo defined lockfiles to be usable by the repository rule.

closes https://github.com/bazelbuild/rules_rust/issues/1447


crate_universe/src/splicing/splicer.rs
```diff
@@ -1317,6 +1365,37 @@ mod test {
...
+                .splice_workspace(&cargo())
+                .unwrap();
+
+        // Check the default resolver version
+        let cargo_manifest = cargo_toml::Manifest::from_str(
+            &fs::read_to_string(workspace_manifest.as_path_buf()).unwrap(),
+        )
+        .unwrap();
+
+        // Due to the addition of directo deps for splicing, this package should have been added to the root manfiest.
```
illicitonion, line 1395:
```suggestion
        // Due to the addition of direct deps for splicing, this package should have been added to the root manfiest.
```

illicitonion (approved):
Nice, thanks!

PR #1600, opened 2022-10-21
Add a build setting to disable the hash prefix on `rust_test` outputs.

jmillikin:
The hash prefix for test outputs was originally added to work around potential filename conflicts between `rust_test` and `rust_binary` on platforms without sandboxed builds (Windows).

However, on platforms that *do* have sandboxed builds, the prefix can be unwanted because it makes path of Rust test binaries different from what would be produced using other language rules such as `cc_test`.

This commit adds a Bazel build setting so that the output prefix can be disabled in cases where it's unwanted.

```
$ bazel build //test/rust:hello_lib_test
Target //test/rust:hello_lib_test up-to-date:
  bazel-bin/test/rust/test-365930506/hello_lib_test
$ bazel build --@rules_rust//rust/settings:rust_test_prefix_output_files=false //test/rust:hello_lib_test
Target //test/rust:hello_lib_test up-to-date:
  bazel-bin/test/rust/hello_lib_test
```

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -15,6 +15,7 @@
 """Rust rule implementations"""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load("//rust/private:common.bzl", "rust_common")
 load("//rust/private:rustc.bzl", "rustc_compile_action")
 load(
@@ -358,6 +359,26 @@ def _rust_binary_impl(ctx):
         ),
     )
 
+def _rust_test_output_name(ctx, toolchain, crate_root):
+    """Generate the output name for a rust_test target.
+
+    Args:
+        ctx (ctx): The ctx object for the current target.
+        toolchain (rust_toolchain): The current `rust_toolchain`
+        crate_root (File): The test's crate root source file.
+
+    Returns:
+        str: The output name for a rust_test binary.
+    """
+    prefix_output_files = ctx.attr._prefix_output_files[BuildSettingInfo].value
+
+    output_name = "{}{}".format(ctx.label.name, toolchain.binary_ext)
+    if not prefix_output_files:
+        return output_name
+
+    output_hash = determine_output_hash(crate_root, ctx.label)
+    return "test-{}/{}".format(output_hash, output_name)
+
 def _rust_test_impl(ctx):
     """The implementation of the `rust_test` rule.
 
@@ -382,14 +403,8 @@ def _rust_test_impl(ctx):
         # Target is building the crate in `test` config
         crate = ctx.attr.crate[rust_common.crate_info] if rust_common.crate_info in ctx.attr.crate else ctx.attr.crate[rust_common.test_crate_info].crate
 
-        output_hash = determine_output_hash(crate.root, ctx.label)
-        output = ctx.actions.declare_file(
-            "test-%s/%s%s" % (
-                output_hash,
-                ctx.label.name,
-                toolchain.binary_ext,
-            ),
-        )
+        output_name = _rust_test_output_name(ctx, toolchain, crate.root)
+        output = ctx.actions.declare_file(output_name)
 
         # Optionally join compile data
         if crate.compile_data:
@@ -428,14 +443,8 @@ def _rust_test_impl(ctx):
             crate_root_type = "lib" if ctx.attr.use_libtest_harness else "bin"
             crate_root = crate_root_src(ctx.attr.name, ctx.files.srcs, crate_root_type)
 
-        output_hash = determine_output_hash(crate_root, ctx.label)
-        output = ctx.actions.declare_file(
-            "test-%s/%s%s" % (
-                output_hash,
-                ctx.label.name,
-                toolchain.binary_ext,
-            ),
-        )
+        output_name = _rust_test_output_name(ctx, toolchain, crate_root)
+        output = ctx.actions.declare_file(output_name)
 
         # Target is a standalone crate. Build the test binary as its own crate.
         crate_info = rust_common.create_crate_info(
@@ -764,6 +773,9 @@ _rust_test_attrs = dict({
         cfg = "exec",
         default = Label("@bazel_tools//tools/cpp:grep-includes"),
         executable = True,
+    ),
+    "_prefix_output_files": attr.label(
+        default = Label("//rust/settings:rust_test_prefix_output_files"),
     ),
 }.items() + _experimental_use_cc_common_link_attrs.items())
 
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -362,6 +362,9 @@ def _rust_binary_impl(ctx):
 def _rust_test_output_name(ctx, toolchain, crate_root):
     """Generate the output name for a rust_test target.
 
+    See `docs/rust_test_output_path_uniqueness.md` for details on why a
+    `rust_test` output path needs special handling.
+
     Args:
         ctx (ctx): The ctx object for the current target.
         toolchain (rust_toolchain): The current `rust_toolchain`
--- a/test/test_prefix_setting/BUILD.bazel
+++ b/test/test_prefix_setting/BUILD.bazel
@@ -1,0 +1,3 @@
+load(":prefix_setting_tests.bzl", "prefix_setting_tests")
+
+prefix_setting_tests(name = "prefix_setting_tests")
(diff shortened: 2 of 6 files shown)


krasimirgg:
This is nice! Just to make sure @jmillikin : is this just for consistency with `cc_test` on supported platforms, and you don't have a workflow that relies on the specifics of the output files layout?

jmillikin:
A little of column A, and a little of column B.

I have a workflow for integration testing using cross-compiled test binaries running in QEMU. To support this without depending on the output file layout would require [`ctx.expand_location()`](https://bazel.build/rules/lib/ctx#expand_location) to support generating "short" paths as used in runfiles, but Bazel doesn't support this[0]. Therefore I am hardcoding the output path of test binaries, which works because `cc_test` has predictable output paths.

[0] That function has a `short_paths=` parameter, but Bazel considers it "Private API" and setting it to `true` raises an exception.

krasimirgg:
could you refresh this PR so the buildkites get re-triggered

krasimirgg:
And could you add a test

jmillikin:
@krasimirgg OK, I rebased on current main HEAD and added a test.

UebelAndre:
If `rust_binary` and `rust_test` are conflicting, wouldn't it make sense to have the hash account for the name of the rule or something so they're always different?

jmillikin:
> If `rust_binary` and `rust_test` are conflicting, wouldn't it make sense to have the hash account for the name of the rule or something so they're always different?

The current behavior (introduced in PR https://github.com/bazelbuild/rules_rust/pull/1434, which you reviewed) is for `rust_test` to have a hash. Per this PR's description, the hash exists as a workaround for non-sandboxed platforms (Windows).

This PR adds the option to remove the hash from outputs, because (again, per PR description) it is unnecessary on platforms with sandboxing and interferes with incremental migration from codebases in other languages (C++).

jmillikin:
Rebased to current `main` HEAD. CI failures appear to be unrelated (caused by bazelci infra issues).

@UebelAndre , it's been about six months since I sent this PR in. Would it be possible to get it reviewed?

UebelAndre:
@scentini @krasimirgg do you know why we have this hash output? Why can't `rules_rust` just write an output matching the name of the label?

rust/private/rust.bzl
```diff
@@ -378,18 +379,17 @@ def _rust_test_impl(ctx):
...
 
-        output_hash = determine_output_hash(crate.root, ctx.label)
-        output = ctx.actions.declare_file(
-            "test-%s/%s%s" % (
-                output_hash,
-                ctx.label.name,
-                toolchain.binary_ext,
-            ),
-        )
+        output_name = "%s%s" % (ctx.label.name, toolchain.binary_ext)
```
UebelAndre, line 388: I personally prefer format strings unless there's a performance reason to do otherwise. Is there?
```suggestion
        output_name = "{}{}".format(ctx.label.name, toolchain.binary_ext)
```

test/test_prefix_setting/BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+)
+
+with_prefix_test_output_files_cfg(
+    name = "test_binary_without_output_hash",
+    testonly = True,
+    srcs = [":test_binary"],
+    prefix_test_output_files = False,
+)
+
+sh_test(
```
UebelAndre, line 30: Do you need `sh_test`? Would it be possible to do this test with the bazel unittest framework? You're just trying to assert on the output path, right?

UebelAndre (changes requested):
Sorry for the delay! Been a busy year!

rust/private/rust.bzl
jmillikin: Done.

test/test_prefix_setting/BUILD.bazel
jmillikin: Just lack of familiarity with the skylib testing system. Updated to replace sh_test with `unittest.bzl%analysistest`.

jmillikin:
> @scentini @krasimirgg do you know why we have this hash output? Why can't `rules_rust` just write an output matching the name of the label?

There's a fairly comprehensive writeup in the PR I linked: https://github.com/bazelbuild/rules_rust/pull/1434

UebelAndre:
> > @scentini @krasimirgg do you know why we have this hash output? Why can't `rules_rust` just write an output matching the name of the label?
> 
> There's a fairly comprehensive writeup in the PR I linked: #1434

Thanks! I'll take a look in a but but in the mean time, could you add some documentation to the new flag that contains a writeup? I think the most ergonomic way to document this would be to use a custom rule instead of the `bazel_skylib` build settings.

jmillikin:
I'm not sure I'm the correct person to write those docs because I don't use Windows, don't have any need for the hash prefix added by @scentini in that PR, and all I want to do here is have an escape hatch to turn it off. I would be perfectly happy with a PR that removed the whole thing.

Would copying chunks of the original PR into a docstring be sufficient?

UebelAndre:
> Would copying chunks of the original PR into a docstring be sufficient?

Yeah probably 😅. Anything we can do to communicate this clearly would be good. I know I've had folks ask me about this directly and it was definitely a challenge to remember and dig up. Having a breakdown in the repo would be awesome!

jmillikin:
@UebelAndre I've added a file `docs/rust_test_output_path_uniqueness.md` that summarizes the issue and links to PR https://github.com/bazelbuild/rules_rust/pull/1434 for folks who want to dig deeper.

jmillikin:
Rebased to current `main`.

@UebelAndre gentle ping. Any remaining blockers?

UebelAndre (changes requested):
Digging into this closer, I don't actually see a reason why `rust_binary` and `rust_test` need to have hashes in the outputs of their names. Why not just delete this behavior and have them always write consistent outputs? In the links below we see the output is based on `ctx.label.name` which should always produce unique outputs, right?

https://github.com/bazelbuild/rules_rust/blob/0.21.1/rust/private/rust.bzl#L329
https://github.com/bazelbuild/rules_rust/blob/0.21.1/rust/private/rust.bzl#L385-L392
https://github.com/bazelbuild/rules_rust/blob/0.21.1/rust/private/rust.bzl#L431-L438

jmillikin:
I believe that @scentini provided detailed steps to reproduce the issue in the original PR, which I linked above, and which you reviewed.

If you want to revert PR #1434 then that would solve *my* problem, but it would leave Windows users out in the cold, and you may want to consider carefully given that in https://github.com/bazelbuild/rules_rust/pull/1427#issuecomment-1171940333 you determined the inter-action file corruption to be a release blocker.

I don't have a Windows dev machine available at the moment, but I can verify that on Linux, with the hashes disabled, build actions for `rust_test` and `rust_binary` rules do generate temporary files with different content that are written to the same path. These are `.d` files -- presumably on Windows, the `.o` files get written to disk instead of piped.

rust_binary:

```
$ bazel build -s --@rules_rust//rust/settings:rust_test_prefix_output_files=false //test/unit/rustdoc:bin_with_transitive_proc_macro --sandbox_debug --spawn_strategy=sandboxed
[...]
1683905179.521713278: src/main/tools/linux-sandbox-pid1.cc:285: working dir: /home/john/.cache/bazel/_bazel_john/bfc02a8f286d82b1e327c569081a7f63/sandbox/linux-sandbox/13/execroot/rules_rust

$ cd /home/john/.cache/bazel/_bazel_john/bfc02a8f286d82b1e327c569081a7f63/sandbox/linux-sandbox/13/execroot/rules_rust
$ shasum bazel-out/k8-fastbuild/bin/test/unit/rustdoc/*.d
800703ec95caef394b403af3da96536ecee58a12  bazel-out/k8-fastbuild/bin/test/unit/rustdoc/bin_with_transitive_proc_macro.d
```

rust_test:

```
$ bazel build -s --@rules_rust//rust/settings:rust_test_prefix_output_files=false //test/unit/rustdoc:bin_with_transitive_proc_macro_test --sandbox_debug --spawn_strategy=sandboxed
[...]
1683905191.591318251: src/main/tools/linux-sandbox-pid1.cc:285: working dir: /home/john/.cache/bazel/_bazel_john/bfc02a8f286d82b1e327c569081a7f63/sandbox/linux-sandbox/14/execroot/rules_rust

$ cd /home/john/.cache/bazel/_bazel_john/bfc02a8f286d82b1e327c569081a7f63/sandbox/linux-sandbox/14/execroot/rules_rust
$ shasum bazel-out/k8-fastbuild/bin/test/unit/rustdoc/*.d
0e3bf6f7752fcdc643bd06a5106d5e4613c48214  bazel-out/k8-fastbuild/bin/test/unit/rustdoc/bin_with_transitive_proc_macro.d
```

jmillikin:
It seems like keeping a perma-fork of `rules_rust` will be a more practical approach moving forward.

PR #1602, opened 2022-10-21
[Crate Universe] Add support for skipped deps

### LINKED ISSUE #1599: Crate Universe feature request: skipping dependencies
Much like my previous issue on here, I'm in the process of moving a repo from using cargo raze over to crate universe, and I've hit one of the stranger packages in our dependencies. We currently (indirectly) depend on a library that generates bindings for a C library in its build.rs file. Because of how it does so, we have to basically replace the whole crate with our own bazel-friendly implementation.

To enable these kinds of scenarios, cargo raze provided a `skipped_deps` option. You can see an excerpt from their readme below:

> In a few cases, the sys crate may need to be overridden entirely. This can be facilitated by removing and supplementing dependencies in the Cargo.toml, pre-generation:

```toml
[package.metadata.raze.crates.sdl2.'0.31.0']
skipped_deps = [
    "sdl2-sys-0.31.0"
]
additional_deps = [
    "@//cargo/overrides/sdl2-sys:sdl2_sys"
]
```

Unfortunately, this usecase is not currently supported by crate universe, making it impossible to implement this crate.

I've started implementing a potential fix, but given I'm not very familiar with the codebase, I wanted to add an issue to track this, and in case someone with a better understanding of this repo beats me to it.

rdelfin:
When incorporating crates using crate_universe, one will every now and then encounter dependencies that can only be resolved by replacing them entirely with a different, bazelified version of a crate. The most common usecase of this are `-sys` crates, which sometimes generate bindings for a C library in some convoluted ways in their `build.rs` file.

To enable these kinds of scenarios, `cargo raze` provides a `skipped_deps` option. You can see an excerpt from their readme below:

> In a few cases, the sys crate may need to be overridden entirely. This can be facilitated by removing and supplementing dependencies in the Cargo.toml, pre-generation:
```toml
[package.metadata.raze.crates.sdl2.'0.31.0']
skipped_deps = [
    "sdl2-sys-0.31.0"
]
additional_deps = [
    "@//cargo/overrides/sdl2-sys:sdl2_sys"
]
```

Unfortunately, this usecase is not currently supported by crate universe, making it impossible to pull in dependencies like `sdl2` or `ibverbs`. This PR contains an implementation of the feature desribed above, letting you skip dependencies of a crate.

This fixes #1599


rdelfin:
I'm gonna close this out in favour of an override solution

PR #1605, opened 2022-10-24
Add `stdlib_deps` to `rust_toolchain`

P1n3appl3:
Currently it's possible to use `stdlib_linkflags` to add things like `-lpthread` when linking against std. However this relies on the toolchain implicitly adding the search dir for that library (iiuc it works because pthread is part of the sysroot). For the [*-fuchsia targets](https://doc.rust-lang.org/nightly/rustc/platform-support/fuchsia.html) the stdlib also depends on the library `fdio` which is _not_ part of the sysroot.

Today to get compilation for fuchsia working you have to add this dependency on fdio manually on _every_ rust target (`deps = if_fuchsia(["@fuchsia_sdk//pkg/fdio"])`), otherwise you'll get `ld.lld: error: unable to find library -lfdio`. To accommodate for this we propose adding a `stdlib_deps` field on the toolchain definition. Its contents will be added as a dep for all rust targets built with that toolchain.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -33,6 +33,7 @@ load(
     "is_exec_configuration",
     "make_static_lib_symlink",
     "relativize",
+    "transform_deps",
 )
 
 BuildInfo = _BuildInfo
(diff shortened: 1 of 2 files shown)


UebelAndre:
Do you know if these dependencies could be added to the target being assigned to target represented by `rust_stdlib_filegroup`?

P1n3appl3:
I originally thought that wasn't possible because fdio comes from a different repository than the rust stdlib and you can't have a filegroup that includes members from multiple repositories. I've since tried it and it seems to actually work:

``` bzl
# In the build file we stick in our rust repository
rust_stdlib_filegroup(
    name = "rust_std",
    srcs = glob(["**/*.rlib"]) + ["@fuchsia_sdk//:arch/x64/dist/libfdio.so"],
)
```

Thanks for the tip.

PR #1611, opened 2022-10-27
Fix ambiguous native dependencies in `proc_macro`s and `staticlib`s

scentini:
`rustc` doesn't know how to deal with native dependencies that have the same name (specifying the full path to the native library does not currently work). `rules_rust` deals with this issue in `bin` and `cdylib` crate types by appending a hash to the different dependencies, so instead of `-lnative-dep -lnative-dep` the linking command line ends up being `-lnative-dep-{hash1} -lnative-dep-{hash2}`. This PR also applies the mechanism to `staticlib` and `proc-macro` crate types.


krasimirgg (approved):
Looks good!

PR #1612, opened 2022-10-28
Renamed `_build_script_run` rule to `cargo_build_script`

UebelAndre:
Today the primary target of a `cargo_build_script` entry in a BUILD file is actually a `_build_script_run` rule. This makes targets hard to query for folks who are not intimately aware of how this functionality is implemented. To reduce confusion for users, the rule seen in their BUILD files should be something they can query for. For example:

```
rules_rust % bazel query 'kind(cargo_build_script, //...)'
//test/build_env:cargo_build_script_env-vars_build_script
//test/cargo_build_script:tools_exec_build_rs
//test/dep_env:read_a
//test/dep_env:read_b
//test/dep_env:read_dep_dir
//test/dep_env:set_a_build
//test/out_dir_in_tests:build_script
//test/transitive_lib:buildscript
//test/unit/transitive_link_search_paths:dep_build_script
//test/unit/transitive_link_search_paths:proc_macro_build_script
Loading: 7 packages loaded
```


illicitonion (approved):
Makes sense to me - thanks!

PR #1616, opened 2022-10-31
[Crate Universe] Add support for package overrides

### LINKED ISSUE #1599: Crate Universe feature request: skipping dependencies

rdelfin:
When incorporating crates using crate_universe, one will every now and then encounter dependencies that can only be resolved by replacing them entirely with a different, bazelified version of a crate. The most common usecase of this are `-sys` crates, which sometimes generate bindings for a C library in some convoluted ways in their `build.rs` file.

To enable these kinds of scenarios, `cargo raze` provides a `skipped_deps` option. You can see an excerpt from their readme below:

> In a few cases, the sys crate may need to be overridden entirely. This can be facilitated by removing and supplementing dependencies in the Cargo.toml, pre-generation:
```toml
[package.metadata.raze.crates.sdl2.'0.31.0']
skipped_deps = [
    "sdl2-sys-0.31.0"
]
additional_deps = [
    "@//cargo/overrides/sdl2-sys:sdl2_sys"
]
```

Unfortunately, this usecase is not currently supported by crate universe, making it impossible to pull in dependencies like `sdl2` or `ibverbs`. This PR contains an implementation of a slightly different form of this feature, where you can provide a dependency override for a given crate. For example, in the case of `ibverbs`, you might choose to override the dependency as follows:

```python
crates_repository(
[...]
    annotations = {
        [...]
        "ibverbs": [crate.annotation(
            gen_build_script = False,
            dep_overrides = {"ibverbs-sys": "@//overrides/ibverbs:ffi"},
        )],
        [...]
    }
    [...]
)
```

This will make all `ibverbs` libraries replace their `ibverbs-sys` dependency with the target `//overrides/ibverbs:ffi` (the `@` at the front tells Bazel the dependency is in our main workspace, as opposed to the internal `crates_repositories` one for this dependency).

This resolves #1599

### DIFF
--- a/crate_universe/src/context/crate_context.rs
+++ b/crate_universe/src/context/crate_context.rs
@@ -396,6 +396,17 @@ impl CrateContext {
                 }
             }
 
+            // Dependency overrides
+            if let Some(dep_overrides) = &crate_extra.dep_overrides {
+                for (crate_name, new_dep) in dep_overrides {
+                    let dep_filter = |dep: &CrateDependency| &dep.id.name == crate_name;
+                    if self.common_attrs.deps.any_matches(dep_filter) {
+                        self.common_attrs.deps.remove_if(dep_filter);
+                        self.common_attrs.extra_deps.insert(new_dep.clone());
+                    }
+                }
+            }
+
             // Compile data glob
             if let Some(extra) = &crate_extra.compile_data_glob {
                 self.common_attrs.compile_data_glob.extend(extra.clone());
--- a/crate_universe/private/crate.bzl
+++ b/crate_universe/private/crate.bzl
@@ -78,6 +78,7 @@ def _annotation(
         data = None,
         data_glob = None,
         deps = None,
+        dep_overrides = None,
         gen_build_script = None,
         patch_args = None,
         patch_tool = None,
@@ -114,6 +115,8 @@ def _annotation(
         data (list, optional): A list of labels to add to a crate's `rust_library::data` attribute.
         data_glob (list, optional): A list of glob patterns to add to a crate's `rust_library::data` attribute.
         deps (list, optional): A list of labels to add to a crate's `rust_library::deps` attribute.
+        dep_overrides (dict, optional): A dictionary of of crate names that should be replaced with a Bazel dependency
+            in the `rust_library::deps` attribute.
         gen_build_script (bool, optional): An authorative flag to determine whether or not to produce
             `cargo_build_script` targets for the current crate.
         patch_args (list, optional): The `patch_args` attribute of a Bazel repository rule. See
@@ -159,6 +162,7 @@ def _annotation(
             data = data,
             data_glob = data_glob,
             deps = deps,
+            dep_overrides = dep_overrides,
             gen_build_script = gen_build_script,
             patch_args = patch_args,
             patch_tool = patch_tool,
--- a/crate_universe/src/config.rs
+++ b/crate_universe/src/config.rs
@@ -150,6 +150,10 @@ pub struct CrateAnnotations {
     /// [deps](https://bazelbuild.github.io/rules_rust/defs.html#rust_library-deps) attribute.
     pub deps: Option<BTreeSet<String>>,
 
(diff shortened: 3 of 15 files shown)


rdelfin:
I'm getting hyper errors in the tests, but they seem unrelated to my change

crate_universe/src/context/crate_context.rs
```diff
@@ -396,6 +396,17 @@ impl CrateContext {
                 }
             }
 
+            // Dependency overrides
+            if let Some(dep_overrides) = &crate_extra.dep_overrides {
+                for (crate_name, new_dep) in dep_overrides {
+                    let dep_filter = |dep: &CrateDependency| &dep.id.name == crate_name;
+                    if self.common_attrs.deps.any_matches(dep_filter) {
+                        self.common_attrs.deps.remove_if(dep_filter);
+                        self.common_attrs.extra_deps.insert(new_dep.clone());
```
illicitonion, line 405: This feels like it should be an atomic replacement, rather than a remove+insert? In particular, it feels like if the old dep was in a select, so should the new dep be, rather than put into an extra_deps list?

I'd suggest instead adding a `replace<T: Clone>(&mut self, old: &T, new: &T)` to `SelectList` which walks all of the relevant `BTreeSet`s doing replacement in each?

illicitonion:
(Also, sorry for such the long delay in getting this reviewed - thanks a lot for the contribution!)

rdelfin:
@illicitonion no worries at all!

You make a good point about having them be a top level override. I've put it into the crate partially for simplicity and to mimic how this works on raze, but it would I can't think of a reason you'd only replace the dependency on one crate. I'll do a second version and re-request a review once it's done

illicitonion:
Amazing, thanks so much!

rbtcollins:
does this patch need someone to pick it up? @rdelfin -  are you still planning on coming back to this?

rdelfin:
Thanks for the ping @rbtcollins! Yeah, I started working on it because I had been maintaining a patch internally for our repo while this got fixed up, but I haven't had the time to dig into it. I'll work on it now that a recent version broke my rebased patch beyond my ability to fix it, but if you're interested in working on it, feel free to! Happy to help too

rbtcollins:
If I may suggest, the way I thought about this before reading this PR was that a crate annotation could inject a custom BUILD rule, but still have the crate source http-archive defined, and still participate in other ways in the dependency graph (getting the same original name and so on). 

That to me feels like a new option on the annotation *of the crate being replaced*

fhilgers:
Does #2674 address the same usecase as this PR?

> Allow external crates to be aliased to local targets

> In some cases it may be desirable to alias a public crate to a locally
define rust_library target, even for dependencies of crates that are not
being overriden.
Support this use case by swapping out alias for rust_library when
a user overrides the target using an annotation.

robin-wayve:
I can confirm that #2674 solves this issue for the codebase that @rdelfin originally posted this for 👍🏻

PR #1617, opened 2022-10-31
Prefix `self` for proto imports

jun-sheaf:
This PR prefixes proto imports with `self`. This eliminates ambiguity of imports.


UebelAndre (changes requested):
Thanks for the pull-request! Is there some kind of regression test you could add for this?

jun-sheaf:
This only matters when using the generated files with cargo, not Bazel.

UebelAndre:
> This only matters when using the generated files with cargo, not Bazel.

What is the error you see? It still feels like something worth having a regression test for if possible.

jun-sheaf:
The 'use' is ambiguous if there is a protobuf file named x containing a message also called x with an enum nested. Since nested enums are namespaced by the message name, 'use x::*' could reference either the enum namespace or the module. the way to fix this is 'self'.

I don't have capacity to build a test.

Issue #1618, opened 2022-11-01
Bazel CI: rules_rust is broken in Bazel@HEAD downstream


SalmaSamy:
[rules_rust](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2715#018432d6-294d-42d6-8634-aaeb76050c85) is failing in the downstream with the following error:
```
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/24b6159646937fb03d5380a181c473a3/external/crates_vendor_pkgs__hyper-0.14.22/BUILD.bazel:27:13: Compiling Rust rlib hyper v0.14.22 (68 files) failed: (Exit 1): process_wrapper failed: error executing command (from target @crates_vendor_pkgs__hyper-0.14.22//:hyper)
--
  | (cd /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/24b6159646937fb03d5380a181c473a3/sandbox/linux-sandbox/908/execroot/examples && \
  | exec env - \
  | CARGO_CFG_TARGET_ARCH=x86_64 \
  | CARGO_CFG_TARGET_OS=linux \
  | CARGO_CRATE_NAME=hyper \
  | CARGO_MANIFEST_DIR='${pwd}/external/crates_vendor_pkgs__hyper-0.14.22' \
  | CARGO_PKG_AUTHORS='' \
  | CARGO_PKG_DESCRIPTION='' \
  | CARGO_PKG_HOMEPAGE='' \
  | CARGO_PKG_NAME=hyper \
  | CARGO_PKG_VERSION=0.14.22 \
  | CARGO_PKG_VERSION_MAJOR=0 \
  | CARGO_PKG_VERSION_MINOR=14 \
  | CARGO_PKG_VERSION_PATCH=22 \
  | CARGO_PKG_VERSION_PRE='' \
  | SYSROOT=bazel-out/k8-fastbuild/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu_tools/rust_toolchain \
[...]

PR #1619, opened 2022-11-01
Fixed CI breakage

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/1618


UebelAndre:
For reviewers, the first commit contains the fix, the second is just regenerating outputs.

scentini (approved):

illicitonion (commented):
Thanks for the fix!

I think you want to remove the `>=`s in all of these version constraints.

`0.14.22` means `>=0.14.22, <1`
`>=0.14.22` means `>=0.14.22` which would allow selecting `1.0.0` if it were released.

I think you want the former?

PR #1620, opened 2022-11-02
Skip adding -lstatic to libtest and libstd on Darwin

### LINKED ISSUE #1573: libtest/libstd being dynamically linked on macos
Presumably related to https://github.com/bazelbuild/rules_rust/issues/1374

On the latest release we're still seeing rustlib being dynamically linked with a rust_binary output (non-trivial project setup, not sure how much this matters), resulting in a binary that cannot be loaded:

```
➜  sdk git:(main) ✗ otool -L bazel-bin/test/client/logger_runner

bazel-bin/test/client/logger_runner:
        @rpath/libtest-5266a2fa65112d20.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libstd-f6811f31090e0239.dylib (compatibility version 0.0.0, current version 0.0.0)
...
```

We've been able to work through this by applying the following patch to rules_rust:

```
# https://github.com/bazelbuild/rules_rust/issues/1374

--- rust/private/repository_utils.bzl
+++ rust/private/repository_utils.bzl
@@ -33,6 +33,7 @@
             "lib/rustlib/{target_triple}/bin/rust-lld{binary_ext}",
             "lib/rustlib/{target_triple}/lib/*{dylib_ext}",
         ],
+        exclude = ["**/*libstd-*{dylib_ext}", "**/*libtest-*{dylib_ext}"],
         allow_empty = True,
     ),
@@ -149,6 +150,7 @@ def BUILD_for_clippy(target_triple):
             "lib/rustlib/{target_triple}/lib/*{staticlib_ext}",
             "lib/rustlib/{target_triple}/lib/self-contained/**",
         ],
+        exclude = ["**/*libstd-*{dylib_ext}", "**/*libtest-*{dylib_ext}"],
         # Some patterns (e.g. `lib/*.a`) don't match anything, see https://github.com/bazelbuild/rules_rust/pull/245
         allow_empty = True,
     ),
```

This unblocks us for now, but I'd love to understand how we can get this fixed for real. Would the approach we're taking here to just exclude the dylibs be a satisfactory solution?

@UebelAndre @scentini @keith

gferon:
This is one way to fix #1573, which I have a feeling can be wrong. The binaries produced with this patch contain all required symbols and are working fine on macOS.


gferon:
Looks like the CI failure is unrelated to this change 🤔

illicitonion (changes requested):
It seems a little weird that we'd express "don't dynamically link libstd" by saying "don't statically link libstd", but, if it seems to work (which CI shows it does), that seems better than the status quo...

Would it be possible to get a couple of tests test showing the desired end-effect here?

1. Our existing binaries all successfully run on CI at the moment - do you have an example that fails without this patch?
2. Could we add a `sh_test` which runs `otool -L` on a produced `rust_binary` and shows that it doesn't dynamically link the libraries we don't expect to be dynamically linked?

gferon:
You're right, this would be much better if the issue was documented and a test would prevent any regression. I'll remove the patch on our codebase and try to come up with a minimal non-working example ASAP!

gferon:
I was able to reproduce the issue with a very simple dependency setup, full workspace available here: https://github.com/gferon/bazel-workspace-rules-rust-pr-1620

The problem is visible when you have a `rust_library` > `cc_library` > `rust_binary`. I don't have any idea why this is a problem, maybe @scentini has some insights?

```starlark
rust_library(
    name = "libtool-rs",
    srcs = ["lib.rs"],
    edition = "2021",
)

cc_library(
    name = "api",
    srcs = ["api.c"],
    deps = [":libtool-rs"],
)

rust_binary(
    name = "app",
    srcs = ["main.rs"],
    edition = "2021",
    deps = [":api"],
)
```

gferon:
@illicitonion since I'm currently working on upstreaming all of our custom patches, I unburried this one from the past. I added a test to demonstrate the issue, and I can confirm the test fails without the patch and passes otherwise.

> Could we add a sh_test which runs otool -L on a produced rust_binary and shows that it doesn't dynamically link the libraries we don't expect to be dynamically linked?

I added a `sh_test` that just runs the test binary, which has the same effect.

test/portable_link_flags_darwin/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -e
+
+$1
```
illicitonion, line 4: Could you instead run `otool -L "$1"` and verify (by grep or whatever) that it doesn't contain "libstd" or "libtest"?

I want us to be sure that we're actually statically linking stuff, rather than setting up rpaths or whatever so that in the bazel-bin tree the binary _happens_ to work but if you moved it somewhere else or copied it to another machine it would fail to.

You can use `target_compatible_with` to make sure this test only runs on darwin where `otool` will be present.

illicitonion (commented):
This looks great, thanks so much! One small comment about the test, then let's get this merged!

illicitonion:
Thanks!

PR #1632, opened 2022-11-05
fix: Fix issue with wasi-0.11.0+wasi-snapshot-preview1

semtexzv:
Handles paths with plus in package name


UebelAndre (changes requested):
Thanks! Do you think you could also add a regression test for this?

UebelAndre (approved):
Awesome! Thank you!

crate_universe/src/utils/starlark/label.rs
```diff
@@ -257,6 +258,14 @@ mod test {
         assert_eq!(label.target, "subdir/target");
     }
 
+    #[test]
+    fn label_contains_plus() {
```
UebelAndre, line 262: Seems this test doesn't pass?
```
---- utils::starlark::label::test::label_contains_plus stdout ----
thread 'utils::starlark::label::test::label_contains_plus' panicked at 'assertion failed: `(left == right)`
  left: `Some("")`,
 right: `None`', crate_universe/src/utils/starlark/label.rs:264:9
```

UebelAndre (changes requested):

UebelAndre (approved):
Thanks!

PR #1635, opened 2022-11-07
Add per-toolchain `rustc_flags`

### LINKED ISSUE #1623: Per-toolchain rustflags
You can already set `stdlib_linkflags` on a `rust_toolchain` so that every target that links to std adds those flags. We have a `rust_toolchain` with a stdlib that's compiled with `Cpanic=abort`, so we need every target that's built with that toolchain to also set panic abort in their _rustflags_.

We initially tried setting this globally by putting `'--@rules_rust//:extra_exec_rustc_flags="-Cpanic=abort"'` in our bazelrc, but we have multiple rust_toolchains in our build, and some of them do have a std built with panic=unwind, so we need a way to apply this flag to only targets built under the specific toolchain that sets panic=abort. Would adding `stdlib_rustflags` to the `ToolchainInfo` and threading it through to the compile actions similar to the linkflags be a reasonable way to do this?

P1n3appl3:
As mentioned in #1623, there are some flags to rustc that you'd like to be able to apply in a per-toolchain fashion rather than globally with the `extra_exec_rustc_flags` config. This adds the field and threads it through to `construct_arguments()`

closes #1623


rust/private/rustc.bzl
```diff
@@ -970,6 +970,8 @@ def construct_arguments(
     if hasattr(ctx.attr, "_extra_exec_rustc_flag") and is_exec_configuration(ctx):
         rustc_flags.add_all(ctx.attr._extra_exec_rustc_flag[ExtraExecRustcFlagsInfo].extra_exec_rustc_flags)
 
+    rustc_flags.add_all(toolchain.rustc_flags)
```
krasimirgg, line 973: I'd like to have 2 toolchain-specific fields, distinguishing between target-specific and exec-specific flags, similarly to _extra_{exec_}_rustc_flags above (say to be able to define a sanitizer-enabled rust_toolchain that uses `-Zsanitizer=address` in target configuration and not in exec configuration). For consistency, let's call them `extra_rustc_flags` and `extra_exec_rustc_flags`. Also let's move them before line 960 so that the user-supplied extra flags appear after the toolchain-supplied ones.
Also it would be nice to come up with a test for these.

test/toolchain/toolchain_test.bzl
```diff
@@ -0,0 +1,128 @@
...
+    return analysistest.end(env)
+
+toolchain_adds_rustc_flags_test = analysistest.make(
+    _toolchain_adds_rustc_flags_impl,
+    attrs = {
+        "lib": attr.label(mandatory = True),
+    },
+    config_settings = {
+        "@//:extra_rustc_flags": [CONFIG_FLAG],
+        # "//command_line_option:extra_toolchains": [":rust_extra_flags_toolchain"],
```
P1n3appl3, line 63: I was also curious if we could avoid the transition by specifying the extra toolchain directly here, but I run into the same issue where it errors because the toolchain hasn't been declared yet.

test/unit/toolchain/toolchain_test.bzl
```diff
@@ -111,6 +155,21 @@ def _toolchain_test():
...
+        os = "linux",
+        rust_doc = ":mock_rustdoc",
+        rust_std = ":std_libs",
+        rustc = ":mock_rustc",
+        staticlib_ext = ".a",
+        stdlib_linkflags = [],
+        extra_rustc_flags = [TOOLCHAIN_FLAG],
+        target_json = ":target_json",
+    )
+
```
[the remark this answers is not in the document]
P1n3appl3: Thanks for the tip! I've tried to make that wrapper rule that invokes the transition, but i'm running into an issue where the transition impl errors out because the toolchain hasn't been defined yet:

```
$ bazel test --test_output=streamed //test/toolchain:toolchain_adds_rustc_flags_test
...
ERROR: /mnt/rules_rust/test/toolchain/BUILD.bazel:3:21: While resolving toolchains for target //test/toolchain:lib: invalid registered toolchain ':rust_extra_flags_toolchain': no such target '//:rust_extra_flags_toolchain': target 'rust_extra_flags_toolchain' not declared in package '' defined by /mnt/rules_rust/BUILD.bazel
```

I'm not sure how to change that ordering, given that the `rust_toolchain` is already defined before the rule that uses it. Also not sure how i'd pass which toolchain i wanted to transition to to avoid hardcoding it in the transition impl, but that's only and issue if we wanted to add more tests without duplicating these definitions.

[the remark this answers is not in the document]
krasimirgg: Try this in `_transition_impl`:
```
return {"//command_line_option:extra_toolchains": ["@rules_rust//test/toolchain:extra_flags_toolchain"]}
```

My understanding is that this function is evaluated from something like a "top-level" context and requires the labels need to be sufficiently fully qualified.

test/toolchain/toolchain_test.bzl
P1n3appl3: as it turns out: yes we can use the `config_settings` from `analysistest` to avoid having to make the transition/wrapper rule.

test/unit/toolchain/toolchain_test.bzl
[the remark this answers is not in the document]
P1n3appl3: ah thanks that was it

P1n3appl3:
The test is passing now, thanks for the help @krasimirgg. Let me know if you'd like any additional changes.

One more note about the test is that I didn't specify any platform constraints for the extra toolchain, so i think that the library defined in the test can be resolved to _either_ the one we passed with `command_line_option:extra_toolchains` or whatever default one it was using before. I'm not sure if it's selecting the one we defined because `extra_toolchains` takes precedence when two toolchains could both match the constraints or if it's just incidental.

krasimirgg:
> The test is passing now, thanks for the help @krasimirgg. Let me know if you'd like any additional changes.
> 
> One more note about the test is that I didn't specify any platform constraints for the extra toolchain, so i think that the library defined in the test can be resolved to _either_ the one we passed with `command_line_option:extra_toolchains` or whatever default one it was using before. I'm not sure if it's selecting the one we defined because `extra_toolchains` takes precedence when two toolchains could both match the constraints or if it's just incidental.

Right, [extra_toolchains takes precedence over the ones registered via the WORKSPACE file](https://bazel.build/reference/command-line-reference#flag--extra_toolchains), so we should be all good.

krasimirgg:
Could you resolve the failing buildkites

P1n3appl3:
Running the test locally passes for me, and I had a couple others pull the change and it passed for them as well. The error it's getting in CI is a bit confusing:

```
ERROR: /workdir/test/toolchain/BUILD.bazel:3:21: While resolving toolchains for target //test/toolchain:lib: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file?
```

I'm not sure what would make it fail to resolve the default cc toolchain, there are other tests that define a `rust_library` and they don't seem to do anything special to make it work.

Also buildifier passes locally with no changes and the buildifier check in CI doesn't have any helpful output, it just fails with `🚨 Error: The command exited with status 4`.

test/toolchain/toolchain_test.bzl
```diff
@@ -0,0 +1,109 @@
...
+        EXEC_TOOLCHAIN_FLAG not in action.argv,
+        "Found exec toolchain flag ({}) in rustc flags: {}".format(EXEC_TOOLCHAIN_FLAG, action.argv),
+    )
+
+    return analysistest.end(env)
+
+toolchain_adds_rustc_flags_test = analysistest.make(
+    _toolchain_adds_rustc_flags_impl,
+    config_settings = {
+        "//command_line_option:extra_toolchains": ["@rules_rust//test/toolchain:extra_flags_toolchain"],
```
krasimirgg, line 42: Did a bit of debugging with @scentini and suspect the issue is that here we're overriding the custom `--extra_toolchains=@buildkite_config//config:cc-toolchain` part that's passed in RBE builds (https://buildkite.com/bazel/rules-rust-rustlang/builds/7469#01848db5-aa45-46dc-b12d-027efe604c8e/237-240).
We could try and either always add this here (not sure what will happen in non-RBE environments) or bring back the user-defined transition where we can do things like: append the custom rust toolchain to the parent `//command_line_option:extra_toolchains` list

P1n3appl3: setting the buildkite toolchain in the config_settings doesn't seem to work locally. I still don't understand where exactly the transition should go.

If I make a wrapper rule that sets `cfg = my_transition` and a target using that rule depends on the `rust_library`, then building the wrapper target will trigger the transition and build the library with the extra configs from the transition. But then the `target_under_test` will be that wrapper target instead of the library, and i can't figure out how to get the actions of the library target in that case.

scentini: PTAL at https://github.com/bazelbuild/rules_rust/blob/main/test/unit/is_proc_macro_dep/is_proc_macro_dep_test.bzl

We create a [`custom rule`](https://github.com/bazelbuild/rules_rust/blob/main/test/unit/is_proc_macro_dep/is_proc_macro_dep_test.bzl#L42) for testing, that applies the [`transition`](https://github.com/bazelbuild/rules_rust/blob/main/test/unit/is_proc_macro_dep/is_proc_macro_dep_test.bzl#L36), and that attaches an [`aspect`](https://github.com/bazelbuild/rules_rust/blob/main/test/unit/is_proc_macro_dep/is_proc_macro_dep_test.bzl#L18) to the `dep` attribute that collects the actions for the dependency (which will be a `rust_library`). The `target_under_test` will be the custom rule we created, and we will inspect the actions collected by the aspect: https://github.com/bazelbuild/rules_rust/blob/main/test/unit/is_proc_macro_dep/is_proc_macro_dep_test.bzl#L128

I know this is all very complicated, but it's the best that the unit testing framework provides at this time.

P1n3appl3: Thanks so much! I never would have thought to use an aspect to collect the action through the transition rule wrapper. Looks like it passes in CI as well now.

krasimirgg (approved):
Looks good! Thank you! (there is one nit in the test)

PR #1638, opened 2022-11-08
[docs] Fixing typos in CargoConfig doc strings

andrewpollack:
Fixing a couple typos I noticed while going through `cargo-vendor`ing workflows


illicitonion (approved):
Thanks!

PR #1641, opened 2022-11-08
Generated rust-project.json files now include sysroot paths

UebelAndre:
Fixes https://github.com/bazelbuild/rules_rust/issues/1576


UebelAndre:
This is an alternate implementation of https://github.com/bazelbuild/rules_rust/pull/1580 which doesn't add any rust-analyzer fields to the core `rust_toolchain`. The somewhat downside of this implementation though is that unlike https://github.com/bazelbuild/rules_rust/pull/1580 I'm updating `rust-analyzer` toolchain to begin to contain it's own sysroot which is defined separately from `rust_toolchain`. I ended up on this implementation due to the fact that `rust_toolchain` does not expose the raw files it ingests and instead, creates a unique sysroot for actions [here](https://github.com/bazelbuild/rules_rust/blob/0.12.0/rust/toolchain.bzl#L444-L454). To my knowledge, the rust-analyzer sysroot is always expected to run on the host so using anything in `execroot` could lead to unexpected behavior. So to have a simple location that won't change out from under rust-analyzer in the `output_user_root` for a workspace, `rust-analyzer` now directly depends on a `rustc` binary. If this is not necessary I think it'd be great to clean up but the thing I like most about this approach is that `rust_toolchain` is left alone.

rust/private/rust_analyzer.bzl
```diff
@@ -215,6 +217,19 @@ rust_analyzer_toolchain = rule(
...
     doc = "A toolchain for [rust-analyzer](https://rust-analyzer.github.io/).",
     attrs = {
+        "proc_macro_srv": attr.label(
+            doc = "The path to a `rust_analyzer_proc_macro_srv` binary.",
+            cfg = "exec",
+            executable = True,
+            allow_single_file = True,
+        ),
+        "rustc": attr.label(
+            doc = "THe path to a `rustc` binary.",
```
freeformstu, line 227:
```suggestion
            doc = "The path to a `rustc` binary.",
```

tools/rust_analyzer/rust_project.rs
```diff
@@ -89,10 +92,12 @@ pub struct Dependency {
 }
 
 pub fn generate_rust_project(
+    sysroot: &str,
     sysroot_src: &str,
     crates: &BTreeSet<CrateSpec>,
 ) -> anyhow::Result<RustProject> {
     let mut project = RustProject {
+        sysroot: Some(sysroot.into()),
```
freeformstu, line 100: Is `sysroot` a valid `rust-project.json` attribute?
I don't see it in the documentation:
https://rust-analyzer.github.io/manual.html#non-cargo-based-projects

UebelAndre: I had the same question but appears to be based on local functional testing! I'll have to file an issue updating the docs to communicate this.

illicitonion (approved):

snowp:
Thanks for taking this over the finish line!

tools/rust_analyzer/rust_project.rs
UebelAndre: https://github.com/rust-lang/rust-analyzer/issues/13608

PR #1645, opened 2022-11-09
Allow missing source annotations for path deps

### LINKED ISSUE #1631: crates_vendor: [patch.crates-io] fails to patch primary/transitive dependencies for `path = "..."`
Using `[patch.crates-io]` in your `Cargo.toml` to patch to local copies of a crate via `CRATE_NAME = { path = "path/to/crate"}` is currently failing with an error of style: `Error: The package 'Name("zerocopy") Version { major: 0, minor: 6, patch: 1 }' has no source info so no annotation can be made`. This fails for patching both primary and transitive dependencies.

This same workflow functions for its `CRATE_NAME = { git = "git_repository"}` equivalent.

Minimal repro:

## Command:
`bazel run //:crates_vendor`

## Directory structure:
```
WORKSPACE
BUILD
Cargo.toml
empty/zerocopy
            Cargo.toml
            src/lib.rs
src/lib.rs
```

`WORKSPACE`
```

http_archive(
    name = "rules_rust",
    sha256 = "696b01deea96a5e549f1b5ae18589e1bbd5a1d71a36a243b5cf76a9433487cf2",
    urls = ["https://github.com/bazelbuild/rules_rust/releases/download/0.11.0/rules_rust-v0.11.0.tar.gz"],
)

rules_rust_dependencies()
rust_register_toolchains()

crate_universe_dependencies()
```

`BUILD`
```
load("@rules_rust//crate_universe:defs.bzl", "crates_vendor")
crates_vendor(
    name = "crates_vendor",
    mode = "local",
    vendor_path = "vendor",
)
```

`Cargo.toml`
```
[package]
name = "my-third-party"
version = "0.1.0"

[dependencies]
zerocopy = "0.6.1"

[patch.crates-io]
zerocopy = { path = 'empty/zerocopy' }

[workspace]
```

`src/lib.rs`
```
// This file is left intentionally blank
```

`empty/zerocopy/Cargo.toml`
```
[package]
name = "zerocopy"
version = "0.6.1"
```

`empty/zerocopy/src/lib.rs`
```
// This file is left intentionally blank
```

djkoloski:
Fixes #1631. Missing source annotations are expected when providing local packages or patching crates that would be vendored.

### DIFF
--- a/crate_universe/src/context/crate_context.rs
+++ b/crate_universe/src/context/crate_context.rs
@@ -247,7 +247,7 @@ impl CrateContext {
     pub fn new(
         annotation: &CrateAnnotation,
         packages: &BTreeMap<PackageId, Package>,
-        source_annotations: &BTreeMap<PackageId, SourceAnnotation>,
+        source_annotations: &BTreeMap<PackageId, Option<SourceAnnotation>>,
         extras: &BTreeMap<CrateId, PairredExtras>,
         include_build_scripts: bool,
     ) -> Self {
@@ -352,7 +352,7 @@ impl CrateContext {
         };
 
         // Save the repository information for the current crate
-        let repository = source_annotations.get(&package.id).cloned();
+        let repository = source_annotations.get(&package.id).cloned().flatten();
 
         // Identify the license type
         let license = package.license.clone();
--- a/crate_universe/src/metadata/metadata_annotation.rs
+++ b/crate_universe/src/metadata/metadata_annotation.rs
@@ -161,7 +161,7 @@ pub enum SourceAnnotation {
 #[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
 pub struct LockfileAnnotation {
     /// A mapping of crates/packages to additional source (network location) information.
-    pub crates: BTreeMap<PackageId, SourceAnnotation>,
+    pub crates: BTreeMap<PackageId, Option<SourceAnnotation>>,
 }
 
 impl LockfileAnnotation {
@@ -191,7 +191,7 @@ impl LockfileAnnotation {
                     )?,
                 ))
             })
-            .collect::<Result<BTreeMap<PackageId, SourceAnnotation>>>()?;
+            .collect::<Result<BTreeMap<PackageId, Option<SourceAnnotation>>>>()?;
 
         Ok(Self { crates })
     }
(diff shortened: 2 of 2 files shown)


UebelAndre:
> Fixes #1631. Missing source annotations are expected when providing local packages or patching crates that would be vendored.

Is there a way to make this more explicit? In general I think that path packages are a hard thing to support correctly but agree that they can be very useful for this `[patch]` case. Would it be possible to create a concrete representation for path packages and ensure they're paths within the repo itself? If this functionality is also available to `crates_repository` then I think there needs to be a way to properly track the manfests at the given path by the repository rule.

djkoloski:
We could try parsing the [`id` field](https://docs.rs/cargo_metadata/0.15.1/cargo_metadata/struct.Package.html#structfield.id) of each package. That contains information about the path to the crate, eg:

```json
"id": "zerocopy 0.6.1 (path+file:///path/to/zerocopy)"
```

It's worth noting that although the `id` field contains this information about a local path, there is a warning in the documentation that the format of the `id` field may change at any time. The information in `id` doesn't appear in the `source` field, which is always `None` for path dependencies. I don't think there's any other way to reliably tell path dependencies from other dependencies.

UebelAndre:
> there is a warning in the documentation that the format of the `id` field may change at any time.

Do you have a link to this? I care most that the case of something being a path dependency is clear but if it can be explicitly represented (even by some internal container) that'd be great. Can you also provide some kind of end to end test for this change? I'm really concerned about introducing any support for paths since I think it could lead to really confusing issues down the road where people start using relative paths outside the workspace.

djkoloski:
> Do you have a link to this?

The [documentation for the output of `cargo metadata`](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html#output-format) does not specify any particular format for the `id`. This is supported by the [documentation for `PackageId` field](https://docs.rs/cargo_metadata/0.15.1/cargo_metadata/struct.PackageId.html) in the `cargo-metadata` crate. So we could technically parse this to get the information we need but should be prepared to take on a maintenance burden if it changes.

> Can you also provide some kind of end to end test for this change?

Sure, those are all in `tests` right? I just have to read up a bit more so I know how to set it up.

> I'm really concerned about introducing any support for paths since I think it could lead to really confusing issues down the road where people start using relative paths outside the workspace.

The only goal of this change is to fix #1631, so if there's another approach that would work and you like better then I'm more than happy to pursue that instead.

djkoloski:
We also have the manifest path in the `manifest_path` field. Maybe we could check the `source` field and fall back to the manifest path?

andrewpollack:
Are there any updates on this?

herry-c:
I have a similar problem. Is there any other way to use [patch] through bazel?

shelbyd:
I've been able to get past the no source info error by forking the patched dependency and patching with a `git = "..."` specification. While I'd prefer to include the vendored dependencies in the same repo, in a `third_party/` dir, it's workable for me to have a `github.com/<my-company>/<my-project>-vendored` repo.

djkoloski:
Closing this PR, the work for this has been incorporated into #1902.

PR #1646, opened 2022-11-09
Fix vendoring when not in a package

djkoloski:
When vendoring from a label that is not in a package, the output directory would be incorrectly set to `/vendor_path` and be treated as an absolute path.

### DIFF
--- a/crate_universe/private/crates_vendor.bzl
+++ b/crate_universe/private/crates_vendor.bzl
@@ -48,7 +48,7 @@ def _get_output_package(ctx):
             ctx.label.package,
             ctx.attr.vendor_path,
         )
-    return output
+    return output.lstrip("/")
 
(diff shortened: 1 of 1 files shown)


crate_universe/private/crates_vendor.bzl
```diff
@@ -41,7 +41,7 @@ def _is_windows(ctx):
 
 def _get_output_package(ctx):
```
UebelAndre, line 42: I think the original intent here was that users could set `/path/to/something` and get a relative path from the workspace root. Would it not also work call `lstrip("/")` on `output` to continue support for both of these uses?

UebelAndre (changes requested):

UebelAndre (approved):
Thanks!

Issue #1417, opened 2022-06-20
`crate_universe` does not consolidate configurations that match the same set of platforms.


pikajude:
Here's the Cargo.toml for rustix-0.33.7: https://docs.rs/crate/rustix/0.33.7/source/Cargo.toml

The resulting generated file: https://gist.github.com/pikajude/4c6eec44aba219e58f5803a7efecf10b

The condition `# cfg(all(windows, not(target_vendor = "uwp")))` duplicates the condition `# cfg(windows)`. See lines 53 and 97. This results in the error:

```
ERROR: Traceback (most recent call last):
        File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/76ed2d273358d4668b08d66cb9d677b9/external/crate_index__rustix-0.33.7/BUILD.bazel", line 101, column 10, in <toplevel>
                ): [
Error: dictionary expression has duplicate key: ("@rules_rust//rust/platform:i686-pc-windows-msvc", "@rules_rust//rust/platform:x86_64-pc-windows-msvc")
ERROR: /home/ubuntu/.cache/bazel/_bazel_ubuntu/76ed2d273358d4668b08d66cb9d677b9/external/crate_index__wasmtime-runtime-0.35.3/BUILD.bazel:33:13: no such target '@crate_index__rustix-0.33.7//:rustix': target 'rustix' not declared in package '' defined by /home/ubuntu/.cache/bazel/_bazel_ubuntu/76ed2d273358d4668b08d66cb9d677b9/external/crate_index__rustix-0.33.7/BUILD.bazel and referenced by '@crate_index__wasmtime-runtime-0.35.3//:wasmtime_runtime'
```

This is *not* a duplicate of #1236 as we aren't using `supported_platform_triples` in our workspace file.

UebelAndre:
What rev of `rules_rust` are you using?

This seems to be caused by `uwp` not being represented in a unique windows platform but is still a unique configuration. The fix is to either define that as a platform and pass it to `supported_platform_triples` or to update `crate_universe` to flatten configurations that cannot be uniquely represented. I thought we did the latter already... I'll need to double check when I get a chance.

pikajude:
0.6.0. How recently did you make the change you mentioned?

UebelAndre:
[...]

PR #1647, opened 2022-11-10
Fix issue #1417 - Group deps by platform triple rather than cfg(...) string.

wmatthews-google:
Group deps and aliases by platform triple rather than by cfg string when generating BUILD files. This avoid Bazel errors due to duplicate keys/deps. See https://github.com/bazelbuild/rules_rust/issues/1417.

This is done purely at render time by adding a new Tera filter which will remap the conditions in the deps SelectList from "cfg(...)" strings to platform triple strings.

This is my first PR here, so please bear with me if I'm doing things wrong!

### DIFF
--- a/crate_universe/src/rendering/template_engine.rs
+++ b/crate_universe/src/rendering/template_engine.rs
@@ -1,12 +1,14 @@
 //! A template engine backed by [Tera] for rendering Files.
 
-use std::collections::HashMap;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use anyhow::{Context as AnyhowContext, Result};
+use serde::{Deserialize, Serialize};
 use serde_json::{from_value, to_value, Value};
 use tera::{self, Tera};
 
 use crate::config::{CrateId, RenderConfig};
+use crate::context::crate_context::CrateDependency;
 use crate::context::Context;
 use crate::rendering::{
     render_crate_bazel_label, render_crate_bazel_repository, render_crate_build_file,
@@ -14,7 +16,7 @@ use crate::rendering::{
 };
 use crate::utils::sanitize_module_name;
 use crate::utils::sanitize_repository_name;
-use crate::utils::starlark::{SelectStringDict, SelectStringList};
+use crate::utils::starlark::{SelectList, SelectStringDict, SelectStringList};
 
 pub struct TemplateEngine {
     engine: Tera,
@@ -187,6 +189,10 @@ impl TemplateEngine {
         tera.register_function(
             "crates_module_label",
             module_label_fn_generator(render_config.crates_module_template.clone()),
+        );
+        tera.register_filter(
+            "remap_deps_configurations",
+            remap_select_list_configurations::<CrateDependency>,
         );
 
         let mut context = tera::Context::new();
@@ -397,4 +403,21 @@ fn crate_repository_fn_generator(template: String, repository_name: String) -> i
             }
         },
     )
+}
+
+/// Tera filter which re-keys a [`SelectList`]. See [`SelectList::remap_configurations`]
+/// for more details. The mapping must be provided as a Tera argument named `mapping`.
+fn remap_select_list_configurations<T: Ord + Clone + for<'a> Deserialize<'a> + Serialize>(
+    input: &Value,
+    args: &HashMap<String, Value>,
+) -> tera::Result<Value> {
+    let mapping = parse_tera_param!("mapping", BTreeMap<String, BTreeSet<String>>, args);
+
+    match from_value::<SelectList<T>>(input.clone()) {
+        Ok(v) => match to_value(v.remap_configurations(&mapping)) {
+            Ok(v) => Ok(v),
+            Err(_) => Err(tera::Error::msg("Failed to remap conditions.")),
+        },
+        Err(_) => Err(tera::Error::msg("The filter input could not be parsed.")),
+    }
 }
--- a/crate_universe/src/rendering.rs
+++ b/crate_universe/src/rendering.rs
@@ -518,4 +518,51 @@ mod test {
                 .join("\n")
         ));
     }
+
+    #[test]
+    fn test_render_build_file_deps() {
+        let config: Config = serde_json::from_value(serde_json::json!({
+            "generate_build_scripts": false,
+            "rendering": {
+                "repository_name": "multi_cfg_dep",
+                "regen_command": "bazel test //crate_universe:unit_test",
+            },
+            "supported_platform_triples": [
+                "x86_64-apple-darwin",
+                "x86_64-unknown-linux-gnu",
+                "aarch64-apple-darwin",
+                "aarch64-unknown-linux-gnu",
+            ],
+        }))
+        .unwrap();
+        let metadata = test::metadata::multi_cfg_dep();
+        let lockfile = test::lockfile::multi_cfg_dep();
+
+        let annotations = Annotations::new(metadata, lockfile, config.clone()).unwrap();
+        let context = Context::new(annotations).unwrap();
+
+        let renderer = Renderer::new(config.rendering);
+        let output = renderer.render(&context).unwrap();
+
+        let build_file_content = output
+            .get(&PathBuf::from("BUILD.cpufeatures-0.2.1.bazel"))
+            .unwrap();
+
+        // This is unfortunately somewhat brittle. Alas. Ultimately we wish to demonstrate that the
+        // original cfg(...) strings are preserved in the `deps` list for ease of debugging.
+        assert!(
+            build_file_content.replace(' ', "")
+            .contains(&
+r#"deps = [
+    ] + select({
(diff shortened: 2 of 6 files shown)


UebelAndre (changes requested):
Thanks! Could you show an example of what these files render out to? One of the reason cfg strings were used was to try and make it clearer what was happening in the select statements so that if a dependency was unexpectedly missing or present, we can see what Cargo says and how that maps to rust. Just wanna make sure that's still clear.

wmatthews-google:
@UebelAndre - thanks for taking a look! Right now it's not clear at all in the output how deps map to cfg blocks... 

e.g. for `errno` whose crate has
```toml
[target."cfg(target_os=\"dragonfly\")".dependencies.errno-dragonfly]
version = "0.1.1"
[target."cfg(target_os=\"hermit\")".dependencies.libc]
version = "0.2"
[target."cfg(target_os=\"wasi\")".dependencies.libc]
version = "0.2"
[target."cfg(unix)".dependencies.libc]
version = "0.2"
[target."cfg(windows)".dependencies.winapi]
version = "0.3"
```
we now generate
```python
    deps = [
    ] + select({
        "@rules_rust//rust/platform:aarch64-apple-darwin": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:aarch64-apple-ios": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:aarch64-apple-ios-sim": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:aarch64-linux-android": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:aarch64-unknown-linux-gnu": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:arm-unknown-linux-gnueabi": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:armv7-linux-androideabi": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:armv7-unknown-linux-gnueabi": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:i686-apple-darwin": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:i686-linux-android": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:i686-pc-windows-msvc": [
            "@crates_vendor__winapi-0.3.9//:winapi",
        ],
        "@rules_rust//rust/platform:i686-unknown-freebsd": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:i686-unknown-linux-gnu": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:powerpc-unknown-linux-gnu": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:s390x-unknown-linux-gnu": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:wasm32-wasi": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:x86_64-apple-darwin": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:x86_64-apple-ios": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:x86_64-linux-android": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:x86_64-pc-windows-msvc": [
            "@crates_vendor__winapi-0.3.9//:winapi",
        ],
        "@rules_rust//rust/platform:x86_64-unknown-freebsd": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "@rules_rust//rust/platform:x86_64-unknown-linux-gnu": [
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        "//conditions:default": [
        ],
    }),
```
... which makes Bazel happy, but I agree is less clear to humans.

A possible idea:

Suppose I make `remap_conditions` return `SelectList<WithOriginalConditions<T>>` where (roughly)
```rust
struct WithOriginalConditions<T> {
    value: T,
    original_conditions: BTreeSet<Option<String>>
}
```
then render the `original_conditions` (`Some("cfg(...)")` etc., or `None` for default) as comments above each dep.

... then the output would be something like
```python
        # ...
        "@rules_rust//rust/platform:x86_64-unknown-linux-gnu": [
            # cfg(unix)
            "@crates_vendor__libc-0.2.137//:libc",
        ],
        # ...
``` 
... somewhat long and verbose but at least explicit. (And hopefully humans rarely read these generated files anyway).

There is still a down side that conditions which don't match any platform triples end up silently dropped in `remap_conditions`. If this is a concern, `remap_conditions` could return an additional `BTreeSet<String>` of unmapped conditions which could be rendered as comments too.

WDYT?

Related question: Should I regenerate (at the same locked version) all BUILD files under `crate_universe/3rdparty` as part of this PR?

crate_universe/src/rendering/template_engine.rs
```diff
@@ -398,3 +404,18 @@ fn crate_repository_fn_generator(template: String, repository_name: String) -> i
         },
     )
 }
+
+fn remap_select_list_conditions<T: Ord + Clone + for<'a> Deserialize<'a> + Serialize>(
```
UebelAndre, line 408: nit: can you add a comment for this function?

UebelAndre (changes requested):
re: https://github.com/bazelbuild/rules_rust/pull/1647#issuecomment-1312023727

Thanks for showing a rendering, the change is way easier to comprehend now! My concern about readability is more for debugability. Humans likely won't look at it unless there's a problem and in developing `crate_universe` I've learned that dependency management concepts can be difficult to keep paged into my brain. I think it'd be a good thing to make sure the future version of maintainers and new developers have an easier time figuring out what's going on. So I like the proposal to retain info for commenting the original configurations. 👍 

> Related question: Should I regenerate (at the same locked version) all BUILD files under `crate_universe/3rdparty` as part of this PR?

It would be helpful if you had a commit with those refreshed but it does make PRs harder to comprehend post-merge. Your call 😄

crate_universe/src/rendering/template_engine.rs
wmatthews-google: Done.

wmatthews-google:
Updated the code as discussed. The implementation is a big uglier than before (`WithOriginalConfigurations<T>` isn't a great value type inside `SelectList`), however we can now render all the original cfg strings in the resulting build files.

See https://github.com/bazelbuild/rules_rust/pull/1655 for a re-rendering of crate_universe/3rdparty BUILD files after this CL. We can choose to merge it too or not as you deem best.

UebelAndre (changes requested):
I think this is looking good! There's some error which I'm not entirely sure isn't an infrastructure issue so you may want to push a new commit to force CI to retrigger.

One last thought is it might be good to have some kind of test for the renderer that ensures a crate with a conditional dependency renders a build file that contains a comment for the original config and the expected triple somewhere in the output. There aren't a ton of render tests so this might be more scope increase than I thought but I think that'd be a really good test to have. But to be clear, the requested change is just to investigate CI.

wmatthews-google:
Thanks!

Added a build file rendering test. LMK if you think it'd be better to frame the assertion in terms of regexes. As-is it's quite clear how the comments map to dependencies, but this comes at the expense of being a bit brittle.

crate_universe/src/rendering/template_engine.rs
```diff
@@ -398,3 +404,79 @@ fn crate_repository_fn_generator(template: String, repository_name: String) -> i
...
+    use crate::metadata::Annotations;
+    use crate::test::*;
+
+    /// Trims leading and trailing whitespace from each line.
+    fn trim_lines(input: &str) -> String {
+        input.lines().map(str::trim).collect::<Vec<_>>().join("\n")
+    }
+
+    #[test]
+    fn test_render_build_file_deps() {
```
UebelAndre, line 438: Could this test be moved to `rendering.rs`? It doesn't seem like it's specific to Tera (though that's not really a strong distinction since there's only one template engine 😅)

UebelAndre (changes requested):
This looks great! One last request but looks good to me otherwise 😄

crate_universe/src/rendering/template_engine.rs
wmatthews-google: Yup... done. That's definitely nicer.

wmatthews-google:
@UebelAndre thanks for the review!

PR #1650, opened 2022-11-10
Inform users that `crate` and `srcs` are mutually exclusive

UebelAndre:
This should avoid confusion in cases where users have defined tests with both `rust_test.crate` and `rust_test.srcs` set and later find `rust_test.srcs` are actually unused and not being tested.

relates to https://github.com/bazelbuild/rules_rust/issues/2324


UebelAndre:
I guess this turned out to not be the case due to: https://github.com/bazelbuild/rules_rust/blob/0.12.0/test/inline_test_with_deps/test_with_srcs/BUILD.bazel
```
(12:38:24) ERROR: /workdir/test/inline_test_with_deps/test_with_srcs/BUILD.bazel:15:10: in rust_test rule //test/inline_test_with_deps/test_with_srcs:inline_test:
	File "/workdir/rust/private/rust.bzl", line 379, column 13, in _rust_test_impl
		fail("rust_test.crate and rust_test.srcs are mutually exclusive. Update {} to use only one of these attributes".format(
Error in fail: rust_test.crate and rust_test.srcs are mutually exclusive. Update //test/inline_test_with_deps/test_with_srcs:inline_test to use only one of these attributes
(12:38:24) ERROR: /workdir/test/inline_test_with_deps/test_with_srcs/BUILD.bazel:15:10: Analysis of target '//test/inline_test_with_deps/test_with_srcs:inline_test' failed
```

However, if I were to have a test source file exist outside of the traditional package structure, it seems tests go ignored
```diff
diff --git a/test/inline_test_with_deps/test_with_srcs/BUILD.bazel b/test/inline_test_with_deps/test_with_srcs/BUILD.bazel
index 64d628e5..bc8c71f4 100644
--- a/test/inline_test_with_deps/test_with_srcs/BUILD.bazel
+++ b/test/inline_test_with_deps/test_with_srcs/BUILD.bazel
@@ -14,7 +14,10 @@ rust_library(

 rust_test(
     name = "inline_test",
-    srcs = ["src/extra.rs"],
+    srcs = [
+        "src/extra.rs",
+        "tests/test.rs",
+    ],
     crate = ":inline",
     deps = ["//test/inline_test_with_deps/dep"],
 )
diff --git a/test/inline_test_with_deps/test_with_srcs/tests/test.rs b/test/inline_test_with_deps/test_with_srcs/tests/test.rs
new file mode 100644
index 00000000..edbfb4cf
--- /dev/null
+++ b/test/inline_test_with_deps/test_with_srcs/tests/test.rs
@@ -0,0 +1,4 @@
+#[test]
+fn some_test() {
+    assert!(false);
+}
```

I wonder if this functionality should be supported, where one can have `crate` and `srcs`. Maybe as an alternative, a check can be added to ensure `srcs` are exclusively "extra sources" to `crate`, and not integration-like test files.

cc @illicitonion @krasimirgg

krasimirgg:
I'd also expect rust_test.crate and rust_test.srcs to be mutually exclusive and that test case is surprising to me.

hlopko:
I did some archeology and I think we should delete the test. The origin of the logic in questoin is from https://github.com/bazelbuild/rules_rust/pull/203/files#r272400341 and I claim that the feature wasn't really needed (used for test-only submodules with test helpers, but authors seem to be fine with not supporting the feature),  alternatives were not explored (for example why not have a test helpers crate in deps?), and as we can see it is buggy. 

So I vote for deleting the test and submitting this PR (thank you @UebelAndre for working on this!)

UebelAndre:
I should have checked for all tests before but it seems there's one more that fails
```
(22:33:06) ERROR: /workdir/test/out_dir_in_tests/BUILD.bazel:25:16: in rust_test rule //test/out_dir_in_tests:suite_tests/out_dir_reader:
	File "/workdir/rust/private/rust.bzl", line 379, column 13, in _rust_test_impl
		fail("rust_test.crate and rust_test.srcs are mutually exclusive. Update {} to use only one of these attributes".format(
Error in fail: rust_test.crate and rust_test.srcs are mutually exclusive. Update //test/out_dir_in_tests:suite_tests/out_dir_reader to use only one of these attributes
(22:33:06) ERROR: /workdir/test/out_dir_in_tests/BUILD.bazel:25:16: Analysis of target '//test/out_dir_in_tests:suite_tests/out_dir_reader' failed
```
This was introduced in https://github.com/bazelbuild/rules_rust/pull/954. I'll have to dig into this to figure out what the expectation is.

UebelAndre:
@sayrer @briansmith maybe one of you could help me understand https://github.com/bazelbuild/rules_rust/pull/1650#issuecomment-1312527359 ?

illicitonion (approved):

PR #1654, opened 2022-11-10
Update `//util/label` to support `+` in packages

UebelAndre:
This is apparently valid which I confirmed with an `sh_test` in a `baz/foo+bar` package.
```
rules_rust % bazel test //baz/...
INFO: Analyzed target //baz/foo+bar:test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Target //baz/foo+bar:test up-to-date:
  bazel-bin/baz/foo+bar/test
INFO: Elapsed time: 0.216s, Critical Path: 0.09s
INFO: 3 processes: 1 internal, 2 darwin-sandbox.
INFO: Build completed successfully, 3 total actions
//baz/foo+bar:test                                                       PASSED in 0.1s

Executed 1 out of 1 test: 1 test passes.
INFO: Build completed successfully, 3 total actions
```

### DIFF
--- a/util/label/label.rs
+++ b/util/label/label.rs
@@ -153,7 +153,7 @@ fn consume_package_name<'s>(input: &'s str, label: &'s str) -> Result<(&'s str, 
         return Err(LabelError(err(
             label,
             "package names may contain only A-Z, \
-        a-z, 0-9, '/', '-', '.', ' ', '$', '(', ')' and '_'.",
+        a-z, 0-9, '/', '-', '.', ' ', '$', '(', ')', '_', and '+'.",
         )));
     }
     if package_name.ends_with('/') {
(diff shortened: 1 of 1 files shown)


util/label/label.rs
```diff
@@ -148,6 +148,7 @@ fn consume_package_name<'s>(input: &'s str, label: &'s str) -> Result<(&'s str,
             || c == '('
             || c == ')'
             || c == '_'
+            || c == '+'
```
illicitonion, line 151: Can you update the error below too on line 156?

illicitonion (approved):

util/label/label.rs
UebelAndre: I can later today but if you wanted to add apply a suggestion I’d be happy with that too 😅

illicitonion: I totally would have, but Github won't let me suggest on a line you didn't edit :(

UebelAndre: Gah! c'mon github!

-- Updated 😄

PR #1655, opened 2022-11-13
Re-render crate BUILD files after #1647

wmatthews-google:
See https://github.com/bazelbuild/rules_rust/pull/1647 for main PR.


UebelAndre (approved):
Thanks!

PR #1661, opened 2022-11-17
Allow passing a bazel path to vendor explicitly

djkoloski:


docs/crate_universe.md
```diff
@@ -360,8 +360,7 @@ bazel run //3rdparty:crates_vendor
 Repinning dependencies is controlled by both the `CARGO_BAZEL_REPIN` environment variable or the `--repin`
 flag to the `crates_vendor` binary. To update dependencies, simply add the flag ro your `bazel run` invocation.
 
-```shell
-bazel run //3rdparty:crates_vendor -- --repin
+```shellbazel
```
UebelAndre, line 363: Was this intentional?

UebelAndre (changes requested):
Thanks! Only had one question 😄

docs/crate_universe.md
djkoloski: Not intentional, sorry about that. Removed that change.

UebelAndre (approved):

PR #1663, opened 2022-11-18
Revert #1564

illicitonion:
It looks like the reported issue is probably related to the ios
toolchain in use, rather than our general toolchain flag propagation.


illicitonion:
@keith - I would be interested in your thoughts on the CI failure we get from reverting this PR...

The original PR filtered out the `-target` flag from the cc_toolchain when propagating to `cargo_build_script`s. This is causing problems for toolchains which actually need the `-target` flag.

AFAICT what's _actually_ currently needed to fix the issues with building iOS applications is to pass `--cpu=ios_x86_64` on the command line, and that this is effectively a dupe of https://github.com/bazelbuild/bazel/issues/16784 - that Bazel decides what cc_toolchain and flags &c to configure based on `--cpu` rather than `--platforms`...

Is there a nicer way of making the iOS stuff work that _doesn't_ require setting the `--cpu` flag? Or do you have thoughts on what kind of manipulating of cc_toolchain stuff we should be doing to satisfy iOS builds? I'm pretty convinced that just stripping out `-target` is incorrect, but it's a little sad if setting `--cpu` is required as a workaround...

Otherwise, presumably we should start setting `--cpu` on both of these CI builds? https://github.com/bazelbuild/rules_rust/blob/ddf2a4c23c9e08ee6a1cc57dc3058ec7b036224d/.bazelci/presubmit.yml#L446-L459 - the first one doesn't appear to _need_ it because the code involved isn't complex enough (or we don't have sufficiently pedantic asserts), but the second _does_ seem to...

UebelAndre:
Seems a test needs to be updated for iOS?

illicitonion:
> Seems a test needs to be updated for iOS?

Yeah - the test is fixed by adding `--cpu=ios_x86_64` to the iOS example build, which I _think_ is the correct fix for the issue, but I don't know the iOS stuff well enough to know if there's a better one...

dae:
Sorry about the unintended breakage; no objections from me on reverting the change.

UebelAndre:
@dae would you be able to confirm what the right solution is for https://github.com/bazelbuild/rules_rust/pull/1663#issuecomment-1320524776 ?

dae:
If @illicitonion confirmed the test I added passes when `--cpu=ios_x86_64` is provided, that sounds like the best solution, as it addresses the problem at the source, and it's the solution I would have used had I known specifying a platform was not enough.

UebelAndre (approved):

keith:
What part of this change breaks it in that way (or maybe it was just broken at HEAD?)

For `ios_build` I think the issue is not having a `platform_mappings` file, did the issue appear for both `ios` and `ios_build`? (not sure why those are separate either

illicitonion:
Aha - yeah, it looks like we were missing the `platform_mappings` file, and instead were directly setting `--platforms` in the CI for the `ios_build` example - we were only seeing the issue in that one repo. Thanks for looking!

I added a `platform_mappings`, and changed the `.bazelci` file to set the flags mentioned in the `platform_mappings` file instead of the one it sets. This fixes things up :)

PR #1671, opened 2022-11-25
Added a build setting for toolchain channels

UebelAndre:


UebelAndre:
This PR introduces select statements to toolchain definitions:
```starlark
toolchain(
    name = "toolchain",
    exec_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    target_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    toolchain = select({"@rules_rust//rust/toolchain/channel:nightly":"@rust_darwin_x86_64__x86_64-apple-darwin_nightly_tools//:rust_toolchain","@rules_rust//rust/toolchain/channel:stable":"@rust_darwin_x86_64__x86_64-apple-darwin_stable_tools//:rust_toolchain"}),
    toolchain_type = "@rules_rust//rust:toolchain",
)
```

The conditions for these are defined as config settings that match on a build_setting rule I wrote in the `//rust/toolchain/channel` package:
```starlark
config_setting(
    name = "nightly",
    flag_values = {
        ":channel": "nightly",
    },
)

config_setting(
    name = "stable",
    flag_values = {
        ":channel": "stable",
    },
)
```

I'm now confused as to why my test is failing at `//test/unit/channel_transitions`
```
Executing tests from //test/unit/channel_transitions:nightly_transition_test
In test _nightly_transition_test_impl from //test/unit/channel_transitions:channel_transiitons_test.bzl: Expected ["bazel-out/k8-opt-exec-2B5CBBC6-ST-54f9b509cdab/bin/util/process_wrapper/process_wrapper", "--subst", "pwd=${pwd}", "--", "bazel-out/k8-fastbuild-ST-444e700f6c5e/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu_stable_tools/rust_toolchain/bin/rustc", "test/unit/channel_transitions/lib.rs", "--crate-name=lib", "--crate-type=rlib", "--error-format=human", "--codegen=metadata=-2309820253", "--codegen=extra-filename=-2309820253", "--out-dir=bazel-out/k8-fastbuild-ST-444e700f6c5e/bin/test/unit/channel_transitions", "--codegen=opt-level=0", "--codegen=debuginfo=0", "--remap-path-prefix=${pwd}=", "--emit=dep-info,link", "--color=always", "--target=x86_64-unknown-linux-gnu", "-L", "bazel-out/k8-fastbuild-ST-444e700f6c5e/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu_stable_tools/rust_toolchain/lib/rustlib/x86_64-unknown-linux-gnu/lib", "--edition=2018"] to contain -Zunstable-options
```

@illicitonion @scentini @krasimirgg Would any of you have an idea as to why the toolchain is not being transitioned?

UebelAndre:
re: The issue was was resolved with the following diff.

```diff
diff --git a/test/unit/channel_transitions/channel_transiitons_test.bzl b/test/unit/channel_transitions/channel_transiitons_test.bzl
index 748b454a..668f2ca6 100644
--- a/test/unit/channel_transitions/channel_transiitons_test.bzl
+++ b/test/unit/channel_transitions/channel_transiitons_test.bzl
@@ -30,7 +30,7 @@ nightly_transition_test = analysistest.make(
     _nightly_transition_test_impl,
     doc = "Test that targets can be forced to use a nightly toolchain",
     config_settings = {
-        "@rules_rust//rust/toolchain/channel:channel": "nightly",
+        "@//rust/toolchain/channel:channel": "nightly",
     },
 )

@@ -41,7 +41,7 @@ stable_transition_test = analysistest.make(
     _stable_transition_test_impl,
     doc = "Test that targets can be forced to use a stable toolchain",
     config_settings = {
-        "@rules_rust//rust/toolchain/channel:channel": "stable",
+        "@//rust/toolchain/channel:channel": "stable",
     },
 )
```

The behavior around labels has been frustratingly inconsistent. I wish it was clearer what style should be used where or if wrapping a string in `Label` could be "correct" in all cases.

katre:
The label syntax is confusing. When you specify the label as `@rules_rust//something`, that means "a package in the `rules_rust` repo, which might be the main repo if it was named with `workspace(name = "rules_rust")` in WORKSPACE". On the other hand, the `@//something` syntax means "a package in the main repo", which might be what you mean in a test but also might not be what users want when they import your rules.

Can you use repo-relative labels like `//something` instead? Those assume that they refer to the current repo, whether that's the main or an imported repo, and might work better for you.

UebelAndre:
> Can you use repo-relative labels like `//something` instead? Those assume that they refer to the current repo, whether that's the main or an imported repo, and might work better for you.

I tried, repo-relative labels didn't work 😞

katre:
With respect to `select` and `toolchain`, I actually prefer to use the (poorly documented) `target_setting` attribute instead.

You currently have this toolchain:
```
toolchain(
    name = "toolchain",
    exec_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    target_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    toolchain = select({"@rules_rust//rust/toolchain/channel:nightly":"@rust_darwin_x86_64__x86_64-apple-darwin_nightly_tools//:rust_toolchain","@rules_rust//rust/toolchain/channel:stable":"@rust_darwin_x86_64__x86_64-apple-darwin_stable_tools//:rust_toolchain"}),
    toolchain_type = "@rules_rust//rust:toolchain",
)
```

This is equivalent to the following:
```
toolchain(
    name = "toolchain-nightly",
    exec_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    target_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    toolchain = "@rust_darwin_x86_64__x86_64-apple-darwin_nightly_tools//:rust_toolchain",
    target_settings = ["@rules_rust//rust/toolchain/channel:nightly"],
    toolchain_type = "@rules_rust//rust:toolchain",
)

toolchain(
    name = "toolchain-stable",
    exec_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    target_compatible_with = ["@platforms//cpu:x86_64","@platforms//os:osx"],
    toolchain = "@rust_darwin_x86_64__x86_64-apple-darwin_stable_tools//:rust_toolchain",
    target_settings = ["@rules_rust//rust/toolchain/channel:stable"],
    toolchain_type = "@rules_rust//rust:toolchain",
)
```

The benefits of the second form is that it's clearer to users when they read it, it's clearer to rule authors when they want to edit it, and the two different toolchains are listed properly in toolchain debugging.

I need to document `target_settings` better, but the upshot is that it takes a list of `config_setting` targets, and the toolchain will only be used if all of them are true. You can, of course, use the expanded config logic from skylib to make more complicated structures.

UebelAndre:
re https://github.com/bazelbuild/rules_rust/pull/1671#issuecomment-1330779691

Thanks for the suggestion @katre! I was able to cleanup some things to come up with a much cleaner implementation. I had no idea [toolchain.target_settings](https://bazel.build/reference/be/platform#toolchain) existed.

scentini:
TIL about `target_settings` too. @katre, do you expect it to work well for `select`s where `//conditions:default` is involved? In other words, if we have a toolchain with `target_settings` and another one without, will Bazel choose the one with `target_settings` (assuming the underlying `config_setting` is `True`), or will there be ambiguity?

katre:
If you have a toolchain with a target setting that matches the config, and one without, then the normal toolchain mechanism of registration order will apply. Note that if you register toolchains using `//package:all` they are registered in the order they are declared in the BUILD file.

UebelAndre:
Anyone have time to review this this week? 😅

krasimirgg:
> Anyone have time to review this this week? 😅

Thank you for the ping! I will review this tomorrow.

crate_universe/private/common_utils.bzl
```diff
@@ -71,11 +71,18 @@ def get_rust_tools(repository_ctx, host_triple):
         cargo_config = repository_ctx.path(repository_ctx.attr.cargo_config)
         repository_ctx.symlink(cargo_config, cargo_home_config)
 
+    if repository_ctx.attr.rust_version.startswith(("beta", "nightly")):
+        channel, _, version = repository_ctx.attr.rust_version.partition("/")
```
krasimirgg, line 75: I'm confused by this line in conjunction with the documentation for rust_version below:
```
The version of Rust the currently registered toolchain is using. Eg. `1.56.0`, or `nightly/2021-09-08
```
This line makes it look that we're going to have a string where partitioning by `/` has 3 components, I'm not sure what the second one is...

rust/repositories.bzl
```diff
@@ -589,8 +626,10 @@ def rust_repository_set(
 
     Args:
         name (str): The name of the generated repository
-        version (str): The version of the tool among "nightly", "beta', or an exact version.
         exec_triple (str): The Rust-style target that this compiler runs on
+        version (str): The version of the tool among "nightly", "beta', or an exact version.
```
krasimirgg, line 630: nit: replace `"beta'` with `"beta"`; also add an example for how an exact version looks like, like the `1.65.0` above.

rust/repositories.bzl
```diff
@@ -589,8 +626,10 @@ def rust_repository_set(
...
+        versions (list, optional): A list of toolchain versions to download. This paramter only accepts one versions
```
krasimirgg, line 631: 1. do we fail with an error message if the user supplied multiple versions for a channel?
2. nit: ... only accepts one version [not versions] ...

krasimirgg (commented):
Looks good! I've got a few nits.

rust/repositories.bzl
UebelAndre: What is the API for defining a stable and nightly toolchain then?

crate_universe/private/common_utils.bzl
UebelAndre: Where is the 3rd component coming from? [string.partition](https://bazel.build/rules/lib/string#partition) is documented as
> Splits the input string at the first occurrence of the separator sep and returns the resulting partition as a three-element tuple of the form (before, separator, after). If the input string does not contain the separator, partition returns (self, '', ''). 

So of the 3 values being returned, only 2 are expected to contain meaningful information, right?

rust/repositories.bzl
krasimirgg: The API as-is makes sense. It was just a question what happens if the user tries to supply multiple versions for the same channel regardless, like what if we pass `["nightly/2022-11-02", "nightly/2022-11-03"]`, do we error out in this case?

krasimirgg: Perfect! Thank you!

krasimirgg (approved):
Thank you!

PR #1665, opened 2022-11-22
fix: add space to crate data exclude list

bazaglia:
Add files with spaces to the exclude list. [Original discussion raised on Slack](https://bazelbuild.slack.com/archives/CSV56UT0F/p1669140022100749?thread_ts=1669139179.947319&cid=CSV56UT0F): 

> Hey! I'm trying to setup my Rust project for the first time following the example from rules_rust/examples/crate_universe but I get the following error. My rust binary compiles and run just fine when running outside Bazel. It seems to be related with one of the deps having a space in its name? How can I address that?
> ```
> INFO: Analyzed target //backend/router:router (1 packages loaded, 5 targets configured).
> INFO: Found 1 target...
> ERROR: /private/var/tmp/_bazel_andre.bazaglia/d99a96be6823e596c135c9df87021555/external/crate_index__v8-0.44.3/BUILD.bazel:98:19: Creating runfiles tree bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crate_index__v8-0.44.3/v8_build_script_.runfiles [for tool] failed: build-runfiles failed: error executing command
>   (cd /private/var/tmp/_bazel_andre.bazaglia/d99a96be6823e596c135c9df87021555/execroot/energy_services && \
>   exec env - \
>     PATH=/bin:/usr/bin:/usr/local/bin \
>   /var/tmp/_bazel_andre.bazaglia/install/796da0a89b8d87ba0127fbf18aa5085d/build-runfiles --allow_relative bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crate_index__v8-0.44.3/v8_build_script_.runfiles_manifest bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crate_index__v8-0.44.3/v8_build_script_.runfiles): Process exited with status 1: Process exited with status 1
[...]

PR #1673, opened 2022-11-26
Common glob excludes

bazaglia:
PR #1665, unfortunately, missed updating the exclude attribute from the glob helper function everywhere it was used. This PR fixes that and also makes it now consistent everywhere it is needed by creating a centralized list.

I created a repro in [bazaglia/rules_rust_apollo_router](https://github.com/bazaglia/rules_rust_apollo_router) and verify a build that was previously failing because of the inconsistency in the exclude globs works with the changes in that PR. Thanks for sharing the diff @UebelAndre. 😃

### DIFF
--- a/crate_universe/src/rendering/template_engine.rs
+++ b/crate_universe/src/rendering/template_engine.rs
@@ -23,6 +23,14 @@ pub struct TemplateEngine {
     context: tera::Context,
 }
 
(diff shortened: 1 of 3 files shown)


UebelAndre (approved):
Hehe thanks for opening the PR!

PR #1675, opened 2022-11-29
Merge cc toolchain flags into build script env

illicitonion:
Right now, setting build_script_env will _overwrite_ the values from the cc_toolchain, which is almost certainly ~always wrong.

Instead, allow _appending_ values, which means that if overrides are necessary people can add them (e.g. disabling specific warnings).

Note: This is technically a breaking change, but one which should improve faithfulness of the build.


UebelAndre (approved):
Seems reasonable to me! Is there an easy way to have a test for this in `//test/cargo_build_script` though? Would be nice to decouple this example/integration test with testing the new functionality.

PR #1677, opened 2022-11-29
Fix confusing/misleading crate_universe docs

### LINKED ISSUE #1674: Hard to follow crate_universe documentation
I'm trying to follow this example in https://bazelbuild.github.io/rules_rust/crate_universe.html#direct-packages:

```bzl
load("@cargo_bazel//:defs.bzl", "crate", "crates_repository", "render_config")

crates_repository(
    name = "crate_index",
    lockfile = "//:Cargo.Bazel.lock",
    packages = {
        "async-trait": crate.spec(
            version = "0.1.51",
        ),
        "mockall": crate.spec(
            version = "0.10.2",
        ),
        "tokio": crate.spec(
            version = "1.12.0",
        ),
    },
    # Setting the default package name to `""` forces the use of the macros defined in this repository
    # to always use the root package when looking for dependencies or aliases. This should be considered
    # optional as the repository also exposes alises for easy access to all dependencies.
        default_package_name = ""
    ),
)

crate_repositories()
```

This produces the following error:

```
ERROR: Failed to load Starlark extension '@cargo_bazel//:defs.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @cargo_bazel
This could either mean you have to add the '@cargo_bazel' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
```

If I replace `@cargo_bazel//:defs.bzl` with `@rules_rust//crate_universe:defs.bzl`, then:

```
ERROR: /Users/john/figma/rules_rust_built/WORKSPACE:21:18: //external:crate_index: missing value for mandatory attribute 'cargo_lockfile' in 'crates_repository' rule
ERROR: error loading package '': Encountered error while reading extension file 'defs.bzl': no such package '@crate_index//': error loading package 'external': Could not load //external package
```

If I replace `lockfile` with `cargo_lockfile`, then:

```
ERROR: An error occurred during the fetch of repository 'crate_index':
	File "/private/var/tmp/_bazel_john/6ebb6fa28311e23a56b6491e5d493861/external/rules_rust/crate_universe/private/crates_repository.bzl", line 34, column 30, in _crates_repository_impl
		lockfiles = get_lockfiles(repository_ctx)
	File "/private/var/tmp/_bazel_john/6ebb6fa28311e23a56b6491e5d493861/external/rules_rust/crate_universe/private/generate_utils.bzl", line 299, column 36, in get_lockfiles
		cargo = repository_ctx.path(repository_ctx.attr.cargo_lockfile),
Error in path: Not a regular file: /Users/john/figma/rules_rust_built/Cargo.Bazel.lock
```

Okay, I read further down in the documentation, it says "If set, this file must exist within the workspace (but can be empty) before this rule will work." That's referring to the `lockfile` attribute, not `cargo_lockfile`. I can maybe guess that it also applies to `cargo_lockfile`, but it would be nice if the error message said something like "Create an empty Cargo.Bazel.lock file in order to proceed" right away, instead of "Error in path: Not a regular file". Or, why can't it just create the file for me?

illicitonion:
Fixes #1674

### DIFF
--- a/crate_universe/docs.bzl
+++ b/crate_universe/docs.bzl
@@ -46,6 +46,7 @@ load("@rules_rust//crate_universe:defs.bzl", "crates_repository")
 
 crates_repository(
     name = "crate_index",
+    cargo_lockfile = "//:Cargo.lock",
     lockfile = "//:Cargo.Bazel.lock",
     manifests = ["//:Cargo.toml"],
 )
@@ -99,10 +100,11 @@ about missing targets or environment variables defined only in Bazel. In workspa
 to have a "Cargo free" setup. `crates_repository` supports this through the `packages` attribute.
 
 ```python
-load("@cargo_bazel//:defs.bzl", "crate", "crates_repository", "render_config")
+load("@rules_rust//crate_universe:defs.bzl", "crate", "crates_repository", "render_config")
 
 crates_repository(
     name = "crate_index",
+    cargo_lockfile = "//:Cargo.lock",
     lockfile = "//:Cargo.Bazel.lock",
     packages = {
         "async-trait": crate.spec(
--- a/crate_universe/private/crates_repository.bzl
+++ b/crate_universe/private/crates_repository.bzl
@@ -155,7 +155,7 @@ crates_repository(
 The above will create an external repository which contains aliases and macros for accessing
 Rust targets found in the dependency graph defined by the given manifests.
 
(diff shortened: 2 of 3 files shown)


UebelAndre (approved):
Thanks!

PR #1678, opened 2022-11-29
crate_universe/private/crates_vendor.bzl typo fix

andrewpollack:
Fixing typo in `crate_universe/private/crates_vendor.bzl`


UebelAndre (approved):
Good catch, Thanks!

PR #1682, opened 2022-11-30
stardoc: Use backtick not `<code>` for attr default values 

illicitonion:
This is a cherry-pick of https://github.com/bazelbuild/stardoc/pull/143


UebelAndre (approved):
Thanks!

PR #1683, opened 2022-11-30
Add RV64GC target

ghost:
RV64GC can use essentially the same set of parameters as RV32IMC, but must be declared in rules_rust for downstream targets to use Rust inside Bazel for RV64GC targets.

Tested against a demonstration project targeting QEMU's RV64 Virt platform.

### DIFF
--- a/examples/crate_universe/cargo_aliases/Cargo.Bazel.lock
+++ b/examples/crate_universe/cargo_aliases/Cargo.Bazel.lock
@@ -326,9 +326,9 @@ checksum = "73473c0e59e6d5812c5dfe2a064a6444949f089e20eec9a2e5506596494e4623"
 
 [[package]]
 name = "syn"
-version = "1.0.104"
+version = "1.0.105"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ae548ec36cf198c0ef7710d3c230987c2d6d7bd98ad6edc0274462724c585ce"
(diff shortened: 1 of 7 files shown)


rust/platform/triple_mappings.bzl
```diff
@@ -61,6 +62,8 @@ _CPU_ARCH_TO_BUILTIN_PLAT_SUFFIX = {
     "powerpc64le": None,
     "riscv32": "riscv32",
     "riscv32imc": "riscv32",
+    "riscv64": "riscv64",
+    "riscv64gc": "riscv64",
```
UebelAndre, line 66: I don't see a constraint for this defined at https://github.com/bazelbuild/platforms/blob/0.0.6/cpu/BUILD. Won't this cause problems if anyone tries to match it?

UebelAndre (changes requested):

rust/platform/triple_mappings.bzl
UebelAndre: oh nevermind, they get mapped to `riscv54`. No issue here.

UebelAndre (approved):
Thanks!

UebelAndre:
A rebase is needed and then the changes can be merged 😄

tdewey-rivos:
Had to re-pin again, syn jumped from .104 to 1.05. Thanks for the review!

PR #1684, opened 2022-11-30
Remove ios/android/wasm support for gen_rust_project deps

P1n3appl3:
Currently, the `crates_vendor` for the `gen_rust_project` binary uses the default list of supported platform triples. This results in the `@build_bazel_apple_support` repository getting unconditionally downloaded when you `bazel run @rules_rust//tools/rust_analyzer:gen_rust_project`, which is undesirable when you're not writing rust targeting ios.

The apple support repository is downloaded because it's referenced whenever `triple_to_constraint_set` is called (because it calls `extra_ios_constraints`). I'm not sure how exactly that's getting invoked through `crates_vendor`, but the easiest solution is to remove the ios entries from the `supported_platform_triples` list. I also took the liberty of removing wasm and android given that it's unlikely that anyone is using bazel+rust-analyzer on those platforms.


UebelAndre:
For this change I think you're gonna need to regenerate dependencies by running the target you just updated.

P1n3appl3:
I've just updated the deps gain. @UebelAndre let me know if there are any other changes you wanted, I think this is ready to merge otherwise.

UebelAndre (approved):
Super sorry about missing this for so long! Looks good, thanks!

PR #1692, opened 2022-12-06
Add `env` attribute to `rust_binary`

UebelAndre:
This change allows users to set an `env` attribute on `rust_binary` targets to then have environment variables defined for `bazel run` invocations via the use of [RunEnvironmentInfo](https://bazel.build/rules/lib/RunEnvironmentInfo).


UebelAndre:
[RunEnvironmentInfo](https://bazel.build/rules/lib/RunEnvironmentInfo) is currently only in Bazel 5.3 and I'm not sure how to conditionally add this provider 😞

rust/private/rust.bzl
```diff
@@ -355,6 +355,16 @@ def _rust_binary_impl(ctx):
         ),
     )
 
+    providers.append(RunEnvironmentInfo(
```
illicitonion, line 358: You can probably use https://github.com/bazelbuild/bazel-skylib/blob/5bfcb1a684550626ce138fe0fe8f5f702b3764c3/lib/versions.bzl#L127 to check `versions.is_at_least("5.3.0")`

UebelAndre: That won't work outside of repository rules until something like https://github.com/bazelbuild/bazel/issues/8305 is implemented and I don't think Bazel rules should need to do gymnastics to get the version of Bazel being used into rules 😞

UebelAndre: With the min version bumped to 6.0 I think this should be good to go now.

illicitonion (approved):
Ship it!

PR #1695, opened 2022-12-09
load_arbitrary_tool uses tool_suburl to look up sha256

ted-logan:
In load_arbitrary_tool, use the suburl (which contains the a subdirectory containing the date, if the version is beta or nightly) to look up the sha256 to to download a specific archive. This allows Bazel to cache the downloaded archive by hash, and supports the correct behavior for dated nightly builds, as provided in the FILE_KEY_TO_SHA dict in //rust:known_shas.bzl.

The most recent nightly entry is given in FILE_KEY_TO_SHA as:

```
    "2022-11-02/cargo-nightly-x86_64-unknown-linux-gnu.tar.gz": "d32e0a9f78ece567627b9b572912b000c53099c0dd9c9f5cea54848de02c6486",
```

This change fixes downloaded archive caching when using the nightly toolchains, including the default nightly toolchain.

### DIFF
--- a/rust/private/repository_utils.bzl
+++ b/rust/private/repository_utils.bzl
@@ -671,7 +671,8 @@ def load_arbitrary_tool(ctx, tool_name, tool_subdirectories, version, iso_date, 
             urls.append(new_url)
 
     tool_path = produce_tool_path(tool_name, target_triple, version)
-    archive_path = tool_path + _get_tool_extension(ctx)
+
+    archive_path = tool_suburl + _get_tool_extension(ctx)
     sha256 = getattr(ctx.attr, "sha256s", dict()).get(archive_path) or FILE_KEY_TO_SHA.get(archive_path) or sha256
 
     for subdirectory in tool_subdirectories:
--- a/rust/private/repository_utils.bzl
+++ b/rust/private/repository_utils.bzl
@@ -630,6 +630,30 @@ def produce_tool_path(tool_name, target_triple, version):
         fail("No tool version was provided")
     return "-".join([e for e in [tool_name, version, target_triple] if e])
 
+def lookup_tool_sha256(ctx, tool_name, target_triple, version, iso_date, sha256):
+    """Looks up the sha256 hash of a specific tool archive.
+
+    The lookup order is:
+
+    1. The sha256s dict in the context attributes;
+    2. The list of sha256 hashes populated in //rust:known_shas.bzl;
+    3. The sha256 argument to the function
+
+    Args:
+        ctx (repository_ctx): A repository_ctx (no attrs required).
+        tool_name (str): The name of the given tool per the archive naming.
+        target_triple (str): The rust-style target triple of the tool
+        version (str): The version of the tool among "nightly", "beta', or an exact version.
+        iso_date (str): The date of the tool (ignored if the version is a specific version).
+        sha256 (str): The expected hash of hash of the Rust tool.
+
+    Returns:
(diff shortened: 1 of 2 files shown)


rust/private/repository_utils.bzl
```diff
@@ -671,7 +671,8 @@ def load_arbitrary_tool(ctx, tool_name, tool_subdirectories, version, iso_date,
             urls.append(new_url)
 
     tool_path = produce_tool_path(tool_name, target_triple, version)
-    archive_path = tool_path + _get_tool_extension(ctx)
+
```
UebelAndre, line 674: Is the new line needed?

UebelAndre (changes requested):
Thanks for opening this PR! Could you show a before and after of what values get passed to `ctx.download_and_extract`?

Also, would you be willing to add a test for this? My thoughts are the url and sha256 value calculations in `load_arbitrary_tool` should be moved into a helper function so it can be called in a starlark unit test without downloading things. Does that make sense?

ted-logan:
> Could you show a before and after of what values get passed to `ctx.download_and_extract`?

Adding a debug print immediately before calling `ctx.download_and_extract()` in the for loop, and running my build, shows the sha256 is empty for the nightly release and populated with my change.  (I'm running with the `--@rules_rust//rust/toolchain/channel=nightly`, because my application requires unstable features only available in the nightly.)

Without my change applied, for the rustc download the sha256 is empty because it's trying to look up `rustc-nightly-aarch64-apple-darwin.tar.gz` in `FILE_KEY_TO_SHA`, which doesn't exist because the nightly archives in `FILE_KEY_TO_SHA` are prefixed by date.
```
DEBUG: load_arbitrary_tool() calling ctx.download_and_extract(urls = ["https://static.rust-lang.org/dist/2022-11-02/rustc-nightly-aarch64-apple-darwin.tar.gz"], sha256 = , auth = {}, stripPrefix = rustc-nightly-aarch64-apple-darwin/rustc)
```

With my change here applied, at the same point I see a valid hash for the same file:

```
DEBUG: load_arbitrary_tool() calling ctx.download_and_extract(urls = ["https://static.rust-lang.org/dist/2022-11-02/rustc-nightly-aarch64-apple-darwin.tar.gz"], sha256 = c905e93a13634a705c29def0d61cf6b877f6637e244c82afab9ab321dd23fc28, auth = {}, stripPrefix = rustc-nightly-aarch64-apple-darwin/rustc)
```

The current stable release is unchanged; both runs give the same hash:

```
DEBUG: load_arbitrary_tool() calling ctx.download_and_extract(urls = ["https://static.rust-lang.org/dist/rustc-1.65.0-aarch64-apple-darwin.tar.gz"], sha256 = bbcf34977e41b9f966746a559aa2af6fa7efd7f695338851b37f722f7a1104fb, auth = {}, stripPrefix = rustc-1.65.0-aarch64-apple-darwin/rustc)
```

> Also, would you be willing to add a test for this? My thoughts are the url and sha256 value calculations in `load_arbitrary_tool` should be moved into a helper function so it can be called in a starlark unit test without downloading things. Does that make sense?

I've moved the sha256 code out of `load_arbitrary_tool` and added a new test to confirm the stable and nightly lookups are working correctly.

rust/private/repository_utils.bzl
ted-logan: My thought was that the archive path is no longer calculated using the tool path created in the previous line, so it sort of made sense to separate it visually.

I've moved this code out into a new function though.

UebelAndre (approved):
Awesome work! Thanks!

PR #1697, opened 2022-12-13
rust_analyzer: fixed aliased crates from accidentally merging in generate_rust_project()

goniz:
The code in `tools/rust_analyzer/rust_project.rs:generate_rust_project()` goes through the list of crates produced by Bazel and tries to uniquely merge all the crates by their `crate_id`.

I've encountered that when using `aliases` in `rust_library`/`rust_binary` then aliased crates will get merged and will not appear in the final `rust-project.json` file.

This change tries to solve this by inserting the `crate.name` into `crate_id`.


goniz:
@scentini @hlopko 🙏🏻 ❤️

scentini:
Hey @goniz, could you please add a test covering the problematic behavior?

PR #1703, opened 2022-12-16
Attach debug info outputs to rust tests

chancila:
Unclear why #1072 explicitly didn't allow tests to generate dsym files via the `dsym_folder` output group.

This change allows them to do so and make debugging easier on macos.
 
Manual verification on macos:

```
bazel build -c dbg //test/unit/debug_info:myrusttest
rust-lldb bazel-bin/test/unit/debug_info/test-2282882243/myrusttest \
    --one-line "image list" \
    --one-line "exit" | \
    grep DWARF &>/dev/null && echo "dSYM module loaded"
```

should print `dSYM module loaded`


illicitonion:
LGTM, thanks!

@ddeville I think you added this originally - was there an important reason to skip debug generation for tests?

illicitonion (approved):

PR #1712, opened 2022-12-20
Added CI for single toolchain channel workspaces

UebelAndre:
This provides regression testing for cases where users have only registered one toolchain in their workspace (meaning they only have a toolchain of a single [channel](https://rust-lang.github.io/rustup/concepts/channels.html) type).

### DIFF
--- a/test/cargo_build_script/tools_exec.rs
+++ b/test/cargo_build_script/tools_exec.rs
@@ -3,7 +3,6 @@ pub fn test_tool_exec() {
     let tool_path = env!("TOOL_PATH");
     assert!(
         tool_path.contains("-exec-"),
-        "tool_path did not contain '-exec-': {}",
-        tool_path
+        "tool_path did not contain '-exec-'",
     );
 }
--- a/test/process_wrapper/rustc_quit_on_rmeta.rs
+++ b/test/process_wrapper/rustc_quit_on_rmeta.rs
@@ -68,8 +68,7 @@ mod test {
         ]);
         assert!(
             !out_content.contains("should not be in output"),
-            "output should not contain 'should not be in output' but did: {}",
-            out_content
+            "output should not contain 'should not be in output' but did",
         );
     }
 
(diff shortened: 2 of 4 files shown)


test/process_wrapper/rustc_quit_on_rmeta.rs
```diff
@@ -68,8 +68,7 @@ mod test {
         ]);
         assert!(
             !out_content.contains("should not be in output"),
-            "output should not contain 'should not be in output' but did: {}",
-            out_content
+            "output should not contain 'should not be in output' but did",
```
UebelAndre, line 71: I believe this change is related to https://github.com/rust-lang/rust-clippy/pull/10055

Unfortunately `#[allow(clippy::uninlined_format_args]` cannot be used here without requiring we raise our min supported Rust version above 1.59.0

test/cargo_build_script/tools_exec.rs
```diff
@@ -1,9 +1,9 @@
+
 #[test]
 pub fn test_tool_exec() {
     let tool_path = env!("TOOL_PATH");
     assert!(
         tool_path.contains("-exec-"),
-        "tool_path did not contain '-exec-': {}",
-        tool_path
+        "tool_path did not contain '-exec-'",
```
UebelAndre, line 7: I believe this change is related to https://github.com/rust-lang/rust-clippy/pull/10055

Unfortunately `#[allow(clippy::uninlined_format_args]` cannot be used here without requiring we raise our min supported Rust version above 1.59.0

test/process_wrapper/rustc_quit_on_rmeta.rs
illicitonion: Can `rustversion::attr` be used here? https://crates.io/crates/rustversion

(I don't know the answer - I know `rustversion` can be used for conditional attrs on structs, and can't be used for conditional enabling of features - I'm not sure how `allow`s fit in :))

UebelAndre: I don't think it'd be appropriate to add a new external dependency just for this. It seems it'll be fixed in a future nightly release so I think it's an acceptable trade off for now. If I can remember it I'll add them back in the future

illicitonion (approved):

PR #1716, opened 2022-12-21
Match prerelease versions with annotation wildcard

### LINKED ISSUE #1690: crate.annotation version wildcard does not match prerelease versions
The `"*"` default value for the `version` attribute appears not to match prerelease versions of crates. For example, I was trying to add an annotation for launchdarkly-server-sdk 1.0.0-beta.4:

```
        "launchdarkly-server-sdk": [crate.annotation(
            patch_args = ["-p1"],
            patches = ["@//bazel/patches:launchdarkly-rust-server-sdk/0001-remove-use-of-built-crate-to-allow-bazel-to-build.patch"],
        )],
```

This produces:

> Error: Unused annotations were provided. Please remove them: [CrateId { name: "launchdarkly-server-sdk", version: "*" }]

If I add `version = "1.0.0-beta.4"`, then it works.

This is like the crates_universe version of https://github.com/google/cargo-raze/issues/419. Do they use the same underlying library for version matching or something?

jfirebaugh:
Fixes #1690

### DIFF
--- a/crate_universe/private/crate.bzl
+++ b/crate_universe/private/crate.bzl
@@ -90,7 +90,8 @@ def _annotation(
     """A collection of extra attributes and settings for a particular crate
 
     Args:
-        version (str, optional): The version or semver-conditions to match with a crate.
+        version (str, optional): The version or semver-conditions to match with a crate. The wildcard `*`
+            matches any version, including prerelease versions.
         additive_build_file_content (str, optional): Extra contents to write to the bottom of generated BUILD files.
         additive_build_file (str, optional): A file containing extra contents to write to the bottom of
             generated BUILD files.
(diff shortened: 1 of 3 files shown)


illicitonion (commented):
LGTM, thanks!

Could you also update the docstring on https://github.com/bazelbuild/rules_rust/blob/2f0511782f3706bc8a64cc8e051864e4daeac2bb/crate_universe/private/crate.bzl#L93 to explicitly say that `*` matches _any_ version, including pre-release versions? (After that you'll need to run `docs/update_docs.sh` to regenerate the markdown docs).

Thanks!

illicitonion (approved):
Thanks!

PR #1717, opened 2022-12-21
Addressed clippy warnings from `clippy 0.1.67 (ec56537c 2022-12-15)`

UebelAndre:
This PR addresses newer clippy warnings that exist in nightly toolchains. https://github.com/bazelbuild/rules_rust/pull/1712 introduces regression testing for this change but since it's larger change I thought it made sense to review separately.

### DIFF
--- a/cargo/bootstrap/bootstrap_installer.rs
+++ b/cargo/bootstrap/bootstrap_installer.rs
@@ -29,7 +29,7 @@ fn install() -> std::io::Result<u64> {
 
 fn main() {
     if let Err(err) = install() {
-        eprintln!("{:?}", err);
+        eprintln!("{err:?}");
         std::process::exit(1);
     };
 }
--- a/cargo/cargo_build_script_runner/bin.rs
+++ b/cargo/cargo_build_script_runner/bin.rs
@@ -127,7 +127,7 @@ fn run_buildrs() -> Result<(), String> {
         format!(
             "Build script process failed{}\n--stdout:\n{}\n--stderr:\n{}",
             if let Some(exit_code) = process_output.status.code() {
-                format!(" with exit code {}", exit_code)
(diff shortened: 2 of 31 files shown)


illicitonion (approved):
I think this is generally reasonable, _but_, it _does_ restrict `rules_rust` support to only rust versions 1.58.0 (released January 2022) and newer. Specifically because the process wrapper will now be using a construct only supported since that release.

If we care about supporting older rusts, we should probably disable this particular lint in the process_wrapper and revert the changes in that one binary.

I personally am fine with dropping support for year-old rusts, but also if someone else _does_ care, I don't think slightly-cleaner-format-strings is a valuable enough feature to justify breaking support for older rusts.

UebelAndre:
> I think this is generally reasonable, _but_, it _does_ restrict `rules_rust` support to only rust versions 1.58.0 (released January 2022) and newer. Specifically because the process wrapper will now be using a construct only supported since that release.
> 
> If we care about supporting older rusts, we should probably disable this particular lint in the process_wrapper and revert the changes in that one binary.
> 
> I personally am fine with dropping support for year-old rusts, but also if someone else _does_ care, I don't think slightly-cleaner-format-strings is a valuable enough feature to justify breaking support for older rusts.

I would agree some formatting fixes would be an upsetting thing to change the min supported version for. However, before making that change I wanted to know what the min version there repo supported was so I opened https://github.com/bazelbuild/rules_rust/pull/1720 and found issues on versions less than `1.59.0`. I only stopped at one and haven't looked to see if there's a work-around. I posted the error on the following comment and think we should discuss that there: https://github.com/bazelbuild/rules_rust/pull/1720#issuecomment-1361728936

illicitonion:
Works for me - ship it!

PR #1731, opened 2022-12-28
Introduce rules_rust_bazel_integration_test; port clippy tests

### LINKED ISSUE #1730: clippy_failure_test is silently failing
> Testing bad_test_clippy... stat: cannot stat '/tmp/ci-vr9W8fqJ3x/rules_rust_test_clippy/bazel-bin/test/clippy/bad_test.clippy.out': No such file or directory

jfirebaugh:
Introduces `rules_rust_bazel_integration_test`, a macro on top of [rules_bazel_integration_test](https://github.com/bazel-contrib/rules_bazel_integration_test) that allows writing integration tests for rules_rust in Rust.

The approach is similar to the one described in ["Bazel Rules to Test Bazel Rules"](https://www.tweag.io/blog/2022-10-06-bazel-rules-to-test-bazel-rules/): An integration test lives in a subdirectory of `test/integration`. Within this directory, create a `workspace` directory containing a self-contained Bazel workspace that sets up the scenario under test. Write a `test.rs` file containing standard Rust tests, which may use the utility functions in `integration_test.rs` to run Bazel commands in the provided workspace and examine their results.

As an example of what integration tests in this style look like, I ported the existing "clippy" tests that were previously run via `clippy_failure_test.sh`. Error handling in shell scripts is notoriously difficult to get right, as evidenced by #1730. I believe that being able to write tests in Rust will have much better ergonomics and reliability.

Fixes #1730.


jfirebaugh:
Tests are failing in CI because the [expected environment variable `BIT_BAZEL_BINARY`](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/9df8426df7edafdff6f8d4588d8fedbe23ba0e93/doc/custom_test_runners.md#implementation-of-a-custom-test-runner) is not set within the test runner. I'm not sure what's causing this, as these tests pass locally. @cgrindel do you have any idea?

cgrindel:
@jfirebaugh Running `bazel test //...` from your branch on my laptop is failing for me. I noticed that you opted to pass the output of a `rules_test` as the test runner. Is the output of that rule a simple binary? I know some rulesets can generate different test outputs depending on the platform. Have you tried just passing in the output of a `rust_binary`?

illicitonion:
I had a quick look, I think there's an issue of tags here?

You don't actually want to run `//test/integration/clippy:test_test` - it's an implementation detail test which doesn't set env vars itself but expects a wrapper to. So we should add a `manual` tag to it.

But by default you _do_ want to run `//test/integration/clippy:test`, but the integration testing framework sets the `manual` tag by default.

This patch may help:

```patch
diff --git a/test/integration/integration_test.bzl b/test/integration/integration_test.bzl
index 7018c111..841d9504 100644
--- a/test/integration/integration_test.bzl
+++ b/test/integration/integration_test.bzl
@@ -17,6 +17,7 @@ def rules_rust_bazel_integration_test(name, srcs, deps = []):
         srcs = srcs,
         deps = deps + ["//test/integration"],
         edition = "2018",
+        tags = ["manual"],
     )
     bazel_integration_test(
         name = name,
@@ -26,4 +27,5 @@ def rules_rust_bazel_integration_test(name, srcs, deps = []):
             "//:all_files",
         ],
         workspace_path = "workspace",
+        tags = ["exclusive"],
     )
```

cgrindel:
Interesting. Unless overridden, `bazel_integration_test` should [set](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/473e3ed9a3bf1d49658253601a8c90c9573c1597/bazel_integration_test/private/bazel_integration_test.bzl#L46) the [tags](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/main/bazel_integration_test/private/integration_test_utils.bzl#L126-L129) `manual` and `exclusive`. Perhaps only the `rust_test` needs the tags?

illicitonion:
> Interesting. Unless overridden, `bazel_integration_test` should [set](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/473e3ed9a3bf1d49658253601a8c90c9573c1597/bazel_integration_test/private/bazel_integration_test.bzl#L46) the [tags](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/main/bazel_integration_test/private/integration_test_utils.bzl#L126-L129) `manual` and `exclusive`. Perhaps only the `rust_test` needs the tags?

Do you know _why_ `bazel_integration_test` sets `manual` by default? That will serve to prevent the tests running when running `//...` as we do on CI...

cgrindel:
In all of the situations where I have used it, I did not want the big integration tests to run with `bazel test //...`. Given the nature of the tests, almost any change to the codebase will cause these tests to run. Since they can take a long time and take up quite a bit of machine resources, I decided to default them to explicit execution. The typical pattern is to define two test suites: one for smoke tests (e.g., run the integration tests against the current Bazel version) and one for all integration tests (e.g., run for all of the supported Bazel versions). I run `bazel test //...` constantly during development. Once I am happy  with my changes and `bazel test //...` is green, I run the smoke tests. Typically, I leave the execution of the full integration test suite to the CI machines.

PR #1275, opened 2022-04-20
rules_rust: enable pipelined compilation.

gigaroby:
Pipelined compilation allows better parallelism during builds as it
allows libraries to generate lightweight metadata files to unlock other
depencies. These metadata files (.rmeta) can only be used to unlock
library -> library dependencies and do not affect builds in any other
way. This is currently the default in cargo:
https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199

PR #1733, opened 2022-12-29
[crate_universe] add an annotation to disable pipelining

Calsign:
The recent support for pipelined compilation (#1275) is great. There are some situations, however, where we need to disable pipelining for specific crates, e.g. #1584. This PR adds a crate annotation option to disable pipelining for rust_library targets generated by cargo_bazel. The alternatives would be patching the crate with the annotations system or disabling pipelining globally.


illicitonion:
LGTM, thanks! Could you add a test to https://github.com/bazelbuild/rules_rust/blob/main/crate_universe/src/rendering.rs which shows this gets properly passed through the system? Then I'm happy to merge!

Calsign:
@illicitonion I have added such a test, I think it should be good to merge now?

illicitonion (approved):
Thanks!

PR #1735, opened 2022-12-30
Re-pinned all dependencies managed by crate_universe

UebelAndre:
This was generated by running the following bash command:
```bash
for target in $(bazel query 'kind(crates_vendor, //...)') ; do bazel run ${target} -- --repin; done
```


dtolnay (commented):
I get a different result after running `crates_vendor` for wasm_bindgen/3rdparty, on the same commit.

It seems two of the `wait-timeout` crate's bins are getting rendered in a different order. 

It's unexpected to me that the generated BUILD files would be nondeterministic like this. 

```diff
diff --git a/wasm_bindgen/3rdparty/crates/BUILD.bazel b/wasm_bindgen/3rdparty/crates/BUILD.bazel
index 51206d6e..ed072257 100644
--- a/wasm_bindgen/3rdparty/crates/BUILD.bazel
+++ b/wasm_bindgen/3rdparty/crates/BUILD.bazel
@@ -207,8 +207,8 @@ alias(
 )
 
 alias(
-    name = "wait-timeout__exit",
-    actual = "@rules_rust_wasm_bindgen__wait-timeout-0.2.0//:exit__bin",
+    name = "wait-timeout__reader",
+    actual = "@rules_rust_wasm_bindgen__wait-timeout-0.2.0//:reader__bin",
     tags = ["manual"],
 )
 
@@ -219,8 +219,8 @@ alias(
 )
 
 alias(
-    name = "wait-timeout__reader",
-    actual = "@rules_rust_wasm_bindgen__wait-timeout-0.2.0//:reader__bin",
+    name = "wait-timeout__exit",
+    actual = "@rules_rust_wasm_bindgen__wait-timeout-0.2.0//:exit__bin",
     tags = ["manual"],
 )
 
diff --git a/wasm_bindgen/3rdparty/crates/BUILD.wait-timeout-0.2.0.bazel b/wasm_bindgen/3rdparty/crates/BUILD.wait-timeout-0.2.0.bazel
index 610cd069..f15ee858 100644
--- a/wasm_bindgen/3rdparty/crates/BUILD.wait-timeout-0.2.0.bazel
+++ b/wasm_bindgen/3rdparty/crates/BUILD.wait-timeout-0.2.0.bazel
@@ -144,7 +144,7 @@ rust_library(
 )
 
 rust_binary(
-    name = "exit__bin",
+    name = "reader__bin",
     srcs = glob(
         include = [
             "**/*.rs",
@@ -171,7 +171,7 @@ rust_binary(
     }),
     crate_features = [
     ],
-    crate_root = "src/bin/exit.rs",
+    crate_root = "src/bin/reader.rs",
     data = select_with_or({
         "//conditions:default": [
         ],
@@ -392,7 +392,7 @@ rust_binary(
 )
 
 rust_binary(
-    name = "reader__bin",
+    name = "exit__bin",
     srcs = glob(
         include = [
             "**/*.rs",
@@ -419,7 +419,7 @@ rust_binary(
     }),
     crate_features = [
     ],
-    crate_root = "src/bin/reader.rs",
+    crate_root = "src/bin/exit.rs",
     data = select_with_or({
         "//conditions:default": [
         ],
```

dtolnay:
I put up #1736 + #1737 to fix the nondeterminism. You might want to redo the re-pin on top of those.

examples/crate_universe/vendor_external/crates/defs.bzl
```diff
@@ -291,7 +291,7 @@ def aliases(
 _NORMAL_DEPENDENCIES = {
     "": {
         _COMMON_CONDITION: {
-            "clap": "@crates_vendor__clap-3.2.23//:clap",
+            "clap": "@crates_vendor__clap-3.1.5//:clap",
```
dtolnay, line 294: How come a bunch of crates are getting **downgraded** under examples/crate_universe/vendor_external?

- clap from 3.2.23 to 3.1.5
- clap_derive from 3.2.18 to 3.1.4
- form_urlencoded from 1.1.0 to 1.0.1
- getrandom from 0.2.8 to 0.2.5
- indexmap from 1.9.2 to 1.8.0
- libc from 0.2.137 to 0.2.119
- memchr from 2.5.0 to 2.4.1
- os_str_bytes from 6.4.1 to 6.0.0
- ppv-lite86 from 0.2.17 to 0.2.16
- proc-macro2 from 1.0.47 to 1.0.36
- quote from 1.0.21 to 1.0.15
- rand_core from 0.6.4 to 0.6.3
- regex from 1.7.0 to 1.5.4
- regex-syntax from 0.6.28 to 0.6.25
- semver from 1.0.14 to 1.0.6
- serde from 1.0.148 to 1.0.136
- syn from 1.0.105 to 1.0.86
- tinyvec from 1.6.0 to 1.5.1
- toml from 0.5.9 to 0.5.8
- unicode-bidi from 0.3.8 to 0.3.7
- unicode-normalization from 0.1.22 to 0.1.19
- url from 2.3.1 to 2.2.2

It's not something I was able to reproduce. I get just normal expected upgrades in that directory from running the command you said -- dfbad06651d80cf98accd012a966e4e40620e7b6.

- libc 0.2.137 to 0.2.139
- once_cell 1.16.0 to 1.17.0
- proc-macro2 1.0.47 to 1.0.49
- quote 1.0.21 to 1.0.23
- semver 1.0.14 to 1.0.16
- serde 1.0.148 to 1.0.152
- syn 1.0.105 to 1.0.107
- toml 0.5.9 to 0.5.10
- unicode-ident 1.0.5 to 1.0.6

Would you be able to share exactly what you ran to end up with this?

dtolnay (changes requested):

examples/crate_universe/vendor_external/crates/defs.bzl
UebelAndre: Good catch. I'd written a bash function that updates these and forgot I had the examples updated using `--repin=workspace`. The downgrade is interesting. The versions correctly match what's found in https://github.com/fnichol/names/blob/v0.13.0/Cargo.lock. This means that at some point, dependencies were re-pinned. I think the behavior though is undesirable. Anytime a lockfile is provided that lives in an external repository, I don't think re-pinning should be possible. Instead, users will need to provide one from within the repo. What do you think?

dtolnay: Okay I see what happened. I didn't notice it was being told to use a lockfile from some other repo.

My expectation would be, given this line:

https://github.com/bazelbuild/rules_rust/blob/532e60ff0ee3ac318333b3ae05392c220b17ce28/examples/crate_universe/vendor_external/BUILD.bazel#L10

`crates_vendor` should **only** change crate versions to exactly match what is in that lockfile, whether or not you passed `--repin`. When I run `//vendor_external:crates_vendor` currently, it violates this expectation.

The only way to "repin" those crates is by changing the above line to be a local lockfile, or by changing the content of the other repo by changing the urls here:

https://github.com/bazelbuild/rules_rust/blob/532e60ff0ee3ac318333b3ae05392c220b17ce28/examples/crate_universe/WORKSPACE.bazel#L289-L295

dtolnay (approved):

examples/crate_universe/vendor_external/crates/defs.bzl
UebelAndre: I filed https://github.com/bazelbuild/rules_rust/issues/1738 and can take a stab at it. Let me know if you have thoughts on the UX there. If repinning external lockfiles should be an error or a no-op or something.

UebelAndre:
@illicitonion I think this is ready to merge now.

illicitonion (approved):

PR #1736, opened 2023-01-01
Collect targets in a deterministic order

dtolnay:
Fixes the nondeterminism pointed out in https://github.com/bazelbuild/rules_rust/pull/1735#pullrequestreview-1233384845.

PR #1737, opened 2023-01-01
Eliminate all use of hash-based collections from crate_universe

dtolnay:
This is necessary together with #1736 to keep the generated BUILD files deterministic. Iterating a hashmap or hashset in hash order is a common source of nondeterminism in code generators (based on my experience with proc macros).

I haven't looked into which of the hashmaps changed in this diff are correctly used. It's *possible* to use a hashmap correctly, if all we do with them is lookups, not iteration. But it's simple enough to just not use hash-based containers entirely and never have hash order nondeterminism problems.


UebelAndre (approved):
Great find and thanks for the pull-request!

PR #1741, opened 2023-01-02
draft: fix data and compile_data for rust_doc

### LINKED ISSUE #1572: rust_doc_test fails for libraries with compile_data
`rust_doc_test` fails if used on a crate that has compile-time data dependencies.
Minimal example:

`BUILD.bazel` file:

```python
load(
    "@rules_rust//rust:defs.bzl",
    "rust_doc_test",
    "rust_library",
)

package(default_visibility = ["//visibility:public"])

rust_library(
    name = "lib",
    srcs = ["lib.rs"],
    compile_data = [":data.txt"],
    rustc_env = {
        "DATA_PATH": "$(execpath :data.txt)",
    },
)

rust_doc_test(
    name = "doc_test",
    crate = ":lib",
)
```

`lib.rs` file:

```rust
const DATA: &str = include_str!(env!("DATA_PATH"));

/// Returns the embedded textual data.
/// ```
/// assert!(!get_data().is_empty());
/// ```
pub fn get_data() -> &'static str {
    &DATA
}
```

`data.txt` file:

```
Hello, World!
```

How to reproduce:
Copy the files above into `examples/doc_test_compile_data` and run bazel test:

```
λ rules_rust/examples/doc_test_compile_data/ main* bazel test :doc_test
bazel test :doc_test

ERROR: /Users/lifted/Projects/rules_rust/examples/doc_test_compile_data/BUILD.bazel:18:14: in deps attribute of rust_doc_test rule //doc_test_compile_data:doc_test: source file '//doc_test_compile_data:data.txt' is misplaced here (expected no files)
ERROR: /Users/lifted/Projects/rules_rust/examples/doc_test_compile_data/BUILD.bazel:18:14: Analysis of target '//doc_test_compile_data:doc_test' failed
ERROR: Analysis of target '//doc_test_compile_data:doc_test' failed; build aborted: 
```

paulfariello:
Try to fix #1572 

Main issue is that data_paths do not inherit from original crate when building rust_doc (and probably rust_test).
The main idea is to rely on crate_info which inherit correctly from original crate.

Issue with this draft is that compile_data is filed with File instead of Target when building 3rd party crate.

```
ERROR: <redacted>__base64-0.13.1/BUILD.bazel:34:13: in rust_library rule @<redacted>__base64-0.13.1//:base64:
	File "external/rules_rust/rust/private/rust.bzl", line 197, column 32, in _rust_library_impl
		return _rust_library_common(ctx, "rlib")
	File "external/rules_rust/rust/private/rust.bzl", line 292, column 32, in _rust_library_common
		return rustc_compile_action(
	File "external/rules_rust/rust/private/rustc.bzl", line 1093, column 46, in rustc_compile_action
		args, env_from_args = construct_arguments(
	File "external/rules_rust/rust/private/rustc.bzl", line 915, column 38, in construct_arguments
		expand_list_element_locations(
	File "external/rules_rust/rust/private/utils.bzl", line 311, column 53, in expand_list_element_locations
		return [_expand_location_for_build_script_runner(ctx, arg, data) for arg in args]
	File "external/rules_rust/rust/private/utils.bzl", line 255, column 30, in _expand_location_for_build_script_runner
		dedup_expand_location(ctx, env, data),
	File "external/rules_rust/rust/private/utils.bzl", line 230, column 31, in dedup_expand_location
		return ctx.expand_location(input, _deduplicate(targets))
Error in expand_location: at index 0 of targets, got element of type File, want Target
```

### DIFF
--- a/proto/proto.bzl
+++ b/proto/proto.bzl
@@ -246,7 +246,7 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, is_gr
             rustc_env = {},
             is_test = False,
             compile_data = depset([target.files for target in getattr(ctx.attr, "compile_data", [])]),
-            compile_data_targets = getattr(ctx.attr, "compile_data", depset([])),
+            compile_data_targets = depset(getattr(ctx.attr, "compile_data", [])),
             wrapped_crate_type = None,
             owner = ctx.label,
         ),
--- a/proto/proto.bzl
+++ b/proto/proto.bzl
@@ -246,6 +246,7 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, is_gr
             rustc_env = {},
             is_test = False,
             compile_data = depset([target.files for target in getattr(ctx.attr, "compile_data", [])]),
+            compile_data_targets = getattr(ctx.attr, "compile_data", depset([])),
             wrapped_crate_type = None,
             owner = ctx.label,
         ),
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -906,10 +906,10 @@ def construct_arguments(
     rustc_flags.add_all(rust_flags)
 
     # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
-    data_paths = getattr(attr, "data", []) + crate_info.compile_data_targets.to_list()
+    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets])
 
     # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
-    data_paths = depset(direct = data_paths).to_list()
+    data_paths = data_paths.to_list()
 
     rustc_flags.add_all(
         expand_list_element_locations(
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -906,16 +906,14 @@ def construct_arguments(
     rustc_flags.add_all(rust_flags)
 
     # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
-    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets])
-
     # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
-    data_paths = data_paths.to_list()
+    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets])
 
     rustc_flags.add_all(
         expand_list_element_locations(
             ctx,
             getattr(attr, "rustc_flags", []),
-            data_paths,
+            data_paths.to_list(),
         ),
     )
     add_edition_flags(rustc_flags, crate_info)
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -907,13 +907,13 @@ def construct_arguments(
 
     # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
     # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
-    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets])
+    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets]).to_list()
 
     rustc_flags.add_all(
         expand_list_element_locations(
             ctx,
             getattr(attr, "rustc_flags", []),
-            data_paths.to_list(),
+            data_paths,
         ),
     )
     add_edition_flags(rustc_flags, crate_info)
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -905,8 +905,11 @@ def construct_arguments(
     rustc_flags.add_all(rust_std_paths, before_each = "-L", format_each = "%s")
     rustc_flags.add_all(rust_flags)
 
+    # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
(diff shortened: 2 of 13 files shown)


illicitonion:
Thanks for putting together this fix!

It looks like the issue is that whenever we construct a `CrateInfo` (which we do several times in rust/private/rust.bzl, for instance), we convert `compile_data` from an attribute (targets) to files as required by the `CrateInfo` provider (by using `ctx.files.compile_data` instead of `ctx.attr.compile_data`).

I think to fix this we'd need to:
1. Modify `CrateInfo` to have an additional field, `compile_data_targets` here: https://github.com/bazelbuild/rules_rust/blob/d5289ad1c933155cf5894e5dd22f5cc4e22e124f/rust/private/providers.bzl#L17-L43
2. Wherever we construct a `CrateInfo`, start also setting this field (to `ctx.attr.compile_data` rather than `ctx.files.compile_data`)
3. In your change, consume `compile_data_targets` rather than `compile_data`.

Then I suspect your change will start working 🤞

proto/proto.bzl
```diff
@@ -246,6 +246,7 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, is_gr
             rustc_env = {},
             is_test = False,
             compile_data = depset([target.files for target in getattr(ctx.attr, "compile_data", [])]),
+            compile_data_targets = getattr(ctx.attr, "compile_data", depset([])),
```
illicitonion, line 249:
```suggestion
            compile_data_targets = depset(getattr(ctx.attr, "compile_data", [])),
```

rust/private/rustc.bzl
```diff
@@ -905,8 +905,11 @@ def construct_arguments(
     rustc_flags.add_all(rust_std_paths, before_each = "-L", format_each = "%s")
     rustc_flags.add_all(rust_flags)
 
+    # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
+    data_paths = getattr(attr, "data", []) + crate_info.compile_data_targets.to_list()
+
     # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
-    data_paths = depset(direct = getattr(attr, "data", []) + getattr(attr, "compile_data", [])).to_list()
+    data_paths = depset(direct = data_paths).to_list()
```
illicitonion, line 912: Rather than flattening this list twice, can you instead initially construct a depset and then to_list it once?

illicitonion (commented):
LGTM! Last up, can you add a test which was failing without this change, which shows rust_doc now works with compile_data?

Thanks!

rust/private/rustc.bzl
paulfariello: done

proto/proto.bzl
paulfariello: done

paulfariello:
For the record, the added test would fail on main branch at build time with the following error:
```
INFO: Invocation ID: 6aec52b5-1be1-471a-8513-af33247b3398
ERROR: /home/redacted/rules_rust/test/unit/compile_data/BUILD.bazel:3:24: in rust_doc rule //test/unit/compile_data:compile_data_env_rust_doc: label '//test/unit/compile_data:compile_data.txt' in $(location) expression is not a declared prerequisite of this rule
ERROR: /home/redacted/rules_rust/test/unit/compile_data/BUILD.bazel:3:24: in (an implicit dependency) attribute of rust_doc rule //test/unit/compile_data:compile_data_env_rust_doc: $(location) not defined. Since this rule was created by the macro 'compile_data_test_suite', the error might have been caused by the macro implementation
ERROR: /home/redacted/rules_rust/test/unit/compile_data/BUILD.bazel:3:24: Analysis of target '//test/unit/compile_data:compile_data_env_rust_doc' failed
ERROR: Analysis of target '//test/unit/compile_data:compile_data_env_rust_doc' failed; build aborted:
INFO: Elapsed time: 8.315s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (107 packages loaded, 1951 targets configured)
ERROR: Couldn't start the build. Unable to run tests
```

test/unit/compile_data/compile_data_test.bzl
```diff
@@ -65,6 +80,21 @@ def _define_test_targets():
         rustc_flags = ["--cfg=test_compile_data"],
     )
 
+    rust_library(
+        name = "compile_data_env",
+        srcs = ["compile_data_env.rs"],
+        compile_data = ["compile_data.txt"],
+        rustc_env = {
+            "COMPILE_DATA_PATH": "$(location :compile_data.txt)",
```
illicitonion, line 88: `location` is soft-deprecated - let's use `execpath` instead:

```suggestion
            "COMPILE_DATA_PATH": "$(execpath :compile_data.txt)",
```

rust/private/rustc.bzl
```diff
@@ -905,8 +905,11 @@ def construct_arguments(
     rustc_flags.add_all(rust_std_paths, before_each = "-L", format_each = "%s")
     rustc_flags.add_all(rust_flags)
 
+    # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
+    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets])
+
     # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
```
illicitonion, line 911: It's the `depset` construction which does the deduplication, so let's leave the comment on that piece

```suggestion
    # Gather data path from crate_info since it is inherited from real crate for rust_doc and rust_test
    # Deduplicate data paths due to https://github.com/bazelbuild/bazel/issues/14681
    data_paths = depset(direct = getattr(attr, "data", []), transitive = [crate_info.compile_data_targets])
```

illicitonion (approved):
LGTM - thanks for putting this all together! Two tiny suggestions and we're good to go!

PR #1742, opened 2023-01-03
Don't emit .pdb output files using mingw-gcc

waych:
mingw-gcc doesn't support emitting debug data in separate .pdb files the way the msvc toolchain does.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1168,12 +1168,12 @@ def rustc_compile_action(
     # The action might generate extra output that we don't want to include in the `DefaultInfo` files.
     action_outputs = list(outputs)
 
-    # Rustc generates a pdb file (on Windows) or a dsym folder (on macos) so provide it in an output group for crate
+    # Rustc generates a pdb file (on Windows MSVC) or a dsym folder (on macos) so provide it in an output group for crate
     # types that benefit from having debug information in a separate file.
     pdb_file = None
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Sorry for the delay! Could you link some issues this lead to or documentation on `.pdb` files and why `mingw-gcc` doesn't support it? There's a lack of windows knowledge among the maintainers so we have to rely on users making thoroughly documented and tested changes to help avoid regressions.

Additionally, the `*-windows-gnu` targets are hard to test since we don't have a way to distinguish it from `*-windows-msvc` so I struggle to think of how this could be tested. If you can think of a way that'd be great 😅

waych:
I haven't had the time to give much attention to this stream of development recently (including e.g. #1740 and #1749).

Since I sent these out, I started to look at how to pull a full mingw build / test into the rust_rules CI, but along the way I discovered a couple more test breaks (~4?) that needed some massaging to un special-case from being "windows" and instead trigger for "windows-msvc". The "windows-gnu" builds should work much like the linux builds in terms of artifacts and toolchain capabilities, so inline DWARF data for debug.

I have mingw + rust building working Windows artifacts in my local buildbarn setup, but it'll also still take more work to clean it up hermetically so that it can be used in the CI neatly and have all tests pass.

I won't be able to get back to this for a while, maybe a couple weeks.

PR #1743, opened 2023-01-03
Convert BUILD.$name-$version.bazel from tera to serde_starlark

dtolnay:
Context: see #1734.

This PR converts all of `crate_build_file.j2`, `partials/crate/*.j2` and `partials/starlark/*.j2` to serde_starlark.

### Remarks

- I kept the output of `bazel run //crate_universe/3rdparty:crates_vendor` out of this PR to keep GitHub's code review interface usable. You can find the output generated code in 44c264a7845701239e50d643b78b2bf4a117d425.

- I tried to keep the conversion 1-to-1 to the largest extent possible. As an example, build_script.j2 duplicated most of the content of common_attrs.j2 even though it probably could have used that template via include, and I have mirrored the same here by not using `#[serde(flatten)] common: CommonAttrs` inside of `CargoBuildScript`. There is probably plenty of opportunity for cleaning up things in followup PRs.

- I came across several bugs in the existing code. Again, I mostly stayed away from these to keep the PR bounded in scope. For example the fact that `rustc_flags` is stored as a BTreeSet\<String\> is not correct; compiler flags are ordered and sorting them is not correct. For example `-C debuginfo=2`...

    https://github.com/bazelbuild/rules_rust/blob/532e60ff0ee3ac318333b3ae05392c220b17ce28/crate_universe/src/context/crate_context.rs#L178

- One bug that I fixed because in the new implementation it was easier to fix than to not fix, is that all existing uses of selectable_list.j2 and selectable_dict.j2 were broken because they all had the same failure mode as #1417. The generated code using `selects.with_or` could end up generating a map with duplicate keys if multiple distinct Cargo cfg expressions mapped to the same subset of supported platform triples. My implementation follows the same approach as #1647, but that PR only applied it to `deps` and `aliases`, not e.g. `rustc_env` and everything else.

- Anecdotally the new implementation has strong "if it compiles, it works" property. I mostly typed out the implementation of this PR top to bottom without running `crates_vendor`, and as soon as the whole thing was written and the compiler was okay with it, it was generating the output I wanted. I have not done much work in the templates in the past but I do not imagine they have this property.

### Remaining work

- `select_with_or` in private/selects.bzl is no longer used by the code generator's output. I had to keep it in this PR because all the generated BUILD files still include calls to it. It can be deleted after all those are regenerated.

- One of the tera templates still uses the old serialization of `SelectList` so the starlark serialization for it isn't a normal `Serialize` impl but instead an associated function. This causes there to be a bunch of `serialize_with = "SelectList::serialize_starlark"` attributes needed. This can be cleaned up after the last tera template is converted.

- ~~A lot of the contents of `crate::context` can be cleaned up after they're no longer used by tera. Tera needs those data structures to be serializable, whereas with serde_starlark we're not serializing anything from `crate::context`, only from `crate::utils::starlark`. For now I've just added a bunch of `serde(skip)` attributes in `crate::context` on all the parts tera is no longer using.~~ *Edit: apparently we still need to serialize all the fields in `Context` so that they end up in cargo-bazel-lock.json.*

- There are 2 remaining tera templates after this PR, module_bzl.j2 and vendor_module.j2.


crate_universe/private/selects.bzl
```diff
@@ -1,41 +1,43 @@
...
-        no_match_error: Optional custom error to report if no condition matches.
+selects = struct(
+    with_unmapped = _with_unmapped,
+    NO_MATCHING_PLATFORM_TRIPLES = _SENTINEL,
+)
 
-    Returns:
-        A native `select()`
-    """
+# FIXME(dtolnay): No longer used by the serde_starlark-based renderer. Delete
```
illicitonion, line 36: TODO

crate_universe/src/utils/starlark/serialize.rs
```diff
@@ -1,7 +1,86 @@
...
+}
+
+pub fn rust_binary<S>(rule: &RustBinary, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    FunctionCall::new("rust_binary", rule).serialize(serializer)
+}
+
+// Serialize an array with each element on its own line, even if there is just a
```
illicitonion, line 34: I think we can remove this - our real goal is to be buildifier-compatible and buildifier accepts either, but happy to keep it also (or to remove it in a follow-up).

crate_universe/src/utils/starlark/serialize.rs
```diff
@@ -23,3 +102,25 @@ impl Serialize for ExportsFiles {
...
+        self.glob.is_empty() && self.select.is_empty()
+    }
+}
+
+impl Serialize for Data {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut plus = serializer.serialize_tuple_struct("+", MULTILINE)?;
```
illicitonion, line 117: It would be nice to only do the plus-ing if both are non-empty (again, we can do in a follow-up)

crate_universe/src/utils/starlark/select.rs
```diff
@@ -19,15 +24,24 @@ pub trait Select<T> {
...
+    // Invariant: any T in `common` is not anywhere in `selects`.
     common: BTreeSet<T>,
+    // Invariant: none of the sets are empty.
     selects: BTreeMap<String, BTreeSet<T>>,
+    // Elements that used to be in `selects` before the most recent
+    // `remap_configurations` operation, but whose old configuration did not get
+    // mapped to any new configuration. They could be ignored, but are preserved
+    // here to generate comments that help the user understand what happened.
+    #[serde(skip_serializing_if = "BTreeSet::is_empty", default = "BTreeSet::new")]
+    unmapped: BTreeSet<T>,
```
illicitonion, line 36: Was adding `unmapped` somehow required by this change, or is it an enhancement for documentation purposes? It looks like the latter (which is fine), but I want to make sure I understand if not.

illicitonion (approved):
This looks great, thanks for putting it together! Feel free to update the vendored files (as a separate commit in this PR)

crate_universe/src/utils/starlark/serialize.rs
dtolnay: This is worth keeping IMO. Note that the only case I use it in is inside `select`:

```bzl
    deps = select({
        "@rules_rust//rust/platform:aarch64-pc-windows-msvc": [
            "@cui__winapi-util-0.1.5//:winapi_util",  # cfg(windows)
        ],
        "@rules_rust//rust/platform:i686-pc-windows-msvc": [
            "@cui__winapi-util-0.1.5//:winapi_util",  # cfg(windows)
        ],
        "@rules_rust//rust/platform:x86_64-pc-windows-msvc": [
            "@cui__winapi-util-0.1.5//:winapi_util",  # cfg(windows)
        ],
        "//conditions:default": [],
    }),
```

compared to this which I find a little too compact to be easily readable, with the cfg comment way out of view:

```bzl
    deps = select({
        "@rules_rust//rust/platform:aarch64-pc-windows-msvc": ["@cui__winapi-util-0.1.5//:winapi_util"],  # cfg(windows)
        "@rules_rust//rust/platform:i686-pc-windows-msvc": ["@cui__winapi-util-0.1.5//:winapi_util"],  # cfg(windows)
        "@rules_rust//rust/platform:x86_64-pc-windows-msvc": ["@cui__winapi-util-0.1.5//:winapi_util"],  # cfg(windows)
        "//conditions:default": [],
    }),
```

Everywhere else I'm having arrays serialize using serde_starlark's default line wrapping, which is multiline for arrays &gt;1 element:

```bzl
    srcs = glob(["**/*.rs"]),
    compile_data = glob(
        include = ["**"],
        exclude = [
            "**/* *",
            "BUILD",
            "BUILD.bazel",
            "WORKSPACE",
            "WORKSPACE.bazel",
        ],
    ),
    tags = [
        "cargo-bazel",
        "manual",
        "noclippy",
        "norustfmt",
    ],
```

crate_universe/src/utils/starlark/serialize.rs
dtolnay: I think that's equivalent to what this is already doing. This as currently written serializes to:

| `self.glob` | `self.select.common` | `self.select.selects` | serialized |
|---|---|---|---|
| empty | empty | empty | `[]` |
| empty | empty | non-empty | `select(…)` |
| empty | non-empty | empty | `[…]` |
| empty | non-empty | non-empty | `[…] + select(…)` |
| non-empty | empty | empty | `glob(…)` |
| non-empty | empty | non-empty | `glob(…) + select(…)` |
| non-empty | non-empty | empty | `glob(…) + […]` |
| non-empty | non-empty | non-empty | `glob(…) + […] + select(…)` |

crate_universe/src/utils/starlark/select.rs
dtolnay: Keeping track of unmapped configurations isn't new, it was added in https://github.com/bazelbuild/rules_rust/pull/1647 (calling it to your attention because I don't see that you were involved in reviewing that PR).

The way that PR implemented it was:

```rust
impl<T: Ord + Clone> SelectList<T> {
    pub fn remap_configurations<'a, I, S>(
        &self,
        mapping: &'a BTreeMap<String, I>,
    ) -> (
        SelectList<WithOriginalConfigurations<T>>,
        BTreeMap<String, BTreeSet<T>>,
    )
```

where "unmapped" configurations are returned separately from SelectList, in the second element of the tuple.

We could continue to keep unmapped configurations out of `SelectList` and shuffle them around separately everywhere, like this:

```rust
pub struct RustLibrary {
    ...
    pub deps: (SelectList<WithOriginalConfigurations<String>>, BTreeMap<String, BTreeSet<String>>),
    pub proc_macro_deps: (SelectList<WithOriginalConfigurations<String>>, BTreeMap<String, BTreeSet<String>>),
    pub rustc_env: (SelectDict<WithOriginalConfigurations<String>>, BTreeMap<String, BTreeMap<String, T>>),
    pub rustc_flags: (SelectList<WithOriginalConfigurations<String>>, BTreeMap<String, BTreeSet<String>>),
```

It's just that... literally everywhere crate_universe wants to  serialize a SelectList, you also want to serialize all the configurations that got mapped away out of that select, to help the user understand what happened to them. So I've combined it together into one struct.

dtolnay:
I made 2 force pushes:

- first, purely a rebase of the already reviewed code to fix merge conflict with #1735.
  https://github.com/bazelbuild/rules_rust/compare/5a39df2dc26e6bbb6805fffb742b6c218bc0d130..d5a493e47989ad51cc38445c4b0500dbfc3257fd

- second a small change to address https://github.com/bazelbuild/rules_rust/pull/1743#discussion_r1060791553 to replace `FIXME` -> `TODO`.<br>https://github.com/bazelbuild/rules_rust/compare/d5a493e47989ad51cc38445c4b0500dbfc3257fd..bba5b8b0550988056d6867be5274ee9d9023a97b

crate_universe/private/selects.bzl
dtolnay: Fixed.

dtolnay:
- Rebased to resolve conflict with #1718

illicitonion (approved):
Ship it! Thanks!

dtolnay:
Clean rebase.

PR #1745, opened 2023-01-03
Propagate `compatible_with` attribute to the underlying `_build_script_run` target

scentini:
Currently, if a target is annotated with `compatible_with` and depends on a `cargo_build_script` (also annotated with the appropriate `compatible_with`), we hit an issue:
```
in rust_test rule //test/cargo_build_script:test_compatible_with: dependency //test/cargo_build_script:empty_build_script doesn't support expected environment: //test/cargo_build_script:bar
```
This is because the `cargo_build_script` macro doesn't forward the `compatible_with` attribute to the underlying `_build_script_run` target. 

This PR fixes the issue by propagating the `compatible_with` attribute to the `_build_script_run` target.


test/cargo_build_script/BUILD.bazel
```diff
@@ -27,3 +27,35 @@ rust_test(
...
+)
+
+cargo_build_script(
+    name = "empty_build_script",
+    srcs = ["do_nothing.rs"],
+    compatible_with = [":bar"],
+    edition = "2018",
+)
+
+rust_test(
```
illicitonion, line 55: I'm not sure the test adds much value as it's a thing you need to manually run to observe, but 👍

illicitonion (approved):

scentini:
@illicitonion thanks for the speedy review!!

scentini:
@UebelAndre I couldn't point you to proper documentation on `compatible_with` in https://github.com/bazelbuild/rules_rust/pull/1537#issuecomment-1255027150, but this PR contains a test using it, in case you're interested.

PR #1746, opened 2023-01-03
Regenerate BUILD files using serde_starlark renderer

dtolnay:
This PR is the result of `for target in $(bazel query 'kind(crates_vendor, //...)') ; do bazel run ${target}; done` in each of `./` and `./examples/crate_universe`.

I saw that this repo only uses squash merges and I was hoping not to get this large generated change squashed into #1743 so that the implementation commit in crate_universe remains understandable afterward. I can rebase the generated files here after #1743 gets merged.


dtolnay:
Note I did not update `./examples/crate_universe_unnamed` because I think that was missed from #1735, so running `crates_vendor` in there involves a bunch of versions getting changed. I wanted this PR to be just regenerating existing BUILD files and not bundled together with version changes.

UebelAndre (approved):
Thank you!

PR #1747, opened 2023-01-03
Exclude generated files from language stats and collapse in code review

dtolnay:
This has 2 effects:

- All the generated Starlark will stop counting for this repo's language stats:<br>

- Generated files will get collapsed by default in code review, similar to how Cargo.lock is always currently collapsed. You can click **"Load diff"** to inspect the generated change.<br>


UebelAndre (approved):
Neat! Thanks!

UebelAndre:
@dtolnay ready to merge pending a rebase 😄

dtolnay:
Rebased, though I'm not sure why you needed this to be rebased. There is no possibility that this would conflict with any other PR unless some other PR were also adding a .gitattributes file, and in that case GitHub would have prevented the merge.

illicitonion:
> I'm not sure why you needed this to be rebased. There is no possibility that this would conflict with any other PR

I believe the repo has "require all PRs to be up to date" enabled which means the merge-base must be the HEAD of the target branch - the goal being to ensure tests were run against the HEAD state of the repo, rather than a stale HEAD, which could cause _logical_ merge conflicts rather than literal ones.

That said, I think the setting causes more hassle than benefit... Personally, I would disable it - @hlopko do you have thoughts?

(Most PR authors use the enabled-by-default "allow maintainers to push to my PRs" setting, which allows us to press a button to merge main on your behalf, which lessens the hassle somewhat for contributors).

dtolnay:
Ah okay -- I know about the "Allow edits and access to secrets by maintainers" but unfortunately it doesn't exist for PRs that come from organizations, only from personal forks. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork emphasizes this too: it only works for _"forks owned by a personal account"_.

UebelAndre:
Unless there's a way to test a change against HEAD of main pre-merge, I think requiring a rebase is an acceptable test strategy.

PR #1749, opened 2023-01-04
mingw: missing rsbegin.o/rsend.o

waych:
When building using mingw-w64, RBE is rendered broken due to missing `rsbegin.o` and `rsend.o` at final executable binary link.  This is due to this filegroup not including these files. The reason is that on the windows target uses `.lib` for `staticlib_ext`, so these files that distributed as part of the stdlib are missed.

Putting a glob for `*.o` here isn't the most elegant solution, however I struggled to come up with something more elegant that wasn't adding even more conditional logic, given the use of `allow_empty = True`.


illicitonion (commented):
This looks reasonable, but I'd feel much more comfortable with it if we had a failing test this was fixing - is it possible to write one? Otherwise I suspect this will end up getting broken again down the line...

waych:
Agreed this would be better with a test.  I'm not very familiar with how to test bazel rules, so I'm studying up on how that is done in the tests directory and reading up unittest.bzl.

illicitonion:
From the look of the PRs you've been putting together, I suspect it may actually be preferable to modify https://github.com/bazelbuild/rules_rust/blob/main/.bazelci/presubmit.yml to run all the existing tests with a new toolchain, rather than writing a unit test - it looks like you're finding places a toolchain has different assumptions than others, which we're realistically only going to discover through running the toolchain.

illicitonion:
> I'm not familiar with buildkite yet, but will take a look at what's required. I imagine it'd be something that looks like a derivative of ubuntu2004_clang_lld with an `--extra_toolchains=` pointing at the toolchain definition somewhere in the tree?

Yeah, that sounds about right :)

PR #1751, opened 2023-01-04
Generate only the needed subset of binaries for bindgen and proto

dtolnay:
This removes, for example, the `protobuf-bin-gen-rust-do-not-use` binary that was previously getting exposed.

https://github.com/bazelbuild/rules_rust/blob/d6e3003594713ba00d03a354cbf7a7c2c65e22d2/proto/3rdparty/crates/BUILD.bazel#L77-L81

I had to change `impl Add for CrateAnnotations` to make crate-level `gen_binaries` work. It turns out in #1718 I didn't fully understand what this `Add` impl was supposed to do, and I had only tested `gen_binaries` _before_ adding the global-level `generate_binaries` argument (https://github.com/bazelbuild/rules_rust/pull/1718#discussion_r1055669481). (I'm still not sure that I understand what this `Add` is supposed to do exactly, but at least now it appears to work in practice.)

The behavior I observed before the change in this PR was that global-level `generate_binaries = True` would do the right thing but crate-level `gen_binaries` would seem to get ignored whether or not global-level `generate_binaries` was set.

### DIFF
--- a/bindgen/3rdparty/BUILD.bazel
+++ b/bindgen/3rdparty/BUILD.bazel
@@ -5,6 +5,7 @@ crates_vendor(
     name = "crates_vendor",
     annotations = {
         "bindgen": [crate.annotation(
+            gen_binaries = ["bindgen"],
             gen_build_script = True,
         )],
         "clang-sys": [crate.annotation(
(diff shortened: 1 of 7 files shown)


UebelAndre (approved):
You’re doing the lord’s work here. Thank you so much!

PR #1752, opened 2023-01-04
Repin ./examples/crate_universe_unnamed

dtolnay:
This directory was missed from #1735 (and so I'd also excluded it from #1746).

Generated by `for target in $(bazel query 'kind(crates_vendor, //...)'); do bazel run ${target} -- --repin; done`.


UebelAndre (approved):
Thank you!

PR #1754, opened 2023-01-05
Don't propagate `compatible_with` to the underlying `cargo_build_script` `rust_binary` target

scentini:
I mostly fixed the interaction between `compatible_with` and `cargo_build_script` in https://github.com/bazelbuild/rules_rust/pull/1745, however I forgot to remove the `compatible_with` attribute from the underlying `rust_binary` target. This target shouldn't have the `compatible_with` specification as we only depend on it in `exec` mode.

### DIFF
--- a/cargo/private/cargo_build_script_wrapper.bzl
+++ b/cargo/private/cargo_build_script_wrapper.bzl
@@ -94,7 +94,10 @@ def cargo_build_script(
         rustc_flags (list, optional): List of compiler flags passed to `rustc`.
         visibility (list of label, optional): Visibility to apply to the generated build script output.
         tags: (list of str, optional): Tags to apply to the generated build script output.
-        **kwargs: Forwards to the underlying `rust_binary` rule.
+        **kwargs: Forwards to the underlying `rust_binary` rule. An exception is the `compatible_with`
+            attribute, which shouldn't be forwarded to the `rust_binary`, as the `rust_binary` is only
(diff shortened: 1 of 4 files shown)


UebelAndre:
For the description, do you mean `rust_binary`?

scentini:
@UebelAndre Indeed, thanks!

UebelAndre (approved):

cargo/private/cargo_build_script_wrapper.bzl
```diff
@@ -112,7 +112,7 @@ def cargo_build_script(
         binary_tags.append("manual")
     build_script_kwargs = {}
     if "compatible_with" in kwargs:
```
krasimirgg, line 114: hm, this is fine logically but it's a bit of mental gymnastics to figure out what's happening.
For readability, could we maybe introduce a new variable `binary_kwargs`, initialized with kwargs and pop the `compatible_with` from that one, leaving the `kwargs` unchanged?
Also, could we update the docs `**kwargs: Forwards to the underlying `rust_binary` rule.` a bit to spell out the compatible_with behavior?

krasimirgg (changes requested):

cargo/private/cargo_build_script_wrapper.bzl
scentini: Done.

krasimirgg (approved):

PR #1755, opened 2023-01-05
Remove unnecessary binary file

scentini:
This `.so` is not used anywhere, and is the cause of https://github.com/bazelbuild/rules_rust/issues/1701. The `.so` files we use for the tests under `//test/versioned_dylib` are compiled from source via [`cc_binary`s](https://github.com/bazelbuild/rules_rust/blob/0.15.0/test/versioned_dylib/c/BUILD.bazel#L4).


UebelAndre (approved):
Thanks!

PR #1772, opened 2023-01-12
Update clippy and rustfmt aspects to require CrateInfo providers

UebelAndre:


illicitonion (approved):

rust/private/clippy.bzl
```diff
@@ -225,6 +225,7 @@ rust_clippy_aspect = aspect(
         ),
     },
     provides = [ClippyInfo],
+    required_providers = [rust_common.crate_info],
```
scentini, line 228: `rust_shared_library` and `rust_static_library` do not emit a `CrateInfo` directly (because their  use case is to produce artifacts that are then used from other languages), but we still want to be able to run clippy and rustfmt on them. This should work:
```
required_providers = [[rust_common.crate_info], [rust_common.test_crate_info]] 
```

scentini: That's a good question, and looks like we currently don't run the checks on targets that don't provide `CrateInfo` already. I'll look into reenabling them independently of this PR.

rust/private/rustfmt.bzl
```diff
@@ -134,6 +149,48 @@ generated source files are also ignored by this aspect.
...
+    srcs = _find_rustfmtable_srcs(crate_info, ctx)
+    manifest = _generate_manifest(edition, srcs, ctx)
+
+    return [
+        OutputGroupInfo(
+            rustfmt_manifest = depset([manifest]),
+        ),
+    ]
+
+# This aspect contains functionality split out of `rustfmt_aspect` which broke when
```
UebelAndre, line 182: This feels like a bug upstream. @scentini have you ever encountered this?

rust/private/clippy.bzl
UebelAndre: Got clippy and rustfmt working with rules that provide `TestCrateInfo`. I updated the tests for `rustfmt` but not `clippy`. Maybe someone could add that in a separate PR?

scentini: Sounds good!

rust/private/rustfmt.bzl
scentini: Nope, but I could check and file a bug against Bazel if the claim still holds.

scentini (approved):

rust/private/rustfmt.bzl
UebelAndre: https://github.com/bazelbuild/bazel/issues/17214

PR #1777, opened 2023-01-13
rustdoc_test: substitute the root of the current crate

bsilver8192:
Otherwise a trivial little crate with no rust_library dependencies, and a build script dependency, fails. This is unusual because most crates complex enough to have a build script also have other dependencies, I ran into while minimizing a test case for another bug.


illicitonion (approved):
Thanks!

PR #1779, opened 2023-01-16
rustdoc_test: fix and test OUT_DIR

bsilver8192:
It needs the same treatment as SYSROOT.

The new test fails without #1777, this will need to be rebased after that is merged to pass CI.


bsilver8192:
For reference, the CI failures now look like GPG keys for an apt repository got rotated, not related to this PR.

Issue #1762, opened 2023-01-06
paths passed to build script by runner that are relative to the `execroot` cause build script to fail


xortive:
I have a custom toolchain that builds it's own `libc++` and sets a dependency from the toolchain to said `libc++`, and sets `execroot`-relative `-isystem` flags. When these are passed to a build script, it fails as the build script is not run from inside the execroot.

It appears that some other flags with this problem like `--sysroot` have special-cased handling, but I'm not sure that manually keeping track of compiler flags that may contain `execroot`-relative paths and replacing them with `{pwd}` for the build script runner is a long-term solution here, but I don't have other ideas.

illicitonion:
Yeah, I definitely agree with this problem statement.

Unfortunately, I also don't have any great answers...

Possibly a slightly better heuristic could be that we could search through flag values for things that look like paths which exist, or look like directories which exist, but that's starting to get a bit scary - if someone's, say, trying to use a custom `remap-path-prefix`, or embed some path as a string rather than by reading the file, we may do the wrong thing for them.

But at its core there's a mismatch between where and how bazel and cargo run things which is hard to resolve. I'd love to hear any ideas people have, though!

bsilver8192:
My current solution is a `cc::Build` wrapper which:
* Canonicalizes all the input paths as they're added
* Generates and uses a wrapper for the compiler itself that `cd`s back to the place all the Bazel toolchain paths expect, using the strategy from #1555

I remember having similar struggles with Go several years ago. They just had an ever-growing list of flags to substitute, which I had to expand. IIRC it was in the Go tooling itself too, not `rules_go`, because the Go tools executed by `rules_go` also `cd`. Might be worth looking at how that currently works to see if they've come up with anything better in the meantime.

Neither of those are very attractive ideas, of course.

PR #1781, opened 2023-01-17
Symlink in the exec-root so that relative paths will work, unchanged.

freeformstu:
#1762

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -27,7 +27,6 @@ default_windows_targets: &default_windows_targets
   - "-//bindgen/..."
   - "-//test/proto/..."
   - "-//test/unit/pipelined_compilation/..."
-  - "-//test/cargo_build_script:test_exec_root_access.build.feature_enabled"
 crate_universe_vendor_example_targets: &crate_universe_vendor_example_targets
   - "//vendor_external:crates_vendor"
   - "//vendor_local_manifests:crates_vendor"
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -27,6 +27,7 @@ default_windows_targets: &default_windows_targets
   - "-//bindgen/..."
   - "-//test/proto/..."
   - "-//test/unit/pipelined_compilation/..."
+  - "-//test/cargo_build_script:test_exec_root_access.build.feature_enabled"
 crate_universe_vendor_example_targets: &crate_universe_vendor_example_targets
   - "//vendor_external:crates_vendor"
   - "//vendor_local_manifests:crates_vendor"
--- a/cargo/cargo_build_script_runner/bin.rs
+++ b/cargo/cargo_build_script_runner/bin.rs
@@ -53,6 +53,27 @@ fn run_buildrs() -> Result<(), String> {
     create_dir_all(&out_dir_abs)
         .unwrap_or_else(|_| panic!("Failed to make output directory: {:?}", out_dir_abs));
 
+    if should_symlink_exec_root() {
+        // Symlink the execroot to the manifest_dir so that we can use relative paths in the arguments.
+        let exec_root_paths = std::fs::read_dir(&exec_root)
+            .map_err(|err| format!("Failed while listing exec root: {err:?}"))?;
+        for path in exec_root_paths {
+            let path = path
+                .map_err(|err| {
+                    format!("Failed while getting path from exec root listing: {err:?}")
+                })?
+                .path();
+
+            let file_name = path
+                .file_name()
+                .ok_or_else(|| "Failed while getting file name".to_string())?;
+            let link = manifest_dir.join(file_name);
+
+            symlink_if_not_exists(&path, &link)
+                .map_err(|err| format!("Failed to symlink {path:?} to {link:?}: {err}"))?;
+        }
+    }
+
     let target_env_vars =
         get_target_env_vars(&rustc_env).expect("Error getting target env vars from rustc");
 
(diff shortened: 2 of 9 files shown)


cargo/cargo_build_script_runner/bin.rs
```diff
@@ -53,6 +53,46 @@ fn run_buildrs() -> Result<(), String> {
...
+            })?
+            .path();
+        let metadata = std::fs::metadata(&path).map_err(|err| {
+            format!(
+                "Failed while getting metadata for path {:?}: {:?}",
+                path, err
+            )
+        })?;
+
+        // Only symlink directories.
```
illicitonion, line 75: Why not symlink files too?

illicitonion (commented):
This is a pretty cute solution, thanks for putting it together!

Could you add add a test-case to https://github.com/bazelbuild/rules_rust/tree/main/test/cargo_build_script which shows that this works, so that we don't break it in the future?

cargo/cargo_build_script_runner/bin.rs
freeformstu: There's actually no reason not to symlink in files as well. I did not need it for my use case so I ignored it. I'll get file symlinks added in.

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -53,6 +53,30 @@ fn run_buildrs() -> Result<(), String> {
...
+    for path in exec_root_paths {
+        let path = path
+            .map_err(|err| format!("Failed while getting path from exec root listing: {err:?}"))?
+            .path();
+
+        let file_name = path
+            .file_name()
+            .ok_or_else(|| "Failed while getting file name".to_string())?
+            .to_str()
+            .ok_or_else(|| "Failed while converting file name to string".to_string())?;
```
illicitonion, line 68:
```suggestion
            .ok_or_else(|| "Failed while getting file name".to_string())?;
```

You can use an `OsString` to `Path::join`

test/cargo_build_script/BUILD.bazel
```diff
@@ -66,3 +66,40 @@ rust_test(
...
+###############################################################################
+# Test that the build script can access files in the exec root.
+#
+# All assertions are done in the build script. If it succeeds in execution, the
+# test passes.
+###############################################################################
+
+write_file(
+    name = "root_file",
+    out = "root_file.txt",
```
illicitonion, line 79:
```suggestion
    name = "cargo_manifest_dir_file",
    out = "cargo_manifest_dir_file.txt",
```

My first read of this in the build script assumed "root" meant "root of repo"

test/cargo_build_script/test_exec_root_access.build.rs
```diff
@@ -0,0 +1,62 @@
...
+                .expect("Failed to convert file name to string")
+        })
+        .collect::<HashSet<_>>();
+
+    assert!(
+        root_files.take(build_script_name).is_some(),
+        "Build script must be in the current directory"
+    );
+    assert!(
+        root_files.take("root_file.txt").is_some(),
```
illicitonion, line 39: Let's assert on the contents too

illicitonion (approved):
LGTM, just some minor comments - thanks!

freeformstu:
@illicitonion Apologies for the delay on this PR. I just addressed your comments - I'm still working through some windows test failures in this PR. I'll try to get those addressed by the end of this week.

Do you mind if I squash and rebase?

illicitonion:
Thanks! Feel free to manage the branch however you want (rewrite history or add commits on top), and I'll squash + rebase into main whenever it's done :)

cargo/cargo_build_script_runner/bin.rs
```diff
@@ -53,6 +53,23 @@ fn run_buildrs() -> Result<(), String> {
     create_dir_all(&out_dir_abs)
         .unwrap_or_else(|_| panic!("Failed to make output directory: {:?}", out_dir_abs));
 
+    // Symlink the execroot to the manifest_dir so that we can use relative paths in the arguments.
+    let exec_root_paths = std::fs::read_dir(&exec_root)
```
UebelAndre, line 57: Ok, I think the issue here is that when running outside of a sandbox there will actually be more files in this directory than expected. Instead of reading from the execroot, is there a MANIFEST file or something that tracks all inputs to an action? I forget if that's created for actions or just tests. The idea would be you can link only the things the build script has as an explicitly registered as an input.

UebelAndre (changes requested):
I think the failure here is caused by the Windows not actually having support for sandboxing. I don't have a windows box so it's kinda hard to confirm but reading https://bazel.build/versions/6.1.0/docs/sandboxing it seems to explicitly mention Linux and MacOS but not Windows which, in addition to the failure here, makes me think there's just no sandboxing. This is potentially a good find for the feature as we should make sure it works correctly with and without sandboxing.

cargo/cargo_build_script_runner/bin.rs
UebelAndre: Looks like MANIFEST is only found in runfiles

freeformstu: Thanks @UebelAndre! I can reproduce the issue locally with this diff now:
```diff
diff --git a/cargo/private/cargo_build_script.bzl b/cargo/private/cargo_build_script.bzl
index e567c1c5..3d1f4858 100644
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -242,6 +242,9 @@ def _cargo_build_script_impl(ctx):
         mnemonic = "CargoBuildScriptRun",
         progress_message = "Running Cargo build script {}".format(pkg_name),
         env = env,
+        execution_requirements = {
+            "local": "",
+        }
     )
 
     return [
```

What do you think the behavior should be in this case? Should I just disable all of this logic if you're building outside of a bazel sandbox since that's the whole point of this change?

freeformstu: My most recent change updates the test to only assert that the execroot is empty if the test is running in a sandbox. This resolves almost all of the issues but now I have 3 remaining. One is definitely symlink related and the other two look unrelated but may be

Questions:
1. Would it be acceptable to make this a feature which does not exist when running on Windows so that it can entirely avoid the windows symlinking issues?
2. If that's not acceptable, would it be acceptable to make this feature opt-in (perhaps through an environment variable) so that consumers of this change can toggle it on where they see fit? That would give folks the ability to enable it for windows in the environments where they have symlinks enabled.

UebelAndre: I think making this a feature would be good. It can be disabled by default and enabled globally via `--features` in user's `.bazelrc` files. My experience is doing anything on windows is a bad idea 😅 but we should probably aim to keep the behavior consistent across operating systems.

UebelAndre (changes requested):

freeformstu:
Now I just have ios and mac build issues which look like they're just generally broken (not due to my changes). @UebelAndre any recommendations here?

.bazelci/presubmit.yml
```diff
@@ -27,6 +27,7 @@ default_windows_targets: &default_windows_targets
   - "-//bindgen/..."
   - "-//test/proto/..."
   - "-//test/unit/pipelined_compilation/..."
+  - "-//test/cargo_build_script:test_exec_root_access.build.feature_enabled"
```
UebelAndre, line 30: Can you delete this in favor of a `target_compatible_with` flag on the target should you need it?

cargo/private/cargo_build_script.bzl
```diff
@@ -3,6 +3,7 @@
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "CPP_COMPILE_ACTION_NAME", "C_COMPILE_ACTION_NAME")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load("//cargo:features.bzl", "SYMLINK_EXEC_ROOT_FEATURE", "feature_enabled")
```
UebelAndre, line 6: Why would this need to be a part of the public api vs living in `//cargo/private`?

UebelAndre (changes requested):
Awesome work! Just a few questions 😅

cargo/private/cargo_build_script.bzl
UebelAndre: That sounds reasonable, but could you then move the `feature_enabled` logic into `//cargo/private`? Surely that doesn't seem necessary, right?

freeformstu: Done.

UebelAndre (approved):
Amazing work! Thank you so much!

dk333333:
Hi @freeformstu apologies for the naive question, but how/where do I enable this feature? I had cherry picked an earlier commit but will need to enable it for real when upgrading. I checked the [docs](http://bazelbuild.github.io/rules_rust/flatten.html) but could not find it there

PR #1782, opened 2023-01-17
Support crate_features and proc macros in doctests

bsilver8192:
Previously setting features for the doctests themselves wasn't supported at all, and attempting to use a proc macro would fail. I needed both of these together to write a test.

This is the last of my changes to get rust_doc_test working for all of my crates at work.

### DIFF
--- a/rust/private/rustdoc_test.bzl
+++ b/rust/private/rustdoc_test.bzl
@@ -67,6 +67,10 @@ def _construct_writer_arguments(ctx, test_runner, opt_test_params, action, crate
     root = crate_info.output.root.path
     if not root in roots:
         roots.append(root)
+    for proc_macro_dep in crate_info.proc_macro_deps.to_list():
+        root = proc_macro_dep.crate_info.output.root.path
+        if not root in roots:
+            roots.append(root)
     for dep in crate_info.deps.to_list():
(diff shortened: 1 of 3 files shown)


bsilver8192:
The tests fail on Windows, I think it's because rustdoc_test_writer's escaping doesn't work. Unfortunately I don't know enough Windows shell scripting to fix it. What should I do? I see these options:
* Somebody else fixes the Windows code before merging, either in this PR or a separate one
* File a bug and disable the problematic test rules on Windows
* File a bug, use `select` to avoid setting the `cfg` for Windows, and comment out the problematic tests
* File a bug, use `select` to avoid setting the `cfg` for Windows, and `cfg(not(target_os = "windows"))` t
* Split out the crate_features support for a separate PR, and fork rustdoc_lib.rs into proc_macro and non-proc_macro variants without any `cfg`

bsilver8192:
@UebelAndre I need help on this. I added a test for the escaping (and found+fixed a bug on Linux), but I have no idea how to do this on Windows. Nothing I read online seems to work the way it's claimed, but it's also hard to confirm what's happening without a local machine to experiment interactively. Seems like there's 2 or 3 nested layers of parsing that all have to be satisfied without escape characters that they strip off, and it may be impossible to do robustly.

PR #1787, opened 2023-01-18
Add "crate-name={}" tag to Crate Universe targets

### LINKED ISSUE #1783: Crate Universe: Expose the original crate name via a tag
When interoperating with Cargo, it's useful to have knowledge of the original crate name for a crate downloaded via Crate Universe (e.g: for distributing a crate compiled via Bazel on crates.io).

Unfortunately this name is not currently exposed anywhere, besides (incidentally) via the auto-generated repository name: `@crates__async-trait-0.36.0`.

As a workaround it is possible to retrieve the crate name through string parsing, but this feels like a hack.

alexjpwalker:
When interoperating with Cargo, it's useful to have knowledge of the original crate name for a crate downloaded via Crate Universe (e.g: for distributing a crate compiled via Bazel on crates.io).

Unfortunately this name is not currently exposed anywhere, besides (incidentally) via the auto-generated repository name: `@crates__async-trait-0.36.0` (for the target `@crates__async-trait-0.36.0//:async_trait`) and via the `alias`, `@crates//:async-trait`, which (to the best of my knowledge) is not generally visible to a Bazel aspect, as the `target` field in an aspect implementation refers to the original target being aliased - in this case `@crates__async-trait-0.36.0//:async_trait`

As a workaround it is possible to retrieve the crate name through string parsing, but this feels like a hack. In this PR we add a tag to each Crate Universe target: for example `"crate-name=async-trait"`.

Fixes #1783 .

### DIFF
--- a/crate_universe/src/rendering.rs
+++ b/crate_universe/src/rendering.rs
@@ -514,6 +514,7 @@ impl Renderer {
                 tags.insert("manual".to_owned());
                 tags.insert("noclippy".to_owned());
                 tags.insert("norustfmt".to_owned());
+                tags.insert(format!("crate-name={}", krate.name));
                 tags
             },
             version: krate.common_attrs.version.clone(),


illicitonion:
Thanks for the PR!

> the repo appears to be unusable (ie: when importing the source code artifact via `http_archive`) without going through the release process in GH Actions

See https://github.com/bazelbuild/rules_rust/blob/main/crate_universe/DEVELOPMENT.md for development tips :)

---

We should also add this tag to generated `cargo_build_script`, `rust_proc_macro` and `rust_binary` targets - could you make that be the case, and add a test in rendering.rs?

crate_universe/src/rendering.rs
```diff
@@ -766,6 +768,7 @@ mod test {
 
         assert!(build_file_content.contains("rust_library("));
```
alexjpwalker, line 769: Validate `crate-name` tag correctness for:
1. `rust_library`
2. `cargo_build_script`
3. `rust_proc_macro`
4. `rust_binary`

crate_universe/src/rendering.rs
```diff
@@ -379,6 +379,7 @@ impl Renderer {
                 tags.insert("manual".to_owned());
                 tags.insert("noclippy".to_owned());
                 tags.insert("norustfmt".to_owned());
+                tags.insert(format!("crate-name={}", krate.name));
```
alexjpwalker, line 382: Add tag to `cargo_build_script`s

illicitonion (approved):
LGTM - thanks!

If you rebase past https://github.com/bazelbuild/rules_rust/pull/1785 we should be able to merge :)

alexjpwalker:
Merged latest upstream changes.

PR #1789, opened 2023-01-19
Added compatibility flags to `.bazelrc` to prevent regressions

UebelAndre:


.bazelrc
```diff
@@ -9,9 +20,19 @@ build:rustfmt --output_groups=+rustfmt_checks
...
 build:clippy --output_groups=+clippy_checks
 
-# https://bazel.build/docs/windows#symlink
-startup --windows_enable_symlinks
+###############################################################################
+## Incompatibility flags
+###############################################################################
+
+# https://github.com/bazelbuild/bazel/issues/8195
+build --incompatible_disallow_empty_glob=true
```
UebelAndre, line 28: I'm nervous that changes to `crate_universe` could trigger a regression here so thought I'd always enable it.

illicitonion (approved):

Issue #1681, opened 2022-11-30
Add support for `--nolegacy_external_runfiles`


UebelAndre:
[--nolegacy_external_runfiles](https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles) can be very helpful by improving sandbox performance by reducing the number of runfiles produced by executables. It would be great if rules_rust could fully support this flag.

UebelAndre:
The repo mostly works with this flag but there's some issues with `rust_doc_test` and `crates_vendor` (crate_universe rule). I've opened https://github.com/bazelbuild/rules_rust/pull/1680 to address the `crates_vendor` issue but `rust_doc_test` is a bit more nuanced and requires a deeper look

Issue #1431, opened 2022-07-04
Compile `rust_doc_test` binaries in an action


UebelAndre:
re: https://github.com/bazelbuild/rules_rust/pull/1427#issuecomment-1174049071

I think it might be time to rethink the `rust_doc_test` rules. Right now the test compiles and performs tests as part of the test itself. Instead of doing this, the test binaries that perform doc tests should be compiled in an action and the binaries be the executable for running tests.

UebelAndre:
https://github.com/rust-lang/rust/issues/87022 is something that might enable this behavior.

UebelAndre:
cc @scentini for awareness.

UebelAndre:
Looks like some of the rfc is implemented. It might be worth revisiting this and trying to compile tests in the build action and run them in the test action. This would also cleanup a ton of code and build issues.

PR #1790, opened 2023-01-19
Added support for `--nolegacy_external_runfiles` to `rust_doc_test`

UebelAndre:
This solution is pretty hacky but should get the job done. I think a hacky solution is acceptable though since the way this rule works is already less than desirable. https://github.com/bazelbuild/rules_rust/issues/1431 tracks the ideal implementation.

closes https://github.com/bazelbuild/rules_rust/issues/1681


illicitonion (approved):
Sure

PR #1791, opened 2023-01-20
Prevent crates_vendor from restarting bazel.

matts1:
Previously, crates_vendor had the potential to restart bazel every time it runs, as the bazel startup options are dependent on PATH.


illicitonion (approved):
Today I learned! Thanks!

PR #1792, opened 2023-01-20
Have `--experimental_use_cc_common_link` cover `rust_shared_library`

scentini:
Currently the `--experimental_use_cc_common_link` build setting only covers `rust_binary` and `rust_test`. This PR adds `rust_shared_library` to the list. `rust_proc_macro` is to be covered in a separate PR.


krasimirgg (approved):
Nice! Thank you!

PR #1795, opened 2023-01-23
Fix `cc_common.link` file output name

scentini:
https://github.com/bazelbuild/rules_rust/pull/1792 added `rust_shared_library` to the set of rules that `--experimental_use_cc_common_link` operates on, however it also introduced the following bug:

We declare the output files early in the process, and store them in `CrateInfo.output`. The output file closely matches the `CrateInfo.name`, but in some cases it can slightly differ, eg for a target named `foo-bar`, the crate name will be `foo_bar`, while the output filename stays `foo-bar` (or `libfoo-bar.so` in the case of `cdylib`s).

https://github.com/bazelbuild/rules_rust/pull/1792 added logic to get rid of the `lib` prefix and `.so` suffix before passing the filename to `cc_common.link` (as `cc_common.link` expects), however instead of manipulating the `CrateInfo.output.basename` to remove the `.so` and `.lib`, I used `CrateInfo.name`. As per above, this is not always correct.

This PR also adds unit tests for the `--experimental_use_cc_common_link` build setting.


krasimirgg (approved):
Thank you!

PR #1798, opened 2023-01-25
Add `CARGO_BAZEL_REPIN_ONLY` repinning allowlist

### LINKED ISSUE #1797: Sync with re-pin for single crate index causes splicing for all of them
(Previous Slack conversation: https://bazelbuild.slack.com/archives/CSV56UT0F/p1674498123988129)

Given that I have multiple cargo workspaces that use `crate_universe`, I'd expect that I can re-pin a single index via the command provided in the documentation (`CARGO_BAZEL_REPIN=true bazel sync --only=myworkspace_index`). However, on the console, I can see that it causes not only the requested workspace to be re-pinned, but that it executes the splicing for all of them. Given that we already have 4 cargo workspaces in our monorepo and splicing for each workspace can take 1-2 minutes, this is slowly becoming a blocker for adding more workspaces.

I can reproduce this issue with this repo that I had already set up: https://github.com/hobofan/rules_rust_multiple_workspaces . In there running either `CARGO_BAZEL_REPIN=true bazel sync --only=workspace1_index` or `CARGO_BAZEL_REPIN=true bazel sync --only=workspace2_index` will cause both workspaces to be spliced.

Some further info:
- Bazel 5.3.0
- macOS 12.5.1

Happy to provide any further information!

hobofan:
FIXES #1797 as discussed

I kept it on the Starlark level, as I don't think `cargo-bazel` needs to be aware of the environment variable.

Thanks for the review! :)


UebelAndre (approved):
Looks good to me, thanks!

hobofan:
Wow, thanks for the quick turnaround! :)

PR #1803, opened 2023-01-27
Use correct dynamic link args fro proc-macro crates

konkers:
Some toolchains, notably https://github.com/bazelembedded/rules_cc_toolchain, use different and incompatbile link arguments for executables and shared libraries.


UebelAndre (changes requested):
Thanks! Would you also be able to add some kind of test for this? I'm not sure if there's an easy way to define a cc toolchain with incompatible args that you could transition to and show the right args are used in the right context. 

Maybe @scentini could suggest a way to test this (they're usually really good at coming up with test strategies 😄)?

konkers:
Yeah, I'm at a total loss as to how to test this change.  Any guidance would be greatly appreciated.

UebelAndre (approved):
Ignoring testing though, this change looks good to me 😄

konkers:
Build failure appears to be caused by https://github.com/bazelbuild/continuous-integration/issues/1540.  Unclear form the UI how to re-run that step.

Issue #1445, opened 2022-07-10
Update `rust_toolchain.target_json` from a label to a json string


UebelAndre:
There have been a couple of cases where the functionality added in https://github.com/bazelbuild/rules_rust/pull/836 to teach `rust_toolchain` to handle json files which represent fine grain information about platform triples has caused confusion or a gap in information in "starlark-land" when writing rules. A common issue is that while you can use `target_json`, you still need to provide `os` and in some cases, a "most similar" `target_arch` but that can differ from what `target_json` actually represents. Instead of supporting either a json or explicitly defined values, I think everything should be written to the toolchain. For folks who still want to use a json file to represent their platform triples, I suggest a repository rule be updated or introduced to allow translating the json file into explicit values available in "starlark-land".

UebelAndre:
cc @hlopko

hlopko:
cc @krasimirgg

PR #1810, opened 2023-01-31
Updated `rust_toolchain.target_json` to take encoded json strings

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/1445

Despite the interface change, users are still able to have Json files committed to their repo that can feed into toolchains. `test/unit/toolchain/toolchain_test_utils.bzl` in this PR contains an example repository rule which can convert target Json files into Starlark that can be passed to toolchain definitions.


krasimirgg:
Sorry, I somehow missed commenting on the feature request.
We have a use case where the target json file is potentially generated by some other action (and its contents potentially vary by configuration). I'd like it if we still could represent this in rules_rust somehow...
Somehow having a json-encoded string attribute that then internally we turn into a file to feed into the rustc actions feels weird in case we already have the json file lying around. I agree that there's overlap between some of the keys from the json file and some toolchain attributes, what if we had an action that checked those for consistency instead?

UebelAndre:
Can you expand on what action would be generating this json file? I’d run into the same and found this lead to multiple sources of truth which lead to issues. The ultimate goal of this feature for me is to have a Bazel platform for each target these files would represent. Housing data in a json file seems to block this

krasimirgg:
>Can you expand on what action would be generating this json file?
Hypothetical, let's say a `genrule` generates the json file (and there would be corresponding manually-defined platform coming along with it.) How does using a file interfere with having a Bazel platform for each target these represent?

krasimirgg:
How about: have a custom "utils" rule like `rust_toolchain_target_json_from_string` that takes an json-encoded string attribute and generates a json file. Then we can feed that into the target_json label of the generated rust_toolchain-s by the repository macros and still have the ability to pass the json directly or generate it some other way in some cases.

UebelAndre:
> How about: have a custom "utils" rule like `rust_toolchain_target_json_from_string` that takes an json-encoded string attribute and generates a json file. Then we can feed that into the target_json label of the generated rust_toolchain-s by the repository macros and still have the ability to pass the json directly or generate it some other way in some cases.

I would rather there just be 1 way to do handle this. I'm still not sure I understand the value in keeping `target_json` as a label maybe other than it's a breaking change for some folks. Why wouldn't or couldn't information about the platform live in starlark?

krasimirgg:
I also chatted a bit with @scentini about that and now I think what you propose here is OK for the reasons you laid out: this json is super platform specific, it would be useful to have the information in starlark, and bazel json support seems good enough to be able to encode this.

krasimirgg (approved):
Looks good! Thank you.

Issue #1802, opened 2023-01-27
Update protobuf crate to 2.18.2


giokara:
I would like to use `rules_rust_library` and `rules_rust_binary` to build Rust applications with Bazel.

I'm already using [rules-proto-grpc](https://github.com/rules-proto-grpc/rules_proto_grpc) to compile common proto files over different languages (C++, Python). As such, I would like to also use `rust_proto_library` from that repository to build the proto library, instead of the same rule from this repository.

[rules-proto-grpc](https://github.com/rules-proto-grpc/rules_proto_grpc) is already on protobuf crate version 2.18.2. In order to avoid incompatibilities between the crate version, it would be helpful to update this repository's dependencies. I have already updated the repo in a fork https://github.com/giokara-oqton/rules_rust and can open a PR with my changes.

UebelAndre:
I'd be happy to review a PR if someone were to put one up 😄

wilsonwang371:
> I would like to use `rules_rust_library` and `rules_rust_binary` to build Rust applications with Bazel.
> 
> I'm already using [rules-proto-grpc](https://github.com/rules-proto-grpc/rules_proto_grpc) to compile common proto files over different languages (C++, Python). As such, I would like to also use `rust_proto_library` from that repository to build the proto library, instead of the same rule from this repository.
> 
> [rules-proto-grpc](https://github.com/rules-proto-grpc/rules_proto_grpc) is already on protobuf crate version 2.18.2. In order to avoid incompatibilities between the crate version, it would be helpful to update this repository's dependencies. I have already updated the repo in a fork https://github.com/giokara/rules_rust and can open a PR with my changes.

Do you have a contact email? I want to discuss with you some of the details in your branch.

wilsonwang371:
> Sure, however not sure how to share without making publicly available. Any suggestions are welcome 🙂 About the PR, I'd like to update to current version (v3.2.0) instead of reported v2.18.2 but haven't found more time to work on it.

Are you able to rebase to the latest release and add support for 3.2.0? Btw, there might be some further dependencies needed such as protobuf-support.

Are you on the slack channel? Or you can send me an email so that I can get in touch with you. My email is wilson.wang[at]bytedance.com

PR #1812, opened 2023-01-31
Update protobuf and bytes crates

giokara:
Update the protobuf crate version (and, because required, the bytes crate). Closes https://github.com/bazelbuild/rules_rust/issues/1802.


proto/3rdparty/crates/BUILD.grpc-compiler-0.6.2.bazel
```diff
@@ -72,7 +72,7 @@ rust_binary(
     version = "0.6.2",
     deps = [
         ":grpc_compiler",
-        "@rules_rust_proto__protobuf-2.8.2//:protobuf",
-        "@rules_rust_proto__protobuf-codegen-2.8.2//:protobuf_codegen",
+            "@rules_rust_proto__protobuf-2.18.2//:protobuf",
+            "@rules_rust_proto__protobuf-codegen-2.18.2//:protobuf_codegen",
     ],
```
dtolnay, line 77: This is not the indentation that would be created by the `:crates_vendor` BUILD file generator.

I see _"missing merge resolve"_ (7e363513bf98cad7a149046afc816e08564a1b02) in the commit history of this PR so I infer this result is from a badly resolved merge conflict that you edited manually, not what `:crates_vendor` actually generated.

It's almost never correct to try manually resolving merges inside generated code. You just regenerate the generated file.

Also from what's visible in that merge, you originally created this update against a very old commit of this repo (at least 4 weeks old), before later rebasing it at least once and then doing at least 2 merges between the upstream branch and your branch. That is not going to work; there have been other crates updated by other contributors in the meantime. You can tell because the Cargo.Bazel.lock in this PR is invalid and makes `bazel run //proto/3rdparty:crates_vendor` fail,  whereas it's supposed to work and it works currently on `main`.

```console
error: failed to select a version for `protobuf`.
    ... required by package `grpc v0.6.2`
    ... which satisfies dependency `grpc = "^0.6.2"` (locked to 0.6.2) of package `direct-cargo-bazel-deps v0.0.1 (/tmp/.tmpgVYSMl)`
versions that meet the requirements `>=2.8.0, <2.10.0` are: 2.8.2, 2.8.1, 2.8.0

all possible versions conflict with previously selected packages.

  previously selected package `protobuf v2.18.2`
    ... which satisfies dependency `protobuf = "^2.18.2"` (locked to 2.18.2) of package `direct-cargo-bazel-deps v0.0.1 (/tmp/.tmpgVYSMl)`

failed to select a version for `protobuf` which could resolve this conflict
```

dtolnay (changes requested):
I recommend recreating this PR from scratch from current master. I see what seems like at least 1 too many incorrectly manually resolved merge conflicts in here that are going to cause problems.

proto/3rdparty/crates/BUILD.grpc-compiler-0.6.2.bazel
giokara: I will make the changes from scratch as suggested. I wasn't aware that these files were auto-generated - will use the mentioned bazel target for the update.

PR #1813, opened 2023-02-01
Use env method recently added to cargo_metadata

illicitonion:
Before we were re-implementing `exec`'s internals because we needed to set an env var. This is cleaner.


UebelAndre (approved):
Nice! Thanks!

PR #1815, opened 2023-02-01
Encapsulate running cargo into a struct

illicitonion:
I'm about to add conditionally setting environment variables depending on the cargo version in-use, so it's useful to route all calls to cargo through controlled methods.

### DIFF
--- a/crate_universe/src/metadata.rs
+++ b/crate_universe/src/metadata.rs
@@ -10,7 +10,9 @@ use std::io::BufRead;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
+use std::sync::{Arc, Mutex};
 
+use crate::lockfile::Digest;
 use anyhow::{anyhow, bail, Context, Result};
 use cargo_lock::Lockfile as CargoLockfile;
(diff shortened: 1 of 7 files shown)


crate_universe/src/metadata.rs
```diff
@@ -80,6 +85,44 @@ impl MetadataGenerator for Generator {
     }
 }
 
+#[derive(Clone)]
+pub struct Cargo {
+    path: PathBuf,
+    full_version: Arc<Mutex<Option<String>>>,
```
UebelAndre, line 91: Why not just `Option<String>`?

UebelAndre (changes requested):
A nit and a question but otherwise seems good 😄

crate_universe/src/metadata.rs
illicitonion: To do that, we'd need every method which takes a `Cargo` to take a mutable reference to it, which would require some futzing around. Instead, we can just pass around shared references, at the cost of acquiring a lock to check whether we've already looked up the full version.

UebelAndre (approved):
Looks good to me!

PR #1817, opened 2023-02-01
Update protobuf crate

giokara:
Update the protobuf crate version (and, because required, the bytes crate). Closes https://github.com/bazelbuild/rules_rust/issues/1802.
Replaces https://github.com/bazelbuild/rules_rust/pull/1812.


UebelAndre:
I haven't looked closely at the errors but noticed an import error and am now wondering if https://github.com/bazelbuild/rules_rust/pull/1617 is related.

giokara:
After looking a bit closer, my first expectation is some 2015/2018 edition incompatibility as described here https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html. Although it's not clear to me how the changes in this PR could lead to this.

UebelAndre:
Closing due to https://github.com/bazelbuild/rules_rust/issues/3858

PR #1819, opened 2023-02-03
Fixes resolver issue with root packages and another dependency format

### LINKED ISSUE #1818: resolver = "2" `cargo tree` doesn't output all deps for workspaces with root packages
I've been trying out the changes in #1710 and I think I've found a minor issue that causes dependencies to be missing.

# Issue 1: Root packages

I believe this is happens when you have a workspace with a [root package](https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package). In this case, `cargo tree` only outputs the dependencies of the root package, not the entire workspace. There is a [--workspace](https://doc.rust-lang.org/cargo/commands/cargo-tree.html#package-selection) flag that configures this output.

# Issue 2: Proc macros

Also, another minor issue is that there is an unhandled `cargo tree` output format that looks like: 

`serde_derive v1.0.152 (proc-macro) (*)`

Dig-Doug:
Fixes #1818


crate_universe/src/metadata.rs
```diff
@@ -441,9 +442,9 @@ where
         // We expect the crate id (parts[1]) to be either
         // "<crate name> v<crate version>" or
         // "<crate name> v<crate version> (<path>)"
-        // https://github.com/rust-lang/cargo/blob/7fb01c68c190c28a43581ae7c719dcf057ecc002/src/cargo/core/package_id.rs#L210-L220
+        // "<crate name> v<crate version> (proc-macro) (<path>)"
```
UebelAndre, line 445: is https://github.com/rust-lang/cargo/blob/7fb01c68c190c28a43581ae7c719dcf057ecc002/src/cargo/core/package_id.rs#L210-L220 still not useful information?

UebelAndre (changes requested):
Thanks! Should the integration test also be updated? I'm not sure the `--workspace` flag is covered there.

crate_universe/src/metadata.rs
Dig-Doug: Added link to the cargo tree code

Dig-Doug:
Added an integration test

UebelAndre (approved):
Thank you so much!

PR #1824, opened 2023-02-07
Added update known shas to include T1-T2 triples

UebelAndre:
This change adds known sha256 entries for all T1 and T2 platform triples for Rust 1.67.0. While not all platforms are supported in the current version of `rules_rust` (or even Bazel), including them can skip a performance regression for users using these unique platforms in their own projects. This change is intended to avoid situations where a switch to Bazel or successfully cross-compiling doesn't feel unnecessarily slow as users wait for Bazel to re-download toolchains (related to https://github.com/bazelbuild/rules_rust/issues/1825).


UebelAndre:
I'm not sure if this is a performance regression. Does anyone have a sense for how big `known_shas.bzl` can get before it becomes problematic?

illicitonion (approved):
Let's see how this goes!

Issue #1711, opened 2022-12-20
Feature request: Add something like per_file_copt to allow customizing rustc flags per module


csmulhern:
For C/C++, [per_file_copt](https://bazel.build/reference/command-line-reference#flag--per_file_copt) allows users to use a regex to apply compiler options on a per-file basis. Common usages include e.g. disabling warnings for third party code, or optimizing certain dependencies even when using fastbuild.

I see rules_rust has `--@rules_rust//:extra_rustc_flags` to allow passing custom flags to rustc. It would be great to have another rule (e.g. `per_module_rustc_flags`) that uses a regex syntax similar to `per_file_copt`, that would allow customization of rustc flags at the module level. The regex could be matched against the crate root path.

For example, `--@rules_rust//:per_module_rustc_flags=^external/.*@-Awarnings` could be used to allow warnings for rust modules in external workspaces.

Anki recently migrated off of Bazel, and cites being able to optimize a subset of their dependencies as a benefit of migrating (see: https://github.com/ankitects/anki/pull/2202).

csmulhern:
@UebelAndre how do you feel about making my current experimental implementation non-experimental?

csmulhern:
Gentle ping on this.

csmulhern:
@UebelAndre we've had experimental_per_crate_rustc_flag for 3 years. There's still no regex support in Starlark, but I still think it's highly useful as-is. How do you feel about stabilizing?

UebelAndre:
We can promote it, I got no issue with that! Do you wanna put up a PR for that?

Also, sorry for the delay 😅

csmulhern:
No worries! I've submitted a PR for stabilization; #4182.

PR #1827, opened 2023-02-08
Adds per_crate_rustc_flag build setting.

csmulhern:
Requested in #1711.

I'm putting up this initial PR to get early feedback, and to gauge the appetite for adding a feature of this sort, before investing further.

In my view, this feature would behave identically to `--per-file-copt`, but mapped to Rust. A good description of `--per-file-copt` is available here: https://bazel.build/docs/user-manual#per-file-copt.

A short TL;DR is that `--per-file-copt` takes values of the format `[+-]regex[,[+-]regex]...@option[,option]...`. The regex(es) is matched against both the label of the target the source belongs to, and the execution path of the source file. If either of these match (the default or if prefixed with `+`), or do not match (if prefixed with `-`), the option(s) is added to the set of compilation flags.

I'm proposing the following behavior changes to map this behavioral model to Rust:

1) Flags would be applied to crates instead of files, as they are the unit of compilation in Rust.
2) The execution path used for the filter would be the path to the crate root.

There are additional factors that make conformance with `--per-file-copt` difficult. Namely, Starlark doesn't support regex matching today. As such, my proposal is to have per_crate_rustc_flag support prefix matching instead, until the date that regex matching comes to Starlark. Regex matching is likely to come at some point, as the C++ rules are being migrated to Starlark, and so `--per-file-copt` will need to be implemented in Starlark.

The current implementation does not support negative matching (`-` above), though that can be added.

The current implementation also only supports a single regex and a single option, though that bit is addressable (and `--per_crate_rustc_flag` can also be used multiple times, so this isn't a real limitation; IMO support for multiple regexes / flags in a single value is more motivated by the desire to mirror `--per-file-copt` than anything else).

A motivating use case for this change is to enable compilation of external crates in fastbuild configurations. With the current implementation, that would look like:
```
bazel build --@rules_rust//:per_crate_rustc_flag="external/@--codegen=opt-level=3" //a:target
```

In the world where regex is supported, this would instead look like:

```
bazel build --@rules_rust//:per_crate_rustc_flag="^external/.*@--codegen=opt-level=3" //a:target
```

It's worth thinking about future changes to this API if there is strong intent to later support regex matching instead of this more limited API. A few options:

1) We can just be comfortable making a breaking change later on.

2) We can name the current implementation something like `experimental_per_crate_rustc_flag` to flag that the API is unstable. When regexes are supported in Starlark, the implementation can be updated, and the `experimental_` prefix can be dropped.

2) We can force prefixes to start with `^` and end with `.*`. Any patterns that included unescaped regex characters or do not start with `^` and end with `.*` could be rejected with a `fail`. This would allow us to support more patterns in the future with proper regex matching, but would make that a backwards compatible change.

Anyway, please let me know what the appetite is to land a feature of this sort, and if there is appetite, what path forward you'd be most comfortable with.

Cheers.


csmulhern:
Any feedback here?

rust/private/rustc.bzl
```diff
@@ -63,6 +63,11 @@ ExtraExecRustcFlagsInfo = provider(
     fields = {"extra_exec_rustc_flags": "List[string] Extra flags to pass to rustc in exec configuration"},
 )
 
+PerCrateRustcFlagsInfo = provider(
+    doc = "Pass each value as an additional flag to non-exec rustc invocations for crates matching the provided filter",
+    fields = {"per_crate_rustc_flags": "List[string] Extra flags to pass to rustc in non-exec configuration"},
```
UebelAndre, line 68: nit, this could probably be shortened to `flags`

rust/private/rustc.bzl
```diff
@@ -1845,6 +1854,32 @@ def _get_dirname(file):
     """
     return file.dirname
 
+def _add_per_crate_rustc_flags(ctx, args, crate_info, per_crate_rustc_flags):
```
UebelAndre, line 1857: Could you write a unit test for this? The function will probably need to be updated to return a list of args instead populating an args reference but to callers I think this would be clearer.

rust/private/rustc.bzl
```diff
@@ -1923,3 +1958,15 @@ extra_exec_rustc_flag = rule(
     implementation = _extra_exec_rustc_flag_impl,
     build_setting = config.string(flag = True, allow_multiple = True),
 )
+
+def _per_crate_rustc_flag_impl(ctx):
+    return PerCrateRustcFlagsInfo(per_crate_rustc_flags = [f for f in ctx.build_setting_value if f != ""])
+
+per_crate_rustc_flag = rule(
+    doc = (
```
UebelAndre, line 1966: Can you update these docs to describe the matching behavior?

UebelAndre (changes requested):
I think this sounds reasonable.

> It's worth thinking about future changes to this API if there is strong intent to later support regex matching instead of this more limited API. A few options:
> 
>     1. We can just be comfortable making a breaking change later on.
> 
>     2. We can name the current implementation something like `experimental_per_crate_rustc_flag` to flag that the API is unstable. When regexes are supported in Starlark, the implementation can be updated, and the `experimental_` prefix can be dropped.
> 
>     3. We can force prefixes to start with `^` and end with `.*`. Any patterns that included unescaped regex characters or do not start with `^` and end with `.*` could be rejected with a `fail`. This would allow us to support more patterns in the future with proper regex matching, but would make that a backwards compatible change.

I think the way to make breaking changes in the future is with experimental or incompatibility flags similar to those found here: https://github.com/bazelbuild/rules_rust/blob/0.18.0/rust/settings/BUILD.bazel

As long as the parsing behavior is documented and tested we can make the appropriate changes as new functionality is made available.

Do you know if there's a feature request somewhere for regex support in Starlark?

rust/private/rustc.bzl
csmulhern: Agreed, but I chose this for consistency with the other providers defined in this file, which use this naming convention. What do you think about being the change we want to see in the world vs stylistic consistency?

rust/private/rustc.bzl
csmulhern: I went ahead and added unit tests for the entire rule.

rust/private/rustc.bzl
csmulhern: Done.

csmulhern:
I went ahead and added the `experimental_` prefix to the public flag. I also went ahead and added unit tests for the existing prefixing matching behavior.

> Do you know if there's a feature request somewhere for regex support in Starlark?

I saw some chatter about it in an issue in the Bazel issue tracker, but am having trouble finding it now unfortunately.

csmulhern:
@UebelAndre any thoughts on the latest changes?

UebelAndre (changes requested):
Looking good! Just one question

BUILD.bazel
```diff
@@ -76,6 +76,12 @@ extra_exec_rustc_flag(
     visibility = ["//visibility:public"],
 )
 
+per_crate_rustc_flag(
+    name = "experimental_per_crate_rustc_flag",
```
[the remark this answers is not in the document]
csmulhern: Wasn't the suggestion to go per_crate_rustc_flag -> experimental_per_crate_rustc_flag?

I use per_crate_rustc_flag for all the internal usages, but experimental_ for the public usages. I can change all the internal usages to experimental_ as well if that is confusing.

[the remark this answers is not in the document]
UebelAndre: I think the way you have it is fine

csmulhern:
I fixed the buildifier issues.

UebelAndre (approved):
Thanks!

PR #1832, opened 2023-02-10
Add `render_config` attribute to `crates_vendor`.

UebelAndre:
`crates_vendor` did not allow users to customize the render config. This is likely because the rule takes control of the majority of the values, but there are still some that are useful to set. The change here introduces a `render_config` attribute to `crates_vendor` which is identical to `crates_repository` but will fail if values controlled by `crates_vendor` are set.


illicitonion (approved):
LGTM - thanks!

PR #1835, opened 2023-02-13
fix: use target_triple struct instead of string

wcalandro:
Kythe is currently updating to a newer rules_rust and we are failing to build because we use `dev_components = True` and currently the string target triple is being passed instead of the struct.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -325,7 +325,7 @@ def _rust_toolchain_tools_repository_impl(ctx):
 
     # Not all target triples are expected to have dev components
     if ctx.attr.dev_components:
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Good catch! Would you be able to add a regression test similar to https://github.com/bazelbuild/rules_rust/blob/0.18.0/examples/WORKSPACE.bazel#L53-L70 ?

This way this block of code is regression tested?

wcalandro:
I attempted to add another set but I when I revert the bug fix the build does not fail because bazel doesn't attempt to download the toolchain. Do you have any suggestions?

PR #1836, opened 2023-02-13
Add CARGO_MANIFEST_DIR to rust-project.json

titanous:
This fixes rust-analyzer's handling of proc macros that expect `CARGO_MANIFEST_DIR`.


UebelAndre:
Can you link an example of this? And the error you’d see?

titanous:
Something like this (untested):

`main.rs`
```rust
use askama::Template;

#[derive(Template)]
#[template(path = "example.html.j2")]
struct ExampleTemplate {}
```

`BUILD.bazel`
```starlark
rust_binary(
    name = "backend",
    srcs = [
        "main.rs",
    ],
    compile_data = ["example.html.j2"],
    deps = [
        "@crate_index//:askama",
    ],
)
```

`example.html.j2` needs to exist but can be empty.

The error will show up as a panic in the rust-analyzer language server output (available from the output tab in VS Code).

PR #1842, opened 2023-02-20
Crate universe

matts1:
Add support for crate universe with bzlmod to rules_rust.

Please ignore all commits by @cameron-martin as they're being reviewed seperately in github.com/bazelbuild/rules_rust/pull/1528.

I'll change this from a draft PR into a real PR once his PR is merged.


matts1:
@gburgessiv FYI

bzlmod/private/toolchains/toolchains.bzl
```diff
@@ -0,0 +1,116 @@
...
+                rustfmt_version = toolchain.rustfmt_version,
+                edition = toolchain.edition,
+                sha256s = toolchain.sha256s,
+                urls = toolchain.urls,
+                dev_components = False,
+                auth = None,
+            )
+
+        for toolchain in mod.tags.toolchain:
+            repo_name = "%s_rust_toolchains" % mod.name
```
cameron-martin, line 57: It seems like you're making it so that the whole module graph can create and register its own toolchains, but isn't the set of available toolchains a global thing? That's why I originally made it so that only the top-level module can define toolchains, but maybe I'm misunderstanding something here.

matts1: To be honest, I'm not particularly familiar with toolchains, so you could definitely be right. I was making the assumption that each module registered its own toolchains, and that, like use_repo, each module had its own visibility into the set of toolchains, and specifically used  ones that they defined themselves?

Eg. I saw that your toolchain definition has the tag "edition". Since editions are specifically not backwards compatible, something designed for a 2018 edition might fail if compiled with a toolchain that specifies the 2021 edition.

@Wyverald Could you confirm how register_toolchain works? Is it global or per-module?

matts1:
Superceded by https://github.com/bazelbuild/rules_rust/pull/1910

PR #1843, opened 2023-02-20
Use `_make_link_flags_darwin` when target os is `ios`.

thb-sb:
Right now, `_make_link_flags_default` is used when the target os is `ios`. However, it makes use of `--whole-archive` and `--no-whole-archive` flags, which are not supported by `dyld`.

As it is done for the `macos` target, we should use `-force_load` instead. `_make_link_flags_darwin` does that, therefore we should use it when linking for iOS.


illicitonion (approved):
Seems reasonable - thanks!

illicitonion:
Thanks so much!

Issue #1839, opened 2023-02-16
Refactor itegration tests out of examples


UebelAndre:
The examples directory is currently a mix of examples and integration tests. The directory should be refactored so there can be well documented examples for users to learn from that are separate from complicated regression tests.

UebelAndre:
I'm going to begin refactoring some examples while leaning more heavily toward deleting ones I don't understand. If anyone strongly feels some particular example is high-value, please let me know or it risks being deleted in the event it's value is not obvious to me.

PR #1844, opened 2023-02-21
Fix tests for new Apple toolchain

keith:
The Apple toolchain that is used to build iOS targets is moving out of bazel and into the apple_support repo. This switches tests to that toolchain so that this repo continues to work once it is removed from bazel.

https://github.com/bazelbuild/bazel/pull/16619

### DIFF
--- a/examples/ios/.bazelrc
+++ b/examples/ios/.bazelrc
@@ -1,0 +1,4 @@
+# Enable CC toolchain that supports iOS from https://github.com/bazelbuild/apple_support
+build --apple_crosstool_top=@local_config_apple_cc//:toolchain
+build --crosstool_top=@local_config_apple_cc//:toolchain
+build --host_crosstool_top=@local_config_apple_cc//:toolchain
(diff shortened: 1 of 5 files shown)


examples/ios/.bazelrc
```diff
@@ -0,0 +1,3 @@
+build --apple_crosstool_top=@local_config_apple_cc//:toolchain
```
UebelAndre, line 1: Is there any documentation on why these are needed? Perhaps some github issue that refers to these?

UebelAndre (approved):
Thanks! 

Just had one question.

examples/ios/.bazelrc
keith: the setup doc talks about the general usage https://github.com/bazelbuild/apple_support/

but the gist is until the incompatible flag for CC toolchain resolution ships this is how you have to specify anything besides the default cc toolchain. in the future once that flag is flipped (and the behavior is ironed out in general) this entire file can be tossed because the setup will happen in the WORKSPACE and that will be enough

UebelAndre: Could you add that link with this context there? I'm planning on working on https://github.com/bazelbuild/rules_rust/issues/1839 and think it'd be nice to have the ios examples act as examples but to do that they need to be well documented. Every little bit helps.

keith: done

PR #1845, opened 2023-02-22
Fix test assertion for arm64 macs

keith:
On M1 macs the bazel-out darwin path is `darwin_arm64-fastbuild`. This just takes whatever is there and allows it just assuming bazel is doing the right thing.


UebelAndre (approved):
Thank you!

PR #1847, opened 2023-02-24
Support repo mapping for bazel runfiles.

matts1:
With bazel 6 + bzlmod, bazel creates a "repo mapping" file. Note that this can be a problem even if rules_rust doesn't support bzlmod, as it can occur if a bzlmod repo includes rules_rust in its WORKSPACE.bazel file.

Discussion detailed at https://groups.google.com/g/bazel-discuss/c/DsVivJhU7Bw


matts1:
@scentini What do I need to do to get this reviewed?

scentini:
Hi @matts1 unfortunately I have not yet found the time to review the PR; however there seem to be some failures on CI, could you please take care of those?

matts1:
I fixed the clippy errors that were causing the failures, but there still appears to be two failures which I believe are unrelated to my code.

tools/runfiles/runfiles.rs
```diff
@@ -100,10 +130,39 @@ impl Runfiles {
...
         }
+
+        let path = self
+            .repo_mapping
+            .as_ref()
+            .and_then(|mapping| {
+                let mut components = path.components();
+                if let Some(std::path::Component::Normal(target_local)) = components.next() {
+                    if let Some(target_local) = target_local.to_str() {
+                        let current_canonical = self.current_repository();
```
scentini, line 141: We're looking at this PR with @Wyverald, and all seems fine apart from this line; we concluded that the implementation of `current_repository()` is currently incorrect. We'll first need to tackle that, because it causes the next couple of lines to be wrong as well. I'll file a separate issue to describe the problem.

matts1: @scentini Has `current_repository` been fixed? If not, could you link to the bug that was filed so I can know when it is fixed.

dzbarsky:
https://github.com/bazelbuild/rules_rust/pull/2566 is the latest attempt

cerisier:
Should this be closed ?

dzbarsky:
yes, this was fixed in https://github.com/bazelbuild/rules_rust/pull/2566

Issue #1814, opened 2023-02-01
macOS CI failing with "xcodebuild: error: SDK "macosx13.1" cannot be located."


illicitonion:
Examples:
https://buildkite.com/bazel/rules-rust-rustlang/builds/8026#01860c93-81ea-4355-bbe0-ed32c053b65c
https://buildkite.com/bazel/rules-rust-rustlang/builds/8031#01860dc1-d38c-47fd-8b46-715a854005bb

Seems to have started happening today for unrelated changes.

UebelAndre:
Maybe https://github.com/bazelbuild/continuous-integration/issues/1540 is related?

UebelAndre:
I've opened a PR to temporarily disable the failing job here: https://github.com/bazelbuild/rules_rust/pull/1816

PR #1856, opened 2023-03-03
Support `[patch]` in crate_universe when using multiple `Cargo.toml`s

UebelAndre:
This change is intended to address the current issues in CI but requires some additional changes to be made.

- https://github.com/bazelbuild/rules_rust/issues/1858

Note that each commit is a unique change but they're all required to fix CI.


.bazelci/presubmit.yml
```diff
@@ -440,17 +440,16 @@ tasks:
...
-  #   name: Crate Universe Examples
-  #   platform: macos
-  #   working_directory: examples/crate_universe
-  #   run_targets: *crate_universe_vendor_example_targets
-  #   build_targets:
-  #     - "//..."
-  #   test_targets:
-  #     - "//..."
-  #   build_flags: *aspects_flags
+  crate_universe_examples_macos:
```
UebelAndre, line 443: By replacing the use of `rules_foreign_cc` the issue mentioned in https://github.com/bazelbuild/rules_rust/issues/1814 is avoided

UebelAndre:
> I'm a bit torn on this one - while I strongly prefer rustls to openssl, I have found it valuable in the past to point people to this example as an example of a complex C++-using crate... What's the motivation behind the replacement? Just working around the bazelci issue?

The use of `rules_foreign_cc` has lead to issues and is not something I want to weigh down these tests. If folks want an example for how to build a complicated C++-using crate I would suggest making one in the root of `./examples`. I think we have sufficient examples of this though unless you would like an example for specifically `openssl`

cargo/private/cargo_build_script_wrapper.bzl
```diff
@@ -119,6 +119,9 @@ def cargo_build_script(
         build_script_kwargs["compatible_with"] = kwargs["compatible_with"]
         binary_kwargs.pop("compatible_with")
 
+    if "toolchains" in kwargs:
```
UebelAndre, line 122: This addresses https://github.com/bazelbuild/rules_rust/issues/1859

crate_universe/src/splicing/splicer.rs
```diff
@@ -296,6 +296,12 @@ impl<'a> SplicerKind<'a> {
         let installations =
             Self::inject_workspace_members(&mut manifest, manifests, workspace_dir)?;
 
+        // Collect all patches from the manifests provided
```
UebelAndre, line 299: This addresses https://github.com/bazelbuild/rules_rust/issues/1858

examples/crate_universe/vendor_local_manifests/BUILD.bazel
```diff
@@ -14,12 +14,22 @@ rust_library(
     aliases = aliases(),
     edition = "2018",
     proc_macro_deps = all_crate_deps(proc_macro = True),
+    # TODO: https://github.com/bazelbuild/rules_rust/issues/1860
```
UebelAndre, line 17: https://github.com/bazelbuild/rules_rust/issues/1860

crate_universe/src/splicing/splicer.rs
```diff
@@ -471,6 +484,34 @@ impl<'a> SplicerKind<'a> {
...
+    fn inject_patches(manifest: &mut Manifest, patches: &cargo_toml::PatchSet) -> Result<()> {
+        for (registry, new_patches) in patches.iter() {
+            // If there is an existing patch entry it will need to be merged
+            if let Some(existing_patches) = manifest.patch.get_mut(registry) {
+                // Error out if there are duplicate patches
+                existing_patches.extend(
+                    new_patches
+                        .iter()
+                        .map(|(pkg, info)| {
+                            if existing_patches.contains_key(pkg) {
```
illicitonion, line 497: It'd be nice to not error if they patches are identical

illicitonion (approved):
Makes sense - thanks!

crate_universe/src/splicing/splicer.rs
UebelAndre: Good idea, done!

PR #1857, opened 2023-03-03
Support sparse indexes

### LINKED ISSUE #1854: Usage of sparse registries not working
https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html introduced usage of a sparse registry.
The [feature is stablized](https://github.com/rust-lang/cargo/issues/9069) and will become available in the stable channel in a few days:

> Sparse registries are now stabilized and will be available in Rust 1.68, which will be released on 2023-03-09

[crate-index](https://github.com/bazelbuild/rules_rust/blob/main/crate_universe/3rdparty/crates/defs.bzl#L301) is used [here](https://github.com/bazelbuild/rules_rust/blob/main/crate_universe/src/splicing.rs#L320-L321) to parse the config that would hold the configuration:

```toml
[registries.artifactory]
index = "sparse+https://artifactory.company.my/artifactory/api/cargo/external-index-crates-io-remote/index/"
[source.artifactory]
registry = "sparse+https://artifactory.company.my/artifactory/api/cargo/external-index-crates-io-remote/index/"
```

However this fails to parse with:
```
Error: Failed to locate crate indexes

Caused by:
    0: Failed to load index for url: sparse+https://artifactory.company.my/artifactory/api/cargo/external-index-crates-io-remote/index/
    1: 'sparse+https://artifactory.company.my/artifactory/api/cargo/external-index-crates-io-remote/index/' is not a valid registry url
```

There is already an issue i could find since the feature is not yet supported by [rust-crates-index](https://github.com/frewsxcv/rust-crates-index): 
https://github.com/frewsxcv/rust-crates-index/issues/87

illicitonion:
And explicitly opt in to them for cargos which support it.

This should very significantly speed up repinning, as only the metadata for crates needed for the build, rather than the whole index, will need fetching.

Fixes #1854


crate_universe/src/splicing.rs
```diff
@@ -294,80 +296,83 @@ impl WorkspaceMetadata {
...
                         } else {
                             &url
                         }
                     } else {
                         &url
+                    }
+                } else {
+                    &url
+                };
+                let index = if cargo.use_sparse_registries_for_crates_io()? && index_url == "https://github.com/rust-lang/crates.io-index" {
```
UebelAndre, line 317: Perhaps this string could live as a global somewhere and also be used here? https://github.com/bazelbuild/rules_rust/blob/0.18.0/crate_universe/src/splicing/cargo_config.rs#L35-L38

crate_universe/src/metadata.rs
```diff
@@ -124,6 +130,29 @@ impl Cargo {
         }
         Ok(full_version.clone().unwrap())
     }
+
+    pub fn use_sparse_registries_for_crates_io(&self) -> Result<bool> {
+        let full_version = self.full_version()?;
+        let version_str = full_version.split(' ').nth(1);
+        if let Some(version_str) = version_str {
+            let version = Version::parse(version_str).context("Failed to parse cargo version")?;
```
UebelAndre, line 138: Do you know if nightly releases return a clean semver value?

UebelAndre (approved):
Looks good to me!

A few questions and is there any kind of unit testing you think would be effective here? I suppose by updating Rust to 1.68.0 here you'll have coverage for this code and coverage for older versions in the min version test so this seems regression sufficient.

crate_universe/src/splicing/cargo_config.rs
```diff
@@ -122,7 +122,13 @@ impl CargoConfig {
 
     /// Look up a registry [Source] by its url.
     pub fn get_source_from_url(&self, url: &str) -> Option<&Source> {
-        self.source.values().find(|v| v.registry == url)
+        if let Some(found) = self.source.values().find(|v| v.registry == url) {
```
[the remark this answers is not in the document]
illicitonion: This is actually a fix for the replace_with logic... Added some tests for it, too :) Still not actually sure what the issue in #1866 was, though...

crate_universe/src/splicing.rs
illicitonion: Good call! Done!

crate_universe/src/metadata.rs
illicitonion: They do, e.g. 1.69.0 parses as:
`Version { major: 1, minor: 69, patch: 0, pre: Prerelease("nightly") }`

Now, technically that means logic is _slightly_ buggy, because 1.68.0-nightly is not >= 1.68.0 - changed the logic to just compare the major and minor version.

illicitonion:
I added some unit tests for a few different bits of logic - an integration test will hopefully still come at some point, but we now actually have a bit of coverage, so that's something... :)

Issue #1859, opened 2023-03-03
`toolchains` is not forwardable to `cargo_build_script`


UebelAndre:
Applying a similar diff (obviously to a project which defines `rules_perl`) will result in an error where `$(PERL)` is not defined.
```diff
diff --git a/test/cargo_build_script/BUILD.bazel b/test/cargo_build_script/BUILD.bazel
index 55b5db2a..99de9113 100644
--- a/test/cargo_build_script/BUILD.bazel
+++ b/test/cargo_build_script/BUILD.bazel
@@ -8,7 +8,8 @@ load("//rust:defs.bzl", "rust_library", "rust_test")
 cargo_build_script(
     name = "tools_exec_build_rs",
     srcs = ["build.rs"],
-    build_script_env = {"TOOL": "$(execpath :tool)"},
+    build_script_env = {"TOOL": "$(execpath :tool)", "PERL_EXECUTABLE": "$(PERL)"},
+    toolchains = ["@rules_perl//:current_toolchain"],
     edition = "2018",
     # Add a flag to test that they're exposed to the build script
     rustc_flags = ["--verbose"],
```

This is caused by the kwargs of the `cargo_build_script`wrapper not making it to the underlying rule. This should be fixed.

PR #1862, opened 2023-03-04
Forward `toolchains` to `cargo_build_script` targets

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/1859

### DIFF
--- a/test/cargo_build_script/build.rs
+++ b/test/cargo_build_script/build.rs
@@ -18,9 +18,24 @@ fn test_encoded_rustflags() {
     assert_eq!(flags[1], "--verbose");
 }
 
+
+/// Ensure Make variables provided by the `toolchains` attribute are expandable.
+fn test_toolchain_var() {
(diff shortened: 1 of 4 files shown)


test/cargo_build_script/build.rs
```diff
@@ -18,9 +18,21 @@ fn test_encoded_rustflags() {
     assert_eq!(flags[1], "--verbose");
 }
 
+
+/// Ensure Make variables provided by the `toolchains` attribute are expandable.
+fn test_toolchain_var() {
+    let tool_path = std::env::var("EXPANDED_TOOLCHAIN_VAR").unwrap();
```
illicitonion, line 24: Let's also assert that the path pointed to exists?

illicitonion (approved):

test/cargo_build_script/build.rs
UebelAndre: Done!

illicitonion (approved):

PR #1865, opened 2023-03-04
fix example link

nyurik:
example link to protobuf github dir does not work in https://bazelbuild.github.io/rules_rust/rust_proto.html


UebelAndre (approved):
Thanks!

nyurik:
@UebelAndre I tried to run `./docs/update_docs.sh` per CI error description, but there are no changed files. I am not sure why CI fails

UebelAndre (changes requested):
The docs are backed by starlark code. You probably updated the outputs and upon regenerating docs in CI your change created a diff.

PR #1868, opened 2023-03-09
Have rustdoc return its output directory instead of zip default.

UebelAndre:
This is a breaking change for direct consumers of the `rust_doc` targets but fortunately the fix would be to simply consume the `.zip` target with the same name. The benefit to this change though is zips are not created by default which saves time.


illicitonion (approved):

UebelAndre:
@scentini @krasimirgg would either of you be able to find out what's going on with the windows environment here? My changes here shouldn't impact these examples at all.
https://buildkite.com/bazel/rules-rust-rustlang/builds/8297#0186f03d-d1c4-4205-a79d-73fce2fce071

scentini:
Sorry for the delay @UebelAndre , I've been quite swamped with other work recently. I'll try to take a look at the failure tomorrow.

PR #1871, opened 2023-03-11
Add support for thumbv7em with hard float

Codetector1374:


hexdae:
Thanks for opening this. Having an example of cross-compilation for this target would also be awesome. I could help out with this if needed

rust/platform/triple_mappings.bzl
```diff
@@ -188,11 +194,32 @@ _SYSTEM_TO_STDLIB_LINKFLAGS = {
...
 def cpu_arch_to_constraints(cpu_arch):
+    if cpu_arch not in _CPU_ARCH_TO_BUILTIN_PLAT_SUFFIX:
+        fail("CPU architecture \"{}\" is not supported by rules_rust".format(cpu_arch))
+
     plat_suffix = _CPU_ARCH_TO_BUILTIN_PLAT_SUFFIX[cpu_arch]
 
-    if not plat_suffix:
+    return ["@platforms//cpu:{}".format(plat_suffix)]
+
+def cpu_arch_constraints_from_triple(triple):
```
UebelAndre, line 204: Why not update `cpu_arch_to_constraints` to match the pattern of `abi_to_constraints`? I don't particularly care which pattern is used but I would like there to be a consistent API.

UebelAndre (changes requested):
Thanks! One question.

rust/platform/triple_mappings.bzl
Codetector1374: `cpu_arch_to_constraints` is also used in 
https://github.com/bazelbuild/rules_rust/blob/7f2dd433ac0f0fee6ee1ef90544a648640a66526/rust/platform/platform.bzl#L36
which I don't think it has access to the full abi / triple.

I can update `cpu_arch_constraints_from_triple` to match `abi_to_constraints` if that is what you are asking for.

UebelAndre: Yeah, I would like that :)

Codetector1374: Updated with the new interface.

UebelAndre (approved):
Thank you!

UebelAndre:
Looks like you need a rebase

Codetector1374:
done

PR #1874, opened 2023-03-12
Parallelize cargo tree calls

JohnAmican:
In metadata.rs `cargo tree` is called per platform triple. Have these calls happen in parallel.

Context: https://bazelbuild.slack.com/archives/CSV56UT0F/p1678470889893829?thread_ts=1678379024.023019&cid=CSV56UT0F


illicitonion (approved):
Thanks so much!

PR #1875, opened 2023-03-13
add cycle detection to RA gen tool

snowp:
The rust-analyzer tool will choke up with an unhelpful "failed to make progress" error when the BUIlD file dependencies form a cycle. This has the script attempt to detect a cycle in the final set of crates to provide a better error message to the user, printing out the path of dependencies that resulted in a cycle being detected 

This is effectively a piece of code I added to debug this for my own usage, figured it might be helpful to upstream (and save myself the pain next time)


tools/rust_analyzer/rust_project.rs
```diff
@@ -177,7 +177,13 @@ pub fn generate_rust_project(
             for unmerged_crate in &unmerged_crates {
                 let mut path = vec![];
                 if let Some(cycle) = detect_cycle(unmerged_crate, &crate_map, &mut path) {
-                    log::warn!("Cycle detected: {:?}", cycle.iter().map(|c| c.display_name.to_string()).collect::<Vec<String>>());
+                    log::warn!(
```
UebelAndre, line 180: Is this just a formatting change?

UebelAndre:
closing as the same change was made in https://github.com/bazelbuild/rules_rust/pull/2487

PR #1876, opened 2023-03-13
Temp debugging ci

bsilver8192:
I'm using this PR to try debugging Windows-only problems without a Windows dev box for #1782.


bsilver8192:
I give up on Windows, don't need this PR any more.

Issue #1850, opened 2023-02-28
Crate universe: bug with missing crate_features from adding an unrelated dependency


ttiurani:
I'm migrating from raze to crate universe and I almost got everything to work. The crate universe way of working is so much better, thank you for the awesome work!

I did stumble on a bug though, a reproduction of which can be found here:

https://github.com/ttiurani/rules_rust/tree/main/examples/crate_universe_failure/missing_crate_features

The problem is that without this (totally unrelated) project in the workspace linking to `age`, `crate_features` are correct. After adding `age` to a project in a different crate, two of the three needed `crate_features` disappear, and the compilation fails for the `aead_dep` crate too.

I don't expect this to happen, and Cargo builds this just fine.

What might be the problem?

fmorency:
I stumbled on this bug myself after upgrading from `rules_rust` 0.17.0 to 0.18.0. It looks like a regression since it works fine on 0.17.0.

![2023-02-28_16-16](https://user-images.githubusercontent.com/1102868/221987687-5750d475-8064-47f6-bfb6-ce5ca4e5e6b7.png)

UebelAndre:
This may be a regression from:
- https://github.com/bazelbuild/rules_rust/pull/1710

wmatthews-google:
I can definitely take a look... 
Thanks for the repro & sorry for the regression!

gferon:
Looks like this might fix your problem, can you maybe try it out? https://github.com/bazelbuild/rules_rust/pull/1884

ttiurani:
@gferon Yes, that fixes my issue, in the minimal replication, thanks! (I'm unfortunately unable to test it in my real, because I don't know how to get bootstrapping to work.)

Hopefully your fix can be merged and released soon!

ttiurani:
I verified with 0.19.1 that the problem is gone in my real workspace as well. Thank you!

Issue #1609, opened 2022-10-26
crates_repository: how to define a crate twice with a different set of features ?  


QuentinPerez:
:wave:
Is there a way to define a crate multiple times with a different set of features ?
I tried something like that but it seems that the crate is filtered out.
crates_repository(
    ...
    packages = {
        "tonic-tls": crate.spec(
            features = ["tls"],
            package = "tonic",
            version = "0.8",
        ),
        "tonic": crate.spec(
            version = "0.8",
        ),
    }
)
❯ bazel query @crate_index//... | grep tonic
Loading: 0 packages loaded
Loading: 0 packages loaded
Loading: 0 packages loaded
# no more tonic dependency
Let me know it it's the right way to do it

UebelAndre:
Odd, I would have expected that to work. I can take a look.

UebelAndre:
Finally got some time to look into this and this seems to actually be an error in Cargo that is not surfaced when running `cargo metadata`. Using a small repro `Cargo.toml` I can generate an error:
```toml
[package]
edition = "2018"
name = "direct"
version = "0.0.1"

[lib]
name = "direct"
path = "direct.rs"
required-features = []

[dependencies.tonic]
features = []
version = "0.8"

[dependencies.tonic-tls]
features = ["tls"]
package = "tonic"
version = "0.8"

[workspace]
members = []
resolver = "1"
```
```
the crate `direct v0.0.1 (/Users/Code/resolve-repro)` depends on crate `tonic v0.8.2` multiple times with different names
```
When splicing this manifest, `crate_universe` receives a json dump that has effectively an empty [resolve section](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html#output-format)
```json
"resolve": {
    "nodes": [
        {
            "id": "direct 0.0.1 (path+file:///Users/Code/resolve-repro)",
            "dependencies": [],
            "deps": [],
            "features": []
        }
    ],
    "root": "direct 0.0.1 (path+file:///Users/Code/resolve-repro)"
},
```

[...]

PR #1884, opened 2023-03-16
Accumulate all features from cargo tree output

gferon:
Consider the following `cargo tree --target aarch64-apple-darwin --workspace --prefix=none --format="{p} {f}"` output (which you can obtain with a dependency to `matrix-sdk = 0.6` for example):

```
js_int v0.2.2 default,std
js_int v0.2.2 default,serde,std
js_int v0.2.2 default,serde,std (*)
js_int v0.2.2 default,serde,std (*)
js_int v0.2.2 default,serde,std (*)
```

without this patch, the `crate_features` map will only contain `default` and `std` which means the dependents of the `js_int` crate will fail to compile when they require serialization. 

Since universe doesn't really support the compilation of the same crate with different feature sets (yet), it makes more sense to me to pick the biggest set of features.


gferon:
I suppose this is somewhat related to https://github.com/bazelbuild/rules_rust/issues/1609

gferon:
Note: Looks like the CI failure is unrelated to this change

gferon:
Confirmed to fix #1850

gferon:
Absolutely, will do!

UebelAndre:
@wmatthews-google would you also be able to take a look?

wmatthews-google:
This basically looks good to me. Thanks!

+1 for add a unit test (or perhaps extend `parse_features_from_cargo_tree_output_prefix_none` in crate_universe/src/metadata.rs.)

I think we could probably take the repro example from #1850 and make it an integration test case (crate_universe/tests/cargo_integration_test.rs), but I can do this as a followup PR if you'd like.

UebelAndre:
Looks like there are some failing CI jobs though
https://buildkite.com/bazel/rules-rust-rustlang/builds/8301#0186ffc5-40d5-4f54-82b1-334eab577698

gferon:
@UebelAndre is the cargo package cache maybe pinned somewhere in CI?

crate_universe/tests/cargo_integration_test.rs
```diff
@@ -261,3 +261,36 @@ fn feature_generator_workspace() {
...
+    // we just skip this test case.
+    use std::net::ToSocketAddrs;
+    if "github.com:443".to_socket_addrs().is_err() {
+        eprintln!("This test case requires network access. Skipping!");
+        return;
+    }
+
+    let runfiles = runfiles::Runfiles::create().unwrap();
+    let metadata = run(
+        "crate_combined_features(",
```
UebelAndre, line 279:
```suggestion
        "crate_combined_features",
```

crate_universe/tests/cargo_integration_test.rs
```diff
@@ -261,3 +261,36 @@ fn feature_generator_workspace() {
...
+    // other environments) disallow or don't support this. In those cases,
+    // we just skip this test case.
+    use std::net::ToSocketAddrs;
+    if "github.com:443".to_socket_addrs().is_err() {
+        eprintln!("This test case requires network access. Skipping!");
+        return;
+    }
+
+    let runfiles = runfiles::Runfiles::create().unwrap();
+    let metadata = run(
```
UebelAndre, line 278: Perhaps `run` should run `cargo fetch` before calling `splice` to make sure the registry is up to date? I bet `cargo tree --locked` does not update the index? which might be a bug there?

Maybe the long term fix is to have https://github.com/bazelbuild/rules_rust/blob/0.19.0/crate_universe/src/splicing.rs#L419-L451 call `cargo fetch` or something that updates the index in the event that the function does not actually call `cargo generate-lockfile`?

gferon: I did a quick test and `cargo tree --locked` does update the registry properly (at least on my machine) even after clearing `~/.cargo/registry`.

gferon: One more feature I didn't really think of... [`matrix-sdk = 0.6.2` requires Rust 1.65](https://github.com/matrix-org/matrix-rust-sdk/blob/main/Cargo.toml#L20), thanks for figuring this out.

gferon:
Changed the test to use another example, which duplicate `serde` with different feature sets in the output of `cargo tree`. This works with both Rust 1.59 and 1.68. I think it's good to go! 🚀

PR #1885, opened 2023-03-17
Implement support for optional crates enabled with dep: features

gferon:
Skip crates that are marked as `optional = true` and not enabled with a `dep:` feature.

For example, the `matrix-sdk` uses this extensively (see: https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk/Cargo.toml) and this patch makes it work with `cargo-bazel`.


gferon:
> Thanks for putting up this PR! Would you be able to add some testing for this? pray smile

For sure, I first wanted to get #1884 merged, which I'm fixing right now.

gferon:
@UebelAndre I added an integration test using `clap` as an example, but unfortunately this will required CI to run with Rust 1.60 since this is when [the new `dep:` cargo feature syntax was stabilized](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features).

gferon:
CI is finally green after fixing conflicts and merging #1893 🎉

crate_universe/tests/cargo_integration_test.rs
```diff
@@ -295,3 +295,67 @@ fn feature_generator_crate_combined_features() {
         json!(["default", "derive", "serde_derive", "std"])
     );
 }
+
+#[test]
+fn feature_generator_crate_optional_deps_disabled() {
```
UebelAndre, line 300: Is this something that can be unit tested instead? I don't know that this feature *needs* to be integration tested since all the info exists in metadata?

UebelAndre (changes requested):

crate_universe/src/metadata/dependency.rs
```diff
@@ -113,6 +114,30 @@ impl DependencySet {
     }
 }
 
+fn is_optional_crate_enabled(parent: &Node, dep: &NodeDep, metadata: &CargoMetadata) -> bool {
```
UebelAndre, line 117:
```suggestion
/// For details on optional dependencies see [the Rust docs](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies)
fn is_optional_crate_enabled(parent: &Node, dep: &NodeDep, metadata: &CargoMetadata) -> bool {
```

UebelAndre (approved):
Thanks! Looks great! I had one suggestion but not absolutely necessary. Can merge tomorrow morning 😄

PR #1888, opened 2023-03-20
Fix triple constraints for iOS and watchOS

### LINKED ISSUE #1887: Linking fails when building for iOS simulator
When building a multi-platform artifact (xcframework) with `rules_rust` 0.19, it fails with the following message: `building for iOS Simulator, but linking in object file built for iOS`

relevant longer log excerpt:

```
ld: in bazel-out/ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-fastbuild-ST-719010c802cc/bin/external/raze__rust_icu_udata__2_0_3/librust_icu_udata-1703024191.a(rust_icu_udata-1703024191.rust_icu_udata.fe6f27bd-cgu.0.rcgu.o), building for iOS Simulator, but linking in object file built for iOS, file 'bazel-out/ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-fastbuild-ST-719010c802cc/bin/external/raze__rust_icu_udata__2_0_3/librust_icu_udata-1703024191.a'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Error in child process '/usr/bin/xcrun'. 1
```

I have tracked it down to #1805 since the problem is solved when reverting this commit.

gferon:
The logic introduced/changed in #1805 was not correct. This PR simplifies and fixes it.

Only triples for iOS and watchOS have a simulator ABI, and the following triples need to be set to `@build_bazel_apple_support//constraints:simulator`:

* aarch64-apple-ios-sim
* aarch64-apple-watchos-sim
* x86_64-apple-watchos-sim
* x86_64-apple-ios

All the other iOS and watchOS triples need to have the `@build_bazel_apple_support//constraints:device` constraints.

I was able to successfully build a multi platform (x86_64 sim, aarch64 sim and aarch64 device) xcframework with this patch as a test.

Resolves #1887

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -243,43 +243,16 @@ def abi_to_constraints(abi, *, arch = None, system = None):
         List: A list of labels to constraint values
     """
 
-    if not abi:
-        return []
-
-    if abi == "sim":
-        if not system:
-            fail("The ABI value {} is ambiguous. Please specify a system to match the right constraints.")
-        if not arch:
-            fail("The ABI value {} is ambiguous. Please specify an architecture to match the right constraints.")
-        if system == "ios":
-            if abi == "sim":
-                return ["@build_bazel_apple_support//constraints:simulator"]
-        elif arch == "aarch64":  # Only add device for archs that have both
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thanks!

Issue #1892, opened 2023-03-21
Re-enable bazel-zig-cc example


UebelAndre:
Builds are currently failing https://buildkite.com/bazel/rules-rust-rustlang/builds/8317#018703e4-250a-4202-b9e9-607627a6a24e

This issue tracks re-enabling it.

PR #1893, opened 2023-03-21
Disable zig example

UebelAndre:
see https://github.com/bazelbuild/rules_rust/issues/1892


UebelAndre:
cc @illicitonion I'd like to merge other things and don't want this job to be blocking those changes. I see this as a temporary change.

illicitonion (approved):
Ship it - I'm debugging, but won't be able to get a fix in today

PR #1891, opened 2023-03-21
Update bazel-zig-cc example

UebelAndre:
Minor cleanup but accounts for the migration of `bazel-zig-cc` from https://git.sr.ht/~motiejus/bazel-zig-cc to https://github.com/uber/bazel-zig-cc

PR #1894, opened 2023-03-21
Update zig toolchain to a released version

illicitonion:


UebelAndre:
I case it helps I made some progress here: https://github.com/bazelbuild/rules_rust/pull/1891

PR #1899, opened 2023-03-24
Support `-Cpanic=abort`

bsilver8192:
This takes some effort to switch back to `-Cpanic=unwind` for tests and
proc macros, which are incompatible with it.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -682,6 +682,9 @@ _common_attrs = {
     "_is_proc_macro_dep_enabled": attr.label(
         default = Label("//:is_proc_macro_dep_enabled"),
     ),
+    "_panic_style": attr.label(
+        default = Label("//:panic_style"),
+    ),
     "_process_wrapper": attr.label(
         doc = "A process wrapper for running rustc on all platforms.",
         default = Label("//util/process_wrapper"),
@@ -1125,11 +1128,25 @@ rust_library_without_process_wrapper = rule(
     incompatible_use_toolchain_transition = True,
 )
 
+def _force_panic_unwind_transition_impl(settings, attr):
+    _ignore = (settings, attr)
+    return {"//:panic_style": "unwind"}
+
+_force_panic_unwind_transition = transition(
+    implementation = _force_panic_unwind_transition_impl,
+    inputs = [],
+    outputs = ["//:panic_style"],
+)
+
 rust_test = rule(
     implementation = _rust_test_impl,
     provides = _common_providers,
     attrs = dict(_common_attrs.items() +
-                 _rust_test_attrs.items()),
+                 _rust_test_attrs.items() + {
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+    }.items()),
     executable = True,
     fragments = ["cpp"],
     host_fragments = ["cpp"],
@@ -1138,6 +1155,7 @@ rust_test = rule(
         str(Label("//rust:toolchain_type")),
         "@bazel_tools//tools/cpp:toolchain_type",
     ],
+    cfg = _force_panic_unwind_transition,
     incompatible_use_toolchain_transition = True,
     doc = dedent("""\
         Builds a Rust test crate.
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -53,6 +53,13 @@ ErrorFormatInfo = provider(
     fields = {"error_format": "(string) [" + ", ".join(_error_format_values) + "]"},
 )
 
+_panic_style_values = ["unwind", "abort"]
+
+PanicStyleInfo = provider(
+    doc = "Set the panic style",
+    fields = {"panic_style": "(string) [" + ", ".join(_panic_style_values) + "]"},
+)
+
 ExtraRustcFlagsInfo = provider(
     doc = "Pass each value as an additional flag to non-exec rustc invocations",
     fields = {"extra_rustc_flags": "List[string] Extra flags to pass to rustc in non-exec configuration"},
@@ -606,7 +613,7 @@ def collect_inputs(
         force_depend_on_objects (bool, optional): Forces dependencies of this rule to be objects rather than
             metadata, even for libraries. This is used in rustdoc tests.
         experimental_use_cc_common_link (bool, optional): Whether rules_rust uses cc_common.link to link
-	    rust binaries.
+        rust binaries.
 
     Returns:
         tuple: A tuple: A tuple of the following items:
@@ -879,6 +886,8 @@ def construct_arguments(
 
     rustc_flags.add("--error-format=" + error_format)
 
+    rustc_flags.add("-Cpanic=" + _get_panic_style(ctx))
+
     # Mangle symbols to disambiguate crates with the same name. This could
     # happen only for non-final artifacts where we compute an output_hash,
     # e.g., rust_library.
@@ -1019,6 +1028,26 @@ def construct_arguments(
     )
 
     return args, env
+
+def _get_panic_style(ctx):
+    # Hard-code this because I can't figure out how perform the "exec" transition + change unwind
+    # styles for proc macros. Proc macros indicate errors by panicking, which need to unwind to be
+    # reported properly.
+    if is_exec_configuration(ctx):
+        return "unwind"
+    panic_style = "unwind"
+    if hasattr(ctx.attr, "_panic_style"):
+        panic_style = ctx.attr._panic_style[PanicStyleInfo].panic_style
+    return panic_style
+
+def _get_libstd_and_allocator_ccinfo(ctx, toolchain):
+    panic_style = _get_panic_style(ctx)
+    if panic_style == "unwind":
+        return toolchain.unwind_libstd_and_allocator_ccinfo
+    elif panic_style == "panic":
+        return toolchain.panic_libstd_and_allocator_ccinfo
+    else:
+        fail("Unrecognized panic style: " + panic_style)
 
 def rustc_compile_action(
         ctx,
@@ -1258,7 +1287,7 @@ def rustc_compile_action(
         )
 
         # Collect the linking contexts of the standard library and dependencies.
-        linking_contexts = [toolchain.libstd_and_allocator_ccinfo.linking_context, toolchain.stdlib_linkflags.linking_context]
+        linking_contexts = [_get_libstd_and_allocator_ccinfo(ctx, toolchain).linking_context, toolchain.stdlib_linkflags.linking_context]
 
         for dep in crate_info.deps.to_list():
             if dep.cc_info:
@@ -1477,9 +1506,10 @@ def establish_cc_info(ctx, attr, crate_info, toolchain, cc_toolchain, feature_co
             else:
                 cc_infos.append(dep.cc_info)
 
-    if crate_info.type in ("rlib", "lib") and toolchain.libstd_and_allocator_ccinfo:
+    libstd_and_allocator_ccinfo = _get_libstd_and_allocator_ccinfo(ctx, toolchain)
+    if crate_info.type in ("rlib", "lib") and libstd_and_allocator_ccinfo:
         # TODO: if we already have an rlib in our deps, we could skip this
-        cc_infos.append(toolchain.libstd_and_allocator_ccinfo)
+        cc_infos.append(libstd_and_allocator_ccinfo)
 
     return [cc_common.merge_cc_infos(cc_infos = cc_infos)]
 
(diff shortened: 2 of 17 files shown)


bsilver8192:
Looks like wasm linking with cc_library is currently broken on master, because wasm only supports `-Cpanic=abort`.

bsilver8192:
> Thank you! This is super exciting and I'm looking to review it but will need some time to read up on the details of panicking. Want to pay special attention to this because of the new test transition and since the default rustc panicking strategy (when we don't pass any -Cpanic seems non-obvious, target-, and final-artifact-type- specific).

Yes indeed.

> Just to start with, what's your high-level expectations about this? Naively, since tests don't support `-Cpanic=abort` right now, if I wanted a rust_library or a rust_binary built with `-Cpanic=abort`, I would try something along the lines of: `bazel build --@rules_rust//:extra_rustc_flags=-Cpanic=abort //path/to:rust_lib`, possibly adding it to a Bazel config file, and not invoking it for tests... Does this suffice for your workflow?

Basically I want things to work like they do with `cargo`, where I can build my binaries with `-Cabort=panic` but still have tests and proc macros work. My initial motivation for not using a flag is wanting `bazel build` and `bazel test` to share the local cache nicely, if you swap the flag then they overwrite things and end up over-rebuilding every time you switch. I later realized that proc macros also don't work correctly if you just add the flag to everything.

Also having the stdlib code dig through `extra_rustc_flag` and `extra_rustc_flags` to determine which library to include seems messy.

> I also wonder how can we evolve this to also support the unstable `-Zpanic-abort-tests` feature ([rust-lang/rust#67650](https://github.com/rust-lang/rust/issues/67650))?

I'm thinking that would be another flag, and then switch tests to use something besides `force_panic_unwind_transition` which looks at the flag. I think proc macros will still use `force_panic_unwind_transition` as-is.

I'm hoping to have time to debug the Windows test failures this week/weekend. I need to get a Windows dev environment set up, for this and another of my open PRs.

rust/toolchain.bzl
```diff
@@ -179,20 +180,20 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_std, allocator_library):
 
         # The libraries panic_abort and panic_unwind are alternatives.
         # The std by default requires panic_unwind.
-        # Exclude panic_abort if panic_unwind is present.
-        # TODO: Provide a setting to choose between panic_abort and panic_unwind.
+        # Exclude panic_abort if panic_unwind is present, and vice versa.
         filtered_between_core_and_std_files = rust_stdlib_info.between_core_and_std_files
-        has_panic_unwind = [
+        if panic == "abort":
+            panic_filter = "panic_unwind"
```
scentini, line 186: I believe there is also `libunwind` that we want to filter out in `--panic=abort` mode?

rust/private/rustc.bzl
```diff
@@ -1029,6 +1038,30 @@ def construct_arguments(
 
     return args, env
 
+def _get_panic_style(ctx, toolchain):
+    panic_style = "unwind"
+
+    # Most targets default to unwind, but a few default to abort. Can't find a list in the
+    # documentation, this list is extracted from `library/panic_unwind/src/lib.rs` as of 1.68.1.
+    target_triple = toolchain.target_triple
+    if target_triple.arch.startswith("wasm") or target_triple.arch == "avr" or target_triple.system in ("none", "uefi", "espidf"):
```
scentini, line 1047: Could we move this logic over to `toolchain.bzl`, and use a `toolchain.default_panic_style` here?

rust/toolchain.bzl
```diff
@@ -121,13 +121,14 @@ def _ltl(library, ctx, cc_toolchain, feature_configuration):
...
 
-def _make_libstd_and_allocator_ccinfo(ctx, rust_std, allocator_library):
+def _make_libstd_and_allocator_ccinfo(ctx, rust_std, allocator_library, panic):
     """Make the CcInfo (if possible) for libstd and allocator libraries.
 
     Args:
         ctx (ctx): The rule's context object.
         rust_std: The Rust standard library.
         allocator_library: The target to use for providing allocator functions.
+        panic: Either "unwind" or "abort" to selection which panic implementation to include.
```
scentini, line 131:
```suggestion
        panic: Either "unwind" or "abort" to select which panic implementation to include.
```

test/unit/toolchain_make_variables/toolchain_make_variables_test.bzl
```diff
@@ -16,6 +16,13 @@ _RUSTFMT_TOOLCHAIN_ENV = {
     "ENV_VAR_RUSTFMT": "$(RUSTFMT)",
 }
 
+def _mangle_path(path):
```
scentini, line 19: I don't quite understand the change here, and even less how it relates to this PR in general, do you mind elaborating?

.bazelci/presubmit.yml
```diff
@@ -496,6 +496,15 @@ tasks:
       - "//..."
     build_flags:
       - "--@rules_rust//rust/settings:experimental_use_cc_common_link=True"
+  ubuntu2004_panic_abort:
+    name: -Cabort=panic
```
scentini, line 500:
```suggestion
    name: -Cpanic=abort
```

test/panic_style/defs.bzl
```diff
@@ -0,0 +1,29 @@
...
+)
+
+def _with_panic_style_cfg_impl(ctx):
+    return [DefaultInfo(files = depset(ctx.files.srcs))]
+
+with_panic_style_cfg = rule(
+    implementation = _with_panic_style_cfg_impl,
+    attrs = {
+        "panic_style": attr.string(mandatory = True),
+        "srcs": attr.label_list(
```
scentini, line 21: This is not really `srcs` but rather `binary`, right? As in, we pass the executable generated by a `rust_binary` here?

scentini:
As for the Windows failures, at least this one is due to too long a path, due to the transition hash we now have in the output path:
```
LINK : fatal error LNK1104: cannot open file 'C:\b\tycizaqx\execroot\examples\bazel-out\x64_windows-opt-exec-2B5CBBC6-ST-e9b575d06caf\bin\external\rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools\rust_toolchain\lib\rustlib\x86_64-pc-windows-msvc\lib\librustc_std_workspace_alloc-15cc9d7e933adbe4.rlib'
```
I think we might be able to shorten this section: `rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools` by  quite a bit.

scentini:
I'm afraid the rest of the Windows failures might be due to too long a path too :( The `Crate Universe Unnamed Examples` fails with
`note: LINK : fatal error LNK1181: cannot open input file 'windows.lib'`
I tracked down `windows.lib` to the [`data`](https://github.com/bazelbuild/rules_rust/blob/main/examples/crate_universe_unnamed/vendor_remote_manifests/crates/BUILD.windows_x86_64_msvc-0.42.0.bazel#L54) section of a build script, which manifests as
```
"/LIBPATH:C:\\b\\vaql5ptt\\execroot\\__main__\\bazel-out/x64_windows-opt-exec-2B5CBBC6-ST-e9b575d06caf/bin/external/crates_vendor_manifests__windows_x86_64_msvc-0.42.2/windows_x86_64_msvc_build_script_.exe.runfiles/crates_vendor_manifests__windows_x86_64_msvc-0.42.2/lib" 
``` 
on the command line - the path being 255 characters. Appending `/windows.lib` puts us over the limit.

We don't have much choice in the matter but trying to trim the generated paths. In this case we could maybe shorten the `crates_vendor_manifests` directory to gain some characters back. @bsilver8192 are you up for pursuing this?

test/unit/toolchain_make_variables/toolchain_make_variables_test.bzl
bsilver8192: With the transitions, the outdir component of the path ends up different, even though it's pointing to the same file. The test fails if you revert this change.

scentini:
Hey @bsilver8192 do you have a timeline for pushing this PR through?
I believe my https://github.com/bazelbuild/rules_rust/pull/1926 has introduced some merge conflicts. Upcoming work on supporting `no_std` builds is going to further intertwine with code you're touching here. Finally, after I'm done with `no_std` I will very soon need `--Cpanic=abort` myself :)

UebelAndre:
@bsilver8192 @scentini  Frinedly ping, any updates here?

bsilver8192:
@UebelAndre I tried poking at #1782 to try getting that one finished first, but the Windows command environment is truly a nightmare. There's two shells plus Rust with four different kinds of processing that all have to pass arbitrary strings through without mangling them. This one is stuck behind issues in the same area, and I'm not going to get around to fixing it anytime soon. I don't have experience with or interest in Windows shells and command line paths, and these are not quick fixes (at least for me). It's kind of frustrating that both of my changes are blocked on pre-existing issues with a platform that I don't use, but at this point I've concluded that fixing them is never going to get to the top of my priority list.

I think the Windows issues are independent of this and my other PR, so somebody else can grab the tests I wrote that surface the issues and fix them. Alternatively we could just agree to skip the new tests on Windows for now. If somebody else can fix the Windows issues, I'd be happy to resolve the merge conflicts and finish the review. Alternatively if somebody else wants to take over the PRs, I'd be fine with that.

scentini:
@bsilver8192 if you rebase and move forward with the review, I could work on clearing the Windows hurdles. Alternatively I'm also happy to take over the PR.

.bazelci/presubmit.yml
bsilver8192: Done

rust/toolchain.bzl
bsilver8192: Done

rust/toolchain.bzl
bsilver8192: I think that's used for calculating stacktraces independent of the panic mode. Also I think I found the part of rustc that implements the switching logic at https://github.com/rust-lang/rust/blob/42faef503f3e765120ca0ef06991337668eafc32/compiler/rustc_metadata/src/creader.rs#L746, and it's only swapping the `panic_unwind` and `panic_abort` crates, not doing anything with `libunwind`.

rust/private/rustc.bzl
```diff
@@ -1907,6 +1941,34 @@ error_format = rule(
...
+    raw = ctx.build_setting_value
+    if raw not in _panic_style_values:
+        fail("{} expected a value in `{}` but got `{}`".format(
+            ctx.label,
+            _panic_style_values,
+            raw,
+        ))
+    return [PanicStyleInfo(panic_style = raw)]
+
+panic_style = rule(
```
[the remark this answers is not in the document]
bsilver8192: I hadn't seen that, that is nicer. Done

rust/private/rustc.bzl
bsilver8192: Done

test/panic_style/defs.bzl
bsilver8192: In this case, yes. If you look at this rule as a more general way of applying that transition to any set of things, then `srcs` is the more common name. I don't really care which name, do you want me to change it?

rust/private/rustc.bzl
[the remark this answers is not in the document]
bsilver8192: Although that does mean it disappears from the docs.

bsilver8192:
Ok @scentini I finally got back to this, sorry for the delay. I rebased and I think I responded to all your comments, when you get a chance could you take another look and work on the Windows problems?

scentini:
Thanks a lot @bsilver8192 , I'll target to fix the windows failures this week.

rust/toolchain.bzl
```diff
@@ -187,39 +187,29 @@ def _make_libstd_and_allocator_ccinfo(ctx, rust_std, allocator_library, std = "s
 
         # The libraries panic_abort and panic_unwind are alternatives.
         # The std by default requires panic_unwind.
-        # Exclude panic_abort if panic_unwind is present.
-        # TODO: Provide a setting to choose between panic_abort and panic_unwind.
+        # Exclude panic_abort if panic_unwind is present, and vice versa.
```
krasimirgg, line 190: note: I'm looking into whether we may entirely remove this filtering as a follow-up to this work (originally, this was put in place due to the specifics of how rustc formed the underlying linker invocation that includes these libraries, and that has changed since). Nothing actionable for this PR, just informational.

rust/private/rust.bzl
```diff
@@ -1115,6 +1124,7 @@ rust_test = rule(
         str(Label("//rust:toolchain_type")),
         "@bazel_tools//tools/cpp:toolchain_type",
     ],
+    cfg = force_panic_unwind_transition,
```
UebelAndre, line 1127: Can this not also be solved by having this flag appear in the `rust_toolchain` instead of for each test and proc-macro? Also, why is this not for binaries as well?

UebelAndre (changes requested):

rust/private/rust.bzl
bsilver8192: It needs to apply only to tests (the overall goal of this PR is build binaries without this transition). `#[should_panic]` doesn't work with `-Cpanic=abort`, so Cargo automatically builds tests with `-Cpanic=unwind`, which is the behavior I'm copying here.

https://doc.rust-lang.org/cargo/reference/profiles.html#panic documents this:
> Tests, benchmarks, build scripts, and proc macros ignore the `panic` setting. The `rustc` test harness currently requires `unwind` behavior. See the [panic-abort-tests](https://doc.rust-lang.org/cargo/reference/unstable.html#panic-abort-tests) unstable flag which enables `abort` behavior.
>
> Additionally, when using the `abort` strategy and building a test, all of the dependencies will also be forced to build with the `unwind` strategy.

goffrie:
Is there a way we could implement this without using transitions? Transitions are really unfortunate if you depend on any non-Rust code as all that stuff has to be rebuilt even though it doesn't change at all, due to the separate bazel-out directory :/

Previously (before using rules_rust) we had internal rules that supported panic=abort and panic=unwind by creating two actions for each rust_library target, and the provider would expose both versions of the library.

PR #1901, opened 2023-03-24
Fix crate deps with condition

### LINKED ISSUE #1880: Cargo.toml with conditions are failing to build
I have this section in my Cargo.toml

```
[target.'cfg(target_os = "macos")'.dependencies]
 security-framework = "2.8.0"
```

and it's generating errors:

```
                aliases = aliases(),
        File "/private/var/tmp/_bazel_yuweiba/f3b87d172301024018ed3e9f856c70fc/external/crate_index/defs.bzl", line 281, column 30, in aliases
                if condition_triples in crate_aliases:
Error: unhashable type: 'list'
```

I think this is essentially from: https://github.com/bazelbuild/rules_rust/blob/d6e3003594713ba00d03a354cbf7a7c2c65e22d2/proto/3rdparty/crates/defs.bzl#L362

which is generated from https://github.com/bazelbuild/rules_rust/blob/d6e3003594713ba00d03a354cbf7a7c2c65e22d2/crate_universe/src/rendering/templates/module_bzl.j2#L317

I tried casting the list to tuples but it still fails with some other errors.

so not sure best way to fix this?

ibigbug:
Fixed #1880

### DIFF
--- a/crate_universe/src/rendering/templates/module_bzl.j2
+++ b/crate_universe/src/rendering/templates/module_bzl.j2
@@ -17,7 +17,6 @@ Expected length = 6 lines
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
-load("@bazel_skylib//lib:selects.bzl", "selects")
 
 ###############################################################################
 # MACROS API
@@ -200,11 +199,22 @@ def all_crate_deps(
         else:
             return []
 
-    crate_deps = list(dependencies.pop(_COMMON_CONDITION, {}).values())
-    for condition, deps in dependencies.items():
-        crate_deps += selects.with_or({_CONDITIONS[condition]: deps.values()})
+    common_deps = dependencies.pop(_COMMON_CONDITION, {}).items()
 
-    return crate_deps
+    # If there are only common items in the dictionary, immediately return them
+    if len(dependencies.keys()) == 0:
+        return dict(common_deps)
+    
+    crate_deps = {"//conditions:default": dict(common_deps)}
+    for condition, deps in dependencies.items():
+        condition_triples = _CONDITIONS[condition]
+        for triple in condition_triples:
+            if triple in crate_deps:
+                crate_deps[triple].update(deps)
+            else:
+                crate_deps.update({triple: dict(deps.items() + common_deps)})
+
+    return select(crate_deps)
 
 def aliases(
         normal = False,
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Thanks! Is this something you'd be able to add an integration test for at https://github.com/bazelbuild/rules_rust/tree/0.19.1/examples/crate_universe? I thought the alias example there would have caught an issue like this but it's clearly lacking. Could you update this so we don't have aliases break again in the future?

ibigbug:
sure I'll add it. but sorry for the confusion I think we still need some polish to this PR https://github.com/bazelbuild/rules_rust/issues/1880#issuecomment-1483128569

ibigbug:
@UebelAndre I think I made some progress, and got a new error

```
DEBUG: /private/var/tmp/_bazel_yuweiba/f3b87d172301024018ed3e9f856c70fc/external/crate_index/defs.bzl:276:10: {"": {}, "cfg(target_os = \"macos\")": {}}
DEBUG: /private/var/tmp/_bazel_yuweiba/f3b87d172301024018ed3e9f856c70fc/external/crate_index/defs.bzl:276:10: {"": {}}
ERROR: /Users/yuweiba/workspace/projects/clash-rs/clash/BUILD.bazel:4:13: //clash:aarch64-apple-darwin is not a valid select() condition for //clash:lib.
To inspect the select(), run: bazel query --output=build //clash:lib.
For more help, see https://docs.bazel.build/be/functions.html#select.
```

any suggestions where to go next step?

ibigbug:
@UebelAndre sorry i wasn't able to make any progress. do you have any suggestions?

k1nkreet:
> @UebelAndre I think I made some progress, and got a new error
> 
> ```
> DEBUG: /private/var/tmp/_bazel_yuweiba/f3b87d172301024018ed3e9f856c70fc/external/crate_index/defs.bzl:276:10: {"": {}, "cfg(target_os = \"macos\")": {}}
> DEBUG: /private/var/tmp/_bazel_yuweiba/f3b87d172301024018ed3e9f856c70fc/external/crate_index/defs.bzl:276:10: {"": {}}
> ERROR: /Users/yuweiba/workspace/projects/clash-rs/clash/BUILD.bazel:4:13: //clash:aarch64-apple-darwin is not a valid select() condition for //clash:lib.
> To inspect the select(), run: bazel query --output=build //clash:lib.
> For more help, see https://docs.bazel.build/be/functions.html#select.
> ```

I've got to the same issue with my small research on this topic. It looks like the problem is `module_bzl.j2` defining `_CONDITIONS` value uses triples from `context.conditions` holding the values which are not transformed with `platforms_template`. I'm not really familiar with the tool  processing those templates to understand whether it's possible to apply `platforms_template` to `triples` inside the template specification. The other option could be to provide additional value in `context` or change the value of `context.conditions` according to `platforms_template`. I wonder if @UebelAndre do you has any thoughts on that

k1nkreet:
Since I've spent some time tinkering on `cargo-bazel-bin` tool to solve the last issue and I've made some progress on it, I took the liberty of opening [the new PR](https://github.com/bazelbuild/rules_rust/pull/2018) where I've added the integration test/example, took the changes from this PR and added fix [this](https://github.com/bazelbuild/rules_rust/pull/1901#issuecomment-1492029546) problem

PR #1906, opened 2023-03-28
Add `rust_dylib_library` rule

wilsonwang371:
add dylib option support for building a rust library.

To use it, we will have to specify a rust_dylib_library rule.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -50,6 +50,7 @@ PAGES = dict([
             "rust_binary",
             "rust_library",
             "rust_static_library",
+            "rust_dylib_library",
             "rust_shared_library",
             "rust_proc_macro",
             "rust_test",
--- a/docs/defs.md
+++ b/docs/defs.md
@@ -4,6 +4,7 @@
 * [rust_binary](#rust_binary)
 * [rust_library](#rust_library)
 * [rust_static_library](#rust_static_library)
+* [rust_dylib_library](#rust_dylib_library)
 * [rust_shared_library](#rust_shared_library)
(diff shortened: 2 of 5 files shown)


UebelAndre:
https://doc.rust-lang.org/reference/linkage.html for more info on the different linkage types

UebelAndre:
I think this seems fine. @illicitonion @scentini What do you folks think? Should this be a dedicated rule or would you expect `rust_shared_library` to be configurable between `cdylib` or `dylib`?

scentini:
I think it might make sense to consider whether `rust_library` should be configurable between `rlib` and `dylib`. https://doc.rust-lang.org/reference/linkage.html has info on `-C prefer-dynamic`
> If an executable is being produced and the -C prefer-dynamic flag is not specified, then dependencies are first attempted to be found in the rlib format. If some dependencies are not available in an rlib format, then dynamic linking is attempted (see below).

If we have separate `rust_library` and `rust_dylib_library` we couldn't implement equivalent behavior.

From Bazel perspective, if we have a separate `rust_library` and `rust_dylib_library` one would have to duplicate the build graph if one wants to produce `.rlib`s in certain cases and `.dylib`s in others. An example of this would be testing - one might want to use `dylib`s during development, because in that case a change to a single `rust_library` would not trigger full static link when running a unit test (which can be quite slow). The `cc_*` rules have `linkstatic` attribute to toggle the wanted behavior - it is set to `True` for `cc_binary` and to `False` for `cc_test` (https://bazel.build/reference/be/c-cpp#cc_test.linkstatic)

PR #1907, opened 2023-03-28
Add empty rustfmt.toml

P1n3appl3:
This prevents the user's global rustfmt settings from being used on rules_rust. I noticed that my other rules_rust commits were introducing spurious changes because I have `wrap_comments` enabled globally, so I added this empty config file to prevent that. I figured that others may run into the same issue and this could save some diffs in the future where someone sets a different line-length limit or something for their personal configuration.


scentini:
Thanks!

PR #1910, opened 2023-03-30
Support third party crates with bzlmod

matts1:
This PR adds bzlmod support for third party crates via rules_rust.

It only adds some subset of crate_universe's features. Specifically, it supports:
* Using a Cargo.toml  and Cargo.lock file as the source of truth for dependencies.
* Automatically generating build targets for your referenced dependencies in Cargo.toml.
* One repo per Cargo.toml / Cargo.lock, to namespace and prevent implicit dependencies.

Out of scope of this PR is:
* Vendoring deps
* all_crate_deps or similar

### DIFF
--- a/crate_universe/extension.bzl
+++ b/crate_universe/extension.bzl
@@ -1,0 +1,192 @@
+"""Module extension for generating third-party crates for use in bazel."""
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("//crate_universe:defs.bzl", _crate_universe_crate = "crate")
+load("//crate_universe/private:crates_vendor.bzl", "CRATES_VENDOR_ATTRS", "generate_config_file", "generate_splicing_manifest")
+load("//crate_universe/private:generate_utils.bzl", "render_config")
+load("//crate_universe/private/module_extensions:cargo_bazel_bootstrap.bzl", "get_cargo_bazel_runner")
+
+def _generate_repo_impl(repo_ctx):
+    for path, contents in repo_ctx.attr.contents.items():
+        repo_ctx.file(path, contents)
+
+_generate_repo = repository_rule(
+    implementation = _generate_repo_impl,
+    attrs = dict(
+        contents = attr.string_dict(mandatory = True),
+    ),
+)
+
+def _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg):
+    cargo_lockfile = module_ctx.path(cfg.cargo_lockfile)
+    tag_path = module_ctx.path(cfg.name)
+
+    rendering_config = json.decode(render_config(
+        regen_command = "Run 'cargo update [--workspace]'",
+    ))
+    config_file = tag_path.get_child("config.json")
+    module_ctx.file(
+        config_file,
+        executable = False,
+        content = generate_config_file(
+            module_ctx,
+            mode = "remote",
+            annotations = {},
+            generate_build_scripts = cfg.generate_build_scripts,
+            supported_platform_triples = cfg.supported_platform_triples,
+            generate_target_compatible_with = True,
+            repository_name = cfg.name,
+            output_pkg = cfg.name,
+            workspace_name = cfg.name,
+            generate_binaries = cfg.generate_binaries,
+            render_config = rendering_config,
+        ),
+    )
+
+    manifests = {module_ctx.path(m): m for m in cfg.manifests}
+    splicing_manifest = tag_path.get_child("splicing_manifest.json")
+    module_ctx.file(
+        splicing_manifest,
+        executable = False,
+        content = generate_splicing_manifest(
+            packages = {},
+            splicing_config = "",
+            cargo_config = cfg.cargo_config,
+            manifests = {str(k): str(v) for k, v in manifests.items()},
+            manifest_to_path = module_ctx.path,
+        ),
+    )
+
+    splicing_output_dir = tag_path.get_child("splicing-output")
+    cargo_bazel([
+        "splice",
+        "--output-dir",
+        splicing_output_dir,
+        "--config",
+        config_file,
+        "--splicing-manifest",
+        splicing_manifest,
+        "--cargo-lockfile",
+        cargo_lockfile,
+    ])
+
+    # Create a lockfile, since we need to parse it to generate spoke
+    # repos.
+    lockfile_path = tag_path.get_child("lockfile.json")
+    module_ctx.file(lockfile_path, "")
+
+    cargo_bazel([
+        "generate",
+        "--cargo-lockfile",
+        cargo_lockfile,
+        "--config",
+        config_file,
+        "--splicing-manifest",
+        splicing_manifest,
+        "--repository-dir",
+        tag_path,
+        "--metadata",
+        splicing_output_dir.get_child("metadata.json"),
+        "--repin",
+        "--lockfile",
+        lockfile_path,
+    ])
+
+    crates_dir = tag_path.get_child(cfg.name)
+    _generate_repo(
+        name = cfg.name,
+        contents = {
+            "BUILD.bazel": module_ctx.read(crates_dir.get_child("BUILD.bazel")),
+        },
+    )
+
+    contents = json.decode(module_ctx.read(lockfile_path))
+
+    for crate in contents["crates"].values():
+        repo = crate["repository"]
+        if repo == None:
+            continue
+        name = crate["name"]
+        version = crate["version"]
+
+        # "+" isn't valid in a repo name.
+        crate_repo_name = "{repo_name}__{name}-{version}".format(
+            repo_name = cfg.name,
+            name = name,
+            version = version.replace("+", "-"),
+        )
+
+        build_file_content = module_ctx.read(crates_dir.get_child("BUILD.%s-%s.bazel" % (name, version)))
+        if "Http" in repo:
+            # Replicates functionality in repo_http.j2.
+            repo = repo["Http"]
+            http_archive(
+                name = crate_repo_name,
+                patch_args = repo.get("patch_args", None),
+                patch_tool = repo.get("patch_tool", None),
+                patches = repo.get("patches", None),
+                remote_patch_strip = 1,
+                sha256 = repo.get("sha256", None),
+                type = "tar.gz",
+                urls = [repo["url"]],
+                strip_prefix = "%s-%s" % (crate["name"], crate["version"]),
+                build_file_content = build_file_content,
+            )
+        elif "Git" in repo:
+            # Replicates functionality in repo_git.j2
+            repo = repo["Git"]
+            kwargs = {}
+            for k, v in repo["commitish"].items():
+                if k == "Rev":
+                    kwargs["commit"] = v
+                else:
+                    kwargs[k.lower()] = v
+            new_git_repository(
+                name = crate_repo_name,
+                init_submodules = True,
+                patch_args = repo.get("patch_args", None),
+                patch_tool = repo.get("patch_tool", None),
+                patches = repo.get("patches", None),
+                shallow_since = repo.get("shallow_since", None),
+                remote = repo["remote"],
+                build_file_content = build_file_content,
+                strip_prefix = repo.get("strip_prefix", None),
+                **kwargs
+            )
+        else:
+            fail("Invalid repo: expected Http or Git to exist for crate %s-%s, got %s" % (name, version, repo))
+
+def _crate_impl(module_ctx):
+    cargo_bazel = get_cargo_bazel_runner(module_ctx)
+    all_repos = []
+    for mod in module_ctx.modules:
+        local_repos = []
+        for cfg in mod.tags.from_cargo:
+            if cfg.name in local_repos:
+                fail("Defined two crate universes with the same name in the same MODULE.bazel file. Use the name tag to give them different names.")
+            elif cfg.name in all_repos:
+                fail("Defined two crate universes with the same name in different MODULE.bazel files. Either give one a different name, or use use_extension(isolate=True)")
+            _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg)
+            all_repos.append(cfg.name)
+            local_repos.append(cfg.name)
+
+_from_cargo = tag_class(
+    doc = "Generates a repo @crates from a Cargo.toml / Cargo.lock pair",
+    attrs = dict(
+        name = attr.string(doc = "The name of the repo to generate", default = "crates"),
+        cargo_lockfile = CRATES_VENDOR_ATTRS["cargo_lockfile"],
+        manifests = CRATES_VENDOR_ATTRS["manifests"],
+        cargo_config = CRATES_VENDOR_ATTRS["cargo_config"],
+        generate_binaries = CRATES_VENDOR_ATTRS["generate_binaries"],
+        generate_build_scripts = CRATES_VENDOR_ATTRS["generate_build_scripts"],
+        supported_platform_triples = CRATES_VENDOR_ATTRS["supported_platform_triples"],
+    ),
+)
+
+crate = module_extension(
+    implementation = _crate_impl,
+    tag_classes = dict(
+        from_cargo = _from_cargo,
+    ),
+)
--- a/crate_universe/extension.bzl
+++ b/crate_universe/extension.bzl
@@ -2,7 +2,6 @@
 
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-load("//crate_universe:defs.bzl", _crate_universe_crate = "crate")
 load("//crate_universe/private:crates_vendor.bzl", "CRATES_VENDOR_ATTRS", "generate_config_file", "generate_splicing_manifest")
 load("//crate_universe/private:generate_utils.bzl", "render_config")
 load("//crate_universe/private/module_extensions:cargo_bazel_bootstrap.bzl", "get_cargo_bazel_runner")
--- a/examples/bzlmod/hello_world/MODULE.bazel
+++ b/examples/bzlmod/hello_world/MODULE.bazel
@@ -9,21 +9,17 @@ bazel_dep(
     name = "bazel_skylib",
     version = "1.5.0",
 )
-
 bazel_dep(
     name = "rules_rust",
     version = "0.0.0",
 )
-
 local_path_override(
     module_name = "rules_rust",
     path = "../../..",
 )
 
 rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
-
 rust.toolchain(edition = "2021")
-
 use_repo(
     rust,
     "rust_toolchains",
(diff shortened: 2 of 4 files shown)


aaronmondal:
Interesting! I attempted turning `crates_repository` into an extension but didn't get too far yet. Being able to declare a "global" `crates_index` would probably be quite useful. It might also be worth looking at the `@gazelle//:extensions.bzl`s `go_deps` implementation as that already works with bzlmod and is really nice to work with.

cc @fmeum who might be interested in this since `crates_repository` functionality similar to `go_deps` is probably the end goal here.

fmeum:
Thanks for the mention. I will have to read up on cargo a bit to understand in which aspects it is similar to Go.

There is an ongoing effort to simplify isolation of extension usages which you may be able to take advantage of: https://github.com/bazelbuild/bazel/pull/18529

rust/private/module_extensions/toolchain.bzl
```diff
@@ -0,0 +1,77 @@
...
+                rustfmt_version = host_tools.rustfmt_version,
+                sha256s = host_tools.sha256s,
+                urls = host_tools.urls,
+                version = host_tools.version,
+            )
+            break
+
+    mod = module_ctx.modules[0]
+    for toolchain in mod.tags.toolchain:
+        rust_register_toolchains(
```
aaronmondal, line 37: The current naming for `rust_register_toolchains` is quite confusing and should probably be reworked separately.

If it's `rust_register_toolchains` it shouldn't have a `register_toolchain` bool flag. If it's just the configuration it should be called `rust_configure_toolchains` and the toolchain registration should be manually handled by users, even in the WORKSPACE case.

rust/private/module_extensions/toolchain.bzl
```diff
@@ -0,0 +1,77 @@
...
+    version = attr.string(default = rust_common.default_version),
+    **common_kwargs
+))
+
+rust = module_extension(
+    implementation = _rust_impl,
+    tag_classes = {
+        "host_tools": rust_host_tools,
+        "toolchain": rust_toolchain,
+    },
```
aaronmondal, line 76: `doc` field unspecified.

crate_universe/module_extensions/crate.bzl
```diff
@@ -0,0 +1,226 @@
...
+        supported_platform_triples = CRATES_VENDOR_ATTRS["supported_platform_triples"],
+        annotation_files = attr.label_list(allow_files = [".json"]),
+    ),
+)
+
+crate = module_extension(
+    implementation = _crate_impl,
+    tag_classes = dict(
+        from_cargo = _from_cargo,
+    ),
```
aaronmondal, line 225: `doc` field unspecified.

fmeum:
It also makes sense to follow https://bazel.build/external/extension#put_each_extension_in_a_separate_file and have a dedicated file per extension.

MODULE.bazel
```diff
@@ -8,10 +8,26 @@ print("WARNING: The rules_rust Bazel module is still highly experimental and sub
...
+# their repo, it's overridden.
+rust = use_extension("//rust:extensions.bzl", "rust")
+rust.host_tools(edition = "2021")
+use_repo(rust, "rust_host_tools")
+
+cargo_bazel_bootstrap = use_extension("//crate_universe/private/module_extensions:cargo_bazel_bootstrap.bzl", "cargo_bazel_bootstrap")
+use_repo(
+    cargo_bazel_bootstrap,
+    "cargo_bazel_bootstrap",
+)
```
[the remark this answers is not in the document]
matts1: I agree, but that's a question for a different PR. There already exists a bzlmod toolchain, and this is a no-op to the existing toolchain.

crate_universe/private/module_extensions/cargo_bazel_bootstrap.bzl
```diff
@@ -0,0 +1,49 @@
...
+
+    Args:
+        module_ctx: The module extension's context.
+
+    Returns:
+        A function that can be called to execute cargo_bazel.
+    """
+    cargo_path = str(module_ctx.path(Label("@rust_host_tools//:bin/cargo")))
+    rustc_path = str(module_ctx.path(Label("@rust_host_tools//:bin/rustc")))
+    cargo_bazel = module_ctx.path(Label("@cargo_bazel_bootstrap//:cargo-bazel"))
```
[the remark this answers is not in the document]
matts1: Fixed the labels. The cargo_bazel used later is the same variable

matts1:
@aaronmondal PTAL at https://github.com/bazelbuild/rules_rust/pull/2021

I've followed your recommendations re naming, but I'm specifically not applying your suggestions about the toolchain, since although I agree with you, it's out of scope of this PR

opicaud:
Hey all,

Thanks for this @matts1 , I tried your PR and I have few things to share : 
- There is an option called `package` in `crates_repository` rules,  i found only `annotations`, am I wrong if I say that this option is missing in the extension ?
- Using the new crate extensions involves also a **breaking change** : the repo name "<module_name>_crates" breaks the existing usage of the "old" repo in BUILD file. Also the `all_crate_deps` is no longer available, the new version is proposing to list all dependencies as mentioned in the example you have provided : this new list involves duplication of dependencies from `Cargo.toml` in BUILD file, duplication not present with the usage of `all_crate_deps`..

I perfectly understand that is a challenge to make it works with minimum impacts : for brand new project, it's very ok to use it like it is, but for existing projects, i prefer limiting the change for now, even if the implementation is not "optimized", with just your PR [#2021](https://github.com/bazelbuild/rules_rust/pull/2021)   and without using the new `crate` extension i managed to get it works without breaking change even if my solution is far to be elegant..

Olivier

opicaud:
Hey @matts1 

Great :), so for now i can't migrate and I keep what I have, please let me know when `packages` and `all_crate_deps` will be supported, so that i will be able to use them and override the `use_repo` as you mentioned.

Thanks a lot!

Olivier

opicaud:
Hey @matts1 

I've pushed changes from alchemy_v0.20.1 to fetch back the deps.bzl from the generated files in order to use `all_crates_deps` sugar syntax in BUILD.bazel in the bzlmod crate_universe example

It's working by adding crates in a `use_repo`, this is this part of code that I want to get rid of now.

I unfortunately needed to declare those repos because `deps.bzl` is generated with: 
`["@crates_repository_crates__env_logger-0.9.3//:env_logger", "@crates_repository_crates__log-0.4.17//:log"]`, 
but we need something that looks more like 
`["@crates_repository_crates//:env_logger", "@crates_repository_crates//:log"]`.

I try to patch the `render_config` templates to get the right generation in deps.bzl but without success, as BUILD.bazel is also using this template.
Perhaps you have an idea to guide me? Here the [changes](https://github.com/matts1/rules_rust/compare/alchemy_v2...opicaud:all_crates_deps?expand=1)
Thanks!

fmeum:
If you can make it `@@crates_repository_crates__env_logger-0.9.3//:env_logger` (not3 the double at), that should also work.

opicaud:
@fmeum 

Unfortunatley, using the canonical repo as proposed don't make the difference.
Also [the documentation about module](https://bazel.build/versions/6.3.0/external/extension#repository_names_and_visibility) is explaining that relying on the canonical name is not recommended, because subject to change.
What I notice is that: 
- When `crates_repository` is used in a non bzlmod context, repo name declared in the BUILD.bazel and BUILD_<dep>_* and deps.bzl are the same, so no issue.
- When `crates_repository` is used in a bzlmod context, only the repo declared in the `use_repo` is publicly available, which means deps.bzl should contains only aliases target, as other generated repo are private to this public one, so not available at root level (eg: `all_crate_deps` in BUILD file).

IMHO:
- In bzlmod context, generation of deps.bzl is different than from a non bzlmod context
- For now, I suspect deps.bzl to be the only file being different from a bzlmod/non-bzlmod context

opicaud:
Hey, 
I have figured it out for `all_crates_deps` by extending `from_cargo` with an `repo_alias`. Its role is to provide a prefix to add into the generated deps.bzl Now the example provided is working, see [changes](https://github.com/matts1/rules_rust/compare/alchemy_v2...opicaud:all_crates_deps?expand=1)
Still examples like `all_crates_deps( build = True)` are yet not supported, when rendering BUILD file, because only dependencies are generated, and thus the alias does not exist

crate_universe/extension.bzl
```diff
@@ -0,0 +1,218 @@
...
+                name = crate_repo_name,
+                init_submodules = True,
+                patch_args = repo.get("patch_args", None),
+                patch_tool = repo.get("patch_tool", None),
+                patches = repo.get("patches", None),
+                shallow_since = repo.get("shallow_since", None),
+                remote = repo["remote"],
+                build_file_content = build_file_content,
+                strip_prefix = repo.get("strip_prefix", None),
+                **repo["commitish"]
```
stephencpope, line 184: Unfortunately this is not working for me when trying to use this branch (at current commit `dd7e67f72a6bc3ef852fe6a424b37e0ebbcaeadc`) to build my own repo, where I need to pull some of my crates via git from non-public github repos (`my-crate = { git = "ssh://git@github.com/...", rev = "..." }`.

The problem is that the generated `cargo-bazel-lock.json` file has keys in the `commitish` object which are capitalized: `Branch`, `Tag`, or `Rev`:

```
    "my-crate ...": {
      "name": "my-crate",
      "version": "...",
      "repository": {
        "Git": {
          "remote": "ssh://git@github.com/...",
          "commitish": {
            "Rev": "..."
          }
        }
      },
 ```
When expanded into the arguments for `new_git_repository` this fails because of course `new_git_repository` needs them lower case. This results in an error like this:

```
$ bazel query @crates//:\*
ERROR: <builtin>: @rules_rust~override//crate_universe:rules_rust~override~crate~crates__my-crate...: no such attribute 'Rev' in 'git_repository' rule
ERROR: Traceback (most recent call last):
        File ".../external/rules_rust~override/crate_universe/extension.bzl", line 194, column 37, in _crate_impl
                _generate_hub_and_spokes(module_ctx, cargo_bazel, mod_path, cfg)
        File ".../external/rules_rust~override/crate_universe/extension.bzl", line 174, column 31, in _generate_hub_and_spokes
                new_git_repository(
Error in repository_rule: failed to instantiate 'git_repository' from this module extension
ERROR: Target parsing failed due to unexpected exception: error evaluating module extension crate in @rules_rust~override//crate_universe:extension.bzl
```

Comment (stephencpope) [reply crate_universe/extension.bzl:184]: Also noting that `cargo-bazel-lock.json` uses `Rev`, but `new_git_repository` uses `commit`. I am finding happiness 
with this line altered to
```
**{"commit" if k == "Rev" else k.lower(): v for k, v in repo["commitish"].items()}
```

Comment (matts1) [reply crate_universe/extension.bzl]: Should be fixed, though I don't have a git-based repo to test it myself. Please try it yourself to confirm.

File: crate_universe/extension.bzl [hunk 7]
```diff
@@ -0,0 +1,224 @@
...
+        for name, annotations_for_crate in content.items():
+            for annotation in annotations_for_crate:
+                annotation = _crate_universe_crate.annotation(**annotation)
+                annotations.setdefault(name, [])
+                annotations[name].append(annotation)
+
+    return annotations
+
+def _generate_hub_and_spokes(module_ctx, cargo_bazel, mod_path, cfg):
+    annotations = _generate_annotations(module_ctx, cfg.annotation_files)
```
illicitonion, line 39: Do you think it makes more sense to do this processing of annotations in starlark here, or to just pass the JSON files to cargo-bazel and let it deal with them?

It feels like the starlark approach basically gives us stardoc at the expense of needing to duplicate the definitions across two languages and keep them in sync? And starlark gives us potentially earlier errors, but also potentially less control over the exact contents of the errors?

crate_universe/extension.bzl
```diff
@@ -0,0 +1,224 @@
...
+    tag_path = mod_path
+    repo_name = "crates"
+    if cfg.suffix:
+        tag_path = mod_path.get_child(cfg.suffix)
+        repo_name += "_" + cfg.suffix
+
+    cargo_lockfile = module_ctx.path(cfg.cargo_lockfile)
+
+    rendering_config = json.decode(render_config(
+        regen_command = "Run 'cargo update [--workspace]'",
```
illicitonion, line 49: What's the story with this regen_command? I don't think`cargo update` is generally something we should be recommending people run?

crate_universe/extension.bzl
```diff
@@ -0,0 +1,224 @@
...
+            generate_binaries = cfg.generate_binaries,
+            rendering_config = rendering_config,
+        ),
+    )
+
+    manifests = {module_ctx.path(m): m for m in cfg.manifests}
+    splicing_manifest = tag_path.get_child("splicing_manifest.json")
+    module_ctx.file(
+        splicing_manifest,
+        executable = True,
```
illicitonion, line 74: This is a json file, right?

```suggestion
        executable = False,
```

crate_universe/extension.bzl
```diff
@@ -0,0 +1,224 @@
...
+    ])
+
+    # Create a lockfile, since we need to parse it to generate spoke
+    # repos.
+    lockfile_path = tag_path.get_child("lockfile.json")
+    module_ctx.file(lockfile_path, "")
+
+    # cargo-bazel generate takes the lockfile as input, but also writes
+    # to the lockfile. This means that even though nothing changes, the
+    # modified timestamp of the file is updated. Since the lock file is
```
illicitonion, line 104: https://github.com/bazelbuild/rules_rust/pull/2386 should hopefully allow removing this?

crate_universe/extension.bzl
```diff
@@ -0,0 +1,224 @@
...
+        cargo_lockfile_copy,
+        "--config",
+        config_file,
+        "--splicing-manifest",
+        splicing_manifest,
+        "--repository-dir",
+        tag_path,
+        "--metadata",
+        splicing_output_dir.get_child("metadata.json"),
+        "--repin",
```
illicitonion, line 126: This looks like it unconditionally repins? This is quite different from how the non-bzlmod logic works (https://github.com/bazelbuild/rules_rust/blob/86379418f04b0144cbf3959641ac89797a7c9a10/crate_universe/private/crates_repository.bzl#L44-L86)

What's the story with the difference here? Is this an intermediate step to making pinning conditional, or is there some reason bzlmod needs to do pinning differently?

crate_universe/extension.bzl
```diff
@@ -0,0 +1,224 @@
...
+        if "Http" in repo:
+            # Replicates functionality in repo_http.j2.
+            repo = repo["Http"]
+            http_archive(
+                name = crate_repo_name,
+                patch_args = repo.get("patch_args", None),
+                patch_tool = repo.get("patch_tool", None),
+                patches = repo.get("patches", None),
+                remote_patch_strip = 1,
+                sha256 = repo["sha256"],
```
illicitonion, line 165: This is an `Option` in the rust code

```suggestion
                sha256 = repo.get("sha256", None),
```

illicitonion (commented):
This generally looks good, thanks!

I've left comments for a handful of design discussions, let's work those out an rebase, and hopefully we can get this merged soon.

Thanks again for all the work you've been putting in here!

examples/bzlmod/hello_world/MODULE.bazel
```diff
@@ -17,3 +17,17 @@ use_repo(
...
+    "crate",
+)
+crate.from_cargo(
+    # This is optional, but if you need crate.annotation, then this is the
+    # bzlmod equivalent.
+    annotation_files = ["//:annotations.json"],
+    cargo_lockfile = "//:Cargo.lock",
+    manifests = ["//:Cargo.toml"],
+)
+use_repo(crate, "crates")
```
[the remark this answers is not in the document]
matts1: Not quite.

```
crate = use_extension("@rules_rust//rust:extensions.bzl", "crate")
crate.from_cargo(
    cargo_lockfile = "//foo:Cargo.lock",
    manifests = ["//foo:Cargo.toml"],
    suffix = "foo",
)
crate.from_cargo(
    cargo_lockfile = "//bar:Cargo.lock",
    manifests = ["//bar:Cargo.toml"],
    suffix = "bar",
)
use_repo(
    crate,
    "crates_foo",
    "crates_bar",
)
```

examples/bzlmod/hello_world/MODULE.bazel
```diff
@@ -17,3 +17,17 @@ use_repo(
...
+
+# Use a toml file to generate your dependencies like so:
+crate = use_extension(
+    "@rules_rust//crate_universe:extension.bzl",
+    "crate",
+)
+crate.from_cargo(
+    # This is optional, but if you need crate.annotation, then this is the
+    # bzlmod equivalent.
+    annotation_files = ["//:annotations.json"],
```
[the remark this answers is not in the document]
matts1: I'll just remove annotations from this PR, since it's not actually required.

crate_universe/extension.bzl
matts1: I think the easiest way to explain this is the different workflow.

If I wanted to update `anyhow` from `1.0.77` to `1.0.78`, the workflow here is:
1. Update the entry in `Cargo.toml` from `1.0.77` to `1.0.78`
2. Run `cargo update --workspace` to update `Cargo.lock`

Since everything is transient and automatically generated, you can then just build and it works. I'm not personally a big fan of the whole `CARGO_BAZEL_REPIN` method, but if people do care about it, I think it's fine to add it in a follow-up PR, but it definitely isn't needed for the MVP.

Note that if you try and build between step 1 and 2 you'll get an error saying that we need to update the lockfile, but that `--locked` was passed, so it'll refuse to compile.

crate_universe/extension.bzl
matts1: Removing annotations from this PR, will do them in a follow-up CL to minimize size of this PR

crate_universe/extension.bzl
illicitonion: > I've made it so that the Cargo.lock is the source of truth. The bazel index is based on the Cargo.lock, and is automatically kept up to date.

I'm definitely excited by the idea of being able to get rid of the cargo-bazel lockfile, but I'm pretty sure there were reasons we hadn't already done this... @UebelAndre do you remember?

Some possible things that come to mind, though I'm not 100% confident in any of them:
* The `crate.spec` (via extra crates a versions) and `crate.annotation` (via features) fields can change the resolution, which means if you're not going via our logic you'll miss those parts of the process when you `cargo update`.
* Splicing is expensive (it involves doing a whole resolve, even if it's offline, as well as potentially running `cargo tree` many times to resolve features for platforms) - the cargo-bazel lockfile includes the results of splicing (which is why we can skip it entirely if not repinning), whereas Cargo.lock doesn't so will necessitate re-doing this work. _Possibly_ `Module.bazel.lock` mitigates this as long as it's checked in? But if so, can we get rid of the cargo-bazel lock completely in bzlmod mode?

> There's no need to repin, because repinning can be done via "cargo update".

Even if it's literally just proxying to a call to `cargo update` via the fetched local toolchain, it's important that we support repinning via bazel - we have users who don't themselves have `cargo` installed at all but who may need to trigger a repin.

illicitonion (approved):
I'm going to merge this, it's a useful stand-alone step on our journey - we can discuss what the next steps are :) (and shout if you want to try to arrange a call at any point, I'm sure we can make time zones work if needed :))

PR #1911, opened 2023-03-30
Fix optional deps platforms

gferon:


gferon:
@UebelAndre / @illicitonion I believe this is ready to review/ship

UebelAndre (approved):
Thanks for fixing this!

PR #1912, opened 2023-03-31
cc_common_link: also respect --custom_malloc if set

durin42:
Without this change, we implicitly would link against some default malloc all the time instead of whatever user-specified custom malloc was preferred.


illicitonion (approved):
Thanks!

PR #1914, opened 2023-04-04
add Cortex-M to supported target lists.

Codetector1374:
These targets are listed as T2 supported target by rust.


rust/platform/platform.bzl
```diff
@@ -19,6 +19,9 @@ _SUPPORTED_CPU_ARCH = [
     "x86_64",
     "riscv32",
     "riscv64",
+    "thumbv6m",
```
UebelAndre, line 22: Are there actually platform definitions for this in https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD ?

I feel like this needs to be mapped to something or a PR to that repo needs to be added, right?

UebelAndre (changes requested):
Thanks for the PR! Just one question.

rust/platform/platform.bzl
Codetector1374: Actually good point. I don't think these are really needed for this change anyways.

Codetector1374:
Also can you give some advice on why the CI is failing? Do you know why / if / how should I repin the lock file?

PR #1915, opened 2023-04-04
Handle corner case for windows architecture detection

FaBrand:
It appears that `wmic os get osarchitecture` can also return

```
OSArchitecture
64-Bit
```

Note the capital `B`!

This fails for the string comparison and therefore fails to detect the existing architecture for windows.

Unfortunately i could not find any documentation on why or when this happens, but a simple `lower()` handles both cases just fine

Closes: https://github.com/bazelbuild/rules_rust/issues/1916


UebelAndre (approved):
Thanks!

PR #1920, opened 2023-04-11
Make BuildInfo provider public

### LINKED ISSUE #1578: Add `BuildInfo` provider to rules_rust public API
This ticket expands on the thread started here: https://github.com/bazelbuild/rules_rust/pull/1537#issuecomment-1256297189

There have been a couple pull-requests around [cargo_build_script](https://github.com/bazelbuild/rules_rust/blob/0.10.0/cargo/cargo_build_script.bzl#L311-L435) which increase it's footprint the core Rust rules. I don't think the Bazel rules should become too entangled with the concept of [Cargo build scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html) but I do think it would make sense to allow users to define their own rules and like `cargo_build_script` and [`cargo_dep_env`](https://github.com/bazelbuild/rules_rust/blob/0.10.0/cargo/cargo_build_script.bzl#L474-L498) to wire up regular `rustc` flags. [BuildInfo](https://github.com/bazelbuild/rules_rust/blob/0.10.0/rust/private/providers.bzl#L59-L69) seems to be a sufficient interface for this and I think it'd be better to try and understand what the provider or core rules lack to keep them "Cargo free" (to clarify, Cargo is Rust's primary build system so i'm of the opinion that Bazel should match the behaviors of Cargo within reason without compromising on the premise that the Bazel rules are primarily focused on invoking `rustc` in a similar way as Cargo). 

One could also think of this issue as "How to remove `# buildifier: disable=bzl-visibility` from [cargo_build_script.bzl](https://github.com/bazelbuild/rules_rust/blob/0.10.0/cargo/cargo_build_script.bzl#L7-L14) as I think there'd be a ton more freedom in making the rules defined there work more correctly if improving them didn't require updating any core Rust rules.

Related Issues/Pull-Requests:
- https://github.com/bazelbuild/rules_rust/pull/1551
- https://github.com/bazelbuild/rules_rust/pull/1537

illicitonion:
Fixes #1578


abrams27:
tests are failing, can we help somehow to investigate this?

scentini (approved):

PR #1922, opened 2023-04-12
Exclude target directory from release tars

illicitonion:
Right now our release tars are ~40MB and contain about 125MB of compiled rust output which shouldn't be used - the release actually instructs Bazel to download the compiled executables from elsewhere.


UebelAndre (approved):
Oops!! Thanks!

Issue #515, opened 2020-11-30
Analog to cargo profiles in rules_rust


djmarcin:
Currently it seems that the compilation mode is used to drive the opt_level and debuginfo settings at the toolchain level. This makes it extremely difficult to override settings for more than one crate at a time. For example, if I wanted to override `-c opt` to use `"s"` instead of `3`, I would need to define an entirely new toolchain and figure out how to get my binaries to use it instead of the default toolchain.

This is substantially more difficult than cargo profiles. It would be nice if a profile-like option was exposed in the repository rules to allow the option to more easily change these options that apply across many crates at once. It would be ideal if such an option was able to be toggled by individual targets rather than globally. For example, we may want to have different optimization settings for code shipped to client devices vs cloud servers without needing to define custom toolchains for one or the other.

I'm not sure after skimming the bazel toolchain docs if this is really possible without defining multiple toolchains, but if not, it would be ideal if rules_rust could make this simpler with a helper rule to define additional toolchains by passing it a profile-like dict.

Looking for feedback on this idea / has it been thought about before? If so, are there ideas about how it might work?

mattbonnell:
Late to the party. 

I'm using rules_bazel in "direct packages"/Cargo.toml-free. We run a polyglot monorepo so being able to use pure Bazel and avoid bringing in language-specific build config formats is a big usability win for us.

That said, the ability to configure compilation profiles that Cargo.toml provides is noticeably absent in this setup. Curious how other folks in the community are working around this and what would need to be done to expose this sort of functionality in rules_rust. I have time and motivation to contribute in this area.

realtimetodie:
@UebelAndre

PR #1925, opened 2023-04-18
toolchain: Add config settings for debug info and opt level

ted-logan:
Add two build flags:

- //rust/settings:opt_level
- //rust/settings:debug_info

Each of these flags takes a string list that allows the rustc
optimization level (`--codegen=opt-level=x`) and debug info
(`--codegen=debuginfo=x`) to be changed.

If either flag has one value, it will be used for all three compilation
modes (dbg, fastbuild, opt). If the flag has three values (separated by
commas) each value will be used in sequence for dbg, fastbuild, and opt.

Note that if one is changing the opt-level, one should consider changing
debug-assertions (via a separate extra rustc flag). By default, in
rustc, debug-assertions are enabled only if opt-level is 0.

Add tests to verify that the expected codegen flags are provided on the
rustc command line.


scentini:
Could this PR be replaced by a
`.bazelrc`
```
build:opt -Copt-level=X
build:opt -Cdebuginfo=Y

build:dbg -Copt-level=X
build:dbg -Copt-level=Y

...
```
?

This PR seems to tie in to wanting easier way to customize flags, which was requested in https://github.com/bazelbuild/rules_rust/issues/515. I would rather look at a more general mechanism than individual flags.

UebelAndre:
> Could this PR be replaced by a `.bazelrc`
> 
> ```
> build:opt -Copt-level=X
> build:opt -Cdebuginfo=Y
> 
> build:dbg -Copt-level=X
> build:dbg -Copt-level=Y
> 
> ...
> ```
> 
> ?
> 
> This PR seems to tie in to wanting easier way to customize flags, which was requested in #515. I would rather look at a more general mechanism than individual flags.

Is `build:dbg` and `build:opt` automatically picked up with the `--compilation_mode` flag? Or follow transitions? It'd be pretty sweet if it did!

scentini:
Not really; but the other way around works, eg one could have a `--config=opt` and put `build:opt -c opt`. As for following transitions - there is no difference whether `--@rules_rust//rust:extra_rustc_flags=-Copt-level=X` is set on the command line or via the `.bazelrc` with `build:opt --@rules_rust//rust:extra_rustc_flags=-Copt-level=X`.

ted-logan:
I can use `--@rules_rust//:extra_rustc_flags=` to add flags to the rustc command line, but they get appended at the end of argv and don't replace the `--codegen=opt-level=x` added by rust/private/rustc.bzl. I don't see clear documentation from rustc about how it handles multiple instances of the same argument with different values.

This does seem to work in practice, so I'll probably use it instead, though it's still undocumented so it feels like undefined behavior in the complier.

PR #1934, opened 2023-04-24
Support for `no_std` mode

scentini:
Initial support for `no_std` mode.
This allows us to
1. Don't pass the whole standard library to compile actions that specify `no_std`
2. Conditionally select `crate_features` and `deps` based on whether `no_std` mode is used.
Currently the only supported modes are `off` and `alloc`, with a possibility to expand in the future.

The `no_std` support comes with the following caveats:
1. Targets in `exec` mode are still built with `std`; the logic here being that if a device has enough space to run bazel and rustc, std's presence would not be a problem. This also saves some additional transitions on `proc_macro`s (they need `std`), as they are built in `exec` mode.
2. Tests are still built with `std`, as `libtest` depends on `libstd`

There is quite an ugly hack to make us be able to `select` on the `no_std` flavor taking `exec` into account; I'm looking forward to the day where Bazel will expose better ways to inspect the cfg.

There is also one part I didn't manage to make work - having a `rust_test` that tests the `rust_shared_library` in `cc_common.link` mode; I got a link error for missing `__rg_alloc` & co. symbols, which should be present as we pass `--@rules_rust//rust/settings:experimental_use_global_allocator=True`. Unfortunately I could only spot this error on CI, and could not reproduce locally. I removed the test because the `rust_shared_library` is already tested via a `cc_test`. I will however give another shot at inspecting how my local setup differs from CI.

The `rust_binary` source code in `main.rs` was borrowed from https://github.com/jfarrell468/no_std_examples, big thanks to @jfarrell468 for letting me use it.


test/no_std/main.rs
```diff
@@ -0,0 +1,56 @@
...
+            lateout("r11") _,
+            options(nostack, preserves_flags)
+        );
+    }
+    r0
+}
+
+#[no_mangle]
+unsafe fn start_main(_stack_top: *const u8) -> ! {
+    write(1, "Hello, world\n".as_bytes());
```
jfarrell468, line 51: Do you need this? Maybe get rid of the output, along with the write() function, and just call exit()

test/no_std/WORKSPACE.bazel
```diff
@@ -0,0 +1,54 @@
...
+    # See load satement for `cargo_bazel_bootstrap`.
+    generator = "@cargo_bazel_bootstrap//:cargo-bazel",
+    lockfile = "//:cargo-bazel-lock.json",
+    packages = {
+        "libc_alloc": crate.spec(
+            version = "1.0.3",
+        ),
+    },
+    # Should match the version represented by the currently registered `rust_toolchain`.
+    rust_version = "nightly/2023-04-15",
```
UebelAndre, line 46: if this version isn't in https://github.com/bazelbuild/rules_rust/blob/0.21.1/util/fetch_shas_NIGHTLY_ISO_DATES.txt then this will be a slow fetch every time. I'd suggest using 
```suggestion
    rust_version = "nightly/2023-04-20",
```

test/no_std/WORKSPACE.bazel
```diff
@@ -0,0 +1,54 @@
...
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")
+
+rules_rust_dependencies()
+
+rust_register_toolchains(
+    allocator_library = "@test_cc_common_link//:allocator_library",
+    edition = "2021",
+    global_allocator_library = "@test_cc_common_link//:global_allocator_library",
+    versions = [
+        "nightly/2023-04-16",
```
UebelAndre, line 22: This should be in sync with `crates_repository`, and again, a version found in https://github.com/bazelbuild/rules_rust/blob/0.21.1/util/fetch_shas_NIGHTLY_ISO_DATES.txt should be used.

test/no_std/WORKSPACE.bazel
scentini: Thanks, I just picked a random date :)

test/no_std/main.rs
scentini: Not really, I just wanted to see it work. Done.

BUILD.bazel
```diff
@@ -107,3 +121,30 @@ capture_clippy_output(
...
 )
+
+# This setting may be used to enable builds without the standard library.
+# Currently only no_std + alloc is supported, which can be enabled with setting the value to "alloc".
+# In the future we could add support for additional modes, e.g "core", "alloc,collections".
+string_flag(
+    name = "no_std",
+    build_setting_default = "off",
+    values = [
+        "alloc",
```
kupiakos, line 132: `no_std` should not be enabling `alloc`! This should be called `no_std_with_alloc`.

kupiakos:
@scentini do you foresee any issues with us adding support for true `no_std` code with this design? As in, with `core` only. Many embedded projects cannot have a global allocator, and so we'll need to add this support one way or another. NGL [this](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L267) was horrendous to run into for something claiming to support `no_std`.

When you write `#![no_std]`, `alloc` is not included - the code has to explicitly write `extern crate alloc`. As such, we should _never_ call `no_std` + `alloc` "no_std" and confuse users into thinking there's actual no_std support. Treating "no_std" as "has alloc" is a poor assumption to make. Can we rename the flags to match this for the next breaking change?

BUILD.bazel
scentini: The flag name itself is `no_std` while the available values currently are `off` and `alloc`, which results in
`--@rules_rust//rust/settings:no_std=off` or `--@rules_rust//rust/settings:no_std=alloc`, which IMO makes it clear when we use alloc.  Do you find `--@rules_rust//rust/settings:no_std=no_std_with_alloc` better?

scentini:
@kupiakos happy to take suggestions. The goal here was to eventually have
```
--@rules_rust//rust/settings:no_std=off
--@rules_rust//rust/settings:no_std=core|pure
--@rules_rust//rust/settings:no_std=alloc
--@rules_rust//rust/settings:no_std=collections
```
as https://docs.rust-embedded.org/book/intro/no-std.html#overview explicitly points to `alloc` and `collections` as a special `no_std` flavor. Then again, I don't do embedded development; renaming the flag values is on the table if other vocabulary is better suited.

> When you write #![no_std], alloc is not included - the code has to explicitly write extern crate alloc

True, but when thinking about remote execution, and building the standard library on demand, one needs to decide whether to build `alloc` or whether to include `liballoc` as an input to the compile actions. In addition, we support a mode where we invoke the linker directly instead of linking via `rustc` (we use this mode in Google's internal codebase), and there using/not using `alloc` makes a difference, in that we need to manually provide some allocator symbols that `rustc` otherwise generates at link time.

> and so we'll need to add this support one way or another. NGL [this](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L267) was horrendous to run into for something claiming to support no_std.

I'm currious, which flags did you pass on the command line in order to get to [this failure](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L267)?

PR #1937, opened 2023-04-25
Allow dotted target names.

ssbr:
Targets like `foo.bar` are often used for generated targets in my experience. For example, `my_macro(name="foo")` might generate a target named `foo.something`.

`name_to_crate_name` claims to convert "all illegal characters" to underscores, but dots are the most notable exception so far.


scentini:
Thanks for the PR! Could you PTAL at the failing tests?

PR #1948, opened 2023-04-29
Exclude .tmp_git_root from globs

### LINKED ISSUE #1927: Cargo crates generated from git repositories are not reproducible
If you have a crate loaded from git in your Cargo.toml, this gets loaded by bazel via git_repository. This leaves the .git directory  [either in the root or in `.tmp_git_root`](https://github.com/bazelbuild/bazel/blob/0d672f3495036ca9e35a8fcd821fd4df6461b7dc/tools/build_defs/repo/git.bzl#L31-L36). The .git directory contains files where the contents is not a function of the input attributes to `git_repository`, so technically this repository rule is not reproducible (see https://github.com/bazelbuild/bazel/issues/18152). Rules_rust [generates globs that depend on these files](https://github.com/bazelbuild/rules_rust/blob/d787525049e4cbaadabb2f531fe86c871a6cf505/examples/crate_universe/vendor_external/crates/BUILD.heck-0.4.0.bazel#L23-L32), making these rules unnecessarily re-run when the repository rule is re-run.

We should probably exclude `.git` from these globs.

cameron-martin:
This directory contains the original git repo when the crate is from git. Including this directory currently makes these rules re-run whenever the repository rule re-runs, although this is fixed by https://github.com/bazelbuild/bazel/pull/18271. Even after this fix, excluding this directory avoids depending on unnecessary files.

Fixes #1927.


cameron-martin:
The only actual change is in `crate_universe/src/rendering.rs` and the other changes are re-generated the vendored BUILD files.

UebelAndre (changes requested):
Thanks! Can you separate this into two different commits? I also think if you retry the job it will succeed.

cameron-martin:
I've split this up, although not sure why the CI is failing. Looks like its trying to build windows targets on linux.

UebelAndre:
@scentini @krasimirgg The build failures here doesn't make much sense. What is explicitly requesting the following target in this change?
```
(15:32:19) ERROR: Target //test/linker_inputs_propagation:depends_on_foo_via_sharedlib is incompatible and cannot be built, but was explicitly requested.
```

scentini:
Please don't merge on green CI as it's not really green, if you look at https://buildkite.com/bazel/rules-rust-rustlang/builds/8831#0188d87a-2695-4598-a8ad-3c8d9581f3e4 the task crashed.

scentini:
I'm trying to figure out what's going on here.

scentini:
Current CI run was correct in skipping the incompatible targets; seems like it's coverage that causes issues.

scentini:
Looks like previous CI run was with Bazel 6.1.2 while my current is with 6.2.1. Release notes for 6.2.0 include changes related with `target_compatible_with`, so I'm fine assuming that this was fixed on the Bazel side.
Seems like @UebelAndre has already raised the issue wrt BEP trying to upload missing `.dat` files over at https://github.com/bazelbuild/bazel/issues/17001. Let me know if the review is still blocked on me.

cameron-martin:
@UebelAndre looks like CI has passed now, and I've regenerated the vendored files.

UebelAndre (approved):
Thank you for fixing this!

Issue #1939, opened 2023-04-25
Crates Universe: The optional-only dependency is not fetched even if needed


golovasteek:
I’m trying to build a rust binary, that has some optional dependencies in the Cargo.toml, for my configuration I want to use these optional dependencies.
But it looks like the optional dependencies are not even fetched, and not added to the “@crate_index” repository.
That means that even if I specify them explicitly in the rust_binar binary:
```
rust_binary(
    name = "gui",
    srcs = glob(["src/**/*.rs"]),
    deps = all_crate_deps() + [
        "@crate_index//:astc-decode",
    ],
)
```
they can not be used:
```
ERROR: /Users/evgenypetrov/work/bazel_playground/rust/gui/BUILD.bazel:5:12: no such target '@crate_index//:astc-decode': target 'astc-decode' not declared in package '' defined by /private/var/tmp/_bazel_evgenypetrov/0e70737357a2a67453272d7916c8876e/external/crate_index/BUILD.bazel (Tip: use `query "@crate_index//:*"` to see all the targets in that package) and referenced by '//rust/gui:gui'
ERROR: Analysis of target '//rust/gui:gui' failed; build aborted: Analysis failed
```

Here is the repo, that reproduces the problem

https://github.com/golovasteek/bazel_playground/blob/6f2ade625c80a73a0fcdfc8f936cd46963674454/rust/gui/BUILD.bazel

golovasteek:
Yes. It still occurs on 0.21.1. I've update the example workspace with new release

https://github.com/golovasteek/bazel_playground/tree/main

golovasteek:
If add a dummy cargo crate, that depends on "astc-decode" unconditionally, then I can use it in the initial package as well.

Here is the example change

https://github.com/golovasteek/bazel_playground/pull/1/files

gferon:
I think it kind of works as designed, because the crate is marked as optional, and by default the `gui` crate is being used with default features only, you'd need to add a `crate.annotation` to enable the `sdl2` feature on the `gui` crate manually in your `WORKSPACE`.

See https://github.com/golovasteek/bazel_playground/pull/2 for a working example.

@UebelAndre maybe we could try to explain this in the documentation somewhere? I don't exactly know what would be the best place for that.

UebelAndre:
[...]

PR #1949, opened 2023-04-29
Crate Universe: enable all features when generating metadata

golovasteek:
Closes https://github.com/bazelbuild/rules_rust/issues/1939

Enabling "all features" in the metadata-generateion, will add all the dependencies to metadata. Later this metadata is used to create bazel packages and fetch dependencies.

this fixes the my problem, described in https://github.com/bazelbuild/rules_rust/issues/1939

Tests seem to pass.

Open questions:
 - where is a good place to add test for this functionality?
 - are there undesired side-effects of this change? If yes probably we need tests for them as well?

### DIFF
--- a/crate_universe/src/metadata.rs
+++ b/crate_universe/src/metadata.rs
@@ -77,6 +77,7 @@ impl MetadataGenerator for Generator {
         let metadata = self
             .cargo_bin
             .metadata_command()?
+            .features(cargo_metadata::CargoOpt::AllFeatures)
             .current_dir(manifest_dir)
             .manifest_path(manifest_path.as_ref())
             .other_options(["--locked".to_owned()])


UebelAndre (changes requested):

UebelAndre:
Thanks for working on this! Got a question for you 😄

crate_universe/src/metadata.rs
```diff
@@ -77,6 +77,7 @@ impl MetadataGenerator for Generator {
         let metadata = self
             .cargo_bin
             .metadata_command()?
+            .features(cargo_metadata::CargoOpt::AllFeatures)
```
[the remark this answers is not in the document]
golovasteek: If I understand correctly this chapter https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies all optional dependencies either belong to some explicitly defined feature, or create an implicit feature (with the same name as dependency).

And the way to enable all dependencies is to enable all the features.

Do you see it the same way? Or I maybe I misinterpret the documentation?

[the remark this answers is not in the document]
golovasteek: To my understanding, it only enables all features in the workspace, but not in the external dependencies.

I've checked with [example from this repo](https://github.com/bazelbuild/rules_rust/blob/main/crate_universe/test_data/metadata/crate_optional_deps_disabled/Cargo.toml) .
Generating metadata with `--all-features` flag doesn't pull the optional dependencies for clap.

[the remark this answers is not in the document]
UebelAndre: Awesome! Thanks for checking!

crate_universe/src/context/crate_context.rs
```diff
@@ -142,6 +142,9 @@ pub struct CommonAttributes {
     #[serde(skip_serializing_if = "SelectList::is_empty")]
     pub proc_macro_deps_dev: SelectList<CrateDependency>,
 
+    #[serde(skip_serializing_if = "SelectList::is_empty")]
+    pub all_optional_deps: SelectList<CrateDependency>,
```
UebelAndre, line 146: Instead of tracking `all_optional_deps`, can you either update `Dependency` to have an `optional` field that we can use when gathering workspace member deps, or can you track each kind of optional dependency separately?

UebelAndre (changes requested):
I think CI should be fixed now. Would you also be able to add a unit test for this? Something that shows optional dependencies can be correctly identified from some metadata?

examples/crate_universe/cargo_remote/BUILD.names.bazel
```diff
@@ -27,6 +27,7 @@ rust_binary(
     version = "0.12.0",
     deps = all_crate_deps(normal = True) + [
         ":names",
+        "@crate_index_cargo_remote//:clap",  # Clap is optional dependency and has to be added manually
```
golovasteek, line 30: I'm not sure at the moment, what this example is testing.

If I understand correctly, the `clap` crate is an optional dependency of the `names` crate. And in this case it is not supposed to be enabled by default.

Therefore we need to specify it manually.

golovasteek: The issue here is that `clap` is an optional dependency. But it is enabled by `default` feature.

And prior to my changes it was resolved and added to dependencies set.

After my changes, the dependency is resolved, but it is filtered out on the later stage, when we remove all optional dependencies.

I assume that the "defalut" features should be working out of the box. I'm looking into how to fix it.

crate_universe/src/metadata.rs
```diff
@@ -715,4 +730,27 @@ mod test {
             ])
         );
     }
+
+    #[test]
+    fn generate_metadata_for_crate_with_optional_deps() {
```
golovasteek, line 735: @UebelAndre , I'm trying to add test here. The idea of the test is to ensure that `MetadataGenerator` adds optional dependencies into the `metadata.json` and correspondingly we have them in `CargoMetadata`.

Is it a right test suite to add it?
  * Other unit-tests do not execute `cargo`, and `rustc` here, so the test is more suitable for integration-tests.
  * on the other hand, the whole metadata module is private and is not visible to the integration test.

What do you think?

PR #1956, opened 2023-05-04
Add T2 support for thumbv7em-none-eabi

cbrune:
We have been using this patch internally for several months without
any issues.

Signed-off-by: Curt Brune <curt@enfabrica.net>


cbrune:
Missing CLA

PR #1957, opened 2023-05-05
Add T2 support for thumbv7em-none-eabi

UebelAndre:


illicitonion (approved):

cbrune:
Thanks for adding our patch.  It turns out we (enfrabica.net) do have the CLA in place, but I had not pushed the right buttons on my end.  Now that is all working.  Cheers!

PR #1958, opened 2023-05-05
fix: lld-link (MSVC) fix flags including `-l` prefix

rickvanprim:
Fixes `rustc` passing a verbatim `-l<library name>` to `lld-link` (or `link.exe`) instead of just passing the library name.


rickvanprim:
Just to make sure this isn't a discrepancy between `lld-link` and `link.exe`, I just tested:
```
> link.exe -lfoo.lib
Microsoft (R) Incremental Linker Version 14.35.32217.1
Copyright (C) Microsoft Corporation.  All rights reserved.

LINK : warning LNK4044: unrecognized option '/lfoo.lib'; ignored
LINK : warning LNK4001: no object files specified; libraries used
LINK : warning LNK4068: /MACHINE not specified; defaulting to X64
LINK : fatal error LNK1561: entry point must be defined
```

UebelAndre (changes requested):
Thanks so much for working on this! 

Do you know if this correct for both `msvc` and `gnu` windows platforms?

rickvanprim:
Good catch @UebelAndre.  You're right this is wrong for the `gnu` version (just checked locally and it wants to invoke `x86_64-w64-mingw32-gcc`).  I'm thinking that instead of checking OS, checking the target triple for the ABI `msvc` might be a better approach.  I'm not sure how I'd get that in `native_deps_test.bzl`, so I made the tests use the prefix+suffix match so look for `-Clink-arg=` + `<libname>`.  Seems like the best way to detect it would be if `@platforms` had an OS entry for `mingw`, but that's a much larger scope 😅

While I was working on this, I noticed that a similar incompatible flag is being passed to GCC:
```starlark
if lib.alwayslink:
    ret.extend(["-C", "link-arg=/WHOLEARCHIVE:%s" % get_preferred_artifact(lib, use_pic).path])
```

Just a thought, but it might be easier to eventually structure the code around "linker flavor" than "Windows"/"Darwin"/"Default".

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -140,19 +148,27 @@ def _cdylib_has_native_dep_and_alwayslink_test_impl(ctx):
...
             "-lstatic=native_dep",
             "link-arg=/WHOLEARCHIVE:bazel-out/x64_windows-{}/bin/{}test/unit/native_deps/alwayslink.lo.lib".format(compilation_mode, workspace_prefix),
         ]
+        want_gnu = [
+            "-lstatic=native_dep",
+            "link-arg=-Wl,--whole-archive",
+            "link-arg=bazel-out/x64_windows-{}/bin/{}test/unit/native_deps/alwayslink.lo.lib".format(compilation_mode, workspace_prefix),
+            "link-arg=-Wl,--no-whole-archive",
+        ]
+        asserts.true(env, want_msvc == linker_args or want_gnu == linker_args)
```
UebelAndre, line 163: Is there a way you can know explicitly which one to test against? Doesn't fundamentally the `rust_toolchain.exec_triple` control these flags? So even though there isn't a given way to represent `msvc` vs `gnu` with the `@platforms` repo, you could still check within this function.

rust/private/rustc.bzl
```diff
@@ -1823,7 +1839,7 @@ def _add_native_link_flags(args, dep_info, linkstamp_outs, ambiguous_libs, crate
     use_pic = _should_use_pic(cc_toolchain, feature_configuration, crate_type, compilation_mode)
 
     if toolchain.os == "windows":
-        make_link_flags = _make_link_flags_windows
+        make_link_flags = _make_link_flags_windows_msvc if toolchain.target_triple.abi == "msvc" else _make_link_flags_windows_gnu
```
UebelAndre, line 1842: Does this need to match the `target_triple` or `exec_triple`? `toolchain.os`, which should be deleted, represents the `exec_triple`.

rust/private/rustc.bzl
```diff
@@ -1748,7 +1748,7 @@ def _portable_link_flags(lib, use_pic, ambiguous_libs, get_lib_name, for_windows
             return [] if for_darwin else ["-lstatic=%s" % get_lib_name(artifact)]
         return [
             "-lstatic=%s" % get_lib_name(artifact),
-            "-Clink-arg=-l%s" % (get_lib_name(artifact) if not for_windows else artifact.basename),
+            "-Clink-arg={}{}".format("-l" if not flavor_msvc else "", get_lib_name(artifact)) if not for_windows else "-Clink-arg={}".format(artifact.basename),
```
UebelAndre, line 1751: Style nit: Could you make a list that you conditionally append the right thing into? This conditional seems kinda long.

UebelAndre (changes requested):

test/unit/native_deps/native_deps_test.bzl
rickvanprim: `rust_toolchain.target_triple` controls them.  Though really it's the linker flavor that controls them, but as far as I'm aware there's not a way to successfully mix those, like saying `ld64.lld` + `x86_64-pc-windows-msvc` so it's not an important distinction.

Is there a way to get access to the toolchain instance inside of the test here?  That would be a vastly better solution 🙂

rust/private/rustc.bzl
rickvanprim: `target_triple`, see above comment.  I did not realize that `toolchain.os` represented `exec_triple`.  My setup that lead me to making this PR, I declared `os` on my custom toolchain based on target, so evidently my "fix" was wrong to start with 😅

rust/private/rustc.bzl
rickvanprim: Will do.  Is the `msvc_flavor` change okay or would it be better to curry in a reference to `toolchain` and get rid of `for_windows` and `for_darwin` in favor of `target_triple` checks?

rust/private/rustc.bzl
UebelAndre: This actually turned out to not be entirely sure. I just merged the other PR which should make the intent clearer

UebelAndre (changes requested):
Hi! Would you be willing to do a rebase of these changes? 🙏

rickvanprim:
@UebelAndre of course 🙂  I'm just looking for guidance on:
1. "Get the toolchain" (or `target_triple`, etc.) inside of the tests.
2. With `target_os` how to gate for `msvc`.  I'm assuming `target_os` governs `for_windows` and then `msvc = target_triple && target_triple.abi == "msvc"`.

UebelAndre:
> @UebelAndre of course 🙂 I'm just looking for guidance on:
> 
>     1. "Get the toolchain" (or `target_triple`, etc.) inside of the tests.
> 
>     2. With `target_os` how to gate for `msvc`.  I'm assuming `target_os` governs `for_windows` and then `msvc = target_triple && target_triple.abi == "msvc"`.

1. You can add a `_rust_toolchain` attribute to the rule that defaults to [`//rust/toolchain:current_rust_toolchain`](https://github.com/bazelbuild/rules_rust/blob/0.21.1/rust/toolchain/BUILD.bazel#L36-L38) and you can index that should match the toolchain the rules were built with.
2. I think this code is the one place that needs to be aware of both exec and target triples. My understanding is the bug you're seeing is that we need to format flags differently based on what linker will be used. I believe the `exec_triple` would represent what linker would be used. But in terms of what to link, that would still be `target_triple`. Though, there's still a chance `target_triple` is not set so there may need to be a `target_env` parameter added that represents this. Then you could detect `msvc` where needed? But if the target platform doesn't matter `exec_triple` is now mandatory and can be treated as a source of truth. Simply check `rust_toolchain.exec_triple.abi` wherever you need to know this info.

rickvanprim:
@UebelAndre in theory I've updated this as discussed.  `Crate Universe Examples on Windows` is failing for reasons I'm not sure about.  My attempts to build locally on Windows fail on `main` right now (though the PR that merged the latest commit passed on it).  Not sure if you have any insight?  I'll rebase and try again if I see any new commits to `main`.

UebelAndre:
Do you have a way to test cross compiling from, say `exec_triple = "x86_64-pc-windows-msvc"` to `target_triple = "x86_64-pc-windows-gnu"`? I'm curious if the use of `target_os` is correct. I suspect the `exec_triple` needs to be taken under consideration.

As for the build failure, looks like some infra failure? I'm not sure, if you make any new commit then it'll trigger new builds. If it doesn't pass then then I can ping some folks.

rickvanprim:
Doesn't seem like an infra thing given that it fails for me locally.  I'm not sure why Buildkite isn't showing any useful logs.  This is the error I get locally (running on `main`):
```
ERROR: C:/users/<user>/_bazel_<user>/rocexn33/external/rules_rust_rust_analyzer__anyhow-1.0.68/BUILD.bazel:51:19: Running Cargo build script anyhow failed: (Exit 101): cargo_build_script_runner.exe failed: error executing command (from target @rules_rust_rust_analyzer__anyhow-1.0.68//:anyhow_build_script) bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\cargo\cargo_build_script_runner\cargo_build_script_runner.exe ... (remaining 10 arguments skipped)
thread 'main' panicked at 'Unable to start binary: Os { code: 267, kind: NotADirectory, message: "The directory name is invalid." }', cargo/cargo_build_script_runner/lib.rs:127:41
```
The expanded command is:
```
bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\cargo\cargo_build_script_runner\cargo_build_script_runner.exe
bazel-out/x64_windows-opt-exec-2B5CBBC6/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script_.exe
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.out_dir
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.env
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.flags
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.linkflags
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.linksearchpaths
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.depenv
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.stdout.log
bazel-out/x64_windows-fastbuild/bin/external/rules_rust_util_import__memchr-2.5.0/memchr_build_script.stderr.log
```
And none of those files exist.  There is a directory under `opt-exec` with a runfiles manifest but nothing else.  Searching through the build output, there is no `memchr` listed anywhere.

Regarding testing `x86-64-pc-windows-gnu` from `x86_64-pc-windows-msvc`, I think I can do that with `bazel build --extra_toolchains=@rust_windows_x86_64__x86_64-pc-windows-gnu__stable//:toolchain //...`.  This is failing presently on `main` as well with some missing symbols trying to link `process_wrapper`.  Once the first issue is sorted out, I'll try again.

I don't suppose any of this is ringing any bells?  I can try to debug further if necessary.

UebelAndre (approved):
Awesome work! Thank you so much!

PR #1960, opened 2023-05-06
Removed `rust_toolchain.os` in favor of `rust_toolchain.exec_triple`.

UebelAndre:
`rust_toolchain.exec_triple` has been a mandatory field for years (https://github.com/bazelbuild/rules_rust/commit/f1b8fd3b30ca082a99c80941d2a358bec8a7339e) but by maintaining legacy behavior, there's now an opportunity for users to define misleading toolchains where `rust_toolchain.os` is ambiguous and/or doesn't match `rust_toolchain.exec_triple`.

Changes:
- Delete the `os` attribute from `rust_toolchain`
- Add `target_os` to the toolchain provider which is derived from the `target_triple` or `target_json` attributes.
- Replace the `os` field on the toolchain provider with `target_os` as it's clearer and more matches prior art with [Cargo custom targets](https://docs.rust-embedded.org/embedonomicon/custom-target.html#use-the-target-file).

Note that all uses of `toolchain.os` have been replaced with `toolchain.target_os` to avoid a change in behavior. `rust_toolchain.os` was near universally set to the target triple ([see repository_utils.bzl](https://github.com/bazelbuild/rules_rust/blob/0.21.1/rust/private/repository_utils.bzl#L315)) unless users defined their own toolchains without using `rules_rust` repository rules. That said, the use of `target_os` may not be correct in all it's uses today.


rust/private/rustc.bzl
```diff
@@ -1822,10 +1818,10 @@ def _add_native_link_flags(args, dep_info, linkstamp_outs, ambiguous_libs, crate
 
     use_pic = _should_use_pic(cc_toolchain, feature_configuration, crate_type, compilation_mode)
 
-    if toolchain.os == "windows":
+    if toolchain.target_os == "windows":
```
UebelAndre, line 1821: I could be wrong but I think this section has a mixed use of the "os". The formatting of the link flags would be dependent on what tools are being used (MSVC's `link.exe` vs `gcc`). This may be something to address in https://github.com/bazelbuild/rules_rust/pull/1958 but worth considering in this change as well.

rust/toolchain.bzl
```diff
@@ -559,11 +559,7 @@ def _rust_toolchain_impl(ctx):
 
     exec_triple = triple(ctx.attr.exec_triple)
 
-    exec_os = ctx.attr.os
-    if not ctx.attr.os:
-        exec_os = exec_triple.system
-
-    if not exec_os:
+    if not exec_triple.system:
         fail("No system was provided for the execution platform. Please update {}".format(
```
UebelAndre, line 563: I think this fail could probably be removed but not entirely sure.

rickvanprim:
What's the point in introducing `target_os` instead of just relying on `target_triple` (which seems more explicit and a superset of `target_os`)?

UebelAndre:
> What's the benefit to introducing `target_os` instead of just relying on `target_triple` (which seems more explicit and a superset of `target_os`)?

`target_triple` is not always set, the target platform might be represented by `target_json`. We only attempt to parse the platform info from each but it might be missing. The new field represents the result attempt and saves users from having to check more things than necessary.

krasimirgg (approved):
Looks good! Thank you!

PR #1967, opened 2023-05-12
Add T2 support for x86_64-unknown-none

### LINKED ISSUE #1263: Add `x86_64-unknown-none` as a Tier 2 target
As announced in [rust-lang/rust#95705](https://github.com/rust-lang/rust/pull/95705), `x86_64-unknown-none` is now a Tier 2 target with distributed build artefacts. It would be great to be able to use this in `extra_target_triples` in `rust_register_toolchains`, but it's currently missing from some of the lists in [`rust/platform/tiple_mappings.bzl`](https://github.com/bazelbuild/rules_rust/blob/main/rust/platform/triple_mappings.bzl).

Would it be possible to add this? I reckon it should take us a lot of the way to being able to cross-compile to `x86_64-unknown-none` natively. I'm very happy to help out with testing if that would help. Many thanks.

amari:
Closes #1263.


UebelAndre (changes requested):
Thanks! This will also require that you update the crate_universe examples. You can do so by navigating to `examples/crate_universe` and running `REPIN=1 bazel query ...`. Pass or fail, that command should update the `cargo-bazel-lock.json` files which you add to your PR.

UebelAndre (changes requested):
Oh, looks like you'll need to also update the `cargo-bazel-lock.json` files in `./test/no_std` 😅

UebelAndre (approved):
Thank you!

PR #1970, opened 2023-05-17
Improve handling of build.rs link dependencies

coffinmatician:
This PR helps clarify the difference between cargo_build_script build dependencies, and link dependencies. It separates the two, and ensures that link_deps are only for direct (non-transitive) normal dependencies of the build crate. It does not fix the issue with how crates universe detects links crates although it does add a comment explaining the issue.

### DIFF
--- a/crate_universe/3rdparty/crates/BUILD.cxx-1.0.86.bazel
+++ b/crate_universe/3rdparty/crates/BUILD.cxx-1.0.86.bazel
@@ -100,6 +100,9 @@ cargo_build_script(
         ],
     ),
     edition = "2018",
+    link_deps = [
+        "@cui__link-cplusplus-1.0.8//:link_cplusplus",
+    ],
     links = "cxxbridge1",
     rustc_flags = [
         "--cap-lints=allow",
--- a/crate_universe/3rdparty/crates/BUILD.iana-time-zone-haiku-0.1.1.bazel
+++ b/crate_universe/3rdparty/crates/BUILD.iana-time-zone-haiku-0.1.1.bazel
@@ -97,6 +97,9 @@ cargo_build_script(
         ],
     ),
     edition = "2018",
+    link_deps = [
+        "@cui__cxx-1.0.86//:cxx",
+    ],
     rustc_flags = [
         "--cap-lints=allow",
     ],


coffinmatician:
Okay, well I have attempted to fill out the CLA under Matician Inc. as far as I can tell everything should work, but for some reason the CLA is still failing. If anyone could explain what I am doing incorrectly I would appreciate that.

cargo/private/cargo_build_script.bzl
```diff
@@ -277,7 +277,16 @@ cargo_build_script = rule(
             allow_files = True,
         ),
         "deps": attr.label_list(
-            doc = "The Rust dependencies of the crate",
+            doc = "The Rust build-dependencies of the crate",
+            providers = [rust_common.dep_info],
+            cfg = "exec",
+        ),
+        "link_deps": attr.label_list(
```
illicitonion, line 284: Rather than introducing a new attribute on `cargo_build_script`, which requires manual maintenance of this information in BUILD files, can we add a new `links` field to `DepInfo` and `BuildInfo` here: https://github.com/bazelbuild/rules_rust/blob/f1b19c394f3edffa041b88be1bc371950476d1d2/rust/private/providers.bzl#L46-L70? That way we can conditionally inspect `dep_env` based on whether the crate had a `links` attribute, but the BUILD file doesn't need to know about the difference.

cargo/private/cargo_build_script.bzl
```diff
@@ -225,7 +225,7 @@ def _cargo_build_script_impl(ctx):
         streams.stderr.path,
     ])
     build_script_inputs = []
-    for dep in ctx.attr.deps:
+    for dep in ctx.attr.link_deps:
```
illicitonion, line 228: I believe we still want to include the `out_dir` stuff for _all_ deps, but only gate the `dep_env` adding based on `links`? (I think this will be simpler with a new provider field - we can just iterate across all `deps`, but conditionally do the `dep_env` stuff based on the presence of the `links` field in the provider we're inspecting)

crate_universe/src/metadata/dependency.rs
```diff
@@ -92,6 +93,11 @@ impl DependencySet {
         // https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key
         // https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages
         // https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate
+
+        // TODO: The above/below is incorrect fyi. It is actually only when any
```
illicitonion, line 97: I agree! Could I bother you to rewrite the comment and filter logic accordingly?

illicitonion (changes requested):
Thanks for putting this PR together, I agree this is a problem we should solve!

With any luck, next time you push to this branch the CLA bot will be happy, if not we can escalate to one of our Googlers to investigate :)

crate_universe/src/metadata/dependency.rs
coffinmatician: I'm not interesting in doing that as we don't use crate_universe internally at all (we only use the rules_rust part). I was just trying to not break existing functionality.

cargo/private/cargo_build_script.bzl
coffinmatician: I am confused as I don't see where I would gather this information from. `link_deps`, and `deps` are disjoint ideas (neither is a subset of the other). One comes from *build* dependencies for the host, and the other comes from *runtime* (normal) dependencies for the target which are not the same. Please let me know if I am misunderstanding you.

coffinmatician: I primarily made this change as we have link_deps that link against .so libraries that we didn't want to have added to the cargo_build_script as they aren't there in cargo and aren't needed.

cargo/private/cargo_build_script.bzl
coffinmatician: I think I may have screwed this part up. I'll take a look at it again some time soon.

coffinmatician: Okay, after double checking how cargo works, it doesn't pay attention to `cargo:rerun-if-changed` for things in the `OUT_DIR` so the only option there is to rely on `cargo:rerun-if-env-changed` which can only come from `link_deps` assuming that the cargo build.rs wants to be correct (admittedly some build.rs scripts do stupid things (such as `pyo3`, but manually applying fixes to make those work at all in Bazel is almost always required so I think that the few extra lines of code there is worth the globally reduced build times elsewhere). Therefore I think in theory we only need to make available the parts of `OUT_DIR` from a `cargo_build_script`s `link_deps` and the `OUT_DIR`s that those `link_deps` can see (as you can keep passing forward paths by repeated uses of `cargo:KEY=VALUE`). In practice this also seems to be what I am observing.

If you agree with this conclusion I will rework line 233:
`            for dep_build_info in dep[rust_common.dep_info].transitive_build_infos.to_list():`
to only be for transitive `link_deps`.

crate_universe/src/metadata/dependency.rs
coffinmatician: Actually I will see if I can quickly do this on Monday, or Tuesday, but no promises haha.

cargo/private/cargo_build_script.bzl
illicitonion: Do you have a cargo equivalent of what you're doing here? Or an example repo you can share that shows what you're trying to do?

I've been assuming you're basically implementing support for the [`links`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key) attribute in a Cargo.toml file.

The way I think cargo models this is: a build script may optionally propagate some settings to the linker to add configuration to the linker action, But a cargo build script can _also_ do other things like generate code which is needed for compiles, or generate environment variables which should be set for targets which depend on this one.

That's why I'm viewing the "link_deps" behaviour as a subset of the general "depending on a cargo_build_script" behaviour - if a cargo_build_script does both codegen (for code that needs compiling as part of depending on this cargo_build_script) and specifying link libraries, then if we're looking to closely model cargo build scripts (which is the aim of the `cargo_build_script` rule), we should expose this as optional subsets.

If you're just looking to link against an existing .so file, you should be able to just add a dependency from your `rust_library` or `rust_binary` on a [`cc_import`](https://bazel.build/reference/be/c-cpp#cc_import) target (or a [`cc_library`](https://bazel.build/reference/be/c-cpp#cc_library) if you're compiling it yourself) and shouldn't need to go via a `cargo_build_script` at all.

coffinmatician: > Do you have a cargo equivalent of what you're doing here?

Yes, but a full example is overly complicated. I have attached a simpler example below.

Here is an example of something that compiles in Bazel that doesn't in Cargo (without this PR). This is because of the lack of difference between `deps`, and `link_deps` that isn't made before this PR. 

> I've been assuming you're basically implementing support for the [links](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key) attribute in a Cargo.toml file.

I am attempting to fix some details of the `link` support which already exists but aren't implemented properly.

> The way I think cargo models this is: a build script may optionally propagate some settings to the linker to add configuration to the linker action, But a cargo build script can also do other things like generate code which is needed for compiles, or generate environment variables which should be set for targets which depend on this one.

I am aware of that already. I attempted to highlight the more subtle nuances above, but I will repeat them again. `build-dependencies` in a Cargo.toml file are associated with `cargo_build_script` `deps` while `dependencies` in a Cargo.toml file (that have a `links` attribute) are associated with `cargo_build_script` `link_deps`.

> That's why I'm viewing the "link_deps" behaviour as a subset of the general "depending on a cargo_build_script" behaviour - if a cargo_build_script does both codegen (for code that needs compiling as part of depending on this cargo_build_script) and specifying link libraries, then if we're looking to closely model cargo build scripts (which is the aim of the cargo_build_script rule), we should expose this as optional subsets.

Neither is a subset of the other although I could see how the Rust documentation could mislead you into thinking that. I initially thought that was the case at well until reading more.

> If you're just looking to link against an existing .so file, you should be able to just add a dependency from your rust_library or rust_binary on a [cc_import](https://bazel.build/reference/be/c-cpp#cc_import) target (or a [cc_library](https://bazel.build/reference/be/c-cpp#cc_library) if you're compiling it yourself) and shouldn't need to go via a cargo_build_script at all.

I am aware of that. My goal was to *not* link against something which it currently is. This is because `deps` link things while `link_deps` won't. (This was because we are cross compiling, and only have the target architecture version of a library available to the rule.)

crate_universe/src/metadata/dependency.rs
coffinmatician: Done.

cargo/private/cargo_build_script.bzl
coffinmatician: I have decided not to make this change, but just to leave a comment about it. The code was hard for me to understand that handled the current `transitive_build_infos` due to the codes indirectness, lack of comments, and lack of types. I spent an hour or so staring at the screen feeling a bit foolish before giving up. I may try again later, but I believe that making the change will just minorly improve build caching.

coffinmatician:
@illicitonion Please let me know your thoughts again, and I am happy to rebase onto `master`, and regenerate the commits that update the `crate_universe/3rdparty/crates` then. I know that Github ruins the comment history when I do that, and some people don't like that so let me know if you prefer that I merge master into my branch instead. I dislike that as it ruins the commit history, but I will defer to your preferences as the reviewer.

illicitonion:
Sounds great, thanks! I'm just getting to the end of a really busy patch, but should be able to take a look next week - thanks for your effort and patience!

test/dep_env/read_c.rs
```diff
@@ -0,0 +1,6 @@
+use std::env::var;
+
+fn main() {
+    assert!(var("DEP_X_A").is_err());
```
illicitonion, line 4: This assertion also passes if I just cherry-pick the changes to `test` onto the `main` branch without your other changes, and `s/link_deps/deps/g` in the BUILD.bazel file.

I _think_ one of the core observable pieces you're trying to fix in this PR is that if there's a _transitive_ dep on `:set_a_build`, this variable _still_ isn't set. I think you could demonstrate this by adding a `link_deps` dependency from `read_c` onto `:set_a` and a `deps` dependenecy from `read_c` onto `:set_a_build` - does that sound right to you? If so, could you add those to the BUILD.bazel file so that we have tests for this?

illicitonion (commented):
Thanks again for the contribution!

I think this makes sense and looks good.

I've made one comment about slightly broader test coverage - otherwise I'd be happy to merge this as-is - when it gets merged it's going to get squashed either way, so feel free to merge or rebase main, whichever you prefer.

test/dep_env/read_c.rs
coffinmatician: Yeah, I am aware that it passes on master too. I just added the test as I felt that it was useful (I guess it probably should go in a separate PR). You may be correct that I didn't write the test correctly (whoops), but I think even your suggested test will pass in master.

I will investigate today, or tomorrow further and confirm if what I wrote above is correct.

coffinmatician: Yeah, it was working as I intended, but I see how you are correct that I could improve the test. I have done so.

cargo/private/cargo_build_script.bzl
coffinmatician: I thought about this more, and made some examples in `cargo` showing how this whole set up is way more complicated than I initially realized. I have switched it to be more like what you asked. OUT_DIR can basically come from anywhere I realized. Build dependencies, link dependencies, regular dependencies, proc macro dependencies, etc. They only "work" in that they assume that the cargo target directory doesn't change at all during the build process which is true unless you `cargo build`, and then move the whole repository, and then `cargo build` again.

I think that this isn't actually possible to handle properly with poorly behaving cargo crates. I think I may try to contribute a cargo clippy lint for this bad behavior in the next weeks, and maybe rules_rust should consider supporting those poorly behaving crates out of scope. I'm not sure what the right decision is.

[foobar.tar.gz](https://github.com/bazelbuild/rules_rust/files/11826078/foobar.tar.gz)

illicitonion (approved):
LGTM, thanks so much!

@scentini do you have any idea what's up with CLA bot here? I'm pretty sure yesterday the check had a green tick, but it seems to have gone red on a new push...

scentini:
@illicitonion I'm looking at the history of cla and looks like it was red all along, are you sure you saw it pass?

illicitonion:
I'm pretty sure it was before a force-push, but I'm not 100% on it - @coffinmatician do you know what the story is with your CLA signing?

scentini:
@coffinmatician from https://github.com/bazelbuild/rules_rust/pull/1970#issuecomment-1552001485 I understand that this is supposed to be corporate CLA, I cannot find Matician Inc. in the list of companies with accepted corporate CLA. Docs say the process can take a few days, although it looks like a month has passed. Do you have a way to check that 1. your company has signed the CLA? and 2. you are listed as an authorized contributor?

Here's some docs on corporate CLA, in case you haven't seen them already: https://cla.developers.google.com/about#add-contributors

scentini:
@illicitonion here's the data I see:
```
 Timestamp                    Event                    Conclusion	  Contributors	    Errors
2023-06-23 04:55:47	Re-run Requested	        failure	        (coffinmatician)  :  ❌
2023-06-23 04:55:28	Re-run Requested	        failure	        (coffinmatician)  :  ❌
2023-06-23 04:51:57	Re-run Requested	        failure	        (coffinmatician)  :  ❌
2023-06-21 14:22:52	Pull Request Synchronized	failure    	(coffinmatician)  :  ❌
2023-06-21 14:13:27	Pull Request Synchronized	failure    	(coffinmatician)  :  ❌
2023-06-14 12:59:47	Pull Request Synchronized	failure	        (coffinmatician)  :  ❌
2023-05-17 13:11:47	Re-run Requested	        failure	        (coffinmatician)  :  ❌
2023-05-17 13:04:52	Re-run Requested	        failure	        (coffinmatician)  :  ❌
2023-05-17 11:17:29	Re-run Requested	        failure	        (coffinmatician)  :  ❌
2023-05-16 19:35:58	Pull Request Synchronized	failure	        (coffinmatician)  :  ❌
2023-05-16 19:24:55	Pull Request Opened	        failure	        (coffinmatician)  :  ❌
```

coffinmatician:
> I'm pretty sure it was before a force-push, but I'm not 100% on it - @coffinmatician do you know what the story is with your CLA signing?

I have never gotten the CLA signing approved by Google. I initially had our head of Operations sign and fill out the form for the CLA and put me as a developer inside of it, this was rejected for reasons that I don't understand. It then didn't seem to want to let me fill out the form with me using the email of our CEO for the signature as we had already submitted one or something. I then became distracted by other things (I forget the exact struggles I faced trying to fill out the CLA), and didn't want to make my life harder by doing anything that made Google's CLA system more angry (I have bad experiences with Google's automated response systems freaking out which is why I no longer use Google Accounts for my personal devices). If anyone can confirm that I should attempt to fill out the form again or do something else I am happy to try again.

UebelAndre:
@coffinmatician All that's needed is a rebase and I think this can be merged!

coffinmatician:
> @coffinmatician All that's needed is a rebase and I think this can be merged!

I will try to do that in the next few hours (if not then definitely tomorrow).

coffinmatician:
@scentini I also just wanted to say thank you again for helping me figure out how to submit the CLA again properly. :)

scentini:
@coffinmatician You're very welcome!

PR #1971, opened 2023-05-17
Allow specifying exec and target compatibility constraints

illicitonion:
Extra target triples may only be appropriate for specific platforms (e.g. if you register a toolchain with the exec_triple `x86_64-unknown-linux-gnu` with an extra target triple of `x86_64-unknown-linux-musl`, you may want differing `target_compatible_with` values for each of those two target triples.

This allows for specifying explicit `exec_compatible_with` lists per toolchain (which was already supported by `rust_toolchain_repository` just not proxied by `rust_repository_set`), and explicit `target_compatible_with` lists, both for the default triple and for extra ones.

I'm not 100% sure on our pre-existing assumption that every toolchain supports compiling targeting its own exec platform, i.e. the '`[exec_triple]`' part of https://github.com/bazelbuild/rules_rust/blob/937e63399b111a6d7ee53b187e4d113300b089e9/rust/repositories.bzl#L845 but it _does_ generally hold, and supplying a way of removing that default assumption is definitely a separate PR for another day if it becomes relevant to anyone.


UebelAndre (approved):
I think this seems fine but it would be nice to see an example of it somewhere to both inform users of how they can be more descriptive with their toolchains while also acting as a regression test.

illicitonion:
Thanks! I'll try to get the relevant pieces open sourced so we could have a working example :) Unfortunately toolchains are a bit fiddly in general, but hopefully this shouldn't be too hard to put together :)

PR #1975, opened 2023-05-23
fix: `crate_type` more accurately corresponds to CC linking actions

rickvanprim:


rickvanprim:
@scentini / @krasimirgg thoughts on this?  I tried a version where `staticlib`, `lib`, and `rlib` used `CPP_LINK_DYNAMIC_LIBRARY_ACTION_NAME` and it exploded horribly (trying to invoke `ar` with a bunch of arguments it didn't like).  I'm trying to fix what seems like an issue to me where my flags from `CPP_LINK_EXECUTABLE_ACTION_NAME` contain `-pie` which conflicts with `-shared`, but I could very well be setting up my `CPP_LINK_EXECUTABLE_ACTION_NAME` action wrong.

scentini:
I believe `dylib` should use [`CPP_LINK_NODEPS_DYNAMIC_LIBRARY_ACTION_NAME`](https://github.com/bazelbuild/bazel/blob/6584feaab993b2396adcec1415723ec30c8a4dc8/tools/build_defs/cc/action_names.bzl#LL68C1-L68C44), `staticlib` should use [`CPP_LINK_STATIC_LIBRARY_ACTION_NAME`](https://github.com/bazelbuild/bazel/blob/6584feaab993b2396adcec1415723ec30c8a4dc8/tools/build_defs/cc/action_names.bzl#L71), while `rlib` and `dylib` are pure compile actions so they shouldn't look into the cc linking action flags at all.

rickvanprim:
Got it passing tests with the latest iteration.  My interpretation is that `rustdoc` ultimately tries to make a `bin` and needs all the various linker args supplied to it.  This behavior is still slightly changed as the old version would potentially use the `CPP_LINK_DYNAMIC_LIBRARY_ACTION_NAME` flags for `rustdoc` on a `proc-macro`.  Aside from that, it's roughly the same as before, but `force_link` renamed to `rustdoc` and the docstring updated to hopefully make things a bit more clear.  

Bikeshed wise, I'm not sure whether `rustdoc` or `rust_doc` is the more correct parameter name, as both forms exist in the codebase 😅

UebelAndre (approved):
This change seems good to me. I'll give @scentini more time to chime in though.

rust/private/rustc.bzl
```diff
@@ -371,13 +374,23 @@ def get_linker_and_args(ctx, attr, crate_type, cc_toolchain, feature_configurati
     """
     user_link_flags = get_cc_user_link_flags(ctx)
 
-    if crate_type == "proc-macro":
+    if crate_type in ("bin") or rustdoc:
+        is_linking_dynamic_library = False
+        action_name = CPP_LINK_EXECUTABLE_ACTION_NAME
+    elif crate_type in ("cdylib", "dylib"):
```
scentini, line 380: `cdylib` actually links in all its dependencies, it should be in the same bucket as `proc-macro`, asking for  `CPP_LINK_DYNAMIC_LIBRARY_ACTION_NAME` flags .

scentini:
Thanks for the patience folks, I was out of office an am getting up to speed with `rules_rust`. Changes look good to me apart for the placement of `cdylib`.

scentini (approved):

scentini:
@rickvanprim could you rebase?

rickvanprim:
@scentini done!  Until the next rebase 😆

PR #1976, opened 2023-05-23
feature: `target_compatible_with` added to `CommonAttrs`

rickvanprim:
This should resolve the problem where a platform is targeted that isn't listed in `supported_platform_triples` (in `crates_repository()`) and no error manifests until a missing dependency (or some similar error due to an omitted branch in a `select()`) is encountered.

### DIFF
--- a/crate_universe/src/config.rs
+++ b/crate_universe/src/config.rs
@@ -508,6 +508,13 @@ pub struct Config {
     /// Whether or not to generate Cargo build scripts by default
     pub generate_build_scripts: bool,
 
+    /// A set of platform triples to use in generated select statements
+    #[serde(
+        default = "default_generate_target_compatible_with",
+        skip_serializing_if = "skip_generate_target_compatible_with"
+    )]
+    pub generate_target_compatible_with: bool,
+
     /// Additional settings to apply to generated crates
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub annotations: BTreeMap<CrateId, CrateAnnotations>,
@@ -528,6 +535,13 @@ impl Config {
         let data = fs::read_to_string(path)?;
         Ok(serde_json::from_str(&data)?)
     }
+}
+
+fn default_generate_target_compatible_with() -> bool {
+    true
+}
+fn skip_generate_target_compatible_with(value: &bool) -> bool {
+    *value
 }
 
 #[cfg(test)]
--- a/crate_universe/src/config.rs
+++ b/crate_universe/src/config.rs
@@ -38,7 +38,7 @@ impl std::fmt::Display for VendorMode {
     }
 }
 
-#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct RenderConfig {
     /// The name of the repository being rendered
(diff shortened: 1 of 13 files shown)


crate_universe/src/lockfile.rs
```diff
@@ -211,7 +211,7 @@ mod test {
 
         assert_eq!(
             digest,
-            Digest("7be4f323ac6a4d0a45d9d430a8056967eb248ca7c86bdba44af33ad90392cb4a".to_owned())
+            Digest("7b4142c24aff865f158da49840f7f23a6921c65f28999dd03fdd4bc83eb3fcb1".to_owned())
```
rickvanprim, line 214: With the updated implementation for `RenderConfig::default()` this hash needs to be updated.

crate_universe/src/config.rs
```diff
@@ -77,10 +77,26 @@ pub struct RenderConfig {
     /// The command to use for regenerating generated files.
     pub regen_command: String,
 
-    /// An optional configuration for rendirng content to be rendered into repositories.
+    /// An optional configuration for rendering content to be rendered into repositories.
     pub vendor_mode: Option<VendorMode>,
 }
 
+impl Default for RenderConfig {
```
UebelAndre, line 84: Why did this need to be explicitly implemented?

crate_universe/src/rendering.rs
```diff
@@ -28,13 +28,13 @@ use crate::utils::{self, sanitize_repository_name};
...
 pub struct Renderer {
-    config: RenderConfig,
+    config: Config,
     engine: TemplateEngine,
 }
 
 impl Renderer {
-    pub fn new(config: RenderConfig) -> Self {
-        let engine = TemplateEngine::new(&config);
+    pub fn new(config: Config) -> Self {
```
UebelAndre, line 36: Why not have this continue to take a `RenderConfig` and additionally take the list of supported platforms? This way if `Config` grows the impact is more contained?

UebelAndre (changes requested):
Thank you so much for the PR and your patience! Just to make sure I understand everything, can you post an example of the failure mode? Is this unique to `crates_vendor`?

crate_universe/src/config.rs
rickvanprim: I noticed some of the tests were explicitly deserializing a JSON literal instead of constructing a Rust literal and my interpretation is that that's because there is an unexpected behavioral difference between the "default" you get from Serde and from `Default::default()`.  So I figured it made sense to unify them by implementing `Default` to call the same default field generators that Serde does.  Probably worth me adding a comment above the `impl` if you're agree with the motivation behind the change.

crate_universe/src/rendering.rs
rickvanprim: Works for me 🙂

crate_universe/src/config.rs
UebelAndre: That sounds pretty good, yeah, I didn't know about this. A comment would be nice so someone doesn't try to delete it because they also didn't know 😄

rickvanprim:
Example error message from forgetting to list a platform in `supported_platform_triples`:
```
ERROR: Target //scripts:develop is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //scripts:develop (86847a)
    //scripts:develop_binary (144c08)
    //scripts:utilities (144c08)
    @crate_index//:anyhow (144c08)
    @crate_index__anyhow-1.0.71//:anyhow (144c08)   <-- target platform (//bazel/platforms:host) didn't satisfy constraint @platforms//:incompatible
    ```

UebelAndre:
> Example error message from forgetting to list a platform in `supported_platform_triples`:
> 
> ```
> ERROR: Target //scripts:develop is incompatible and cannot be built, but was explicitly requested.
> Dependency chain:
>     //scripts:develop (86847a)
>     //scripts:develop_binary (144c08)
>     //scripts:utilities (144c08)
>     @crate_index//:anyhow (144c08)
>     @crate_index__anyhow-1.0.71//:anyhow (144c08)   <-- target platform (//bazel/platforms:host) didn't satisfy constraint @platforms//:incompatible
>     ```
> ```

Can you show me `bazel query --output=build` for the two `@crate_index*` targets?

UebelAndre (approved):
Awesome work! Thank you so much!

PR #1979, opened 2023-05-24
fix: `rustc_compile_action` emits `CcInfo` for `wasm32` targets

rickvanprim:
This change allows `cc_binary()`, `cc_library()`, etc. targets to depend on `rust_libary()`, `rust_static_library()`, etc. targets when targeting `wasm32`.


rickvanprim:
@titanous your initial addition gated `CcInfo` on not being `wasm32`, so I'm hoping you can weigh in here on whether this is a bad idea/I'm overlooking something 😅

titanous:
I don't actually have any code using the patch from #732 right now, so I'm not certain... I don't remember why I added that condition 😰

titanous:
#627 said this:

>For the rule to succeed it is also necessary to skip the output of the CcInfo, since it won't be generated. This issue also shows up and is resolved when trying to build a rust_library with crate_type=cdylib and --platforms=@rules_rust//rust/platform:wasm.

rickvanprim:
Hmm that surprises me.  I'd expect `rust_wasm_bindgen` to emit its own set of providers and whatever `rust_library` emitted to be ignored (beyond the purposes of `bindgen` running and producing the `.wasm`).  I'll try to dig in further.

rickvanprim:
It seems like that case should be covered by the targets in `examples/wasm/BUILD.bazel` which build just fine with this change.  I'm thinking the original scenario where `rust_library()` accepted a `crate_type` may have been the culprit and it's no longer an issue with `rust_shared_library()`.  Any thoughts on other stuff to test?

rickvanprim:
@titanous any further thoughts on this or what I should do to move this forward?

UebelAndre (changes requested):
@rickvanprim @titanous I think this change is fine but could you add a unit test of sorts with some comments on the rationale for needing this? None of the maintainers really use WASM so we need to rely on the community and heavy documentation to make sure we don't break things.

examples/nix_cross_compiling/bazel/toolchain_rules/cc_tools/wasm-ld.bzl
UebelAndre, line 1: Can this and neighboring tools be renamed to snake case?

examples/nix_cross_compiling/BUILD.bazel
```diff
@@ -0,0 +1,167 @@
+load("@aspect_bazel_lib//lib:transitions.bzl", "platform_transition_binary")
```
UebelAndre, line 1: Can this instead be implemented in this example? I'd love to reduce dependencies where possible and this transition should be pretty simple.

examples/nix_cross_compiling/bazel/cargo/platforms/BUILD
UebelAndre, line 1: Can these be renamed to `BUILD.bazel` files?

UebelAndre (changes requested):
Awesome work! Sorry it took me so long to look at this 😅

Just a couple of requests

examples/nix_cross_compiling/bazel/toolchain_rules/cc_tools/wasm-ld.bzl
rickvanprim: Will do.

examples/nix_cross_compiling/BUILD.bazel
rickvanprim: I'm hoping to make (very) soon is rust_binary() (and whatever Rust dylib is, etc.) take an optional platform transition.  Given Bazel's lack of "rule wrappers", platform_transition_binary() causes a whole bunch of headaches with things like aspects, bazel test, etc.  If it's okay, I think it would be easier to ship this as is, and then remove it in a follow up PR with that change (if you're not opposed to having an optional transition on those rules).

examples/nix_cross_compiling/bazel/cargo/platforms/BUILD
rickvanprim: Will do.

examples/nix_cross_compiling/BUILD.bazel
UebelAndre: I think it'd still be better to add a `.bzl` file here
```starlark
"""Rules for working with transitions."""

load("@bazel_skylib//lib:paths.bzl", "paths")

def _transition_platform_impl(_, attr):
    return {"//command_line_option:platforms": str(attr.target_platform)}

# Transition from any input configuration to one that includes the
# --platforms command-line flag.
_transition_platform = transition(
    implementation = _transition_platform_impl,
    inputs = [],
    outputs = ["//command_line_option:platforms"],
)

def _platform_transition_binary_impl(ctx):
    # We need to forward the DefaultInfo provider from the underlying rule.
    # Unfortunately, we can't do this directly, because Bazel requires that the executable to run
    # is actually generated by this rule, so we need to symlink to it, and generate a synthetic
    # forwarding DefaultInfo.

    result = []
    binary = ctx.attr.binary[0]

    default_info = binary[DefaultInfo]
    files = default_info.files
    new_executable = None
    original_executable = default_info.files_to_run.executable
    runfiles = default_info.default_runfiles

    if not original_executable:
        fail("Cannot transition a 'binary' that is not executable")

    new_executable_name = original_executable.basename

    # In order for the symlink to have the same basename as the original
    # executable (important in the case of proto plugins), put it in a
    # subdirectory named after the label to prevent collisions.
    new_executable = ctx.actions.declare_file(paths.join(ctx.label.name, new_executable_name))
    ctx.actions.symlink(
        output = new_executable,
        target_file = original_executable,
        is_executable = True,
    )
    files = depset(direct = [new_executable], transitive = [files])
    runfiles = runfiles.merge(ctx.runfiles([new_executable]))

    result.append(
        DefaultInfo(
            files = files,
            runfiles = runfiles,
            executable = new_executable,
        ),
    )

    return result

platform_transition_binary = rule(
    doc = "Transitions the binary to use the provided platform.",
    implementation = _platform_transition_binary_impl,
    attrs = {
        "binary": attr.label(
            doc = "The target to transition.",
            allow_files = True,
            cfg = _transition_platform,
            mandatory = True,
        ),
        "target_platform": attr.label(
            doc = "The target platform to transition the binary.",
            mandatory = True,
        ),
        "_allowlist_function_transition": attr.label(
            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
        ),
    },
    executable = True,
)
```
I would add this to `examples/nix_cross_compiling/transition.bzl` and load that here instead to eliminate a dependency and keep folks from needing to jump around to other github repos when looking at this example

UebelAndre (changes requested):

examples/nix_cross_compiling/bazel/toolchain_rules/cc_tools/wasm-ld.bzl
rickvanprim: For context, the current naming matches the binaries (hence the weird `ld64.lld.bzl`).  Let me know if there's a different overall naming convention that would be preferred.

.bazelci/presubmit.yml
```diff
@@ -328,9 +328,15 @@ tasks:
     platform: ubuntu2004
     working_directory: examples
     build_targets:
+      - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
       - "//..."
+      # Nix example is only supported on NixOS container
+      - "-//nix_cross_compiling/..."
```
rickvanprim, line 334: @UebelAndre is there a better way to exclude the Nix example from all the non-NixOS platforms?

rickvanprim: Hmm looks like this doesn't even work as it still tries to load the `WORKSPACE.bazel`.

examples/nix_cross_compiling/bazel/toolchain_rules/cc_tools/wasm-ld.bzl
UebelAndre: I think that one was fine. I’m familiar with it 😄

.bazelci/presubmit.yml
UebelAndre: Update the working directory and make sure you add the nix directory to the examples .bazelignore file

.bazelci/presubmit.yml
```diff
@@ -614,6 +614,12 @@ tasks:
   #   working_directory: examples/zig_cross_compiling
   #   build_targets:
   #     - "//..."
+  nix_cross_compiling:
+    name: Nix cross compiling test
+    platform: nixos
+    working_directory: examples/nix_cross_compiling
+    build_targets:
+      - "//:nix_cross_compiling"
```
UebelAndre, line 622: why not `//...`?
You should also add `test_targets`
```suggestion
      - "//..."
    test_targets:
      - "//..."
```

UebelAndre (changes requested):

.bazelci/presubmit.yml
rickvanprim: While I was working on the next part to try to upstream the `@platforms//os:nixos` (instead of the constraint being defined in the example here), I deleted the code and everything still built just fine.  Evidently that's because if `target_compatible_with` isn't being satisfied, it gets silently excluded from the `//...` glob.  Hence explicitly enumerating them to catch if a platform goes missing.  It's definitely not ideal though, so maybe there's a better way to catch it?

image

rickvanprim: To repro on this branch, comment off:
```bzl
    # render_config = render_config(
    #     platforms_template = "@//bazel/cargo/platforms:{triple}",
    # ),
```

Then `//...` still passes presumably because the `filegroup()` gets transitively disabled:
```
> bazel build //...
INFO: Analyzed 15 targets (6 packages loaded, 9 targets configured).
INFO: Found 15 targets...
INFO: Elapsed time: 0.274s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
```
But:
```
> bazel build //:nix_cross_compiling                       
ERROR: Target //:nix_cross_compiling is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //:nix_cross_compiling (47a001)
    //:rust_binary_aarch64-linux-android (47a001)
    //rust_binary:rust_binary (101ecc)
    @crate_index//:tokio (101ecc)
    @crate_index__tokio-1.34.0//:tokio (101ecc)
    @crate_index__tokio-macros-2.2.0//:tokio_macros (6cf94f)   <-- target platform (//bazel/platforms:host) didn't satisfy constraint @platforms//:incompatible
INFO: Elapsed time: 0.867s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (24 packages loaded, 606 targets configured)
```

UebelAndre: I would recommend adding a [build_test](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/build_test_doc.md) target for all the things you want to ensure are built. But I'd rather avoid having custom settings in the presubmit definition, that way anyone can navigate to that part of the workspace and run `bazel test //...` to make sure everything works. Ideally all bazel jobs would be covered by a single `bazel test //...`.

rickvanprim: Looks like that has the same issue unfortunately 😞
```
> bazel test //... 
INFO: Analyzed 27 targets (0 packages loaded, 0 targets configured).
INFO: Found 26 targets and 1 test target...
INFO: Elapsed time: 0.249s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
//:nix_cross_compiling                                                  SKIPPED

Executed 0 out of 1 test: 1 was skipped.
```
Maybe we should revisit https://github.com/bazelbuild/rules_rust/pull/1976 to see if we can come up with a better way to indicate that the generated build configs don't support a particular target triple given subtle behavior of `target_compatible_with`?

rickvanprim:
Regarding the example and `cargo-bazel`, it looks like the other examples build `cargo-bazel` from source (`cargo_bazel_bootstrap()`) and that's what I have set up here (Nix builds it with Cargo via `buildRustPackage`), but this is fairly slow.  Is there some trick that's speeding it up in the other examples?

.bazelci/presubmit.yml
UebelAndre: Ah dang, makes sense though

nix/flake.nix
UebelAndre, line 1: Can you explain what this file is (sorry, I'm pretty unfamiliar with nix)? It'd be nice to avoid new top level packages and files.

UebelAndre (changes requested):
I think this is looking good! I just had one last question 😄

nix/flake.nix
rickvanprim: The top level `//nix` is a very recent realization on my part that lead to https://github.com/tweag/rules_nixpkgs/pull/450.  Bazel's Nix integration uses Nix as a build system (as do things like the `devShells`), where Nix works by copying the package (Nix Flake) to its immutable store and hashing it, and then building against that.  If you're actually building things via Nix, that's necessary and makes sense, but for our Bazel usage where it's just being used as a declarative package manager, that means unnecessarily copying the whole Git repository to `/nix/store` whenever it's built (which thanks to Bazel is only when it changes/on clean builds, so some mitigation).

It sounds like when https://github.com/NixOS/nix/pull/6530 lands, that won't be an issue (or less of an issue), so leaving `flake.nix` at the root of the repo is certainly an option as opposed to specifically scoping it down to minimize unnecessary copying.

Conversely, this is why I have `//crate_universe/flake.nix` instead of `//crate_universe/nix/flake.nix` because the `crate_universe` directory copied to the Nix Store for compilation (the Nix bootstrap approach for `cargo-bazel`).  If we stick with `//flake.nix`, we can merge in the contents of `//crate_universe/flake.nix` into it though I'm not sure if that will copy the entire Git repo or just the referenced `crate_universe` directory once https://github.com/NixOS/nix/pull/6530 lands.

Let me know if I can further clarify 😅

.bazelci/presubmit.yml
rickvanprim: I'm hoping in real projects that this is less of an issue.  People can still get tripped up with a `build //...` passing when it shouldn't but it would hopefully be immediately apparent when an expected artifact is suddenly missing 😅

UebelAndre:
Looks like you have some nix errors? I can take another pass when CI is green.

rickvanprim:
I think I've come up with an approach that just installs Nix as part of the `setup` step, which would mean no new container needed (yay).  I'm not sure why I'm hitting the current error, which is presumably due to how `bazelci.py` invokes things, so unfortunately my iteration right now is "test in CI" 😞  Apologies for the email spam.

rickvanprim:
It went green! 😱  That said, for whatever reason until I forced my local Nix to rebuild `cargo-bazel`, it was producing a different hash for the lockfile.  I'm a little concerned that might manifest again, although I'm also wondering if the examples should be running with `CARGO_BAZEL_REPIN=true` anyways to test that portion of things, though at the expense of not testing that it works with the existing lockfile?

nix/flake.nix
rickvanprim: @UebelAndre how would you like me to proceed here?

.bazelci/presubmit.yml
```diff
@@ -610,6 +610,21 @@ tasks:
   #   working_directory: examples/zig_cross_compiling
   #   build_targets:
   #     - "//..."
+  nix_cross_compiling:
+    name: Nix cross compiling test
+    platform: ubuntu2204
+    working_directory: examples/nix_cross_compiling
+    setup:
```
UebelAndre, line 617: This is something that can be flushed into the container now that you know what works, right?

UebelAndre (changes requested):

.bazelci/presubmit.yml
rickvanprim: I'm not sure what you mean by "flushed into the container"?  This is just using one of the normal Ubuntu containers now and I've updated my latest PR in the other repo to remove the "Ubuntu (Nix)" image: https://github.com/bazelbuild/continuous-integration/pull/1827

rickvanprim: If you mean baked into the image, no, not reasonably.  Having Nix ship with the image only saves a matter of seconds.  Almost all of the time is spent building `cargo-bazel` and fetching the toolchains.  Baking those into the image would be difficult and brittle, as either their definition needs to live in that repo and somehow be referenced here, or as soon as the respective lockfiles are out of sync, there will no longer be a cache hit on the Nix store in the image and we're back to where we are now.

UebelAndre: I think I'm confused, how is this setup doing anything other than machine configuration? Looks like you're installing nix, copying some stuff around and updating some config file? This to me feels like something that could go into the new nix image. But doesn't need to be a PR blocker. The commands looked like it took 13 seconds so no issue there.

.envrc
```diff
@@ -0,0 +1,3 @@
+# Use `path:` syntax to avoid copying the entire repo to the Nix Store.
```
UebelAndre, line 1: Can you add a comment to what the `.envrc` file is and link to some docs? For those who don't know anything about nix like me 😅

nix/flake.nix
```diff
@@ -0,0 +1,25 @@
+{
```
UebelAndre, line 1: Can you add a `BUILD.bazel` file to this package and a README explaining what the package is for?

UebelAndre (changes requested):
Oh sorry! I didn't realize CI had been passing. I just had a few more requests and then this otherwise looks good to me!

nix/flake.nix
UebelAndre: If we can avoid the top level package that'd be great, but if not then lets turn it into a package and make sure nothing is publicly visible.

.bazelci/presubmit.yml
rickvanprim: The problem there is that Nix has two modes, single user and multi-user.  Multi-user means install as `root` and run `nix daemon` to handle requests from all users.  This is turn means having an entrypoint script in the container to start the daemon and extra complexity over single user mode.  For single user, Nix needs to be installed by that user, which meant the `Dockerfile` either needs to have ahead of time knowledge that BuildKite will use 997:997 (UID/GID) or that the installation needed to happen as part of the test run.  I've opted for the latter as that seems like the simplest and most robust option.

rickvanprim:
@UebelAndre I have hopefully addressed your feedback and CI passed 🙂

UebelAndre (approved):
Awesome! Thank you for all your patience and hard work!

PR #1981, opened 2023-05-27
Fix typo in crate_universe-generated defs.bzl comment

dtolnay:
```diff
        vv
- dependnecies
+ dependencies
        ^^
```

This typo ends up in the caller's repo after they run crates_vendor, so it's worth fixing to keep from tripping a spell checker they might run on their repo.


illicitonion (approved):
Thanks!

Issue #1953, opened 2023-05-03
Rust-analyzer binaries should be added to `rust_analyzer_toolchain`


UebelAndre:
This would mostly be a fantastic way to add regression tests to the generated `rust-project.json` files where we can actually have rust-analyzer process them.

UebelAndre:
This was implemented in https://github.com/bazelbuild/rules_rust/pull/3670

PR #1982, opened 2023-05-30
Shorten rules_rust_rust_analyzer into rrra

KoltesDigital:
I'm on Windows. I just tried to set up Bazel for rust-analyzer, and I got this error:

```
> bazel run @rules_rust//tools/rust_analyzer:gen_rust_project
INFO: Analyzed target @rules_rust//tools/rust_analyzer:gen_rust_project (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: XYZ/external/rules_rust_rust_analyzer__proc-macro-error-attr-1.0.4/BUILD.bazel:49:19: Running Cargo 
build script proc-macro-error-attr [for tool] failed: (Exit 101): cargo_build_script_runner.exe failed: error executing command (from target @rules_rust_rust_analyzer__proc-macro-error-attr-1.0.4//:proc-macro-error-attr_build_script) bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\rules_rust\cargo\cargo_build_script_runner\cargo_build_script_runner.exe ... (remaining 10 arguments skipped)
thread 'main' panicked at 'Unable to start binary: Os { code: 267, kind: NotADirectory, message: "The directory name is invalid." }', external/rules_rust/cargo/cargo_build_script_runner/lib.rs:127:41
Target @rules_rust//tools/rust_analyzer:gen_rust_project failed to build
ERROR: XYZ/external/rules_rust_rust_analyzer__proc-macro-error-attr-1.0.4/BUILD.bazel:18:16 Compiling Rust 
proc-macro proc_macro_error_attr v1.0.4 (4 files) [for tool] failed: (Exit 101): cargo_build_script_runner.exe failed: error executing command (from target @rules_rust_rust_analyzer__proc-macro-error-attr-1.0.4//:proc-macro-error-attr_build_script) bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\rules_rust\cargo\cargo_build_script_runner\cargo_build_script_runner.exe ... (remaining 10 arguments skipped)
INFO: Elapsed time: 2.112s, Critical Path: 0.28s
INFO: 9 processes: 9 internal.
ERROR: Build failed. Not running target
```

I've observed that setting the command's `current_dir` is the issue. Although the directory exists, surely symlinks don't play nice.

However, looking at the code, the command don't need to be in a particular working directory, since all input and output paths are given as environment variables.

By removing this line, generation of `rust-project.json` succeeds, and rust-analyzer takes it into account so far.


illicitonion:
Thanks for putting together this pull request!

Unfortunately, there are many cargo build scripts which _do_ assume they're run from a specific directory (because cargo documents that they are), so do things like try to read files using relative paths.

Could you put together a test case which we can run on our CI that shows the problem? We have Windows CI set up here: https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/.bazelci/presubmit.yml#L62-L68 and if you add a `run_targets` key (e.g. https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/.bazelci/presubmit.yml#L286-L287) you should be able to `bazel run` things. It'd be useful to know exactly what flags you're running with (e.g. are you enabling runfiles, etc) which you can encode in a test in CI, so that we can reproduce the failure and work out what the problem with the directory is.

KoltesDigital:
Hi @illicitonion,

Agreed for the working directory. [(for the record)](https://doc.rust-lang.org/cargo/reference/build-scripts.html#inputs-to-the-build-script)

I saw you run rust-analyzer tests on Ubuntu during CI, so I've just added the same tests on Windows (you'll surely use YAML anchors). Sadly they fail but due to another reason: https://buildkite.com/bazel/rules-rust-rustlang/builds/8639#01886ea9-9147-45d8-8a83-7c38a731b8c8 . Do you have any idea for that one?

illicitonion:
Thanks for digging in to investigate!

One of the things we've done in the past where we hit path length issues (which is definitely a bit of a whack a mole situation!) is to rename repositories to shorten their directory names (e.g. in https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/crate_universe/3rdparty/BUILD.bazel#L33-L36); https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/tools/rust_analyzer/3rdparty/BUILD.bazel#L36 looks like a fine example of something which we could rename to something like `rrra` - because this ends up in the path twice, that would save us 40 characters. Maybe that's enough, or maybe we need to add some other path shortenings (e.g. we could have a mode where `proc-macro-error-attr-1.0.4` gets shortened down to maybe a 5-character hash of the input string or something - that starts to feel like an actual usability trade-off, but the `rrra` thing seems like a pretty unambiguous win)

KoltesDigital:
Agreed for the name. Though it hurts!

It works on my machine, but I still find relevant to add it to your CICD, and it is still failing because of the linking error (while running the tests). I played with flags but they had no effects. Would you have any clue about that?

KoltesDigital:
@illicitonion Do you have time to look at the CI error, or do you prefer that I remove it from the branch?

scentini:
I'm lacking context here, but I believe the current CI failure comes from too long a path:
C:\tools\msys64\home\b\_bazel_b\v2y4dvlc\execroot\rules_rust_test_rust_analyzer\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools\rust_toolchain\lib\rustlib\x86_64-pc-windows-msvc\lib\libstd-43ecf4bdbb5e1e31.rlib'
Could we also shorten `rules_rust_test_rust_analyzer` as part of this PR?

illicitonion:
It looks like in CI `C:\tools\msys64\home\b\_bazel_b\tmssgypz\execroot\rrt_ra\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools\rust_toolchain\lib\rustlib\x86_64-pc-windows-msvc\lib\librustc_demangle-cae8325b8f11cb55.rlib` is still ~6 characters too long :(

@UebelAndre any thoughts on shortening these? Some things off the top of my head some things that should be low-hanging fruit here:
1. removing the `_tools` suffix from toolchains (and maybe adding a `_proxy` suffix to the proxy toolchain names?)
2. Shortening `windows` to `win` or `windows_x86_64` to `win64` or similar in the platform triple names - `windows_x86_64__x86_64-pc-windows-msvc` is pretty repetative and redundant...

scentini:
`rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools` <- I think we should see if we can make this shorter, but also, in CI we have `--output_user_root=C:/b`, however the too long a path starts with `C:\tools\msys64\home\b\`. Any idea why that's the case?

illicitonion:
> in CI we have `--output_user_root=C:/b`, however the too long a path starts with `C:\tools\msys64\home\b\`. Any idea why that's the case?

Good spot!

bazelci passes these as explicit values to the `bazel` invocations it makes: https://github.com/bazelbuild/continuous-integration/blob/111311466e7468a53c7b7e5aa1ea5234614c52aa/buildkite/bazelci.py#L1850-L1857

But this test runs bazel itself explicitly: https://github.com/bazelbuild/rules_rust/blob/9314b1b0c0fb3ffd87b4911e59cb8ea31cbf99be/test/rust_analyzer/rust_analyzer_test_runner.sh#L70-L80

Ideally bazelci would write these settings to a .bazelrc file somewhere where we could source them, rather than needing to re-implement them... I don't know if we could easily rewrite this shell script to instead be a workspace which we can just pass straight to bazelci rather than writing our own orchestrator for... @UebelAndre?

UebelAndre:
I haven't read the entire thread but just to throw it out there, what if [generate_build_scripts](https://bazelbuild.github.io/rules_rust/crate_universe.html#crates_vendor-generate_build_scripts) was set to false for [//tools/rust_analyzer/3rdparty:crates_vendor](https://github.com/bazelbuild/rules_rust/blob/0.24.1/tools/rust_analyzer/3rdparty/BUILD.bazel#L4-L56)? This usually eliminates long path issues.

Also, I opened https://github.com/bazelbuild/rules_rust/issues/1953 some time ago to try and provide as simplified means of testing rust analyzer. Perhaps that would help too? I can try to take a closer look here before the end of the week.

UebelAndre:
Maybe the tests could be simplified by using https://github.com/bazelbuild/continuous-integration/issues/1688 ?

UebelAndre:
Closing this out as the change was made in https://github.com/bazelbuild/rules_rust/pull/2065 but feel free to reopen it if there's still some debugging going on here.

PR #1984, opened 2023-05-31
Making rust_std attr in rust_toolchain mandatory

daivinhtran:
`rust_toolchain`'s `rust_std` attribute should be mandatory because when it's empty, [this line](https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/rust/toolchain.bzl#L147) will always fail with

```
if not rust_common.stdlib_info in rust_std:
Error: unsupported binary operation: Provider in NoneType
```


UebelAndre:
Shouldn’t this instead be updated to not fail here? This way no_std toolchains can succeed?

daivinhtran:
@UebelAndre 

Given the current [_rust_toolchaim_impl](https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/rust/toolchain.bzl#LL456C5-L456C25), I don't think it's ever possible for `rust_std` to be unset. In [this block](https://github.com/bazelbuild/rules_rust/blob/56e7604879ac74cb8c3d65cbb5d41611f7424977/rust/toolchain.bzl#L261-L265) that handles "no_std_with_alloc", we're still expecting `rust_std` to be defined. 

Do you have an example of `rust_toolchain` instance that doesn't set `rust_std`?

scentini:
This LGTM. Even with `no_std` mode `rustc` still requires some crates from the set of files we call `std`, namely `libcore` (and maybe `libcompiler_builtins`).

scentini (approved):

PR #1986, opened 2023-06-01
Shorten build script target names

illicitonion:
I don't love this, but it helps with the windows path length limits, saving 10 characters.


UebelAndre:
These `MAX_PATH` issues (https://github.com/bazelbuild/rules_rust/issues/1120) on Windows are the real `_bs` 😆

PR #1987, opened 2023-06-01
Fix up anchor link

illicitonion:
Stardoc escapes raw HTML, so this was outputting an escaped anchor tag, and a broken link, rather than working.


UebelAndre (approved):
Thanks!

PR #1988, opened 2023-06-01
Set windows flags in platform-specific bazelrc

illicitonion:
Rather than writing them in .bazelci yaml blocks.

This makes running the examples on Windows more likely to work out of the box.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -60,10 +60,6 @@ tasks:
     test_targets: *default_macos_targets
     coverage_targets: *default_macos_targets
   windows:
-    build_flags:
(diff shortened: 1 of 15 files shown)


UebelAndre (approved):
Love it!

UebelAndre:
Wait, does this really work? I'm seeing the following when building the examples workspace
```
WARNING: while reading option defaults file '/Users/user/Code/rules_rust/examples/.bazelrc':
  invalid command name 'startup:windows'.
```

illicitonion:
Good point, fixed in https://github.com/bazelbuild/rules_rust/pull/2012 :)

PR #1990, opened 2023-06-02
Fixes crates_vendor workspace name detection when using bzlmod

csmulhern:
This fixes a similar issue to #1514, fixed by #1575.

When using Bzlmod, the main default repository name is now `_main`, instead of `__main__`. This PR extends detection of this use case to both scenarios.


illicitonion (approved):
Thanks!

PR #1991, opened 2023-06-02
Added Rust 1.70.0

UebelAndre:
https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html


UebelAndre:
457bec9ce4a2276aab010bc3d9747cc18f680521 is solving for the following [build failure](https://buildkite.com/bazel/rules-rust-rustlang/builds/8657#01887ad7-3ae3-4581-98f4-cf579fe7cb84)
```
error: ambiguous glob re-exports
 --> bazel-out/k8-fastbuild/bin/proto/helloworld/helloworld_proto.grpc.rust/lib.rs:5:9
  |
5 | pub use common::*;
  |         ^^^^^^^^^ the name `file_descriptor_proto` in the value namespace is first re-exported here
...
9 | pub use helloworld::*;
  |         ------------- but the name `file_descriptor_proto` in the value namespace is also re-exported here
  |
  = note: `-D ambiguous-glob-reexports` implied by `-D warnings`
 
```

If anyone has a better fix I'd happily apply it but I'd like to avoid making any changes to protobuf and let it start to die out in favor of https://github.com/protocolbuffers/protobuf/tree/v23.2/rust

@scentini @hlopko Curious on your thoughts here. I'd love to delete the proto rules here in favor of ones defined in protobuf.

illicitonion (approved):
LGTM - and I'd in general love to delete the proto rules

hlopko:
Please do not use the rules in the Protobuf repository just yet, they don't work with protobuf-rust or prost, only with the new Rust implementation that we're experimenting with there. Definitely not worth anybody's time ATM.

Also, once it is ready, we'll have to discuss where the rules should be. It's not clear if the protobuf repo is the best location.

UebelAndre:
> Please do not use the rules in the Protobuf repository just yet, they don't work with protobuf-rust or prost, only with the new Rust implementation that we're experimenting with there. Definitely not worth anybody's time ATM.
> 
> Also, once it is ready, we'll have to discuss where the rules should be. It's not clear if the protobuf repo is the best location.

I'm not migrating anything or anyone yet. I don't think any active maintainers have context with the proto rules in this repo so any issues with them or pull requests are pretty taxing. I don't want this PR to be blocked by this issue so I created https://github.com/bazelbuild/rules_rust/issues/1992 to track re-enabling the targets. If you or anyone you know has context on the proto rules that'd help a ton.

PR #1993, opened 2023-06-02
Add support for armv8-m

afoxley:
* Regenerate documentation
 * Update crate_universe example lock files


afoxley:
@UebelAndre There's still one lock file to update that doesn't seem to get done with the existing instructions.

https://buildkite.com/bazel/rules-rust-rustlang/builds/8670#01887cbf-16db-4112-a64a-6f9391cb6eb1

What's the command to do that one?

UebelAndre (approved):
Thanks! I forgot about that directory. Glad you found it!

Issue #919, opened 2021-08-29
Bindgen rules are not supported on all platforms


UebelAndre:
The `//bindgen` rules are currently not working on Windows and in RBE environments. This should be fixed

sayrer:
Yeah, they are only defined for Mac and Linux:

https://github.com/bazelbuild/rules_rust/blob/bb38e76dc148b2460e326f7f2181c40d13b97fc9/bindgen/BUILD.bazel#L14

UebelAndre:
Yeah, makes sense. This is more of a tracking issue for comments being added in https://github.com/bazelbuild/rules_rust/pull/918

Ideally we should support all platforms but for the time being, my hope is that users see this issue instead of needing to dig through the code.

UebelAndre:
`rust_bindgen` fails with the following error in remote build environments (remote execution):
```
(15:08:59) ERROR: /workdir/examples/bindgen/BUILD.bazel:10:21: Generating bindings for bindgen/simple.h.. failed: (Exit 101): cargo_bin_bindgen failed: error executing command
  (cd /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/90e2e9b2d72fa9b00c19ff0535bb31ae/execroot/examples && \
  exec env - \
    CLANG_PATH=bazel-out/k8-opt-exec-60723E07/bin/external/bindgen_clang_linux/clang \
    LIBCLANG_PATH=bazel-out/k8-opt-exec-60723E07/bin/_solib_k8/_U@bindgen_Uclang_Ulinux_S_S_Clibclang___Ulib \
    RUST_BACKTRACE=1 \
  bazel-out/k8-opt-exec-60723E07/bin/external/rules_rust_bindgen__bindgen__0_58_1/cargo_bin_bindgen --no-rustfmt-bindings '--allowlist-var=SIMPLE_.*' bindgen/simple.h --output bazel-out/k8-fastbuild/bin/bindgen/simple_bindgen__bindgen.rs.unformatted -- -iquote . -iquote bazel-out/k8-fastbuild/bin)
# Configuration: 2098c73efc81d9b6effdf6ab8b32d3798f8ca72c71a5005bd40c6a1e8a36ddb5
# Execution platform: @buildkite_config//config:platform
[...]

PR #1998, opened 2023-06-07
Update bindgen rules to build clang from source.

UebelAndre:
This change updates the default `rust_bindgen` rules to build `libclang` and `clang` from source using the [Bazel rules sourced in the llvm-project repository](https://github.com/llvm/llvm-project/blob/llvmorg-14.0.6/utils/bazel/README.md) for use in `RustBindgen` actions. The version of Clang that Bindgen uses by default is updated `14.0.6` which was what I was able to get building.

Some breaking changes include:
- Users must now load one extra bzl file if they don't already have have `llvm-project` defined in their WORKSPACE.
```
load("@rules_rust//bindgen:transitive_repositories.bzl", "rust_bindgen_transitive_dependencies")

rust_bindgen_transitive_dependencies()
```
- `rust_bindgen_repositories` is deleted as a means of forcing users to take a look at the updated dependencies.

closes https://github.com/bazelbuild/rules_rust/issues/919


8573:
Is there an option not to build Clang from source, for users who don't need that?

UebelAndre:
You should still be able to register your own toolchains

UebelAndre:
Is there an issue you’re running into building clang from source?

aaronmondal:
@UebelAndre Cool change! I want to integrate bindgen-similar functionality between `rules_ll` and `rules_rust` via the `cxx` crate and with this change I can test interop with our bzlmod variant for LLVM.

For reference, LLVM/bzlmod in rules_ll looks like this: https://github.com/eomii/rules_ll/blob/bdd243466d7c5e8be8a9d7f7fcb179b2a1ae084b/MODULE.bazel#L25-L50.

The upstream patch is at https://reviews.llvm.org/D137008. I'll double-check that things are compatible with the usage in `rules_rust` when I find time to revise that patch.

PR #2007, opened 2023-06-13
Make `rust_doc_test` inherit it's crate aliases attribute

### LINKED ISSUE #2006: `rust_doc_test` is failing to compile when `crate` target uses renamed dependencies
There is no way to provide the `aliases` used by crate to `rust_doc_test`. I've created an [example workspace](https://github.com/k1nkreet/bazel-workspaces/tree/master/rust_doc_test_aliases) to reproduce

k1nkreet:
Fixes #2006


UebelAndre:
> Sure! Do you want me to create an example workspace in `examples/` or there is other way to add new tests?

We have some tests you could probably add to here: https://github.com/bazelbuild/rules_rust/tree/0.22.0/test/unit/rustdoc

Does that work for you?

k1nkreet:
Yup, thank you

k1nkreet:
Done

test/unit/rustdoc/rustdoc_test_dep_with_alias.rs
```diff
@@ -0,0 +1,6 @@
+/// ```
+/// let v = 42;
```
UebelAndre, line 2: Does the test still work if you use `aliased_adder` in the test? Feels like that'd be good regression testing to have too.

UebelAndre (changes requested):
Thanks! I had just one last request 😅

test/unit/rustdoc/rustdoc_test_dep_with_alias.rs
k1nkreet: No it will not, because tests also are not inherit `aliases` from crate. In test you still can workaround it because it has it's own `aliases` you can specify explicitly. I thought it's probably a topic for the separate PR when I was thinking about fixing it here :sweat_smile:

k1nkreet: Or maybe I misunderstood you question :sweat_smile: Did you mean if I use `aliased_adder` in doc?

UebelAndre: Yes! :)

test/unit/rustdoc/rustdoc_test_dep_with_alias.rs
```diff
@@ -1,5 +1,8 @@
+/// Here we are trying to use renamed dependency
+/// inside a doc string to check rust_doc_test
 /// ```
-/// let v = 42;
+/// use aliased_adder;
+/// let v = aliased_adder::inc(42);
```
UebelAndre, line 5: One last nit
```suggestion
///  assert_eq!(43, aliased_adder::inc(42));
```

UebelAndre (approved):
One last nit but otherwise this is exactly what was thinking, thank you!

UebelAndre (approved):

Issue #2013, opened 2023-06-14
Rust Analyzer Missing ENV vars


tdbgamer:
When using proc-macros that rely on rustc env vars they blow up and break the intellisense in my project. Looks like the root cause is this line:
https://github.com/bazelbuild/rules_rust/blob/e80582e7516a47c3fb80592f8f24bf38cdf04ee3/rust/private/rust_analyzer.bzl#L89

It's using the attributes defined in the rule rather than `crate_info.rustc_env` which has all the env vars used at compile time. I can PR a fix, but just want to be sure this wasn't on purpose. I applied a patch file locally and everything works like a charm.

scentini:
Replacing the line you pointed above with `env = crate_info.rustc_env` seems fine to me. Does that solve the failures? Also could you be more specific on what the failures/breakage is?

tdbgamer:
That alone won't fix the issue. If you put a print statement there, you'll see that all the `rustc` env vars are missing. The issue is that `CrateInfo` is being constructed _too early_ in the rust_binary/rust_library rules and passed down to the `construct_arguments` function that actually creates all the env vars later, but at that point I believe the `CrateInfo` is already frozen. Could probably fix it by just refactoring that code to not crate crate_info until the end, but I didn't want to make a big change to the structure of all this code.

With both of those changes, `gen_rust_project` sees all the env vars and it fixes all my IDE issues. `rust_test` seems to run into issues as well because the `CARGO_MANIFEST_DIR` has `${pwd}` in it that is replaced in the wrapper script, but messes with rust analyzer (only seems to print out warns luckily) and causes the make variable expansion on `rust_test` to fail (only with inline tests since they copy the `CrateInfo`). So I had to also store a snapshot of the original env vars passed by the user to get those working again. Would probably be better to expand all those variables before constructing CrateInfo to prevent that failure as well.

tdbgamer:
Looks like the `${pwd}` also breaks clippy 🥲

daivinhtran:
> When using proc-macros that rely on rustc env vars they blow up and break the intellisense in my project.

Hi @tdbgamer , could you provide some steps for me to reproduce the issues? And what did you mean by "they blow up and break the intellisense"? Can you be more specific?

tdbgamer:
[...]

PR #2014, opened 2023-06-14
Make rust-analyzer use the same rustc_env as is used at compile time.

timbess:
Addresses https://github.com/bazelbuild/rules_rust/issues/2013

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1391,13 +1391,18 @@ def rustc_compile_action(
 
         # The path to the package dir, including a trailing "/".
         package_dir = ctx.bin_dir.path + "/"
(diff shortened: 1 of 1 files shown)


tdbgamer:
There is likely a better way to do this, but I'm not super familiar with the project or bazel. It fixes all the issues I'm seeing with my app though. Some guidance would be helpful.

scentini:
@daivinhtran could you take a look at better ways to do this? I believe it will help if we refactor the code to construct `CrateInfo` later on inside `rustc_compile_action`.

daivinhtran:
Refactoring the rule implementations to construct `CrateInfo` later in `rustc_compile_action` seems like the right approach to accomplish this. I'll take a stab at it.

PR #2017, opened 2023-06-16
fix: load cargo manifest without resolving abs path of deps

### LINKED ISSUE #2016: `Cargo.Bazel.lock` checksum is unstable when using path dependencies in cargo workspace
The `Cargo.Bazel.lock` files produced on mac vs linux are byte-to-byte the same except for the checksum. Could you point me to where this checksum is calculated? Is this expected or a bug? I'm currently using `rules_rust 0.23.0`.

liningpan:
As described in the original issue, `cargo_toml` automatically resolves path dependencies inherited from the cargo workspace to their absolute paths, if the manifest is loaded with  `cargo_toml::Manifest::from_path`. 

The fix I propose here uses the `cargo_toml::Manifest::from_slice`, which does not attempt to load information from the workspace, and dependencies will only show up `Dependency::Inherited` instead of `Dependency::DependencyDetail`.

https://docs.rs/cargo_toml/latest/cargo_toml/struct.Manifest.html#method.from_str
> It does not call `complete_from_path`, so may be missing implicit data.

Even though the documentation says that it may be missing implicit data, it is not a problem for our use case, since the workspace manifest is also loaded and included in the hash. 

fix #2016


crate_universe/src/splicing.rs
```diff
@@ -118,8 +118,10 @@ impl TryFrom<SplicingManifest> for SplicingMetadata {
             .manifests
             .into_iter()
             .map(|(path, label)| {
-                let manifest = cargo_toml::Manifest::from_path(&path)
+                let manifest_content = fs::read(&path)
```
UebelAndre, line 121: Can you add a comment on the significance of not using `cargo_toml::Manifest::from_path`?

crate_universe/src/splicing.rs
```diff
@@ -582,4 +584,35 @@ mod test {
             PathBuf::from("/tmp/abs/path/workspace/.cargo/config.toml"),
         )
     }
+
+    #[test]
+    fn splicing_metadata_workspace_path() {
```
UebelAndre, line 589: Can you explain how this tests the issue you're seeing? At a glance it seems unrelated.

crate_universe/src/splicing.rs
```diff
@@ -118,9 +118,16 @@ impl TryFrom<SplicingManifest> for SplicingMetadata {
             .manifests
             .into_iter()
             .map(|(path, label)| {
-                let manifest = cargo_toml::Manifest::from_path(&path)
+                // We read the content of a manifest file to buffer and use `from_slice` to
+                // parse it. The reason is that the `from_path` version will resolve indirect
+                // path dependencies in the workspace to absolute path, which causes the hash
+                // to be unstable. Not resolving implicit data is okay here because the
+                // workspace manifest is also included in the hash.
+                // See issue #2016
```
UebelAndre, line 126:
```suggestion
                // See https://github.com/bazelbuild/rules_rust/issues/2016
```

crate_universe/src/splicing.rs
```diff
@@ -582,4 +589,39 @@ mod test {
             PathBuf::from("/tmp/abs/path/workspace/.cargo/config.toml"),
         )
     }
+
+    #[test]
+    fn splicing_metadata_workspace_path() {
+        let runfiles = runfiles::Runfiles::create().unwrap();
+        let workspace_path =
```
UebelAndre, line 596: I think this is undefined behavior with the runfiles API. You should always look for files and then you check for files relative to that. I'd suggest doing something like
```suggestion
        let workspace_manifest_path = runfiles
            .rlocation("rules_rust/crate_universe/test_data/metadata/workspace_path/Cargo.toml");
        let workspace_path = workspace_manifest_path.parent().unwrap();
```

crate_universe/src/splicing.rs
UebelAndre: Awesome! Thank you!

UebelAndre (changes requested):
Looks good! Just one last request 😄

crate_universe/src/splicing.rs
```diff
@@ -582,4 +589,41 @@ mod test {
...
         )
     }
+
+    #[test]
+    fn splicing_metadata_workspace_path() {
+        let runfiles = runfiles::Runfiles::create().unwrap();
+        let workspace_manifest_path = runfiles
+            .rlocation("rules_rust/crate_universe/test_data/metadata/workspace_path/Cargo.toml");
+        let workspace_path = workspace_manifest_path.parent().unwrap();
+        runfiles.rlocation("rules_rust/crate_universe/test_data/metadata/workspace_path");
```
UebelAndre, line 599:
```suggestion
```

crate_universe/src/splicing.rs
```diff
@@ -582,4 +589,41 @@ mod test {
...
+            "rules_rust/crate_universe/test_data/metadata/workspace_path/child_a/Cargo.toml",
+        );
+        let child_b_path = runfiles.rlocation(
+            "rules_rust/crate_universe/test_data/metadata/workspace_path/child_b/Cargo.toml",
+        );
+        let child_b_label = Label::from_str("//child_b:Cargo.toml").unwrap();
+        let manifest = SplicingManifest {
+            direct_packages: BTreeMap::new(),
+            manifests: BTreeMap::from([
+                (parent_path, Label::from_str("//:Cargo.toml").unwrap()),
```
UebelAndre, line 612:
```suggestion
                (workspace_manifest_path, Label::from_str("//:Cargo.toml").unwrap()),
```

UebelAndre (approved):

crate_universe/src/splicing.rs
liningpan: Sorry, I didn't look carefully either. This won't compile because `parent()` borrows.
The suggestion ui is really confusing.

UebelAndre: No worries! And yeah, it could be better by supporting multi line suggestions. But glad everything is sorted!

PR #2022, opened 2023-06-19
Update `rust_library_group` to use `DepVariantInfo`

UebelAndre:
This seems like a much cleaner interface and gives users more confidence that `CrateInfo` providers are paired with the right `DepInfo` providers.

A follow-up to https://github.com/bazelbuild/rules_rust/pull/1848

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -49,6 +49,7 @@ PAGES = dict([
         symbols = [
             "rust_binary",
             "rust_library",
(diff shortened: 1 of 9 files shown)


UebelAndre:
cc @william-smith-skydio does this change also work for your needs?

scentini (approved):

william-smith-skydio:
Yes, this looks cleaner. Thanks.

PR #2036, opened 2023-06-27
Add `opt_level` argument to `rust_repository_set`.

konkers:
`rust_toolchain` allows setting of optimization level. This changes plumbs that into `rust_repository_set` to allow control when declaring toolchains through that rule.


rust/private/repository_utils.bzl
```diff
@@ -253,6 +253,7 @@ rust_toolchain(
     visibility = ["//visibility:public"],
     extra_rustc_flags = {extra_rustc_flags},
     extra_exec_rustc_flags = {extra_exec_rustc_flags},
+    {opt_level_arg}
```
UebelAndre, line 256: nit: you can instead pass `None` to `opt_level`
```suggestion
    opt_level = {opt_level_arg},
```

rust/private/repository_utils.bzl
```diff
@@ -324,6 +331,7 @@ def BUILD_for_rust_toolchain(
         llvm_profdata_label = llvm_profdata_label,
         extra_rustc_flags = extra_rustc_flags,
         extra_exec_rustc_flags = extra_exec_rustc_flags,
+        opt_level_arg = opt_level_arg,
```
UebelAndre, line 334: I think this cleans up the formatting a bit
```suggestion
        opt_level = repr(opt_level),
```

UebelAndre (changes requested):
Thanks! Just one minor nit if you don't mind 😄

rust/private/repository_utils.bzl
konkers: When `None` gets passed to `opt_level`, that overrides the default value and among other issues this line will error out: https://github.com/bazelbuild/rules_rust/blob/main/rust/toolchain.bzl#L471

UebelAndre: Sorry for the delay, `None` appears to allow the default to be used. I tested by hard-coding the value to `None` in this template and added a print to `toolchain.bzl`
```
DEBUG: /Users/rules_rust/rust/toolchain.bzl:465:10: {"dbg": "0", "fastbuild": "0", "opt": "3"}
```

UebelAndre (changes requested):

rust/private/repository_utils.bzl
konkers: Thanks for the feedback!  Just got back from out of town.   Did some more digging in.  The issues I encounter stems from how the `opt_level` argument to `rust_toolchain_tools_repository` is declared (https://github.com/bazelbuild/rules_rust/pull/2036/commits/3531265ea521516e1a1076014fb40205a693b8ef#diff-e1e53bc6cac78a1c114f07108364ea2a4956bba844b457bdd3c57ad30845fa09R417).  `attr.string_dict` defaults to `{}` when no value is given which get's passed down to the invocation to `rust_toolchain`.  `attr.string_dict` does not allow for `default = None` (giving a "parameter 'default' got value of type 'NoneType', want 'dict'" error).  A simple `opt_level = ctx.attr.opt_level if ctx.attr.opt_level else None,` in `_rust_toolchain_tools_repository_impl`'s invocation of `BUILD_for_rust_toolchain` fixes this issue.  I'll push a new patch today.

rust/repositories.bzl
```diff
@@ -413,6 +414,9 @@ rust_toolchain_tools_repository = repository_rule(
         "iso_date": attr.string(
             doc = "The date of the tool (or None, if the version is a specific version).",
         ),
+        "opt_level": attr.string_dict(
+            doc = "Rustc optimization levels.",
```
UebelAndre, line 418:
```suggestion
            doc = "Rustc optimization levels. For more details see the documentation for `rust_toolchain.opt_level`.",
```

UebelAndre (approved):
Thank you, looks good to me!

One final nit for docs but can merge after a rebase 😄

rust/repositories.bzl
konkers: Done

PR #2043, opened 2023-07-01
Moved legacy protobuf rules to `proto/protobuf`

UebelAndre:
This aligns better with the `//proto` package now having multiple kinds of rules.


illicitonion (approved):
Any thoughts on whether we should rename and/or merge the repository rules for setting up the legacy and the prost/tonic toolchains?

Right now I think they have pretty similar names, and the difference isn't super clear?

illicitonion:
> > Any thoughts on whether we should rename and/or merge the repository rules for setting up the legacy and the prost/tonic toolchains?
> > Right now I think they have pretty similar names, and the difference isn't super clear?
> 
> I was thinking there should be no root level macros anymore to disambiguate the two. What do you think?

Sounds reasonable - I'd be tempted to `s/rust_proto_dependencies/rust_proto_protobuf_dependencies/` or something similar to make the call sites slightly more clear, but I don't really mind :)

UebelAndre:
> Sounds reasonable - I'd be tempted to `s/rust_proto_dependencies/rust_proto_protobuf_dependencies/` or something similar to make the call sites slightly more clear, but I don't really mind :)

Would you do the same for `prost`?
`rust_proto_prost_dependencies` and `rust_proto_protobuf_dependencies`?

illicitonion:
> > Sounds reasonable - I'd be tempted to `s/rust_proto_dependencies/rust_proto_protobuf_dependencies/` or something similar to make the call sites slightly more clear, but I don't really mind :)
> 
> Would you do the same for `prost`? `rust_proto_prost_dependencies` and `rust_proto_protobuf_dependencies`?

Honestly happy either way. I lean towards no because we want it to be the only way in the future, but I would be ok with yes because it's not the only way today.

PR #2044, opened 2023-07-01
Fix prost proto packages not sanitizing to valid module names

UebelAndre:
This normalizes module names to be consistently written as lowercase. When given extravagant package names `prost-build` will snake_case package names `PackageX.FooBar` would become `package_x.foo_bar`. This change contains two commits where the first solves a bug where the extern paths that the process wrapper generated did not match the the paths `prost-build` would render. Sanitizing these to lowercase is the simplest fix. The second commit contains a change to add snake_case support by bringing in the same library `prost-build` uses to do the same logic but is a more complicated implementation. I personally lean toward the simpler solution but leave it up to my reviewers.

### DIFF
--- a/proto/prost/private/3rdparty/BUILD.heck.bazel
+++ b/proto/prost/private/3rdparty/BUILD.heck.bazel
@@ -1,0 +1,20 @@
+load("@rules_rust//rust:defs.bzl", "rust_library")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_library(
+    name = "heck",
+    srcs = glob(["**/*.rs"]),
+    crate_features = [
+        "default",
+    ],
+    crate_root = "src/lib.rs",
+    edition = "2018",
+    rustc_flags = ["--cap-lints=allow"],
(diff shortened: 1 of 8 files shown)


UebelAndre:
cc @freeformstu curious on your thoughts if this continues to work for your needs too.

freeformstu:
I think this looks great. I'm fine with the additional dependency, that doesn't bother me, and it makes it more correct.

UebelAndre:
> TLDR: Should commit 2 be dropped in favor of the simpler implementation?
> 
>     * Commit 1 - Sanitize rust module names to lowercase
> 
>     * Commit 2 - Bring in a new dependency to sanitize to snake case the way `prost-build` does.

Also curious if @hlopko has thoughts here. How would you expect the package names to be normalized and do you think the added complexity should be added to match the way Prost does things?

illicitonion (approved):
Thanks!

I would lean towards faithfully minicing `prost-build`'s behaviour as the second commit does.

PR #2061, opened 2023-07-13
Example of rust dynamically linking cc shared library

daivinhtran:


UebelAndre (changes requested):
Thanks, Great idea!

I had just one request 😄

PR #2063, opened 2023-07-13
Updated crate_universe dependencies

UebelAndre:
Changes:
- Updates various dependencies in `//crate_universe:Cargo.toml`
- Update the Rust edition for crate_universe to 2021
- Update crate_unverse version to `0.9.0`
- Pin back dependencies for rust_analyzer to support Rust 1.62.0


UebelAndre:
@illicitonion friendly ping 😄

.bazelci/presubmit.yml
```diff
@@ -3,7 +3,7 @@ aspects_flags: &aspects_flags
   - "--config=rustfmt"
   - "--config=clippy"
 min_rust_version_shell_commands: &min_rust_version_shell_commands
-  - sed -i 's|^rust_register_toolchains(|rust_register_toolchains(versions = ["1.63.0"],\n|' WORKSPACE.bazel
+  - sed -i 's|^rust_register_toolchains(|rust_register_toolchains(versions = ["1.64.0"],\n|' WORKSPACE.bazel
```
illicitonion, line 6: What necessitates this bump? I think we were originally aiming for 2 years of rust versions, which would be 1.62.0 - is there an important reason we need to bump MSRV, or is there a small update we could not perform to keep 1.63.0 compatibility?

UebelAndre: The failure can be seen here: https://buildkite.com/bazel/rules-rust-rustlang/builds/9038#018954c9-91f8-44cc-a55d-a26c40cfc903

[rustix](https://crates.io/crates/rustix) seems to be the culprit

illicitonion: Could you downgrade `is-terminal` to `0.4.7` which I think will bring us back to 1.62.0?

UebelAndre: Yeah, that then leaves [libgit2-sys](https://crates.io/crates/libgit2-sys) https://buildkite.com/bazel/rules-rust-rustlang/builds/9040#018954e3-34cf-46ef-83d9-3cc5348d11e2

illicitonion: Shall we drop that one back to its previous version?

UebelAndre: Alas, there's now https://buildkite.com/bazel/rules-rust-rustlang/builds/9044#018954f7-378c-4b72-9c44-62f375aa617d for `ignore`

UebelAndre: Ok, everything seems to be happy now 😄

illicitonion (approved):
Thanks!

PR #2070, opened 2023-07-20
Remove cfg exec from linker_scripts attribute.

granaghan:
When cross compiling and generating linker scripts based on previous rust_binary outputs, this constraint causes the rust_binary to incorrectly get rebuilt for exec.


krasimirgg:
Looks reasonable! My interpretation is that linker scripts are like source files to be fed to the linker, analogously how source files are fed to a compiler, hence non-exec.

Thank you!

krasimirgg (approved):

PR #2076, opened 2023-07-26
Only add coverage instrumentation to tests

neilisaac:
Linking a rust_shared_library with ld.gold fails when building a rust_shared_library that exposes FFI bindings used by a test written in another language.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -996,7 +996,7 @@ def construct_arguments(
         rustc_flags.add("--extern")
         rustc_flags.add("proc_macro")
 
-    if toolchain.llvm_cov and ctx.configuration.coverage_enabled:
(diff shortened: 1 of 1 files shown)


krasimirgg:
Could you provide a minimal example where this reproduces and/or add a regression test for this? How does the linking fail exactly?

It feels like this PR will disable coverage collection of rust dependencies of a rust test, which seems wrong. Also, I'd expect for coverage to be able to work across at least C++ and rust whenever the toolchain definitions are compatible, e.g.: running with coverage a cc_test that depends on a rust_library should be able to produce coverage information from the rust sources.

Since shared libraries are tricky and intended for interfacing with other languages that may have incompatible coverage support, a possibility would be to disable coverage instrumentation just for the rust_shared_library crate types themselves.

Issue #2079, opened 2023-07-27
Coverage tests failing on min Bazel version jobs


UebelAndre:

UebelAndre:
I'm actually seeing the same issue for other jobs as well: https://buildkite.com/bazel/rules-rust-rustlang/builds/9072#018993cc-4443-4449-904c-e2695fbd9a86

UebelAndre:
I opened https://github.com/bazelbuild/bazel/issues/19098 to track this as a larger issue.

PR #2080, opened 2023-07-27
Disable coverage checks for min Bazel version jobs

UebelAndre:
https://github.com/bazelbuild/rules_rust/issues/2079

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -210,7 +210,8 @@ tasks:
     build_targets: *default_linux_targets
     test_targets: *default_linux_targets
     coverage_targets: *default_linux_targets
-    post_shell_commands: *coverage_validation_post_shell_commands
+    # TODO: https://github.com/bazelbuild/rules_rust/issues/2079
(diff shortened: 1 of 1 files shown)


scentini (approved):

scentini:
Should we consider bumping up the min supported Bazel to 6.3?
https://github.com/bazelbuild/rules_rust/pull/2072 is also a change that fails on 6.0, albeit for that one we can ask to make it compatible with both 6.0 and 6.3.

UebelAndre:
> Should we consider bumping up the min supported Bazel to 6.3? #2072 is also a change that fails on 6.0, albeit for that one we can ask to make it compatible with both 6.0 and 6.3.

I'm actually not sure this should be merged. I saw [here](https://buildkite.com/bazel/rules-rust-rustlang/builds/9072#018993cc-4443-4449-904c-e2695fbd9a86) a job that didn't produce any coverage and it ran on 6.3.0. This issue feels like an issue in Bazel itself so I opened https://github.com/bazelbuild/bazel/issues/19098

PR #2087, opened 2023-07-27
Remove unnecessary eval to fix PATH variables containing a space

### LINKED ISSUE #2086: Running `crates_vendor` with a PATH containing a space fails
I end up seeing `env: ‘Text.app/Contents/SharedSupport/bin:/Applications/Visual’: No such file or directory`

This is caused by the following path entries:
```
/Applications/Sublime Text.app/Contents/SharedSupport/bin
/Applications/Visual Studio Code.app/Contents/Resources/app/bin
```

I've been able to narrow this down to the `_UNIX_WRAPPER` in `crate_universe/private/crates_vendor.bzl`. As far as I can tell, there's an issue with the shell dropping the quotes from the final expression, causing `PATH=/usr/bin:/Applications/Sublime Text.app/Contents/SharedSupport/bin` to be omitted when the eval is used.

You can see the underlying issue in the following:
```
$ TMP="/usr/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin"
$ echo $TMP
/usr/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin
$ eval printf "%s\n" PATH="${TMP}"
PATH=/usr/bin:/Applications/VisualnStudionCode.app/Contents/Resources/app/binn
$ eval 'printf "%s\n" PATH="${TMP}"'
PATH=/usr/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin
```

belak:
Fixes #2086

As far as I can tell, this `eval` is unnecessary at best, and breaks with certain edge cases, like `PATH` (or other referenced variables) which include a space.

I'm not very familiar with this repo, but if there's a way to add a test for this I'd be happy to do it.


UebelAndre (approved):
Thanks!

PR #2088, opened 2023-07-27
Look up CARGO_BAZEL_GENERATOR_PATH correctly

HackAttack:
The intent is to allow this variable to be passed in an --action_env, but ctx.var is not where these show up: ctx.configuration.default_shell_env is.


illicitonion (approved):
Thanks!

PR #2090, opened 2023-07-31
Linker script support for `rust_shared_library` and `cc_common.link`

scentini:


rust/private/rust.bzl
```diff
@@ -954,6 +954,12 @@ rust_shared_library = rule(
     implementation = _rust_shared_library_impl,
     attrs = dict(
         _common_attrs.items() + _experimental_use_cc_common_link_attrs.items() + {
+            "linker_script": attr.label(
+                doc = dedent("""\
+            Link script to forward into linker via rustc options.
```
krasimirgg, line 959: nit: could you indent this and the next line so that they both start at +4 spaces relative to where the `doc =` on the previous line starts at

rust/private/rustc.bzl
```diff
@@ -1347,6 +1347,15 @@ def rustc_compile_action(
             # Append the name of the library
             output_relative_to_package = output_relative_to_package + output_lib
 
+        linker_script = getattr(ctx.file, "linker_script", None)
+        if linker_script:
+            if toolchain.target_os == "windows":
+                fail("Linking with cc_common.link for windows is currently not supported")
+            else:
+                user_link_flags = ["-Wl,--version-script={}".format(linker_script.path)]
```
krasimirgg, line 1355: Are these user_link_flags necessary? Looking a bit into --version-script, https://sourceware.org/binutils/docs/ld/VERSION.html, it seems it's for a special case of linker scripts. From reading up there, naively I'd expect that the linker will automatically pick up linker scripts by just directly listing the linker script as an input to the linker command line invocation, so I'd expect naively that just passing it to the additional_inputs will have the same effect?
Otherwise if that doesn't work for some reason, could we see if passing this via `-T` works: https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html#:~:text=You%20may%20supply%20a%20command,use%20the%20%60%2DT'%20option.

krasimirgg (changes requested):

PR #2103, opened 2023-08-08
Option to generate extra targets with crate_universe

adsnaider:
This is similar to what is provided by cargo-raze's `extra_aliased_targets`. As an example, we use this feature with the cxx crate as the repo provides the headers needed in C++ sources. This feature goes hand in hand with `additive_build_file` as we can create the rule and generate the alias for that rule.


adsnaider:
Issue in question: https://github.com/bazelbuild/rules_rust/issues/2104

adsnaider:
Thanks @illicitonion, the use-case I have is cxx, and in particular, adding it as a regular third_party dependency doesn't work since autocxx depends on cxx so then I end up with duplicated versions of cxx which makes everything really sad. I went ahead and added an example that uses cxx (not autocxx though) in examples/crate_universe/using_cxx

crate_universe/private/crate.bzl
```diff
@@ -123,6 +124,7 @@ def _annotation(
         data (list, optional): A list of labels to add to a crate's `rust_library::data` attribute.
         data_glob (list, optional): A list of glob patterns to add to a crate's `rust_library::data` attribute.
         deps (list, optional): A list of labels to add to a crate's `rust_library::deps` attribute.
+        extra_aliased_targets (list, optional): A list of targets to add to the generated aliases.
```
illicitonion, line 127:
```suggestion
        extra_aliased_targets (list, optional): A list of targets to add to the generated aliases in the root crate_universe repository.
```

examples/crate_universe/WORKSPACE.bazel
```diff
@@ -269,6 +269,98 @@ load(
...
+                # file for C++ code to depend on.
+                cc_library(
+                    name = "cxx_cc",
+                    visibility = ["//visibility:public"],
+                    hdrs = ["include/cxx.h"],
+                    srcs = ["src/cxx.cc"],
+                    includes = ["include"],
+                )
+            """,
+            extra_aliased_targets = ["cxx_cc"],
```
illicitonion, line 294: I think it would be better if we took a dict here, where we map "what the alias in `@using_cxx//:BUILD.bazel` should be called" and map it to "the target in `@using_cxx__cxx-1.0.104//:BUILD.bazel`" it should alias for. It's a nice coincidence that your target here and how you want to expose it are the same, but that's unlikely to always be the case.

(Ideally we'd just add an `additive_build_file_content` for that `@using_cxx` BUILD.bazel file, but we can do that because we don't know the name mangling scheme and the version number, so "a mapping of stuff to push up" seems reasonable).

In your case here, this would be:
```starlark
extra_aliased_targets = {
  "cxx_cc": "cxx_cc",
}
```

but it opens up the ability for you to be more flexible about that mapping, so that someone could create:
```starlark
extra_aliased_targets = {
  "generated_openssl": "openssl_sys_build_script",
}
```

creating "generated_openssl" in the aliases, pointing at "openssl_sys_build_script" in the generated per-crate repo.

or something along those lines.

crate_universe/src/rendering.rs
```diff
@@ -164,6 +164,14 @@ impl Renderer {
                     tags: BTreeSet::from(["manual".to_owned()]),
                 });
             }
+
+            for extra_aliased_target in &krate.extra_aliased_targets {
+                dependencies.push(Alias {
```
illicitonion, line 169: We should check this `dependencies` `Vec` for duplicates after we've pushed these all in, to make sure there are no duplicates, and give a descriptive error message if we find any duplicates.

Before they happened to be de-duplicated by construction, but now that we're taking in user-input, we don't have that guarantee any more.

illicitonion (commented):
Thanks so much for adding the example, it makes much more clear what's going on and why!

I think this generally makes sense as a feature, but left a few pieces of feedback on the API surface - basically, let's make the list a dict, clean up a couple of things, and get this merged.

Thanks again for putting this together!

crate_universe/private/crate.bzl
adsnaider: Done.

examples/crate_universe/WORKSPACE.bazel
adsnaider: Done.

illicitonion (approved):
Looks great, thanks so much!

PR #2110, opened 2023-08-10
Remove `mandatory = True` from the `prost_plugin_flag` attribute in `rust_prost_toolchain`.

thb-sb:
Since `prost_plugin_flag` has a default value (`--plugin=protoc-gen-prost=%s`), we don't need to flag it as mandatory, otherwise people have to explicitely set it to the same value as the default one.


cameron-martin:
I can confirm this is a problem

scentini (approved):

PR #2112, opened 2023-08-10
Fix rust_doc_test dependency to allow either rust_library or cc_library

HackAttack:
It is documented to allow either, but as written, requires both.


scentini (approved):
Thanks!

PR #2116, opened 2023-08-15
Fix argfile length rustc bug (Windows)

tdbgamer:
Fix rustc bug on windows by symlinking all transitive deps into a linked directory. This only affects windows because `rustc` uses an environment variable to track all the paths passed as `-Ldependency=` on that platform and it silently overflows causing "no such crate" errors when compiling rust projects with large transitive dependency trees.


UebelAndre (changes requested):
Thanks! This seems reasonable to me but it seems that some test need to be updated in order to land this
```
(15:02:04) ERROR: C:/b/bk-windows-zh48/bazel/rules-rust-rustlang/test/unit/exports/BUILD.bazel:3:19: Analysis of target '//test/unit/exports:lib_exports_test' failed
```
Once they're updated I think it's mergable 😄

tdbgamer:
@UebelAndre Awesome! Yeah I need to come back to this and figure out why those tests fail. It looks like it expects the first action to always be a Rust sources but it's finding my symlinks instead? Not totally sure what's going on there.

UebelAndre:
> @UebelAndre Awesome! Yeah I need to come back to this and figure out why those tests fail. It looks like it expects the first action to always be some Rust artifact, but it's finding my symlinks instead? Not totally sure what's going on there.

that's what it looks like, the fix is probably to iterate over all actions and find the `Rustc` action. Maybe even assert you only found one.

walter-zeromatter:
FWIW I'm working on migrating our codebase to use rules_rust and just hit this bug

timbess:
@Walter-Reactor did this fix your issue as well? I dropped the ball on this cause our patch file worked, and Windows has become a bit less of a priority for us recently. If it worked for you, I'll try to fix up the test issues and get it merged.

walter-zeromatter:
I'm actually adopting https://github.com/dzbarsky/rules_rust/commit/f2f5b3d03d579b628dcbb3bb5808cb074e4ce6e6 instead, I like the angle it takes

UebelAndre:
So is this change still desired?

timbess:
I'll close this for now because I don't have a Windows machine to repro it on anymore, sorry!

PR #2118, opened 2023-08-16
Add crate_name attribute

daivinhtran:
This PR adds `crate_name` attribute to `rust_proto_library` and `rust_grpc_library` rules. This aligns with `crate_name` attribute provided in other rules to allow custom crate name different from label name.

Setting `--crate-name` flag to `rustc_flags` attribute is not an option because `rustc` does not allow `--crate-name` flag to be set twice.


scentini (approved):

scentini:
LGTM, waiting for @UebelAndre to review as per request.

scentini (approved):
Thanks!

PR #1311, opened 2022-05-03
Updated wasm_bindgen rules to only require core `rules_nodejs` rules

UebelAndre:
Since [rules_nodejs 4.4.2](https://github.com/bazelbuild/rules_nodejs/releases/tag/4.4.2) the project has been shipping a "core" artifact that contains the core requirements for `rules_nodejs` rules. The `wasm_bindgen` rules are currently only staffed to maintain this minimal contract. This change updates `rules_rust` to only require the `rules_nodejs` core artifact for the `wasm_bindgen` rules.

PR #2119, opened 2023-08-18
Update wasm_bindgen to not use deprecated/unmaintained JS rules

dzbarsky:
The providers the rule is using were deleted (I believe in https://github.com/bazelbuild/rules_nodejs/pull/3655).
It seems the community has standardized on https://github.com/aspect-build/rules_js as the future.

I removed the JS dependency from the bzl_lib as well, I didn't see an easy way to pull in the bzl_lib from providers, but please let me know if I missed something.

`cd examples && bazel test //wasm/...` works for me with this branch.

### DIFF
--- a/examples/WORKSPACE.bazel
+++ b/examples/WORKSPACE.bazel
@@ -79,15 +79,20 @@ rust_repository_set(
 # Examples dependencies
 ###############################################################################
 
-http_archive(
-    name = "build_bazel_rules_nodejs",
-    sha256 = "c78216f5be5d451a42275b0b7dc809fb9347e2b04a68f68bad620a2b01f5c774",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.5.2/rules_nodejs-5.5.2.tar.gz"],
+load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
+
+rules_js_dependencies()
+
+load("@rules_nodejs//nodejs:repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
+
+nodejs_register_toolchains(
+    name = "nodejs",
+    node_version = DEFAULT_NODE_VERSION,
 )
 
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
 
-node_repositories()
+bazel_features_deps()
 
 http_archive(
     name = "rules_foreign_cc",
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -7,9 +7,9 @@ package(default_visibility = ["//visibility:private"])
 bzl_library(
     name = "docs_deps",
     srcs = [
+        "@aspect_rules_js//js:providers",
         "@bazel_tools//tools:bzl_srcs",
         "@com_google_protobuf//:bzl_srcs",
-        "@rules_nodejs//nodejs:bzl",
     ],
     deps = [
         "@bazel_skylib//lib:paths",
--- a/docs/WORKSPACE.bazel
+++ b/docs/WORKSPACE.bazel
@@ -37,6 +37,14 @@ rust_wasm_bindgen_dependencies()
 
 rust_wasm_bindgen_register_toolchains()
 
+load("@aspect_rules_js//js:repositories.bzl", "rules_js_dependencies")
+
+rules_js_dependencies()
+
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
+bazel_features_deps()
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
--- a/examples/wasm/BUILD.bazel
+++ b/examples/wasm/BUILD.bazel
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_test")
+load("@aspect_rules_js//js:defs.bzl", "js_test")
 load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_shared_library")
 load("@rules_rust//wasm_bindgen:wasm_bindgen.bzl", "rust_wasm_bindgen")
 
@@ -65,7 +65,7 @@ rust_wasm_bindgen(
     wasm_file = ":hello_world_lib_wasm",
 )
 
-nodejs_test(
+js_test(
     name = "hello_world_wasm_test",
     data = [
         ":hello_world_bundler_wasm_bindgen",
--- a/wasm_bindgen/providers.bzl
+++ b/wasm_bindgen/providers.bzl
@@ -1,10 +1,8 @@
 """A module for re-exporting the providers used by the rust_wasm_bindgen rule"""
 
 load(
-    "@rules_nodejs//nodejs:providers.bzl",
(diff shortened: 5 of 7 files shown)


dzbarsky:
@UebelAndre @illicitonion I saw you messing with this previously in https://github.com/bazelbuild/rules_rust/pull/1311, could you review?

dzbarsky:
I see there is one CI job failing but I'm not sure what's up with the RBE ubuntu16 config, any ideas?

UebelAndre (changes requested):
Hi! Sorry for the delay in the review, thanks so much for putting this together! Do you know if it'd be possible to maintain both? I thought the `wasm_bindgen` rules only relied on the core `rules_nodejs` repo which is stated to be in use within `rules_js`.

dzbarsky:
`rules_js` only uses `rules_nodejs` to define the hermetic node toolchain. The JSInfo providers have changed and live in `rules_js`. From the `rules_nodejs` README: "This ruleset provides a NodeJS development toolchain and runtime with Bazel. It does not have any rules for using NodeJS, such as nodejs_binary. For that, we recommend [rules_js](https://github.com/aspect-build/rules_js)."

Concretely, here is the current state of the world:
- Using `wasm_bindgen` requires pinning to an old version of `rules_nodejs`, from before https://github.com/bazelbuild/rules_nodejs/pull/3655. This is now 6 months old and will only get older.
- The newest supported nodejs at that commit was 18.15. The current 18-series LTS is 18.16.1, and also node19/node20 are out now. While you can define your own custom node toolchain at newer versions with some gymnastics, it's not ergonomic.
- Node18 enters maintenance mode in a month: https://endoflife.date/nodejs

So if you want to use a newish nodejs version, you need to load your own (newer) `rules_nodejs`, which breaks `wasm_bindgen` unless you apply patches to `rules_rust` (basically what I submitted here). 

It seems our options are:
1. Do nothing, stay with current state. This prioritizes projects that are using the deprecated JS rules, others can apply a patch like this one.
2. Apply this patch. This prioritizes projects using the non-deprecated JS rules, legacy users can apply the reverse of this patch.

Personally I think (1) is better, especially because projects that are not upgrading their rulesets probably won't upgrade this one either. I'm also happy to continue carrying this patch locally (I have other rules_rust patches anyway), but I figured this would help the majority of users who are using bazel for JS.

Re: supporting both, since the issue here is a load of an undefined symbol, I'm not aware of a way to handle/recover that failure in Starlark, do you have any ideas? Failing that, I don't think it's feasible to support both.`

examples/WORKSPACE.bazel
```diff
@@ -79,15 +79,20 @@ rust_repository_set(
...
+
+load("@rules_nodejs//nodejs:repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
+
+nodejs_register_toolchains(
+    name = "nodejs",
+    node_version = DEFAULT_NODE_VERSION,
 )
 
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
```
UebelAndre, line 93: What is this? The repo seems to come out of nowhere. Is it required to use `aspect_rules_js`?

dzbarsky: tl;dr This is a bit of a wart, yes.

The package we load the JsInfo provider from depends on that repo to define a config setting: https://github.com/aspect-build/rules_js/blob/63862adcfb565b0a53fb45c60dded0f8a7893c49/js/BUILD.bazel#L3C1-L15 This seems designed to support multiple Bazel versions so hopefully it can go away at some point. That repo also has a small usability trick - it automatically calls `bazel_features_deps` if you use `npm_translate_lock`, and since the vast majority of JS projects use npm packages, they don't have to do anything special here.

I do agree it's a bit weird to have to pull this in just to use the provider, @alexeagle perhaps you can consider rearranging the layout to make the providers more standalone?

dzbarsky: @UebelAndre What do you think?

alexeagle: Hm, yeah I see why this is undesirable. We had to make those part of the public API in https://github.com/aspect-build/rules_js/pull/1187 - but maybe rules_rust can load() the providers without having to load that package? providers.bzl is in its own file for this reason.

dzbarsky: The problem is the `//docs:docs_deps` rule in this repo. The bzl_library requires depending on `"@aspect_rules_js//js:providers"` to satisfy the stardoc rules/lint, which means anything loaded in `//js:BUILD` is a dependency. I think if you moved the providers.bzl to its own package (say `//js/providers/defs.bzl`) it would work.

alexeagle: The bzl_library here can just depend on the bzl file directly in srcs to avoid that

dzbarsky: Not sure I'm following the suggestion, do you mean like so?
```
bzl_library(
    name = "docs_deps",
    srcs = [
        "@aspect_rules_js//js:providers.bzl",
        "@bazel_tools//tools:bzl_srcs",
        "@com_google_protobuf//:bzl_srcs",
    ],
   deps = [...]
)
```

That hits visibility issues: https://github.com/aspect-build/rules_js/blob/main/js/BUILD.bazel#L6-L9
Even ignoring visibility, it will have a problem with missing the js_info bzl
```
com.google.devtools.build.skydoc.SkydocMain$StarlarkEvaluationException: File /private/var/tmp/_bazel_zbarsky/e27223f43abb18db045469720c4d6593/sandbox/darwin-sandbox/28/execroot/rules_rust_docs/bazel-out/darwin-opt-exec-2B5CBBC6/bin/rust_repositories_md_stardoc.runfiles/aspect_rules_js/js/providers.bzl imported '//js/private:js_info.bzl', yet /private/var/tmp/_bazel_zbarsky/e27223f43abb18db045469720c4d6593/sandbox/darwin-sandbox/28/execroot/rules_rust_docs/bazel-out/darwin-opt-exec-2B5CBBC6/bin/rust_repositories_md_stardoc.runfiles/aspect_rules_js/js/private/js_info.bzl was not found.
```

dzbarsky: @UebelAndre ping? What can I do to get this PR moving?

UebelAndre: I’ll try to make time this weekend to do another pass. Sorry for the delay!

UebelAndre:
I have not forgotten about this change. I have uses of these rules that are still reliant on `rules_nodejs` and this would be a breaking change to projects that are unfortunately extremely hard to migrate. I feel a good path forward is to separate the `rust_wasm_bindgen` rules from any external implementation of javascript bazel rules. Maybe it'd be good to have subdirectories in `//wasm_bindgen` for `rules_js` and `rules_nodejs` that provide the interface to each ruleset. If I can find the time I will incorporate these changes into something like this. Admittedly my behavior here is something I strongly dislike where a single maintainer prevents community progress. For that I apologize. https://github.com/UebelAndre/rules_rust/tree/wasm/wasm_bindgen is the branch I'd be using to make these changes in case it provides any insights. If the code paths for raw wasm-bindgen, rules_js, and rules_nodejs were independent then I think there would be absolutely no hesitation to accept any change to the rules_js section as I understand that to be an active community.

dzbarsky:
> I have not forgotten about this change

No problem, thanks for the heads up! The approach you're taking in the branch seems reasonable to me

UebelAndre:
Closing in favor of https://github.com/bazelbuild/rules_rust/pull/2284

PR #2124, opened 2023-08-23
Read default annotation values from package's Cargo.toml metadata

dtolnay:
This PR allows Cargo package authors to specify metadata that replaces the need for crate_universe users to write `annotations`.

Crate_universe will pick up default values for various `annotations` fields from the metadata provided by the package author. It still remains possible for the crate_universe user to write their own `annotations` for the crate anyway, in which case these override the default values distributed by the package author. 

For example the `cxx` crate contains some C++ source code that must get linked by the user. To accomplish this, all users of the crate from Bazel would _previously_ have needed to fill in the following annotation:

```bzl
crates_repository(
    name = "...",
    annotations = {
        "cxx": [crate.annotation(
            additive_build_file_content = """
            cc_library(
                name = "cxx_cc",
                srcs = ["src/cxx.cc"],
                hdrs = ["include/cxx.h"],
                include_prefix = "rust",
                includes = ["include"],
                linkstatic = True,
                strip_include_prefix = "include",
            )
            """,
            extra_aliased_targets = {"cxx_cc": "cxx_cc"},
            gen_build_script = False,
        )],
    },
    ...
)
```

Now `cxx` ships this annotation as part of [the package metadata in its Cargo.toml](https://github.com/dtolnay/cxx/blob/1.0.105/Cargo.toml#L50-L64), and Bazel users can pull in `cxx` without handwriting their own `crate.annotation`.

```toml
[package.metadata.bazel]
additive_build_file_content = """
cc_library(
    name = "cxx_cc",
    srcs = ["src/cxx.cc"],
    hdrs = ["include/cxx.h"],
    include_prefix = "rust",
    includes = ["include"],
    linkstatic = True,
    strip_include_prefix = "include",
)
"""
extra_aliased_targets = { cxx_cc = "cxx_cc" }
gen_build_script = false
```


crate_universe/src/config.rs
```diff
@@ -335,6 +335,87 @@ impl Sum for CrateAnnotations {
     }
 }
 
+/// A subset of `crate.annotation` that we allow packages to define in their
+/// free-form Cargo.toml metadata.
+///
+///     [package.metadata.bazel]
+///     additive_build_file_contents = """
+///         ...
+///     """
```
dtolnay, line 344: I considered a more TOML-native form for `additive_build_file_content`, but just using multiline string literal containing Starlark seems all right.

Support for the structured form can be introduced separately if desired.

```toml
[package.metadata.bazel]
extra_aliased_targets = { cxx_cc = "cxx_cc" }
gen_build_script = false

[[package.metadata.bazel.additive_build_file_content]]
[package.metadata.bazel.additive_build_file_content.cc_library]
name = "cxx_cc"
srcs = ["src/cxx.cc"]
hdrs = ["include/cxx.h"]
include_prefix = "rust"
includes = ["include"]
linkstatic = true
strip_include_prefix = "include"
visibility = ["//visibility:public"]
```

UebelAndre (changes requested):
This is a pretty awesome change! Could you also add a unit test for this? I'm thinking a `Cargo.toml` file somewhere in [crate_universe/test_data](https://github.com/bazelbuild/rules_rust/tree/0.26.0/crate_universe/test_data) and have a test that reads the `[package.metadata.bazel]` section?

dtolnay:
I made a couple attempts at a test, but I haven't been able to figure out how to get crate_universe to pay attention to local `path` dependencies. It seems to only want to generate targets for crates.io dependencies (`version`) and `git`.

Here is what I have so far: affc1b5602e9e6231d27f8d05b329ff2d41c2416.

crate_universe/src/metadata/metadata_annotation.rs
```diff
@@ -395,18 +395,20 @@ impl Annotations {
                         extra
                     })
                     .cloned()
-                    .collect();
+                    .sum();
 
-                if !extras.is_empty() {
+                crate_extra.apply_defaults_from_package_metadata(&pkg.metadata);
```
UebelAndre, line 400: I was thinking that the testing would just be in this file and cover the changes around here. Not necessarily a brand new test target. I think that'd get sufficient coverage of the functionality, right?

UebelAndre (changes requested):

crate_universe/src/metadata/metadata_annotation.rs
dtolnay: Ah, I get it now. Good call. I have added that test.

UebelAndre (approved):
Fantastic! Thank you so much!

PR #2126, opened 2023-08-24
Advertise CcInfo provider on rules

### LINKED ISSUE #2101: Libraries not linked by cc_shared_library
Since Bazel 6.3.0, the aspect used in `cc_shared_library` only propagates to dependencies that advertise themselves as providing `CcInfo` via the `provides` attribute of the rule constructor. `rust_static_library` does not do this, so these libraries are not linked by `cc_shared_library`.

cameron-martin:
The aspect used in the `cc_shared_library` implementation in bazel 6.3 and 7 only traverses attributes that advertise the `CcInfo` provider via the `provides` attribute. This means in these versions of bazel, rust libraries are currently omitted from libraries built with `cc_shared_library`.

Fixes #2101.


cameron-martin:
It seems that the `CcInfo` provider isn't returned when the target arch is wasm, which causes an error in this case. However, you don't have this information when creating the rule. What do you think the best way around this is? Is there a reason why it can't return a `CcInfo` provider in the wasm case?

cameron-martin:
Looks like the issue with wasm not emitting CcInfo is fixed by #1979

test/cc_shared_library/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+cc_library(
+    name = "c_lib",
+    srcs = ["lib.c"],
+    hdrs = ["lib.h"],
+    deps = [":rust_lib"],
+)
+
+# Tests that cc_shared_library correctly traverses into
+# `rust_static_library` when linking.
+cc_shared_library(
```
UebelAndre, line 19: Will this lead to the same issue as https://github.com/bazelbuild/rules_rust/issues/2359 if `experimental_cc_shared_library` is not set?

UebelAndre (changes requested):
Thanks! I think this looks fine but I had one question.

test/cc_shared_library/BUILD.bazel
cameron-martin: Do you mean if people are querying it as an external repository, where the root workspace does not have this set?

UebelAndre: yeah

cameron-martin: I suppose it will, although in 7.0.0 this is enabled by default so it shouldn't be an issue with the latest version.

UebelAndre: @illicitonion given that the issue was referring to a cquery within the `rules_rust` repo, do you think this is an acceptable change? I do like the `procides` changes and love the test. I just don't wanna immediately regress something here.

illicitonion: I sent out https://github.com/bazelbuild/rules_rust/pull/2398 which add tests for what I think we should be aiming to support (i.e. `bazel query ...` and `bazel cquery ...` from inside the `rules_rust` repo) - if we have more things we want to support (e.g. `bazel query @rules_rust//...` from a repo that depends on `rules_rust`) , we should add tests to CI for them.

This PR passes the tests I sent out, so I'm happy :)

UebelAndre:
Sounds good. Just needs a rebase!

PR #2133, opened 2023-08-25
Add target_compatible_with from kwargs to build_script_kwargs

daivinhtran:
The top-level `_build_script_run` target should also have `target_compatible_with` in order to work with platforms compatibility.


UebelAndre (changes requested):
The way args are passed here seems backward to me. I think target folks think about when defining `cargo_build_script` targets is actually `_build_script_run`. Thus we should default to passing all args to that and set the `rust_binary` target to manual. You can then constrain the target with both `target_compatible_with` which I feel most applies to running the script where `exec_compatible_with` would apply to compiling the script as the script is in `cfg = "exec"`. How do you feel about passing kwargs to `_build_script_run` instead?

cc @scentini

daivinhtran:
@UebelAndre Yes that's an approach I would prefer. Passing `kwargs` to `_build_script_run` make sense.

PTAL: https://github.com/bazelbuild/rules_rust/pull/2133/commits/3385c97f454fdaaa1c92f49fa382f36cca025759

I also introduced `edition` and `srcs` attributes to `cargo_build_script` macro so they can be passed to `rust_binary` to build.

UebelAndre (approved):
Seems good to me! I’ll give @scentini a chance to look since I think they have some more nuanced uses with “compatible_with”.

scentini (approved):

PR #2136, opened 2023-08-26
Use _validation output group for rustfmt/clippy

dzbarsky:
This is a bazel6 feature, see https://bazel.build/extending/rules#validation-actions
Since Bazel6 is the current LTS and bazel7 is coming soon, I think the support policy allows us to drop Bazel5 support, is that right?

This has 2 benefits:
- Bazel knows these actions are not critical to the build and can schedule them better (towards the end, more parallelism)
- Users only need to pass one flag per aspect to enable it

I wasn't sure about the change to the `rust_clippy` rule, I couldn't figure out from the examples in the repo if it was working correctly or not (both with my change and on main). Hopefully it has tests?


UebelAndre (changes requested):
Is this something that could perhaps be behind an experimental flag?

dzbarsky:
> Is this something that could perhaps be behind an experimental flag?

I suppose so, are you envisioning a setup where we keep the current output groups and also add the validation output groups (when a flag is enabled)? I think that should work, but it doesn't seem like a big win for users since they'll need to enable this flag instead of enabling a different flag for the output group. I guess it lets us split the rollout over a few releases, do you think folks tend to flip the experimental flags on when taking upgrades?

Issue #2120, opened 2023-08-21
 Error: 'apple_common' value has no field or method 'multi_arch_split'


sgowroji:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3252#018a15dd-b383-4df3-bffa-8839e4ec4fb6

Platform : MacOS

Logs:
```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_buildkite/dfc7b38205384553a1c6cabe90c472e2/external/build_bazel_rules_apple/apple/internal/rule_support.bzl", line 217, column 36, in <toplevel>
		deps_cfg = apple_common.multi_arch_split,
Error: 'apple_common' value has no field or method 'multi_arch_split'
(04:22:57) WARNING: Target pattern parsing failed.
(04:22:57) ERROR: Skipping '//...': error loading package under directory '': error loading package '': at /private/var/tmp/_bazel_buildkite/dfc7b38205384553a1c6cabe90c472e2/external/build_bazel_rules_apple/apple/macos.bzl:33:5: at /private/var/tmp/_bazel_buildkite/dfc7b38205384553a1c6cabe90c472e2/external/build_bazel_rules_apple/apple/internal/macos_binary_support.bzl:50:5: initialization of module 'apple/internal/rule_support.bzl' failed
(04:22:57) ERROR: error loading package under directory '': error loading package '': at /private/var/tmp/_bazel_buildkite/dfc7b38205384553a1c6cabe90c472e2/external/build_bazel_rules_apple/apple/macos.bzl:33:5: at /private/var/tmp/_bazel_buildkite/dfc7b38205384553a1c6cabe90c472e2/external/build_bazel_rules_apple/apple/internal/macos_binary_support.bzl:50:5: initialization of module 'apple/internal/rule_support.bzl' failed

```

Fix is to update it with rules_apple once [this](https://github.com/bazelbuild/rules_apple/issues/2035#issuecomment-1663943174
) is fixed.

CC Greenteam @SalmaSamy

keith:
https://github.com/bazelbuild/rules_rust/pull/2137

PR #2137, opened 2023-08-28
Update rules_apple

keith:
Fixes https://github.com/bazelbuild/rules_rust/issues/2120


keith:
I don't think the CI failures are related 🤔

UebelAndre (approved):

UebelAndre:
The failure is not related… Bazel seems to inconsistently produce coverage reports and thus we have flakey CI.

Issue #1780, opened 2023-01-17
Build fail when depending on a library in a different repo that mixes generated and non-generated sources


dmitrii-ubskii:
rules_rust v0.16.1, Bazel v5.1.1 & v6.0.0.

Scenario: a common library has bindings for a number of different languages, one of which is Rust. `rust/BUILD` in the library defines a rule to generate the bindings as well as the `rust_library` rule that includes some helpful wrapper functions. 

```bazel
# @lib//rust:BUILD
load("@rules_rust//rust:defs.bzl", "rust_library")

genrule(
    name = "gen",
    cmd = "echo 'pub fn hello() { println!(\"Hello, world!\"); }' > $(@D)/gen.rs",
    outs = ["src/gen.rs"],
)

rust_library(
    name = "lib-with-gen",
    srcs = ["src/lib.rs", ":gen"],
    visibility = ["//visibility:public"],
)
```

Another repository wants to use the bindings as a library dependency. We import the `git_repository` as `lib` and attempt to use `@lib//rust:lib-with-gen` as a dependency. 

```bazel
# @bin//:BUILD
load("@rules_rust//rust:defs.bzl", "rust_binary")

rust_binary(
    name = "bin",
    srcs = ["src/main.rs"],
    deps = ["@lib//rust:lib-with-gen"],
)
```

However, the build fails:

```
...$ bazel build //:bin
ERROR: /private/var/tmp/.../external/lib/rust/BUILD:9:13: in rust_library rule @lib//rust:lib-with-gen: 
Traceback (most recent call last):
        File "/private/var/tmp/.../external/rules_rust/rust/private/rust.bzl", line 197, column 32, in _rust_library_impl
                return _rust_library_common(ctx, "rlib")
        File "/private/var/tmp/.../external/rules_rust/rust/private/rust.bzl", line 253, column 42, in _rust_library_common
                srcs, crate_root = _transform_sources(ctx, ctx.files.srcs, getattr(ctx.file, "crate_root", None))
        File "/private/var/tmp/.../external/rules_rust/rust/private/rust.bzl", line 172, column 68, in _transform_sources
                src_symlink = ctx.actions.declare_file(paths.relativize(src.short_path, package_root))
        File "/private/var/tmp/.../external/bazel_skylib/lib/paths.bzl", line 186, column 17, in _relativize
                fail("Path '%s' is not beneath '%s'" % (path, start))
[...]

PR #2138, opened 2023-08-29
Handle files from external repos (fix breaking change in Bazel 3.7.0)

jgsogo:
Fixes https://github.com/bazelbuild/rules_rust/issues/1780

In Bazel 3.7.0, the path returned by `build_file_path` changed (https://github.com/bazelbuild/bazel/issues/12344), it no longer contains the `external/<repo>` prefix for external files. This breaks the expectations of this rule in the [`paths.relativize` function](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/paths_doc.md#pathsrelativize) call.

In this PR I prepend the `workspace_root` (`external/<repo>`) to `build_file_path` (`<path/in/the/repo>`), and use `path` (`external/<repo>/path/in/the/repo`) instead of `short_path` (`../<repo>/path/in/the/repo`).

For files that are generated in different configurations I need to strip the `root` which contains the `bazel-out/darwin-fastbuild-ST-a1a0f4088294/bin/` that will never appear in the `package_root` path.

~Note.- I will probably need some help to add tests for this change~ I've tried to add one test


jgsogo:
I think the CI failure is not related to these changes, it fails to create a symlink for a `.rlib` because of "Permission denied":

```
(09:36:34) ERROR: C:/b/dz26jhdy/external/rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools/BUILD.bazel:89:15: Creating symlink bazel-out/x64_windows-fastbuild/bin/external/rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools/rust_toolchain/lib/rustlib/x86_64-pc-windows-msvc/lib/libcfg_if-7d7ae0874b44dfcc.rlib failed: failed to create symbolic link 'bazel-out/x64_windows-fastbuild/bin/external/rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools/rust_toolchain/lib/rustlib/x86_64-pc-windows-msvc/lib/libcfg_if-7d7ae0874b44dfcc.rlib' to 'external/rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools/lib/rustlib/x86_64-pc-windows-msvc/lib/libcfg_if-7d7ae0874b44dfcc.rlib' due to I/O error: C:/b/dz26jhdy/execroot/__main__/bazel-out/x64_windows-fastbuild/bin/external/rust_windows_x86_64__x86_64-pc-windows-msvc__stable_tools/rust_toolchain/lib/rustlib/x86_64-pc-windows-msvc/lib/libcfg_if-7d7ae0874b44dfcc.rlib (Permission denied)
```

rust/private/rust.bzl
```diff
@@ -154,10 +154,11 @@ def _transform_sources(ctx, srcs, crate_root):
     generated_sources = []
 
     generated_root = crate_root
-    package_root = paths.dirname(ctx.build_file_path)
+    package_root = paths.dirname(paths.join(ctx.label.workspace_root, ctx.build_file_path))
 
     if crate_root and (crate_root.is_source or crate_root.root.path != ctx.bin_dir.path):
-        generated_root = ctx.actions.declare_file(paths.relativize(crate_root.short_path, package_root))
+        crate_root_path = paths.relativize(crate_root.path, crate_root.root.path)
```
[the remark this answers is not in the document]
jgsogo: Honestly, I find my approach easier to read without the `if/else` and no hardcoded strings (very unlikely, but they might change in the future). Maybe I can rename the variable `crate_root_path` to `crate_root_rel_path` to make it a bit more explicit? 

In any case, no strong opinion here. If you feel like your approach is easier to read and maintain, I'm fine. You probably have more Bazel experience than me.

jgsogo:
I moved the logic to another function and added some description to it with a (hopefully) good enough comment. Thanks a lot for the reviews!

UebelAndre (changes requested):
I can take another look later today but I’m wondering why this regression wasn’t caught. Is this fixing behavior for Bazel 3? Is there some kind of regression test that can be added as part of this PR?

jgsogo:
> I can take another look later today but I’m wondering why this regression wasn’t caught. Is this fixing behavior for Bazel 3? Is there some kind of regression test that can be added as part of this PR?

I think this has never been tested. The breaking change was reported in 2020 (https://github.com/bazelbuild/bazel/issues/12344), and it looks like there are not so many people using `rules_rust` to consume Rust targets from external repositories that involve generated files 🤷 

In order to test this we would need to use an external repository. I don't have enough Bazel knowledge to know if this can be reproduced without actually using another repository (maybe using another `WORKSPACE` in this same repo?)

scentini:
Yeah, I introduced the code initially, and coming from a monorepo, external dependencies didn't come to mind.

There is no way to test it without external dependencies, however we have examples of using external dependencies under `//test/cc_common_link` and `test/cc_common_link/with_global_alloc`

jgsogo:
Thanks for the pointers, @scentini . I've added one basic test to validate the new implementation (I've checked it fails with previous logic).

rust/private/rust.bzl
```diff
@@ -127,6 +127,34 @@ def get_edition(attr, toolchain, label):
     else:
         return toolchain.default_edition
 
+def _symlink_for_non_generated_source(ctx, src_file, package_root):
+    """Creates and returns a symlink for non-generated source files.
+
+    This rule uses the full path to the source files and the rule directory to compute
+    the relative paths. This is needed, instead of using `short_path`, because of non-generated
+    source files in external repositories returning relative paths.
```
UebelAndre, line 135:
```suggestion
    source files in external repositories possibly returning relative paths depending on the current version of Bazel.
```

rust/private/rust.bzl
```diff
@@ -127,6 +127,34 @@ def get_edition(attr, toolchain, label):
...
+        package_root (File): The full path to the directory containing the current rule.
+
+    Returns:
+        File: The created symlink if a non-generated file, or the file itself.
+    """
+
+    if src_file.is_source or src_file.root.path != ctx.bin_dir.path:
+        src_short_path = paths.relativize(src_file.path, src_file.root.path)
+        src_symlink = ctx.actions.declare_file(paths.relativize(src_short_path, package_root))
+        ctx.actions.symlink(
```
UebelAndre, line 149: Small nit, can you put the action to symlink back in `_transform_sources`? The thing I wanted to make crystal clear is what is happening with the path manipulation here.

.bazelci/presubmit.yml
```diff
@@ -575,6 +575,14 @@ tasks:
     test_flags:
       - "--@rules_rust//rust/toolchain/channel=nightly"
       - "--@rules_rust//:no_std=alloc"
+  generated_inputs_external_repo_ubuntu2004:
```
UebelAndre, line 578: Instead of adding a new job, could you not instead make a repository rule that writes a small target to an external repo and add it to https://github.com/bazelbuild/rules_rust/blob/0.27.0/test/deps.bzl ?

UebelAndre (changes requested):

rust/private/rust.bzl
jgsogo: Added here: https://github.com/bazelbuild/rules_rust/pull/2138/commits/a8911831549e237e784c52edd9dba9c1dcb44333

.bazelci/presubmit.yml
jgsogo: I think I did the right thing here https://github.com/bazelbuild/rules_rust/pull/2138/commits/8d0b467a0d2c5d6b392adcbb433871a829f17fa9. 

FYI. If the rust library is created in the root of the repository, then current implementation doesn't fail.

.gitignore
```diff
@@ -11,6 +11,7 @@
 /examples/crate_universe/*/bazel-*
 /test/cc_common_link/bazel-*
 /test/cc_common_link/with_global_alloc/bazel-*
+/test/generated_inputs/external_repo/bazel-*
```
UebelAndre, line 14: is this still needed?

UebelAndre (changes requested):

.gitignore
jgsogo, line 14:
```suggestion
```

jgsogo: Sorry, it was a leftover from the previous testing approach.

test/generated_inputs/BUILD.bazel
```diff
@@ -178,3 +179,9 @@ bool_flag(
     name = "change_cfg",
     build_setting_default = False,
 )
+
+rust_test(
+    name = "generated_inputs_external_repo_test",
+    crate = "@generated_inputs_in_external_repo//lib:generated_inputs_external_repo",
```
UebelAndre, line 185:
```suggestion
    # This is regression testing a specific failure case for generated files _not_ in the root
    # of an external repository.
    crate = "@generated_inputs_in_external_repo//lib:generated_inputs_external_repo",
```

UebelAndre (approved):
Awesome work, thank you so much!

jgsogo:
It's been a pleasure. Looking forward to the next release. Thanks for the hard work on this project 🙌

PR #2139, opened 2023-08-29
crate_universe: Allow platform-specific build script env vars

illicitonion:
My use-case here is needing to point at different paths for libclang.so when using bindgen on different platforms.

This allows specifying build script env vars in annotations like:

```starlark
build_script_env = {
  "BORING_BSSL_PATH": "$(execpath @//third_party/boringssl:gen_dir)",
  "LIBCLANG_PATH": {
      "x86_64-unknown-linux-gnu": "$(execpath @libclang_linux_x86_64//file:libclang.so)",
  },
},
```

which will unconditionally set `$BORING_BSSL_PATH`, but will only set `$LIBCLANG_PATH` on x86-64 Linux.

We could in theory support this kind of construct for all annotation fields, but I've only added it where I happened to need it for now.


UebelAndre (changes requested):
Thanks! Seems reasonable to me. Could you show an example of what the rendered select statement looks like?

illicitonion:
> Thanks! Seems reasonable to me. Could you show an example of what the rendered select statement looks like?

Absolutely; here's my example:
```starlark
    build_script_env = select({
        "@rules_rust//rust/platform:x86_64-unknown-linux-gnu": {
            "BORING_BSSL_PATH": "$(execpath @//third_party/boringssl:gen_dir)",
            "LIBCLANG_PATH": "$(execpath @libclang_linux_x86_64//file:libclang.so)",  # x86_64-unknown-linux-gnu
        },
        "//conditions:default": {
            "BORING_BSSL_PATH": "$(execpath @//third_party/boringssl:gen_dir)",
        },
    }),
```

UebelAndre:
What produces the default condition? Will one always be produced?

dzbarsky:
Any chance we could get https://github.com/bazelbuild/rules_rust/pull/2119 in too? Would love to drop my local patches

illicitonion:
> What produces the default condition? Will one always be produced?

I added some unit tests for `Select{Dict,List}` in 3b08ec8ce80f4014e2161154262cf148df6d2acf which hopefully makes this more clear :)

UebelAndre (approved):
Fantastic! Awesome change! Maybe you could also add something to docs so folks know about this feature? But I leave that up to you 😄

PR #2144, opened 2023-08-31
Add android in runtime linking support check

gferon:
Android (like `linux` and `darwin`) supports runtime linking just fine. When building a library for Flutter with `rust_shared_library`, I hit the assert which is incorrect.

Another question: maybe the check is entirely unnecessary, and could thus be removed, as the comment references Bazel 4.1? (does `rules_rust` still support 4.x?)


UebelAndre (changes requested):
Thanks for the change! For android though, maintainers of `rules_rust` don't really develop for Android and thus lack context and experience to effectively maintain it's use with rules_rust. That being said, is there some kind of regression test that can be added to [examples/android](https://github.com/bazelbuild/rules_rust/tree/0.27.0/examples/android) that could act as a regression test?

> Another question: maybe the check is entirely unnecessary, and could thus be removed, as the comment references Bazel 4.1? (does rules_rust still support 4.x?)

I think our min target now is `6.0`. If this could be cleaned up then I think that'd be good.

gferon:
> maintainers of rules_rust don't really develop for Android

Makes a lot of sense, I think we could remove this test entirely because it really does apply to older versions of Bazel. I'll try to think of a regression test that makes `rules_rust` fail in this case. 

To be honest, `rules_rust` works absolutely fine with the latest Android NDK, because it uses transitions and platforms as one would expect with Bazel 6+. The only problem that I think is worth fixing is in your CI: `ANDROID_NDK_HOME` is set to /opt/android-ndk-r15c` which is both old (from 2017) and unsupported. If you can point me in the right direction, I can bump it to a newer one (it's already mounted in a directory by Buildkite it seems) and fix the errors with the example.

gferon:
@UebelAndre we are still running with this patch and it works very well. I see you approved it, is there anything else I can bring so you can merge it?

PR #2150, opened 2023-09-04
Bump min Bazel version to 6.3

scentini:


.bazelci/presubmit.yml
```diff
@@ -221,15 +221,15 @@ tasks:
   ubuntu1804:
     name: "Min Bazel Version"
     # If updating the minimum bazel version, please also update /docs/index.md
```
illicitonion, line 223: TODO :)

illicitonion (commented):
Can you document in the PR description what motivated the change?

illicitonion (commented):
LGTM, thanks! Is the update because of new coverage stuff? Could you describe / link to the changes we depend on in the PR description?

scentini:
Unfortunately I think coverage issues are present also in 6.3 and onwards https://github.com/bazelbuild/bazel/issues/19098. I updated the PR description though.

illicitonion (approved):
Thanks!

PR #2152, opened 2023-09-04
cargo_build_scripts can be run from custom directories

illicitonion:
While running from CARGO_MANIFEST_DIR is the simplest thing for compatibility, there are cases where a cargo build script may be easier to run from the exec root as most bazel actions do (e.g. where a C++ toolchain specifies in-repo include paths).

This mirrors the `rundir` attribute of `go_test`, which has similar concerns: https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/rules.md#go_test-rundir


cargo/private/cargo_build_script_wrapper.bzl
```diff
@@ -93,6 +94,11 @@ def cargo_build_script(
         data (list, optional): Files needed by the build script.
         tools (list, optional): Tools (executables) needed by the build script.
         links (str, optional): Name of the native library this crate links against.
+        rundir (str, optional): A directory to cd to before the cargo_build_script is run. This should be a path relative to the exec root.
```
UebelAndre, line 97:
```suggestion
        rundir (str, optional): A directory to `cd` to before the cargo_build_script is run. This should be a path relative to the exec root.
```

crate_universe/.idea/vcs.xml
UebelAndre, line 1: Is this intended to be committed?

illicitonion: Oops, no, removed!

UebelAndre (approved):
Sweet!

PR #2161, opened 2023-09-15
Refactor CrateInfo construction

daivinhtran:
## Description

This PR addresses https://github.com/bazelbuild/rules_rust/issues/2013 to ensure rust-analyzer can access all the env vars the rust rules pass to `rustc` at compile time. This is a large refactoring in almost all the rules and aspects so I aim this PR to just fix `rust_library` rule. The follow-up PR will address the remaining rules and aspects.

## Summary
* Create `create_crate_info_dict` function in `rust/private/utils.bzl` to create a mutable dict repsenting CrateInfo
* Move `_determine_lib_name`, `get_edition`, `_transform_sources` functions to `rust/private/utils.bzl` to avoid cyclic dependency when `create_crate_info_dict` function use these
* Introduce optional `create_crate_info_callback` attribute to `rustc_compile_action` to allow creating `CrateInfo` inside `rustc_compile_action`. This optional attribute allows scoping the refactoring in this PR to just `rust_library` rule.
* Introduce optional `skip_expanding_rustc_env` attribute to `rustc_compile_action` to skip expanding `rustc_env` attr. This is useful for `clippy` aspect and `rust_test` rule because the `CrateInfo` provided from the depended `rust_library` already expands all the env vars before returning the provider downstream.

## Notes
* `rustc_env_attr` is a temporary field in `CrateInfo` and needed by the rules and aspects not migrated to create `CrateInfo` in `rustc_compile_action` yet. It will be remove in the next PR after `CrateInfo.rustc_env` is always expanded.
* `create_crate_info_callback` will be removed from `rustc_compile_action` after all `CrateInfo`s are created inside `rustc_compile_action`.


rust/private/rustc.bzl
```diff
@@ -1075,32 +1082,219 @@ def construct_arguments(
 
     return args, env
 
+def get_edition(attr, toolchain, label):
+    """Returns the Rust edition from either the current rule's attirbutes or the current `rust_toolchain`
```
dzbarsky, line 1086: nit: "attirbutes"

rust/private/rustc.bzl
```diff
@@ -1075,32 +1082,219 @@ def construct_arguments(
...
+        toolchain (rust_toolchain): The `rust_toolchain` for the current target
+        label (Label): The label of the target being built
+
+    Returns:
+        str: The target Rust edition
+    """
+    if getattr(attr, "edition"):
+        return attr.edition
+    elif not toolchain.default_edition:
+        fail("Attribute `edition` is required for {}.".format(label))
```
dzbarsky, line 1099: Preexisting, but the first time I hit this error it wasn't obvious that you could set a default edition in the toolchain, might be worth a mention in this error?

rust/private/rustc.bzl
```diff
@@ -1075,32 +1082,219 @@ def construct_arguments(
...
+        )
+        return src_symlink
+    else:
+        return src_file
+
+def transform_sources(ctx, srcs, crate_root):
+    """Creates symlinks of the source files if needed.
+
+    Rustc assumes that the source files are located next to the crate root.
+    In case of a mix between generated and non-generated source files, this
```
dzbarsky, line 1136: I think we want to snip "this" at the end of this line?

rust/private/rustc.bzl
```diff
@@ -1075,32 +1082,219 @@ def construct_arguments(
...
+    Args:
+        ctx (struct): The current rule's context.
+        srcs (List[File]): The sources listed in the `srcs` attribute
+        crate_root (File): The file specified in the `crate_root` attribute,
+                           if it exists, otherwise None
+
+    Returns:
+        Tuple(List[File], File): The transformed srcs and crate_root
+    """
+    has_generated_sources = len([src for src in srcs if not src.is_source]) > 0
```
dzbarsky, line 1151: nit: `any([src for src in srcs if not src.is_source])` is shorter and you can bail out a little quicker?

rust/private/rustc.bzl
daivinhtran: I'm leaving your comments unresolved in this PR and will address in a separate PR because they're not within scope.

rust/private/rustc.bzl
daivinhtran: I'm leaving your comments unresolved in this PR and will address in a separate PR because they're not within scope.

rust/private/rustc.bzl
daivinhtran: I'm leaving your comments unresolved in this PR and will address in a separate PR because they're not within scope.

scentini (approved):
Thanks a lot!!

daivinhtran:
@scentini #2188 is the follow-up PR. PTAL!

PR #2166, opened 2023-09-21
rustdoc: respect error_format config

goffrie:
This makes it easier to make a test ensuring rustdoc doesn't emit warnings.

### DIFF
--- a/rust/private/rustdoc.bzl
+++ b/rust/private/rustdoc.bzl
@@ -321,6 +321,9 @@ rust_doc = rule(
             cfg = "exec",
             executable = True,
         ),
+        "_error_format": attr.label(
+            default = Label("//:error_format"),
+        ),
         "_process_wrapper": attr.label(
             doc = "A process wrapper for running rustdoc on all platforms",
             default = Label("@rules_rust//util/process_wrapper"),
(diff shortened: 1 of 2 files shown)


UebelAndre (changes requested):
Hi, thanks for the PR! Can you provide a bit more context here as well as a unit test of sorts?

goffrie:
The context is that we have automation using the output of rustdoc that uses json error format. We used to use extra rustc flags to pass `--error-format=json` but that broke since #1275 now unconditionally passes `--error-format=human` to rustdoc (and it's an error to specify error-format twice on the rustc/rustdoc command line). I figured that the cleaner solution would be to respect the global error_format configuration instead of trying to revert the old change.

I'll see when I can find some time to work on a unit test, but this isn't urgent to merge since we run a patched version.

UebelAndre (changes requested):
In https://github.com/bazelbuild/rules_rust/blob/0.27.0/test/unit/rustdoc/rustdoc_unit_test.bzl you should be able to add a test that hard codes some value for `//:error_format` that you should be able to assert made it's way into the resulting action. 

Does that seem reasonable?

goffrie:
Thanks for the pointer! I've added a test - is that what you had in mind?

UebelAndre (approved):
This is perfect, thank you!

PR #2174, opened 2023-09-28
Add examples for cpp calling rust by including header

daivinhtran:


UebelAndre (changes requested):
Thanks! Could you add a README.md to `examples/ffi/cpp_calling_rust_with_header` to explain what's going on with words? I opened https://github.com/bazelbuild/rules_rust/issues/1839 a while ago and would like to at some point refactor the examples to be demonstrating how to do common things vs treating them as integration tests like we do now.

PR #2180, opened 2023-10-03
Make more annotations select-able

illicitonion:
This is a follow-up to #2139 making more annotations support `select`-ing.


UebelAndre (approved):
Sweet! Thanks!

PR #2183, opened 2023-10-06
Simplify CPU arch detection in get_host_triple

dzbarsky:
The repository_ctx has already populated the arch and normalized it. Tested this locally on Mac/Linux amd/arm, but don't have an easy windows setup. Let's see what CI says.


UebelAndre (approved):
Love the cleanup, thanks!

PR #2188, opened 2023-10-09
Move output_hash creation upstream

daivinhtran:


rust/private/utils.bzl
```diff
@@ -849,68 +849,3 @@ def _symlink_for_non_generated_source(ctx, src_file, package_root):
         return src_symlink
     else:
         return src_file
-
-def create_crate_info_dict(ctx, toolchain, crate_type):
```
daivinhtran, line 853: The logic of this function differs significantly across the rules and aspects so it's not sharable anymore.

scentini (approved):
Thanks!

PR #2189, opened 2023-10-09
Remove _rustc_env_attr from CrateInfo

daivinhtran:
#2188's follow-up PR to remove unused `_rustc_env_attr` field in `CrateInfo` provider. We can rely on `skip_expanding_rustc_env` to determine whether `rustc_env` is already expanded or not.


scentini (approved):
Thanks!

PR #2193, opened 2023-10-12
[prost] Fix handling of message names with acronyms

william-smith-skydio:
For example, the message name `SomeAPI`. Internally protoc-gen-prost [converts](https://github.com/neoeinstein/protoc-gen-prost/blob/1a6d3593622af18b75a4a79f545f8530cdaf444f/protoc-gen-tonic/src/util.rs#L14) this to upper camel case, but rules_rust is not doing that when populating the `--extern_path` args. The result is that when a message with an acroynym in its name is included in another proto file, the generated code has a type name with the wrong casing.

Without this fix, the added test fails to build like this:

```
error[E0412]: cannot find type `NameWithCAPS` in module `camel_case_proto::camel_case`
 --> bazel-out/k8-fastbuild/bin/proto/prost/private/tests/camel_case/another_proto.lib.rs:9:75
  |
9 |         pub inner: ::core::option::Option<::camel_case_proto::camel_case::NameWithCAPS>,
  |                                                                           ^^^^^^^^^^^^ help: a struct with a similar name exists: `NameWithCaps`
 --> bazel-out/k8-fastbuild/bin/proto/prost/private/tests/camel_case/camel_case_proto.lib.rs:7:5
  |
  = note: similarly named struct `NameWithCaps` defined here

```

### DIFF
--- a/proto/prost/private/protoc_wrapper.rs
+++ b/proto/prost/private/protoc_wrapper.rs
@@ -9,7 +9,7 @@ use std::path::PathBuf;
 use std::process;
 use std::{env, fmt};
 
-use heck::ToSnakeCase;
+use heck::{ToSnakeCase, ToUpperCamelCase};
 use prost::Message;
 use prost_types::{
     DescriptorProto, EnumDescriptorProto, FileDescriptorProto, FileDescriptorSet,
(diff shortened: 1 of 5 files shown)


UebelAndre (approved):
Thanks!

PR #2194, opened 2023-10-12
[prost] Handle argfiles

william-smith-skydio:
If the command line passed to protoc is too long, `proto_common.compile` will [spill the args to an argfile](https://github.com/bazelbuild/bazel/blob/release-6.3.0/src/main/starlark/builtins_bzl/common/proto/proto_common.bzl#L105). This PR changes the arg parsing logic to handle the argfile.

Without this change, a prost pbtype build with enough dependencies will fail in a strange way. For me this manifested as the `--prost_out` flag being missing, because it was in the argfile.


UebelAndre (approved):
Great catch, thanks!

Issue #2210, opened 2023-10-24
A crate that appears both as a build and a normal dependency is not handled correctly


ulan:
The [wasmtime-runtime](https://github.com/bytecodealliance/wasmtime/tree/main/crates/runtime) depends on [wasmtime-versioned-export-macros](https://github.com/bytecodealliance/wasmtime/tree/main/crates/versioned-export-macros) both as a build and a normal dependency.

Here is an excerpt from `cargo metadata`:
```
          {
            "name": "wasmtime_versioned_export_macros",
            "pkg": "wasmtime-versioned-export-macros 15.0.0 ...",
            "dep_kinds": [
              {
                "kind": null,
                "target": null
              },
              {
                "kind": "build",
                "target": null
              }
            ]
          }
```

The problem is that `wasmtime-versioned-export-macros` appears in the final lock file only in `build_script_attrs` and doesn't appear in `common_attrs`, which leads to a build error:
```
error[E0432]: unresolved import wasmtime_versioned_export_macros
```

I think the bug is in this line:
https://github.com/bazelbuild/rules_rust/blob/cd3e53e9b8555543071ea109218baf41fc4b88f7/crate_universe/src/metadata/dependency.rs#L62

That line should instead be
```
.filter(|dep| is_dev_dependency(dep) || is_normal_dependency(dep))
```

That is because `!is_build_dependency(dep)` is not equivalent to `is_dev_dependency(dep) || is_normal_dependency(dep)` since a dependency can be both a build and a normal dependency.

UebelAndre:
I’d be happy to review a PR 😄

ulan:
@UebelAndre: Do you know approximately when the next version that includes this fix will be released?

UebelAndre:
https://github.com/bazelbuild/rules_rust/releases/tag/0.30.0 should have your change in it. Closing this out since I think that addresses everything, let me know if I missed something!

ulan:
Thanks a lot @UebelAndre!

PR #2211, opened 2023-10-24
Fix handling of proc macro deps with multiple kinds

ulan:
A proc macro dependency may have multiple kinds such as build
and normal. Due to a bug in filtering predicate, such dependencies
are considered only as build dependency.

This PR fixes the bug and adds a regression test.

Issue: https://github.com/bazelbuild/rules_rust/issues/2210


crate_universe/src/metadata/dependency.rs
```diff
@@ -604,6 +604,23 @@ mod test {
         );
     }
 
+    #[test]
+    fn multi_kind_proc_macro_dep() {
```
UebelAndre, line 608: Can you also assert that the `build` dependency is correctly identified for build script targets?

UebelAndre (changes requested):
Thanks! One request 😄

crate_universe/src/metadata/dependency.rs
ulan: Done, thanks!

UebelAndre (approved):
Thank you!

ulan:
@UebelAndre thanks for merging! Do you know approximately when the next version that includes this fix will be released?

Issue #2092, opened 2023-08-02
Tinyjson compilation is not reproducible


cameron-martin:
When compiling tinyjson, the hashes of the outputs differ between clean builds:

```
***************
*** 188019,188040 ****
    }
  }
  listed_outputs: "bazel-out/k8-opt-exec-CF7D65A3/bin/external/rules_rust_tinyjson/libtinyjson-4031717389.rlib"
  remotable: true
  cacheable: true
  progress_message: "Compiling Rust (without process_wrapper) rlib tinyjson (4 files) [for tool]"
  mnemonic: "Rustc"
  actual_outputs {
    path: "bazel-out/k8-opt-exec-CF7D65A3/bin/external/rules_rust_tinyjson/libtinyjson-4031717389.rlib"
    digest {
!     hash: "92a2c8e00550cc2e26a93d9e53beaa4a7b826d21c5112556eed9cb008435c88e"
!     size_bytes: 633710
      hash_function_name: "SHA-256"
    }
  }
  runner: "linux-sandbox"
  remote_cacheable: true
  target_label: "@rules_rust_tinyjson//:tinyjson"
  
  
  ---------------------------------------------------------
  
--- 188019,188040 ----
    }
  }
  listed_outputs: "bazel-out/k8-opt-exec-CF7D65A3/bin/external/rules_rust_tinyjson/libtinyjson-4031717389.rlib"
  remotable: true
  cacheable: true
  progress_message: "Compiling Rust (without process_wrapper) rlib tinyjson (4 files) [for tool]"
  mnemonic: "Rustc"
  actual_outputs {
    path: "bazel-out/k8-opt-exec-CF7D65A3/bin/external/rules_rust_tinyjson/libtinyjson-4031717389.rlib"
    digest {
!     hash: "3ea14e1e2f634961bfeb2d3ba2e9e690dd16c1991bd2a353583994b0098bd509"
!     size_bytes: 633702
      hash_function_name: "SHA-256"
    }
  }
  runner: "linux-sandbox"
  remote_cacheable: true
  target_label: "@rules_rust_tinyjson//:tinyjson"
```

I ran diffoscope on the output and this is the result: [tj_diff.txt](https://github.com/bazelbuild/rules_rust/files/12239586/tj_diff.txt). It looks like the paths of files in the sandbox are being written to the library. This is similar to #1530, but this is using Linux (Ubuntu 22.04).

UebelAndre:
[...]

PR #2216, opened 2023-10-26
Fix issues causing the `Rustc` process wrapper to be built non-determ…

UebelAndre:
This change aims to ensure the process wrapper is built determinisitically, thus reducing churn in cases where multiple developers share a bazel remote cache. The shared cache is notable as the binaries can be deterministically produced for a single user for a single checkout as some host information is embedded in the compiled binaries. The host information is stored in the `debug-info` section of the executable. This can be resolved by stripping it. The other issue of determinism is `rust_binary_without_process_wrapper` does not resolve `${pwd}` like the process wrapper does, thus stripping sandbox paths from rustc outputs. To solve for this a small bash and batch script was added to account for this behavior. The intent of these scripts is to use __no__ host executables and instead rely on pure bash and batch to accomplish this and keep any actions which use this as hermetic as possible.

The following tables were produced on macOS. Note that `rules_rust_2` is simply another checkout of `rules_rust` on the same commit

Before the changes in this PR:

| path | rules_rust sha256 | rules_rust_2 sha256 |
| --- | --- | --- |
| bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/rules_rust_tinyjson/libtinyjson-4031717389.rlib | 60db194b3e98b67cc0702c6b6c48c5bc8fcf7d723f3ece6a7a24a53888158c7e | 3eac0b443ba160e3a1bde3b023f4e953bb9fc9722e430b5ded6dbb723bc2b532 |
| bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/util/process_wrapper/process_wrapper | 32e7840602c977b76bba0bc39da768d35db3e7a17dbf96e455727be2b2f0151f | 0d248279bbc9b17be5914b41a66759f939ef4da77f8a40a5ce6fa3bf339648ad |

After the changes:

| path | rules_rust sha256 | rules_rust_2 sha256 |
| --- | --- | --- |
| bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/rules_rust_tinyjson/libtinyjson-4031717389.rlib | 07432f5d207da854266ccde212243c67c29907f81a7619434ce9f608b1658d96 | 07432f5d207da854266ccde212243c67c29907f81a7619434ce9f608b1658d96 |
| bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/util/process_wrapper/process_wrapper | cbf17392338aabfc942d975f95a49a67b731c5e597a7d27e3d9cf4d4a06b8f2c | cbf17392338aabfc942d975f95a49a67b731c5e597a7d27e3d9cf4d4a06b8f2c |

closes https://github.com/bazelbuild/rules_rust/issues/2092

### DIFF
--- a/util/process_wrapper/process_wrapper.bat
+++ b/util/process_wrapper/process_wrapper.bat
@@ -1,0 +1,31 @@
+@ECHO OFF
+SETLOCAL enabledelayedexpansion
+
+SET command=%*
+
+:: Resolve the `${pwd}` placeholders
+SET command=!command:${pwd}=%CD%!
+
+:: Strip out the leading `--` argument.
+SET command=!command:~3!
+
+:: Find the rustc.exe argument and sanitize it's path
+for %%A in (%*) do (
+    SET arg=%%~A
+    if "!arg:~-9!"=="rustc.exe" (
+        SET sanitized=!arg:/=\!
+
+        SET command=!sanitized! !command:%%~A=!
+        goto :break
+    )
+)
+
+:break
+
+%command%
+
+:: Capture the exit code of rustc.exe
+SET exit_code=!errorlevel!
+
+:: Exit with the same exit code
+EXIT /b %exit_code%
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -1060,6 +1060,13 @@ def _common_attrs_for_binary_without_process_wrapper(attrs):
         cfg = "exec",
     )
 
+    new_attr["_bootstrap_process_wrapper"] = attr.label(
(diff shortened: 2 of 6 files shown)


util/process_wrapper/process_wrapper.bat
UebelAndre, line 1: It's unclear if this actually results in deterministic outputs. I do not have a windows machine to verify.

scentini:
I don't understand the windows code, but LGTM otherwise, and I think we can leave windows determinism to be confirmed by a windows user. Thanks!

PR #2217, opened 2023-10-26
Patch github.com/bazelbuild/rules_rust/pull/1899

daivinhtran:


.bazelci/presubmit.yml
```diff
@@ -24,8 +24,10 @@ default_macos_targets: &default_macos_targets
 default_windows_targets: &default_windows_targets
   - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
   - "//..."
-  - "-//test/proto/..."
   - "-//test/unit/pipelined_compilation/..."
+  # The proto rules do not work on windows
+  - "-//proto/..."
```
UebelAndre, line 29: Could we instead tag specific targets as incompatible with windows? How many targets fail on windows?

UebelAndre (changes requested):
Thanks! Just one question

.bazelci/presubmit.yml
daivinhtran: Pretty much all `proto` targets fail on windows because `rust_prost_toolchain` fails building on windows after https://github.com/bazelbuild/rules_rust/pull/1899 introduces a transition to make the filepath even longer with the config hash.

scentini: @UebelAndre could we move this forward and disable windows the path length issues are taken care of?

UebelAndre: I’m not convinced we need to disable the unit tests but haven’t had time to figure out if transitioning for the panic PR is necessary.

Issue #2039, opened 2023-06-29
Replace SYSROOT env var with --sysroot flag


daivinhtran:
`rust_toolchain` [generates a rust sysroot](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L495) from collection of toolchain components. The sysroot path is then assigned to [`RUST_SYSROOT` make variable](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L545) and [ToolchainInfo.sysroot](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L629). When constructing rustc compile action, we set `ToolchainInfo.sysroot` to [`SYSROOT` env var](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/private/rustc.bzl#L1022). 

Apparently, `SYSROOT` env var isn't read by rustc and I suspect it is currently a no-op. To reproduce the no-op of `SYSROOT`, simply change [it](https://github.com/bazelbuild/rules_rust/blob/9442aed8ca9e07e024f9daaf06030c968623c9c1/rust/toolchain.bzl#L545) to any random string and expect the builds to still succeed.

Per [rustc doc](https://doc.rust-lang.org/rustc/command-line-arguments.html#--sysroot-override-the-system-root]), we can set `--sysroot` to override the system sysroot (i.e. `rustc --print sysroot). I believe this is needed to prevent rustc from managing to escape the Bazel sandbox and seeing unspecified inputs.

As a concrete example, say we have a toolchain defined as

```
rust_toolchain(
  name = "foo",
  rustc = "prebuilt/bin/rustc",
  rustc_lib = "sysroot_with_stdlibs", # outside of $WORKSPACE/prebuilt
  rust_std = "stdlibs_built_from_source", # outside of $WORKSPACE/prebuilt
)
```

The system sysroot is
```
$ ~/prebuilt/bin/rustc --print sysroot
~/prebuilt
```

Under `~/prebuilt`, there are prebuilt stdlibs
```
<target-tripple>/
  lib/
    libstd-<hash>.rlib|so
```

[...]

PR #2223, opened 2023-10-27
Set --sysroot to sysroot generated by rust_toolchain

daivinhtran:
Addresses https://github.com/bazelbuild/rules_rust/issues/2039


rust/private/rustc.bzl
```diff
@@ -1267,14 +1264,22 @@ def rustc_compile_action(
             dsym_folder = ctx.actions.declare_directory(crate_info.output.basename + ".dSYM", sibling = crate_info.output)
             action_outputs.append(dsym_folder)
 
+    rustc_sysroot_arg = ctx.actions.args()
```
UebelAndre, line 1267: Why is this not just in `rustc_args`?

UebelAndre (changes requested):

UebelAndre (changes requested):
Could you also add an experimental flag for this?

rust/private/rustc.bzl
daivinhtran: Are you talking about `args.rustc_flags`?

UebelAndre: Maybe? There should be 3 `ctx.actions.args` objects that’s created for Rustc actions. This seems to be a new one. That’s what I’m referring to

daivinhtran: Ideally, we should set `--sysroot` inside `construct_arguments` (where rustc args is initially created). However, `--sysroot` has been customly added by other rules after `constructor_arguments` is called.

1. https://github.com/daivinhtran/rules_rust/blob/a7eba053b9b8964f83d5a9ee76ed919a8e385362/rust/private/rustdoc.bzl#L143
2. https://github.com/daivinhtran/rules_rust/blob/a7eba053b9b8964f83d5a9ee76ed919a8e385362/bindgen/private/bindgen.bzl#L147
3. https://github.com/daivinhtran/rules_rust/blob/a7eba053b9b8964f83d5a9ee76ed919a8e385362/cargo/private/cargo_build_script.bzl#L170

Technically, I could just handle it similar to these three cases but it broke [this test](https://github.com/daivinhtran/rules_rust/blob/a7eba053b9b8964f83d5a9ee76ed919a8e385362/test/toolchain/toolchain_test.bzl#L26-L29) that checks rustc flags end with toolchain flag and config flag. I could also just fix the test allow `--sysroot` being the last element.

UebelAndre: The test seems like it could be fixed by adding the the new flag here https://github.com/bazelbuild/rules_rust/blob/0.30.0/rust/private/rustc.bzl#L967

As far as the other uses of `--sysroot`, could `construct_arguments` be expanded to indicate whether or not to include `--sysroot` in the returned `rustc_flags`? I think this would allow for enough control and wouldn't need to change the return value.

daivinhtran: Done. PTAL.

daivinhtran:
> Could you also add an experimental flag for this?

@UebelAndre Done.

rust/private/rustc.bzl
```diff
@@ -819,6 +820,7 @@ def construct_arguments(
         build_metadata (bool): Generate CLI arguments for building *only* .rmeta files. This requires use_json_output.
         force_depend_on_objects (bool): Force using `.rlib` object files instead of metadata (`.rmeta`) files even if they are available.
         skip_expanding_rustc_env (bool): Whether to skip expanding CrateInfo.rustc_env_attr
+        sandboxed (bool): Whether the action using the constructed arguments is sandboxed
```
UebelAndre, line 823: This seems like a weird name. I would have expected to see `add_sysroot` or something. Can you explain what the relevance of sandboxing is here?

rust/private/rustc.bzl
```diff
@@ -1038,9 +1047,6 @@ def construct_arguments(
             data_paths,
         ))
 
-    # Ensure the sysroot is set for the target platform
-    env["SYSROOT"] = toolchain.sysroot
```
UebelAndre, line 1042: This should probably be gated behind the new feature flag.

.bazelci/presubmit.yml
```diff
@@ -92,8 +92,10 @@ tasks:
     test_targets: *default_linux_targets
     build_flags:
       - "--compilation_mode=opt"
+      - "--@rules_rust//rust/settings:experimental_toolchain_generated_sysroot=True"
```
UebelAndre, line 95: Instead of updating the CI jobs, can you update a unit tests to transition this setting? Or add it to the `.bazelrc`, but in general I would like `presubmit.yaml` to be used as an absolute last resort.

UebelAndre (changes requested):

rust/private/rustc.bzl
daivinhtran: Per https://github.com/bazelbuild/rules_rust/issues/2039, `SYSROOT` is no-op. Are you expecting it might be used in user's source code?

UebelAndre: I also believe that to be true but I don't think it hurts to gate this. Maybe some power users have done something with this flag. This is a change to the core rules so I think we should have a controlled rollout.

daivinhtran: Should we grate `SYSROOT` and `--sysroot` behind two different feature flags? Even though it sounds redundant, I think it's a good practice since they're pretty orthogonal in this context - one is used by rustc and one is used by user's source code.

rust/private/rustc.bzl
daivinhtran: Sorry, I was trying to make sure `rustdoc`'s implementation doesn't change. We shouldn't need this if we use `toolchain. sysroot`. I removed it.

.bazelci/presubmit.yml
daivinhtran: How will eventually roll out (e.g. remove the bool flag)?

daivinhtran: To me, the unit tests aren't really useful in this context because we're trying to verify whether `rustc` executes successfully. The unit tests (aka Bazel's analysis test) only verifies that the action has the correct argv (`--sysroot`) but doesn't really verify that `rustc` builds the source successfully.

UebelAndre: https://github.com/bazelbuild/rules_rust/blob/main/COMPATIBILITY.md#how-to-make-a-backwards-incompatible-change is the process we should probably follow here.



> IMO, the unit tests aren't really useful in this context because we're trying to verify whether `rustc` executes successfully given the new `--sysroot` flag pointing to a generated directory. The unit tests (aka Bazel's analysis tests) only verify the action has the correct argv (`--sysroot`) but don't verify that `rustc` builds the source successfully.

The analysistest has [config_settings](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/unittest_doc.md#analysistest.make-config_settings) which would allow you to transition `target_under_test` to enable this specific feature. I think largely we'd just wanna make sure `--sysroot` appears in the Rustc invocation (which would probably still be a test we keep after the flag is deleted). We could cut a release while the flag is disabled and then quickly enable it so users who like to stick closer to HEAD can identify issues. What do you think about that?

rust/private/rustc.bzl
UebelAndre: I personally think we'd be fine without a second flag. I see this change as correcting the definition of `--sysroot` by moving it from the unused `SYSROOT` environment variable to the actual `--sysroot` rustc flag. It would be nice to delete the env var unless it's consumed by rustc in some meaningful way (which I don't believe it is). Thus, I feel the new feature flag should gate both to more clearly communicate the transition. Does that sound reasonable?

UebelAndre (changes requested):

rust/private/rustc.bzl
daivinhtran: > I see this change as correcting the definition of --sysroot by moving it from the unused SYSROOT environment variable to the actual --sysroot rustc flag.

Because we have came to the conclusion that `SYSROOT` is not used by `rustc` but might be used by end users (as stated in this thread), `SYSROOT` and `--sysroot` are orthogonal.

Here is why:

Before adding `--sysroot` to `contruct_arguments`, `SYSROOT` is not used anywhere in rust internals but only (possibly) end users. Adding `--sysroot` does not change that fact for `SYSROOT`. Hence, this PR is not meant to move to `--sysroot` from `SYSROOT`. 

Adding `--sysroot` flag to rustc ensures that rustc finds stdlibs and binaries from a directory generated by `rust_toolchain`.

So ideally, we should just keep `SYSROOT` as is in this PR. We can clean up `SYSROOT` in the future or in a separate PR.

UebelAndre: that also sounds fine to me 😄

rust/private/rustc.bzl
```diff
@@ -965,6 +965,10 @@ def construct_arguments(
     if linker_script:
         rustc_flags.add(linker_script, format = "--codegen=link-arg=-T%s")
 
+    if hasattr(ctx.attr, "_experimental_toolchain_generated_sysroot"):
```
UebelAndre, line 968: Can this be safely added unconditionally? Or are there rules that will break when this check is removed?

UebelAndre (changes requested):

.bazelci/presubmit.yml
daivinhtran: The analysis tests will ensure `--sysroot` exists in the action command (i.e. **analysis phase**). However, it does not ensure whether `rustc` uses the stdlibs and binaries to build end users's rust code successfully (i.e. **execution phase**).

Having it built successfully on CI makes sure that rust code in `rules_rust` all build successfully with `--sysroot` being set. So on my mind, the rollout phases will be the following:
1. Add a build setting per the instruction in https://github.com/bazelbuild/rules_rust/blob/main/COMPATIBILITY.md#how-to-make-a-backwards-incompatible-change
2. Update analysis tests to enable the build setting
3. Update CI to build all rules_rust's code with the build setting enabled. (Rollout to `rules_rust`'s code)
4. Remove the build setting. (Rollout to end users's code)

(3) ensures that isn't any regression in execution phase before rolling it out to end users.

Let me know what you think.

rust/private/rustc.bzl
daivinhtran: Good point. Some rules don't use `common_attrs` but we should add `_experimental_toolchain_generated_sysroot` to all rules that use `rustc`.

UebelAndre: Which rules break?

daivinhtran: `rustdoc`.

UebelAndre: `rustdoc` directly calls `construct_arguments`, maybe restoring the parameter to this function would do the trick?

daivinhtran: I think we should just add the attribute to `rustdoc` rule. https://github.com/bazelbuild/rules_rust/pull/2223/commits/b1a7e1a318faddaff0a486bf36e68c9af313c4b2#diff-ac2237497d8db708da38cbb6a5fda472b241e592c0c148c215eb7687c11b4d1f

UebelAndre: Oh, I thought there were some rules that didn't support `--sysroot`. If that's not the case then this approach seems good to me!

daivinhtran: > Oh, I thought there were some rules that didn't support `--sysroot`. If that's not the case then this approach seems good to me!

Ok. Sounds good to me.

.bazelci/presubmit.yml
daivinhtran: PTAL https://github.com/bazelbuild/rules_rust/pull/2223/commits/d985ad88cb7c183e3e5789c7160fc94eaafd3f74

rust/private/rust.bzl
```diff
@@ -609,6 +609,13 @@ _common_attrs = {
     "_error_format": attr.label(
         default = Label("//:error_format"),
     ),
+    "_experimental_toolchain_generated_sysroot": attr.label(
```
UebelAndre, line 612: Normally we put these flags on `rust_toolchain`. Any reason not to follow that pattern here? It would allow you to avoid adding it to two rules.

UebelAndre (changes requested):
Looking good! Just one last question 😄

rust/private/rust.bzl
daivinhtran: Ahh good point. PTAL https://github.com/bazelbuild/rules_rust/commit/f8027f2e0ed683b9eabf3364fd43bc17ab162e13

PR #2225, opened 2023-10-30
Split write_splicing_manifest and write_config_file.

matts1:
This commit creates the "generate" prefix for each of these functions, which generates the content of these files in a way that doesn't touch the ctx variable. This allows us to use these functions in module extensions, which is required for https://github.com/bazelbuild/rules_rust/pull/1910.


UebelAndre (approved):
Seems good to me, thanks!

Issue #2168, opened 2023-09-22
Add support for binary-only crates with crate_universe


adsnaider:
The documentation specifies that crate_universe doesn't support binary-only dependencies out of the box: http://bazelbuild.github.io/rules_rust/crate_universe.html#binary-dependencies. It would be really nice if it did. How much work would it be to implement? I would be happy to try contributing this feature.

illicitonion:
If we're going to support this, it would likely be via https://github.com/bazelbuild/rules_rust/issues/1739 (see also https://github.com/bazelbuild/rules_rust/issues/2031).

I expect it's probably a relatively small amount of work to do, but it's using an unstable cargo feature so you may run into bugs on the way (and will also need to fork the `cargo_metadata` crate to modify some structs in there) :)

mikea:
this limitation is very painful with `cxxbridge-cmd`. I see quite a few projects on github (google's skia including) trying to implement the workaround.

UebelAndre:
> this limitation is very painful with `cxxbridge-cmd`. I see quite a few projects on github (google's skia including) trying to implement the workaround.

There are some known work-arounds. The `wasm_bindgen_library` rule uses one of them. If anyone is willing to try out nightly releases of Rust (Cargo) that support [-zbindeps](https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html) I'd be happy to review pull-requests to make any updates needed for `crate_universe` to parse the new info from `cargo metadata` (since that's the source of truth `crate_universe` uses to generate dependencies).

mgeisler:
The documentation has [moved a little](https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html#binary-dependencies).

I'm trying to make use of binaries as well, without using nightly features. I've found that there is a difference between a pure-binary crate and one which is also a library crate:

- Library crate: you get library targets as expected,
- Library + binary crate: you get _both_ when `generate_binaries = True` is set,
- Binary crate: you get neither.

So with

```python

crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
crate.spec(
    package = "anyhow",               # library
    version = "1.0.77",
)
crate.spec(
    package = "mdbook-svgbob",        # binary
    version = "0.2.2",
)
crate.spec(
    package = "mdbook-i18n-helpers",  # library + binary
    version = "0.4.0",
)
crate.spec(
    package = "mdbook-pandoc",        # library + binary
[...]

PR #2226, opened 2023-10-30
Adds support for bindeps in crate universe

### LINKED ISSUE #1739: Support "artifact dependencies" (bin, cdylib, staticlib)
Historically Cargo has not supported depending on binary packages, only on libraries.

In the past, crate_universe used to expose `extra_workspace_members` for declaring binary dependencies to augment the library dependencies established by your Cargo.toml.

```bzl
crates_repository(
    name = "crate_index_extra_members",
    extra_workspace_members = {
        "texture-synthesis-cli": crate.workspace_member(
            sha256 = "a7dbdf13f5e6f214750fce1073279b71ce3076157a8d337c9b0f0e14334e2aec",
            version = "0.8.2",
        ),
    },
    ...
)
```

This was deleted in #1406 in favor of "external manifests":

```bzl
http_archive(
    name = "names",
    urls = ["https://github.com/fnichol/names/archive/refs/tags/v0.13.0.zip"],
    ...
)

crates_repository(
    name = "crate_index_cargo_remote",
    manifests = ["@names//:Cargo.toml"],
    ...
)
```

Recently though, Cargo has begun implementing support for binary dependencies natively in Cargo.toml.

- RFC: https://github.com/rust-lang/rfcs/pull/3028
- Tracking issue: https://github.com/rust-lang/cargo/issues/9096
- Overview: https://doc.rust-lang.org/1.66.0/cargo/reference/unstable.html#artifact-dependencies
- Guide-level explanation: https://github.com/rust-lang/rfcs/blob/master/text/3028-cargo-binary-dependencies.md#guide-level-explanation

Supporting this feature in crate_universe would make it substantially more convenient to manage binary dependencies. In particular, `--repin` would be able to pull in updates, which is not possible with the current "external manifests" way.

Binary dependencies would be declared in Cargo.toml as follows, and would work with both `crates_repository` and `crates_vendor`.

```toml
[dependencies]
# Export rust_binary for all bins in the package. If the
# package also contains a library, then a rust_library still
# needs to be generated in BUILD.example-1.0.0.bazel because
# the binary depends on it, but it will be private.
example = { version = "1", artifact = "bin" }

# Export rust_binary for those specific bins, do not export rust_library.
example = { version = "1", artifact = ["bin:faketty", "bin:cmake"] }

# Export rust_library as well as rust_binary.
example = { version = "1", artifact = "bin", lib = true }

# Export rust_library, do not generate any rust_binary targets.
example = { version = "1" }
```

### LINKED ISSUE #2031: Feature request: Support -Zbindeps in crate universe when using a nightly compiler
This would allow easy dependence on binary dependencies (e.g. `bindgen` or  `wasm-bindgen`) by directly adding a dependency in Cargo.toml, rather than following the convoluted process here: https://bazelbuild.github.io/rules_rust/crate_universe.html#binary-dependencies.

For more info on bindeps, see: https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html.

Initially, my idea was just to add "-Zbindeps" to other_options here, when using a nightly version of cargo (we can also add a separate CLI option to splicing to make this more explicitly opt-in):
https://github.com/bazelbuild/rules_rust/blob/ca750fa83d75a2408be93519e9c4d1a2b8b2a087/crate_universe/src/metadata.rs#L77-L83

However, this does not work, due to use of cargo_metadata. The actual Cargo.toml file managed by the end user is parsed by `cargo_metadata` (which serializes packages into this struct: https://github.com/oli-obk/cargo_metadata/blob/010cd78751e51dc48d0ca08ca0f4be3ac054a689/src/lib.rs#L289), and then written out as a generated file (which includes various additional attributes used by crate universe). The generated Cargo.toml file is the actual file passed to `cargo` when splicing. The serialization / deserialization through `cargo_metadata` is stripping out any unknown fields defined within `cargo_metadata`. Critically, this is dropping the `artifact` field defined on a bin-only package when using bindeps.

1) Why do we need to add additional fields to a generated Cargo.toml before splicing?

2) It seems to me like we could generate the generated Cargo.toml with a generic TOML library, rather than round tripping through `cargo_metadata`. Would maintainers be open to such an approach? Are there any concerns about such an approach?

csmulhern:
Fixes #1739, #2031, #2168.


UebelAndre:
I think this looks good! Just gotta fix up the new test and rebase 😄

UebelAndre (approved):

UebelAndre:
Ah, there's one more directory which needs to be repinned via crate_universe. Try running the following?

```bash
cd ./test/no_std/ && REPIN=1 bazel query //...
```

illicitonion (approved):
Looks great, thanks!

csmulhern:
> Ah, there's one more directory which needs to be repinned via crate_universe. Try running the following?
> 
> ```shell
> cd ./test/no_std/ && REPIN=1 bazel query //...
> ```

Bah, thanks.

Happy we were able to land this. Thanks all!

PR #2228, opened 2023-10-31
Updated crate_universe dependencies

UebelAndre:


UebelAndre:
@illicitonion a dependency on `gix` requires [Rust 1.65.0](https://github.com/rust-lang/rust/releases/tag/1.65.0), given that that was released nearly a year ago (Nov 3, 2022), would it be acceptable to bump to that at this time?
```
error[E0658]: `let...else` statements are unstable
  --> external/cui__powerfmt-0.2.0/src/ext.rs:24:9
   |
24 | /         let Some(fill_width @ 1..) = final_width.checked_sub(args_width) else {
25 | |             // No padding will be present. Write the arguments as-is.
26 | |             return self.write_fmt(args);
27 | |         };
   | |__________^
   |
   = note: see issue #87335 <https://github.com/rust-lang/rust/issues/87335> for more information
```

illicitonion:
Seems pretty reasonable - let's ship it - worst case we can revert if it causes problems for people.

UebelAndre:
Dang, now there's an issue that requires [1.67.0](https://github.com/rust-lang/rust/releases/tag/1.67.0) ()
```
error[E0658]: `let...else` statements are unstable
  --> external/cui__powerfmt-0.2.0/src/ext.rs:24:9
   |
24 | /         let Some(fill_width @ 1..) = final_width.checked_sub(args_width) else {
25 | |             // No padding will be present. Write the arguments as-is.
26 | |             return self.write_fmt(args);
27 | |         };
   | |__________^
   |
   = note: see issue #87335 <https://github.com/rust-lang/rust/issues/87335> for more information
```

UebelAndre:
@illicitonion so, it turns out I would need Rust 1.67.0, but there's some overly aggressive flag in clippy that requires a bunch of code to be changed. If I go to 1.68.0 then I can avoid changing anything. I think if you're ok with going as far as 1.67.0, that it'd be better to go to 1.68.0 since it allows me to avoid making code changes, thus older versions should still be compatible. What do you think?

PR #2229, opened 2023-10-31
Support artifact dependencies in crate.spec

dtolnay:
Followup to #2226. That PR made bindeps work for crates repositories that use `manifests`.

https://github.com/bazelbuild/rules_rust/blob/d86faee8b85cc334e0529193142f4b61362e79a3/examples/crate_universe/WORKSPACE.bazel#L60

This PR makes it work for crates repositories that use `packages`.

https://github.com/bazelbuild/rules_rust/blob/43b101e6b3e679b333c2622d7615233edf344624/examples/crate_universe/WORKSPACE.bazel#L317-L320

PR #2230, opened 2023-10-31
Convert cxxbridge-cmd dependency to bindep

dtolnay:
This is how I tested #2229.

I am not sure whether we want this in rules_rust yet because the existing `@cxxbridge-cmd` serves as a useful example for users to copy into their own workspaces, and not all of those users use nightly. If we are open to using `-Zbindeps` unconditionally on all toolchains (via `RUSTC_BOOTSTRAP=1`) then I'll finish this up, otherwise defer to after artifact dependencies is stable.


dtolnay:
Okay! That sounds good to me.

PR #2232, opened 2023-10-31
[prost] includes prost crate infos in rust_analyzer_aspect

### LINKED ISSUE #2121: Rust Analyzer Generation should also add Prost generated code to the seach paths
As mentioned in the title, the rust analyzer generation aspect should also dig into prost rules and generate the specs for them.

This is slightly complicated because unlike the other protobuf rules, prost uses aspects to traverse the proto's dependency tree to compile the protos. This is nice from a user perspective because it allows the user to only define one `rust_prost_library` for a proto_library that includes all the proto_library depndencies instead of having to define a `rust_prost_library` for every dependency as well and populate the dependency tree. However, this is problematic for the rust analyzer aspect because it doesn't allow the aspect to accumulate dependencies, you only get the final `DepVariantInfo`.

I propose that we attack this in steps:
 - Just get rust analyzer working for the top level proto (not the transitive dependnecies)
 - Then we can work on recreating the dependency tree and getting rust analyzer working for the transitive dependencies too.

I wrote a quick PoC for this here: https://github.com/DolceTriade/rules_rust/commit/244eaea3fd8dcf9bf51e86ebf15b8f15a2e8f214

iamricard:
The change is fairly straightforward, but I did need to move prost
providers up to `prost/providers.bzl` and out of `private/prost.bzl` to
avoid a cyclic dependency.

What we do is:
* change `rust_prost_library` to return the `rust_proto_info` provider
* in the `rust-analyzer` aspect we iterate over transitive dep infos and
  generate the appropriate `RustAnalyzerInfo`
* lastly, we use the `crate_info` from `ProstProtoInfo` if we're
  processing a prost target as the "root" `crate_info`

There are a couple of things I'm not sure are correct:
1. For every prost dep I set `deps` to `[]` - I think that's correct but
   I'm not sure-sure.
2. Similarly I set `proc_macro_dylib_path` to always `None` - again I
   think that's correct, but not sure-sure.

Fixes #2121.


UebelAndre (approved):
Awesome! Looks good to me!

PR #2234, opened 2023-11-02
process_wrapper: write all output from rustc if json parsing fails.

gigaroby:
Previously we were only reporting the first line of any error rustc may emit when it panic'd or otherwise unexpectedly stopped outputting json.


scentini (approved):
Thanks!

PR #2245, opened 2023-11-08
Run "bazel run //*:crates_vendor"

matts1:
This avoids very large diffs from some an unrelated PR I have coming.


UebelAndre (approved):
Thanks!

PR #2246, opened 2023-11-09
Declare dependencies in vendored crates

matts1:
Currently, when you run `bazel build --enable_bzlmod //crate_universe:bin`, you get the error `undefined repo @cui`.

When you add a dependency on CUI, you get errors like `undefined repo @cui__<crate>-<version>` (the repo for the third party crate).

This change will allow us to automatically declare dependencies on our third party crates from bzlmod (see https://docs.google.com/document/d/1dj8SN5L6nwhNOufNqjBhYkk5f-BJI_FPYWKxlB3GAmA/edit#heading=h.5mcn15i0e1ch).


UebelAndre:
Do you wanna update the examples in a separate PR as well to reduce the noise?

examples/sys/complex/3rdparty/crates/BUILD.vcpkg-0.2.15.bazel
```diff
@@ -35,9 +33,44 @@ rust_library(
     rustc_flags = ["--cap-lints=allow"],
     tags = [
         "cargo-bazel",
+        "crate-name=vcpkg",
         "manual",
         "noclippy",
         "norustfmt",
     ],
+    target_compatible_with = select({
```
tbaing, line 41: This block gets repeated in a lot of places. Is putting it somewhere common and just referencing it from each of these a viable option to reduce the number of places it need to be (and the number of changes needed when the list changes)?

tbaing (approved):

examples/sys/complex/3rdparty/crates/BUILD.vcpkg-0.2.15.bazel
matts1: No - BUILD files are supposed to not follow the DRY principle.  See https://rsvp.withgoogle.com/events/bazelcon2023/sessions/day2-talk5

If you were to use a variable instead, then it becomes a lot harder for tooling to do analysis.

matts1:
@UebelAndre I can do so if you prefer, but I'm personally fine with it as is.

Also, do you mean just the `examples/sys/complex`, which I missed in #2245, or all the examples?

UebelAndre:
Just the missed examples

matts1:
Ok, I just rebased, diffs look much cleaner now

examples/sys/complex/3rdparty/crates/BUILD.vcpkg-0.2.15.bazel
tbaing: SG, thanks.

UebelAndre (changes requested):
Thanks! I had one more question 😄

crate_universe/src/rendering/templates/module_bzl.j2
```diff
@@ -339,7 +342,18 @@ def crate_repositories():
...
 {%- endfor %}
 {%- endfor %}
-{%- else %}
-    pass
-{%- endif %}
+    return [
+        {%- for id in context.direct_deps %}
+        {%- set crate = context.crates | get(key=id) %}
+        {%- if crate.repository %}
+       struct(repo="{{ crate_repository(name = crate.name, version = crate.version) }}", is_dev_dep = False),
```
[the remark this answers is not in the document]
matts1: I do agree with you, but unfortunately not. It needs to return something that we can make compatible with https://bazel.build/rules/lib/builtins/module_ctx#extension_metadata

Unfortunately, since that takes in two lists of strings as parameters, there's no actual struct built into bazel for this.

My options I considered when doing this were:
```
# These were hard to merge with others from the caller
return struct(regular_deps=[...], dev_deps=[...])
return ([...], [...])
struct(root_module_direct_dev_deps=[...], root_module_direct_dev_deps=[...])

# I chose this because it was easy to use by the caller
[struct(repo="...", is_dev_dep=bool), ...]

# This is probably the option I would choose if I had to pick an alternative. I didn't do it because it seemed like overkill to add another bzl file just for this one purpose.
# TitleCase for duck-typing because it acts like a class.
def RepoDependency(repo, is_dev_dep):
  return struct(repo=repo, is_dev_dep=is_dev_dep)
RepoDependency(repo="...", is_dev_dep=bool)
```

If you'd prefer that last option, just tell me where you'd like me to create the .bzl file and I can go ahead with it.

UebelAndre:
@matts1 I think this is good to go pending a rebase

UebelAndre (approved):

PR #2250, opened 2023-11-09
Fix docs for Bazel@HEAD

Wyverald:
Otherwise the 'docs' pipeline on Bazel CI complains that docs are out of date: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3447#018bb1e5-7c49-4729-88d9-893149ed99e2


UebelAndre (approved):
Thanks!!

PR #2265, opened 2023-11-15
Fix optional dependency check when crate is renamed

### LINKED ISSUE #2262: Missing crate features when `Cargo.toml` specifies `default-features = false`
Whenever a crate specified in `Cargo.toml` specifies a feature flag AND `default-features = false`, rules_rust generates a rust_library with no features instead of the required features. This can cause crates to no compile. An example of such a crate is `indexmap` that depends on `hashbrown` and specifies
```toml
[dependencies.hashbrown]
version = "0.12"
default-features = false
features = ["raw"]
```
Despite this, the rust_library generated in [BUILD.hashbrown-0.12.3.bazel](https://github.com/pamaury/bugs/blob/main/rules_rust_bug/crates/BUILD.hashbrown-0.12.3.bazel) has no feature enabled.

I have a minimal reproduction of the problem [here](https://github.com/pamaury/bugs/tree/main/rules_rust_bug#mishandling-of-no_default_features) using rules_rust 0.30.0.

### LINKED ISSUE #2263: Optional and unused dependencies that are renamed are pulled
Whenever a crate specified in `Cargo.toml` specifies an optional dependency that is not used but whose name does not match the original name of the crate, rules_rust still generates a dependency on that crate due to an alias. This can cause rules_rust to pull many unneeded dependencies. One such example is `serde_with` with `indexmap`:
```toml
indexmap_1 = {package = "indexmap", version = "1.8", optional = true, default-features = false, features = ["serde-1"]}
```
Even though serde_with does not require this dependency by default, it is still used in the generated BUILD file in aliases (not in the deps):
```python
rust_library(
    name = "serde_with",
    srcs = glob(["**/*.rs"]),
    aliases = {
        "@crate_index__indexmap-1.9.3//:indexmap": "indexmap_1",
    },
   ...
)
```

I have a minimal reproduction of the problem [here](https://github.com/pamaury/bugs/tree/main/rules_rust_bug#optional-and-unused-dependencies-that-are-renamed-are-pulled) using rules_rust 0.30.0.

nbdd0121:
Fix #2263 

I think this also fix #2262, because with this change the renamed optional dependency wouldn't be pulled in at all, rendering the issue moot.

cc @pamaury


UebelAndre (approved):
Thanks! Looks good 😄

PR #2272, opened 2023-11-18
replace pwd in rust-project.json output #1

ibigbug:
The generated content uses ${pwd} and in the rust-analyzer server output I see error logs saying can't find dir with ${pwd} references

Though I'm not sure if this actually affect anything as it seems work fine most of the time, but I think it maybe good to not have rust-analyzer errors


illicitonion (approved):
Works for me - thanks!

PR #2276, opened 2023-11-21
cargo_build_script can use rustc_env_files

### LINKED ISSUE #2248: `rustc_env_files` should work with `cargo_build_script`
We currently use `rustc_env_files` with `cargo_build_script` (with version 0.27.0). rules_rust version 0.30.0 no longer supports this. According to the documentation this should work since `cargo_build_script`  should pass `kwargs` to `rust_binary`.

illicitonion:
Fixes #2248


UebelAndre (approved):
Thanks!

PR #2277, opened 2023-11-21
Default experimental_toolchain_generated_sysroot to True

daivinhtran:
https://github.com/bazelbuild/rules_rust/pull/2223 only sets `--@rules_rust//rust/settings:experimental_toolchain_generated_sysroot` in a few targets on CI. This PR defaults the feature flag to `True`. 

This ensures the feature flag works at a larger scope because we remove it.


daivinhtran:
@UebelAndre PTAL

UebelAndre:
Thanks! Why not flip the flag though?

daivinhtran:
@UebelAndre Yes. Let's just flip the flag. I changed the PR. PTAL.

UebelAndre:
I don’t think you should remove the flag just yet. Can you only flip the default?

daivinhtran:
@UebelAndre Done. PTAL.

UebelAndre (approved):
Thanks!

gferon:
I'm not 100% sure, but this might have introduced an issue with the `clippy` aspect. I'm getting:

```
error: Option 'sysroot' given more than once
```

when trying to upgrade `rules_rust`. Any thoughts?

UebelAndre:
> I'm not 100% sure, but this might have introduced an issue with the `clippy` aspect. I'm getting:
> 
> ```
> Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
> error: Option 'sysroot' given more than once
> ```
> 
> when trying to upgrade `rules_rust`. Any thoughts?

Interesting, I'm curious why this wasn't caught by existing CI. Can you file a bug with a small repro?

neilisaac:
@UebelAndre same issue here.  I had to set `build --@rules_rust//rust/settings:experimental_toolchain_generated_sysroot=false` to work around this.  

clippy-driver will [automatically add --sysroot based on SYSROOT](https://github.com/rust-lang/rust-clippy/blob/master/src/driver.rs#L196) if it doesn't detect a --sysroot command line flag.  Is the @bazel-out/k8-opt/bin/.../test-NNN/my_test.clippy.ok-0.params parameter passed literally to clippy-driver, or is it expanded by process_wrapper?  If it's passed literally, that would explain why it appears twice.

Should an "experimental" flag really be enabled by default?

daivinhtran:
The fix to this issue should be done upstream in https://github.com/rust-lang/rust-clippy/pull/12203.

PR #2282, opened 2023-11-23
Add repin script

illicitonion:
Every time I need to update these I incrementally run these commands, finding out what I missed via CI each time.

Instead, let's write this down.


UebelAndre (approved):
Good idea!

PR #2284, opened 2023-11-25
Added support for multiple javascript rule sets to wasm_bindgen package

UebelAndre:
This PR converts the core `rust_wasm_bindgen` rule into a standalone rule that users can user to write custom interfaces into any rules/targets they desire while also supporting interfaces for existing Javascript rule sets. Users loading the root `@rules_rust//wasm_bindgen` symbols will be using the core Rust rules, where users loading symbols from the new sub directories `rules_nodejs` and `rules_js` will have variants that offer better integrations.


UebelAndre:
This change also implements https://github.com/bazelbuild/rules_rust/pull/2119

dzbarsky (commented):
this looks great, thanks for finding another path forward. We'll need to do a bit of work here to support bzlmod as well but that can be in other PRs

docs/rust_wasm_bindgen.md
```diff
@@ -5,24 +5,45 @@
...
+This should enable users to start using the [rust_wasm_bindgen](#rust_wasm_bindgen)
+rule. However, it's common to want to control the version of `wasm-bindgen` in the
+workspace instead of relying on the one provided by `rules_rust`. In this case, users
+should avoid calling `rust_wasm_bindgen_register_toolchains` and instead use the
+[rust_wasm_bindgen_toolchain](#rust_wasm_bindgen_toolchain) rule to define their own
+toolchains to register in the workspace.
+
+### Interfacing with Javascript rules
+
+While it's recommended for users to mantain their own , in the
```
illicitonion, line 37: Missing word after "own"?

illicitonion (approved):

PR #2290, opened 2023-11-28
rust-analyzer: Include prost_runtime as a dep for rust_prost_library

DolceTriade:
This allows rust-analyzer to autocomplete definitions like ::decode() and other methods that are found in ::prost::Message.


UebelAndre (approved):
Thanks!

Issue #2302, opened 2023-12-03
js_rust_wasm_bindgen (aspect_rules_js binding) results in error when including in js_library


jstuder-gh:
## Description

I've encountered an error when attempting to include the results of the `js_rust_wasm_bindgen` rule into a `js_library` target as data (with the intent of ultimately composing WASM binaries into NPM package).

Attempting to build this `js_library` artifact will result in an error. The `js_library` rule attempts to retrieve data from a `JsInfo` provider field that is not populated.

```
js_library(
    name = "hello_world_wasm_lib",
    srcs = [
        ":hello_world_wasm_test.js",
    ],
    data = [
        ":hello_world_bundler_wasm_bindgen",
        ":hello_world_deno_wasm_bindgen",
        ":hello_world_nodejs_wasm_bindgen",
        ":hello_world_nomodules_wasm_bindgen",
        ":hello_world_web_wasm_bindgen",
    ],
    deps = [],
)
```

```
ERROR: /path/to/repos/rules_rust/examples/wasm_bindgen/rules_js/BUILD.bazel:61:11: in js_library rule //wasm_bindgen/rules_js:hello_world_wasm_lib:
Traceback (most recent call last):
        File "/path/to/repos/aspect_rules_js/js/private/js_library.bzl", line 197, column 59, in _js_library_impl
	                npm_package_store_deps = gather_npm_package_store_deps(
			        File "/path/to/repos/aspect_rules_js/js/private/js_helpers.bzl", line 134, column 23, in gather_npm_package_store_deps
				                target[JsInfo].npm_package_store_deps
						Error: 'JsInfo' value has no field or method 'npm_package_store_deps'
						Available attributes: declarations, sources, transitive_declarations, transitive_sources
						ERROR: /path/to/repos/rules_rust/examples/wasm_bindgen/rules_js/BUILD.bazel:61:11: Analysis of target '//wasm_bindgen/rules_js:hello_world_wasm_lib' failed
						ERROR: Analysis of target '//wasm_bindgen/rules_js:hello_world_wasm_test' failed; build aborted:
						INFO: Elapsed time: 1.681s
						INFO: 0 processes.
						FAILED: Build did NOT complete successfully (136 packages loaded, 5244 targets configured)
						ERROR: Couldn't start the build. Unable to run tests
```

## Affected version

0.32.0

## Steps to reproduce:

Use the following branch:

https://github.com/jstuder-gh/rules_rust/tree/jsinfo_error_example

[...]

PR #2303, opened 2023-12-03
Use js_info macro to populate JsInfo provider (js_rust_wasm_bindgen)

jstuder-gh:
Uses the `js_info` macro in aspect_rules_js to create the `JsInfo` provider and provide defaults for any field that is not explicitly set.

Resolves https://github.com/bazelbuild/rules_rust/issues/2302


UebelAndre:
I’d like to keep the original test target but add your more library + test combo

UebelAndre (approved):
Thanks!

jstuder-gh:
Ok, I split it out into two tests as you suggested. One that consumes the WASM data directly and one the consumes it indirectly (via an intermediate `js_library`)

PR #2310, opened 2023-12-09
Platform transition added to Rust binary rules.

rickvanprim:
Without much insight into core Bazel's plans for supporting "flagless" builds across platforms, this seems like the most direct way to select the platform for a given Rust binary target (all the terminal rules) that doesn't need additional work to forward providers.

Currently I'm using `platform_transition_binary` (https://github.com/aspect-build/bazel-lib/blob/main/docs/transitions.md), but it doesn't work with Rust Analyzer/Clippy/etc. presumably due to not forwarding the providers.  In the case of this particular rule, it also doesn't work for something like `rust_shared_library()` which is explicitly not an `executable = True` target.

@illicitonion brought to my attention `with_cfg` (https://github.com/fmeum/with_cfg.bzl) which could be used as an alternative by consumers of `rules_rust` instead of this PR.  That feels reasonable, though certainly more work for individual users and appears to rely on a lot of Bazel black magic.

Along the same lines, this could wait on rule inheritance as presumably a more formally supported `with_cfg` approach, which would also be up to individual users and not `rules_rust`.

Tagging @aiuto, @katre, and @comius per @illicitonion's suggestion of relevant inputs 🙂


illicitonion:
Tagged folks - @aiuto @katre @fmeum - I know there's been long-standing discussion around flagless builds and arbitrary targets being transitionable - this PR seems like a reasonable step in that direction, but feels like a very specific instance/implementation of a very general pattern... Any thoughts on whether we should do this as-is, and whether there's anything we should change to be forwards-compatible with any larger more general efforts folks may be considering?

katre:
This change is compatible with our goals, as outlined in [Standard Platforms Transitions](https://github.com/bazelbuild/proposals/blob/main/designs/2023-06-08-standard-platform-transitions.md).

Once we have the common `change_platform` transition available, it'd be nice if you migrated to use it, but the intent is the same.

rust/private/rust.bzl
```diff
@@ -836,11 +836,35 @@ rust_library = rule(
...
+)
+
 rust_static_library = rule(
     implementation = _rust_static_library_impl,
-    attrs = dict(_common_attrs.items()),
+    attrs = dict(_common_attrs.items() + {
+        "platform": attr.label(
+            doc = "Optional platform to transition the static library to.",
+            default = None,
+        ),
```
illicitonion, line 860:
```suggestion
            default = None,
            providers = [platform_common.PlatformInfo],
        ),
```

rust/private/rust.bzl
```diff
@@ -859,15 +883,36 @@ rust_static_library = rule(
...
-    attrs = dict(
-        _common_attrs.items() + _experimental_use_cc_common_link_attrs.items() + {
-            "_use_grep_includes": attr.bool(default = True),
-        }.items(),
-    ),
+    attrs = dict(_common_attrs.items() + _experimental_use_cc_common_link_attrs.items() + {
+        "platform": attr.label(
+            doc = "Optional platform to transition the shared library to.",
+            default = None,
+        ),
```
illicitonion, line 907:
```suggestion
            default = None,
            providers = [platform_common.PlatformInfo],
        ),
```

rust/private/rust.bzl
```diff
@@ -958,13 +1003,37 @@ _rust_binary_attrs = dict({
...
+
 rust_binary = rule(
     implementation = _rust_binary_impl,
     provides = _common_providers,
-    attrs = dict(_common_attrs.items() + _rust_binary_attrs.items()),
+    attrs = dict(_common_attrs.items() + _rust_binary_attrs.items() + {
+        "platform": attr.label(
+            doc = "Optional platform to transition the binary to.",
+            default = None,
+        ),
```
illicitonion, line 1028:
```suggestion
            default = None,
            providers = [platform_common.PlatformInfo],
        ),
```

rust/private/rust.bzl
```diff
@@ -1096,10 +1165,19 @@ def _common_attrs_for_binary_without_process_wrapper(attrs):
 rust_binary_without_process_wrapper = rule(
     implementation = _rust_binary_impl,
     provides = _common_providers,
-    attrs = _common_attrs_for_binary_without_process_wrapper(_common_attrs.items() + _rust_binary_attrs.items()),
+    attrs = _common_attrs_for_binary_without_process_wrapper(_common_attrs.items() + _rust_binary_attrs.items() + {
+        "platform": attr.label(
+            doc = "Optional platform to transition the binary to.",
+            default = None,
+        ),
```
illicitonion, line 1172:
```suggestion
            default = None,
            providers = [platform_common.PlatformInfo],
        ),
```

rust/private/rust.bzl
```diff
@@ -1120,14 +1198,37 @@ rust_library_without_process_wrapper = rule(
...
 rust_test = rule(
     implementation = _rust_test_impl,
     provides = _common_providers,
-    attrs = dict(_common_attrs.items() +
-                 _rust_test_attrs.items()),
+    attrs = dict(_common_attrs.items() + _rust_test_attrs.items() + {
+        "platform": attr.label(
+            doc = "Optional platform to transition the test to.",
+            default = None,
+        ),
```
illicitonion, line 1223:
```suggestion
            default = None,
            providers = [platform_common.PlatformInfo],
        ),
```

illicitonion (approved):
LGTM!

I'll leave this open for a day or two in case any other maintainers have thoughts.

rickvanprim:
@katre thanks for the link.  I'd be happy to switch this over to `change_platform` whenever that's available 🙂

rust/private/rust.bzl
```diff
@@ -836,11 +836,36 @@ rust_library = rule(
         """),
 )
 
+def _rust_static_library_transition_impl(settings, attr):
```
fmeum, line 839: This looks good to me, but couldn't all these transitions be a single one shared by all rule types?

rust/private/rust.bzl
```diff
@@ -859,15 +884,37 @@ rust_static_library = rule(
...
 rust_shared_library = rule(
     implementation = _rust_shared_library_impl,
-    attrs = dict(
-        _common_attrs.items() + _experimental_use_cc_common_link_attrs.items() + {
-            "_use_grep_includes": attr.bool(default = True),
-        }.items(),
-    ),
+    attrs = dict(_common_attrs.items() + _experimental_use_cc_common_link_attrs.items() + {
+        "platform": attr.label(
+            doc = "Optional platform to transition the shared library to.",
```
fmeum, line 906: Ideally this would also describe how using this attribute interacts with `select` on other attributes: As far as I know, they will only work with Bazel 7+ and will see the "post-transition" platform.

fmeum (approved):

rust/private/rust.bzl
rickvanprim: An original version I did also had an optional `compilation_mode` transition that didn't apply to `rust_library()`, so I kept them different.  I've just moved them to using a `_change_platform()` call based on @katre's proposal.

rust/private/rust.bzl
rickvanprim: Could you elaborate on what you mean?  I'm viewing `rust_shared_library()` and `rust_static_library()` as terminal rules, where a platform transition makes sense, but technically they could be depended on by another binary that's targeting another platform, which would be broken (at least as far as C/Rust style dependencies go, totally fine for bundling).  A thought that comes to mind typing this would be putting the target platform into those providers and then `fail()`ing if they don't match when used as a dependency.  Assuming that's possible 😅

illicitonion: I think Fabian is referring to https://github.com/bazelbuild/bazel/issues/15157 which got fixed in Bazel 7.

fmeum: Thanks for providing the reference, @illicitonion. This is what I meant: Before Bazel 7, this may actually break `rust_*` targets with `select` in attributes. Do you happen to have a test for this? I'm not 100% sure what the behavior is in case the transition is a no-op, so maybe this only affects targets with `platform` set.

rickvanprim: Doesn't look like we can tag `attr`s as `nonconfigurable` unfortunately (I'm guessing that's builtin only), but that seems like it would be the ideal solution here?

fmeum: @illicitonion That's the extent of it as far as I understand, but it does warrant some testing. 

It would be possible to wrap the rules in macros that check the type of `platforms` and `fail` if they see a select, but that's cumbersome.

illicitonion (approved):

PR #2311, opened 2023-12-09
NixOS added as a Platform target including a pseudo target triple

rickvanprim:
Happy to rework this if there's a better approach.  Motivation is to get rid of this hack I've been using (including in the `nix_cross_compiling` example):

```bazel
# This seems like the best way for the Cargo generated dependencies to be
# aware of `x86_64-unknown-nixos-gnu`.
selects.config_setting_group(
    name = "x86_64-unknown-linux-gnu",
    match_any = [
        ":x86_64-unknown-linux-gnu_linux",
        ":x86_64-unknown-linux-gnu_nixos",
    ],
)

config_setting(
    name = "x86_64-unknown-linux-gnu_linux",
    constraint_values = [
        "@platforms//cpu:x86_64",
        "@platforms//os:linux",
    ],
)

config_setting(
    name = "x86_64-unknown-linux-gnu_nixos",
    constraint_values = [
        "@platforms//cpu:x86_64",
        "@platforms//os:nixos",
    ],
)
```


illicitonion (approved):
Looks reasonable based on how Bazel seems to handle `@platforms//os:nixos` - thanks!

PR #2312, opened 2023-12-09
`cargo-bazel` now supports `alias_rule` to customize the aliases generated

rickvanprim:
Motivator is to be able to set `default_alias_rule = "opt"` so that all Cargo fetched 3rd party dependencies are built with `compilation_mode=opt` regardless of what `compilation_mode` the local code is being built with.

Per my conversation with @illicitonion.  Please bikeshed on naming 🙂


crate_universe/src/rendering.rs
```diff
@@ -123,6 +139,13 @@ impl Renderer {
         let header = self.engine.render_header()?;
         starlark.push(Starlark::Verbatim(header));
 
+        // Load `transition_opt_alias` if `transition_opt=True`.
+        if self.transition_opt {
+            starlark.push(Starlark::Verbatim(
+                r#"load("//:transition_opt_alias.bzl", alias = "transition_opt_alias")"#.to_owned(),
```
rickvanprim, line 145: Instead of overwriting `alias`, which was the easiest for Starlark serialization, I could allow this to be specified on a per `crate.annotation()` basis and serialize explicitly below `alias()` or `transition_opt_alias()`.  Does that seem worth doing?  As is, to individually opt a crate out means directly referring to the target instead of through the alias.

crate_universe/test_data/serialized_configs/BUILD.bazel
```diff
@@ -32,6 +32,7 @@ write_file(
                 "x86_64-pc-windows-msvc",
                 "x86_64-apple-darwin",
             ],
+            transition_opt = True,
```
illicitonion, line 35: Can you write an integration test for this behaviour?

This should be observable; from a `cargo_build_script` you can get `std::env::var("OPT_LEVEL")` which will be `0` for fastbuild and `3` for opt.

Hopefully we can in two different crate_universe invocations, one which transitions and one which doesn't, verify that the opt levels are set correctly.

crate_universe/3rdparty/crates/transition_opt_alias.bzl
illicitonion, line 1: Why are there two copies if this file?

crate_universe/private/crates_repository.bzl
```diff
@@ -315,6 +315,10 @@ CARGO_BAZEL_REPIN=1 CARGO_BAZEL_REPIN_ONLY=crate_index bazel sync --only=crate_i
             doc = "A set of all platform triples to consider when generating dependencies.",
             default = SUPPORTED_PLATFORM_TRIPLES,
         ),
+        "transition_opt": attr.bool(
```
illicitonion, line 318: I would lean towards the API here being to take an optional transition, rather than a boolean for whether to do one specific transition. Something like:

```
crate.annotation(
    transition_alias = "@rules_rust//crate_universe/3rdparty/crates:transition_opt_alias.bzl:transition_opt_alias",
)
```

We could allow `"opt"` as a magic value which would alias to that transition. We could also allow `alias` (or `native.alias`) as a magic value (also default value) if that makes sense.

That means that if someone needs to apply some other transition, they can do so, rather than also needing to bolt that on as well.

We can agree that the API for this transition_alias is that it should be a rule which takes exactly one attr named `actual`.

WDYT?

illicitonion (commented):
This looks great, thanks!

I'll leave this one open for a day or two in case anyone else wants to jump in, otherwise I left a few comments :))

crate_universe/3rdparty/crates/transition_opt_alias.bzl
rickvanprim: Everything under `3rdparty/crates` is the output from `cargo-bazel` is my understanding, as `cargo-bazel` needs its own output to bootstrap itself.  So in this case, now that `cargo-bazel` outputs `transition_opt_alias.bzl`, it's showing up here in the repo.  I could make it conditionally output, but I figured unconditionally being output is better incase anyone during development/debugging/whatever wants to turn `transition_opt=False` and then manually make use of `transition_opt_alias` (and would otherwise be frustrated when that file is suddenly no longer present).

illicitonion: Makes sense :) I'd maybe be tempted to just check this file in to `rules_rust` as part of our public API, rather than re-generating it per repository, but also, not needing to grow our public API is probably a better trade-off.

(Answers to my other comment about taking a `transition_alias` attr instead of `transition_opt` attr may change the trade-offs here :))

crate_universe/src/rendering.rs
```diff
@@ -33,6 +33,7 @@ pub struct Renderer {
     config: RenderConfig,
     supported_platform_triples: BTreeSet<TargetTriple>,
     generate_target_compatible_with: bool,
+    alias_rule: AliasRule,
```
illicitonion, line 36: This feels like it should live in `RenderConfig` and be called `default_alias_rule`?

crate_universe/src/rendering.rs
```diff
@@ -124,6 +138,26 @@ impl Renderer {
...
+                .flat_map(|dep| &context.crates[&dep.id].alias_rule),
+        ) {
+            if let Some(bzl) = alias_rule.bzl() {
+                loads.entry(bzl).or_default().insert(alias_rule.rule());
+            }
+        }
+        for (bzl, items) in loads {
+            starlark.push(Starlark::Load(Load {
+                bzl: bzl,
+                items: items,
```
illicitonion, line 157: We should hard error if any of these alias rules are literally called `"alias"`, as it will conflict with the fact that we reference `alias` below at line 243.

crate_universe/src/rendering.rs
```diff
@@ -866,6 +906,7 @@ mod test {
             mock_render_config(None),
             mock_supported_platform_triples(),
             true,
+            true,
```
illicitonion, line 909: I must be reading something wrong, but I don't see a `new` which takes two booleans? Does this need updating? (And throughout)

illicitonion (commented):
Generally LGTM modulo the comments :)

crate_universe/src/config.rs
```diff
@@ -181,6 +186,43 @@ pub enum Checksumish {
     },
 }
 
+#[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Clone)]
+pub enum AliasRule {
+    #[default]
+    #[serde(rename = "default")]
```
illicitonion, line 192: I would be tempted to call this `alias` or `native.alias` rather than `default`? WDYT?

illicitonion (approved):
Surprisingly clean! Nicely put together, thanks!

Just one bikeshed to paint :)

crate_universe/src/config.rs
rickvanprim: Will do!

PR #2314, opened 2023-12-10
Allow additional targets to compile with bzlmod

matts1:
See #2246 for how we declare dependencies automatically for third party crates. I'm using the same mechanism for all other repositories.

Note that the MODULE.bazel changes in use_repo are autogenerated, and if you change the third party crates you depend on, you should get an error message saying something like "run this command to fix it" on your bzlmod presubmit environment.


rust/private/extensions.bzl
```diff
@@ -1,11 +1,48 @@
...
+load("//util/import:deps.bzl", "import_deps")
+load("//wasm_bindgen:repositories.bzl", "rust_wasm_bindgen_dependencies")
 
-def _internal_deps_impl(_module_ctx):
+def _internal_deps_impl(module_ctx):
+    # This should contain the subset of WORKSPACE.bazel that defines
+    # repositories.
+
+    # We don't want rules_rust_dependencies, as they contain things like
+    # rules_cc, which is contained in the BCR.
```
illicitonion, line 18: Excuse my bzlmod ignorance - why is this a problem? If we switched to `rules_rust_dependencies`, what would go wrong? Can you update this comment?

MODULE.bazel
```diff
@@ -7,19 +7,122 @@ module(
...
+bazel_dep(
+    name = "protobuf",
+    version = "21.7",
+    repo_name = "com_google_protobuf",
+)
 
 internal_deps = use_extension("//rust/private:extensions.bzl", "internal_deps")
 use_repo(
     internal_deps,
+    "bazelci_rules",
```
illicitonion, line 40: It's not super clear to me how much of this a real user would need to emulate and if there's any tooling we should be supplying to help... Is there an example somewhere?

illicitonion (approved):

MODULE.bazel
matts1: Not sure what you mean by emulate? The tooling is already there, if I understand your question correctly.

For example, I removed the bazelci_rules, then tried to build, and got the following:

```
$ bazel build --enable_bzlmod
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 5c961a83-2149-45ba-a6b6-c0d403c9bfa9
DEBUG: /usr/local/google/home/msta/chromiumos_head/bazel_deps/rules_rust/MODULE.bazel:8:6: WARNING: The rules_rust Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
WARNING: /usr/local/google/home/msta/chromiumos_head/bazel_deps/rules_rust/MODULE.bazel:42:30: The module extension internal_deps defined in @rules_rust//rust/private:extensions.bzl reported incorrect imports of repositories via use_repo():

Not imported, but reported as direct dependencies by the extension (may cause the build to fail):
    bazelci_rules

 ** You can use the following buildozer command to fix these issues:

buildozer 'use_repo_add @rules_rust//rust/private:extensions.bzl internal_deps bazelci_rules' //MODULE.bazel:all
```

rust/private/extensions.bzl
matts1: Done

MODULE.bazel
illicitonion: What I mean is that things like `rules_rust_wasm_bindgen__anyhow-1.0.71` are internal dependencies of `rules_rust`, but are required if you want to use some parts of functionality in `rules_rust` from an external repo.

From what little I do know about bzlmod (again, apologies for my ignorance), one of the unergonomic pieces around it is that things like rules_go's third-party support requires you to `use_repo` every third-party library, even though they were initially loaded by a module extension via a `go.mod` file.

So my question is: Will users themselves need to `use_repo` the 21 `rules_rust_wasm_bindgen`-prefixed repos themselves if they want to use bindgen in their repo, or will this automatically be done because `rules_rust` does it, or is there some middle-ground where we can provide a utility function which will do it for them if they call it?

(None of this blocks merging the PR, which I will gladly do shortly, I'm just curious, and want to understand if there are things we should be doing to make things easier for our users to consume, or feedback we should be giving to the bzlmod folks about things which aren't ideal)

matts1: Ah, I understand what you mean now. To clarify, this does two things:
* Building `//wasm_bindgen` with bzlmod should now work.
* Building `@rules_rust//wasm_bindgen` should now work

Longer explanation:
Users will just write `bazel_dep(name = "rules_rust", version = "...")` in their MODULE.bazel. This implicitly loads rules_rust's MODULE.bazel, which then declares `rules_rust_wasm_bindgen__anyhow-1.0.71`. However, since you didn't (and can't, because it's a private module extension) declare a dependency on it in your own MODULE.bazel, it's hidden from the main repo.

However, if I were to write `bazel build @rules_rust//wasm_bindgen`, it should work just fine, because it knows that `@rules_rust` can access it.

matts1:
Build appears to be failing because of the upgrade to bazel 7.0.0. I created an empty PR that's failing these same presubmits (#2316)

matts1:
Filed #2317

PR #2315, opened 2023-12-12
Introduced `TargetTriple` newtype.

rickvanprim:
As discussed on Slack with @illicitonion (hopefully).  With the `Select` refactor, we presumably want `selects` to be keyed by `TargetTriple` instead of `String`.  Please let me know if you have a better interface for `TargetTriple` and/or any bikesheds 🙂

### DIFF
--- a/crate_universe/src/rendering.rs
+++ b/crate_universe/src/rendering.rs
@@ -22,6 +22,7 @@ use crate::utils::starlark::{
     Package, RustBinary, RustLibrary, RustProcMacro, Select, SelectDict, SelectList, SelectMap,
     Starlark, TargetCompatibleWith,
 };
+use crate::utils::target_triple::TargetTriple;
 use crate::utils::{self, sanitize_repository_name};
 
 // Configuration remapper used to convert from cfg expressions like "cfg(unix)"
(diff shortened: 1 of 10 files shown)


crate_universe/src/rendering.rs
```diff
@@ -372,8 +373,12 @@ impl Renderer {
                 attrs.map_or(&empty_list, |attrs| &attrs.compile_data),
             ),
             crate_features: SelectList::from(&krate.common_attrs.crate_features)
+                // Removed in big `Select` refactor PR.
```
illicitonion, line 376: Good to know, but can you remove these comments in the mean time? :)
```suggestion
```

crate_universe/src/rendering.rs
```diff
@@ -530,8 +535,12 @@ impl Renderer {
                 &krate.common_attrs.compile_data,
             ),
             crate_features: SelectList::from(&krate.common_attrs.crate_features)
+                // Removed in big `Select` refactor PR.
```
illicitonion, line 538:
```suggestion
```

illicitonion (approved):
Very nice, thanks!

PR #2322, opened 2023-12-13
Uprev bazel CI to bazel 7.0.0

### LINKED ISSUE #2317: Bazel 7.0.0 breaks presubmits
See #2316 (an empty PR that  presubmits fail on).

We implicitly use the latest released version of bazel, and that breaks the presubmits. I tried investigating the rustfmt negative fail case, and bazelisk bisect gave some very unclear results. I think it's the version number bump that broke it, because it also fails on [7.0.0-pre.20231011.2](https://github.com/bazelbuild/bazel/releases/tag/7.0.0-pre.20231011.2), but not the commit before the revbump, and the only thing that changed was the version number. I assume there was some flag flip, but I couldn't work out why.

To repro, remove the `rm -rf` in the rustfmt_failure_test.sh, and then you can just cd into the directory. FYI, you'll need to do bazel clean before each invocation, because if you manage to get it to  succeed it'll just cache hit instead.

I think this might be a bug with bazel itself, because the error looks really wierd - the file it claims not to exist does exist, but it's a symlink (though it's still a symlink in 6.4.0, which it works on).

matts1:
Fixes #2317


illicitonion (approved):
Thanks!

PR #2323, opened 2023-12-13
Remove `.bazelversion` file to always use latest release available.

UebelAndre:
relates to https://github.com/bazelbuild/rules_rust/issues/2317


illicitonion (commented):
I'm 50:50 on this one... I generally like pinning things and having CI be deterministic, but _also_ value testing the latest versions... In an ideal world, I think we'd pin, but have some process which automatically makes an update PR when there are new versions, but I realise that's non-trivial and am not suggesting we build this... WDYT?

UebelAndre:
> I'm 50:50 on this one... I generally like pinning things and having CI be deterministic, but _also_ value testing the latest versions... In an ideal world, I think we'd pin, but have some process which automatically makes an update PR when there are new versions, but I realise that's non-trivial and am not suggesting we build this... WDYT?

I think pinning is good but would rather that be something handled by bazel-ci in general. The pin should be whatever version is either explicitly defined by presubmit.yaml or by the CI environment where ideally before releasing the Bazel team tests all active projects in the bazel org by bumping the pin there and running CI. But as it is now this will get stale and I think we'll get less info than if we allow for the volatile floating pin.

PR #2325, opened 2023-12-13
cargo-bazel: Corrected sparse registry splicing code

JulianSchmid:
This is a fix for supporting sparse-registries (private).

### Issue:

When we tried to use crate_universe together with an internal sparse registry (Artifactory) we ran into an issue. `cargo-bazel splice` was not correctly detecting sparse registries as "sparse", even though the URLs correctly started with the "sparse+https://" prefix in the Cargo.lock file. Instead it re-interpreted them as registries with a git index.

The root cause of this is that`cargo_lock::Lockfile` automatically removes the prefix `sparse+` from the URL and the only way to determine that a registry is sparse is to check the `kind`. The code previously assumed it could check for sparse registries by checking if the url starts with "sparse+https://".

### Problem solution:

This pull requests adapts the code so it uses the lockfile "SourceKind" instead of the URL to check if a registry is sparse or not.

### DIFF
--- a/crate_universe/src/splicing.rs
+++ b/crate_universe/src/splicing.rs
@@ -11,6 +11,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use anyhow::{anyhow, bail, Context, Result};
+use cargo_lock::package::SourceKind;
 use cargo_toml::Manifest;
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Thanks!

PR #2326, opened 2023-12-13
Add crate for testing build configs

illicitonion:
I intend to publish this to crates.io after review.

This will be used to test
https://github.com/bazelbuild/rules_rust/pull/2312.


crate_universe/test_data/test_data_passing_crate/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub fn get_opt_level() -> &'static str {
```
rickvanprim, line 1: This should be `pub const fn` so we can static assert on it.

crate_universe/test_data/test_data_passing_crate/src/lib.rs
```diff
@@ -0,0 +1,7 @@
...
+    env!("BUILD_SCRIPT_OPT_LEVEL")
+}
+
+pub fn get_out_dir() -> &'static str {
```
rickvanprim, line 5: Same as above.

rickvanprim (approved):

UebelAndre (approved):

PR #2328, opened 2023-12-13
rust_analyzer: run bazel aquery with --include_artifacts

neilisaac:
Explicitly set --include_artifacts to override users' .bazelrc file settings.

Evidently I had set `aquery --include_artifacts=false` in ~/.bazelrc, which was very difficult to track down from the error when generating the project file: 
```
Error: missing field `outputIds` at line 9 column 3
```


UebelAndre (approved):
Thanks!

PR #2336, opened 2023-12-15
Include all toolchain files in the current toolchain

illicitonion:
rustc assumes it can read things like librustc_driver, and if you don't
include it, it gets sad.


test/toolchain/BUILD.bazel
```diff
@@ -1,3 +1,19 @@
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load(":toolchain_test.bzl", "toolchain_test_suite")
 
 toolchain_test_suite(name = "toolchain_test_suite")
+
+genrule(
```
UebelAndre, line 6: Can you use `run_binary` instead? I hate genrule lol

UebelAndre (approved):
Seems fine, just had one request 😄

test/toolchain/BUILD.bazel
illicitonion: I don't think so, because `run_binary` requires that `argv[0]` is the tool, but we don't have a single tool here we have a toolchain we need to expand it from `$(RUSTC)`?

UebelAndre: You’re right, never mind!

PR #2343, opened 2023-12-18
fix: use `watch_file` in `.envrc`

cgrindel:
`nix_direnv_watch_file` is deprecated and appears to have been removed from some `direnv` distributions.


rickvanprim (approved):

UebelAndre (approved):
Thanks!

PR #2347, opened 2023-12-20
chore: add `//bzlmod:hello_world_test` and use it as BCR presubmit

cgrindel:
- Create `bzlmod/hello_world` package based upon `examples/bzlmod/hello_world`.
- Add `rules_bazel_integration_test` so that `//bzlmod:hello_world_test` can be executed as a test from the parent workspace.
- Add `filegroup` targets named `all_files` to collect files for all of the packages that are necessary to use `rules_rust`. These are an input to the `rules_bazel_integration_test` test.
- Replace `.bcr/presubmit.yml` to test `//bzlmod:hello_world_test`.


.gitignore
```diff
@@ -14,6 +14,7 @@
 /test/no_std/bazel-*
 /docs/bazel-*
 user.bazelrc
+MODULE.bazel.lock
```
matts1, line 17: I believe it's recommended to commit MODULE.bazel.lock, similar to how you'd commit Cargo.lock.

.bazelrc
```diff
@@ -54,9 +54,7 @@ build --nolegacy_external_runfiles
 ## Bzlmod
 ###############################################################################
 
-# TODO: migrate all dependencies from WORKSPACE to MODULE.bazel
-# https://github.com/bazelbuild/rules_rust/issues/2181
-common --noenable_bzlmod
+common --experimental_enable_bzlmod
```
matts1: Can we put this in its own PR and get it submitted first? Speaking from experience when migrating our product to bzlmod, I don't have faith that this won't break anything, and I'd like to be able to revert this change individually..

MODULE.bazel
```diff
@@ -31,6 +31,10 @@ bazel_dep(
     version = "21.7",
     repo_name = "com_google_protobuf",
 )
+bazel_dep(
+    name = "rules_bazel_integration_test",
```
matts1, line 35: dev_dependency = True

MODULE.bazel
```diff
@@ -140,3 +144,14 @@ use_repo(rust, "rust_host_tools")
 
 cargo_bazel_bootstrap = use_extension("//crate_universe/private/module_extensions:cargo_bazel_bootstrap.bzl", "cargo_bazel_bootstrap")
 use_repo(cargo_bazel_bootstrap, "cargo_bazel_bootstrap")
+
+# rules_bazel_integration_test Configuration
+bazel_binaries = use_extension(
+    "@rules_bazel_integration_test//:extensions.bzl",
+    "bazel_binaries",
```
matts1, line 151: Maybe dev_dependency = True.

Not sure if it's required if you already have it in the bazel_dep

.bazelrc
```diff
@@ -5,6 +5,14 @@
 ## https://bazel.build/docs/best-practices#bazelrc-file
 ###############################################################################
 
+# These are workspaces that are tested using rules_bazel_integration_test. The
+# files need to be found by the parent workspace for packaging, but we do not
+# want the child workspace packages to be evaluated by the parent workspace.
+# To update these lines, execute 
+# `bazel run @rules_bazel_integration_test//tools:update_deleted_packages`
+build --deleted_packages=bzlmod/hello_world
```
matts1, line 13: Should we just pull all your changes here into examples/bzlmod/hello_world?

matts1 (changes requested):

.bazelrc
cgrindel: Yeah. That was an accident.

.gitignore
cgrindel: My bad. I mistakenly thought this repo was not checking them in. I see now that bzlmod-specific repos are. I will add the lock file for the new workspace.

MODULE.bazel
cgrindel: I purposefully did not add that. The issue is that a `dev_dependency`  in a dependent module will not be loaded. However, if the release archive ships with any references to the `dev_dependency`, queries will fail. [This issue](https://github.com/keith/buildifier-prebuilt/issues/85#issuecomment-1860509919) represents such a case.

MODULE.bazel
cgrindel: Please see [my other response](https://github.com/bazelbuild/rules_rust/pull/2347#discussion_r1435297736).

.bazelrc
cgrindel: This PR allows someone to run `bazel test //bzlmod:hello_world` from the root module. Underneath the covers, it will launch another Bazel process and execute tests in the `bzlmod/hello_world` workspace. I use this pattern in [rules_swift_package_manager](https://github.com/cgrindel/rules_swift_package_manager/blob/main/bzlmod/BUILD.bazel) and [bazel-starlib](https://github.com/cgrindel/bazel-starlib/blob/main/examples/BUILD.bazel). In essence, one can run all tests for a  child workspace from the parent workspace. 

For this to work, the parent workspace needs to see the files for the child workspace, but not process them as Bazel packages. That is what `--deleted_packages` does. The other way to prevent the parent workspace from processing the child workspaces is to add them to `.bazelignore`. However, this makes those directories completely inaccessible to the parent.

I noticed that in `rules_rust` the child workspaces are ignored by the parent workspace by adding them to `.bazelignore`. The child workspaces are exercised by changing into the directories and running `bazel test //...`.  

So, the question for the maintainers is whether they are interested in being able to test the child workspaces from the parent workspace. If so, then we need the `--deleted_packages` and the `filegroup(name="all_files")`. Otherwise, we can use the pattern that is currently being used for other child workspaces in this repo.

@matts1 @illicitonion What do you prefer?

cgrindel: One benefit of collecting the runtime files in a filegroup and testing them using `rules_bazel_integration_test` is that you have the list of files that should be included in the release archive. [In `rules_swift_package_manager`](https://github.com/cgrindel/rules_swift_package_manager/blob/main/release/BUILD.bazel#L54-L61), I create the release archive using the same files that are used for the `rules_bazel_integration_test` tests.

MODULE.bazel
matts1: Ah, I see.

@Wyverald This seems to be a problem with `dev_dependency` in general. Is there any particular recomendation from the bzlmod team on standard practice for dealing with issues like this?

.bazelrc
matts1: > So, the question for the maintainers is whether they are interested in being able to test the child workspaces from the parent workspace.

I'm not especially interested in this - I find it much easier to debug something when it's not a generated workspace, and having to manage the filegroups seems like a steep price to pay for me. That being said, I'll defer to to others, since I'm not really representative of most developers since I'm very familiar with the bzlmod examples and the workflows for dealing with changes that affect bzlmod.

> One benefit of collecting the runtime files in a filegroup and testing them using rules_bazel_integration_test is that you have the list of files that should be included in the release archive.

I'm personally quite fine with just tarring up everything except `examples`, unless you have some justification otherwise.

illicitonion: I'd lean towards another child workspace, because of the simplicity it provides - not needing the filegroup, being able to trivially just `cd` in to debug, etc. I don't have a strong objection to the `--deleted_packages` dance - open to the discussion, but would prefer another child workspace absent that discussion :)

cgrindel: No problem. I will modify the PR and ping folks when it is ready.

cgrindel: > I'm personally quite fine with just tarring up everything except examples, unless you have some justification otherwise.

For repos that declare dev dependencies in their `MODULE.bazel`, you would not want to include any files that reference the dev dependencies. It can cause issues with queries executed by a client workspace, because the dev dependencies are not loaded. I don't see any dev dependencies in this repo. So, it should not matter.

cgrindel:
@matts1 @illicitonion I removed `rules_bazel_integration_test` from the PR. This should be ready for another review. 

BTW, the GitHub UI is showing the `.bcr/presubmit.yml` file as being deleted. It should not be deleted. The branch shows that it is there. If it appears deleted for you, please let me know.

bzlmod/hello_world/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
...
+)
+
+sh_test(
+    name = "hello_world_test",
+    srcs = ["hello_world_test.sh"],
+    data = [
+        ":hello_world",
+    ],
+    deps = [
+        "@bazel_tools//tools/bash/runfiles",
```
illicitonion, line 22:
```suggestion
    args = [
        "$(rootpath :hello_world)",
```

bzlmod/hello_world/hello_world_test.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# MARK - Functions
+
+fail() {
+  echo >&2 "$@"
+  exit 1
+}
+
+# MARK - Test
+
```
illicitonion, line 29:
```suggestion
fail() {
   echo >&2 "$@"
   exit 1
 }
 
if [[ "$#" -ne 1 ]]; then
  fail "Usage: $0 /path/to/hello_world"
fi
hello_world="$1"
```

illicitonion (approved):
LGTM, thanks!

PR #2358, opened 2023-12-25
Include more cc_toolchain flags in RustBindgen actions

UebelAndre:
This change parses `cc_toolchain` flags for `c++-compile` actions for use in bindgen. This change aims to allow the action to gain toolchain specific include paths as well as a more accurate `--sysroot`.

### DIFF
--- a/bindgen/private/bindgen.bzl
+++ b/bindgen/private/bindgen.bzl
@@ -14,6 +14,11 @@
 
 """Rust Bindgen rules"""
 
+load(
+    "@bazel_tools//tools/build_defs/cc:action_names.bzl",
+    "CPP_COMPILE_ACTION_NAME",
+)
+load("@rules_cc//cc:defs.bzl", "CcInfo")
 load("//rust:defs.bzl", "rust_library")
 
 # buildifier: disable=bzl-visibility
@@ -69,7 +74,7 @@ def rust_bindgen_library(
         cc_lib = cc_lib,
         bindgen_flags = bindgen_flags or [],
         clang_flags = clang_flags or [],
-        tags = tags,
+        tags = ["manual"],
     )
 
     rust_library(
@@ -81,8 +86,6 @@ def rust_bindgen_library(
     )
 
 def _rust_bindgen_impl(ctx):
-    toolchain = ctx.toolchains[Label("//bindgen:toolchain_type")]
-
     # nb. We can't grab the cc_library`s direct headers, so a header must be provided.
     cc_lib = ctx.attr.cc_lib
     header = ctx.file.header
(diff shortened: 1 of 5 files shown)


UebelAndre:
@scentini @krasimirgg @hlopko do you folks know if there's a way just to grab arguments from `cc_toolchain` using `cc_common` or something? As you can see in this PR I'm only interested in include flags. Ideally I would be able to detect those directly in `_rust_bindgen_impl` by allowing users to optionally pass a list of toolchain features to include in the `RustBindgen` action.

UebelAndre:
> @illicitonion @scentini @krasimirgg @hlopko do you folks know if there's a way just to grab arguments from `cc_toolchain` using `cc_common` or something? As you can see in this PR I'm only interested in include flags. Ideally I would be able to detect those directly in `_rust_bindgen_impl` by allowing users to optionally pass a list of toolchain features to include in the `RustBindgen` action.

I think I found a solution!

bindgen/private/bindgen.bzl
```diff
@@ -129,26 +129,46 @@ def _rust_bindgen_impl(ctx):
...
+    )
+    compile_flags = cc_common.get_memory_inefficient_command_line(
+        feature_configuration = feature_configuration,
+        action_name = CPP_COMPILE_ACTION_NAME,
+        variables = compile_variables,
+    )
+
+    # Parse known arguments from the list of potential compile flags
+    open_arg = False
+    for arg in compile_flags:
```
illicitonion, line 150: Why are we filtering these, rather than just appending all of them?

bindgen/private/bindgen.bzl
```diff
@@ -129,26 +129,46 @@ def _rust_bindgen_impl(ctx):
...
+        if open_arg:
+            args.add(arg)
+            open_arg = False
+            continue
+
+        if arg in ctx.attr.clang_flags:
```
illicitonion, line 156: This filtering seems a little odd? If someone has specified a duplicate flag, that seems fine (and the order may matter) and if it's a problem, they should probably remove it from the `clang_flags` attr?

illicitonion (commented):
This generally looks good but I have a couple of questions on it

bindgen/private/bindgen.bzl
UebelAndre: My hope was to allow the toolchain to specify the order of the flags

bindgen/private/bindgen.bzl
UebelAndre: Yeah, I’m not super worried about duplicates. And again, I was hoping to use the toolchain to dictate order. In an ideal world I would be able to only get flags from a user defined list of features but that doesn’t seem possible.

bindgen/private/bindgen.bzl
illicitonion: What's stopping us from just making this whole `for`-loop instead just be:

```
args.add_all(compile_flags)
args.add_all(ctx.attr.clang_flags)
```

I don't think I'm understanding the feature stuff enough to grasp this - are there specific features' flags we're trying to include/exclude which this for-loop-plus-filter is a reasonable proxy for?

UebelAndre: I originally tried that and then got bullied by non clang flags. I would love to do this eventually but am nervous to roll out too large a change at once since I don't feel like these rules are that well understood.

illicitonion: Aha, I see - sounds good; I just pushed a few comments to your branch, PTAL and merge if you're happy!

illicitonion (approved):

bindgen/private/bindgen.bzl
UebelAndre: Love it! Thanks!

PR #2361, opened 2023-12-27
Update `rust_bindgen_library` to only link into the direct consumer

UebelAndre:
Most uses of [bindgen](https://github.com/rust-lang/rust-bindgen) are via [build scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html) which typically use either `-Lnative=` to specify linker paths for the current target being built. The way the rules are currently defined, has the bindgen library linked directly into the final library which leads to some annoying ordering errors compared to the same project building with Cargo. This PR aims to support the Cargo style of linking via a `leak_symbols = False` flag, thus changing the dependency graph when linking from:
```
cc_lib -> bindgen -> lib_a -> lib_b -> bin
    \__________________________________^
```
to:
```
cc_lib -> bindgen -> lib_a -> lib_b -> bin
```

Note that `leak_symbols` is intended to be used as an escape hatch to restore the original behavior and will be deleted in the future.


UebelAndre:
`linkstatic` is probably a bad name. Open to suggestions but I'll think harder about it.

bindgen/private/bindgen.bzl
```diff
@@ -64,27 +65,109 @@ def rust_bindgen_library(
...
+        output = link_search_paths,
+        content = "\n".join([
+            "-Lnative=${{pwd}}/{}".format(path)
+            for path in depset(linker_search_paths).to_list()
+        ]),
+    )
+
+    return BuildInfo(
+        out_dir = ctx.file._bindgen_out_dir,
+        rustc_env = empty_file,
```
illicitonion, line 163: I'd be tempted to make theses optional rather than having to pass an empty file (and make an entire rule to make an empty dir), but I don't feel strongly

bindgen/private/bindgen.bzl
```diff
@@ -220,6 +316,17 @@ rust_bindgen = rule(
             allow_single_file = True,
             mandatory = True,
         ),
+        "linkstatic": attr.bool(
```
illicitonion, line 319: Yeah, I'm not sure about this naming, as it's not the same as `cc_library`'s `linkstatic`, but I don't have a better one off the top of my head... What may motivate someone to set this to `False`? Could you describe in a sentence what the `False` version of this attr does/means?

illicitonion (commented):
Generally looks reasonable, added a few thoughts

bindgen/private/bindgen.bzl
UebelAndre: Good idea, Done!

bindgen/private/bindgen.bzl
UebelAndre: if you have 
```
cc_lib -> bindgen -> lib_a -> lib_b -> bin
```
the `False` version of this flag will be the equivalent to the following when linking.
```
cc_lib -> bindgen -> lib_a -> lib_b -> bin
    \__________________________________/
```
This can cause link order issues that are annoying to deal with.

illicitonion: I see... When would someone _want_ the other behavior?

UebelAndre: The `False` behavior? I couldn't tell you... My hope was only to leave the escape hatch for the situations I couldn't think of.

illicitonion: Cool, let's ship this as is, treating `leak_symbols` as just a "way to get unblocked if this release breaks you" mechanism (and add a print warning if it's set asking people to file an issue), and aim to delete the setting completely in a few months?

UebelAndre: That's the plan!

UebelAndre:
@illicitonion this good to go?

bindgen/private/bindgen.bzl
```diff
@@ -200,6 +268,19 @@ def _rust_bindgen_impl(ctx):
         tools = tools,
     )
 
+    if ctx.attr.leak_symbols:
+        providers = [cc_common.merge_cc_infos(
```
illicitonion, line 272:
```suggestion
        print("WARN: ctx.attr.leak_symbols is set to True for {} - please file an issue at https://github.com/bazelbuild/rules_rust/issues explaining why this was necessary, as this support will be removed soon.".format(ctx.label))
        providers = [cc_common.merge_cc_infos(
```

illicitonion (approved):

Issue #1408, opened 2022-06-13
Allowing Label to own it's content


UebelAndre:
Looking at https://github.com/bazelbuild/rules_rust/tree/0.5.0/util/label I'm wondering why it effectively only wraps a some owned data elsewhere. Would it make sense to allow `Label` to be a struct which owns a string and can comfortably parse out `repository_name`, `package_name`, and `target`?

UebelAndre:
cc @hlopko

illicitonion:
Yeah, I think using a `Cow` should work fine here?

(I suspect Label _actually_ owning its backing data also wouldn't be the end of the world - @hlopko can hopefully speak to that :))

PR #2363, opened 2023-12-27
Replaced multiple usages of `String` with `Label`.

rickvanprim:


crate_universe/3rdparty/crates/BUILD.ahash-0.7.6.bazel
```diff
@@ -245,6 +245,6 @@ cargo_build_script(
 
 alias(
     name = "build_script_build",
-    actual = "ahash_build_script",
+    actual = "//ahash_build_script:ahash_build_script",
```
rickvanprim, line 248: Is this change acceptable/desirable?

crate_universe/src/rendering.rs
```diff
@@ -674,7 +677,8 @@ impl Renderer {
                         &dependency.id.version,
                         &dependency.target,
                     );
-                    aliases.insert((label, alias.clone()), configuration.clone());
+                    // Should `Select` support `BTreeMap<Label, T>` or `BTreeMap<U, T>`?
```
rickvanprim, line 680: The above change would be easy enough, but then how to update `SelectDict<T>` is also a question.

UebelAndre:
I opened https://github.com/bazelbuild/rules_rust/issues/1408 some time ago. I think it'd be really good to consolidate implementations so that improvements to one implementation can be more impactful. Do you wanna first take a stab at being able to use this class in `crate_universe`? (without adding a dependency on regex 😅)

rickvanprim:
@UebelAndre how do you see `cargo-bazel` depending on `util/label.rs`?  Also switching implementations won't remove `regex` from `cargo-bazel` as at least `tera` also depends on it (from a very cursory search).

Where `util/label.rs` is used?  Is it meant for customers of `rules_rust` via `@rules_rust/util/label:label`?  I'm not seeing anything depending on it in the repo.

Regarding the linked issue, I'm not sure what the other usages are, but certainly in `cargo-bazel` the current direction is to favor ease of code, not minimizing allocations, so we would want the owned form of `Label` (like the current Starlark one).

crate_universe/3rdparty/crates/BUILD.ahash-0.7.6.bazel
```diff
@@ -245,6 +245,6 @@ cargo_build_script(
 
 alias(
     name = "build_script_build",
-    actual = "ahash_build_script",
+    actual = "//:ahash_build_script",
```
rickvanprim, line 248: Hmm another issue, `Label::from_str(":foo").unwrap().repr() == "//:foo"` which is not equivalent.  Same issue with a plain `foo` instead of `:foo`.

rickvanprim: Seems like making emitting `//` conditional on either `repository` or `package` being present would work.  I have mixed feelings on whether that's "correct" though.

UebelAndre:
> @UebelAndre how do you see `cargo-bazel` depending on `util/label.rs`? Also switching implementations won't remove `regex` from `cargo-bazel` as at least `tera` also depends on it (from a very cursory search).
> 
> Where `util/label.rs` is used? Is it meant for customers of `rules_rust` via `@rules_rust/util/label:label`? I'm not seeing anything depending on it in the repo.
> 
> Regarding the linked issue, I'm not sure what the other usages are, but certainly in `cargo-bazel` the current direction is to favor ease of code, not minimizing allocations, so we would want the owned form of `Label` (like the current Starlark one).

I would like it if `cargo-bazel` could delete `//crate_universe/src/util/starlark/label.rs` or at least not define a `Label` struct in favor of `//util/label`. That target is intended for external consumers of rules_rust. I guess ultimately the thing I want to consolidate is the actual parsing logic. Maybe that can be done separately but had hoped to invest more in `//util/label` vs the implementation in `cargo-bazel`.

rickvanprim:
I'm partial to `//crate_universe/src/util/starlark/label.rs` strictly being for rendering purposes (presumably as `struct Label(String)` and dumb otherwise (sorry @illicitonion 😅).  If we're aligned on direction, I can add a dependency on `//util/label:label` (and a Cargo path dependency on a new `Cargo.toml` there) and simplify the Starlark Label.

crate_universe/src/rendering.rs
illicitonion: I think you probably have better instincts on this than I do... WDYT?

It doesn't seem awful to keep all the keys here as `String`s; introducing either a generic generic, or a "bounded to just `String` or `Label`" generic all seem pretty reasonable.

crate_universe/3rdparty/crates/BUILD.ahash-0.7.6.bazel
illicitonion: Yeah, this is definitely not correct as-is, and the conditional we want isn't on what the label looks like, it should be conditional on the relationship between where the label was parsed and where the label is being serialized - there are really two different meanings here - there's a concept of a "relative" label and an "absolute" label.

We need to be aware of this at parse time. I can imagine two ways of handling this reasonably:
1. Change `Label` to be something like:
```rust
struct Label {
    package: PackageOrRelative,
    target: String,
}

enum PackageOrRelative {
    Package {
        repository: String,
        package: String,
    },
    Relative,
}
```
and at render time we output the `//` if `PackageOrRelative` is a `Package`, and omit it if it's a `Relative`. This is a little sketchy, as "label parse time" isn't really a time when we know whether we're going to serialize the label in the same package or a different one, but to simplify things we can make the assumption that the caller knows.

2. At parse time, pass in the package a label is being parsed relative to (and change `package` from an `Option<String>` to a `String` - every label _is_ in a package so this shouldn't _really_ be an `Option` (but I can believe that at existing use sites, e.g. in the `FromStr` implementation we don't _know_ the package we're parsing from). Then we would unconditionally serialize the `//` because we'd always know the package we were in. If we do this, there's an interesting question as to whether at serialisation time we want to accept "the package being serialised into" so that we can output a relative label if the package happens to match.

All of which goes back to the "how many label implementations" question we have. There are some really different use-cases for parsing/serialising a label - "This is just a typed string" lends itself to our current implementation (and leans towards option 1). "This is a type which can be used in different contexts to decide e.g. whether a package should be serialised or not based on where it's used" is more complex to use (as we need to understand parsing and serializing contexts) but may allow solving more general problems (and leans towards option 2). I'm generally fine with us having two types - one which is "basically a newtype around a string" and one which is aware of packages, but also happy for us to end up with one implementation as long as it's reasonably ergonomic...

I'd probably lean towards 1 here, because there isn't really a correct way for us to implement `FromStr`, `Serialize`, and `Deserialize` with option 2...

illicitonion (commented):

rickvanprim:
Based on my Slack discussion with @illicitonion, I'm not merging the Label definitions as evidently `//util/label/label.rs` has some external users, has the unresolved question about ownership, and I don't want to change its parsing to differentiate between "relative" and "absolute" labels without more context there.  For Starlark Label, an "absolute" label must contain `//` (with the exception of a pain repository label).  All the cases I'm aware of are now exhaustively spelled out in the parse function, with a test case per case.

UebelAndre (approved):

UebelAndre:
I'll defer to @illicitonion for merging 😄

crate_universe/src/utils/starlark/label.rs
```diff
@@ -11,60 +11,174 @@ use serde::{Deserialize, Serialize, Serializer};
...
+                target: repository.to_owned(),
+            }),
+
+            // Absolute (Package)
+            (None, true, Some(package), None) => Ok(Self {
+                package: AbosluteOrRelativePackage::Absolute {
+                    repository: None,
+                    package: Some(package.to_owned()),
+                },
+                target: package.to_owned(),
```
illicitonion, line 111: This should just be the basename of the package:

`foo/bar` expands to `@//foo/bar:bar` not `@//foo/bar:foo/bar`

(And it should be an error if the target isn't specified and the package is empty)

crate_universe/src/utils/starlark/label.rs
```diff
@@ -11,60 +11,174 @@ use serde::{Deserialize, Serialize, Serializer};
...
+                target: target.to_owned(),
+            }),
+
+            // Absolute (Repository + Package)
+            (Some(repository), true, Some(package), None) => Ok(Self {
+                package: AbosluteOrRelativePackage::Absolute {
+                    repository: Some(repository.to_owned()),
+                    package: Some(package.to_owned()),
+                },
+                target: package.to_owned(),
```
illicitonion, line 129: Basename

crate_universe/src/utils/starlark/label.rs
```diff
@@ -193,66 +309,129 @@ mod test {
...
+        Label::from_str("@repo:target").unwrap_err();
+    }
+
+    #[test]
+    fn invalid_relative_package_target() {
+        Label::from_str("package:target").unwrap_err();
+    }
+
+    #[test]
+    fn full_label_bzlmod() {
```
illicitonion, line 428: This is an interesting test-case... If you hadn't added the test case, I wouldn't mind (and happy to remove it), but I think that if we're going to care about this, we should preserve the `@@` at the start. So we may need to keep an `is_canonical` bool in the label, set it to whether the prefix of the label is `@@`, and prefix the stringification with `@@` (even if the repo is empty) if `is_canonical`.

(And if we're being pedantic, we also should technically preserve an `@` at the start of a label, so `@//foo:bar` should stringify as `@//foo:bar` not `//foo:bar`).

Maybe the real way of handling this is to make make `repository` be an
```rust
enum Repository {
    Canonical(String), // stringifies to `@@self.0` where `self.0` may be empty
    Explicit(String), // stringifies to `@self.0` where `self.0` may be empty
    Local, // stringifies to the empty string
}
```

illicitonion (commented):
LGTM, thanks! Just a few fun edge-cases to clean up :)

crate_universe/src/utils/starlark/label.rs
```diff
@@ -11,60 +11,174 @@ use serde::{Deserialize, Serialize, Serializer};
...
+    target: String,
+}
+
+#[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Clone)]
+enum AbosluteOrRelativePackage {
+    #[default]
+    Relative,
+    Absolute {
+        repository: Option<String>,
+        package: Option<String>,
```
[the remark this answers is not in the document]
rickvanprim: Works for me!

crate_universe/src/utils/starlark/label.rs
rickvanprim: Adding a test for invalid input of `":"`.

crate_universe/src/utils/starlark/label.rs
rickvanprim: Good catch!

crate_universe/src/utils/starlark/label.rs
rickvanprim: I like the `enum Repository` approach!  Also I didn't add this test, just moved it 😉

crate_universe/src/utils/starlark/label.rs
rickvanprim: I believe the only case where you have an empty package but no target would be `//` (which is covered under `invalid_empty()`).  Something like `@repo//` would be an empty package and empty target but will default to `repo`.

crate_universe/src/utils/starlark/label.rs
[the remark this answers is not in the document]
rickvanprim: This does add some small complexity, see in `rendering.rs` and `splicing.rs` where the match on `label.package()` goes from `Some(package)` to `Some(package) if !package.is_empty()`.

crate_universe/3rdparty/crates/BUILD.libc-0.2.149.bazel
```diff
@@ -29,99 +29,160 @@ rust_library(
             "WORKSPACE.bazel",
         ],
     ),
-    crate_features = [
-        "default",
-        "std",
-    ] + select({
+    crate_features = select({
```
rickvanprim, line 32: I'm not sure why the shared `crate_features` is gone, nor why `extra_traits` wasn't part of it in the first place.  This happens on `main` presently and is independent from this PR.  Maybe a good idea to add a check like `docs` has that runs the vendor command fails if anything is dirty.

illicitonion (approved):
Thanks!

PR #2370, opened 2023-12-28
Consolidate bzlmod example workspaces

UebelAndre:
This aims to de-duplicate code and tests as well as keep the root of `rules_rust` clean.

PR #2371, opened 2023-12-28
Ensure bzlmod example tests are run

UebelAndre:
A followup to https://github.com/bazelbuild/rules_rust/pull/2370#discussion_r1437744310

### DIFF
--- a/examples/bzlmod/hello_world/BUILD.bazel
+++ b/examples/bzlmod/hello_world/BUILD.bazel
@@ -16,7 +16,7 @@ sh_test(
     name = "hello_world_test",
     srcs = ["hello_world_test.sh"],
     args = [
-        "$(rootpath :hello_world)",
+        "$(rlocationpath :hello_world)",
     ],
     data = [
         ":hello_world",
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -630,13 +630,13 @@ tasks:
     name: bzlmod BCR presubmit
     platform: ubuntu2004
     working_directory: examples/bzlmod/hello_world
-    build_targets:
+    test_targets:
       - "//..."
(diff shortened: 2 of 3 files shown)


examples/bzlmod/hello_world/BUILD.bazel
```diff
@@ -16,7 +16,7 @@ sh_test(
     name = "hello_world_test",
     srcs = ["hello_world_test.sh"],
     args = [
-        "$(rootpath :hello_world)",
+        "$(rlocationpath :hello_world)",
```
illicitonion, line 19: This seems more complicated - are there benefits to using `rlocationpath` and needing to pull in a runfiles library, rather than just using `rootpath`?

UebelAndre: I find `rlocationpath` (use of runfiles) to generally be the most consistent and clear thing to do. I don't want to have to think about where the binary is running relative to some other file. Using the runfiles API feels like a better contract but in this case I'm indifferent. I can remove runfiles.

illicitonion: I don't have a super strong preference either way, I'm mostly just a fan of not having to copy+paste 10 lines of runfiles setup into shell scripts :)

illicitonion (approved):

examples/bzlmod/hello_world/BUILD.bazel
UebelAndre: I feel that. I'd rather this not have been a shell script but don't know if there's a strong reason for this. I think a `rust_test` would have been a better usage here. Just something to execute a rust binary.

illicitonion: I think a `rust_test` would be fine :)

PR #2385, opened 2024-01-02
Allow using cargo_bazel_bootstrap from module extensions

matts1:
This function isn't yet used. It will be used in #1910. I'm splitting it into its own PR because no-one seems to want to review #1910 because it's too big.


illicitonion (approved):

illicitonion:
I'll review #1910 today - sorry it's gone for so long, and thanks so much for piecing everything together and splitting it all up!

PR #2386, opened 2024-01-02
Don't re-write an identical lockfile

illicitonion:
https://github.com/bazelbuild/rules_rust/pull/1910 has a workaround for this because changing the timestamp may invalidate the repository rule.

Instead, just don't do the write if the contents is identical.


UebelAndre (approved):
Great thinking!

PR #2392, opened 2024-01-04
Add annotation support for transient crates.

matts1:


matts1:
@illicitonion I've done a complete rewrite of annotations according to your suggestions of inlining them into MODULE.bazel.

gferon:
I tested this PR on our current setup (still trying to move to `bzlmod`) and it fails on [`json.encode`](https://github.com/bazelbuild/rules_rust/pull/2392/files#diff-9cf43cf6220c1588d6370e635e7e573300d64f0f75a42474b759d20709cc1cd1L178) when supplying labels, like `build_script_tools = ["@com_google_protobuf//:protoc"]` with the following error: 

```
(11:07:32) ERROR: Traceback (most recent call last):
        File "/Users/gferon/.cache/bazel/8d38801da57cbc2fb5fa58e853c562bd/external/rules_rust~override/crate_universe/extension.bzl", line 182, column 58, in _crate_impl
                annotation = _crate_universe_crate.annotation(**{
        File "/Users/gferon/.cache/bazel/8d38801da57cbc2fb5fa58e853c562bd/external/rules_rust~override/crate_universe/private/crate.bzl", line 179, column 23, in _annotation
                return json.encode((
Error in encode: at tuple index 1: in struct field .build_script_tools: at list index 0: cannot encode Label as JSON
```

I'm not sure yet what the right fix would be though.

crate_universe/extension.bzl
```diff
@@ -160,16 +167,52 @@ def _crate_impl(module_ctx):
...
         local_repos = []
         for cfg in mod.tags.from_cargo:
             if cfg.name in local_repos:
                 fail("Defined two crate universes with the same name in the same MODULE.bazel file. Use the name tag to give them different names.")
             elif cfg.name in all_repos:
                 fail("Defined two crate universes with the same name in different MODULE.bazel files. Either give one a different name, or use use_extension(isolate=True)")
-            _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg)
+
+            annotations = {k: v for k, v in module_annotations.items()}
+            for crate, values in repo_specific_annotations.get(cfg.name, []).items():
```
gferon, line 206:
```suggestion
            for crate, values in repo_specific_annotations.get(cfg.name, {}).items():
```

I guess this should be a dict, when the key doesn't exist? Otherwise `.items()` fails.

crate_universe/extension.bzl
```diff
@@ -160,16 +167,52 @@ def _crate_impl(module_ctx):
...
+            # For the tag-based method, because it has type safety, we have to
+            # split it into two parameters.
+            if annotation_dict.pop("gen_all_binaries"):
+                annotation_dict["gen_binaries"] = True
+            annotation = _crate_universe_crate.annotation(**{
+                k: v
+                for k, v in annotation_dict.items()
+                # Tag classes can't take in None, but the function requires None
+                # instead of the empty values in many cases.
+                if v != "" and v != [] and v != {}
```
illicitonion, line 187: This doesn't cover the boolean arguments, which feels problematic for the booleans which we allow setting on the `from_cargo` level and overriding per crate with annotations (e.g. `gen_build_script`).

Is there an issue upstream about not being able to use `None` in `tag_class`es?

In the mean time, I guess we should probably ban booleans in our `tag_class`es and follow rules_go's style of having the `tag_class` take a string here which may be one of "on", "off", or "auto" (default "auto"), and converting "auto" to None here?

illicitonion:
> I tested this PR on our current setup (still trying to move to `bzlmod`) and it fails on [`json.encode`](https://github.com/bazelbuild/rules_rust/pull/2392/files#diff-9cf43cf6220c1588d6370e635e7e573300d64f0f75a42474b759d20709cc1cd1L178) when supplying labels like `build_script_tools = ["@com_google_protobuf//:protoc"]` with the following error:
> 
> ```
> (11:07:32) ERROR: Traceback (most recent call last):
>         File "/Users/gferon/.cache/bazel/8d38801da57cbc2fb5fa58e853c562bd/external/rules_rust~override/crate_universe/extension.bzl", line 182, column 58, in _crate_impl
>                 annotation = _crate_universe_crate.annotation(**{
>         File "/Users/gferon/.cache/bazel/8d38801da57cbc2fb5fa58e853c562bd/external/rules_rust~override/crate_universe/private/crate.bzl", line 179, column 23, in _annotation
>                 return json.encode((
> Error in encode: at tuple index 1: in struct field .build_script_tools: at list index 0: cannot encode Label as JSON
> ```
> 
> I'm not sure yet what the right fix would be though.
> 
> EDIT: I tried to "stringify" those labels, but the produced JSON can't be parsed anymore by cargo-bazel, so I'm definitely doing something wrong here.

https://github.com/bazelbuild/rules_rust/pull/2394 should fix this

illicitonion (commented):
This LGTM, thanks! Just one comment around booleans, then I think we're good to go.

crate_universe/extension.bzl
illicitonion: I couldn't find an issue so I filed https://github.com/bazelbuild/bazel/issues/20744 - could you add a reference to it in this comment?

matts1: This is only an issue for `gen_build_script`. There are only two boolean parameters to annotation - `disable_pipelining` and `gen_build_script`, and `disable_pipelining` defaults to `False` in the annotation rather than None.

matts1:
I've just rebased. There was also apparently some problems with using relative labels, as they were being evaluated relative to the wrong package. I've fixed it up now by turning them into strings.

illicitonion (approved):
Thanks so much!

PR #2396, opened 2024-01-04
Fix missing .wasm output for rust_wasm_bindgen

alexkirsz:
https://github.com/bazelbuild/rules_rust/pull/2284 removed the .wasm output file from the `DefaultInfo` provider, which makes it unavailable to other downstream rules.

This PR restores the behavior pre-0.32.0.

### DIFF
--- a/wasm_bindgen/private/wasm_bindgen.bzl
+++ b/wasm_bindgen/private/wasm_bindgen.bzl
@@ -96,7 +96,7 @@ def _rust_wasm_bindgen_impl(ctx):
 
     return [
         DefaultInfo(
-            files = depset(transitive = [info.js, info.ts]),
+            files = depset(transitive = [info.js, info.ts, depset([info.wasm])]),
         ),
         info,
     ]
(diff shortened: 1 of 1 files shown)


wasm_bindgen/private/wasm_bindgen.bzl
```diff
@@ -96,7 +96,7 @@ def _rust_wasm_bindgen_impl(ctx):
 
     return [
         DefaultInfo(
-            files = depset(transitive = [info.js, info.ts]),
+            files = depset(transitive = [info.js, info.ts, depset([info.wasm])]),
```
UebelAndre, line 99: Why not this?
```suggestion
            files = depset([info.wasm], transitive = [info.js, info.ts]),
```

UebelAndre (changes requested):
Thanks! Just one question

wasm_bindgen/private/wasm_bindgen.bzl
alexkirsz: I wasn't aware this was an option! Let's go with this instead.

UebelAndre (approved):
Thanks!

PR #2404, opened 2024-01-09
Fix the patch_tool annotation.

matts1:
It was documented incorrectly as a list, but the parameter is actually an Option<String> in the rust code (https://github.com/bazelbuild/rules_rust/blob/a1e9f9600cd22ecfdd08fc4a7572ccbdeba97395/crate_universe/src/metadata/metadata_annotation.rs#L152)


UebelAndre (approved):
Thanks!

PR #2407, opened 2024-01-09
Fix #2406: use codegen option `link-args` to forward `user_link_flags` in bindgen.

### LINKED ISSUE #2406: Incorrect use of user link flags in bindgen.
Hello,

In [`generate_cc_link_build_info`], the field [`user_link_flags`] of the [`LinkerInput`] is used to extend the command line of `rustc` through `--arg-file`:

https://github.com/bazelbuild/rules_rust/blob/012af3ba8fda6acb91a440f8c86b5a45d2c4e64e/bindgen/private/bindgen.bzl#L127

It doesn't work with platform-specific linker flags such as Apple's `-framework`.

Here is a PoC:

```starlark
# BUILD
load("@rules_rust//bindgen:defs.bzl", "rust_bindgen_library")                    
                                                                                 
cc_library(                                                                      
    name = "foo",                                                                
    srcs = ["foo.c"],                                                            
    hdrs = ["foo.h"],                                                            
    linkopts = ["-framework", "Security"],                                       
)                                                                                
                                                                                 
rust_bindgen_library(                                                            
    name = "foo_bindgen",                                                        
    cc_lib = ":foo",                                                             
    header = "foo.h",                                                            
)
```

Bazel fails with the following error:

```bash
error: Unrecognized option: 'f'
```

The content of the `.link_flags` is the following:

```
-lstatic=foo
-framework
Security
```

`-framework` is treated as `-f ramework`, leading to the `Unrecognized option` error.

Since [`user_link_flags`] is supposed to contain linker flags, I think you should instead use the codegen option [`link-args`]:

```
-C
link-args=-framework Security
```

(Note that `-lstatic=` arguments won't end up in `link-args` since they don't come from [`user_link_flags`])

[`generate_cc_link_build_info`]: https://github.com/bazelbuild/rules_rust/blob/012af3ba8fda6acb91a440f8c86b5a45d2c4e64e/bindgen/private/bindgen.bzl#L102
[`user_link_flags`]: https://bazel.build/rules/lib/builtins/LinkerInput#user_link_flags
[`LinkerInput`]: https://bazel.build/rules/lib/builtins/LinkerInput
[`link-args`]: https://doc.rust-lang.org/rustc/codegen-options/index.html#link-args

thb-sb:
This commit fixes issue #2406 by using the [`link-args`] codegen option of `rustc`:

```
-C
link-args=<user_link_flags>
```

where `<user_link_flags>` are the flags from the [`user_link_flags`] field of [`LinkerInput`].

[`user_link_flags`]: https://bazel.build/rules/lib/builtins/LinkerInput#user_link_flags
[`LinkerInput`]: https://bazel.build/rules/lib/builtins/LinkerInput
[`link-args`]: https://doc.rust-lang.org/rustc/codegen-options/index.html#link-args

### DIFF
--- a/bindgen/private/bindgen.bzl
+++ b/bindgen/private/bindgen.bzl
@@ -124,7 +124,9 @@ def _generate_cc_link_build_info(ctx, cc_lib):
                 linker_search_paths.append(lib.pic_static_library.dirname)
                 compile_data.append(lib.pic_static_library)
 
-        linker_flags.extend(linker_input.user_link_flags)
+        if linker_input.user_link_flags:
+            linker_flags.append("-C")
+            linker_flags.append("link-args={}".format(" ".join(linker_input.user_link_flags)))
 
     if not compile_data:
         fail("No static libraries found in {}".format(


UebelAndre (approved):
Thanks!

bindgen/private/bindgen.bzl
```diff
@@ -124,7 +124,9 @@ def _generate_cc_link_build_info(ctx, cc_lib):
                 linker_search_paths.append(lib.pic_static_library.dirname)
                 compile_data.append(lib.pic_static_library)
 
-        linker_flags.extend(linker_input.user_link_flags)
+        if linker_input.user_link_flags:
```
daivinhtran, line 127: @UebelAndre I think this will only work if we link using `rustc`. If we run the linker directly (i.e. enabling [experimental_use_cc_common_link](https://github.com/bazelbuild/rules_rust/blob/012af3ba8fda6acb91a440f8c86b5a45d2c4e64e/rust/settings/BUILD.bazel#L69), it won't work because the linker doesn't understand `-C link-args`. Only rustc understands `-C link-args=<args>` and passes `<args>` to the linker.

daivinhtran: Another subtle thing I notice in the example provided in https://github.com/bazelbuild/rules_rust/issues/2406 is that `rust_bindgen_library` shouldn't trigger any linking (neither calling `rustc` with `-C linker=<> -C link-args` or a separate linking action) because the macro (which creates rust_library -> rust_bindgen) only produces an rlib, not a shared library or binary.

Is 
```
# BUILD
load("@rules_rust//bindgen:defs.bzl", "rust_bindgen_library")                    
                                                                                 
cc_library(                                                                      
    name = "foo",                                                                
    srcs = ["foo.c"],                                                            
    hdrs = ["foo.h"],                                                            
    linkopts = ["-framework", "Security"],                                       
)                                                                                
                                                                                 
rust_bindgen_library(                                                            
    name = "foo_bindgen",                                                        
    cc_lib = ":foo",                                                             
    header = "foo.h",                                                            
)
```
all you need to reproduce the error with the link flags? Or is there another downstream target like `rust_binary` that causes a linking?

daivinhtran: IIUC, the right fix isn't simply
1. If use `cc_common.link`, don't use `-C link-args=<link flags>`
2. Else, use `-C link-args`
at the bindgen rule level.

IMO the right fix is to keep link flags without `-C link-args=` as before. At the downstream (shared library or binary) targets which run linking, we then decide whether to add `-C link-args` or not based on the setting of [experimental_use_cc_common_link](https://github.com/bazelbuild/rules_rust/blob/012af3ba8fda6acb91a440f8c86b5a45d2c4e64e/rust/settings/BUILD.bazel#L69).

daivinhtran: More context: In https://github.com/bazelbuild/rules_rust/blob/90e4505c19e8276311cd5ad7d9d894aa659105dd/rust/private/rustc.bzl#L998-L1013. 

The rules' implementation already ignore linking when building `rlib` and `lib`. `rust_bindgen_library` produces `rlib` which shouldn't involve any link flags at all.

daivinhtran: I filed https://github.com/bazelbuild/rules_rust/issues/2413 to triage the issue.

Issue #2260, opened 2023-11-13
Support rustfmt toolchains for bzlmod


stephencpope:
When running under bazel 6.4.0 using `--enable_blzmod`, the `rustfmt` toolchain doesn't get wired up as it should. A simple no-op bazel module fails to support the `@rules_rust//:rustfmt` target like this:
```
$ bazel run --enable_bzlmod @rules_rust//:rustfmt
Starting local Bazel server and connecting to it...
DEBUG: @@rules_rust~override//:MODULE.bazel:6:6: WARNING: The rules_rust Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
ERROR: .../external/rules_rust~override/rust/toolchain/BUILD.bazel:40:26: While resolving toolchains for target @rules_rust~override//rust/toolchain:current_rustfmt_toolchain: No matching toolchains found for types @rules_rust~override//rust/rustfmt:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@rules_rust~override//rust/rustfmt:toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
ERROR: Analysis of target '@rules_rust~override//:rustfmt' failed; build aborted: 
INFO: Elapsed time: 13.555s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (42 packages loaded, 168 targets configured)
    currently loading: @rules_rust~override//tools/runfiles
ERROR: Build failed. Not running target
```
when using this `MODULE.bazel`:
```
bazel_dep(name = "rules_rust", version = "0.30.0")
git_override(
    module_name = "rules_rust",
    remote = "https://github.com/bazelbuild/rules_rust",
    commit = "54bef51d8e3458ec60ea2b6a27e15b62502c673d",
)

rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")

rust.toolchain()

use_repo(rust, "rust_toolchains")

register_toolchains(
    "@rust_toolchains//:all",
    dev_dependency = True,
)
```
The problem can also be seen in the output of
```
$ bazel query --enable_bzlmod @rust_toolchains//:\*
@rust_toolchains//:BUILD.bazel
@rust_toolchains//:rust_analyzer_1.73.0
@rust_toolchains//:rust_darwin_aarch64__aarch64-apple-darwin__nightly
[...]

PR #2410, opened 2024-01-10
Fix rustfmt toolchains when consuming rules_rust with bzlmod.

jondo2010:
Fixes https://github.com/bazelbuild/rules_rust/issues/2260


illicitonion (commented):
This looks great, thanks!

Would you be able to add a test which shows this works? https://github.com/bazelbuild/rules_rust/pull/2395/files has an example for when we applied a similar fix to the rust analyzer toolchain.

jondo2010:
Done!

PR #2414, opened 2024-01-10
Implement support for dylib linkage

daivinhtran:
This PR implements dylib linkage against the standard library behind a feature flag `--@rules_rust//rust/settings:experimental_use_dylib_linkage`.

The main part of this feature is [here](https://github.com/bazelbuild/rules_rust/pull/2414/files#diff-2a806da393e47c07ffe67c78ace69eb488b4ac44b029a46d8237b8e2a05637beR258) where we skip exporting static rust stdlibs and export only `libstd.so` instead.

This feature is useful when the subset of libstd being statically linked to downstream shared libraries and binaries is **larger** than the entire dylib version of libstd. The following diagram is the high level of what dylib linkage is trying to achieve.

Untitled Diagram drawio

Running the feature against `android_binary` yields a size reduction on the shared library produced by `android_binary` because it doesn't statically link the rust stdlibs anymore.

```
> bazel build //:android_app --config=android_x86_64 
> unzip -l bazel-bin/android_app.apk
Archive:  bazel-bin/android_app.apk
  Length      Date    Time    Name
---------  ---------- -----   ----
  1381968  2010-01-01 00:00   lib/x86_64/libandroid_app.so <--- static link with rust stdlibs
---------                     -------
  1390294                     9 files
```

```
> bazel build //:android_app --config=android_x86_64 --config=dylib_linkage
> unzip -l bazel-bin/android_app.apk
Archive:  bazel-bin/android_app.apk
  Length      Date    Time    Name
---------  ---------- -----   ----
     8080  2010-01-01 00:00   lib/x86_64/libandroid_app.so <--- reduced size because of dynamic linking
 13055776  2010-01-01 00:00   lib/x86_64/libstd-8d416d49cf02ecea.so
---------                     -------
 13072400                     10 files
 ```

Here, the benefit comes when there are enough shared libraries statically linking against the rust stdlibs. "Enough" here means that the total up size of those libraries being more than just the entire `libstd.so`.

TODO: I'm leaving this PR without unit tests until I get some feedback or suggestions on my approach.

### CODE REVIEW

File: examples/android/BUILD.bazel [hunk 1]
```diff
@@ -18,6 +18,35 @@ cc_library(
...
 
+rust_library(
+    name = "rust_lib2",
+    srcs = ["demo2.rs"],
+    edition = "2018",
+)
+
+cc_library(
+    name = "jni_shim2",
+    srcs = ["android_link_hack.c"],  # Required because of https://github.com/bazelbuild/rules_rust/issues/1271
```
keith, line 29: this should only be in a single cc_library in the whole dependency tree

daivinhtran: It seems like I need to change the function name too? 

```
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild-android-ST-b08757e62318/bin/_objs/jni_shim2/android_link_hack2.pic.o: multiple definition of 'call_link_hack'
/usr/bin/ld.gold: bazel-out/k8-fastbuild-android-ST-b08757e62318/bin/_objs/jni_shim/android_link_hack.pic.o: previous definition here
```

daivinhtran: So the build isn't happy if the two rust files (used in two different `rust_library`s) have any duplicate function signatures.

```
/usr/bin/ld.gold: error: bazel-out/k8-fastbuild-android-ST-b08757e62318/bin/librust_lib2-855537059.a(rust_lib2-855537059.rust_lib2.a90be88302ad5382-cgu.0.rcgu.o): multiple definition of 'print_something_from_rust'
/usr/bin/ld.gold: bazel-out/k8-fastbuild-android-ST-b08757e62318/bin/librust_lib-443237737.a(rust_lib-443237737.rust_lib.6f4c6aab038b4cfe-cgu.0.rcgu.o): previous definition here
```

This seems like an unintended restriction to be resolved.

daivinhtran: If there should only be one `cc_library` in the whole dependency tree, doesn't that mean there is only one `android_library` in the whole tree as well?

keith: i think the ideal is that every rust_library (any number) depends on the single cc_library containing the shim. if that doesn't work because of the linked `alwayslink` issue, then i think you'd have to change the names of everything in the "link hack" (and ofc rust symbols can't overlap just like normal)

rust/settings/BUILD.bazel
```diff
@@ -99,6 +99,19 @@ incompatible_flag(
     issue = "https://github.com/bazelbuild/rules_rust/issues/2324",
 )
 
+# A flag to control whether to link libstd dynamically.
+bool_flag(
+    name = "experimental_use_dylib_linkage",
```
illicitonion, line 104: I'm not quite sure about the full impact of this change - is it _just_ on the stdlib (in which case I'd expect stdlib to be in the flag name), or is it for all dependencies?

illicitonion (commented):
Generally LGTM, just one naming question (and some tests would be great :))

rust/settings/BUILD.bazel
daivinhtran: I agree. It's only dynamically linking libstd. I'll fix the naming and add tests.

Per https://doc.rust-lang.org/reference/linkage.html, it's worth nothing

> For most situations, having all libraries available as a dylib is recommended if dynamically linking. 

So ideally, I think it makes sense to switch `rust_library` to build `dylib` in this mode. But that's a future :) .

rust/toolchain.bzl
```diff
@@ -75,6 +76,14 @@ def _rust_stdlib_filegroup_impl(ctx):
                 print("File partitioned: {}".format(f.basename))
             fail("rust_toolchain couldn't properly partition rlibs in rust_std. Partitioned {} out of {} files. This is probably a bug in the rule implementation.".format(partitioned_files_len, len(dot_a_files)))
 
+    std_dylib = None
+
+    for file in rust_std:
+        # find libstd-*.so
+        if file.basename.startswith("libstd-") and file.basename.endswith(".so"):
```
illicitonion, line 83: It feels like we should probably extract out this check to a function somewhere common with `is_std_dylib` in rust/private/utils.bzl, and maybe add a `or file.basename.endswith(".dll")` case for Windows?

test/link_std_dylib/link_std_dylib_test.bzl
```diff
@@ -0,0 +1,116 @@
...
+                    return True
+    return False
+
+def _export_libstd_dylib_in_cc_info(target):
+    linker_inputs = target[CcInfo].linking_context.linker_inputs
+    for linker_input in linker_inputs.to_list():
+        for library in linker_input.libraries:
+            if hasattr(library, "dynamic_library") and library.dynamic_library != None:
+                basename = library.dynamic_library.basename
+                if basename.startswith("libstd") and basename.endswith(".so"):
```
illicitonion, line 61: This should probably also use the same extracted helper?

~
```diff
@@ -0,0 +1,13 @@
+Fix buildifier error
```
illicitonion, line 1: Remove?

illicitonion (approved):
I'm afraid nothing jumps out about why the tests are failing, but happy to spend a bit of time debugging if no one else has any insights!

rust/toolchain.bzl
daivinhtran: Yes. I think this is why the tests are failing.

PR #2415, opened 2024-01-11
Fix #2413: don't use rustc codegen option `link-args` if `cc_common.link` is used.

### LINKED ISSUE #2413: Incorrect linker flags when using `cc_common.link`
See https://github.com/bazelbuild/rules_rust/pull/2407#discussion_r1446339662 for more context. I believe https://github.com/bazelbuild/rules_rust/pull/2407 breaks any user that uses `cc_common.link` for linking.

Filing a bug so we can triage and prioritize fixing it.

thb-sb:
This commit is an attempt to fix #2413 (see https://github.com/bazelbuild/rules_rust/pull/2407#discussion_r1446339662 for more context).

When `cc_common.link` is used (using [`experimental_use_cc_common_link`]), we must not use the rustc codegen option [`link-args`] since it is unknown to the `cc_common.link` executable. Instead, we must forward values of [`user_link_flags`] directly to the `cc_common.link` command line.

[`experimental_use_cc_common_link`]: https://github.com/bazelbuild/rules_rust/blob/012af3ba8fda6acb91a440f8c86b5a45d2c4e64e/rust/settings/BUILD.bazel#L69
[`link-args`]: https://doc.rust-lang.org/rustc/codegen-options/index.html#link-args
[`user_link_flags`]: https://bazel.build/rules/lib/builtins/LinkerInput#user_link_flags

### DIFF
--- a/bindgen/private/bindgen.bzl
+++ b/bindgen/private/bindgen.bzl
@@ -113,6 +113,8 @@ def _generate_cc_link_build_info(ctx, cc_lib):
     linker_flags = []
     linker_search_paths = []
 
+    rust_toolchain = ctx.toolchains[str(Label("//rust:toolchain_type"))]
+
     for linker_input in cc_lib[CcInfo].linking_context.linker_inputs.to_list():
         for lib in linker_input.libraries:
             if lib.static_library:
@@ -124,7 +126,9 @@ def _generate_cc_link_build_info(ctx, cc_lib):
                 linker_search_paths.append(lib.pic_static_library.dirname)
                 compile_data.append(lib.pic_static_library)
 
-        if linker_input.user_link_flags:
+        if rust_toolchain._experimental_use_cc_common_link:
+            linker_flags.extend(linker_input.user_link_flags)
+        elif linker_input.user_link_flags:
             linker_flags.append("-C")
             linker_flags.append("link-args={}".format(" ".join(linker_input.user_link_flags)))
 


UebelAndre:
@daivinhtran does this work for your use cases?

I’m also wondering if this is something we can add a test for to catch future regressions.

daivinhtran:
@UebelAndre @thb-sb I'm afraid https://github.com/bazelbuild/rules_rust/pull/2407 was not the best fix for https://github.com/bazelbuild/rules_rust/issues/2406.

Let's look at the command line from the aquery output of a `rust_bindgen_library`

```
$ cd examples
$ bazel aquery //bindgen:simple_bindgen

...
  Command Line: (exec bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/external/rules_rust/util/process_wrapper/process_wrapper \
    --arg-file \
    bazel-out/k8-fastbuild/bin/bindgen/simple_bindgen__bindgen.link_search_paths \
    --arg-file \
    bazel-out/k8-fastbuild/bin/bindgen/simple_bindgen__bindgen.link_flags \
    --subst \
    'pwd=${pwd}' \
    -- \
    bazel-out/k8-fastbuild/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/rustc \
    bazel-out/k8-fastbuild/bin/bindgen/simple_bindgen__bindgen.rs \
    '--crate-name=simple_bindgen' \
    '--crate-type=rlib' \
    '--error-format=human' \
    '--codegen=metadata=-3215433401' \
    '--codegen=extra-filename=-3215433401' \
    '--out-dir=bazel-out/k8-fastbuild/bin/bindgen' \
    '--codegen=opt-level=0' \
    '--codegen=debuginfo=0' \
    '--remap-path-prefix=${pwd}=' \
    '--emit=dep-info,link' \
    '--color=always' \
    '--target=x86_64-unknown-linux-gnu' \
    -L \
    bazel-out/k8-fastbuild/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/lib/rustlib/x86_64-unknown-linux-gnu/lib \
    '--edition=2018' \
    '--sysroot=bazel-out/k8-fastbuild/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain')
...
```

Per https://github.com/bazelbuild/rules_rust/blob/caa3959740d0f4133ef9ceabf8fa27ed809fb87e/util/process_wrapper/options.rs#L78-L81, the content of `bazel-out/k8-fastbuild/bin/bindgen/simple_bindgen__bindgen.link_flag` is passed directly to `rustc`. **However**, `rustc` doesn't link when producing an `rlib` (aka `'--crate-type=rlib'`).

Link flags should only be needed when the `rustc` command sets `--codegen=linker=...` and produces a binary or shared library.

daivinhtran:
I'll put up a PR and ask for your feedback!

daivinhtran:
@UebelAndre

So I'd expect for 

```
binary -> rust_library -> bindgen -> cc
```

dependency graph

daivinhtran:
@UebelAndre

So I expected that for 
```
rust_binary -> rust_library -> rust_bindgen -> cc_library
```
dependency graph, the `CcInfo` from the `cc_library` would get propagated to `rust_binary` so that the `rust_library` doesn't need the link artifacts from `cc_library`. But after https://github.com/bazelbuild/rules_rust/pull/2361/files, that's not the case anymore.

Also my statement that `rustc`, when producing `rlib`, does not link seems to be incorrect. Do you happen to know why `rustc` command for producing `rlib` doesn't set `--codegen=linker=/usr/bin/gcc` like when it produces a `bin`?

daivinhtran:
Ok so I think there is a convoluted set of things that seems suspicious here.

`<rust_library_bindgen>__bindgen.link_flags` includes `-lstatic=<cc-dep>` which is passed directly to `rustc` by `process_wrapper`. Additionally, `link_flags` file also includes the `linkopts` set on the cc dependencies.

At head currently, a BUILD file like
```
cc_library(
    name = "simple",
    srcs = ["simple.cc"],
    hdrs = ["simple.h"],
    linkopts = [
        "-framework",
    ],
    visibility = ["//bindgen:__pkg__"],
)

rust_bindgen_library(
    name = "simple_bindgen",
    bindgen_flags = [
        "--allowlist-function=simple_.*",
        "--allowlist-var=SIMPLE_.*",
    ],
    cc_lib = "//bindgen/simple",
    header = "//bindgen/simple:simple.h",
    leak_symbols = False,
)
```
will produces a `link_flags` file as
```
-lstatic=simple
-C
link-args=-framework
```

If we attempt to run the action command for `simple_bindgen` directly with `--print link-args`, we get nothing. This is because rustc does not invoke any linker internally. So I suspect that `-framework` doesn't get applied. I believe so because for
```
rust_binary(
    name = "simple_example",
    srcs = ["main.rs"],
    deps = [":simple_bindgen"],
)
```
passing `--print link-args` to the command yield a linking command back with what we pass to `-C link-args`.

@thb-sb can you confirm whether `-framework` actually gets used?

Ideally, we split `link_flags` into two files

`rustc_flags`
```
-lstatic=simple
```
`linker_flags`
```
-framework
```

Then at upstream (`rust_binary` for example), we can decide whether to use `-C link-args` or pass `linker_flags` directly to the linker based on `experimental_use_cc_common_link`. The logic `experimental_use_cc_common_link` is already implemented.

daivinhtran:
Closing the PR since this was fixed in https://github.com/bazelbuild/rules_rust/pull/2422.

PR #2417, opened 2024-01-11
Update android example to use Starlark version of android_ndk_repository

daivinhtran:
The native version of `android_ndk_repository` rule no longer work with the newer ndk versions.

When I set `ANDROID_NDK_HOME` to `/usr/local/vinhdaitran/Android/Sdk/ndk/26.1.10909125`, the rule expects a different structure from the Android NDK path.
```
ERROR: /usr/local/vinhdaitran/github/rules_rust/examples/android/WORKSPACE.bazel:67:23: fetching android_ndk_repository rule //external:androidndk: java.io.IOException:
Expected directory at /usr/local/vinhdaitran/Android/Sdk/ndk/26.1.10909125/platforms but it is not a directory or it does not exist.
Unable to read the Android NDK at /usr/local/vinhdaitran/Android/Sdk/ndk/26.1.10909125, the path may be invalid. Is the path in android_ndk_repository() or ANDROID_NDK_HOME set correctly?
If the path is correct, the contents in the Android NDK directory may have been modified.
```

Using the Starlark version of the rule, as recommended in https://bazel.build/reference/be/android#android_ndk_repository, fixes the issue.

cc: @keith

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -582,6 +582,10 @@ tasks:
     build_targets:
       - "//:android_app"
     environment:
+      # This ndk version matches with rules_android_ndk repo's CI
+      # https://github.com/bazelbuild/rules_android_ndk/blob/877c68ef34c9f3353028bf490d269230c1990483/.bazelci/presubmit.yml#L37
+      # The ndk is installed by this script 
+      # https://github.com/bazelbuild/continuous-integration/blob/ba56013373821feadd9f2eaa6b81eb19528795f0/macos/mac-android.sh
       ANDROID_NDK_HOME: /opt/android-ndk-r25b
   android_examples_macos:
     name: Android Examples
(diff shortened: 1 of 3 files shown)


keith:
fine with me, but CI might only have old versions

daivinhtran:
> fine with me, but CI might only have old versions

@keith do you happen to know the version CI is using?

keith:
https://github.com/bazelbuild/continuous-integration/blob/ba56013373821feadd9f2eaa6b81eb19528795f0/buildkite/android.sh
https://github.com/bazelbuild/continuous-integration/blob/ba56013373821feadd9f2eaa6b81eb19528795f0/macos/mac-android.sh

katre:
You need to set the `ANDROID_NDK_HOME` env correctly in the config file. See [`rules_android_ndk`](https://github.com/bazelbuild/rules_android_ndk/blob/877c68ef34c9f3353028bf490d269230c1990483/.bazelci/presubmit.yml#L37) as an example.

katre (approved):

daivinhtran:
Thanks @katre for the reference!

.bazelci/presubmit.yml
```diff
@@ -581,6 +581,8 @@ tasks:
       - "--android_platforms=//:arm64-v8a"
     build_targets:
       - "//:android_app"
+    environment:
+      ANDROID_NDK_HOME: /opt/android-ndk-r25b
```
UebelAndre, line 585: Can you leave a comment on how this value was found in case it needs to be updated in the future?

UebelAndre (approved):
One small request but otherwise looks good to me!

.bazelci/presubmit.yml
daivinhtran: Good catch. Done.

PR #2419, opened 2024-01-12
Make cross-compiling with bzlmod possible, added example

jondo2010:
Before it wasn't possible to add your desired cross-compilation targets to `extra_target_triples` if they also happened to be in `DEFAULT_TOOLCHAIN_TRIPLES`.


examples/bzlmod/cross_compile/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+load("@rules_rust//rust:defs.bzl", "rust_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "hello_world_aarch64",
```
illicitonion, line 6: Can you throw in a `sh_test` or `rust_test` which takes this as a data dependency, runs `file` on it, and asserts on the output to show that it was _actually_ compiled for the correct target platform?

illicitonion (commented):
This looks great, thanks! Just one question about testing

examples/bzlmod/cross_compile/BUILD.bazel
jondo2010: Done!

illicitonion (approved):
Thanks!

jondo2010:
@illicitonion any chance of landing this?

illicitonion (approved):

PR #2424, opened 2024-01-14
Prepare rust rules for Starlark CcToolchainInfo.

buildbreaker2021:
https://github.com/bazelbuild/rules_rust/issues/2425

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -671,8 +671,10 @@ def collect_inputs(
     """
     linker_script = getattr(file, "linker_script") if hasattr(file, "linker_script") else None
 
-    # TODO(gnish): rules_rust is not coupled with Bazel release. Remove conditional and change to
-    # _linker_files once CcToolchainInfo changes are visible to Bazel.
+    # TODO: As of writing this comment Bazel used Java CcToolchainInfo.
+    # However there is ongoing work to rewrite provider in Starlark.
+    # rules_rust is not coupled with Bazel release. Remove conditional and change to
+    # _linker_files once Starlark CcToolchainInfo is visible to Bazel. 
     if hasattr(cc_toolchain, "_linker_files"):
         linker_depset = cc_toolchain._linker_files
     else:
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -674,7 +674,8 @@ def collect_inputs(
     # TODO: As of writing this comment Bazel used Java CcToolchainInfo.
     # However there is ongoing work to rewrite provider in Starlark.
     # rules_rust is not coupled with Bazel release. Remove conditional and change to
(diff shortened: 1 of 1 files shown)


rust/private/rustc.bzl
```diff
@@ -671,7 +671,12 @@ def collect_inputs(
     """
     linker_script = getattr(file, "linker_script") if hasattr(file, "linker_script") else None
 
-    linker_depset = cc_toolchain.linker_files()
+    # TODO(gnish): rules_rust is not coupled with Bazel release. Remove conditional and change to
```
UebelAndre, line 674: Can you instead reference a public github issue or something? The Google `TODO(name)` comments are not that useful for external folks.

UebelAndre (changes requested):

rust/private/rustc.bzl
buildbreaker2021: Unfortunately there is no public github issue - I will remove TODO. I think the comment is useful even without an issue it clearly explains why this change is necessary and when it should be removed. I can extend it if you'd like to.

buildbreaker2021: I removed username, and updated the comment to better describe the issue.

buildbreaker2021:
Friendly ping

rust/private/rustc.bzl
```diff
@@ -671,7 +671,14 @@ def collect_inputs(
     """
     linker_script = getattr(file, "linker_script") if hasattr(file, "linker_script") else None
 
-    linker_depset = cc_toolchain.linker_files()
+    # TODO: As of writing this comment Bazel used Java CcToolchainInfo.
```
daivinhtran, line 674: Do you mind filing a Github issue and put it here so we can triage later? It's hard to keep track of all the TODOs.

buildbreaker2021: Done.

rust/private/rustc.bzl
buildbreaker2021: Done.

UebelAndre (approved):
The underscore naming convention makes me nervous that we’re setting ourselves up to need to do some crazy migration and would rather not start relying on deprecated/private interfaces in the first place. This change seems fine to me but am more concerned about the conversion process for rules_cc. I also think there should be an issue made there that tracks this effort

buildbreaker2021:
> The underscore naming convention makes me nervous that we’re setting ourselves up to need to do some crazy migration and would rather not start relying on deprecated/private interfaces in the first place.

About private part - API used by `rules_rust` was private in the first place. `rules_rust` is allowlisted so it can access Bazel's internal private API.

About migration part - I *think* the most likely scenario here is that we make that parameter public eventually. Most likely the migration which will need to be done would be to migrate from `_linker_files` to `linker_files`.

Issue #2418, opened 2024-01-12
Clippy Builds fail on 0.36.2


SvenKratzEasel:
Building of clippy targets fail with the following message : `error: Option 'sysroot' given more than once`

UebelAndre:
Can you give an example workspace for this!

daivinhtran:
Yeah I'm surprised CI isn't failing. Several folks have [reported this](https://github.com/bazelbuild/rules_rust/pull/2277#issuecomment-1868029606) and it seems like something I should look into fixing.

@SvenKratzEasel I would appreciate if you have some steps to reproduce the bug.

SvenKratzEasel:
Here are the rust-related items from the WORKSPACE file:

```
http_archive(
    name = "rules_rust",
    sha256 = "9d04e658878d23f4b00163a72da3db03ddb451273eb347df7d7c50838d698f49",
    urls = ["https://github.com/bazelbuild/rules_rust/releases/download/0.26.0/rules_rust-v0.26.0.tar.gz"],
)

load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")

rules_rust_dependencies()

rust_register_toolchains(
    edition = "2021",
)

load("@rules_rust//crate_universe:repositories.bzl", "crate_universe_dependencies")

crate_universe_dependencies(bootstrap = True)

load("@rules_rust//crate_universe:defs.bzl", "crates_repository")

crates_repository(
    name = "crate_index",
    cargo_config = "//:.cargo/config.toml",
    cargo_lockfile = "//:Cargo.lock",
    # See load satement for `cargo_bazel_bootstrap`.
    generator = "@cargo_bazel_bootstrap//:cargo-bazel",
    lockfile = "//:cargo-bazel-lock.json",
    manifests = [
       ...
    ],
)

load(
    "@crate_index//:defs.bzl",
    cargo_workspace_crate_repositories = "crate_repositories",
)

cargo_workspace_crate_repositories()
```

illicitonion:
There's a full minimal-ish repro at https://github.com/dmeister/rules_rules_clippy_repo from https://bazelbuild.slack.com/archives/CSV56UT0F/p1704923287157729

A workaround is to set `--@rules_rust//rust/settings:experimental_toolchain_generated_sysroot=False`.

I suspect the fix may be just to not add the flag in clippy specifically.

daivinhtran:
Thanks! I'll take a look.

[...]

Issue #2298, opened 2023-12-01
`experimental_toolchain_generated_sysroot`


UebelAndre:
`experimental_toolchain_generated_sysroot` tracks a feature which explicitly passes `--sysroot` to `Rustc` actions. The sysroot path is defined by the toolchain.

Relates to https://github.com/bazelbuild/rules_rust/issues/2039

UebelAndre:
This flag was added in https://github.com/bazelbuild/rules_rust/pull/2223 and flipped in https://github.com/bazelbuild/rules_rust/pull/2277

PR #2428, opened 2024-01-17
Added `incompatible_no_rustc_sysroot_env`

UebelAndre:
To remove `SYSROOT` from Rustc actions

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1059,7 +1059,8 @@ def construct_arguments(
         ))
 
     # Ensure the sysroot is set for the target platform
-    env["SYSROOT"] = toolchain.sysroot
+    if not toolchain._incompatible_no_rustc_sysroot_env:
+        env["SYSROOT"] = toolchain.sysroot
     if toolchain._experimental_toolchain_generated_sysroot:
         rustc_flags.add(toolchain.sysroot, format = "--sysroot=%s")
 
--- a/rust/settings/BUILD.bazel
+++ b/rust/settings/BUILD.bazel
@@ -98,3 +98,10 @@ incompatible_flag(
     build_setting_default = True,
     issue = "https://github.com/bazelbuild/rules_rust/issues/2324",
 )
+
+# A flag to remove the SYSROOT environment variable from `Rustc` actions.
+incompatible_flag(
+    name = "incompatible_no_rustc_sysroot_env",
+    build_setting_default = True,
+    issue = "https://github.com/bazelbuild/rules_rust/issues/2429",
+)
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -643,6 +643,7 @@ def _rust_toolchain_impl(ctx):
         _experimental_use_global_allocator = experimental_use_global_allocator,
         _experimental_use_coverage_metadata_files = ctx.attr._experimental_use_coverage_metadata_files[BuildSettingInfo].value,
         _experimental_toolchain_generated_sysroot = ctx.attr._experimental_toolchain_generated_sysroot[IncompatibleFlagInfo].enabled,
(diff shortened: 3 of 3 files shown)


UebelAndre:
Ping @illicitonion

illicitonion (approved):
I'm not sure we want to be uncoditionally removing this from all actions; I can believe clippy and rustc want different envs here... But also, no tests are failing, so hopefully if this breaks something someone can report a failing test case we can commit.

UebelAndre:
It’s my understanding that this has no impact on rustc and potentially negative impacts with clippy. Its existence feels low value and the intent satisfied by the —sysroot flag. Is that not correct?

daivinhtran:
Hi @UebelAndre I've ran into issues with `--sysroot` flag during bootstrapping after checking in https://github.com/bazelbuild/rules_rust/pull/2277. I was going to propose a fix in this space but haven't had time to to get to it?

Could you hold on to merging this PR? I'll add some notes by the end of today. I'm just making sure the fixes don't step on each other.

If you still prefer to merge the PR, could you add some tests that will fail without this PR? Just to protect from further regression.

cc: @krasimirgg

UebelAndre:
> If you still prefer to merge the PR, could you add some tests that will fail without this PR? Just to protect from further regression.

The issue is I don't know exactly what causes a failure with this environment variable set. I would have expected failures to occur when enabling `--sysroot` (aka `experimental_toolchain_generated_sysroot` - https://github.com/bazelbuild/rules_rust/issues/2298) but alas, nothing failed. This PR adds a feature flag with the variable enabled to expedite the integration. I don't see an issue with merging but if you want to take on addressing https://github.com/bazelbuild/rules_rust/issues/2418 then that sounds good to me and I can hold.

daivinhtran:
SGTM. I'll address https://github.com/bazelbuild/rules_rust/issues/2418 now. Hope to send out some PR by end of today.

UebelAndre:
> It’s my understanding that this has no impact on rustc and potentially negative impacts with clippy. Its existence feels low value and the intent satisfied by the —sysroot flag. Is that not correct?

@daivinhtran would you be able to confirm or deny this? I still think this would be a good change to make if it's true.

daivinhtran:
> > It’s my understanding that this has no impact on rustc and potentially negative impacts with clippy. Its existence feels low value and the intent satisfied by the —sysroot flag. Is that not correct?
> 
> @daivinhtran would you be able to confirm or deny this? I still think this would be a good change to make if it's true.

If we're sure that we always pass `--sysroot`, then `clippy-driver` will recognize and prefer that over `SYSROOT`. So removing `SYSROOT` now does feel safe to me.

One thing I'm cautious about is that `SYSROOT` might be use somewhere else out of our sight. For example, we didn't know `SYSROOT` is used by `clippy-driver` but now we know.

If the intent was to fix https://github.com/bazelbuild/rules_rust/issues/2418, the bug is fixed in `clippy-driver` now. If the intent is now to remove redundancy, that seems fine to me too.

UebelAndre:
> One thing I'm cautious about is that `SYSROOT` might be use somewhere else out of our sight.

This is why this change in behavior is introduced as a flag so users can report if this causes a breakage. As of now I know of no known use for it and only 1 issue caused by it. So again, I think we should disable it and see what happens and if a user reports an issue we can add a regression test and maintain the behavior after assessing the use case.

To be clear though, the intent is to drop an unnecessary flag. This change allows us to determine this through user feedback.

daivinhtran:
> > One thing I'm cautious about is that `SYSROOT` might be use somewhere else out of our sight.
> 
> This is why this change in behavior is introduced as a flag so users can report if this causes a breakage. As of now I know of no known use for it and only 1 issue caused by it. So again, I think we should disable it and see what happens and if a user reports an issue we can add a regression test and maintain the behavior after assessing the use case.
> 
> To be clear though, the intent is to drop an unnecessary flag. This change allows us to determine this through user feedback.

SGTM!

daivinhtran (approved):

PR #2431, opened 2024-01-20
Add extra_rustc_flags_for_crate_types.

granaghan:
This allows extra rustc flags that will only apply to library targets to be set by the toolchain. Cross language LTO needs '-C linker-plugin-lto' on libraries, how passing this flag to rust_binary generated bloat. Adding this attribute resolves this issue.

Change-Id: Iba725fab1b1941e9586ff97cd71ec3bc1dfc1523


rust/private/rustc.bzl
```diff
@@ -1063,6 +1063,9 @@ def construct_arguments(
     else:
         rustc_flags.add_all(toolchain.extra_rustc_flags)
 
+    if crate_info.type in ["rlib", "lib"] and not is_exec_configuration(ctx):
+        rustc_flags.add_all(toolchain.extra_rustlib_rustc_flags)
```
daivinhtran, line 1067: I think the attribute naming is too generic whereas its responsibility is too specific (i.e. only applying to rlib or lib and not in exec configuration).

1. If we want to apply the flags to non-exec configuration, we can just set the flags to non-exec toolchains. Exec and non-exec are separate toolchains.

2. "rustlib" means more than just rlib and lib. It can include dylib (not implemented yet), cdylib as well.

I feel like the rust_tochain rule isn't the right level for this kind of flag.

Normally for scenarios like this, we could create macros to wrap the rust_library and rust_static_library rules with the custom flags instead? Would this be a solution?

daivinhtran (changes requested):
I left a comment. I think macros would be the solution for your use cases. Let me know if macros won't work.

rust/private/rustc.bzl
matts1: I agree with these concerns, but I think with a refactor it could be useful. How about instead of `extra_rustlib_rustc_flags = ["--foo"]`, we add `extra_rustc_flags_for_crate_types = {"rlib": ["--foo"], "lib": ["--foo"]}` (and remove the `is_exec_configuration` check). I believe that would be generic enough to be worth adding to the toolchain.

daivinhtran (commented):
This LGTM. Do you mind adding some tests?

granaghan:
I added a simple test. Is there a way to get the CrateInfo provider in the test so I can test with multiple crate types? I poked around a bit and didn't  manage to, but I'm pretty new to working with rule implementations.

matts1:
To get any provider, you just need to do `target[Provider]` on a target that defines a CrateInfo (cc_binary and cc_library). However, looking at the test, it seems quite confusing, but it only seems to do assertions on the toolchain itself, so doesn't have access to cc_binary and cc_library targets, so there's no CrateInfo to get.

granaghan:
From the CI:
```
bazel-out/darwin_x86_64-opt-exec-ST-a84ee23024d0/bin/util/process_wrapper/process_wrapper.sh: line 18: /private/var/tmp/_bazel_buildkite/e4140ff87fa9a41eb6ddd3bb20d07fdc/sandbox/darwin-sandbox/22/execroot/rules_rust/bazel-out/darwin_x86_64-opt-exec-ST-a84ee23024d0/bin/test/toolchain/rust_extra_flags_toolchain/bin/mock_rustc.exe: Undefined error: 0
```

Did I do something to make it try and run the fake binaries? Is it the `rust_binary` I added?

daivinhtran:
@granaghan I merged the main branch into your branch to rerun CI. Now it's failing with different error.

Can you pull the updated branch, run `./docs/update_docs.sh` locally, and then `git push` again?

granaghan:
It looks like the coverage check is still tying to use the mock compiler. For some reason it's building for OS `none` on both Ubuntu and Mac. It passes for me locally though.
```
    CARGO_CFG_TARGET_OS=none \
```

granaghan:
Okay, I managed to repro it after gathering all the additions flags. It seems that `//rust/settings:experimental_use_coverage_metadata_files` is causing this, but I'm not sure why yet.

granaghan:
It seems this only happens if I declare a `rust_binary` in the tests.  I switched `rust_shared_library` and it passes.

granaghan:
Any other changes here or is this good to go?

daivinhtran (approved):
LGTM. Can you also file an issue about the failure with rust_binary? Thanks!

granaghan:
Added https://github.com/bazelbuild/rules_rust/issues/2583. I don't have write access, so you'll need to merge. Thanks!

PR #2440, opened 2024-01-24
Make runfiles library bzlmod compatible

purkhusid:
## What?
Adds handling of repository mapping in the runfiles library when bzlmod is enabled.

## Why?
My original motivation is to get the rust analyzer project generator working and it uses the runfiles library.


purkhusid:
@matts1 @illicitonion @UebelAndre The regex approach I'm using to get the source repo is kind of fragile. I'm not that deep into Rust land, do you have any suggestions that I could use to figure out the source repo?

tools/runfiles/runfiles.rs
```diff
@@ -53,23 +55,70 @@ enum Mode {
...
+                    mode: r.mode,
+                    source_repository: r.source_repository,
+                    repo_mapping: HashMap::new(),
+                }
+            }
+        })
+    }
+
+    fn get_source_repository() -> String {
+        let caller = std::env::current_exe().expect("failed to get current executable path");
```
fmeum, line 103: This looks problematic: What if a `rust_binary` in repo `foo` depends on a `rust_library` in repo `bar`? Lookup should be relative to `bar` in that case, but this would result in `foo`.

Maybe this is a use case for `std::env!` together with rules support that adds the repository name of the current target to the compilation environment when the target has a direct dependency on the runfiles library? This is very similar to what I did for `cc_*` rules.

matts1: I think the correct approach would be to use a macro instead of a function. We need the compilation context of the caller.

Rlocation!(r, "foo") would expand to r.rlocation(foo, env!("workspace"))

Note that that environment variable is evaluated at compile time, not runtime. Not sure if it's the same idea that @fmeum had

Sorry for formatting, on phone

fmeum: @matts1 Yes, that's what I had in mind, a macro that itself uses `std::env!` to evaluate to the repository name of the target containing the `rlocation` call. Compile-time evaluation seems necessary as runtime evaluation wouldn't be able to distinguish the calls based on source file.

purkhusid: Thanks for the input. I'll take a look if I can change the library so that it uses a macro. Are we fine with a breaking change? 

Just to sum up, the approach you both are suggesting is that I add the canonical workspace name as an env to the rustc compilation action and then use a macro that uses `std:env!` to embed the workspace name into the compiled binary that is using the macro.

That leaves the question if there is a good way to get the canonical workspace name within a rule? AFAIK `ctx.workspace_name` always  returns `_main` when bzlmod is enabled. It's possible to use `native.repo_name` but then all rules would have to be wrapped in a macro.

fmeum: The term "workspace name" has always been confusing and isn't really relevant with Bzlmod anymore as you noticed. Instead, you want the canonical repository name of the current target, which you can obtain via `ctx.label.repo_name` (available as of 6.5.0) or `ctx.label.workspace_name` (a terribly misnamed alias of `repo_name`, but available in all versions of Bazel).

purkhusid: Alrighty, I'll use `ctx.label.workspace_name` for increased compatibility.

tools/runfiles/runfiles.rs
```diff
@@ -43,6 +43,8 @@ const RUNFILES_DIR_ENV_VAR: &str = "RUNFILES_DIR";
 const MANIFEST_FILE_ENV_VAR: &str = "RUNFILES_MANIFEST_FILE";
 const MANIFEST_ONLY_ENV_VAR: &str = "RUNFILES_MANIFEST_ONLY";
 const TEST_SRCDIR_ENV_VAR: &str = "TEST_SRCDIR";
+const EXTERNAL_GENERATED_FILE_REGEX: &str = "bazel-out[/][^/]+/bin/external/([^/]+)/";
```
dzbarsky, line 46: wdyt of avoiding the regex dep here? I think `str.split('/')` should get you what you want?

dzbarsky:
@purkhusid I'd love to get this landed so we can switch to bzlmod. Anything I can do to help?

purkhusid:
@dzbarsky This PR has to be reworked from scratch after some good input in the comments and changed to use a macro instead of the regexp approach.

I don't have capacity to work on this at the moment or in the near future so if someone wants to take a stab at this then that would be ideal.

dzbarsky:
No worries. I'll pick up the torch in https://github.com/bazelbuild/rules_rust/pull/2566

dzbarsky:
@purkhusid want to abandon this PR? I landed the fix

PR #2444, opened 2024-01-24
Reproduce bug in clippy-driver

daivinhtran:
Steps to reproduce https://github.com/bazelbuild/rules_rust/issues/2418

1. https://github.com/dmeister/rules_rules_clippy_repo and replace `http_archive(name = "rules_rust")` to `local_repository` pointing to a local repo of `rules_rust`
2. In `rules_rust` repo, apply the patch https://github.com/bazelbuild/rules_rust/pull/2444.diff
3. Flip the counter back and forth between 22 and 23 to reproduce the bug

I believe when there are too many arguments to `clippy-driver`, it breaks.

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -93,6 +93,7 @@ def _clippy_aspect_impl(target, ctx):
         aliases = crate_info.aliases,
         # Clippy doesn't need to invoke transitive linking, therefore doesn't need linkstamps.
         are_linkstamps_supported = False,
+        from_clippy = True,
     )
 
     compile_inputs, out_dir, build_env_files, build_flags_files, linkstamp_outs, ambiguous_libs = collect_inputs(
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -195,7 +195,8 @@ def collect_deps(
         deps,
         proc_macro_deps,
         aliases,
-        are_linkstamps_supported = False):
+        are_linkstamps_supported = False,
+        from_clippy = False):
     """Walks through dependencies and collects the transitive dependencies.
 
     Args:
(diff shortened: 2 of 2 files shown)


daivinhtran:
Intereresting `//crate_universe:cargo_bazel_bin` has similar setup but `clippy-driver` doesn't fail on it.

daivinhtran:
> So the complaint about duplicate `--sysroot` is a red herring?

I believe so. Have you seen https://github.com/bazelbuild/rules_rust/issues/2447?

`process_wrapper` parses (and processes?) the args and pass `"@bazel-out/k8-fastbuild/bin/hello_world/hello_world.clippy.ok-0.params.expanded", "-Lnative=/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/external/crate_index__ring-0.16.20/ring_build_script.out_dir", "-Lnative=/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/external/crate_index__ring-0.17.7/ring_build_script.out_dir"` to `clippy-driver`. 

`clippy-driver` then sets `--syroot` based on `SYSROOT`.

I believe, at runtime within the execution of the `process_wrapper`, `clippy-driver` sees another `--sysroot` in the `.expanded` file and fails. That's why either removing "--sysroot" (which causes it to be removed from `.expanded`) or removing `SYSROOT` (which causes `--sysroot` not to be created by `clippy-driver` before calling `rustc`) fixed; but it's not the true fix. The true fix should be fixing `process_wrapper` to not wrap `--sysroot` into `@bazel-out/k8-fastbuild/bin/hello_world/hello_world.clippy.ok-0.params.expanded` or make `@bazel-out/k8-fastbuild/bin/hello_world/hello_world.clippy.ok-0.params.expanded` more deterministic.

PR #2445, opened 2024-01-25
Use a path instead of a string for emit in rustc command-line arguments.

matts1:
This can improve the cache hit rate, as it will work with path mapping.

Also generate the emit path unconditionally, as this works better with the `binary_ext` parameter better (as it guarantees that you won't get an error where the file names don't match).


krasimirgg:
Looks good!

>This can improve the cache hit rate, as it will work with path mapping.

Could you go into more detail what's path mapping?

matts1:
Suppose you enter a transition from foo=False to foo=True, and foo doesn't affect a rust binary you're building.

Bazel will still have to rebuild, because when entering a transition it changes the output path from `bazel-out/k8-fastbuild/...` to `bazel-out/<hash>`. However, with path mapping, you can tell bazel that these are equivalent, and that one of these paths maps to another in the action key.

krasimirgg (approved):
Thank you! Looks good!

PR #2451, opened 2024-02-01
clippy: use --cap-lints=warn; apply clippy_flags when capture_output=True

goffrie:
Instead of using `-Wclippy::all` to override deny-by-default clippy lints into warnings, use `--cap-lints=warn`.
This allows users to provide whatever lint flags they like and the compiler will still succeed.


rust/private/clippy.bzl
```diff
@@ -141,12 +141,11 @@ def _clippy_aspect_impl(target, ctx):
         args.process_wrapper_flags.add("--stderr-file", clippy_out)
 
         if clippy_flags:
-            fail("""Combining @rules_rust//:clippy_flags with @rules_rust//:capture_clippy_output=true is currently not supported.
-See https://github.com/bazelbuild/rules_rust/pull/1264#discussion_r853241339 for more detail.""")
+            args.rustc_flags.add_all(clippy_flags)
```
krasimirgg, line 144: Could you ping that discussion thread to mention that this PR lifts this restriction?

krasimirgg (commented):
This is cool! Just want to make sure folks on the old discussion thread are aware of this PR.

buchgr:
Thanks @goffrie. IIUC this allows users to set additional lints as warnings and anything more strict will be ignored. The need to specify additional lints as warnings makes sense to me. However, it'd expect the rules to error if a an unsupported deny lint is specified, rather than silently ignore it. What do you think? Would it be possible to add this failure behavior?

goffrie:
I don't see this as a problem. `--deny` flags will be treated as though they were `--warn`, but specifying those flags as deny can still be useful. For example, a workspace may have clippy targets that are sometimes built directly by the developer _without_ capture_clippy_output, in which case the --deny would do the right thing; and sometimes depended upon by other rules with a capture_clippy_output transition, in which case --deny would be downgraded.

krasimirgg:
Thank you! Chatted with @buchgr this looks good to merge, with a follow-up issue https://github.com/bazelbuild/rules_rust/issues/2548 about the potential user confusion around user-defined --deny flags.

PR #2454, opened 2024-02-01
Fix bzlmod examples on windows

cameron-martin:
On windows (and some other platforms), the file extension of cargo, rustc, etc have an extension. The module extension for loading crates did not take this into account, causing it to error on windows.

Additionally, when using bzlmod to build vendored crates, the runfiles tree would exceed the 260 char windows path limit. To mitigate this, I have shortened the internal_deps module extension to just `i` and changed the build script suffix to `_bs` from `_build_script`. This makes the path names below the 260 char limit.

This makes the bzlmod CI run on windows, to avoid regressing this. Currently gen_rust_project does not run on windows for other reasons, although we do build this.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -667,6 +667,16 @@ tasks:
       - "@rules_rust//tools/rust_analyzer:gen_rust_project"
     test_targets:
       - "//..."
+  windows_bzlmod_bcr:
+    name: bzlmod BCR presubmit
+    platform: windows
+    working_directory: examples/bzlmod/hello_world
+    run_targets:
+      - "//third-party:vendor"
+    build_targets:
+      - "@rules_rust//tools/rust_analyzer:gen_rust_project"
+    test_targets:
+      - "//..."
 
 buildifier:
   version: latest
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -14,3 +14,20 @@ bcr_test_module:
         - "@rules_rust//tools/rust_analyzer:gen_rust_project"
       test_targets:
         - "//..."
+# Windows is run separately because currently gen_rust_project doesn't run on windows, although it does build
+bcr_test_module_windows:
+  module_path: ""
+  matrix:
+    bazel: ["6.x", "7.x"]
+  tasks:
+    run_tests:
+      working_directory: examples/bzlmod/hello_world
+      name: "Run test module"
+      platform: windows
+      bazel: ${{ bazel }}
+      run_targets:
+        - "//third-party:vendor"
+      build_targets:
+        - "@rules_rust//tools/rust_analyzer:gen_rust_project"
+      test_targets:
+        - "//..."
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -36,7 +36,7 @@ bazel_dep(
     repo_name = "com_google_protobuf",
 )
 
-internal_deps = use_extension("//rust/private:extensions.bzl", "internal_deps")
+internal_deps = use_extension("//rust/private:extensions.bzl", "i")
 use_repo(
     internal_deps,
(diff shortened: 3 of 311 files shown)


cameron-martin:
It looks like running `gen_rust_project` doesn't work on windows, so I've omitted that from the windows CI for now.

cameron-martin:
It seems that we can't build `gen_rust_project` on windows. Do we not already have that building on CI on windows?

cameron-martin:
So the issue is that you can't run cargo build scripts when using bzlmod when rules_rust is an external repository. When trying to launch the build script, it fails with

```
Unable to start binary: Os { code: 267, kind: NotADirectory, message: "The directory name is invalid." }                                                                                                                                                                               
```

Maybe its something to do with weird characters in the path name. Here's the difference between the paths of the executables when running from rules_rust vs an external repository:

```
C:\users\camer\_bazel_camer\5ot5qtif\execroot\rules_rust\bazel-out/x64_windows-opt-exec-ST-13d3ddad9198/bin/external/rrra__io-lifetimes-1.0.11/io-lifetimes_build_script_.exe
C:\users\camer\_bazel_camer\5cp6anq6\execroot\_main\bazel-out/x64_windows-opt-exec-ST-13d3ddad9198/bin/external/rules_rust~override~internal_deps~rrra__io-lifetimes-1.0.11/io-lifetimes_build_script_.exe
```

The first example is from rules_rust, the second is when running from the bzlmod example. Possibly its the tilde that's causing issues?

cameron-martin:
Okay I know what is happening here. The issue is actually with the current_dir argument [here](https://github.com/bazelbuild/rules_rust/blob/bbe2cf42b9788dcae8e9e02349436a0e2aa52571/cargo/cargo_build_script_runner/bin.rs#L85). This goes beyond the 260 char file name limit on windows when running from an external repository.

Not sure what the best solution to this is. Can we just expect people to [enable long paths](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell), and do so on the CI machines too?

cameron-martin:
Now there's a failure when linking. This directory is too long:

```
 "/LIBPATH:C:\\b\\5cp6anq6\\execroot\\_main\\bazel-out/x64_windows-opt-exec-ST-13d3ddad9198/bin/external/rules_rust~override~i~rrra__windows_x86_64_msvc-0.48.0/windows_x86_64_msvc_build_script_.exe.runfiles/rules_rust~override~i~rrra__windows_x86_64_msvc-0.48.0/lib"
```

Any ideas on how to cut this down? Does the build script name actually need to be `windows_x86_64_msvc_build_script_.exe`?

dtolnay:
That comment is from https://github.com/abrisco/cargo-bazel/pull/37.

scentini:
> Not sure what the best solution to this is. Can we just expect people to [enable long paths](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell), and do so on the CI machines too

I believe we already enable long paths on the CI machines, however this is not sufficient because a lot of the Windows tools do not work with them (MSVC's `link.exe` included, I believe). Maybe @UebelAndre has more context on the comment above; I wonder however if we can instead omit the prefix leading up to `build_script`, if there is a single build script in the package?

cameron-martin:
@scentini It looks like the crate name is read from the executable name to set the `CARGO_CRATE_NAME` env var. See:

https://github.com/bazelbuild/rules_rust/blob/53daac71c0755680c29f4c55ac5238cc063e8b31/cargo/private/cargo_build_script.bzl#L409-L420

https://github.com/bazelbuild/rules_rust/blob/53daac71c0755680c29f4c55ac5238cc063e8b31/cargo/private/cargo_build_script.bzl#L141

I think changing this suffix to something shorter should work though.

PR #2456, opened 2024-02-02
[docs/cargo.md] Add missing comma

NilsIrl:


illicitonion (approved):
Thanks for the fix!

For the future, a lot of our markdown is actually generated, so you need to run `./docs/update_docs.sh` to make sure both the original and generated versions are up to date (and this shows up in CI) - I ran the script in your branch to fix up this PR :)

PR #2457, opened 2024-02-02
Add argument `include_coverage` to toggle if `InstrumentedFilesInfo` is created

kiwiwarmnfuzzy:
If `rustc_compile_action` is invoked within an aspect of a rule, without this change, an `InstrumentedFilesInfo` would be created, which conflicts with the `InstrumentedFilesInfo` with the underlying target, causing a bazel action conflict. Thus, we add an option (default to `True`) so that aspect can specify this option when invoking `rustc_compile_action`, to avoid creating `InstrumentedFilesInfo`.


krasimirgg (approved):
Looks good, thank you!

PR #2459, opened 2024-02-03
Add script to revendor all crates

cameron-martin:
It seems that vendored crates commonly get out of sync. Re-vendoring seems to currently be a quite manual process, and finding everything to re-vendor is a hassle.

This adds a bash script, `vendor.sh`, that finds all crate vendor rules in all workspaces in this repository and runs them. This significantly improves the workflow for updating these.

This also adds `.bazelrc` files to disable bzlmod to non-bzlmod workspaces, since otherwise bazel 7 create a MODULE.bazel and MODULE.bazel.lock when re-vendoring.

This also re-vendors all crates (in a separate commit).


vendor.sh
UebelAndre, line 1: Can we put this file in `util` or something? I'd like to keep the root of the repo clean. Maybe `crate_universe` would be better with an accompanied change to some developer docs?

UebelAndre (changes requested):
So sorry I didn't see this sooner, This would be fantastic! Just one request

vendor.sh
cameron-martin: I've put it in util since there is already a similar script, `repin_all.sh` in there. Where are the developer docs? I couldn't see any except CONTRIBUTING.md, that only seems to contain information about the CLA.

UebelAndre: Maybe https://github.com/bazelbuild/rules_rust/blob/0.39.0/crate_universe/DEVELOPMENT.md ? Maybe that should be replaced by a `CONTRIBUTING.md` though. Docs are kinda weak at the moment.

cameron-martin: Okay done.

UebelAndre (approved):
Thanks!

PR #2461, opened 2024-02-05
Add support for `all_crate_deps` when using bzlmod

AmeliasCode:
This adds support for the `all_crate_deps` functionality in repositories created via bzlmod. An [earlier update](https://github.com/bazelbuild/rules_rust/pull/1910/files) that added support for 3rd party crates with bzlmod intentionally left this functionality out. 

Some key things that I think should be reviewed: 

- Is the method of getting `crate_label_template` in `extensions.bzl` the best approach? It requires a lot of hard-coding, and it seems like it might be possible to simplify but I'm not familiar enough with the code base to know how. 
- Is the `generate_config_file` method from `crates_vendor.bzl` the right method to use for generating the cargo-bazel config? It seems like there are two places with similar logic (here and in `crates_repositories.bzl` (see https://github.com/bazelbuild/rules_rust/pull/1832)


AmeliasCode:
Sorry for the buildifier failures, wasn't sure exactly what command the buildkite workflow was running. Appears to be passing now! For anyone stuck in the future, try `buildifier -v --warnings all --lint fix`

AmeliasCode:
@illicitonion Judging by similar pull requests this seems like it might be in your arena? Could you please take a look when you have time? Thanks!

PR #2467, opened 2024-02-08
fix rustdoc: omit link flags for rustdoc

bgeron:
In our project, rust_doc fails because it is generating `-l` arguments to
rustdoc, and rustdoc does not support `-l` arguments.

In my understanding, these arguments are only relevant for calling the
linker at some point, and it won't be called for rustdoc anyway.

With this PR, our project can build with rust_doc.


krasimirgg:
There are some CI failures that suggest that for some actions in rustdoc mode at least, the linking step or something in the `if ("link" in emit and crate_info.type not in ["rlib", "lib"]):` section is necessary.
It seems that the rustdoc param should already account for not invoking linking since https://github.com/bazelbuild/rules_rust/commit/9442aed8ca9e07e024f9daaf06030c968623c9c1. Naively, could you adapt the code to filter out the necessary `-l` flags depending on the existing `rustdoc` param instead?

Also could you add an example integration test about the failure you're seeing.

bgeron:
Thanks for looking at this! I adapted the code correspondingly, to omit `-l` flags in a more fine-grained way. I thin the point of that flag is rustdoc will still compile binaries for running doctests. I don't understand it 100% though, but this patch is much more minimal, the tests are now passing, and our repo is building.

Most of the new patch is just threading variables through. But we now omit `-l` flags in three circumstances that before all failed for different reasons:

- rustdoc for bin, linking to `cc_library`. Before, we only had rustdoc for a lib that links to a `cc_library`. 
- Rust build script emitting `-llibrary`.
- Nonempty `cc_toolchain.dynamic_runtime_lib` or `cc_toolchain.static_runtime_lib`.

I made regression tests for the first two. But I didn't know how to for the third one; common platforms seem to have empty `cc_toolchain.dynamic_runtime_lib` and `cc_toolchain.static_runtime_lib`. I think it's okay to not cover this unlikely case with a test for now.

krasimirgg (approved):
Looks good! Thank you!

PR #2473, opened 2024-02-12
Release 0.39.0

freeformstu:


krasimirgg:
Could you add details on which changes are we picking up in this version.

UebelAndre:
> Could you add details on which changes are we picking up in this version.

That’s autogenerated in the release notes. Or maybe I’m confused by what you’re looking for.

krasimirgg:
> > Could you add details on which changes are we picking up in this version.
> 
> That’s autogenerated in the release notes. Or maybe I’m confused by what you’re looking for.

I just assumed there are some new API changes incompatible with 0.38? Or it's more of a: release at a regular time intervals?

UebelAndre:
rules_rust releases are intended to be regular interval things

UebelAndre (approved):

PR #2480, opened 2024-02-16
Introduce @rules_testing for Starlark tests

daivinhtran:
This is a follow-up PR of https://github.com/bazelbuild/rules_rust/pull/2422.

I'm scoping it to a separate PR to introduce [@rules_testing](https://github.com/bazelbuild/rules_testing) (a new Starlark testing framework). The framework removes a lot of boilerplate code (unncessarily) required when writing analysis tests.


UebelAndre (approved):
Thanks! Seems totally fine to me! I think if we want to introduce this it'd be good to do so in a change that converts more so we can get a better sense of the practical improvement the new framework offers to the repo.

daivinhtran:
>  it'd be good to do so in a change that converts more

That's a good idea. I'll hold on to this PR until I convert a few more tests to surface the differences more clearly.

daivinhtran:
Since I need to use `rules_testing` to write tests for https://github.com/bazelbuild/rules_rust/pull/2414, I'll go ahead and merge the PR. I'll follow-up with converting a few more tests to `rules_testing` before we introduce it more broadly.

Issue #2483, opened 2024-02-16
crate_universe broken with bzlmod


jondo2010:
This regressed in #2461 

At  https://github.com/AmeliasCode/rules_rust/blob/13ba928eddf944ff6f8636f53f5f869e8b01894d/crate_universe/extension.bzl#L49 breaks most uses of `rules_rust` in bzlmod since it hard codes the repository name to `~override`. This works in CI since the `rules_rust` module in the tests are all overridden.

UebelAndre:
cc @AmeliasCode @illicitonion

AmeliasCode:
@UebelAndre @jondo2010 Any ideas on how best to get the canonical repository name for the generated repositories? If I understand correctly, the actual repositories are created here: https://github.com/AmeliasCode/rules_rust/blob/13ba928eddf944ff6f8636f53f5f869e8b01894d/crate_universe/extension.bzl#L131 but the BUILD files are generated before-hand.

AmeliasCode:
I'm thinking something like this (names will be improved later): 

```
    lbl = Label(":target")
    idx = str(lbl).find("//")
    rep = str(lbl)[:idx]
    rendering_config = json.decode(render_config(
        regen_command = "Run 'cargo update [--workspace]'",
        crate_label_template = "{}~crate~{{repository}}__{{name}}-{{version}}//:{{target}}".format(rep),
    ))
```

but this seems kind of hacky? It works, but I'm still not sure where `~crate~` is coming from, or if this would have other unintended consequences.

AmeliasCode:
>I'm still not sure where ~crate~ is coming from

Answering my own question here, it's the name of the extension.

cameron-martin:
Why does the canonical name need to be used here, rather than utilising the repo mappings?

AmeliasCode:
@cameron-martin Can you expand on that a bit? I'm fairly new to Bazel

cameron-martin:
Bzlmod uses a feature called "repo mappings" to map non-canonical repository names to canonical repository names. I'm not sure that the form of the canonical repository names is designed to be a stable public API, so it would be better to specify non-canonical repository names and let bazel do this mapping.

[...]

PR #2485, opened 2024-02-17
Fix `crate_universe` when using bzlmod

AmeliasCode:
Addresses the issue raised here: https://github.com/bazelbuild/rules_rust/issues/2483

Now, we programmatically determine the canonical repository name prefix. This assumes that the repositories created by `crate_universe` will all inherit the same prefix as `rules_rust`.

If there is a more ideal approach to determining the canonical repository name, please suggest! I was unable to find anything else. 

Additionally, I'm not sure how to code a test case against this, since we have to override the `rules_rust` location in order to test it?


cameron-martin:
Btw I don't think this is the correct fix. See https://github.com/bazelbuild/rules_rust/issues/2483#issuecomment-1951390416

AmeliasCode:
Withdrawing, as @cameron-martin has raised a better fix: https://github.com/bazelbuild/rules_rust/pull/2491

PR #2491, opened 2024-02-18
Use non-canonical repo name in hub repo

### LINKED ISSUE #2483: crate_universe broken with bzlmod

cameron-martin:
#2461 changed the rendering of the hub repo to use canonical repository names, but incorrectly. Doing so wasn't necessary - instead we just need to do the resolution of the non-canonical label from the perspective of the hub repo rather than the consumer.

Fixes #2483

### DIFF
--- a/crate_universe/extension.bzl
+++ b/crate_universe/extension.bzl
@@ -46,7 +46,6 @@ def _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg, annotations):
 
     rendering_config = json.decode(render_config(
         regen_command = "Run 'cargo update [--workspace]'",
-        crate_label_template = "@@rules_rust~override~crate~{repository}__{name}-{version}//:{target}",
     ))
     config_file = tag_path.get_child("config.json")
     module_ctx.file(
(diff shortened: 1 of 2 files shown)


UebelAndre (approved):
Thanks! Do you know if there's something we can do to regression test this? To ensure bzlmod works in pre-commit testing?

cameron-martin:
I couldn't immediately think of a way, since this bug only arises due to the difference in canonical repository names for overridden and non overridden repositories. Open to ideas however.

cameron-martin:
Then you can only test the last the latest released version, so can't catch issues before you release.

I guess you could set up a local registry, but that sounds like a lot of hassle.

PR #2505, opened 2024-02-20
Add MODULE.bazel.lock to gitignore

keith:
This way this file doesn't need to be manually managed, but if someone wants to build rules_rust offline they can have a local version


illicitonion (approved):
Thanks!

Issue #2495, opened 2024-02-19
Error: macos_application has to declare 'j2objc' as a required fragment in order to access it



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3666#018dbf86-8436-4052-8905-72823801bfc5

 Platform: MacOS

 Logs: 
```
ERROR: /Users/buildkite/builds/bk-imacpro-17/bazel-org-repo-root/rules_rust/examples/ios/BUILD.bazel:36:18: in macos_application rule //:macos_app:
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_buildkite/332ed97c1e1b0f15afaaf0f1012ec808/external/build_bazel_rules_apple/apple/internal/macos_rules.bzl", line 228, column 65, in _macos_application_impl
		link_result = linking_support.register_binary_linking_action(
	File "/private/var/tmp/_bazel_buildkite/332ed97c1e1b0f15afaaf0f1012ec808/external/build_bazel_rules_apple/apple/internal/linking_support.bzl", line 242, column 58, in _register_binary_linking_action
		linking_outputs = apple_common.link_multi_arch_binary(
	File "/virtual_builtins_bzl/common/objc/linking_support.bzl", line 265, column 86, in _link_multi_arch_binary
	File "/virtual_builtins_bzl/common/objc/compilation_support.bzl", line 807, column 62, in _register_configuration_specific_link_actions
	File "/virtual_builtins_bzl/common/objc/compilation_support.bzl", line 579, column 26, in _register_j2objc_dead_code_removal_actions
Error: macos_application has to declare 'j2objc' as a required fragment in order to access it. Please update the 'fragments' argument of the rule definition (for example: fragments = ["j2objc"])
(05:02:27) ERROR: /Users/buildkite/builds/bk-imacpro-17/bazel-org-repo-root/rules_rust/examples/ios/BUILD.bazel:36:18: Analysis of target '//:macos_app' failed
(05:02:27) ERROR: Analysis of target '//:macos_app' failed; build aborted
``` 
 Culprit: 

 Steps:
  ``` 
 git clone https://github.com/bazelbuild/rules_rust
git reset 547f6a0666024c4b14b794598ecac3ce56a5e950  --hard
export USE_BAZEL_VERSION=38c1d5d27061481b1a688abb37c6f1dd8b288c26
bazel build  //... 
 ``` 
 CC Greenteam @fweikert

[...]

PR #2506, opened 2024-02-20
Update apple rules

keith:
Fixes https://github.com/bazelbuild/rules_rust/issues/2495


daivinhtran (approved):

keith:
it looks like it got removed from the queue because it took too long(?) not sure If it's going to merge or not

PR #2507, opened 2024-02-21
Updated label crate to understand canonical vs apparent repositories

UebelAndre:
Doing my best to adhere to https://bazel.build/concepts/labels. This change updates the label utility to account for modern changes to Bazel labels (e.g. from bzlmod).

### DIFF
--- a/util/label/label.rs
+++ b/util/label/label.rs
@@ -9,49 +9,144 @@ use label_error::LabelError;
 /// TODO: validate . and .. in target name
 /// TODO: validate used characters in target name
 pub fn analyze(input: &'_ str) -> Result<Label<'_>> {
-    let label = input;
-    let (input, repository_name) = consume_repository_name(input, label)?;
-    let (input, package_name) = consume_package_name(input, label)?;
-    let name = consume_name(input, label)?;
-    let name = match (package_name, name) {
-        (None, None) => {
-            return Err(LabelError(err(
-                label,
-                "labels must have a package and/or a name.",
-            )))
+    Label::analyze(input)
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum Repository<'s> {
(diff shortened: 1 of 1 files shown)


UebelAndre:
@hlopko if you have time, would love to know if this still works for your use cases. I did my best to keep the spirit of the original `Label` impl you added in https://github.com/bazelbuild/rules_rust/pull/724

util/label/label.rs
```diff
@@ -9,49 +9,144 @@ use label_error::LabelError;
...
-    pub repository_name: Option<&'s str>,
-    pub package_name: Option<&'s str>,
-    pub name: &'s str,
+pub enum Label<'s> {
+    Relative {
+        target_name: &'s str,
+    },
+    Absolute {
+        repository: Option<Repository<'s>>,
+        package_name: Option<&'s str>,
```
illicitonion, line 39: An absolute label always has a package name (which may happen to be empty), so I don't think this should be an `Option`?

util/label/label.rs
```diff
@@ -9,49 +9,144 @@ use label_error::LabelError;
...
+    }
+
+    pub fn repo_name(&self) -> Option<&'s str> {
+        match self {
+            Label::Absolute { repository, .. } => repository.as_ref().map(|repo| repo.repo_name()),
+            Label::Relative { .. } => None,
         }
     }
 
     pub fn packages(&self) -> Vec<&'s str> {
```
illicitonion, line 129: I'm not sure what this function exists for - it seems unused, and its semantics seem generally unhelpful? Should we delete it?

illicitonion (approved):

util/label/label.rs
UebelAndre: Good point. I can update this

util/label/label.rs
UebelAndre: It’s mostly around because it was already there. But yeah, I think the semantics around “packages” is not knowable by just the label. If a package is a directory containing a BUILD file then you can have a directory with another BUILD file multiple directories deep. I think it’d be better to have users accept the semantics applied here within their own repo. Simple enough to split the package in user code

PR #2514, opened 2024-02-23
Re-vendor all crates

cameron-martin:
It looks like the vendored crates got out of sync again. This updates them all, using the new script `//crate_universe/tools:vendor`.


illicitonion (approved):
Thanks!

PR #2515, opened 2024-02-23
Add public interface to lockfile interospection

### LINKED ISSUE #1725: crate_universe: exposing the bazel lockfile format
Hello. I am making a [gazelle](https://github.com/bazelbuild/bazel-gazelle) plugin for rust for automatic build file management: https://github.com/Calsign/gazelle_rust

One of the things that the gazelle plugin handles is dependencies on external crates pulled in through crate_universe. This requires parsing lockfiles. Depending on the configuration, rules_rust currently uses both cargo lockfiles (`Cargo.lock`) and custom "bazel lockfiles" (`Cargo.Bazel.lock`). While it is not currently possible to go fully cargo-free (i.e., no `Cargo.lock`), [this comment](https://github.com/bazelbuild/rules_rust/issues/1674#issuecomment-1330900533) suggests that allowing a cargo-free setup is desired. Thus to support cargo-free mode, the gazelle plugin needs to be able to parse bazel lockfiles.

Parsing bazel lockfiles is tricky because the interface is not exposed and it is subject to change across rules_rust versions. My current approach involves [depending on cargo_bazel](https://github.com/Calsign/gazelle_rust/blob/main/rust_parser/BUILD.bazel#L22) from the version of rules_rust used in each project and [patching rules_rust](https://github.com/Calsign/gazelle_rust/blob/main/patches/rules_rust.patch) to expose the required modules from cargo_bazel (`context` and `utils`). This is suboptimal.

I would like the gazelle plugin to be usable without needing to patch rules_rust. What's the best approach here? I'm thinking the `context` and `utils` modules could be split out into a separate crate that is publicly exposed. (This would also make the gazelle plugin build faster.) Is that reasonable? If so, I'm happy to put up the PR, just wanted to open a discussion first.

Separately it could make sense to integrate the gazelle plugin into rules_rust if that is desired. But I'll open a different issue for that.

illicitonion:
https://github.com/Calsign/gazelle_rust currently relies on a patch to rules_rust which makes the `Context` struct public.

Instead, supply the information they need via a public API.

All names (and whether we want to hide these structs behind traits and just return `impl Trait`s) are open to discussion.

Fixes #1725

cc @Calsign

### DIFF
--- a/crate_universe/src/api/lockfile.rs
+++ b/crate_universe/src/api/lockfile.rs
@@ -1,0 +1,220 @@
+//! The lockfile::public module represents a reasonable stable API for inspecting the contents of a lockfile which others can code against.
+
+use std::collections::BTreeSet;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
+
+use anyhow::Result;
+use serde::Deserialize;
+
+pub use crate::config::CrateId;
+use crate::context::crate_context::{CrateDependency, Rule};
+use crate::context::{CommonAttributes, Context};
+use crate::select::Select;
+
+/// Parse a lockfile at a path on disk.
+pub fn parse(path: &Path) -> Result<impl CargoBazelLockfile> {
+    let reader = BufReader::new(File::open(path)?);
+    let lockfile: CargoBazelLockfileImpl = serde_json::from_reader(reader)?;
+    Ok(lockfile)
+}
+
+/// CargoBazelLockfile provides a view over cargo-bazel's lockfile format,
+/// providing information about the third-party dependencies of a workspace.
+/// While the lockfile's format doesn't provide any kind of compatibility guarantees over time,
+/// this type offers an interface which is likely to be publicly supportable.
+/// No formal compatibility guarantees are offered around this type - it may change at any time,
+/// but the maintainers will attempt to keep it as stable they reasonably can.
+pub trait CargoBazelLockfile {
+    /// Get the members of the local workspace.
+    /// These are typically not very interesting on their own, but can be used as roots for navigating what dependencies these crates have.
+    fn workspace_members(&self) -> BTreeSet<CrateId>;
+
+    /// Get information about a specific crate (which may be in the local workspace, or an external dependency).
+    fn crate_info(&self, crate_id: &CrateId) -> Option<CrateInfo>;
+}
+
+#[derive(Deserialize)]
+#[serde(transparent)]
+struct CargoBazelLockfileImpl(Context);
+
+impl CargoBazelLockfile for CargoBazelLockfileImpl {
+    fn workspace_members(&self) -> BTreeSet<CrateId> {
+        self.0.workspace_members.keys().cloned().collect()
+    }
+
+    fn crate_info(&self, crate_id: &CrateId) -> Option<CrateInfo> {
+        let crate_context = self.0.crates.get(crate_id)?;
+        Some(CrateInfo {
+            name: crate_context.name.clone(),
+            version: crate_context.version.clone(),
+            library_target_name: crate_context.library_target_name.clone(),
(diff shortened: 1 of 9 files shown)


Calsign:
Thanks! This looks great. I haven't gotten around to doing this yet so thanks for picking up the slack.

Just for reference, there are additional changes which could make this situation even better:
1. Split the logic for parsing lockfiles into a separate crate so that we don't need to build all of cargo_bazel to build gazelle_rust
2. Add some kind of test checking compatibility of the public lockfile API

For 1, I'll do this at some point in the future if I'm sufficiently motivated. For now it's not that big of a deal.

For 2, that would need some kind of guarantee from rules_rust which the project is understandably not ready to provide. Improving the status quo to not require a patch is great and definitely good enough for now. But at some point in the future it would be nice to have a stronger guarantee.

crate_universe/src/lockfile/public.rs
```diff
@@ -0,0 +1,194 @@
...
+                .targets
+                .iter()
+                .any(|t| matches!(t, Rule::ProcMacro(_))),
+            common_attributes: crate_context.common_attrs.clone(),
+        })
+    }
+}
+
+#[derive(Deserialize, PartialEq, Eq, Debug)]
+pub struct CrateInfo {
```
UebelAndre, line 47: If this is gonna be a public interface can we get full docstrings on these?

crate_universe/src/lockfile/public.rs
```diff
@@ -0,0 +1,194 @@
...
+    name: String,
+    // Or semver::Version?
```
UebelAndre, line 49: Sounds like a good idea?

crate_universe/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub mod cli;
 
 mod config;
 mod context;
-mod lockfile;
+pub mod lockfile;
```
UebelAndre, line 7: Maybe we should have an `api` module instead? I worry about leaking things from `lockfile` since now we'd need to explicitly state `pub(crate)` Maybe that's a good thing though? I'll defer to you 😄

UebelAndre (approved):
This seems fine to me!

Had a few nits/questions but nothing urgent or worth blocking merge 😄

illicitonion:
I think this is ready to go (and agree with @Calsign's future extensions, but would treat them as extensions :)) - the first three commits are just https://github.com/bazelbuild/rules_rust/pull/2525

crate_universe/test_data/cargo_bazel_lockfile/multi_package-cargo-bazel-lock.json
UebelAndre, line 1: @illicitonion question: how was this file generated? Could you add some docs for it?

crate_universe/src/api/lockfile.rs
```diff
@@ -0,0 +1,220 @@
...
+#[cfg(test)]
+mod test {
+    use super::{parse, CargoBazelLockfile};
+    use crate::config::CrateId;
+    use crate::context::crate_context::CrateDependency;
+    use semver::Version;
+    use std::collections::BTreeSet;
+
+    #[test]
+    fn test() {
```
UebelAndre, line 125: @illicitonion What is this testing?

illicitonion: See https://github.com/bazelbuild/rules_rust/pull/2637 :)

crate_universe/test_data/cargo_bazel_lockfile/multi_package-cargo-bazel-lock.json
illicitonion: See https://github.com/bazelbuild/rules_rust/pull/2637 :)

PR #2516, opened 2024-02-24
Fix all-crate-deps with bzlmod and conditionals.

criemen:
I believe this was an oversight from #2491, that this is lacking.


UebelAndre (approved):
Thanks!

PR #2525, opened 2024-02-28
Split CrateID CrateNameAndVersionReq

illicitonion:
Currently we use the same CrateId type for two contexts: A known crate
name + version, and in an annotation to specify a version requirement
that an annotation applies to.

These are different use-cases, and neither should use String to
represent a the version.

Also, update CrateId to use a `semver::Version` (which is required, not optional).

### DIFF
--- a/crate_universe/src/config.rs
+++ b/crate_universe/src/config.rs
@@ -519,12 +519,12 @@ pub struct CrateId {
     pub name: String,
 
     /// The crate's semantic version
-    pub version: String,
+    pub version: semver::Version,
 }
 
 impl CrateId {
     /// Construct a new [CrateId]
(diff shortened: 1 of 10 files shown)


crate_universe/src/config.rs
```diff
@@ -708,6 +681,175 @@ impl Config {
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        deserializer.deserialize_str(CrateNameAndVersionReqVisitor)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct VersionReqString {
+    original: String,
```
UebelAndre, line 742: Can you add docs for anything public?

UebelAndre (approved):
Sounds good to me, just had one request 😄

PR #2526, opened 2024-02-28
Use pub(crate) rather than pub

illicitonion:
Now that we actually have some pub items, this will help us to avoid accidentally leaking more (e.g. in
https://github.com/bazelbuild/rules_rust/pull/2515 we are leaking `CrateId`, `Select`, and `CrateDependency` to be public because they're marked `pub` not `pub(crate)`.

I'm not sure I like this, but wanted send as an RFC to discuss.


UebelAndre (approved):
I might be out of the loop but is this common practice for crates? Was the before "lazy"?

I can do some digging on my own, but this seems just fine to me 😄

illicitonion:
I think the before was maybe a bit lazy, but also honestly fine? But happy to merge and see how we go!

Issue #2522, opened 2024-02-27
Genquery: cycle in dependency graph


UebelAndre:
It seems using `genquery` on any Rust targets leads to failures due to a cycle in the dependency graph. The diff below can be used to repro the issue and get the output that follows.
```diff
diff --git a/crate_universe/BUILD.bazel b/crate_universe/BUILD.bazel
index 370dc5c3..ffb18c11 100644
--- a/crate_universe/BUILD.bazel
+++ b/crate_universe/BUILD.bazel
@@ -76,6 +76,12 @@ rust_binary(
     deps = [":cargo_bazel"],
 )

+genquery(
+    name = "genquery",
+    scope = [":cargo_bazel_bin"],
+    expression = "deps(:cargo_bazel_bin)",
+)
+
 alias(
     name = "bin",
     actual = ":cargo_bazel_bin",
```

```
~/Code/rules_rust (main ✗) bazel build //crate_universe:genquery
ERROR: /Users/user/Code/rules_rust/util/process_wrapper/BUILD.bazel:31:36: in rust_binary_without_process_wrapper rule //util/process_wrapper:process_wrapper: cycle in dependency graph:
    //crate_universe:genquery (e5528793e6772dcf5ff2b9774ea78fbb21b115ab96d65cadb8cb5b68206d2fe6)
    //crate_universe:genquery (577c459edef5fa899d167957b8c740aeb5edc881970dd81d94adeef309381026)
    //crate_universe:cargo_bazel_bin
.-> //util/process_wrapper:process_wrapper
|   //util/import:import
|   //util/import:import_macro_label
|   //util/import:import_macro
|   //util/import:import_macro_impl
`-- //util/process_wrapper:process_wrapper
ERROR: Analysis of target '//crate_universe:genquery' failed; build aborted
```

UebelAndre:
cc @scentini @cfredric

UebelAndre:
I believe the issue is caused by https://github.com/bazelbuild/rules_rust/blob/0.39.0/rust/private/rust.bzl#L531-L534

daivinhtran:
Could you the fix simply be removing `_import_macro_dep` from `rust_binary_without_process_wrapper`?

UebelAndre:
[...]

Issue #2383, opened 2023-12-29
Bazel query is broken for Rust targets


csmulhern:
bazel version: 7.0.0
rules_rust version: 0.36.1

Sample: [example.zip](https://github.com/bazelbuild/rules_rust/files/13795396/example.zip)

To reproduce, run:
```bash
bazel query "deps(//:main)"
```

Output:
```bash
ERROR: /private/var/tmp/_bazel_cameron/bd7f4772f50608f24ec4c16fe5369dcc/external/rules_rust/util/import/3rdparty/crates/BUILD.bazel:52:6: no such package '@@rules_rust_util_import__quote-1.0.10//': The repository '@@rules_rust_util_import__quote-1.0.10' could not be resolved: Repository '@@rules_rust_util_import__quote-1.0.10' is not defined and referenced by '@@rules_rust//util/import/3rdparty/crates:quote'
ERROR: Evaluation of query "deps(//:main)" failed: preloading transitive closure failed: no such package '@@rules_rust_util_import__quote-1.0.10//': The repository '@@rules_rust_util_import__quote-1.0.10' could not be resolved: Repository '@@rules_rust_util_import__quote-1.0.10' is not defined
```

It looks like there is an implicit dependency on `@rules_rust/util/import` for all rust rules, but these dependencies are not included through `rules_rust_dependencies`.

I'm guessing this dependency is coming from: 
https://github.com/bazelbuild/rules_rust/blob/e6a1ee48cd98d2e1c277a2c5823e1b0c0307297e/rust/private/rust.bzl#L531-L533

The simplest fix seems like it would be to add `import_deps` to `rules_rust_dependencies`. However, I'm also wondering if it's really necessary for all rules to depend on `@rules_rust/util/import`.

This issue prevents usage of `bazel query` with rust targets, and breaks workflows that depend on `bazel query`, e.g. `ibazel`.

illicitonion:
It looks like Bazel 7 made `label_flag` more eager during query: https://github.com/bazelbuild/bazel/commit/658b3a6a4811c7907dd07e949182ba571e96138a

cc @cfredric @hlopko as authors of the feature

cc @gregestren - it looks like rules_rust was depending on _not_ reading through `label_flag` to avoiding dependencies needing to be set up for a niche use-case - do you know of any handy workarounds?

scentini:
@gregestren Could you please advise what to do here?

katre:
[...]

PR #2559, opened 2024-03-18
Fixed genquery for rust targets

UebelAndre:
relates to https://github.com/bazelbuild/rules_rust/issues/2522

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -528,10 +528,6 @@ RUSTC_ATTRS = {
     "_extra_rustc_flags": attr.label(
         default = Label("//:extra_rustc_flags"),
     ),
-    "_import_macro_dep": attr.label(
-        default = Label("//util/import"),
-        cfg = "exec",
-    ),
     "_is_proc_macro_dep": attr.label(
         default = Label("//rust/private:is_proc_macro_dep"),
     ),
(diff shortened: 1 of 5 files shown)


daivinhtran (commented):
I'm not an expert in this corner of the rules but shouldn't we also remove [this](https://github.com/bazelbuild/rules_rust/blob/8b977b5f7020fc3a3b3bbc318e414269014f126a/rust/private/utils.bzl#L486) altogether?

UebelAndre:
> I'm not an expert in this corner of the rules but shouldn't we also remove [this](https://github.com/bazelbuild/rules_rust/blob/8b977b5f7020fc3a3b3bbc318e414269014f126a/rust/private/utils.bzl#L486) altogether?

I'm not sure this should be completely deleted to follow the spirit of https://github.com/bazelbuild/rules_rust/issues/2383#issuecomment-1883287133. Sounds like Google has a use for this code but I don't think it's acceptable to have basic functionality be broken by a shell of a tool. If there is a more desired outcome then I'm happy to update the PR if the change is small. But I think it should fall to Googlers to improve `//util/import` to the point where it can coexist with the rest of rules_rust.

daivinhtran:
@UebelAndre 

Would you mind waiting until @scentini is back next week? @scentini has more context and is on vacation this week. I need to sync with them to have a better understanding.

scentini:
Thanks folks! We indeed still use the import macro in google's internal codebase and the PR in its current form works the best;
I will do some investigation on what is expected from genquery in this context.

scentini (approved):

PR #2563, opened 2024-03-19
Cleanup buildifier conditions

dzbarsky:


UebelAndre (changes requested):
The use of `@bazel_tools` seems like a step back. Can this instead use the `@platforms` repository?

dzbarsky:
> The use of `@bazel_tools` seems like a step back. 
Why is that? I thought that was the canonical place for these conditions. There's plenty of other things in `@bazel_tools` that you are supposed to use, such as `@bazel_tools//tools/build_defs/repo:http.bzl` or `@bazel_tools//tools/bash/runfiles`. Looks like the repo is already using `@bazel_tools//src/conditions` for the LLVM patches (though those usages can be moved to @platforms since they are OS-only)

> Can this instead use the `@platforms` repository?
I think @platforms only defines the OS and CPU constraints; I don't think it has the conditions forming their cross-product. Maybe I'm missing something?

dzbarsky:
Another option, if you're OK taking on the dependency, is to use https://github.com/keith/buildifier-prebuilt. This also has the advantage of `bazel query` not downloading buildifiers for all the platforms, and users may have this in their repo already

dzbarsky:
OK if `@bazel_tools//src/conditions` is going away, this isn't worth it

PR #2565, opened 2024-03-19
Allow a no-cargo setup for bzlmod

dzbarsky:
Usage looks like so:
```
crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
crate.spec(package = "anyhow", version = "1.0.77")
....
crate.from_specs(name = "crates")
```

It might make sense to merge the annotation attributes into the spec so we don't have to duplicate things, but we can probably iterate on this in the future, this API is still experimental, yeah?


dzbarsky:
@matts1 You're probably best-equipped to review this if you have the cycles

examples/bzlmod/hello_world_no_cargo/hello_world_test.sh
matts1, line 1: Nit: I've recently been made aware of a much better way to do this.

Can you add `bazel_dep(name = "aspect_bazel_lib", version = "1.36.0")` to `MODULE.bazel`, then add the following to the build file:

```
load("@aspect_bazel_lib//lib:run_binary.bzl", "run_binary)
load("@bazel_skylib//rules:diff_test.bzl", "diff_test")

run_binary(
  name = "run_hello_world",
  tool = ":hello_world",
)

diff_test(
    name = "hello_world_diff_test",
    file1 = ":run_hello_world",
    file2 = "testdata/hello_world.out",
)
```

examples/bzlmod/hello_world_no_cargo/src/main.rs
matts1, line 1: Nit: Can we make most of the files in the directory a symlink to `examples/bzlmod/hello_world/*`? That should help us keep them in sync - the only file that should actually be different is MODULE.bazel, but the rest should be the same.

crate_universe/extension.bzl
```diff
@@ -225,7 +258,23 @@ def _crate_impl(module_ctx):
...
             for crate, values in repo_specific_annotations.get(cfg.name, {}).items():
                 _get_or_insert(annotations, crate, []).extend(values)
-            _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg, annotations)
+            cargo_lockfile = module_ctx.path(cfg.cargo_lockfile)
+            manifests = {str(module_ctx.path(m)): str(m) for m in cfg.manifests}
+            _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg, annotations, cargo_lockfile = cargo_lockfile, manifests = manifests)
+            all_repos.append(cfg.name)
+            local_repos.append(cfg.name)
+
+        for cfg in mod.tags.from_specs:
```
matts1, line 267: Nit: Looks like a lot of duplicated code between this and the above section. Could we pull the common stuff into a function?

matts1 (changes requested):
A few nitpicks, but generally looks good to me

examples/bzlmod/hello_world_no_cargo/hello_world_test.sh
dzbarsky: This would need more changes to the binary to direct the output to a file, or another wrapper. But really I don't even need to run it, I'll just switch this to a `build_test`. As long as it compiles we're ok :)

matts1: That's a good point, sounds good to me. Don't know if we even need a build_test. `bazel test :all` actually builds all non-test targets as well IIRC.

examples/bzlmod/hello_world_no_cargo/src/main.rs
dzbarsky: I added `--noenable_workspace` to the `.bazelrc`, so now we only have BUILD.bazel, MODULE.bazel, .bazelrc and src/main.rs. The last one is the only one that is the same. I tried symlinking it but Bazel `rust_binary` seems to not pick up the symlinked source file. Given the source is pretty trivial, it didn't seem worth poking at it more, but let me know if you feel strongly!

examples/bzlmod/hello_world_no_cargo/hello_world_test.sh
dzbarsky: that's true, I'll just leave the `rust_binary`

crate_universe/extension.bzl
dzbarsky: good call, I rearranged this a bit to share the repo validation. So the duplication now is the 3 lines for `annotations`. I'm hesitating a bit because sharing this either requires using a lambda or a function that takes `module_annotations`, `repo_specific_annotations`, and `cfg.name`. Alternately, it can look like `this_repo_annotations = update_annotations(module_annotations, repo_specific_annotations.get(cfg.name, {})` or something. 

Do you still think it's worth pulling out or keep it as-is?

matts1: I quite like the `repo_specific_annotations` idea. I think that'd probably be the cleanest.

examples/bzlmod/hello_world_no_cargo/src/main.rs
matts1: That's very curious, but yeah, don't think it's worth digging into.

crate_universe/extension.bzl
```diff
@@ -290,10 +325,39 @@ _annotation = tag_class(
     ),
 )
 
+_from_specs = tag_class(
+    doc = "Generates a repo @crates from a Cargo.toml / Cargo.lock pair",
```
matts1, line 329: Nit: This looks like it needs to be updated

matts1 (approved):

crate_universe/extension.bzl
dzbarsky: yep good catch, ty!

dzbarsky:
@UebelAndre @illicitonion can we merge this?

crate_universe/extension.bzl
```diff
@@ -290,10 +325,39 @@ _annotation = tag_class(
...
+        generate_binaries = CRATES_VENDOR_ATTRS["generate_binaries"],
+        generate_build_scripts = CRATES_VENDOR_ATTRS["generate_build_scripts"],
+        supported_platform_triples = CRATES_VENDOR_ATTRS["supported_platform_triples"],
+    ),
+)
+
+# This should be kept in sync with crate_universe/private/crate.bzl.
+_spec = tag_class(
+    attrs = dict(
+        package = attr.string(doc = "The explicit name of the package.", mandatory = True),
```
UebelAndre, line 342: nit can you expand these attrs?
```suggestion
        package = attr.string(
            doc = "The explicit name of the package.",
            mandatory = True,
        ),
```

crate_universe/extension.bzl
```diff
@@ -290,10 +325,39 @@ _annotation = tag_class(
...
+        name = attr.string(doc = "The name of the repo to generate", default = "crates"),
+        cargo_config = CRATES_VENDOR_ATTRS["cargo_config"],
+        generate_binaries = CRATES_VENDOR_ATTRS["generate_binaries"],
+        generate_build_scripts = CRATES_VENDOR_ATTRS["generate_build_scripts"],
+        supported_platform_triples = CRATES_VENDOR_ATTRS["supported_platform_triples"],
+    ),
+)
+
+# This should be kept in sync with crate_universe/private/crate.bzl.
+_spec = tag_class(
```
UebelAndre, line 340: nit: can you have globals be ALL_CAPS?
```suggestion
_SPEC = tag_class(
```

examples/bzlmod/hello_world_no_cargo/.bazelrc
```diff
@@ -0,0 +1,3 @@
+common --experimental_enable_bzlmod
+common --noenable_workspace
+common --enable_runfiles
```
UebelAndre, line 3: Nit: add a newline (I really need to add a lint for this)
```suggestion
common --enable_runfiles

```

crate_universe/extension.bzl
```diff
@@ -40,8 +40,16 @@ _generate_repo = repository_rule(
...
+def _annotations_for_repo(module_annotations, repo_specific_annotations):
+    if not repo_specific_annotations:
+        return module_annotations
+
+    annotations = dict(module_annotations)
+    for crate, values in repo_specific_annotations.items():
+        _get_or_insert(annotations, crate, []).extend(values)
+    return annotations
+
+def _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg, annotations, cargo_lockfile = None, manifests = {}, packages = {}):
```
UebelAndre, line 52: Sorry to ask but can you add a docstring here? I'm still very new to bzlmod so having the params and intent documented here would help a lot.

UebelAndre (changes requested):
Thanks! A couple of nits and can you add a build in CI that runs on this workspace? Or some kind of regression test somewhere else?

~~I'd also love to know if @matts1 approves too.~~ (They do!)

crate_universe/extension.bzl
dzbarsky: sure, will do

crate_universe/extension.bzl
dzbarsky: That wouldn't match the existing `tag_class` definitions for `from_cargo` and `annotation`, and I think it would make the `module_extension` call look funky. These are basically functions and `tag_class` is sort of a function factory if you squint a little.

dzbarsky:
> Thanks! A couple of nits and can you add a build in CI that runs on this workspace? Or some kind of regression test somewhere else?

Oh I had assumed we were already testing the examples in CI. Should we fix that instead?

UebelAndre:
> Oh I had assumed we were already testing the examples in CI. Should we fix that instead?

A new workspace requires a new job. Had this been an expansion of an existing case the it would have been picked up

examples/bzlmod/hello_world_no_cargo/MODULE.bazel
UebelAndre, line 1: Can you make sure this directory is added to `examples/.bazelignore`?

crate_universe/extension.bzl
```diff
@@ -40,8 +40,28 @@ _generate_repo = repository_rule(
     ),
 )
 
-def _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg, annotations):
-    cargo_lockfile = module_ctx.path(cfg.cargo_lockfile)
+def _annotations_for_repo(module_annotations, repo_specific_annotations):
```
UebelAndre, line 43: nit: Do you wanna document this one as well 😅

UebelAndre (changes requested):
Nice! I had just one more request to fix the `.bazelignore` file and then I think it's good to go!

examples/bzlmod/hello_world_no_cargo/MODULE.bazel
dzbarsky: `bzlmod` is already in `examples/.bazelignore` so I think it's not necessary?

crate_universe/extension.bzl
dzbarsky: Sure, done

matts1:
> > Thanks! A couple of nits and can you add a build in CI that runs on this workspace? Or some kind of regression test somewhere else?
> 
> Oh I had assumed we were already testing the examples in CI. Should we fix that instead?

Since you made a new workspace, you have to create a new entry. We are testing bzlmod/hello_world in CI already, you should be able to just copy-paste that one.

UebelAndre (approved):
Thank you!

examples/bzlmod/hello_world_no_cargo/MODULE.bazel
UebelAndre: you're correct, my bad!

PR #2566, opened 2024-03-20
Bzlmod-aware runfiles library

dzbarsky:
This implements the repo_mapping capabilities and provides a new runfiles API. The new API can be accessed explicitly as `runfiles::Runfiles::rlocation_from` or with the `runfiles::rlocation!` macro, which adds compile-time support for correctly embedding the external repo. This is a purely new API, existing usage continues to work, although we mark it deprecated because it's not fully correct. We can remove it at some point in the future.

This PR also transitions in-repo examples/tests to using it, in case anyone copies them.


tools/runfiles/runfiles.rs
```diff
@@ -19,13 +19,13 @@
 //!     use runfiles::Runfiles;
 //!     ```
 //!
-//! 3.  Create a Runfiles object and use rlocation to look up runfile paths:
+//! 3.  Create a Runfiles object and use Rlocation! to look up runfile paths:
```
matts1, line 22: Should it be title-case? I was under the impression that naming conventions dictated that since it's a function, it should be `rlocation`

tools/runfiles/runfiles.rs
```diff
@@ -44,36 +44,49 @@ const MANIFEST_FILE_ENV_VAR: &str = "RUNFILES_MANIFEST_FILE";
...
 }
 
 impl Runfiles {
     /// Creates a manifest based Runfiles object when
     /// RUNFILES_MANIFEST_ONLY environment variable is present,
     /// or a directory based Runfiles object otherwise.
     pub fn create() -> io::Result<Self> {
-        if is_manifest_only() {
+        let mode = if is_manifest_only() {
             Self::create_manifest_based()
```
matts1, line 75: IIRC @fmeum said that the canonical implementation is at [rules_python](https://github.com/bazelbuild/bazel/blob/master/tools/python/runfiles/runfiles.py) and that other implementations should seek to follow that. In all other implementations, we create the manifest based one if RUNFILES_MANIFEST_FILE exists, otherwise we create it based on RUNFILES_DIR.

I believe we should instead have:
```
let mode = if let Ok(manifest_file) = std::env::var(MANIFEST_FILE_ENV_VAR) {
  create_manifest_based(Path::from(manifest_file))?
} else {
  Mode::DirectoryBased(find_runfiles_dir()?)
}
```

tools/runfiles/runfiles.rs
```diff
@@ -44,36 +44,49 @@ const MANIFEST_FILE_ENV_VAR: &str = "RUNFILES_MANIFEST_FILE";
 const MANIFEST_ONLY_ENV_VAR: &str = "RUNFILES_MANIFEST_ONLY";
 const TEST_SRCDIR_ENV_VAR: &str = "TEST_SRCDIR";
 
+#[macro_export]
+macro_rules! Rlocation {
```
matts1, line 48: I'm not going to say we should do it one way or the other, but we should definitely at least consider the folllowing API as an alternative:

```
let  r = runfiles::create!() // Expands to runfiles::create(env!("REPOSITORY_NAME"))
r.rlocation(path)
```

Pros:
* More clean API (less macros)
* More performant (only need to calculate the repo mapping once)
* Simpler to migrate to (only migrate the runfiles constructor, not the rlocation calls)

Cons:
* If you're passing the runfiles constructor across repo boundaries, it might do the wrong repo mapping. I can't see this ever happening though.

tools/runfiles/runfiles.rs
dzbarsky: sure, I wasn't sure what the convention was for macros but I think you're right

tools/runfiles/runfiles.rs
dzbarsky: I'm happy to swap to that. It's a bit riskier since it's a behavior change but it's more correct you point out

tools/runfiles/runfiles.rs
dzbarsky: Yep I asked @fmeum the same and he mentioned passing across repo boundaries is a nice to have. I'm fine either but let's figure it out in this thread before I refactor more :)

Re: more performant - I think it's the same? The mapping is only calculated at startup? Or if you're referring to the map lookups + string ops - you're about to do File IO, I don't think it matters :)

The migration path is also a good point. I think as-written the code doesn't _require_ any client changes, but they can move to the new API to become more correct. I think with your proposal everyone using runfiles needs a minor code change?

rust/private/rustc.bzl
```diff
@@ -1095,6 +1095,9 @@ def construct_arguments(
     if _is_no_std(ctx, toolchain, crate_info):
         rustc_flags.add('--cfg=feature="no_std"')
 
+    # Needed for bzlmod-aware runfiles resolution.
+    env["REPOSITORY_NAME"] = ctx.label.workspace_name
```
UebelAndre, line 1099: Why wasn't the `workspace_name.env` file sufficient? Have other projects been adding `REPOSITORY_NAME` for runfiles support? I'd like to avoid adding things to the `Rustc` action if possible.

UebelAndre (dismissed):

rust/private/rustc.bzl
dzbarsky: If my understanding is correct, that file always encodes the rules_rust repo name. What we want is the repo_name of the calling code. So if the `_main` repo depends on `foo` which uses runfiles, it should be the mapped name for `foo` when code in `foo` looks up runfiles.

I checked the python and Go implementations and they use reflection/stack-walking to figure out the caller. Java [uses an annotation](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java#L61-L74), C++ allows you to pass it into the constructor at creation time https://github.com/bazelbuild/bazel/blob/5594166dcbcdf746117ee4106ff5b6679434016c/tools/cpp/runfiles/runfiles_src.h#L129-L133

fmeum: @dzbarsky is right here, this kind of magic is indeed required to get the correct lookup behavior. I looked into this a bit more a while back and couldn't find any way to do this without `Rustc` changes.

tools/runfiles/runfiles.rs
fmeum: `RUNFILES_MANIFEST_ONLY` is a trap that breaks cross-platform builds and should not be used. Instead, I would say the following flow is the best you can do at this point in time:
1. If at least one of `RUNFILES_MANIFEST_FILE` or `RUNFILES_DIR` is set:
    1. If `RUNFILES_MANIFEST_FILE` is set and exists, use it.
    2. If `RUNFILES_DIR` is set and exists, use it.
    3. Error.
2. If the runfiles manifest exists in one of the well-known locations, use it.
3. If the runfiles directory exists in one of the well-known locations, use it.
4. Error.

rust/private/rustfmt.bzl
```diff
@@ -13,7 +13,7 @@ def _get_rustfmt_ready_crate_info(target):
     """
 
     # Ignore external targets
-    if target.label.workspace_root.startswith("external"):
+    if target.label.workspace_name:
```
dzbarsky, line 16: driveby cleanup to not depend on `external`

tools/runfiles/runfiles.rs
matts1: Ah, good point about the migration path, I hadn't notice that we left the old method, so I thought this whole PR was backwards incompatible. If that's the case, I think we definitely just do it the way you've already done it.

rust/private/rustc.bzl
matts1: This looks good to me. I spent multiple hours trying to understand how the old one worked about a year ago, and never did work out how - turns out my understanding was correct and it was never working.

dzbarsky:
@UebelAndre this is passing tests now. I'm planning to add an external-repo integration test as well to double check, but assuming that is OK, can we merge this? None of us have a better idea how to make this work correctly and if I understand correctly the state of the world, enabling bzlmod in a repo with rust code will make rust-analyzer not work correctly, even if the rust rules are not using bzlmod. Would be good to get a fix in

fmeum (approved):

tools/runfiles/runfiles.rs
```diff
@@ -19,62 +19,72 @@
...
 
 impl Runfiles {
     /// Creates a manifest based Runfiles object when
     /// RUNFILES_MANIFEST_ONLY environment variable is present,
     /// or a directory based Runfiles object otherwise.
     pub fn create() -> io::Result<Self> {
-        if is_manifest_only() {
-            Self::create_manifest_based()
+        let mode = if let Ok(manifest_file) = std::env::var(MANIFEST_FILE_ENV_VAR) {
+            Self::create_manifest_based(&PathBuf::from(manifest_file))?
```
illicitonion, line 73:
```suggestion
        let mode = if let Some(manifest_file) = std::env::var_os(MANIFEST_FILE_ENV_VAR) {
            Self::create_manifest_based(&Path::new(&manifest_file))?
```

tools/runfiles/runfiles.rs
```diff
@@ -85,46 +95,78 @@ impl Runfiles {
...
-        })
+        Ok(Mode::ManifestBased(path_mapping))
     }
 
     /// Returns the runtime path of a runfile.
     ///
     /// Runfiles are data-dependencies of Bazel-built binaries and tests.
     /// The returned path may not be valid. The caller should check the path's
     /// validity and that the path exists.
+    /// @deprecated - this is not bzlmod-aware. Prefer the `Rlocation!` macro or `rlocation_from`
```
illicitonion, line 106:
```suggestion
    /// @deprecated - this is not bzlmod-aware. Prefer the `rlocation!` macro or `rlocation_from`
```

tools/runfiles/runfiles.rs
```diff
@@ -85,46 +95,78 @@ impl Runfiles {
...
+            .clone(),
     }
 }
 
+fn parse_repo_mapping(path: PathBuf) -> io::Result<RepoMapping> {
+    Ok(std::fs::read_to_string(path)?
+        .lines()
+        .map(|line| {
+            let parts: Vec<String> = line.splitn(3, ',').map(String::from).collect();
+            ((parts[0].clone(), parts[1].clone()), parts[2].clone())
```
illicitonion, line 162: It would be good to error rather than panic if this file is malformed and doesn't contain 3 elements

illicitonion (commented):
LGTM, thanks! It'd be great to have that external repo test you mentioned.

tools/runfiles/runfiles.rs
dzbarsky: done, I also fixed the duplicative string copying while I was here

dzbarsky:
> LGTM, thanks! It'd be great to have that external repo test you mentioned.

I added it as a separate commit. It's a bit complicated setup but it's what we need to really verify we are using the correct repo_mapping of transitive repos instead of just the main one

matts1 (approved):

UebelAndre:
Does this break the existing API? Can you update the PR description to reflect the changes some more?

illicitonion (approved):
Works for me! Will leave for @UebelAndre to review/merge when happy

dzbarsky:
@UebelAndre does the PR description make sense now? Anything else you'd like to see before this lands?

dzbarsky:
@UebelAndre @illicitonion What can I do to get this unstuck and landed?

This PR is blocking other issues, such as https://github.com/bazelbuild/rules_rust/issues/2615. 

There have been other attempts to fix this for over a year now, such as: [first one](https://github.com/bazelbuild/rules_rust/pull/1847), [second one](https://github.com/bazelbuild/rules_rust/pull/2440).

I think I have spent more time trying to get this PR landed than it took to write the code, which is a little frustrating as you can imagine.

scentini:
This PR LGTM, and as it was already approved by @illicitonion a while back, I'm going to to ahead and merge it. @UebelAndre please feel free to do post-merge review, and we'll follow up.

UebelAndre:
Sorry, this one fell through the cracks. Didn't mean to delay the change!

PR #2575, opened 2024-03-27
Mark the bzlmod extension reproducible as appropriate

dzbarsky:
@matts1 @fmeum PTAL :)

### DIFF
--- a/crate_universe/extension.bzl
+++ b/crate_universe/extension.bzl
@@ -1,5 +1,6 @@
 """Module extension for generating third-party crates for use in bazel."""
 
+load("@bazel_features//:features.bzl", "bazel_features")
 load("@bazel_skylib//lib:structs.bzl", "structs")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
@@ -275,6 +276,8 @@ def _crate_impl(module_ctx):
     cargo_bazel = get_cargo_bazel_runner(module_ctx, cargo_bazel_output)
 
     all_repos = []
+    reproducible = True
+
     for mod in module_ctx.modules:
         module_annotations = {}
         repo_specific_annotations = {}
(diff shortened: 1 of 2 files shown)


matts1 (approved):
Ah, it's nice to see that they've added reproducible.

DavidZbarsky-at:
@UebelAndre can we get this in as well?

UebelAndre:
@dzbarsky can you update the PR description to explain what this change does? What does it mean to "Mark the bzlmod extension reproducible"?

matts1:
A reproducible extension doesn't get put in the bzlmod lockfile.

This code basically says that if we're already using a Cargo.lock lockfile for, we don't need to add it to MODULE.bazel.lock, which is the repository rule lockfile.

crate_universe/extension.bzl
```diff
@@ -339,6 +347,12 @@ def _crate_impl(module_ctx):
             if repo not in local_repos:
                 fail("Annotation specified for repo %s, but the module defined repositories %s" % (repo, local_repos))
 
+    metadata_kwargs = {}
+    if bazel_features.external_deps.extension_metadata_has_reproducible:
+        metadata_kwargs["reproducible"] = reproducible
+
+    return module_ctx.extension_metadata(**metadata_kwargs)
```
fmeum, line 354: Depending on your Bazel requirement, it may be safer to test for `extension_metadata` via `hasattr`.

fmeum (approved):
LGTM from the bzlmod perspective. I don't know rust well enough to say whether there are other sources of non-reproducibility.

UebelAndre (approved):

crate_universe/extension.bzl
UebelAndre: @dzbarsky thoughts?

dzbarsky: I believe `extension_metadata` is present in 6.2.0+ and we support a min of 6.3.0

dzbarsky: https://bazel.build/versions/6.2.0/rules/lib/extension_metadata

UebelAndre: @fmeum was your comment suggesting that we could eliminate the new dependency on on `bazel_features` by using `hasattr`?

DavidZbarsky-at: We cannot. The 'reproducible' argument is newer than 'extension_metadata'

Issue #2524, opened 2024-02-28
Vendoring crates with a "+" in the version breaks the deps graph


mattboardman:
It appears that "+" characters in a crate version are translated to a hyphen ("-") character in the generated BUILD files for a crate when using the vendor local configuration. [code pointer](https://github.com/bazelbuild/rules_rust/blob/33860abf8c3c28f9a1f5b3477347650bdaee9cc3/crate_universe/extension.bzl#L137)

So all crates that depend on `abc-1.0+v1.0` will have the crate formatted as `abc-1.0-v1.0:abc` in any rdeps.

However, the directory name for the vendored crate will still contain the original name with the + character in the version. [example crate that will cause the issue](https://crates.io/crates/libbpf-sys)

This leads to an error when building the dependency graph:

```
no such package 'third_party/rust/crates/libbpf-sys-1.3.0-v1.3.0': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
```

PR #2579, opened 2024-03-27
Fix (+) In vendored crates semver metadata

ericmcbride:
- Resolves https://github.com/bazelbuild/rules_rust/issues/2524
- Revendored examples deps.
- Semver Metadata can break vendored dependencies, by cargo vendoring (foo-1.2.3+1.2.3) for example.  The bazel labels will look for `foo-1.2.3-1.2.3` which does not work.
- Root of the issue is, rules rust uses cargo to vendor the dependencies, which takes the semver + metadata, and generates the file path based off that (hence the + for the dependency in the issue)
- Tested by taking that dependency and make sure it loaded using the `vendor_local_manifests` example
- Did not include that dependency, because it would introduce a sys dep in CI 
- Opted to do it within the scope of the vendor cli.  When trying to change the version in the `render_build_files` call, it fixes the version in terms of the `outputs` and `files` declared in the vendor cli file, but then we do not have a handle to the generated vendor files from cargo, meaning we cant move the contents over near as easily.  I also didn't want to break anything downstream that may rely on that info.


crate_universe/src/rendering.rs
```diff
@@ -802,14 +802,29 @@ pub(crate) fn write_outputs(
...
+            // metadata on the semver, which may not be compatible with bazel (+) for example.
+            // So we need to check the paths to see if they contain a +.  If they do we need to
+            // move the directory with all of its contents to a new directory that does not
+            // contain special characters so its compatible with the bazel labels defined
+            // in the generated defs.bzl
+            let original_path_str = path
+                .parent()
+                .expect("All file paths should have valid directories")
+                .to_str()
+                .expect("All file paths should be strings");
```
ericmcbride, line 823: I can use anyhow here to propagate an error to the caller.  Was following the precedence already set in this function with `path.parent()` called above.

crate_universe/src/rendering.rs
```diff
@@ -802,14 +802,29 @@ pub(crate) fn write_outputs(
...
+            if original_path_str.contains('+') {
```
ericmcbride, line 824: Are worried about someones path on their computer having a + sign in it?  If so i may need to not re use the sanitize function

UebelAndre (changes requested):
Thanks! Could you also add a unit test for this? 🙏

crate_universe/src/rendering.rs
UebelAndre: Why not have this logic happen in the place where we decide the output name?
https://github.com/bazelbuild/rules_rust/blob/main/crate_universe/src/rendering.rs#L286

That way we don't need to bother with an additional file op and this code function would be unchanged, right?

UebelAndre (changes requested):
Thank you! I had just one question

crate_universe/src/rendering.rs
ericmcbride: I think I tried this and it had side effects somewhere . Let me change this line and make sure.  Im all for smaller code changes.

ericmcbride:
> Thank you! I had just one question

I had previously manipulated the version and not the label directly here.  Thanks for that catch!

crate_universe/src/rendering.rs
ericmcbride: If you just change the label, it will put the build file in the correct path, but the Crate itself will be stored in the + file path, due to how cargo vendors the dep.    We use cargo to run the vendor command, and it will auto generate the semver metadata with the plus sign

```
libbpf-sys-1.3.0+v1.3.0 
libbpf-sys-1.3.0-v1.3.0
```

```
>  ls libbpf-sys-1.3.0+v1.3.0 
bindings.h  build.rs  Cargo.toml  elfutils  libbpf  LICENSE  README.md  rebuild.sh  src  tests  zlib
```

```
>  ls libbpf-sys-1.3.0-v1.3.0 
BUILD.bazel
```

UebelAndre: What if you did the rename here?

https://github.com/bazelbuild/rules_rust/blob/0.41.0/crate_universe/src/rendering.rs#L292

If all this new code is doing is renaming a file, it seems like we could update the code to have a correct name in the first place. If the label shouldn't change then perhaps we can update just the output file? But the thing I'd like to avoid is renaming a file. We should be producing the correct data before writing.

UebelAndre: AAAAH! I see now!! Thank you for helping me and for being so patient! 

In this case, could you make a new function that does the renames? I'd totally forgotten cargo is what writes the directory names. I think it would be much clearer to future readers if there was something that consumed the `outputs` value from the render, and then looked to rename existing directories as well as mutate the target output path and we hand that off to `write_outputs` just as we do today. That rename function could have a nice and noticeable docstring that can communicate what's going on.

Hopefully this isn't too much of a detour but if I know myself I'll forget about this in a month and be just as confused for similar changes 😅

ericmcbride: Awesome, I will get on it!  And no problem.  Took me a while to grok it too.  I went down the exact same path as you so I know exactly the thought pattern you were going down  😅

crate_universe/src/utils.rs
```diff
@@ -16,3 +19,94 @@ pub(crate) fn sanitize_module_name(name: &str) -> String {
...
+/// `VendorGenerator::generate()` method.  This means that the semver metadata will
+/// always contain a (+) symbol, which is not compatible with bazel's labels.  
+/// This function will rename the cargo vendor generated file paths to be compatible with bazel
+/// labels by simply replacing the (+) with a (-). If this file is called by any other cli mod,
+/// it just simplies joins the out dir to the path
+pub(crate) fn normalize_cargo_file_paths(
+    outputs: BTreeMap<PathBuf, String>,
+    out_dir: &Path,
+) -> BTreeMap<PathBuf, String> {
+    outputs
```
ericmcbride, line 48: I moved the re mapping of  the outputs from write_outputs into this function, that way the iteration still only happens one time.  So it combines 
```
    let outputs: BTreeMap<PathBuf, String> = outputs
        .into_iter()
        .map(|(path, content)| (out_dir.join(path), content))
        .collect();
```
With the renaming, so we get both the join from  https://github.com/bazelbuild/rules_rust/pull/2579/files#diff-52bdc04b030a02035a19a3c805288b7720191185a18f6b8f16f61a59baa42d3fL787 and renaming of the files in one swoop..  

if nothing needs to be renamed, it simply just does the `out_dir` join, previously in the `write_outputs` function

UebelAndre (approved):
Awesome! Thank you so much!

PR #2584, opened 2024-03-29
Allow protobuf-backed rust_proto_library to be used with bzlmod.

scottpledger:
This is my attempt at fixing `rust_proto_library` for bzlmod, as part of https://github.com/bazelbuild/rules_rust/issues/2181.


illicitonion (commented):
Hello! Thanks for the PR!

This generally looks good, but I think a lot of it got merged in the similar #2601 - but either way I think having the example is super valuable - would you mind rebasing?

PR #2591, opened 2024-03-31
Added Rust 1.77.1

UebelAndre:
https://blog.rust-lang.org/2024/03/28/Rust-1.77.1.html


rust/private/common.bzl
```diff
@@ -31,7 +31,7 @@ load(":providers.bzl", "CrateGroupInfo", "CrateInfo", "DepInfo", "DepVariantInfo
 #
 # Note: Code in `.github/workflows/crate_universe.yaml` looks for this line, if
 # you remove it or change its format, you will also need to update that code.
-DEFAULT_RUST_VERSION = "1.77.0"
+DEFAULT_RUST_VERSION = "1.77.1"
 
 DEFAULT_NIGHTLY_ISO_DATE = "2024-03-21"
```
UebelAndre, line 36: Nightly is not being bumped due to https://github.com/rust-lang/rust/issues/123156

.bazelrc
```diff
@@ -28,6 +28,9 @@ common --experimental_cc_shared_library
 ## Unique configuration groups
 ###############################################################################
 
+# Enable use of the nightly toolchains by default.
```
daivinhtran, line 31: I don't see where we enable nightly toolchain by default though. Do we use `--config=nightly` by default anywhere?

UebelAndre: It's used by me every time I bump the Rust version 😅

daivinhtran: Can we reword the comment to clarify?

UebelAndre: Is this any less clarifying then the comments for clippy and rustfmt? While those are used in CI the comments don't refer to their use. Just what they do. Is this not following a similar pattern? Should they all be updated?

daivinhtran: IMO "by default" here is kinda misleading. It's not applied unless we run `bazel build --config=nightly`.

So unless we have

```
build (or common) --//rust/toolchain/channel=nightly
```

it's not really the default.

So just `enable use of the nightly toolchains` (without `by default`) is more accurate.

UebelAndre: Done!

UebelAndre:
@daivinhtran ready for another pass :)

daivinhtran (approved):

daivinhtran:
Sorry for the delay. I thought I approved it.

PR #2596, opened 2024-04-01
fix(cargo-bazel): ignore example crates when checking if proc-macro

### LINKED ISSUE #2577: crates_universe lockfile associating a package's examples as a proc_macro_deps
## What

When depending on certain crates, the bazel lockfile from crates_universe incorrectly associates a crate's dependency as a `proc_macro_dep` if the transitive dependency contains an `example`  that is also a crate type of `proc-macro`.

rules_rust: v0.40.0
bazel: v6.4.0

## Example Scenario

[deno-ops crate](https://github.com/denoland/deno_core/blob/79e73e500e409063cfbc165b58314d2c66cbf371/ops/Cargo.toml#L18) has a dependency on proc-macro-rules crate. The proc-macro-rules crate has an [example](https://github.com/nrc/proc-macro-rules/blob/ce363947a762d0667e1040c39b8596433026c594/Cargo.toml#L24-L26) that is `proc-macro`. This causes the `proc-macro-rules` crate to be added to both the `deps` and `proc_macro_deps` attributes, when it should only be added to `deps`.

e.g.
```sh
$ bazel query @crate_index__deno_ops-0.78.0//:all --output=build | grep deps
  deps = ["@crate_index__deno-proc-macro-rules-0.3.2//:deno_proc_macro_rules", "@crate_index__lazy-regex-2.5.0//:lazy_regex", "@crate_index__once_cell-1.19.0//:once_cell", "@crate_index__pmutil-0.6.1//:pmutil", "@crate_index__proc-macro-crate-1.3.1//:proc_macro_crate", "@crate_index__proc-macro2-1.0.79//:proc_macro2", "@crate_index__quote-1.0.35//:quote", "@crate_index__regex-1.10.4//:regex", "@crate_index__strum-0.25.0//:strum", "@crate_index__syn-1.0.109//:syn", "@crate_index__syn-2.0.55//:syn", "@crate_index__thiserror-1.0.58//:thiserror"],
  proc_macro_deps = ["@crate_index__deno-proc-macro-rules-0.3.2//:deno_proc_macro_rules", "@crate_index__strum_macros-0.25.3//:strum_macros"],
  ```
 
When building the deno-ops target, this is the error:
```
in rust_proc_macro rule @crate_index__deno_ops-0.78.0//:deno_ops: 
<traceback redacted>
Error in fail: @crate_index__deno_ops-0.78.0//:deno_ops listed @crate_index__deno-proc-macro-rules-0.3.2//:deno_proc_macro_rules in its proc_macro_deps, but it is not proc-macro, it is a rlib. It should probably instead be listed in deps.
```

This is because `deno_proc_macro_rules` is properly generated as a `rust_library`.

## Potential Fix 
 
After skimming through the rules_rust generator source code, I believe this following function is where we can benefit from an extra filter.
 https://github.com/bazelbuild/rules_rust/blob/fc601ba32f21ec034baebc487646dea92afbcd04/crate_universe/src/metadata/dependency.rs#L199-L204
 
We should only be looking for `proc-macro` targets that are not an `example`. 

With the following patch, the bazel lockfile was generated without adding the `proc-macro-rule` as a `proc_macro_dep` and the build succeeds.

```patch
diff --git a/crate_universe/src/metadata/dependency.rs b/crate_universe/src/metadata/dependency.rs
index 979719b8..7f82c860 100644
--- a/crate_universe/src/metadata/dependency.rs
+++ b/crate_universe/src/metadata/dependency.rs
@@ -197,10 +197,17 @@ fn is_lib_package(package: &Package) -> bool {
 }
 
 fn is_proc_macro_package(package: &Package) -> bool {
-    package
-        .targets
-        .iter()
-        .any(|target| target.crate_types.iter().any(|t| t == "proc-macro"))
+    package.targets.iter().any(|target| {
+        let is_proc_macro = target
+            .crate_types
+            .iter()
+            .any(|t| t == "proc-macro");
+        let is_example = target
+            .kind
+            .iter()
+            .any(|t| t == "example");
+        is_proc_macro && !is_example
+    })
 }
 
 fn is_dev_dependency(node_dep: &NodeDep) -> bool {
```

qtica:
# WHAT

* ignore example crate types when checking if it is a proc-macro

# WHY

When building the list of `deps` and `proc_macro_deps` for a given crate, 
dependency crates which contain examples in their Cargo.yaml of of type
`proc-macro` cause the dependency to be added to the `proc_macro_deps`. The
example crate types should be ignored.

Fixes #2577


qtica:
@UebelAndre would you mind taking a look at this PR for fixing the issue I opened the other day? 

I took at glance at the failing tests and they seem unrelated?

UebelAndre (approved):
Thank you! Has a test and everything! <3

Issue #2586, opened 2024-03-29
crates.io download URLs


Turbo87:
crates.io lately changed the way download requests are routed to reduce the load on their API servers  (see https://blog.rust-lang.org/2024/03/11/crates-io-download-changes.html). This was done via https://github.com/rust-lang/crates.io-index/commit/0f11dc08f3ac7acfc356ed2d4e2711e6c023566c, but it looks like Bazel has the URL patterns hardcoded at the moment (see https://github.com/bazelbuild/rules_rust/blob/0.41.0/crate_universe/src/metadata/metadata_annotation.rs#L276-L279) so these requests are still going through the crates.io API.

It would be great if either the hardcoded pattern could be adjusted for the new URLs, or ideally the code adjusted to use the `config.json` file of the package index.

PR #2597, opened 2024-04-01
Fix crates.io URL

ericmcbride:
- Resolves https://github.com/bazelbuild/rules_rust/issues/2586
- All other statements already use the correct CDN urls / git repos, based off of the `config.json`
- The hardcoded path seems to be based off a last ditch effort if theres not metadata spliced from the crate.  I think using the hardcoded path is fine here, due to the fact that its last ditch, and I do not think the crates download url will be changing often.


UebelAndre (approved):
Thank you!

PR #2598, opened 2024-04-01
Add support for `--compile_one_dependency`

william-smith-skydio:
The `--compile_one_dependency` flag ([docs](https://bazel.build/docs/user-manual#compile-one-dependency)) changes `bazel build` etc. to accept a file path and build the target corresponding to that file. This is useful for check-on-save with rust-analyzer in combination with the newly-added `$saved_file` command substitution (https://github.com/rust-lang/rust-analyzer/pull/15476).

Officially `--compile_one_dependency` only supports the builtin C++ and Java rules, but an [undocumented flag](https://github.com/bazelbuild/bazel/blob/7.1.1/src/main/java/com/google/devtools/build/lib/packages/Attribute.java#L102) can be added to attributes to turn them into sources supporting `--compile_one_dependency`. I'm not sure what the status of this support is, but it appears to work for all bazel versions up to at least 7.1.1, and if support is removed the flag is pretty harmless.

Before this change:
```
> bazel build --compile_one_dependency tools/rust_analyzer/main.rs
WARNING: Target pattern parsing failed.
ERROR: Couldn't find dependency on target '//tools/rust_analyzer:main.rs'
ERROR: Couldn't find dependency on target '//tools/rust_analyzer:main.rs'
INFO: Elapsed time: 0.956s
INFO: 0 processes.
ERROR: Build did NOT complete successfully
```

After:
```
> bazel build --compile_one_dependency tools/rust_analyzer/main.rs
INFO: Analyzed target //tools/rust_analyzer:gen_rust_project (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //tools/rust_analyzer:gen_rust_project up-to-date:
  bazel-bin/tools/rust_analyzer/gen_rust_project
INFO: Elapsed time: 0.341s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
```


illicitonion:
This looks reasonable - do you think there's a reasonable way to add a test for this behaviour?

william-smith-skydio:
I added a test, which I think should work. I'm not 100% sure since I haven't worked with the bazel CI system before.

illicitonion (approved):
Thanks!

PR #2600, opened 2024-04-02
Fixed bug where crate_universe could match aliases to bench/example deps

UebelAndre:
A dependency which aliases a transitive dependency with a name that happens to match another Cargo target in that (Cargo) package will result in an incorrect dependency being collected. For the example of [surrealdb-core](https://github.com/surrealdb/surrealdb/blob/v1.3.1/core/Cargo.toml#L83), the `executor = { package "async-executor" }` dependency was incorrectly matching [a benchmark target](https://github.com/smol-rs/async-executor/blob/v1.9.1/Cargo.toml#L38-L40), thus causing a broken dep to be rendered. This PR fixes this issue.


UebelAndre:
This change also contains https://github.com/bazelbuild/rules_rust/pull/2599 which should be reviewed separately to reduce scope from this change.

illicitonion (approved):

PR #2601, opened 2024-04-03
Support building more things with bzlmod

matts1:
This allows us to build more things with bzlmod with `WORKSPACE.bazel` disabled. This is required because when you use it as a module, it can't load dependencies from `WORKSPACE.bazel`.

`bazel build --enable_bzlmod --noenable_workspace --lockfile_mode=off //... --keep_going`

Before: Analysis succeeded for 937 of 978 top-level targets
After: Analysis succeeded for 978 of 982 top-level targets

Once we get those remaining targets working, we can then change our CI to add `--noenable_workspace` to our bzlmod configuration.


proto/protobuf/repositories.bzl
```diff
@@ -18,34 +18,35 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
...
-        http_archive,
-        name = "rules_proto",
-        sha256 = "dc3fb206a2cb3441b485eb1e423165b231235a1ea9b031b4433cf7bc1fa460dd",
-        strip_prefix = "rules_proto-5.3.0-21.7",
-        urls = [
-            "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/refs/tags/5.3.0-21.7.tar.gz",
-            "https://github.com/bazelbuild/rules_proto/archive/refs/tags/5.3.0-21.7.tar.gz",
-        ],
-    )
+def rust_proto_protobuf_dependencies(bzlmod = False):
```
illicitonion, line 21: Nothing seems to call this with `bzlmod = True` - why do we need to add it?

matts1: Good catch, thanks. It should have been called with bzlmod=True in the internal deps.

illicitonion (approved):
Thanks!

PR #2605, opened 2024-04-05
rust_bindgen: Allow specifying multiple headers

ParkMyCar:
The Rust Bindgen tool allows specifying multiple header files, but the `rust_bindgen` bazel rule does not. This PR updates the `rust_bindgen` bazel rule to take a list of files by adding a new `"headers"` attribute, and marking the previous `"header"` attribute as deprecated.

Note: I'm pretty sure I need to re-gen the user facing documentation by I couldn't figure out how to.


ParkMyCar:
Spoke too soon! The Bindgen library supports specifying multiple headers, but the cli tool does not

Issue #2595, opened 2024-04-01
`default_features = False` not respected in bzlmod 


firekind:
Hello,
I have a non cargo setup via bzlmod. When i set `default_features = False` in `crate.spec` for a crate (in this case, the `ort` crate), the `rust_library` target generated in the crate's BUILD file still has the `crate_features` attribute set. This does not seem to happen if I use WORKSPACE instead of bzlmod.

firekind:
here's a simple [repro](https://github.com/firekind/2595-repro)

firekind:
This was fixed by #2606, so closing.

PR #2606, opened 2024-04-07
Keep default_features parity from bzlmod to workspace

Lev1ty:
This PR contains two changes:
1. Set the default value of `crate.spec` `default_features` to `True` as `WORKSPACE` handles it.
2. Prevent filtering out `default_features` during package json encoding as both truthy and falsy values for booleans are significant.

### DIFF
--- a/crate_universe/extension.bzl
+++ b/crate_universe/extension.bzl
@@ -224,7 +224,7 @@ def _package_to_json(p):
     return json.encode({
         k: v
         for k, v in structs.to_dict(p).items()
-        if v
+        if v or type(v) == "bool"
     })
 
 def _get_generator(module_ctx):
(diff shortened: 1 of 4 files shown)


Lev1ty:
Targeted at fixing #2595

illicitonion (commented):
Thanks for your contribution! This PR looks good to me, but can you add a test so we don't regress this behaviour in the future? Maybe add another dep to https://github.com/bazelbuild/rules_rust/tree/main/examples/bzlmod/hello_world_no_cargo, and write a small rust file which consumes the dep and would fail to compile if default features weren't enabled?

Lev1ty:
Glad I could contribute to OSS I depend on :) Added the test, let me know any follow-ups too

illicitonion (approved):
Perfect, thanks so much!

Issue #1950, opened 2023-04-29
Support conditional compilation dependencies


tthebst:
### Discussed in https://github.com/bazelbuild/rules_rust/discussions/1947

<div type='discussions-op-text'>

<sup>Originally posted by **tthebst** April 28, 2023</sup>
I'm trying to compile tokio with `--cfg tokio_unstable` with rust rules. This works fine by adding it in the `crate.annotations` rustflag options. It also seems to get picked up by when compiling because I see that the compiler invocations have (`bazel build -s --config local @crate_index//:tokio`). But when compiling I get an error `use of undeclared crate or module tracing` indicating that the tracing [dependency](https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml#L116) was not added when compiling. In the bazel lock file I see this:

```
"deps": {
 ...
  "selects" :{
      ....
      "cfg(tokio_unstable)": [
          {
            "id": "tracing 0.1.38",
            "target": "tracing"
           }
       ],
   }
}
```

How can I instruct bazel to add the tracing dependency when compiling?

I will try to come up with a small repro for this. </div>

UebelAndre:
From https://github.com/bazelbuild/rules_rust/discussions/1947#discussioncomment-5755216
> I don't believe `crate_universe` knows anything about configurations when it resolves dependencies. So the select statement can never be matched. Similar to [@rules_rust//crate_universe/src/context/platforms.rs](https://github.com/bazelbuild/rules_rust/blob/0.21.1/crate_universe/src/context/platforms.rs#L12-L103), I would expect a `resolve_cfg_rustc_flags` that goes through annotations and finds `cfg` flags that can be used to enable or disable features. I don't think there will be a way to dynamically enable and disable the dependencies per bazel invocation. Would you be willing to convert this to an Issue? I'd be happy to collaborate on a pull request as well if you'd be open to working on this feature 😄

ericmcbride:
Im gonna try to knock this out.  I have a use case for this, for tokio console.

PR #2607, opened 2024-04-08
DRAFT: Conditional compilation deps

ericmcbride:
- addresses https://github.com/bazelbuild/rules_rust/issues/1950
- This is a WIP and i want more input before I continue.  This takes care of the `cfg_expr` use case of  the `Flag` predicate.  AKA `cfg(tokio_unstable)` or `cfg(tokio_dump)`
- Added a new bzl mod example, to show this base use case is fufilled . 
- This takes care of the rustc_flags use case of `tokio_unstable` or `tokio_coredump`
- @UebelAndre I would like your input on this approach before I take care of any of the other predicate use cases in `cfg_expr`  Some of my thoughts are, do we need to add metadata to CrateDependency to say we injected this value?   Could this be leveraged in the future for optional dependencies so they dont have to be declared explicitly in the library?  What were you looking to resolve in 1950?
- After the questions / vision of this feature is resolved, will add full documentation + unit tests + more conditional cfg types in the bzl mod example.


kyleconroy:
I'd also love to be able to use tokio-console. Anything I can do to help this along?

ericmcbride:
I just need some feedback on this approach.  Im about to pick up another issue thats more pressing for me this week and i can circle back around to this one

PR #2618, opened 2024-04-18
Add aspect_rules_js to MODULE.bazel

dzbarsky:
Otherwise wasm_bindgen with rules_js is unusable.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -39,6 +39,10 @@ bazel_dep(
     version = "21.7",
     repo_name = "com_google_protobuf",
 )
+bazel_dep(
+    name = "aspect_rules_js",
+    version = "1.39.0",
+)
 
 bazel_dep(
     name = "rules_python",


UebelAndre:
This feels like a very heavy dependency. Is there no way for users to opt in on their own? I feel the non-bzlmod workflow was quite nice as it didn’t force big dependencies upon devs that didn’t need them.

dzbarsky:
There's no way currently to inject in outside deps. However, modules are lazily loaded so this should be fairly cheap. Fmeums recommendation: https://bazelbuild.slack.com/archives/C014RARENH0/p1713461860883669?thread_ts=1713457638.539409&cid=C014RARENH0

illicitonion (approved):

illicitonion:
Works for me! Can you buildifier? Thanks!

dzbarsky:
@illicitonion done, thanks!

PR #2619, opened 2024-04-18
Substitute `${SDKROOT}` and `${DEVELOPER_DIR}` by their environment variables on darwin.

thb-sb:
Substitute `${SDKROOT}` and `${DEVELOPER_DIR}` by their environment variables on darwin.

This PR attempts to fix an issue with bindgen when used on darwin with clang
compiled from source (i.e. from `@llvm-project`).

When using clang compiled from source on darwin, libclang (used through `clang-sys`) failed
to find system header files (such as `stdarg.h` or `stdbool.h`) because these
files are actually living under the SDK path (under `<SDK_PATH>/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/` and `<SDK_PATH>/usr/include`).

Similar to what Bazel did in their [`xcrunwrapper.sh`](https://github.com/bazelbuild/bazel/blob/d8c27bfcd37a74dfbf1bdb9a1e3df13af8360a01/tools/objc/xcrunwrapper.sh#L17), we must have a way to tell bindgen's libclang
where to find these system headers.

In this commit, I add the following two substitutes variables:
  - `${SDKROOT}`: replaced by `std::env::var("SDKROOT")` if found.
  - `${DEVELOPER_DIR}`: replaced by `std::env::var("DEVELOPER")` if found.

This is similar to what it is currently done for `${pwd}`.

The following rule should now work on darwin:

```c
// mylib.h
#include <stdarg.h>
```

```starlark
# BUILD
cc_library(
    name = "mylib",
    srcs = ["mylib.c"],
    hdrs = ["mylib.h"],
)

rust_bindgen_library(
    name = "mylib_bindgen",
    cc_lib = ":mylib",
    clang_flags = select({
        "@platforms//os:macos": [
            "--sysroot=${SDKROOT}",
            "-isystem${SDKROOT}/usr/include",
            "-isystem${SDKROOT}/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/",
        ],
        "//conditions:default": [],
    }),
    crate_name = "mylib",
    header = "mylib.h",
)
```

It may also fix #1834.

### DIFF
--- a/bindgen/private/bindgen.bzl
+++ b/bindgen/private/bindgen.bzl
@@ -198,6 +198,22 @@ def _rust_bindgen_impl(ctx):
 
     args = ctx.actions.args()
 
+    executable = bindgen_bin
+
+    # When running on darwin, we have to forward the SDKROOT environment variable,
+    # otherwise some system include files may not be found if the clang compiled
+    # from source is being used.
+    # Since the sdk directory given by the apple toolchain[1] may be a dummy
+    # variable, we have to use the process_wrapper.
+    #
+    # [1]: https://bazel.build/rules/lib/builtins/apple_toolchain.html#sdk_dir
+    if apple_common != None:
+        args.add_all(["--subst", "SDKROOT=${SDKROOT}"])
+        args.add_all(["--subst", "DEVELOPER_DIR=${DEVELOPER_DIR}"])
+        args.add_all(["--", bindgen_bin])
+        tools = depset([bindgen_bin], transitive = [tools])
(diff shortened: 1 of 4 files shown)


illicitonion (commented):
This seems pretty reasonable, thanks! Is it possible to add in a test which would fail without this change?

UebelAndre (changes requested):
https://github.com/bazelbuild/rules_rust/pull/2619#pullrequestreview-2010931917

PR #2621, opened 2024-04-19
Add Apple visionOS support

QuentinPerez:
👋 

A small PR to support VisionOS. As it's a tier 3 toolchain, you'll need to build and host the standard library. Below is an example demonstrating the use of the `rust_repository_set` rule to configure the toolchain.

```
rust_repository_set(
    name = "aarch64_apple_visionos_on_aarch64_apple_darwin",
    edition = "2024",
    exec_triple = "aarch64-apple-darwin",
    extra_target_triples = {
        "aarch64-apple-visionos": [
            "@platforms//os:visionos",
            "@platforms//cpu:arm64",
            "@build_bazel_apple_support//constraints:apple",
            "@build_bazel_apple_support//constraints:device",
        ],
        "aarch64-apple-visionos-sim": [
            "@platforms//os:visionos",
            "@platforms//cpu:arm64",
            "@build_bazel_apple_support//constraints:apple",
            "@build_bazel_apple_support//constraints:simulator",
        ],
    },
    urls = ["URLS_TO_YOUR_RUST_STD"],
)
```

### DIFF
--- a/crate_universe/src/context/platforms.rs
+++ b/crate_universe/src/context/platforms.rs
@@ -1,8 +1,44 @@
+use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 
 use anyhow::{anyhow, Context, Result};
-use cfg_expr::targets::{get_builtin_target_by_triple, TargetInfo};
+use cfg_expr::targets::Abi;
+use cfg_expr::targets::{
+    get_builtin_target_by_triple, Arch, Endian, Families, HasAtomics, Os, Panic, TargetInfo,
+    Triple, Vendor,
+};
 use cfg_expr::{Expression, Predicate};
+
+// since we use binary_search_by on it, the array must be sorted.
+// NOTE: can be removed once, the visionos target is added to cfg_expr
+pub const TARGET_INFO: &[TargetInfo] = &[
+    TargetInfo {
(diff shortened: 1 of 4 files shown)


QuentinPerez:
can we merge it ?

UebelAndre:
It looks like there’s some build failures that will need to be addressed first

QuentinPerez:
> It looks like there’s some build failures that will need to be addressed first

You're right, I missed that one! I think we should not be able to merge it as long as the target isn't available on stable.

```
Caused by:
  process didn't exit successfully: `/private/var/tmp/_bazel_quentinperez/5845dd8b2f37c7097e5eb050b26837b8/external/rust_darwin_aarch64__aarch64-apple-darwin__stable_tools/bin/rustc - --crate-name ___ --print=file-names --target aarch64-apple-visionos --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=split-debuginfo --print=crate-name --print=cfg` (exit status: 1)
  --- stderr
  error: Error loading target specification: Could not find specification for target "aarch64-apple-visionos". Run `rustc --print target-list` for a list of built-in targets

```

QuentinPerez:
I am not sure to understand why the blzmod BCR step fails only on window

PR #2623, opened 2024-04-20
Support loading http credentials from netrc

golovasteek:
Bazel's `http_archive` and `http_file` use `~/.netrc` file to load HTTP credentials. 

Now rules_rust use the same approach if no other authorization parameters are provided.

NOTE: `http_archive` supports passing a custom `netrc` and `auth_patterns`. But this parameters are not added to the rust rules at the moment.


krasimirgg:
Thank you!

>NOTE: http_archive supports passing a custom netrc and auth_patterns. But this parameters are not added to the rust rules at the moment.

Any reason why not? Naively, a big chunk of that new `get_auth` function is to handle those and we'll need to add them at some point anyways, so let's just add them now (like they did in Python, https://github.com/bazelbuild/rules_python/pull/1417).

rust/private/repository_utils.bzl
```diff
@@ -804,10 +810,39 @@ def load_arbitrary_tool(
 
     return {archive_path: sha256}
 
+# The function is copied from the main branch of bazel_tools.
+# It should become available there from version 8.0.0,
+# We should remove this function when we upgrade minimum supported
+# version to 8.0.0.
```
illicitonion, line 816: This got back-ported: https://github.com/bazelbuild/bazel/pull/20915

```suggestion
# It should become available there from version 7.1.0,
# We should remove this function when we upgrade minimum supported
# version to 7.1.0.
```

illicitonion (approved):
Works for me! Will leave @krasimirgg to discuss extending support :)

golovasteek:
> Thank you!
> 
> > NOTE: http_archive supports passing a custom netrc and auth_patterns. But this parameters are not added to the rust rules at the moment.
> 
> Any reason why not? Naively, a big chunk of that new `get_auth` function is to handle those and we'll need to add them at some point anyways, so let's just add them now (like they did in Python, [bazelbuild/rules_python#1417](https://github.com/bazelbuild/rules_python/pull/1417)).

I've basically followed the YAGNI rule. I though it would be straight forward to add and test these parameters, once they are need.

If you say it is valuable to add it.

rust/repositories.bzl
```diff
@@ -639,6 +651,12 @@ _RUST_ANALYZER_TOOLCHAIN_TOOLS_REPOSITORY_ATTRS = {
             "See [repository_ctx.download](https://docs.bazel.build/versions/main/skylark/lib/repository_ctx.html#download) for more details."
         ),
     ),
+    "netrc": attr.string(
```
illicitonion, line 654: I think buildifier is failing because these keys aren't sorted?

golovasteek: Fixed.
BTW, somehow I didn't manage to make buildifier to complain about it locally. Is there some specific configuration or something like that?

PR #2624, opened 2024-04-22
Register a default rust toolchain.

matts1:
See https://github.com/bazelbuild/bazel/discussions/22024 for details.


matts1:
Oh right, I forgot to unmark it as draft. It's ready to merge, thanks.

PR #2625, opened 2024-04-22
Add clippy_flag flag to allow flags to be added in succession.

granaghan:
This helps keep my bazelrc cleaner since I can pass clippy flags on multiple lines with comments about why they're being added.


illicitonion (approved):
Thanks!

PR #2631, opened 2024-04-27
also rewrite -isystem in addition to -sysroot

adrianimboden:
I build in a hermetic c++ toolchain and use `-isystem` args in addition to `-sysroot` which did not get rewritten before.

this is one of the errors I get (indirect dependency when using `aspect_rules_py`):
```
cargo:warning=ToolExecError: Command "/tmp/bazel-working-directory/_main/external/local_config_cc/bin/clang" ... "-nostdinc" "-isystem" "external/local_config_cc/cxxlib/fastbuild/include/clang" "-isystem" "external/local_config_cc/cxxlib/fastbuild/include/x86_64-linux-gnu" "-isystem" "external/local_config_cc/cxxlib/fastbuild/include" "--sysroot=/tmp/bazel-working-directory/_main/external/local_config_cc/cxxlib/fastbuild" ... "-o" "/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/external/rules_rust~~crate~crate_index__bzip2-sys-0.1.11-1.0.8/bzip2-sys_bs.out_dir/lib/bzip2-1.0.8/decompress.o" "-c" "bzip2-1.0.8/decompress.c"
cargo:warning=In file included from bzip2-1.0.8/decompress.c:22:
cargo:warning=bzip2-1.0.8/bzlib_private.h:25:10: fatal error: 'stdlib.h' file not found
cargo:warning=#include <stdlib.h>
cargo:warning=         ^~~~~~~~~~
cargo:warning=1 error generated.
```

I think it is unfortunate and probably brings more weird errors up in the future when the build command does not run in the same place where every other bazel rule runs the compiler.


cargo/private/cargo_build_script.bzl
```diff
@@ -79,6 +79,30 @@ def _pwd_flags(args):
...
+        args (list): List of tool arguments.
+
+    Returns:
+        list: The modified argument list.
+    """
+    res = []
+    fix_next_arg = False
+    for arg in args:
+        if fix_next_arg:
+            res.append("${{pwd}}/{}".format(arg))
```
illicitonion, line 95: We should only do this is the next arg is a relative path - could you throw in a call to https://github.com/bazelbuild/bazel-skylib/blob/2b546aff9e92966ac5276e2b747aaf9f02690371/lib/paths.bzl#L236 in the `if` guarding this?

illicitonion (commented):
LGTM with one comment - thanks for the contribution!

cargo/private/cargo_build_script.bzl
adrianimboden: you are absolutely right. I missed that, sorry.

Issue #2626, opened 2024-04-23
How to deal with crates that depend on cmake on macos


johnynek:
I am trying to use the zip crate. When I try to build after repinning and adding this dependency I get this error:

https://gist.github.com/johnynek/c53ea36910352d607eb6bf19612db3d9

Note, it has portions like:
```
CMAKE_TOOLCHAIN_FILE_aarch64-apple-darwin = None
CMAKE_TOOLCHAIN_FILE_aarch64_apple_darwin = None
HOST_CMAKE_TOOLCHAIN_FILE = None
CMAKE_TOOLCHAIN_FILE = None
CMAKE_GENERATOR_aarch64-apple-darwin = None
CMAKE_GENERATOR_aarch64_apple_darwin = None
HOST_CMAKE_GENERATOR = None
CMAKE_GENERATOR = None
cargo:warning=xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation
cargo:warning=xcrun: error: unable to lookup item 'PlatformVersion' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
cargo:warning=xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation
cargo:warning=xcrun: error: unable to lookup item 'PlatformVersion' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
cargo:warning=xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation
cargo:warning=xcrun: error: unable to lookup item 'PlatformVersion' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
CMAKE_PREFIX_PATH_aarch64-apple-darwin = None
CMAKE_PREFIX_PATH_aarch64_apple_darwin = None
HOST_CMAKE_PREFIX_PATH = None
CMAKE_PREFIX_PATH = None
CMAKE_aarch64-apple-darwin = None
CMAKE_aarch64_apple_darwin = None
HOST_CMAKE = None
CMAKE = None
```

So it seems to be saying that it can't find cmake. That makes sense, I didn't have cmake installed. But then I did `brew install cmake` and get the exact same error.

Finally I did:

```
sudo ln -s /opt/homebrew/bin/cmake /usr/local/bin/
```
and then reran the `bazel build` and it worked.

[...]

PR #2635, opened 2024-04-30
Add docs and examples of complicated build scripts

illicitonion:
Fixes https://github.com/bazelbuild/rules_rust/issues/2626


cargo/private/cargo_build_script.bzl
```diff
@@ -315,6 +315,8 @@ def _cargo_build_script_impl(ctx):
         progress_message = "Running Cargo build script {}".format(pkg_name),
         env = env,
         toolchain = None,
+        # Set use_default_shell_env so that $PATH is set, as tools like cmake may want to probe $PATH for helper tools.
+        use_default_shell_env = True,
```
UebelAndre, line 319: I always struggle to balance convenience with correctness. I think it's more correct to omit this value but agree it's an inconvenience. If CMake is a known tool that requires `PATH` then maybe instead the rules_foreign_cc CMake toolchain can provide that?

UebelAndre (changes requested):

cargo/private/cargo_build_script.bzl
illicitonion: > maybe instead the rules_foreign_cc CMake toolchain can provide that?

Do you have any suggestions for how it could provide it? I guess it would need to run an action with `use_default_shell_env` enabled, capture `$PATH` as an output file, and write a wrapper script around its `cmake` binary which sets `$PATH` as per the previous action? Which feels very fragile and error-prone, and not necessarily better than this...

Alternatively, we could do something like https://github.com/bazelbuild/rules_rust/pull/2638 - we already special-case `$AR` in the build script runner, so this doesn't feel _that_ far down the slippery slope, but it's definitely ugly, but at least it's reasonably scope-limited (just macOS, just if you don't specify better values for any of the things it sets). If you'd prefer, I can write some tests for that draft PR, and remove the `use_default_shell_env` here?

(My _real_ feature request here is for the Bazel C++ toolchain to have a way of finding a usable `ar`, rather than a tool which happens to be able to do the same thing. I _suppose_ that could be a thing rules_foreign_cc provides via a `preinstalled_ar_toolchain` or something, but... That's a lot of work to go through as a user to actually wire everything up properly to use that toolchain).

UebelAndre: I have no great solutions. Maybe `rules_foreign_cc` can provide a toolchain with some make variable for `PATH`? That aside though, I find piece of mind when actions do not have `use_default_shell_env` enabled so I know `PATH` will never get added and what I see on the action implementation is what I get. I feel this philosophy is unpopular with the community though so I'm not gonna block the PR for it. But if this were to be a config setting developers could disable, I'd certainly be a happy user of it 😄

UebelAndre (approved):

Issue #2212, opened 2023-10-25
Inconsistent cargo bazel lockfile between MacOS and Linux


jvliwanag:
Given a rust project with the dependency: `pbjson_types`, the generated cargo bazel lockfile differs when generated on Mac and Linux.

See sample project [here](https://github.com/jvliwanag/repro-cargo-bazel-lock-issue). The diff switching from the mac generated one to the linux one is on this [commit](https://github.com/jvliwanag/repro-cargo-bazel-lock-issue/commit/665303fb7271ac10d4232a1a14907899cb484cd1).

Mac build is from an M2 Sonoma. Linux build is from a docker ubuntu arm64 image.

gferon:
I encountered the exact same issue in the past few weeks. It is related to a (current) hard limitation of `cargo metadata` which does not expose features per target platform. Because of that, `rules_rust` relies on `cargo tree` to determine which features are enabled for each platform.

The problem with this approach is that `cargo tree` will _always_ compute build dependencies for the host platform, which in your case can either be macOS or Linux.

One quick & dirty solution could be to add a `--host-triple` to `cargo tree` to be able to list `build-dependencies` for each platform Bazel is going to build on _and_ for. I [started a discussion with the `cargo` maintainers](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/cargo.20tree.20--host-triple) which pointed me towards `cargo metadata`.

It looks like we could improve `cargo_universe` a lot if `cargo metadata` would be able to use `resolver = 2` but this sounds complicated, see: https://rust-lang.github.io/rfcs/2957-cargo-features2.html#cargo-metadata

This whole story was already discovered by @illicitonion a while ago in https://github.com/bazelbuild/rules_rust/issues/1662 - could you maybe give your two cents to help move in one direction or another?

Some extra relevant links:
- https://github.com/rust-lang/cargo/issues/9863
- https://github.com/rust-lang/rust-analyzer/issues/15233

jvliwanag:
Thanks for looking into this. As this seems blocked though by changes we’d need on cargo, what would be viable options to workaround this issue for the meantime?

illicitonion:
> One quick & dirty solution could be to add a `--host-triple` to `cargo tree` to be able to list `build-dependencies` for each platform Bazel is going to build on _and_ for.

[...]

PR #2636, opened 2024-05-01
Use `cargo tree` to determine `resolver = 2` dep trees

UebelAndre:


UebelAndre:
I recommend reviewers take this change one commit at a time. There are two very noisy commits with a bunch of rendered output.

UebelAndre:
I also don't think this closes https://github.com/bazelbuild/rules_rust/issues/2212 since if I understand the issue correctly, `[build-dependencies]` are always evaluated by the current host triple so if there are any platform specific `[build-dependencies]` then there's no current way to get `cargo tree` to resolve those for a different platform than the current platform.

illicitonion (approved):
Very nice! Thank you!!

PR #2643, opened 2024-05-07
Fix cc_common_link when using sibling repository layout

yuzhy8701:
Linking with cc_common is broken for external repositories if you also specify `--experimental_sibling_repository_layout`. 

The rule would complain `The package dir path should be a prefix of the crate_info.output.path`. It happens because the package path derived from `bin_dir`, `workspace_root` and `package` did not match how sibling layout handles external repositories.

This change ignores the `workspace_root` component if the path signifies the usage of sibling layout, as it is not needed.


scentini (approved):
Thanks!

PR #2646, opened 2024-05-09
Support protoc-gen-prost-serde in prost_toolchain

mrmeku:
protoc-gen-prost-serde is a protoc plugin that generates [serde](https://serde.rs/) serialization implementations that follow the conventions of protobuf-JSON. Prost does not have JSON bindings itself, so this plugin adds them. Since it is augmenting the structs generated by prost itself, it needs to be part of the same crate which is why I've added it to the existing prost_toolchain

### DIFF
--- a/docs/flatten.md
+++ b/docs/flatten.md
@@ -830,7 +830,8 @@ Builds a Rust proc-macro crate.
 ## rust_prost_toolchain
 
 <pre>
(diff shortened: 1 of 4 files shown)


illicitonion (changes requested):
This generally looks reasonable, thanks! 

Two questions:
1. Can you add a test or example which shows this working, so that we don't break it in the future?
2. Why is the rustfmt stuff getting removed? It feels like it should probably still stay?

PR #2647, opened 2024-05-10
Support alwayslink for linking with clang

### LINKED ISSUE #1271: Support alwayslink for linking with clang?
When linking rust libraries into a final C/C++ binary, sometimes you may need a library to be force loaded in order to include symbols that may otherwise not be referenced, for example if you're interfacing with the JNI. This can be controlled in bazel by passing `alwayslink = True/False` to `cc_common.create_library_to_link`. Should the rules expose a `alwayslink` attribute, similar to `cc_library` and others, to control this? I'm happy to submit this if so. I'm not sure how this interacts with https://github.com/bazelbuild/rules_rust/issues/637 and https://github.com/bazelbuild/rules_rust/issues/1268 since they mostly seem to focus on binaries being linked with `rustc`, and not `clang`

cerisier:
Simple approach to fix #1271


cerisier:
Thanks for having a look at this @scentini
Just applied your suggestions.

scentini:
Thank you for your contribution! CI says you also need to run `./docs/update_docs.sh` before we can merge.

archshift:
Any progress on this? It would be really great to see!

cerisier:
@scentini up to date and docs generated.

scentini (approved):

scentini:
Thanks @cerisier !

PR #2648, opened 2024-05-10
Add target for running rustfmt directly

illicitonion:
Previously you either needed to go via our bazel-aware wrapper, or write a rule to grab the file from the toolchain.

### DIFF
--- a/tools/rustfmt/src/upstream_rustfmt_wrapper.rs
+++ b/tools/rustfmt/src/upstream_rustfmt_wrapper.rs
@@ -1,0 +1,38 @@
+use std::path::PathBuf;
+use std::process::{exit, Command};
+
+fn main() {
+    let runfiles = runfiles::Runfiles::create().unwrap();
+
+    let rustfmt = runfiles::rlocation!(runfiles, env!("RUSTFMT"));
+    if !rustfmt.exists() {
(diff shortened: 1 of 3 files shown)


tools/rustfmt/src/upstream_rustfmt_wrapper.rs
```diff
@@ -0,0 +1,31 @@
...
+
+fn main() {
+    let runfiles = runfiles::Runfiles::create().unwrap();
+
+    let rustfmt = runfiles::rlocation!(runfiles, env!("RUSTFMT"));
+    if !rustfmt.exists() {
+        panic!("rustfmt does not exist at: {}", rustfmt.display());
+    }
+
+    let working_directory = std::env::var_os("BUILD_WORKSPACE_DIRECTORY")
```
UebelAndre, line 12: isn't there:
```suggestion
    let working_directory = std::env::var_os("BUILD_WORKING_DIRECTORY")
```
?

UebelAndre (approved):

tools/rustfmt/src/upstream_rustfmt_wrapper.rs
illicitonion: Ooh yes, even better, good call!

PR #2651, opened 2024-05-15
Allow runfiles object to be an arbitrary expression.

matts1:
Our use case is `runfiles::rlocation!(cache.runfiles, "foo")` but this would also allow `runfiles::rlocation!(Runfiles::create()?, "foo")`


DavidZbarsky-at (approved):
makes sense! Sorry I am a rust macro noob :)

matts1:
No worries, me too. I reviewed your PR and didn't spot it either.

illicitonion (approved):

PR #2652, opened 2024-05-15
Allow searching for the rlocation of a module.

matts1:
Currently, if the repo mapping maps "foo" to "real_foo", then:
* `rlocation!(r, "foo/bar/baz")` will return "real_foo/bar/baz"
* `rlocation!(r, "foo")` will return `foo`

This PR handles the case where the user attempts to do something like `rlocation!("rules_rust")` (which is valid for directory based runfiles).


illicitonion (approved):
Thanks!

PR #2654, opened 2024-05-15
Add support for generated sources to rust-analyzer.

matts1:
At the moment, rust_analyzer's rust-project.json will list the root of a proto crate as, for example: `$HOME/.cache/bazel/_bazel_msta/60654fdd7bbd86377eab8595eb8577e8/execroot/_main/bazel-out/k8-dbg/bin/examples/prost/proto/example_proto.lib.rs`. However, this file does not exist, since it wasn't actually declared as the output of any rule.

With this change, tools/rust-analyzer will actually generate the generated sources.

### DIFF
--- a/proto/prost/private/prost.bzl
+++ b/proto/prost/private/prost.bzl
@@ -275,6 +275,7 @@ def _rust_prost_aspect_impl(target, ctx):
             package_info = package_info_file,
         ),
         rust_analyzer_info,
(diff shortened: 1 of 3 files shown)


illicitonion:
Oh, looks like CI is failing?

matts1:
Whoops, should be fixed now.

PR #2656, opened 2024-05-15
Default linux abi to musl not gnu

illicitonion:
The gnu-abi build cargo-bazel doesn't run in older linux distributions (e.g. centos7) because its libc implementation is too old.

This defaults to musl-linked binaries which don't link libc, so is more likely to be compatible with the execution environment.


illicitonion:
I think this showed up because upgrading cross in #2645 appears to now link against a newer glibc, but I haven't dug super deep into it.

PR #2660, opened 2024-05-17
fix: default rustfmt version to supplied rust version

mattem:
Modifies the default of `rustfmt_version` so that it uses the first version in the `versions` list. If there are multiple versions given then falls back to the original default of `DEFAULT_NIGHTLY_VERSION`.

We tripped over this where supply a single Rust version (non-nightly), but started using nightly features in `rustfmt.toml` and could no longer format directly with `cargo fmt`. This helped align the versions of both toolchains.


illicitonion (approved):
Thanks!

Issue #2520, opened 2024-02-26
Long path issues on Windows with bzlmod


criemen:
> For the prefix - you're right; I believe that we could still make it work, by adding a `crate_name` attribute to the `cargo_build_script` rule and making it mandatory. Normally this would be a breaking change, but we don't directly use the `cargo_build_script` rule, we have a [wrapper macro](https://github.com/bazelbuild/rules_rust/blob/53daac71c0755680c29f4c55ac5238cc063e8b31/cargo/private/cargo_build_script_wrapper.bzl#L11), where we could pass `crate_name` to the underlying `_build_script_run` target.

_Originally posted by @scentini in https://github.com/bazelbuild/rules_rust/issues/2454#issuecomment-1961099787_

As discussed in #2454, there's a possibility of shortening the target name of the cargo build script that's auto-generated for crates in `crate-universe` (ideally to only `b` or `bs`).
This is necessary for a dependency of mine, where we run into long path issues with constructions like
```
    SET CARGO_MANIFEST_DIR=bazel-out/x64_windows-opt-exec-ST-13d3ddad9198/bin/external/rules_rust~override~crate~r__tree-sitter-embedded-template-0.20.0/tree-sitter-embedded-template_bs_.exe.runfiles/rules_rust~override~crate~r__tree-sitter-embedded-template-0.20.0/
```
This fails with long path issues, even though the bazel output base is already on `C:/tmp`, and the repository name is `r` to be as short as possible. There's not much I can do if `rules_rust` repeats the (very long) dependency name three times, while having the whole repository name twice in there.

If you want to try yourself, plug
```
tree-sitter-embedded-template = { git = "https://github.com/tree-sitter/tree-sitter-embedded-template.git", rev = "203f7bd3c1bbfbd98fc19add4b8fcb213c059205" }
```
into your `Cargo.toml`, and try compiling your project on latest `rules_rust` and `bzlmod` on Windows.

criemen:
With 

```
common --nolegacy_external_runfiles
common --experimental_sibling_repository_layout
```
in `.bazelrc`, it's possible to get rid of the `external` directory, and turn the quoted path above into

```
[...]

PR #2663, opened 2024-05-23
crate_universe: Don't include crate name in build script path.

criemen:
This PR shortens the path to the build script (potentially significantly), which helps with long-path issues on Windows.
This change is motivated by the crate `tree-sitter-embedded-template`, which has a too long path to the compiled build script otherwise, and which then fails to build on Windows in a `bzlmod` setting.
Fixes https://github.com/bazelbuild/rules_rust/issues/2520.

The cargo build script name was used to automatically derive the cargo package name, this now went into a separate override parameter that is generated by the crate-universe tooling.
There's never two build scripts in a single crate, so I don't see how having a single build script target with a single name would be a problem.

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -151,7 +151,9 @@ def _cargo_build_script_impl(ctx):
         stderr = ctx.actions.declare_file(ctx.label.name + ".stderr.log"),
     )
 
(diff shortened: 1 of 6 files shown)


illicitonion (approved):
Seems reasonable - obviously this doesn't feel super tidy, but the constraints Windows imposes are real. Thanks for putting this together!

PR #2671, opened 2024-06-01
Ensure dynamic library dependencies end up in the runfiles directory

EdSchouten:
Right now the runfiles directory only contains "data" that is attached to a rust_binary() or one of its dependencies. This is inconsistent with cc_binary() and go_binary(), which also include an `_solib_*/` directory containing shared library dependencies.

When doing a plain 'bazel run' against the resulting executable, this is not a noticeable issue. The reason being that the runtime dynamic linker will use `${execroot}/_solib_*`. However, it does become a problem when `pkg_tar(include_runfiles = True)` is used to construct an archive of the binary so that it can be placed in a container image.

This change extends the runfiles gathering logic to merge all LibraryToLink.dynamic_library files in there as well.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1444,8 +1444,18 @@ def rustc_compile_action(
 
     experimental_use_coverage_metadata_files = toolchain._experimental_use_coverage_metadata_files
 
+    dynamic_libraries = [
+        library_to_link.dynamic_library
+        for dep in getattr(ctx.attr, "deps", [])
+        if CcInfo in dep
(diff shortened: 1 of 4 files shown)


illicitonion (commented):
This change LGTM, thanks!

Could you add a test to verify this behaviour? This feels like it could probably be tested in an analysistest like some examples in https://github.com/bazelbuild/rules_rust/tree/main/test/unit/native_deps or if that's hard could maybe a `sh_test` wrapping a `pkg_tar` in https://github.com/bazelbuild/rules_rust/blob/main/test/cc_shared_library/BUILD.bazel ?

illicitonion (approved):
Thanks!

PR #2674, opened 2024-06-02
Allow overriding crate universe packages to local packages

ograff:
Allow external crates to be aliased to local targets

In some cases it may be desirable to alias a public crate to a locally
define rust_library target, even for dependencies of crates that are not
being overriden.
Support this use case by swapping out `alias` for `rust_library` when
a user overrides the target using an annotation.


illicitonion:
LGTM! I made a PR with one suggested modification in https://github.com/ograff/rules_rust/pull/1 - PTAL and merge into this branch or discuss, and I'm happy to merge this :) Thanks!

ograff:
@illicitonion that change looks great, I went ahead and merged it in.

Is your preference for me to squash all these commits to one pre-merge or leave the history as is?

illicitonion:
We auto-squash on merge, so don't worry about it, GitHub will deal with it for us! Thanks!

illicitonion (approved):

PR #2675, opened 2024-06-03
Fix cargo_build_script executables.

matts1:
This allows it to work with new-style toolchains that use action configs (see https://github.com/bazelbuild/bazel/issues/22561).

It also allows you to specify a seperate C compiler and C++ compiler.


illicitonion (approved):
I had no idea action configs existed, but seems reasonable if it works!

PR #2676, opened 2024-06-03
Fix protobuf generated srcs to include all sources

matts1:


illicitonion (approved):
Seems eminently reasonable!

PR #2683, opened 2024-06-06
Add support for override_target when using bzlmod

AmeliasCode:
The ability to override crate targets when using crate_universe was added here: https://github.com/bazelbuild/rules_rust/pull/2674

However, this change did not expose that functionality when using bzlmod. 

This change adds that functionality in. Because of the limited set of options we have for dictionaries in tag classes, I had to split out "override_targets" into four different options, each taking a Label. 

I have added in an example based on the example given in #2674 , but using bzlmod instead. I have also tested that example and found it to be working.


illicitonion (commented):
Thanks, this LGTM! (@matts1 do you have strong feelings on the API here?)

Could you add a new entry to `.bazelci/presubmit.yml` which runs this on our CI (similar to the existing `bzlmod_no_cargo` entry)?

Other maintainers (non-blocking for this change): I'm not sure whether we want to consolidate our bzlmod examples into a single workspace with many crate_universe calls, or keep having nice separate independent small workspaces (and accordingly many bazelci top-level tasks) - any strong thoughts? cc @UebelAndre @hlopko @scentini @krasimirgg @matts1

AmeliasCode:
@illicitonion Thanks! I've gone ahead and added the tests that you mentioned, it looks like it worked but please let me know if I missed anything.

illicitonion (approved):
Thanks!

Issue #2684, opened 2024-06-07
@rules_rust//:rustfmt depends on a copy of rustfmt built for the 'exec' platform, not the 'target' platform


EdSchouten:
Imagine you're running Bazel on a Mac, but you're making use of a remote execution service that provides workers for both Linux and macOS. You want to prefer the use of Linux workers over macOS, so your .bazelrc looks like this:

```sh
build --host_platform=//:macos --extra_execution_platforms=//:linux,//:macos
```

Now if we invoke `@rules_rust//:rustfmt`, we see the following:

```
$ bazel run @rules_rust//:rustfmt //my/package/...
INFO: Invocation ID: 66ff4b58-bb69-42aa-969e-7bf54e029f10
INFO: Analyzed target @@rules_rust//:rustfmt (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target @@rules_rust//tools/rustfmt:target_aware_rustfmt up-to-date:
  bazel-bin/external/rules_rust/tools/rustfmt/target_aware_rustfmt
INFO: Elapsed time: 1.127s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/external/rules_rust/tools/rustfmt/target_aware_rustfmt //my/package/...
Loading: 0 packages loaded
INFO: Empty results
Loading: 0 packages loaded
INFO: Empty results
Loading: 0 packages loaded
INFO: Empty results
Loading: 0 packages loaded
/private/var/tmp/_bazel_ed/2cc046446ffd8acd61a49fc7cb12510e/execroot/myworkspace/bazel-out/darwin_x86_64-fastbuild/bin/external/rules_rust/tools/rustfmt/target_aware_rustfmt.runfiles/rustfmt_nightly-2024-05-02__x86_64-unknown-linux-gnu_tools/bin/rustfmt: /private/var/tmp/_bazel_ed/2cc046446ffd8acd61a49fc7cb12510e/execroot/myworkspace/bazel-out/darwin_x86_64-fastbuild/bin/external/rules_rust/tools/rustfmt/target_aware_rustfmt.runfiles/rustfmt_nightly-2024-05-02__x86_64-unknown-linux-gnu_tools/bin/rustfmt: cannot execute binary file
```

[...]

PR #2685, opened 2024-06-07
Depend on a copy of rustfmt for the target

EdSchouten:
The toolchain_files() and current_rust_toolchain() rules return files for the current 'exec' platform. This is perfecly reasonable if you want to use these tools as part of ctx.actions.run(). But if you want to use these as part of 'data = []' (runfiles), they must be built for the target.

Fixes: https://github.com/bazelbuild/rules_rust/issues/2684

### DIFF
--- a/rust/private/toolchain_utils.bzl
+++ b/rust/private/toolchain_utils.bzl
@@ -53,7 +53,7 @@ def _toolchain_files_impl(ctx):
     )]
 
 toolchain_files = rule(
-    doc = "A rule for fetching files from a rust toolchain.",
+    doc = "A rule for fetching files from a rust toolchain for the exec platform.",
     implementation = _toolchain_files_impl,
     attrs = {
         "tool": attr.string(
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
Thanks!

`transition_to_target` feels like it should maybe be built in to Bazel itself? @katre any idea where would be a good place for this transition to live for real?

(Happy to host a fork of it here in the mean time!)

katre:
@illicitonion I just read over the code, and I have a lot of questions about `transition_to_target`. It's a transition that sets the current target platform as an execution platform? Which makes the `:current_rustfmt_toolchain` dependency analyze with the toolchain (which is the rustfmt toolchain) as if the execution platform were the target platform of the actual `transition_to_target`?

I kind of feel like this is a lot of indirection: how many rustfmt toolchain implementations are there?

I'm happy to continue this discussion: if we made this a standard transition, it would belong in https://github.com/bazelbuild/rules_platform. Can you open a feature request there with a clear problem statement and any background context? Is this related to https://github.com/bazelbuild/bazel/issues/19645? It feels similar but I'm not sure.

EdSchouten:
Hey @katre,

> It's a transition that sets the current target platform as an execution platform? Which makes the :current_rustfmt_toolchain dependency analyze with the toolchain (which is the rustfmt toolchain) as if the execution platform were the target platform of the actual transition_to_target?

Exactly!

> I kind of feel like this is a lot of indirection: how many rustfmt toolchain implementations are there?

As far as I know, it's equal to the number of Rust toolchains being registered. There could be multiple versions registered, but also for multiple operating systems/hardware architectures.

> I'm happy to continue this discussion: if we made this a standard transition, it would belong in https://github.com/bazelbuild/rules_platform. Can you open a feature request there with a clear problem statement and any background context? Is this related to [bazelbuild/bazel#19645](https://github.com/bazelbuild/bazel/issues/19645)? It feels similar but I'm not sure.

This is indeed related to https://github.com/bazelbuild/bazel/issues/19645. Note that the approach we just merged uses a transition, but there's absolutely no hard requirement for that. If https://github.com/bazelbuild/bazel/issues/19645 would result in some kind of direct syntax for doing platform transitions, or having a special 'target' exec group or something, that would also be fine. I will file an issue at https://github.com/bazelbuild/rules_platform regardless.

Issue #1773, opened 2023-01-13
Allow path dependencies without having everything in one Cargo workspace


shelbyd:
My team has a monorepo for our Rust + Go + JS project. We're trying to get all of our rust code compiling in Bazel. We have path dependencies for our own code inside the monorepo. 

AFAICT this function https://github.com/bazelbuild/rules_rust/blob/52e02c25b6fb6f035eb453fec6162243e2d48daa/crate_universe/src/splicing/splicer.rs#L566 effectively requires us to have all of our Rust code in a single Cargo workspace if we're going to use path dependencies. This isn't a huge problem except for dependency resolution.

Cargo can't find compatible versions of dependencies, and I'm not aware of a way to resolve that. It's possible that specifying versions in the WORKSPACE file could work, or vendoring the dependencies. But I'd prefer not to do either of those options.

If we remove the root `[workspace]`, normal Cargo builds everything fine (path dependencies included). So it seems that the requirement of a Cargo workspace when using path dependencies is unnecessary and would be the most effective way to resolve our issue.

Does this sound reasonable? Is there a better way to resolve our situation?

hobofan:
I think this is a duplicate (or at least has large overlap) with https://github.com/bazelbuild/rules_rust/issues/1525.

shelbyd:
This is a large issue for my team. We have a few binaries in our repo that we cannot bring into Bazel because of this.

Looking at this more, the issue is that we have:
1. `bin0` indirectly requires `some_crate-v0` and `our_lib`
2. `bin1` indirectly requires `some_crate-v1` and `our_lib`
3. `our_lib` does not require `some_crate` (but may in the future)

There should be no target that actually ends up with both `some_crate-v0` and `some_crate-v1` in the build.

I tried to specify `packages` in our root `crates_universe` by aliasing the different versions of `some_crate`. But `{ "some_crate-v0": crate.spec(package = "some_crate", version = "0") }` did not generate any rules related to `some_crate`, so I don't know what to do there. I could probably get this working if that package spec worked.

The only remaining solution I see is to vendor our crates and have a single version of `some_crate` vendored. They're almost surely compatible, or at least minimal changes required. But I'd prefer not to include _all_ of the crates we use. Is there a way to vendor only certain crates that we need to modify but keep them in the same `crate_universe`?

As I said, this is a serious issue for my team, so I'd at least like some idea of which direction to go in, or potential alternatives. Thanks.

hobofan:
[...]

PR #2687, opened 2024-06-11
Ignore path deps in metadata hash

illicitonion:
They don't end up factoring into what we generate, so we pretend they don't exist. This avoids people who are using both Cargo and Bazel needing to spuriously repin just because they edited Cargo-only dependency information.


UebelAndre (changes requested):
Are path dependencies not desired?
- https://github.com/bazelbuild/rules_rust/issues/1773

If this support was implemented, would there be anything else being filtered out?

illicitonion:
> Are path dependencies not desired?

I'm actually not sure what this would reasonably look like. It feels like _maybe_ they are via https://github.com/bazelbuild/rules_rust/pull/2674 and _maybe_ we can automate setting some overrides up for path dependencies, but it's not super clear to me what target we can practically implicitly add a dependency on from a generated file into the local workspace. If we were to implement that kind of support, we would want to factor the replacement target, rather than the path dep value, into the cache key.

> If this support was implemented, would there be anything else being filtered out?

Yeah, we should probably be filtering out _everything_ that doesn't factor into dep resolution. So we should be stripping out things like authors, description, documentation, readme, homepage, ... - all of these things will currently require a repin if they change, even though they're just metadata for Cargo which Bazel will completely ignore.

PR #2688, opened 2024-06-11
Remote vendor doesn't require srcs present

illicitonion:
Before this, if we re-generate all of the BUILD.bazel files for our
vendored deps, they are invalid because they only have BUILD.bazel files
and the globs present in them expect srcs.

First two commits are real changes, third commit is just a repin.


UebelAndre (approved):

UebelAndre:
Thank you!

PR #2691, opened 2024-06-12
`crate_universe`: Make module extension restarts less expensive.

criemen:
This PR reduces the number of expensive module extension implementation restarts in combination with multiple `from_cargo` configurations.

As each configuration processing calls `module_ctx.path` (which can and will cause restarts), we executed _generate_hub_and_spokes (i.e. `cargo-bazel`) _a lot_ and then threw that away on the next restart.
Triggering `module_ctx.path` early is therefore a significant performance optimization.

On our repository, this gives us a 10sec speedup on module extension processing (local M1 mac, nothing happening in parallel), see our [MODULE.bazel](https://github.com/github/codeql/blob/main/MODULE.bazel) if you're interested in what we're doing.

This excessive restarting also exposed an upstream bazel bug on Windows 2019, where bazel spuriously fails to clean up the working directory (c.f. https://github.com/bazelbuild/bazel/issues/22710).


criemen:
Help with CI would be appreciated, I can't spot the problem right now.

crate_universe/extension.bzl
```diff
@@ -348,16 +348,24 @@ def _crate_impl(module_ctx):
             all_repos.append(cfg.name)
             local_repos.append(cfg.name)
 
+        # Postpone actually calling _generate_hub_and_spokes as it's rather expensive, and _every_ module_ctx.path can cause
+        # a restart of the module extension. In partigular with multiple cargo configurations, we'd run `cargo-bazel`
```
redsun82, line 352:
```suggestion
        # a restart of the module extension. In particular with multiple cargo configurations, we'd run `cargo-bazel`
```

criemen:
I significantly reworked the implementation of this, as the previous iteration would've still caused expensive restarts with from_cargo was used in different modules.

illicitonion (approved):
Seems reasonable, thanks!

The CI job should be fixed by https://github.com/bazelbuild/rules_rust/pull/2688 - should get that merged and this unblocked soon :)

criemen:
I checked with the bazel devs, and `module_ctx.path` on a new file _always_ causes a restart at the moment, as bazel watches the file for changes, and that for some reason needs a restart. That explains why this was so expensive, and why also moving the download of `cargo-bazel` after all the other path lookups makes sense.

Issue #2662, opened 2024-05-18
crate_universe: "Error: Failed to query cargo version"


mvukov:
```
Computing main repo mapping: 
INFO: Repository rules_ros2_crate_index instantiated at:
  /__w/rules_ros2/rules_ros2/examples/WORKSPACE:68:19: in <toplevel>
  /github/home/.cache/bazel/_bazel_runner/0f445b8fd31f922fe083ae63ab47da39/external/com_github_mvukov_rules_ros2/repositories/rust_setup_stage_3.bzl:4:22: in rust_setup_stage_3
Repository rule crates_repository defined at:
  /github/home/.cache/bazel/_bazel_runner/0f445b8fd31f922fe083ae63ab47da39/external/rules_rust/crate_universe/private/crates_repository.bzl:107:36: in <toplevel>
ERROR: An error occurred during the fetch of repository 'rules_ros2_crate_index':
   Traceback (most recent call last):
	File "/github/home/.cache/bazel/_bazel_runner/0f445b8fd31f922fe083ae63ab47da39/external/rules_rust/crate_universe/private/crates_repository.bzl", line 45, column 28, in _crates_repository_impl
		repin = determine_repin(
	File "/github/home/.cache/bazel/_bazel_runner/0f445b8fd31f922fe083ae63ab47da39/external/rules_rust/crate_universe/private/generate_utils.bzl", line 395, column 21, in determine_repin
		result = execute(
	File "/github/home/.cache/bazel/_bazel_runner/0f445b8fd31f922fe083ae63ab47da39/external/rules_rust/crate_universe/private/common_utils.bzl", line 54, column 13, in execute
		fail(_EXECUTE_ERROR_MESSAGE.format(
[...]

PR #2698, opened 2024-06-16
Pass HOME env var when fetching version

mvukov:
Fixes https://github.com/bazelbuild/rules_rust/issues/2662.

So, in a github action https://github.com/mvukov/rules_ros2/blob/980afb82f4af92480815e723b47a0cde4bea63a7/.github/workflows/main.yml I use my custom build image https://github.com/mvukov/bazel_builder/blob/49178bdf9c151b1fdb1fb0bac580da5d53a990fd/Dockerfile, no user defined. The github action passes user and HOME folder. The home folder seems to be `/github/home` and fetching the version seems to have problems with this. Therefore, I just pass the HOME env var here.


mvukov:
@illicitonion Can you provide some feedback on this, please?

mvukov:
So, the issue seems to be the following: I build code in a container w/o user defined (hence, root). In a github action, I pass --user argument to the container. I think that cargo eventually resorts to read std::env::home that reads /etc/passwd and that fails in my case (as HOME env var is not routed). Hence, I kindly ask for this PR to be taken into consideration.

@illicitonion @UebelAndre PTAL.

UebelAndre (changes requested):
Sorry for the delay here! Is https://github.com/bazelbuild/rules_rust/issues/2662#issuecomment-2152758543 not the root cause here? Does this fix the behavior inside your container?

mvukov:
@UebelAndre That's unfortunately not going to work. I removed user from my container, so dockerfile only has root. The thing is that the github action passes [user](https://github.com/mvukov/rules_ros2/blob/980afb82f4af92480815e723b47a0cde4bea63a7/.github/workflows/main.yml#L52) (that's how I got github action working with my container). The action also passes HOME env var which is ` /github/home` (valid folder). Since HOME is not passed to cargo, cargo looks at `/etc/passwd`. However, when docker passes user  via `--user` that user is "homeless", it doesn't exist in `/etc/passwd`. Therefore, we need HOME env var for cargo to work.

mvukov:
@illicitonion @UebelAndre PTAL.

mvukov:
@illicitonion @UebelAndre PTAL.

mvukov:
@illicitonion @UebelAndre PTAL.

mvukov:
@illicitonion @UebelAndre PTAL.

mvukov:
@krasimirgg @illicitonion @UebelAndre PTAL. Please let me know if I can help somehow to move this forward.

mvukov:
@krasimirgg @illicitonion @UebelAndre PTAL.

UebelAndre (approved):
Thank you for your patience and persistence! I've finally had mental bandwidth to dig into this and for just getting the version I think access to `HOME` is not going to introduce much if any volatility. My original concern was that `HOME` was gonna be set when doing splicing or something which could lead to configs sneaking into generated outputs.

Thank you so much!

mvukov:
Many thanks for the feedback and efforts.

PR #2703, opened 2024-06-18
upstream wrappers for cargo, rustc, cargo-clippy

kolloch:
...and a bazel_env example.

As discussed in https://bazelbuild.slack.com/archives/CSV56UT0F/p1718207171652029?thread_ts=1718182474.631279&cid=CSV56UT0F


rust/toolchain.bzl
```diff
@@ -713,13 +714,13 @@ rust_toolchain = rule(
             allow_single_file = True,
             cfg = "exec",
         ),
-        "clippy_driver": attr.label(
-            doc = "The location of the `clippy-driver` binary. Can be a direct source or a filegroup containing one item.",
+        "cargo_clippy": attr.label(
```
UebelAndre, line 717: Is cargo_clippy a real binary?

kolloch: Yep, it is. Or `cargo-clippy` is, to be exact.

If you call `cargo clippy`, `cargo-clippy` is executed. And that uses `clippy-driver` somehow.

examples/bazel_env/all_crate_deps_test.sh
```diff
@@ -0,0 +1,35 @@
...
+}
+
+# MARK - Args
+
+if [[ "$#" -ne 1 ]]; then
+  fail "Usage: $0 /path/to/hello_world"
+fi
+HELLO_WORLD="$(rlocation "$1")"
+
+# MARK - Test
```
illicitonion, line 31: Could you modify this test to invoke all of the tools provided by the upstream_wrapper package, to verify that they all work?

illicitonion (commented):
LGTM - just one question around tests :)

examples/bazel_env/all_crate_deps_test.sh
kolloch: Yep, good idea.

kolloch: Maybe separate tests?

kolloch: More like `rustc --version` or a real compilation? (same for the other stuff)

illicitonion: I don't have strong opinions - separate tests sound good, but if it's easier to do everything together that's fine too :)

illicitonion: > More like rustc --version or a real compilation? (same for the other stuff)

For each, if it's easy to put together a real-world example (i.e. a trivial hello-world binary, with a Cargo project, with a single failing clippy lint or something), that'd be really nice, but if that's really fiddly, just checking that `--version` or `--help` or something for each binary works would be good enough

rust/toolchain.bzl
UebelAndre: Wait, so then there is no `cargo_clippy` binary? There is `cargo` and then there is `clippy-driver`. Right?

kolloch: There is a `cargo` binary, a `cargo-clippy` and a `clippy-driver` binary.

The `cargo-clippy` binary is in the same "component" archive as `clippy-driver`.

UebelAndre: Would it be worth keeping `clippy-driver` around and additionally exposing `cargo-clippy`?

examples/bazel_env/all_crate_deps_test.sh
kolloch: Hmmm... I have been thinking about it. Depending on the context, it is somewhat different what you want.

A) If the user executes something like `bazel run @rules_rust//tools/upstream_wrapper:cargo check`, they probably want it to find the hermetic `rustc`. By default, it will not.

B) Whereas when you create a `bazel_env.bzl`, all the hermetic tools are already in the path and automatically found. The current version works without effort.

---

A1) special handling of sub-commands 

For A), one could bake the dependencies to a couple of things like `rustc` into the wrapper. E.g. it could set the [`RUSTC`](https://doc.rust-lang.org/cargo/reference/environment-variables.html) variable to the hermetic `rustc`.

It might be weird though for a user that if they override `RUSTC` for a reason. So one could maybe check for an existing one? Any ways, it might still be confusing to the user if they have a `rustc` in the path and it doesn't get picked up. 

For `rustc` that might be really an edge case. For external commands, and `cargo clippy` is one, the `cargo-clippy` binary has to be in the `PATH` as far as I know. There is no env variable to specifically tell cargo where to search for this sub command, I think.

A2) put the sysroot bin in the PATH in the wrapper

More logical for me would be to put the sysroot bin directory into the PATH in the wrapper.

To ensure that we don't by accident pick up existing binaries that we have in our sysroot, we'd need to prepend it. Again it might be somewhat confusing in special situations but overall, I think it is the best.

kolloch:
@illicitonion as described above, I am now prepending the `bin` directory to the PATH in the wrapper so that `cargo` can find `cargo-clippy`, `rustc` and `rustfmt` -- even if they are not setup with bazel_env but only called with `bazel_run`.

Also added some tests as requested that test this.

rust/toolchain.bzl
kolloch: I did keep `clippy-driver` around. The diff is misleading. Some lint forced me to sort the attributes alphabetically which makes it appear as if I removed clippy_driver. Instead I just moved it.

kolloch:
@illicitonion I rebased it on https://github.com/bazelbuild/rules_rust/pull/2700 since it had conflicts. (Thanks BTW)

illicitonion (approved):
Looks great, thanks so much!

Issue #2701, opened 2024-06-17
Bzlmod has no support for strip_level or opt_level


chrisstaite-menlo:
There's currently no way to pass the strip_level or opt_level to rust_toolchain using Bzlmod.

marvin-hansen:
You can do this with manual compiler option pass through in 3 steps

1) In your root folder BUILD.bazel, add the following entry:

```Starlark
config_setting(
    name = "release",
    values = {
        "compilation_mode": "opt",
    },
)
```

This maps Rusts release mode to Bazel's -c opt.

2) In your binary, add the opt flags & strip settings: 

```Starlark 
# Build binary
rust_binary(
    name = "bin",
    crate_root = "src/main.rs",
    srcs = glob([
        "src/*/*.rs",
        "src/*.rs",
    ]),
    # Compiler optimization
    rustc_flags = select({
       "//:release": [
            "-Clto",
            "-Ccodegen-units=1",
            "-Cpanic=abort",
            "-Copt-level=3",
            "-Cstrip=symbols",
            ],
        "//conditions:default":
        [
           "-Copt-level=0",
        ],
    }),

    deps = [   ],
    visibility = ["//visibility:public"],
)
```

3) Run bazel build with optimization

``bazel build -c opt //...`

Example code:
https://github.com/marvin-hansen/bazel_rust_example

Hope that helps.

chrisstaite-menlo:
Oh, that's great, thank you!

marvin-hansen:
I'm drafting currently the new Bazelmod documentation that should get upstreamed hopefully soon. Do you think this issue can be closed given this will be documented?

Issue #2670, opened 2024-05-31
documentation for using via module.bazel


pdeva:
current docs are focused on using `WORKSPACE` file..

marvin-hansen:
@pdeva I've just made an end to end example of a Rust repo using module.bazel and added some basic docs. 

https://github.com/marvin-hansen/bazel_rust_example

However, there is still one open issue, which I try to resolve:
https://github.com/bazelbuild/rules_rust/issues/2689

marvin-hansen:
The first draft of the Rust with Bzlmod documentation is available now:

https://github.com/marvin-hansen/bazel_rust_example/blob/main/bzlmod.md#rust-with-bzlmod

Once the doc is reasonably ready, it will be merged into the official documentation. In the meantime, 
feedback is welcome.

UebelAndre:
Closing now that https://github.com/bazelbuild/rules_rust/pull/2713 has been merged, thank you!

Issue #2668, opened 2024-05-28
Support for rules_proto 6.x


aira-xite:
In the latest version of [rules_proto](https://github.com/bazelbuild/rules_proto/releases/tag/6.0.0) several breaking changes have been made, in particular `rules_proto_toolchains` has been moved and it breaks build of rules_rust.
```
ERROR: Traceback (most recent call last):
        File "..../external/rules_rust/proto/protobuf/transitive_repositories.bzl", line 4, column 75, in <toplevel>
                load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_proto_toolchains")
Error: file '@rules_proto//proto:repositories.bzl' does not contain symbol 'rules_proto_toolchains'
``` 

Please consider adding support for it.

UebelAndre:
I would imagine with bzlmod this is no longer an issue. Closing this out and I advise users to switch to bzlmod.

Issue #2534, opened 2024-02-29
Can't find instructions for protobuf toolchains from bzlmod


riking:
The protobuf instructions make no mention of bzlmod: https://bazelbuild.github.io/rules_rust/rust_proto.html

UebelAndre:
Most `rules_rust` maintainers are pretty new to bzlmod. If anyone wants to open a PR adding some docs I think we'd be happy to review and learn 😅

PR #2713, opened 2024-06-24
More Bazelmod code examples.

marvin-hansen:
This PR addresses  a number of issues related and Bazelmod.

First and foremost it paves the way for a meaningful update the Bazelmod documentation that references these and existing code examples. This touches at least the following two issues. 
* documentation for using via module.bazel #2670
* Migrate to Bzlmod for managing external dependencies #2181

The compile_opt example addresses or resolves:
* Analog to cargo profiles in rules_rust #515
* Bzlmod has no support for strip_level or opt_level #2701

The musl_cross_compilling example addresses or resolves
* Static compilation of rust binaries using musl #390 
* Document how to do cross-compilation #276

The oci_container does not relate to any open issue, 
although the tokio example in it gives a nice end to end example so that never hurts.

The proto example addresses or resolves:
* Support for rules_proto 6.x #2668
* Protobuf Version #302
* Can't find instructions for protobuf toolchains from bzlmod #2534
* Possibly a few more if I were to search longer

Formalities
* I've signed the CLA
* I've signed all commits


marvin-hansen:
Updated the crate_universe_bzlmod.md documentation. 

This PR updates both, the example folder and the doc folder. Although in the doc folder, it only
updates the newly created  crate_universe_bzlmod.md file and filled the blanks with some content.

Note, the documentation references some of the code examples in this PR, specifically the dependency vendoring example. 

Please review to

docs/crate_universe_bzlmod.md
```diff
@@ -8,13 +8,318 @@ This doc describes using crate_universe with bzlmod.
 
 If you're using a WORKSPACE file, please see [the WORKSPACE equivalent of this doc](crate_universe.html).
 
-There are some examples of using crate_universe with bzlmod:
+There are some examples of using crate_universe with bzlmod in the [example folder](../examples/bzlmod).
```
UebelAndre, line 11: Is this gonna render correctly?

UebelAndre (changes requested):
I haven't finished going through all the examples but could you update the presubmit yaml to make sure the examples are built and regression tested?

So far this is looking amazing though!

docs/crate_universe_bzlmod.md
marvin-hansen: Yes, that works.

marvin-hansen:
@UebelAndre I'm looking at the presubmit.yaml and how the other examples are build. I've to look closer into regression testing because I don't think I've added much yet. 

Also, I need a bit of time to figure this out because I've stopped using GH actions a long time. The good news is, the same config used in the examples builds on my CI (BuildBuddy on Ubuntu 20.04) and the way I understand this pre-sub config is that the bzlmd job also runs on ubuntu 20.04 so I hope this is just a matter of tine.

marvin-hansen:
Looks like this was a lot easier than I though.

* All bazelmod examples build on CI
* All linter errors reported fixed.

The only remaining CI issue seems to be related to a git detached file that prevents the doc build.

I have no idea how that happened, but would you please take a look how to fix the dod build?

Doc build: https://buildkite.com/bazel/rules-rust-rustlang/builds/11570#019058a0-1986-4b05-bafc-6ed8ce8443c7

examples/bzlmod/oci_container/README.md
UebelAndre, line 1: I think I'm on board with all examples you've posted but I think the oci example should live in the `rules_oci` repo. Is this something you could move to that repo? Or curious if @illicitonion has any thoughts here.

UebelAndre (changes requested):
Sorry for the delay! I had one last question and seems like the docs job is out of date? Maybe you just need to run https://github.com/bazelbuild/rules_rust/blob/0.46.0/docs/update_docs.sh

examples/bzlmod/oci_container/README.md
marvin-hansen: OK, I had a similar thought. I think it's best to remove the OCI example.
I've currently another PR over at the[ bazelbuild example repo ](https://github.com/bazelbuild/examples/pull/477) and it might be better located there as the rules_oci repo already has a number of good example.

marvin-hansen:
OK, 
just a quick update:

1) I've renovated the cross compile example, added comprehensive Readme, and added it to the CI Build. 

It's actually possible to move the sysroot download into MODULE so the workspace file is not needed anymore.
I fixed an hermeticity issue when building on ARM by adding the proper X86 sysroot and, on that occasion, I've updated the sysroot to the latest 04/2024 release to prevent issued with too old sys libs. 

2) For the same reason, I've updated the MUSL example and readme. Same change set as the cross example.

Not sure if you want to re-review those two examples, but if so, it would be quick. 

3) Removed the OCI example, as mentioned before as its better placed elsewhere.

4) Re-generated and tested the docs. Thanks for the hint

Build seems green to me. 

Please let me know if the current state can be merged or if you seek any further changes.

Thank you

examples/bzlmod/oci_container/README.md
UebelAndre: I also noticed https://github.com/bazel-contrib/rules_oci/blob/v1.7.6/docs/rust.md so sounds like there's some interest in guides there already!

UebelAndre:
Oh, and clang-format seems unhappy

examples/bzlmod/musl_cross_compiling/build/.DS_Store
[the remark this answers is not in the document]
marvin-hansen: Done. Updated .gitignore.

marvin-hansen:
Alright,

I've fixed the issue you've mentioned. Also, I've added platform tests to the cross compile tests, which are all passing.

However, MUSL became a bit of an issue. I've added both, platform tests and static tests.

The test pass on my Macbook, but fail on CI.

[All tests are defined here](https://github.com/bazelbuild/rules_rust/blob/1854074e41050614d6519744ce23e03d29eff6dd/examples/bzlmod/musl_cross_compiling/musl_scratch/BUILD.bazel).

[And the script used to test.](https://github.com/bazelbuild/rules_rust/blob/1854074e41050614d6519744ce23e03d29eff6dd/examples/bzlmod/musl_cross_compiling/musl_scratch/test_platform.sh)

Out of 4 tests, only one passes, strangely, and all the other tests fail with the following error

Wrong file type: musl_scratch/bin_linux_x86_64_musl/bin: symbolic link to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/57f73e3b9e45c5287764d9e8d83b2c8b/execroot/_main/bazel-out/k8-fastbuild/bin/musl_scratch/bin_linux_x86_64_musl/bin

[Full Testlog](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/0190812d-b1fa-4639-89e9-c5522c59f975/musl_scratch/test_musl_x86_64_static/test_attempts/attempt_1.log)

[Failed CI Run for MUSL example](https://buildkite.com/bazel/rules-rust-rustlang/builds/11606#01908139-d7e7-4b23-9be4-a21d7e8553a9)

I've added these tests because I've notices very recently that on Macos, MUSL produces correct static binaries, but not on my CI (BuildBuddy on Ubuntu 20.04) and I've reported that yesterday in Slack, although with a different error. In order to verify if this example is affected, I added the platform and static tests, which fail. 

From the error alone, I cannot tell if the script takes issue with the symbolic link, or if the binary really is incorrectly linked. 

Moving forward, I suggest to exclude the MUSL example, put it separately for investigation, and merge the rest when ready.

Once I've figured out the exact issue with MUSL, I would then fill a separate PR with a fixed MUSL example.

Please let me know.

examples/bzlmod/oci_container/README.md
marvin-hansen: Thanks for the link. I look into the OCI stuff once the MUSL issue is resolved.

marvin-hansen:
For all practical purpose, it's best to exclude the MUSL example from this PR, merge it, and then then move on. I can re-add the example anytime after the toolchain issue has been resolved. 

I've tried quite a few things to resolve the MUSL issue, but to no avail. 

The way I see it is, that the toolchain resolution touches quite a few sensitive code blocks and somehow it's not so easy 
to separate resolution on linux x86_64 for different ABI's. At least I could not find a way that resolves the issue and wthat isn't breaking something else along the way. However, I suggested a possible alternative in the [MUSL toolchain resolution issue](https://github.com/bazelbuild/rules_rust/issues/2726), but that needs more research. Therefore I suggest to remove the MUSL example, merge this PR, and then take it from there.

examples/bzlmod/compile_opt/README.md
```diff
@@ -0,0 +1,51 @@
+# Hello World with compiler optimization
+
+By default, rules_rust you can define compiler option for each binary target.
```
illicitonion, line 3:
```suggestion
Each binary target can have its own compiler options, and these can be customised differently for different optimisation levels.
```

examples/bzlmod/cross_compile/LLVM_Troubleshooting.md
```diff
@@ -0,0 +1,29 @@
...
+
+`
+apt update && apt install -y libtinfo5
+`
+
+The libtinfo5 library may have different package names on other distributions, but it is a well known
+issue. [See this SO discussion](https://stackoverflow.com/questions/48674104/clang-error-while-loading-shared-libraries-libtinfo-so-5-cannot-open-shared-o)
+for various solutions.
+
+On MacOX, it is sufficient to have the Apple Clang compiler installed.
```
illicitonion, line 18: @keith @brentleyjones do y'all have specific recommendations or docs we can link to here around when to have Xcode CLT vs full install?

examples/bzlmod/cross_compile/README.md
```diff
@@ -1,6 +1,290 @@
...
+You cross-compile by calling the target.
+
+`bazel build //:hello_world_x86_64`
+
+or
+
+`bazel build //:hello_world_aarch64`
+
+Notice, the default target `//...` does not know about the 
+many different target platform and will report an error.
```
illicitonion, line 23: I'm not sure what problem this is referring to, and this seemed to work for me - can you elaborate / give an example of the issue?

examples/bzlmod/musl_cross_compiling/README.md
```diff
@@ -0,0 +1,779 @@
...
+
+This is a non-trivial constraint, but not every C/C++ target compiles with MUSL. Unfortunately, there is no
+other way to find out than actually trying to build it. As a rule of thumb, it is best to avoid Rust crates that just wrap an existing C/C++ whenever possible. Many Rust crates have some implicit C/C++ dependencies, but many of these actually build with MUSL. However, the larger the underlying C/C++ library, the higher the chance that something breaks. 
+
+The exact problem isn't C/C++ per se, it's the problem that, just one call in std libc that tries to open a file or something else that's not supported in MUSL binaries simply breaks the build. This is equally true for Rust crates, but many Rust crates use feature flags to mitigate IO related issues. Some code out there actually requires system IO, and that just doesn't compile with MUSL. If somehow one or more major library is needed and cannot be build with MUSL, it's sensible to just move on, a build a dynamically linked binary and use a distroless image instead, as shown in the [OCI example](../07-oci-container).
+
+### 3) MUSL builds best with network only applications
```
illicitonion, line 54: Is a "network only application" a common term? I've not come across it before, and am not entirely sure what it's referring to here...

examples/bzlmod/musl_cross_compiling/README.md
```diff
@@ -0,0 +1,779 @@
...
+
+When building network only applications, then MUSL builds most of the time out of the box. However, the only remaining issue comes from DNS resolution. For historical reasons, the DNS resolution implementation in the standard library expects a HOST file with a default DNS server. Because it's a file, this doesnt work with MUSL. However, if you use a custom DNS resolver such as the excellent [Hickory DNS crate](https://github.com/hickory-dns/hickory-dns), you simply switch off the default features in the crate dependency and then construct the hickory resolver with a custom configuration containing the network DNS server. From there, MUSL usually compiles your code. 
+
+
+### 4) Test cross compiled MUSL Containers on CI or with Docker
+
+There is non-zero chance of a segfault for any number of reasons. It's rare, but Tte compiler cannot catch those corner cases so the best you can do is to pull the actual image, and start it either locally or on your CI and run some basic integration tests. Integration tests are generally considered best practice,
```
illicitonion, line 61:
```suggestion
There is non-zero chance of a segfault for any number of reasons. It's rare, but the compiler cannot catch those corner cases so the best you can do is to pull the actual image, and start it either locally or on your CI and run some basic integration tests. Integration tests are generally considered best practice, 
```

examples/bzlmod/proto/.DS_Store
illicitonion, line 1: Can you remove this and any other .DS_Store files?

examples/bzlmod/oci_container/README.md
illicitonion: cc @alexeagle @thesayyn do y'all have thoughts on where language-specific rules_oci examples should live?

illicitonion (commented):
This is generally looking great, thanks so much for taking the time to put it all together!

I agree we should probably remove the musl example until we resolve https://github.com/bazelbuild/rules_rust/issues/2726 but would love to re-add it after then!

examples/bzlmod/musl_cross_compiling/README.md
marvin-hansen: I just made the term up while writing this. it just means, your service is using only network to load whatever it needs. I am writing these kinds of micro services for years and build them static binaries w/o (g)libc due to security requirements. In GoLang, that's largely a non-event, but in Rust with MUSL, well, it's still a learning process for me. 

However, I agree that's not the best wording so let me edit this

examples/bzlmod/proto/.DS_Store
marvin-hansen: sure

examples/bzlmod/cross_compile/README.md
marvin-hansen: Correct, I just checked it a moment ago. 
Seems to be a leftover from. Thanks for catching it. 

I've updated the Readme to reflect that  target `//...` works as expected.

examples/bzlmod/musl_cross_compiling/README.md
marvin-hansen: As this example gets removed for this PR, and most likely renovated once the MUSL issue is fixed,.
I have to re-write this Readme anyways, but  I put in a note for this.

examples/bzlmod/proto/.DS_Store
marvin-hansen: Done.

marvin-hansen:
Please let me know if you want any further changes.

Otherwise, build is green.

marvin-hansen (commented):
LGTM

examples/bzlmod/oci_container/README.md
thesayyn: we accept language specific examples as docs. there's the rust one that @UebelAndre sent above. Can be updated as needed.

examples/bzlmod/proto/grpc_server/.DS_Store
illicitonion, line 1: Please remove

examples/bzlmod/proto/proto_bindings/.DS_Store
illicitonion, line 1: Please remove

examples/bzlmod/proto/.DS_Store
illicitonion, line 1: Please remove

illicitonion (approved):
LGTM, thanks!

examples/bzlmod/proto/grpc_server/.DS_Store
marvin-hansen: Sorry for the blunder, somehow the .gitignore didn't worked or was updated after the push. Don't know.  

All fixed.

marvin-hansen (commented):
Green. Good to go?

marvin-hansen:
@UebelAndre Would you please take a final look and decide whether to merge?

That would make the Bazelmod docs available online and allow to close a number of related issues.

Thank you

UebelAndre (approved):
Thank you so much!

Issue #2716, opened 2024-06-27
rust_analyzer ignores repo srcs in crates with generated srcs


UebelAndre:
Any target which has a single generated source will have all sources converted to generated sources so they live next to each other. This works fine for rustc but when using rust-analyzer on your repo, it is confusing as to why entire crates are getting completely ignored. The root cause is the source code in the repo is never referenced in the generated `rust-project.json` file and thus rust-analyzer does not know to process those files. An easy repro is to run the following in the `rules_rust` repo:
```
bazel build //...
bazel run @rules_rust//tools/rust_analyzer:gen_rust_project
```

From here, observe any of the targets in [tests/generated_inputs](https://github.com/bazelbuild/rules_rust/tree/0.46.0/test/generated_inputs) package.
```json
{
  "display_name": "gen_src_in_different_cfg",
  "root_module": "/private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/execroot/rules_rust/bazel-out/darwin_arm64-fastbuild/bin/test/generated_inputs/root.rs",
  "edition": "2018",
  "deps": [],
  "is_workspace_member": true,
  "cfg": [
    "test",
    "debug_assertions"
  ],
  "target": "aarch64-apple-darwin",
  "env": {
    "CARGO_CFG_TARGET_ARCH": "aarch64",
    "CARGO_CFG_TARGET_OS": "darwin",
    "CARGO_CRATE_NAME": "gen_src_in_different_cfg",
    "CARGO_MANIFEST_DIR": "/private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/execroot/rules_rust/test/generated_inputs",
    "CARGO_PKG_AUTHORS": "",
    "CARGO_PKG_DESCRIPTION": "",
    "CARGO_PKG_HOMEPAGE": "",
    "CARGO_PKG_NAME": "gen_src_in_different_cfg",
    "CARGO_PKG_VERSION": "0.0.0",
    "CARGO_PKG_VERSION_MAJOR": "0",
    "CARGO_PKG_VERSION_MINOR": "0",
    "CARGO_PKG_VERSION_PATCH": "0",
    "CARGO_PKG_VERSION_PRE": "",
    "REPOSITORY_NAME": ""
  },
  "is_proc_macro": false
},
```

[...]

PR #2717, opened 2024-06-28
Updated rust-analyzer to use repo vs generated files as crate roots

UebelAndre:
This change updates the `rust_analyzer_aspect` to attempt to match generated crate root sources to `srcs` inputs and use the the path of the source file in place of the generated path. The generated path is then added to `sources.include_dirs` for compatibility. The impact of this can be seen in the following diff of the `@rules_rust//:rust-project.json` file:
```diff
--- rust-project.old.json	2024-06-28 09:00:51
+++ rust-project.json	2024-06-28 09:04:43
@@ -821,10 +821,16 @@
         },
         {
             "display_name": "libgensrc_with_crate_root",
-            "root_module": "/private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/execroot/rules_rust/bazel-out/darwin_arm64-fastbuild/bin/test/generated_inputs/lib.rs",
+            "root_module": "test/generated_inputs/lib.rs",
             "edition": "2018",
             "deps": [],
             "is_workspace_member": true,
+            "source": {
+                "include_dirs": [
+                    "/private/var/tmp/_bazel_user/76282c66b0dfe3c5cb9a230bdc913a52/execroot/rules_rust/bazel-out/darwin_arm64-fastbuild/bin/test/generated_inputs"
+                ],
+                "exclude_dirs": []
+            },
             "cfg": [
                 "test",
                 "debug_assertions"
@@ -850,10 +856,16 @@
...
...
...
```

closes https://github.com/bazelbuild/rules_rust/issues/2716

### DIFF
--- a/test/rust_analyzer/generated_srcs_test/rust_project_json_test.rs
+++ b/test/rust_analyzer/generated_srcs_test/rust_project_json_test.rs
@@ -1,0 +1,43 @@
+#[cfg(test)]
+mod tests {
+    use std::env;
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_deps_of_crate_and_its_test_are_merged() {
+        let rust_project_path = PathBuf::from(env::var("RUST_PROJECT_JSON").unwrap());
+
+        let content = std::fs::read_to_string(&rust_project_path)
+            .unwrap_or_else(|_| panic!("couldn't open {:?}", &rust_project_path));
+
+        let output_base = content
+            .lines()
+            .find(|text| text.trim_start().starts_with("\"sysroot_src\":"))
+            .map(|text| {
+                let mut split = text.splitn(2, "\"sysroot_src\": ");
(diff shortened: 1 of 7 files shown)


test/rust_analyzer/generated_srcs_test/rust_project_json_test.rs
```diff
@@ -0,0 +1,43 @@
...
+      "is_workspace_member": true,
+      "source": {
+        "include_dirs": [
+          "#
+        .to_owned()
+            + output_base;
+
+        println!("{}", content);
+        assert!(
+            content.contains(&expected),
```
illicitonion, line 38: FWIW since https://github.com/bazelbuild/rules_rust/pull/2709 we can easily parse and assert over JSON in these tests, rather than doing string equality comparisons :) Not saying we necessarily should, just that we can

illicitonion (approved):

PR #2718, opened 2024-06-29
Clean up downloads that go through the API instead of the CDN

ograff:
This is useful as it seems crates.io is starting to deny some direct requests, not sure if its based off user agent or remote how IP.


UebelAndre (approved):
Thank you @ograff!

PR #2723, opened 2024-07-03
Introduce feature `rules_rust_link_cc`.

yuzhy8701:
This feature is similar to `rules_rust_unsupported_feature`. The difference is that, `rules_rust_unsupported_feature` gets disabled if *any* rust is involved, while this feature gets disabled if *only* rust is involved.

My use case for this feature is to static link to c++ / objc stdlibs. They are needed only when any C object is being linked, and will produce invalid rust_proc_macro if linked blindly.

This implementation uses the fact that `DepInfo.transitive_noncrates` would be empty unless there is a `cc_library` or `cc_import` in the (transitive) deps. As a side effect, `collect_deps()` can no longer use the output of `_are_linkstamps_supported()` so it now always returns a populated linkstamp depset, to be discarded later if linkstamps are not supported.


rust/private/rustc.bzl
```diff
@@ -1149,7 +1155,10 @@ def rustc_compile_action(
     rustc_output = crate_info_dict.get("rustc_output", None)
     rustc_rmeta_output = crate_info_dict.get("rustc_rmeta_output", None)
 
-    cc_toolchain, feature_configuration = find_cc_toolchain(ctx)
+    extra_disabled_features = [RUST_LINK_CC_FEATURE]
+    if crate_info_dict["type"] in ["bin", "cdylib"] and any([d.cc_info for d in crate_info_dict["deps"].to_list()]):
```
scentini, line 1159: This would only be true if a direct dependency is a C/C++ dependency; what's the intended behavior when there is a transitive C/C++ dependency?

yuzhy8701: That's a good question. Looks like I can utilize the collected `dep_info` to include all the transitive deps as well. Updated the PR.

scentini (approved):
Thanks!

PR #2732, opened 2024-07-08
Use newer macos runners

ograff:


version.bzl
```diff
@@ -1,3 +1,3 @@
 """The version of rules_rust."""
 
-VERSION = "0.47.0"
+VERSION = "0.48.0"
```
UebelAndre, line 3: Nit: could we do 0.47.1?

UebelAndre (approved):
Thank you!

UebelAndre (approved):

PR #2734, opened 2024-07-09
Add support for shared test-only modules

IvanIsCoding:
Currently, `rust_test_suite` is not capable of supporting integration tests with shared modules. A very basic example is:
```
[crate]/
    src/
        lib.rs
    tests/
        integration_test_a.rs
        integration_test_b.rs
        util/
            mod.rs
```
With `integration_test_a.rs` and `integration_test_b.rs` importing code from `util`.

This PR adds a `shared_srcs` argument to `rust_test_suite`. That way, we can finally compile and run those tests with:
```python
rust_test_suite(
    name = "integrated_tests_suite",
    srcs = glob(["tests/**"]),
    shared_srcs = ["tests/util/mod.rs"],
    deps = [":example_dep"],
)
```

### DIFF
--- a/docs/defs.md
+++ b/docs/defs.md
@@ -650,7 +650,7 @@ Run the test with `bazel test //hello_lib:greeting_test`.
 ## rust_test_suite
 
 <pre>
-rust_test_suite(<a href="#rust_test_suite-name">name</a>, <a href="#rust_test_suite-srcs">srcs</a>, <a href="#rust_test_suite-kwargs">kwargs</a>)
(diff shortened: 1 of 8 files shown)


scentini:
Thanks!

scentini (approved):

IvanIsCoding:
@scentini I think "iOS build script cross compile test on macOS" is broken and blocking the merge? It fails with this message that seems unrelated
```
Caused by:
Error: Failed to read /tmp/tmp76kc680t/test_bep.json
```

PR #2739, opened 2024-07-11
Allow multiple uses of extra_rustc_flags

vabr-g:
Currently, only one extra_rustc_flags is taken into account when compiling, which might lead to puzzling errors when combining multiple configs.

This was manually tested on an internal system.


vabr-g:
I'm withdrawing this PR, because after talking to a colleague who is a Rust expert, we decided to fix the issue differently (by fixing a bug in extra_rustc_flag instead of changing the behaviour of extra_rustc_flags).

PR #2740, opened 2024-07-12
Expose bindgen toolchain through bzlmod

jwnrt:
With this change I was able to use the default bindgen toolchain with bzlmod by adding this to my `MODULE.bazel`:

```bazel
bindgen = use_extension("@rules_rust//bindgen:extension.bzl", "bindgen")
use_repo(bindgen, "bindgen_toolchains")

register_toolchains("@bindgen_toolchains//:default_bindgen_toolchain")
```

Had to backport a change from recent versions of LLVM which use this `WORKSPACE_ROOT` environment variable to fix paths to generated files. Updating the LLVM archive would also fix that problem.

Happy to add tests and documentation if this matches what you had in mind for getting this working.


bindgen/transitive_repositories.bzl
```diff
@@ -8,7 +8,6 @@ def rust_bindgen_transitive_dependencies():
 
     llvm_configure(
         name = "llvm-project",
-        repo_mapping = {"@llvm_zlib": "@zlib"},
```
UebelAndre, line 11: Why delete this?

bindgen/extension.bzl
```diff
@@ -0,0 +1,24 @@
...
+        name = "bindgen_toolchains",
+        build = "//bindgen/toolchain:BUILD.bazel",
+    )
+
+bindgen = module_extension(implementation = _bindgen_impl)
+
+def _bindgen_toolchains_impl(ctx):
+    ctx.file("BUILD.bazel", ctx.read(ctx.attr.build))
+
+bindgen_toolchains = repository_rule(
```
UebelAndre, line 19: Can you document the rule and attributes here?

bindgen/extension.bzl
```diff
@@ -0,0 +1,24 @@
...
+    implementation = _bindgen_toolchains_impl,
+    attrs = {
+        "build": attr.label(mandatory = True),
```
UebelAndre, line 22: I'm not sure if this is the case for bzlmod but isn't the precedent something like
```suggestion
        "build_file_content": attr.string(mandatory = True),
        "build_file": attr.label(mandatory = True),
```

UebelAndre (changes requested):
Thank you! Just had some questions and nits 😄

lalten:
@jwnrt any chance you could pick this up again? Seems like this would fix https://github.com/bazelbuild/rules_rust/issues/2787

bindgen/extension.bzl
jwnrt: Replaced with `local_repository`

jwnrt:
> @jwnrt any chance you could pick this up again? Seems like this would fix https://github.com/bazelbuild/rules_rust/issues/2787

Sorry, I took my eye off this when I realised I wouldn't need it for OpenTitan.

I'm happy to finish it off but I won't have time for at least a week. Happy for anyone else to take it over if they get there before me.

Aside, it turns out the `local_repository` thing I switched to using isn't available in Bazel 6, so that might be blocked on https://github.com/bazelbuild/rules_rust/pull/2812. I'll switch it back to the manual re-implementation for now.

I still have no resolution for the `repo_mapping` thing. It's not supported in Bzlmod, but may break non-bzlmod users who have a different version of zlib installed. The best solution would be for Bzlmod to stop erroring on that parameter and just ignore it. That would be an upstream change.

AlexOrozco1256:
@jwnrt Hello James, is there any plan to fix this issue? Has a bug been filed with bazelbuild for the repo_mapping issue? Thanks!

bindgen/transitive_repositories.bzl
jwnrt: I've made the `repo_mapping` parameter optional and disabled it when using the Bzlmod extension. Even if Bazel is updated to silently ignore the parameter under Bzlmod, `rules_rust` would have to increase its minimum supported Bazel version to avoid errors.

bindgen/extension.bzl
jwnrt: Reverted that, added the `build_file_content` parameter.

bindgen/extension.bzl
jwnrt: Reverted the `local_repository` change since it's only available in more recent Bazel versions.

Added the docs, partially copied from `local_repository`.

jwnrt:
Thanks for the reminder, I've updated the PR to resolve those issues.

Based on https://github.com/bazelbuild/bazel/issues/21846 it seems to me that Bazel are moving away from supporting `repo_mapping` entirely. Even if upstream Bazel stops erroring on `repo_mapping` for Bzlmod, `rules_rust` would have to greatly increase its minimum supported Bazel version to benefit.

I've worked around it by making `repo_mapping` optional (but used by default). I've documented it as being unstable / subject to change.

ahans:
@jwnrt Thanks for your effort, sounds like a good solution to me. @UebelAndre Could you take another look, so we can merge this PR? The issue blocks the transition of [rules_ros2](https://github.com/mvukov/rules_ros2) to [bzlmod](https://github.com/mvukov/rules_ros2/pull/238).

AlexOrozco1256:
@jwnrt Thank you!! Sorry it took me a while to get back but I appreciate your help!

hlopko:
(just a drive by that this is an exciting change and it would be great to have it merged - thanks @jwnrt and @UebelAndre)

ajwerner:
I also keep rebasing this because I need it. Anything I can do to help get this merged?

AlexOrozco1256:
Hello, is there anything I can do to help get this merged? I've received inquiries from other teams that would like to use bindgen + bzlmod. Thanks!

PR #2741, opened 2024-07-13
Support incompatible enable proto toolchain resolution

illicitonion:
This allows using a shared (generally pre-built) protoc definition, rather than needing to supply one via a label.

We allow (but deprecate) users to still bring along their own protoc as a label, but if they want to do so, they will need to ensure they have the correct repositories registered in order to build/bring their protoc, rather than us setting them up by default.


illicitonion:
The docs CI failure should be fixed by https://github.com/bazel-contrib/bazel_features/pull/62 - otherwise this should be good to go.

alexeagle:
bazel_features release is out, so this should be unblocked again!
thanks!

test/deps.bzl
```diff
@@ -63,10 +63,19 @@ def rules_rust_test_deps():
         sha256 = "b8c487191eb942361af905e40172644eab490190e717c3d09bf83e87f3994fff",
     )
 
+    maybe(
+        http_archive,
+        name = "rules_python",
```
UebelAndre, line 68: Why is python needed?

UebelAndre (approved):
Looks good to me!

PR #2742, opened 2024-07-15
Remove failure when bindgen cc_lib does not contain static library

### LINKED ISSUE #2704: rust_bindgen should not require compile_data
`rust_bindgen` currently asserts that the cc_lib has an associated static library to link.  This is problematic for use cases where this cannot be satisfied.  

Example use cases:
* 3rd party vendor provides a pre-compiled shared object and header only (used via cc_library + cc_import)
* generating rust constants from a header file without any code to link to

Ideally rust_bindgen should just be a code generator that does not impose any additional requirements not needed by bindgen itself, or a separate un-opinionated interoperable code generation rule should be provided.

jwnrt:
I don't think this failure is required: it's fine to bindgen a library with only headers and no static library. Should this error instead be triggering when the `cc_library` has non-header files but isn't static?

Fixes #2704


hlopko:
Thanks @jwnrt, the PR makes sense. Would you mind adding a separate example in examples/bindgen (for example a header that is used from both Rust and C++, and both libraries meet in a common binary) showing everything links together well?

jwnrt:
Sure, I've added an example, let me know if it matches what you expected

jwnrt:
Fixed those formatting problems

jwnrt:
Friendly bump: this is ready for review again. We can apply this as a patch downstream, so no rush from me.

PR #2743, opened 2024-07-15
Add support for --wrap-extern-fns to bindgen rules

hlopko:


krasimirgg (approved):
Thank you! (the output_group trick is cool!)

PR #2191, opened 2023-10-10
Use canonical labels for analysistest config_settings

nickgooding:
With bzlmod enabled, errors like the following would occur:
```
Error in analysis_test_transition: invalid transition output '@[unknown repo '' requested from @bazel_skylib~1.3.0]//:clippy_flags': no repo visible as @ from repository '@bazel_skylib~1.3.0'
```

To resolve this, use str(Label("...")) to get the canonical label to use as the key in config_settings.

Related issues: #2181, bazelbuild/bazel#19286

PR #2744, opened 2024-07-16
Make rustfmt toolchain optional in bindgen

hlopko:


bindgen/private/bindgen.bzl
```diff
@@ -375,10 +375,11 @@ rust_bindgen = rule(
     outputs = {"out": "%{name}.rs"},
     fragments = ["cpp"],
     toolchains = [
-        str(Label("//bindgen:toolchain_type")),
-        str(Label("//rust:toolchain_type")),
-        str(Label("//rust/rustfmt:toolchain_type")),
-        "@bazel_tools//tools/cpp:toolchain_type",
+        config_common.toolchain_type("//bindgen:toolchain_type"),
+        config_common.toolchain_type("//rust:toolchain_type"),
+        config_common.toolchain_type("//rust/rustfmt:toolchain_type"),
```
krasimirgg, line 380: 1. This line now looks redundant with the next one?
2. I'm not sure if the removal of `str(Label(...))` is consistent with bzlmod. IIRC, it's here for some bzlmod support, e.g., https://github.com/bazelbuild/rules_rust/pull/2191. Could you confirm or else add this canonization back? 
3. Optionally, the whole `config_common.toolchain_type` is very nice, could you file a bug or prep a PR to update all such instances?

krasimirgg (changes requested):

bindgen/private/bindgen.bzl
hlopko: (1) - right, fixed.
(2) - I believe it's fine, I see it used in other rules.
(3) - will do :)

krasimirgg (approved):

PR #2749, opened 2024-07-18
Add fake dependency roots for all transitive proc-macros

illicitonion:
See the extensive comment for details, but tl;dr: when cross-compiling,
just doing top-level resolves of direct dependencies can miss resolving
platform-specific features for proc-macros which are only used on
certain platforms.


UebelAndre (approved):
Nice! Thank you!

Issue #2368, opened 2023-12-28
Explicitly tracked lists of generated dependencies in MODULE.bazel


UebelAndre:
I just came across https://github.com/bazelbuild/rules_rust/blob/0.35.0/MODULE.bazel#L35-L124 and think this is a pretty egregious regression in maintainability for me. Any dependency here that's generated by a `crate_universe` rule needs to be updatable by the same rule or some other `.bzl` file rendered to contain the appropriate list.

UebelAndre:
cc @illicitonion @matts1

illicitonion:
From https://github.com/bazelbuild/rules_rust/pull/2314:

> Note that the MODULE.bazel changes in use_repo are autogenerated, and if you change the third party crates you depend on, you should get an error message saying something like "run this command to fix it" on your bzlmod presubmit environment.

I didn't test this out - did you run into this? What happened?

UebelAndre:
> I didn't test this out - did you run into this? What happened?

```
(13:20:44) WARNING: /workdir/MODULE.bazel:35:30: The module extension internal_deps defined in @rules_rust//rust/private:extensions.bzl reported incorrect imports of repositories via use_repo():
 
Imported, but not created by the extension (will cause the build to fail):
    rules_rust_bindgen__bindgen-0.65.1, rules_rust_bindgen__bindgen-cli-0.65.1
 
Not imported, but reported as direct dependencies by the extension (may cause the build to fail):
    rules_rust_bindgen__bindgen-0.69.1, rules_rust_bindgen__bindgen-cli-0.69.1
 
 ** You can use the following buildozer command to fix these issues:
 
buildozer 'use_repo_add @rules_rust//rust/private:extensions.bzl internal_deps rules_rust_bindgen__bindgen-0.69.1 rules_rust_bindgen__bindgen-cli-0.69.1' 'use_repo_remove @rules_rust//rust/private:extensions.bzl internal_deps rules_rust_bindgen__bindgen-0.65.1 rules_rust_bindgen__bindgen-cli-0.65.1' //MODULE.bazel:all
```
I don't have buildozer installed. I also don't think relying on buildozer is a good solution. I intended to update dependencies by running a `crates_vendor` target and that should have been all I needed to do. Can this list be rendered into another `bzl` file and loaded in `MODULE.bazel` instead?

illicitonion:
[...]

PR #2751, opened 2024-07-19
crate_universe: Only compile label regex once.

criemen:
I hit this when profiling for something else, we spent 1.1sec of 1.3sec of `cargo-bazel generate` compiling this regular expression. It seems to be quite bad on the regex compiler, as even compiling it once takes 8ms.

I had to add the `once_cell` crate to unwrap the Result type - the corresponding stdlib function is still nightly-only unfortunately. However, that crate was already in the transitive dependency tree, so it's a net-same change.


criemen:
Can someone please push the correct, re-vendored dependencies here? I spent now 30min trying to get this to work locally, but it's not working for me.

criemen:
Revendoring the dependencies helped with some CI jobs, but the BCR jobs still complain about a lockfile being out of date. I ran the suggested command locally in `examples/bzlmod/hello_world`, but that doesn't seem to satisfy CI. Can someone help me out here?

crate_universe/Cargo.toml
```diff
@@ -49,6 +49,7 @@ textwrap = "0.16.0"
 toml = "0.8.10"
 tracing = "0.1.40"
 tracing-subscriber = "0.3.17"
+once_cell = "1.19.0"
```
UebelAndre, line 52: nit, can you sort this into the list above?

UebelAndre (changes requested):
Thank you for this! I think there are some extra things you'll need manually to update on the bzlmod side of this. See https://github.com/bazelbuild/rules_rust/issues/2368

criemen:
Thanks! CI finally passed on this.
That was harder than expected given the size of the core change...

UebelAndre (approved):
Thank you! For this change I probably would have just regenerated dependencies for `crate_universe` directly but thank you for going above and beyond here 😄

PR #2765, opened 2024-07-28
Respect "--no-typescript" flag in wasm_bindgen

duarten:
Don't expect typescript outputs if this flag is set.

### DIFF
--- a/wasm_bindgen/private/wasm_bindgen.bzl
+++ b/wasm_bindgen/private/wasm_bindgen.bzl
@@ -53,11 +53,14 @@ def rust_wasm_bindgen_action(ctx, toolchain, wasm_file, target_output, bindgen_f
     bindgen_wasm_module = ctx.actions.declare_file(ctx.label.name + "_bg.wasm")
 
     js_out = [ctx.actions.declare_file(ctx.label.name + ".js")]
-    ts_out = [ctx.actions.declare_file(ctx.label.name + ".d.ts")]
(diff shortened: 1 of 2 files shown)


wasm_bindgen/private/wasm_bindgen.bzl
```diff
@@ -53,11 +53,14 @@ def rust_wasm_bindgen_action(ctx, toolchain, wasm_file, target_output, bindgen_f
     bindgen_wasm_module = ctx.actions.declare_file(ctx.label.name + "_bg.wasm")
 
     js_out = [ctx.actions.declare_file(ctx.label.name + ".js")]
-    ts_out = [ctx.actions.declare_file(ctx.label.name + ".d.ts")]
+    ts_out = []
+        if not "--no-typescript" in bindgen_flags:
+            ts_out.append(ctx.actions.declare_file(ctx.label.name + ".d.ts"))
```
illicitonion, line 58:
```suggestion
    if not "--no-typescript" in bindgen_flags:
        ts_out.append(ctx.actions.declare_file(ctx.label.name + ".d.ts"))
```

illicitonion (commented):
LGTM, could you add a test/example which sets `--no-typescript` which we can use to validate this keeps working in the future? Thanks!

duarten:
I added the test/example :)

illicitonion (approved):

PR #2775, opened 2024-08-02
Add virtual deps to package, not virtual workspace root

illicitonion:
Otherwise on repinning, we get errors because virtual workspaces can't have deps.

Repinning the multi_package example hit this problem.


UebelAndre (approved):

UebelAndre:
Should there be a new release for this?

illicitonion:
> Should there be a new release for this?

Yeah, I'll kick one off :) Thanks!

PR #2776, opened 2024-08-02
Release 0.49.1

illicitonion:


version.bzl
```diff
@@ -1,3 +1,3 @@
 """The version of rules_rust."""
 
-VERSION = "0.49.0"
+VERSION = "0.50.0"
```
UebelAndre, line 3:
```suggestion
VERSION = "0.49.1"
```

MODULE.bazel
```diff
@@ -2,7 +2,7 @@
 
 module(
     name = "rules_rust",
-    version = "0.49.0",
+    version = "0.50.0",
```
UebelAndre, line 5:
```suggestion
    version = "0.49.1",
```

UebelAndre (changes requested):
Can we make this a patch release? I like to use these to suggest to users there is a problem with a particular release.

UebelAndre (approved):

UebelAndre:
> Can we make this a patch release? I like to use these to suggest to users there is a problem with a particular release.

Apparently I have that power! Bwahaha (hope you don't mind 😅)

PR #2794, opened 2024-08-15
Ensure proto compiler attribute is consistent in prost_toolchain

UebelAndre:
The deprecated attribute was passing a `Target` where the `proto_toolchain` is passing a `FilesToRun` provider. This ensures the two are consistent.


illicitonion (approved):
Thanks!

PR #2803, opened 2024-08-21
Build `rust_test` targets using a crate name different from the underlying lib

felipeamp:
This PR also makes `rust_test` put its compilation outputs in the same directory as the `rust_library` rule (i.e. not in a `test-{hash}` subdirectory anymore).

After this change both the `rust_library` and `rust_test` rules will put all its compilation outputs in the same directory, but there won't be any name collisions in non-sandboxed environments (see https://github.com/bazelbuild/rules_rust/pull/1427 for more context).

This is a partial rollback of https://github.com/bazelbuild/rules_rust/commit/10185339ddd5db93a7ca4b10e2255e97c81f99cb and https://github.com/bazelbuild/rules_rust/commit/26344d4cd7bca958ab289eda3ac4c24f3a36fbeb.


scentini (approved):
Thanks!

PR #2804, opened 2024-08-21
Support generated files with sibling repo layout.

yuzhy8701:
When using `experimental_sibling_repository_layout`, `file.path` won't contain a segment matching `workspace_root`, as the workspace name is encoded in the `root` path instead. In this change we first remove the root path prefix from each path. This allows patching the toolchain sysroot with rules like `copy_file()`, etc.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -326,12 +326,16 @@ def _symlink_sysroot_tree(ctx, name, target):
         # Parse the path to the file relative to the workspace root so a
         # symlink matching this path can be created within the sysroot.
 
+        file_path = file.path
+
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Thanks! Can you show the before and after of a toolchain as well as add a regression test where a toolchain is created with some fake generated files? We have some similar tests here https://github.com/bazelbuild/rules_rust/tree/0.49.3/test/unit/toolchain

yuzhy8701:
Since the sibling repo layout is being dropped (https://github.com/bazelbuild/bazel/issues/23576), this fix no longer make a lot of sense. I'm going to close it for now.

PR #2812, opened 2024-08-26
Update min bazel version

scentini:
Bazel `6.4` was released in, Oct 2023, we're well beyond the support window.

I updated straight to `7.3.1` because according to the [docs](https://github.com/bazelbuild/bazel/releases/tag/7.3.1) it's fully compatible with `7.0`:

> Bazel 7.3.1 is a patch LTS release. It is fully backward compatible with Bazel 7.0 and contains selected changes by the Bazel community and Google engineers.


UebelAndre:
> well beyond the support window

Where is the support window documented? I feel like the right move is to go to 7.0 but if there is some existing guidance on min supported versions then I'd say "fire away" :D

illicitonion:
Per https://github.com/bazelbuild/rules_rust/issues/415#issuecomment-841157191 I think it's fine to drop 6.x support.

I'm also ok with us jumping straight to 7.3.1, because none of our users should be forced to be on an older 7.x release (they should be able to upgrade without worry), but think it's also fine if we want to support 7.0.

illicitonion (approved):

Issue #2818, opened 2024-08-28
Provide a mechanism for disabling toolchain registration when using bzlmod.


konkers:
Some projects want to vendor and tightly control their toolchains.  With `rules_rust` always registering toolchains when used through blzmod, it is easy to misconfigure a project's toolchain configuration and accidentally fallback on the `rules_rust` toolchains without knowing.  Providing a way to explicitly disable the `rules_rust` toolchains would turn this case into and error instead of a silent fallback.

konkers:
#2819 Implements a mechanism for inhibiting the registration of default rust toolchains.    It does not address:
* proto toolchain
* prost toolchain
* bindgen toolchain
* host_tools repository
* cargo_bazel_bootstrap repository

proto, prost, and bindgen toolchains are directly referenced instead of going through an extension.   They would probably need to follow the same pattern as rust_toolchains to allow for disabling their registration.

The host_tools and cargo_bazel_bootstrap repositories are a bit trickier as rules_rust references these repos internally.

PR #2819, opened 2024-08-28
Allow bzlmod users to inhibit toolchain registration

konkers:
Partial fix for #2818

### DIFF
--- a/rust/extensions.bzl
+++ b/rust/extensions.bzl
@@ -31,6 +31,17 @@ def _find_modules(module_ctx):
 
     return root, our_module
 
+def _empty_repository_impl(repository_ctx):
+    repository_ctx.file("WORKSPACE.bazel", """workspace(name = "{}")""".format(
+        repository_ctx.name,
+    ))
+    repository_ctx.file("BUILD.bazel", "")
+
+_empty_repository = repository_rule(
+    doc = ("Declare an empty repository."),
+    implementation = _empty_repository_impl,
+)
+
 def _rust_impl(module_ctx):
     # Toolchain configuration is only allowed in the root module, or in
     # rules_rust.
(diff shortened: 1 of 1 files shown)


rust/extensions.bzl
```diff
@@ -40,18 +51,25 @@ def _rust_impl(module_ctx):
...
-        rust_register_toolchains(
-            dev_components = toolchain.dev_components,
-            edition = toolchain.edition,
-            allocator_library = toolchain.allocator_library,
-            rustfmt_version = toolchain.rustfmt_version,
-            rust_analyzer_version = toolchain.rust_analyzer_version,
-            sha256s = toolchain.sha256s,
-            extra_target_triples = toolchain.extra_target_triples,
-            urls = toolchain.urls,
-            versions = toolchain.versions,
```
krasimirgg, line 52: Why not modify the implementation of `rust_register_toolchains` to not register any toolchains when passed `versions = []`? This way we avoid the extra check and _empty_repository setup.

konkers: There did not seem to be a clean way to get an empty list passed as `toolchain_names` to `toolchain_repository_hub` here (https://github.com/bazelbuild/rules_rust/blob/main/rust/repositories.bzl#L252) without nesting most of that function under an `if len(versions) > 0:`.   `versions` is not considered until inside calls to `rust_repository_set` and `_get_toolchain_repositories`.   I could add the following at the beginning of `rust_register_toolchains` to avoid having to declare `_empty_repository`: 

```
if len(versions) == 0:
    toolchain_repository_hub(
        name = "rust_toolchains",
        toolchain_names = [],
        toolchain_labels = {},
        toolchain_types = {},
        exec_compatible_with = {},
        target_compatible_with = {},
    )
```

Other refactor suggestions are welcome.

krasimirgg: Thanks! I think what you have now it cleaner than the `toolchain_repository_hub` approach.

krasimirgg (approved):

Issue #2827, opened 2024-09-02
Put compilation outputs of `rust_library` and `rust_test` rules in same directory without conflicts


felipeamp:
Currently the compilation outputs of a `rust_library` are put in that specific directory. For example, one of the compilation outputs of `my/package:my_library` will be the `my_library` file in `my/package`.

However, the compilation outputs of `rust_test` rules are put in a specific test subdirectory containing a hash in its name. For example, the compilation output of `my/package:my_test`, which tests `my/package:my_library`, will contain the `my_library` file in `my/package/test-<hash_id>`.

This was done as a way to unblock the flaky/failing tests of https://github.com/bazelbuild/rules_rust/pull/1427.

In order to have all compilation outputs in the same directory (i.e. without the `test-<hash_id>` subdirectory for test rules) we will add a `_test` suffix to the crate name of `rust_test` targets. This is a backward-incompatible change, since if the `crate_name` attribute is set in code (e.g. `#![crate_name = "my_crate"]`), we need to set the `crate_name` attribute in the `rust_test` rule to match it.

Therefore this change will be done behind a feature flag.

Issue #2082, opened 2023-07-27
`experimental_use_coverage_metadata_files`


UebelAndre:
`experimental_use_coverage_metadata_files` is a flag to have coverage tooling added as [coverage_common.instrumented_files_info.metadata_files](https://bazel.build/versions/6.3.0/rules/lib/coverage_common?hl=en#instrumented_files_info) (taking advantage of https://github.com/bazelbuild/bazel/issues/18697) instead of reporting tools like `llvm-cov` and `llvm-profdata` as runfiles to each test. To enable this flag pass the following:
```
--@rules_rust//rust/settings:experimental_use_coverage_metadata_files
```

PR #2828, opened 2024-09-02
Build `rust_test` targets using a crate name different from the underlying lib

felipeamp:
This is a rollforward of
https://github.com/bazelbuild/rules_rust/pull/2803, but behind the `incompatible_change_rust_test_compilation_output_directory` incompatible flag.

This PR also makes `rust_test` put its compilation outputs in the same directory as the `rust_library` rule (i.e. not in a `test-{hash}` subdirectory anymore).

After this change both the `rust_library` and `rust_test` rules will put all its compilation outputs in the same directory, but there won't be any name collisions in non-sandboxed environments (see https://github.com/bazelbuild/rules_rust/pull/1427 for more context).

Issue with context: https://github.com/bazelbuild/rules_rust/issues/2827


test/unit/rust_test_outputs_are_in_subdirectory/BUILD.bazel
UebelAndre, line 1: Can you replace these tests with another set that demonstrates the expected behavior?

test/rust/src/lib.rs
```diff
@@ -12,5 +12,4 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![crate_name = "hello_lib"]
```
UebelAndre, line 15: Why remove this?

UebelAndre (changes requested):
Thanks! Had a few questions.

test/unit/rust_test_outputs_are_in_subdirectory/BUILD.bazel
felipeamp: Done. I've also undeleted this directory, since it still tests current behavior (when the incompatible flag is not set/set to False).

UebelAndre (changes requested):
Thanks! Could I also get you to make (or modify an existing) issue ticket to track the new incompatibility flag? We have some guidance on https://github.com/bazelbuild/rules_rust/blob/0.49.3/COMPATIBILITY.md#how-to-make-a-backwards-incompatible-change but since this could be a breaking change we try to give users a chance to raise concerns during the rollout. I've been writing mine to look like https://github.com/bazelbuild/rules_rust/issues/2082

felipeamp:
I have filed an issue for this, see https://github.com/bazelbuild/rules_rust/issues/2827. Is there anything specific that's missing there?

felipeamp:
I've also fixed the windows tests (that weren't working due to the toolchain binary extension) and added an assert on the directory name as well.

UebelAndre:
Just gotta resolve the conflicts!

felipeamp:
All done.

PR #2836, opened 2024-09-05
[bug] fix coverage linker flags when using --@rules_rust//rust/settin…

Ryang20718:
When running bazel coverage with `--@rules_rust//rust/settings:experimental_use_cc_common_link=true`, we fail to generate coverage properly because of some missing linker flags (the c library as well as some missing symbols)

This PR fixes that

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1432,6 +1432,12 @@ def rustc_compile_action(
             # Append the name of the library
             output_relative_to_package = output_relative_to_package + output_lib
 
+        coverage_link_flags = []
+        if include_coverage:
+            # coverage fails due to gcc lacking a few linker args
+            # this fixes missing symbols and links against the c lib
+            coverage_link_flags = ["-u", "__llvm_profile_runtime", "-lc"]
+
         cc_common.link(
             actions = ctx.actions,
             feature_configuration = feature_configuration,
@@ -1441,6 +1447,7 @@ def rustc_compile_action(
             name = output_relative_to_package,
             stamp = ctx.attr.stamp,
             output_type = "executable" if crate_info.type == "bin" else "dynamic_library",
+            user_link_flags = coverage_link_flags,
         )
 
         outputs = [crate_info.output]


.bazelci/presubmit.yml
```diff
@@ -565,6 +572,8 @@ tasks:
       - "//..."
     test_targets:
       - "//..."
+    coverage_targets:
```
UebelAndre, line 575: I think you'll need to add `coverage_flags` here.

UebelAndre (changes requested):
Thank you! I think there's one change left to get passing CI.

.bazelci/presubmit.yml
Ryang20718: 😓 I thought `test_flags` are inherited as part of coverage?

on main, without my change, `bazel coverage` is broken for `Build with no_std + alloc using cc_common.link infrastructure for linking`

UebelAndre: This is true of `.bazelrc` flags but I don't think bazel CI works that way. Perhaps I'm misinterpreting https://github.com/bazelbuild/continuous-integration

.bazelci/presubmit.yml
```diff
@@ -565,6 +572,8 @@ tasks:
       - "//..."
     test_targets:
       - "//..."
+    # coverage_targets: # TODO: fix https://github.com/bazelbuild/rules_rust/issues/2838
```
UebelAndre, line 575: Isn't this used for regression testing?

UebelAndre (changes requested):

.bazelci/presubmit.yml
Ryang20718: correct, but bazel coverage doesn't work on main with this target (even without my changes)

on main, we have 3 broken coverage scenarios whereas with my pr, we'll have 1 broken coverage scenario

.bazelci/presubmit.yml
Ryang20718: are there `coverage flags`? https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md

I don't see any mention of that?

rust/private/rustc.bzl
```diff
@@ -1432,6 +1432,11 @@ def rustc_compile_action(
             # Append the name of the library
             output_relative_to_package = output_relative_to_package + output_lib
 
+        coverage_link_flags = []
+        if include_coverage:
+            # fixes missing symbols coverage fails due to gcc lacking a few linker args
+            coverage_link_flags = ["-u", "__llvm_profile_runtime"]
```
UebelAndre, line 1438: What happens if you're not using gcc?

UebelAndre (changes requested):

.bazelci/presubmit.yml
UebelAndre: Coverage is not very well documented in the Bazel ecosystem. I remember seeing this deep in the python code some time ago but don't remember exactly where.

Pryz:
Looks like I'm might be hitting this bug. Any chance we can move this PR forward? Thanks :)

Issue #2801, opened 2024-08-20
Support for `--incompatible_use_plus_in_repo_names`


criemen:
Bazel 7.3.0 introduced `--incompatible_use_plus_in_repo_names` which changes the bzlmod separator from ~ to + (for performance reasons on Windows). This flag will be flipped with 8.0.0.
Flipping that flag currently is incompatible with `rules_rust`, as (at least) the `crate_universe` code assumes that the separator is ~.
Im getting
```
ERROR: Traceback (most recent call last):
        File "/private/var/tmp/_bazel_criemen/267ec2c545f0139b2caed9502387afe1/external/rules_rust+/crate_universe/extension.bzl", line 369, column 37, in _crate_impl
                _generate_hub_and_spokes(module_ctx, cargo_bazel, cfg, annotations, cargo_lockfile = cargo_lockfile, manifests = manifests)
        File "/private/var/tmp/_bazel_criemen/267ec2c545f0139b2caed9502387afe1/external/rules_rust+/crate_universe/extension.bzl", line 132, column 16, in _generate_hub_and_spokes
                cargo_bazel(splice_args)
        File "/private/var/tmp/_bazel_criemen/267ec2c545f0139b2caed9502387afe1/external/rules_rust+/crate_universe/private/module_extensions/cargo_bazel_bootstrap.bzl", line 66, column 17, in run
                fail("%s returned with exit code %d:\n%s" % (pretty_args, result.return_code, result.stderr))
[...]

PR #2844, opened 2024-09-06
crate_universe: Don't preform path translation on `out_dir`

konkers:
Bug: #2801


krasimirgg:
Thank you! Could you also set up a new presubmit running some of the blzmod examples with `--incompatible_use_plus_in_repo_names`.

konkers:
Apologies for the noise.  Trying to get CI working.

konkers:
OK churn done.  I have added tests using `--incompatible_use_plus_in_repo_names`.   This was a bit complicated by the fact that

1. The command line flag does not exist in the bazel version use by the test by default
2. The repo names leak into the lock file causing `--lockfile_mode=error` to fail

krasimirgg (approved):
Thank you!

PR #2850, opened 2024-09-10
Prune `known_shas.bzl` to only include versions used in BazelCI

UebelAndre:


UebelAndre:
To correct the description, the versions retained are just anything greater than the min tested version.

PR #2853, opened 2024-09-10
Error on absolute src paths

illicitonion:
These lead to non-deterministic lockfile contents because they depend on the depth of the directory the repo is checked out into.

This may possibly introduce false positives because of hard-coded /dev/null or similar in actual published crates - hopefully it doesn't (ideally people aren't publishing this kind of thing to crates.io), but if it does, we can loosen those to a warning or add a per-crate opt-out or something.


UebelAndre (approved):
Thanks!

If you get a chance before merging can you bump the crate_universe version for this?

PR #2854, opened 2024-09-10
Improve crate_universe debug logging for `cargo tree` parsing

UebelAndre:
It's much easier to understand what's going on if I can see a standard `cargo tree` output.


illicitonion (approved):
Thanks!

PR #2856, opened 2024-09-11
Propagate `cargo_build_script.data` to `Rustc` compile actions

UebelAndre:
This way any outputs that the build script may refer to in flags it emits can be referenced without needing to also explicitly pass the data target to the `compile_data` attribute of it's consumer.

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -433,7 +433,7 @@ def _cargo_build_script_impl(ctx):
             flags = flags_out,
             linker_flags = link_flags,
             link_search_paths = link_search_paths,
-            compile_data = depset(),
+            compile_data = depset(transitive = script_data),
         ),
         OutputGroupInfo(
             streams = depset([streams.stdout, streams.stderr]),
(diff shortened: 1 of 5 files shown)


illicitonion (approved):

goffrie:
Is there any way to opt _out_ of this? We have some rules where we want data to _only_ be provided to the build script (admittedly the situation is a pile of hacks, and this is just what broke the camel's back).

UebelAndre:
> Is there any way to opt _out_ of this? We have some rules where we want data to _only_ be provided to the build script (admittedly the situation is a pile of hacks, and this is just what broke the camel's back).

If the data is only for the build script would assigning it to `tools` do the trick for you? That shouldn't be propagated beyond the build script
https://github.com/bazelbuild/rules_rust/blob/0f0b354380d15046ee3852e79c7c4475f4e589bb/cargo/private/cargo_build_script_wrapper.bzl#L112

I'm otherwise not opposed to an incompatibility flag, but when making this change I thought that would cover such a use case.

illicitonion:
I'm not sure an incompatibility flag would make sense for this one long-term, but an extra attr on `cargo_build_script` for `non_propagated_data` or something may?

goffrie:
`tools` is built for the exec configuration unfortunately, which does matter for my use case. I agree that adding an incompatibility flag doesn't make much sense, but adding yet another attr seems unfortunate too :/

PR #2858, opened 2024-09-11
[bug] only generate rust libraries with coverage instrumentation only when coverage is enabled

Ryang20718:
Currently, binaries produced by rust_binary generate profraw files which means the binary is built with code coverage instrumentation with bazel build. We should only generate coverage with bazel coverage. This PR fixes this bug


UebelAndre (changes requested):
Thanks! Is this something you can add an analysis test for? Maybe add a `coverage` directory to https://github.com/bazelbuild/rules_rust/tree/0.50.0/test/unit ?

PR #2866, opened 2024-09-12
Add wasm64 support.

freeformstu:
This PR adds support for `wasm64-unknown-unknown` although there are no prebuilt `wasm64-unknown-unknown` rustc toolchains. Compiling for `wasm64` will be left up to anyone using this target triple until we get T2 support for wasm64.

### DIFF
--- a/rust/platform/platform.bzl
+++ b/rust/platform/platform.bzl
@@ -67,9 +67,24 @@ def declare_config_settings():
         )
 
     native.platform(
-        name = "wasm",
+        name = "wasm32",
         constraint_values = [
             "@platforms//cpu:wasm32",
+            str(Label("//rust/platform/os:unknown")),
+        ],
+    )
+
+    # Add alias for wasm to maintain backwards compatibility.
+    native.alias(
+        name = "wasm",
+        actual = ":wasm32",
+        deprecation = "Use `@rules_rust//rust/platform:wasm32` instead",
+    )
+
+    native.platform(
+        name = "wasm64",
+        constraint_values = [
+            "@platforms//cpu:wasm64",
             str(Label("//rust/platform/os:unknown")),
         ],
     )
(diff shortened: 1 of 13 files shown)


rust/platform/platform.bzl
```diff
@@ -67,13 +67,21 @@ def declare_config_settings():
         )
 
     native.platform(
-        name = "wasm",
+        name = "wasm32",
```
UebelAndre, line 70: Can you add an alias for backward compatibility?

rust/private/dummy_cc_toolchain/BUILD.bazel
```diff
@@ -1,20 +1,30 @@
+"""When compiling Rust code for wasm (wasm32 or wasm64), we avoid linking to cpp code so we introduce a dummy cc
+toolchain since we know we'll never look it up.
+
+TODO(jedmonds@spotify.com): Need to support linking C code to rust code when compiling for wasm32 or wasm64.
```
UebelAndre, line 4: What's this TODO and who is this?

UebelAndre (changes requested):

rust/private/dummy_cc_toolchain/BUILD.bazel
freeformstu: I moved this up from the comment below.

UebelAndre:
I think the failing windows example is Ok to disable with a reference to:
- https://github.com/bazelbuild/rules_rust/issues/2878

UebelAndre (changes requested):
`wasm64-unknown-unknown` doesn't have a `std` artifact according to https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-3

This would lead to failures in analysis (failing to fetch an artifact that doesn't exist) should anything match that platform. I think it should be removed as a default platform and users can use `crates_repository.supported_platform_triples` if they were to have a usable toolchain.

UebelAndre (approved):

PR #2869, opened 2024-09-13
Added tests for `runfiles::parse_repo_mapping`

UebelAndre:


UebelAndre:
Admittedly I am not familiar with the `_repo_mapping` files and so have only written tests for what I found in my `bazel-out` directory.

illicitonion (approved):

Issue #1156, opened 2022-02-27
Fully support windows without --enable_runfiles


UebelAndre:
`rules_rust` currently has some components that require [--enable_runfiles](https://bazel.build/reference/command-line-reference#flag--enable_runfiles) and [--windows_enable_symlinks](https://bazel.build/reference/command-line-reference#flag--windows_enable_symlinks) to be set on windows. This should not be needed and all rules should work without requiring any flags.

PR #2871, opened 2024-09-14
Update `cargo_build_script` to work without runfiles support.

UebelAndre:
Partially addresses https://github.com/bazelbuild/rules_rust/issues/1156


cargo/private/runfiles_enabled.bzl
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    runfiles_enabled_setting(
+        name = name,
+        value = select({
+            # If either of the runfiles are set, use the flag
+            ":{}_enable_runfiles".format(name): True,
+            ":{}_disable_runfiles".format(name): False,
+            # Otherwise fall back to the system default.
+            "@platforms//os:windows": False,
```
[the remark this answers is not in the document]
UebelAndre: It complained when you actually specified `--noenable_runfiles`. I've updated this to work on windows and unix systems where the flag is enabled, disabled, and unset.

UebelAndre:
> Works for me!
> 
> cc @fmeum just as an FYI in case there's anything you think should live more upstream somewhere

It would be great because I made this a lot more complicated lol

cargo/private/cargo_build_script.bzl
```diff
@@ -208,16 +242,35 @@ def _cargo_build_script_impl(ctx):
...
+    for target in script_info.data:
+        script_data.append(target[DefaultInfo].files)
+        script_data.append(target[DefaultInfo].default_runfiles.files)
+    for target in script_info.tools:
+        script_tools.append(target[DefaultInfo].files)
+        script_tools.append(target[DefaultInfo].default_runfiles.files)
+
+    workspace_name = ctx.label.workspace_name
+    if not workspace_name:
+        workspace_name = ctx.workspace_name
```
fmeum, line 266: See above, this can be simplified.

cargo/private/runfiles_enabled.bzl
```diff
@@ -0,0 +1,167 @@
...
+    return {
+        _RUNFILES_ENABLED_ATTR_NAME: attr.label(
+            doc = "A flag representing whether or not runfiles are enabled.",
+            providers = [RunfilesEnabledInfo],
+            default = default,
+            cfg = "exec",
+        ),
+    }
+
+def runfiles_enabled_build_setting(name, **kwargs):
```
fmeum, line 42: Nice hack ;-)

I think that this can be simplified since you don't need a single build setting to match both the platform and the value of `--enable_runfiles`. This currently also matches the target platform, but the default is based on the host platform.

How about only matching the flag here and then using `ctx.configuration.host_path_separator` to filter out a Windows host in `runfiles_enabled_build_setting`?

cargo/private/runfiles_maker/runfiles_maker.rs
```diff
@@ -0,0 +1,71 @@
+//! A tool for building runfiles directories for Bazel environments that don't
+//! support runfiles or have runfiles enabled.
```
fmeum, line 2: enabled --> disabled

cargo/private/runfiles_enabled.bzl
UebelAndre: This took an embarrassing amount of brain power 😅 , could I get you to slice and dice this code to be the simpler form?

UebelAndre:
This change ended up being more complicated and needs another review but I forgot to disable auto-merge. Revert is here https://github.com/bazelbuild/rules_rust/pull/2885

PR #2887, opened 2024-09-19
Update `cargo_build_script` to work without runfiles support.

UebelAndre:
Partially addresses https://github.com/bazelbuild/rules_rust/issues/1156

This pull-request implements an additional change to enable this behavior which aggregates all transitive `BuildInfo.compile_data` into `Rustc` actions. While this seems to bloat these actions with unnecessary data, it addresses a catastrophic flaw in how Windows works at all.

As of Bazel 7.3.1, Windows does not run any actions in a sandbox (https://github.com/bazelbuild/bazel/discussions/18401), this means that references to the current working directory will be consistent since they always refer to the execroot. On top of this fact, `cargo_build_script` will assign [CARGO_MANIFEST_DIR](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) to a [path within the runfiles directory of the build script](https://github.com/bazelbuild/rules_rust/blame/d9ee6172d1dfc2801ba809441668e60e797f89de/cargo/private/cargo_build_script.bzl#L218). The combination of these two facts leads crates like [windows_x86_64_msvc](https://crates.io/crates/windows_x86_64_msvc), which assign a linker path using `CARGO_MANIFEST_DIR` ([@windows_x86_64_msvc//build.rs](https://github.com/microsoft/windows-rs/blob/0.59.0/crates/targets/x86_64_msvc/build.rs#L1-L8)) to introduce un-tracked dependencies into dependent `Rustc` actions. This then leads to build failures if the `windows_x86_64_msvc` crate is ever a remote cache hit for the dependents as runfiles (or `.cargo_runfiles` in the case of this PR) are not fetched and will not exist on the host at link time.

This change addresses this issue of untracked dependencies by ensuring the runfiles of `cargo_build_script.script` targets are aggregated into `Rustc` actions.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -36,6 +36,21 @@ default_windows_targets: &default_windows_targets
   - "//..."
   - "-//test/proto/..."
   - "-//test/unit/pipelined_compilation/..."
+default_windows_no_runfiles_targets: &default_windows_no_runfiles_targets
+  - "--" # Allows negative patterns; hack for https://github.com/bazelbuild/continuous-integration/pull/245
+  - "//..."
+  # TODO: https://github.com/bazelbuild/rules_rust/issues/1156
+  - "-//crate_universe/..."
+  - "-//test/chained_direct_deps:mod3_doc_test"
+  - "-//test/out_dir_in_tests:demo_lib_doc_test"
+  - "-//test/proto/..."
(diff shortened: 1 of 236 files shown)


UebelAndre:
@scentini @krasimirgg would love your thoughts on this as well if you have the time.

illicitonion (approved):

Issue #2884, opened 2024-09-18
Prost: Handling of reserved keywords in package names is broken


purkhusid:
It seems like handling of proto package names is broken in the prost wrapper in rules_rust.

I've created a repro here: https://github.com/bazelbuild/rules_rust/compare/main...purkhusid:rules_rust:prost-keyword-repro

If you run `bazel test //proto/prost/private/tests/keywords/...` you will get the following error:

```
hread 'main' panicked at external/rules_rust_prost__prost-build-0.13.1/src/config.rs:1048:49:
called `Result::unwrap()` on an `Err` value: Error("expected identifier, found keyword `type`")
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: prost_build::config::Config::generate
   4: <protoc_gen_prost::generator::core::CoreProstGenerator as protoc_gen_prost::generator::Generator>::generate
   5: <protoc_gen_prost::generator::ChainedGenerator<G1,G2> as protoc_gen_prost::generator::Generator>::generate
   6: protoc_gen_prost::execute
   7: protoc_gen_prost__bin::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
--prost_out: protoc-gen-prost: Plugin failed with status code 101.
thread 'main' panicked at proto/prost/private/protoc_wrapper.rs:751:9:
protoc failed with status: 1
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: protoc_wrapper::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace
```

I have tried to reproduce this error using prost without Bazel and also `protoc-gen-prost` without Bazel and so far I've not been able to. This points to the issue being in how rules_rust wraps `protoc-gen-prost`.

purkhusid:
This is fixed in https://github.com/bazelbuild/rules_rust/commit/62dbb7756c80c7d01c6d7bc3d642b07486ee4c68

PR #2888, opened 2024-09-19
Fix keyword handling in prost wrapper

purkhusid:
## What?
Fixes Prost: Handling of reserved keywords in package names is broken #2884

## How?
Added escaping on Rust keywords in proto package names


illicitonion (approved):
Thanks!

PR #2897, opened 2024-09-24
Added utility for parsing action Args param files

UebelAndre:
This change introduces the `action_args` crate which is something I feel I keep writing in various repos now. It's original design is to make it easier to pass args to built binaries. E.g.

```rust
use action_args;
use clap::Parser;
use runfiles::{rlocation, Runfiles};

#[command(version, about, long_about = None)]
struct ClapArgs {}

fn main() {
    let args = {
        let runfiles = Runfiles::create().unwrap();

        let var = std::env::var("ARGS_FILE").unwrap();
        let runfile = rlocation!(runfiles, var).unwrap();
        let text = std::fs::read_to_string(runfile).unwrap();
        let argv = action_args::parse_args(text);
        ClapArgs::parse_from(std::env::args().take(1).chain(argv))
    };

    // ...
    // ...
    // ...
}
```

This utility will likely be unnecessary should https://github.com/bazelbuild/bazel/issues/16076 ever be implemented.

### DIFF
--- a/tools/action_args/BUILD.bazel
+++ b/tools/action_args/BUILD.bazel
@@ -1,0 +1,13 @@
+load("//rust:defs.bzl", "rust_library", "rust_test")
+
+rust_library(
+    name = "action_args",
+    srcs = ["action_args.rs"],
+    edition = "2021",
+    visibility = ["//visibility:public"],
+)
+
+rust_test(
+    name = "action_args_test",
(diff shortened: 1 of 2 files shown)


UebelAndre:
I'm happy to reject this if it seems too trivial 😅

illicitonion (approved):
Sure!

PR #2900, opened 2024-09-24
Accept rust_library_group in proc_macro_deps

### LINKED ISSUE #2892: proc_macro_deps should accept rust_library_group
`proc_macro_deps` requires `CrateInfo` provider though `CrateGroupInfo` is also supported.

ashi009:
Fixes #2892

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -61,15 +61,16 @@ def _assert_correct_dep_mapping(ctx):
                     ),
                 )
     for dep in ctx.attr.proc_macro_deps:
-        type = dep[rust_common.crate_info].type
-        if type != "proc-macro":
-            fail(
-                "{} listed {} in its proc_macro_deps, but it is not proc-macro, it is a {}. It should probably instead be listed in deps.".format(
-                    ctx.label,
-                    dep.label,
-                    type,
-                ),
-            )
+        if rust_common.crate_info in dep:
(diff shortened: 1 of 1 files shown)


rust/private/rust.bzl
```diff
@@ -647,7 +648,7 @@ _common_attrs = {
             List of `rust_proc_macro` targets used to help build this library target.
         """),
         cfg = "exec",
-        providers = [rust_common.crate_info],
+        providers = [[rust_common.crate_group_info], [rust_common.crate_info]],
```
illicitonion, line 651: This accepts `crate_group_info` providers, but then doesn't look for them? I think you'd need a second branch in the `if` you added which looks for `crate_group_info` and if it's present, iterates across them to look in their `crate_info` values?

Please could you add a test which uses this functionality, so we can see the code definitely works?

illicitonion (changes requested):

rust/private/rust.bzl
ashi009: > This accepts crate_group_info providers, but then doesn't look for them? I think you'd need a second branch in the if you added which looks for crate_group_info and if it's present, iterates across them to look in their crate_info values?

This follows the established checking logic for deps, on top of this change. So I didn't add a second branch. Also, as `crate_group_info` returns a depset, do we want to flatten that just for checking?

PR #2902, opened 2024-09-26
Allowlist more clang flags in bindgen

hlopko:
Especially `--target` is very usable for cross-compilation use cases.


krasimirgg:
could you look into the failing CI? Looks related.

krasimirgg (changes requested):
(just marking as needs attn)

hlopko:
The CI is green now, thanks for the ping!

krasimirgg (approved):

PR #2905, opened 2024-09-30
Fix `override_targets` keys in docs

max-heller:
Updates the documentation to match the implementation: https://github.com/bazelbuild/rules_rust/blob/144d34fdaa3c3c5fe27a9666b760aa31aa9155a5/crate_universe/src/context/crate_context.rs#L65-L74


illicitonion (approved):
Thanks!

PR #2908, opened 2024-10-01
Allow `+` in repo names

fmeum:
Bazel 8 uses `+` as the separator for segments of canonical repo names, whereas Bazel 7 uses `~` or `+` depending on the value of `--incompatible_use_plus_in_repo_names`.


fmeum:
CC @Wyverald

fmeum:
@illicitonion Could we get a release this week? Bazel 8rc1 will come out on Monday and needs this.

PR #2911, opened 2024-10-01
Use absolute path with "-fsanitize-ignorelist"

vitalybuka:
Also generalize `_pwd_flags`. Flags can use both, " " and "=" as argument separators.


krasimirgg (approved):
Thank you!

PR #2914, opened 2024-10-03
Feature/wasm UI example (DRAFT)

kolloch:
I thought it might be nice to provide a WASM UI example for bzlmod with yew.


kolloch:
There are rolling release failures but on first glance they seem rather unrelated?

I could fix some by explicitly adding `--enable_workspace` to `.bazelrc`

PR #2922, opened 2024-10-06
Restore lost "is_absolute"

### LINKED ISSUE #2917: 0.52.0 breaks cross compilation of sys dependencies
The new [0.52.0](https://github.com/bazelbuild/rules_rust/releases/tag/0.52.0) release breaks my build completely.

The issue seems to be related to vendored sys-crates not compiling the underlying c deps anymore.
So far I was able to pinpoint the problem to the following crates:
* mimalloc
* lz4-sys

The only viable workaround is to pin rules_rust to any of the previous versions. 

I made a repro of the issue that builds with the previous 0.51 version and virtually all before, but when using the new 0.52 version, it throws the exact same error that I see in my main mono repo.

Link to the repro: 
https://github.com/marvin-hansen/sys-crates-repro/tree/main

vitalybuka:
`is_absolute` was lost for "=" case in #2911

Fixes  #2917

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -182,7 +182,7 @@ def _pwd_flags(args):
             fix_next_arg = False
         else:
             opt, path = split_flag(arg)
-            if opt and path:
+            if opt and path and not paths.is_absolute(path):
                 res.append("{}${{pwd}}/{}".format(opt, path))
             else:
                 fix_next_arg = (opt != None)


UebelAndre (changes requested):
Thanks! Don’t mean to burden you after dealing with the regression but Is this something that can be regression tested?

vitalybuka:
I am not sure how to do that in reasonable way. Looks like there is no a single test which fail if I "return []" from _pwd_flags.

I am very new to this code, could you please suggest which test could be the likely example for requested test?

UebelAndre (approved):

PR #2923, opened 2024-10-06
[Bugfix] protos with the same package name prefix fail to compile

houqp:
When proto files sharing the same package name prefixes reference types from each other, protoc_wrapper will generate code with incorrect module scope resulting in the following compilation error:

```
error[E0412]: cannot find type `CMessage` in module `super`
  --> bazel-out/darwin_arm64-fastbuild/bin/proto/prost/private/tests/shared_package_prefix/shared_pkg_prefix_proto.lib.rs:21:46
   |
21 |         pub c: ::core::option::Option<super::CMessage>,
   |                                              ^^^^^^^^ not found in `super`
   |
help: consider importing this struct
   |
15 +     use crate::pkg::CMessage;
   |
help: if you import `CMessage`, refer to it directly
   |
21 -         pub c: ::core::option::Option<super::CMessage>,
21 +         pub c: ::core::option::Option<CMessage>,
```

This Patch refactors the module tree walk implementation to ensure generated rust code are wrapped with the correct rust module hierarchy.

### DIFF
--- a/proto/prost/private/protoc_wrapper.rs
+++ b/proto/prost/private/protoc_wrapper.rs
@@ -68,7 +68,7 @@ struct Module {
     contents: String,
 
     /// The names of any other modules which are submodules of this module.
-    submodules: BTreeSet<String>,
+    submodules: BTreeMap<String, Module>,
 }
 
 /// Generate a lib.rs file with all prost/tonic outputs embeeded in modules which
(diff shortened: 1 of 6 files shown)


houqp:
@UebelAndre the CI error looks like infra related and not caused by my change?

PR #2924, opened 2024-10-07
Fix fix_next_arg condition

vitalybuka:
Follow up to #2922 and #2917.

Before #2922 the only possible "else" with "opt"
was space separated flags. Now we get there for
already `is_absolute` path.


vitalybuka:
Thanks for reminder.
It's not needed.

PR #2926, opened 2024-10-08
Updated Bazelmod cross compilation to catch MacOS sys dependency issues

marvin-hansen:
As discussed in issue https://github.com/bazelbuild/rules_rust/issues/2917,
this updated cross compilation example serves as a bridge until the exact cause of the sys crates not building on Mac issue has been found and better regression testing has been added. 

Also, I took the freedom to trim and streamline the LLVM setup making it easier to configure while preserving the property that it caches failing sys crate builds. Updated the Readme accordingly.


marvin-hansen:
Update:
* Apparently none of the Bazelmod examples run on MacOS on CI; at least that explains how things slipped through.
* I added one MacOS test runner for the cross compile example following the format of all the other tests. 
* Since the MacOS runner uses X86, I updated the LLVM config so that Bazel downloads the correct Intel/Mac version.

However, I am not familiar with BazelCI, but is there a way to add a single MacOS runner that basically runs all Bazelmod tests similar to the crate universe runner for the Workspace examples?

Would be great if someone with a solid understanding of the CI would ensure that all Bazelmod examples get build and tested on MacOS.

.bazelci/presubmit.yml
```diff
@@ -857,12 +865,6 @@ tasks:
     working_directory: examples/bzlmod/proto
     build_targets:
       - "//..."
-  bzlmod_proto:
-    name: Proto and Prost with prebuilt protoc with bzlmod
-    platform: ubuntu2004
-    working_directory: examples/bzlmod/proto_with_toolchain
-    build_targets:
-      - "//..."
```
krasimirgg, line 865: is this removal intentional?

examples/bzlmod/cross_compile/README.md
```diff
@@ -1,15 +1,18 @@
 # Cross Compilation
 
-For cross compilation, you have to specify a custom platform to let Bazel know that you are compiling for a different platform than the default host platform.
+For cross compilation, you have to specify a custom platform to let Bazel know that you are compiling for a different platform than the default host platform.
```
krasimirgg, line 3: nit: drop trailing space

examples/bzlmod/cross_compile/README.md
```diff
@@ -38,113 +41,95 @@ The setup requires three steps, first declare dependencies and toolchains in you
...
-# ARM (aarch64) LLVM Toolchain with sysroot.
-# https://github.com/bazelbuild/rules_rust/blob/main/examples/bzlmod/cross_compile/WORKSPACE.bzlmod
-llvm.toolchain(
-    name = "llvm_toolchain_aarch64_with_sysroot",
-    llvm_versions = LLVM_VERSIONS,
-)
 llvm.sysroot(
-    name = "llvm_toolchain_aarch64_with_sysroot",
-    label = "@org_chromium_sysroot_linux_aarch64//:sysroot",
+    name = "llvm_toolchain",
```
krasimirgg, line 117: I don't quite understand how the new version works: naively this has the same name as the previosly-defined llvm_toolchain, so it will create a name collision?

examples/bzlmod/cross_compile/.bazelrc
```diff
@@ -0,0 +1,85 @@
...
+build --incompatible_enable_cc_toolchain_resolution
+
+###############################################################################
+## Custom user flags
+##
+## This should always be the last thing in the `.bazelrc` file to ensure
+## consistent behavior when setting flags in that file as `.bazelrc` files are
+## evaluated top to bottom.
+###############################################################################
+try-import %workspace%/user.bazelrc
```
krasimirgg, line 85: For the .blazerc files in in test subdirectories, it looks like we try to keep them tiny with only the settings necessary for the specific example to work, e.g.,: https://github.com/bazelbuild/rules_rust/blob/main/examples/bzlmod/hello_world/.bazelrc

Could you remove the settings that are not necessary?

krasimirgg (changes requested):
Left some comments. Stuff looks largely reasonable, (note unfortunately I'm not really familiar with the bzlmod presubmits).

examples/bzlmod/cross_compile/README.md
marvin-hansen: No it doesn't. I got confused by this as well since the WORKSPACE config is quite different. Previously, I declared 3 different LLVM, one for host and 2 for the target platform, but that caused to download LLVM 3 times... 
I asked another engineer who is more senior and who pointed out, I actually can just declare one, add the sysroot to it, and then it works as good and that is exactly what I am doing here. I don't exactly understand the details either, but in my experience, it works just fine in 3 different projects.

examples/bzlmod/cross_compile/.bazelrc
marvin-hansen: sure, let me clean this up

.bazelci/presubmit.yml
marvin-hansen: Nope, that was an accident. Probably a cut instead of copy. 
 Good catch. Thanks for point it out.

marvin-hansen:
@krasimirgg I addresses all of your issues with the latest commits.
CI is green.

Please review again.
Thank you

krasimirgg (approved):
Thank you! 

In general having a macos CI sounds reasonable.

PR #2927, opened 2024-10-08
Add target specific rustc_flags last to the command line

erenon:
This allows target specific flags to override toolchain
specific flags, e.g: to disable LTO on examples.


krasimirgg (approved):
Thank you!

PR #2928, opened 2024-10-08
Add proc_macro_deps for rust_doc_test

erenon:
This allows representing doctests that
depend on proc macros their crate do not depend on.


havasd:
@UebelAndre sorry to tag you but if you have some time, could you please review this MR?

UebelAndre (approved):
Thanks for the ping, looks good to me!

PR #2934, opened 2024-10-14
Update wasm_bindgen to non-deprecated rules_js API

dzbarsky:
`js_info` needs to jump through some hoops to support the legacy names. The new ones are shorter and clearer.
Also, set the toolchain on the action to support Auto Exec Groups while I'm here.


UebelAndre (approved):
Thanks! Does this break any backward compatibility?

dzbarsky:
> Thanks! Does this break any backward compatibility?

This means you have to use rules_js 2.x instead of 1.x with rules_rust. We can add it to release notes but I think it's unlikely anyone cares, I suspect I might be the only one using this given I keep finding it broken :)

DavidZbarsky-at:
@UebelAndre I think CI is "green", I retried twice and got different flakes every time

PR #2925, opened 2024-10-07
Revert #2911 and #2922

UebelAndre:
Revert "Restore lost "is_absolute" (#2922)"

This reverts commit 2a85e13c74239d2a0d0c47a37f10a4c9739fa5e6.

Revert "Use absolute path with "-fsanitize-ignorelist" (#2911)"

This reverts commit 9594fa70b5b6f556a7b4f57ca0ac7c48149bf280.

PR #2935, opened 2024-10-14
cargo_build_script: Add support for `-fsanitize-ignorelist=`

vitalybuka:
This is trivial copy/paste implementation, to
minimize regression probability.

Previously it was combined with refactoring, but
introduced regression and was reverted #2911.

This time I split refactoring into follow up PR.

PR #2936, opened 2024-10-14
cargo_build_script: Add regression test for revert #2925

vitalybuka:
#2931 added tests for _pwd_flags, however 
the test as-is passes with #2911 and #2922,
which still caused problems and reverted.
Here I introduce a test case for likely
unfixed in reverted #2925.

The test needs strictter check for cflags. As-ls
`flag in cflags` is a just substring search and
very weak test for abs path.

For #2917

### DIFF
--- a/test/cargo_build_script/cc_args_and_env/BUILD.bazel
+++ b/test/cargo_build_script/cc_args_and_env/BUILD.bazel
@@ -5,12 +5,15 @@ load(
     "isystem_absolute_test",
     "isystem_relative_test",
     "sysroot_absolute_test",
+    "sysroot_absolute2_test",
     "sysroot_relative_test",
 )
 
 sysroot_relative_test(name = "sysroot_relative_test")
 
 sysroot_absolute_test(name = "sysroot_absolute_test")
+
+sysroot_absolute2_test(name = "sysroot_absolute2_test")
 
 isystem_relative_test(name = "isystem_relative_test")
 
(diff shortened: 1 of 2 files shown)


vitalybuka:
@krasimirgg

vitalybuka:
Please review @krasimirgg @UebelAndre 
But to avoid conflicts with #2935 marking as draft.
I would appreciate if land #2935 and I update this one,
and then upload refactoring to avoid code duplication,
which contained the cause of the revert.

krasimirgg:
Could you try to re-format the BUILD.bazel file per the Buildifier message:
```
##### :bazel: buildifier: found 1 format issue in your WORKSPACE, BUILD and *.bzl files
--
  | 2024-10-16 12:43:50 CEST | If this repo uses a pre-commit hook, then you should install it. Otherwise, please download <a href="https://github.com/bazelbuild/buildtools/releases/tag/v7.3.1">buildifier 7.3.1</a> and run the following command in your workspace:<br/><pre><code>buildifier test/cargo_build_script/cc_args_and_env/BUILD.bazel</code></pre>
```

krasimirgg (approved):

vitalybuka:
This is rebased version ready to merge.

vitalybuka:
From merge queue error:
`fatal: couldn't find remote ref gh-readonly-queue/main/pr-2936-63fa643067d5b006900b880695a6cc4187b325c3`

I don't know what to do about that, maybe rebased one will work.

PR #2941, opened 2024-10-16
gen_rust_project: optimize aquery with many targets

TheLortex:
Hi, if I'm not mistaken, in Bazel `deps(a)+deps(b)` is equivalent to `deps(a+b)`. The latter is more efficient because it doesn't have to deduplicate common dependencies between a and b. This performance issue has been triggered in our configuration, as we track roughly 700 targets and 1600 transitive dependencies in our monorepo. 
Running `aquery` with the original implementation takes 15 seconds on my machine while the implementation proposed in this PR takes 1.5 seconds.


UebelAndre (approved):
Great find! Thank you!

PR #2943, opened 2024-10-17
[Prost] Remove extra generated modules and opt-in to transitively including deps

freeformstu:
This PR makes two changes:
1. It removes the duplicately generated code in the prost `lib.rs`.
2. It allows opting into transitively including all of the proto deps.

## Duplicately generated code.

Previously there was a bug where we generated the same code at every proto package level. This created separate Rust types for each module which made it easy to accidentally import an incompatible type.

Before:
```rust
// @generated

pub mod b_ar {
    pub mod b_az {
        pub mod qaz {
            pub mod qu_x {
                // @generated
                // This file is @generated by prost-build.
                #[allow(clippy::derive_partial_eq_without_eq)]
                #[derive(Clone, PartialEq, ::prost::Message)]
                pub struct Bar {
                    #[prost(string, tag = "1")]
                    pub name: ::prost::alloc::string::String,
                    #[prost(message, optional, tag = "2")]
                    pub foo: ::core::option::Option<
                        ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::Foo,
                    >,
                    #[prost(message, optional, tag = "3")]
                    pub nested_foo: ::core::option::Option<
                        ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::foo::NestedFoo,
                    >,
                }
                /// Nested message and enum types in `Bar`.
                pub mod bar {
                    #[allow(clippy::derive_partial_eq_without_eq)]
                    #[derive(Clone, PartialEq, ::prost::Message)]
                    pub struct Baz {
                        #[prost(string, tag = "4")]
                        pub name: ::prost::alloc::string::String,
                    }
                }
                // @@protoc_insertion_point(module)
            }
        }
    }
}
pub mod b_az {
    pub mod qaz {
        pub mod qu_x {
            // @generated
            // This file is @generated by prost-build.
            #[allow(clippy::derive_partial_eq_without_eq)]
            #[derive(Clone, PartialEq, ::prost::Message)]
            pub struct Bar {
                #[prost(string, tag = "1")]
                pub name: ::prost::alloc::string::String,
                #[prost(message, optional, tag = "2")]
                pub foo:
                    ::core::option::Option<::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::Foo>,
                #[prost(message, optional, tag = "3")]
                pub nested_foo: ::core::option::Option<
                    ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::foo::NestedFoo,
                >,
            }
            /// Nested message and enum types in `Bar`.
            pub mod bar {
                #[allow(clippy::derive_partial_eq_without_eq)]
                #[derive(Clone, PartialEq, ::prost::Message)]
                pub struct Baz {
                    #[prost(string, tag = "4")]
                    pub name: ::prost::alloc::string::String,
                }
            }
            // @@protoc_insertion_point(module)
        }
    }
}
pub mod qaz {
    pub mod qu_x {
        // @generated
        // This file is @generated by prost-build.
        #[allow(clippy::derive_partial_eq_without_eq)]
        #[derive(Clone, PartialEq, ::prost::Message)]
        pub struct Bar {
            #[prost(string, tag = "1")]
            pub name: ::prost::alloc::string::String,
            #[prost(message, optional, tag = "2")]
            pub foo: ::core::option::Option<::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::Foo>,
            #[prost(message, optional, tag = "3")]
            pub nested_foo: ::core::option::Option<
                ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::foo::NestedFoo,
            >,
        }
        /// Nested message and enum types in `Bar`.
        pub mod bar {
            #[allow(clippy::derive_partial_eq_without_eq)]
            #[derive(Clone, PartialEq, ::prost::Message)]
            pub struct Baz {
                #[prost(string, tag = "4")]
                pub name: ::prost::alloc::string::String,
            }
        }
        // @@protoc_insertion_point(module)
    }
}
pub mod qu_x {
    // @generated
    // This file is @generated by prost-build.
    #[allow(clippy::derive_partial_eq_without_eq)]
    #[derive(Clone, PartialEq, ::prost::Message)]
    pub struct Bar {
        #[prost(string, tag = "1")]
        pub name: ::prost::alloc::string::String,
        #[prost(message, optional, tag = "2")]
        pub foo: ::core::option::Option<::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::Foo>,
        #[prost(message, optional, tag = "3")]
        pub nested_foo: ::core::option::Option<
            ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::foo::NestedFoo,
        >,
    }
    /// Nested message and enum types in `Bar`.
    pub mod bar {
        #[allow(clippy::derive_partial_eq_without_eq)]
        #[derive(Clone, PartialEq, ::prost::Message)]
        pub struct Baz {
            #[prost(string, tag = "4")]
            pub name: ::prost::alloc::string::String,
        }
    }
    // @@protoc_insertion_point(module)
}

```

Now:
```rust
// @generated

pub use foo_proto;

pub mod b_ar {
    pub mod b_az {
        pub mod qaz {
            pub mod qu_x {
                // @generated
                // This file is @generated by prost-build.
                #[allow(clippy::derive_partial_eq_without_eq)]
                #[derive(Clone, PartialEq, ::prost::Message)]
                pub struct Bar {
                    #[prost(string, tag = "1")]
                    pub name: ::prost::alloc::string::String,
                    #[prost(message, optional, tag = "2")]
                    pub foo: ::core::option::Option<
                        ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::Foo,
                    >,
                    #[prost(message, optional, tag = "3")]
                    pub nested_foo: ::core::option::Option<
                        ::foo_proto::foo::quu_x::co_rg_e::grault::ga_rply::foo::NestedFoo,
                    >,
                }
                /// Nested message and enum types in `Bar`.
                pub mod bar {
                    #[allow(clippy::derive_partial_eq_without_eq)]
                    #[derive(Clone, PartialEq, ::prost::Message)]
                    pub struct Baz {
                        #[prost(string, tag = "4")]
                        pub name: ::prost::alloc::string::String,
                    }
                }
                // @@protoc_insertion_point(module)
            }
        }
    }
}
```

## Including transitive depenencies.

Because the proto rules rely on aspects, we are compiling targets which may not be directly accessible as a label in Bazel. Out of convenience, the original implementation automatically included all transitive dependencies to the `rust_library` or `rust_binary` that depended on the proto crate. 

As an example, let's say we have the following hierarchy:
```
my_crate -> a.proto
a.proto -> b.proto
b.proto -> c.proto
```
The previous setup made it possible for `my_crate` to directly import `a_proto`, `b_proto`, and `c_proto`.

This is problematic though because it leads to significant crate dependency bloat. `my_crate` may only use `a_proto` directly but it now also depends on the unused crates `b_proto` and `c_proto`.

This PR makes every generated crate re-export its direct dependencies. So, for the same example we now can import via those reexports:
```rust
//! my_crate
use a_proto::A;
use a_proto::b_proto::B;
use a_proto::b_proto::c_proto::C;
```

This means that you can also add the rustc flag `-Dunused_crate_dependencies` to ensure you're not depending on crates which are unused. In our monorepo we've had to disable this flag for crates depending on protos.

If you need to depend on an intermediate crate directly, you can just add a `rust_prost_library` definition and since it's all generated with aspects, it will be the identical underlying crate.

Finally, if you really want the extra crates included, you can set the flag `include_transitive_deps` on the prost toolchain and that will bring back the previous behavior.

### DIFF
--- a/proto/prost/private/tests/transitive_dependencies/a_test.rs
+++ b/proto/prost/private/tests/transitive_dependencies/a_test.rs
@@ -1,14 +1,13 @@
 //! Tests transitive dependencies.
 
 use a_proto::a::A;
-use a_proto::b_proto::a::b::B;
 use a_proto::b_proto::c_proto::a::b::c::C;
 use a_proto::duration_proto::google::protobuf::Duration;
 use a_proto::timestamp_proto::google::protobuf::Timestamp;
 use a_proto::types_proto::Types;
 
 #[test]
-fn test_b() {
+fn test_a() {
     let duration = Duration {
         seconds: 1,
         nanos: 2,
@@ -16,7 +15,8 @@ fn test_b() {
 
     let a = A {
         name: "a".to_string(),
-        b: Some(B {
+        // Ensure the external `b_proto` dependency is compatible with `a_proto`'s `B`.
+        b: Some(b_proto::a::b::B {
             name: "b".to_string(),
             c: Some(C {
                 name: "c".to_string(),
--- a/proto/prost/private/tests/transitive_dependencies/a_test.rs
+++ b/proto/prost/private/tests/transitive_dependencies/a_test.rs
@@ -1,11 +1,11 @@
 //! Tests transitive dependencies.
 
 use a_proto::a::A;
-use b_proto::a::b::B;
-use c_proto::a::b::c::C;
-use duration_proto::google::protobuf::Duration;
-use timestamp_proto::google::protobuf::Timestamp;
-use types_proto::Types;
+use a_proto::b_proto::a::b::B;
+use a_proto::b_proto::c_proto::a::b::c::C;
+use a_proto::duration_proto::google::protobuf::Duration;
+use a_proto::timestamp_proto::google::protobuf::Timestamp;
+use a_proto::types_proto::Types;
 
 #[test]
 fn test_b() {
--- a/proto/prost/private/tests/transitive_dependencies/transition.bzl
+++ b/proto/prost/private/tests/transitive_dependencies/transition.bzl
@@ -2,16 +2,16 @@
 
 load("//rust:defs.bzl", "rust_common")
 
-def _extra_toolchain_transition_impl(_settings, attr):
+def _extra_toolchain_transition_impl(settings, attr):
     return {
         "//command_line_option:extra_toolchains": [
             attr.toolchain,
-        ],
+        ] + settings["//command_line_option:extra_toolchains"],
     }
 
 _extra_toolchain_transition = transition(
     implementation = _extra_toolchain_transition_impl,
-    inputs = [],
+    inputs = ["//command_line_option:extra_toolchains"],
     outputs = ["//command_line_option:extra_toolchains"],
 )
 
@@ -23,6 +23,7 @@ def _extra_toolchain_wrapper_impl(ctx):
 
 extra_toolchain_wrapper = rule(
     implementation = _extra_toolchain_wrapper_impl,
+    doc = "Transition rule used for unit testing the prost toolchains.",
     attrs = {
         "dep": attr.label(),
         "toolchain": attr.label(),
--- a/proto/prost/private/tests/transitive_dependencies/transition.bzl
+++ b/proto/prost/private/tests/transitive_dependencies/transition.bzl
@@ -1,0 +1,32 @@
+load("//rust:defs.bzl", "rust_common")
+
+def _extra_toolchain_transition_impl(_settings, attr):
+    return {
+        "//command_line_option:extra_toolchains": [
+            attr.toolchain,
+        ],
+    }
+
+_extra_toolchain_transition = transition(
+    implementation = _extra_toolchain_transition_impl,
+    inputs = [],
+    outputs = ["//command_line_option:extra_toolchains"],
+)
+
+def _extra_toolchain_wrapper_impl(ctx):
+    return [
+        ctx.attr.dep[DefaultInfo],
+        ctx.attr.dep[rust_common.crate_group_info],
+    ]
+
+extra_toolchain_wrapper = rule(
+    implementation = _extra_toolchain_wrapper_impl,
+    attrs = {
+        "dep": attr.label(),
+        "toolchain": attr.label(),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+    },
+    cfg = _extra_toolchain_transition,
+)
--- a/proto/prost/private/tests/transitive_dependencies/transition.bzl
+++ b/proto/prost/private/tests/transitive_dependencies/transition.bzl
@@ -1,3 +1,5 @@
+"""Transition rule for the transitive dependencies test case."""
+
 load("//rust:defs.bzl", "rust_common")
 
 def _extra_toolchain_transition_impl(_settings, attr):
--- a/docs/src/flatten.md
+++ b/docs/src/flatten.md
@@ -795,8 +795,9 @@ Builds a Rust proc-macro crate.
 ## rust_prost_toolchain
 
 <pre>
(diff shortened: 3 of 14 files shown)


proto/prost/private/tests/transitive_dependencies/transition.bzl
```diff
@@ -0,0 +1,32 @@
...
+def _extra_toolchain_transition_impl(_settings, attr):
+    return {
+        "//command_line_option:extra_toolchains": [
+            attr.toolchain,
+        ],
+    }
+
+_extra_toolchain_transition = transition(
+    implementation = _extra_toolchain_transition_impl,
+    inputs = [],
```
UebelAndre, line 14: It seems like RBE needs previously defined extra toolchains? Might want to append your new toolchain here.

proto/prost/private/tests/transitive_dependencies/transition.bzl
```diff
@@ -0,0 +1,32 @@
...
+    outputs = ["//command_line_option:extra_toolchains"],
+)
+
+def _extra_toolchain_wrapper_impl(ctx):
+    return [
+        ctx.attr.dep[DefaultInfo],
+        ctx.attr.dep[rust_common.crate_group_info],
+    ]
+
+extra_toolchain_wrapper = rule(
```
UebelAndre, line 24: nit: can you fill out docs?

UebelAndre (changes requested):
Thanks! Just a few questions!

proto/prost/private/tests/transitive_dependencies/a_test.rs
```diff
@@ -1,11 +1,11 @@
 //! Tests transitive dependencies.
 
 use a_proto::a::A;
-use b_proto::a::b::B;
-use c_proto::a::b::c::C;
-use duration_proto::google::protobuf::Duration;
-use timestamp_proto::google::protobuf::Timestamp;
-use types_proto::Types;
+use a_proto::b_proto::a::b::B;
```
UebelAndre, line 4: Can you also add a test that shows transitive dependencies become directly consumable as they did before if a `rust_prost_library` is made for it and added as a direct dependency to the the calling crate?

UebelAndre (changes requested):

proto/prost/private/tests/transitive_dependencies/a_test.rs
freeformstu: I did that in the `transitve_deps_test.rs` which is beside this test.

proto/prost/private/tests/transitive_dependencies/transition.bzl
freeformstu: Thanks!

proto/prost/private/tests/transitive_dependencies/transition.bzl
freeformstu: Done

proto/prost/private/tests/transitive_dependencies/a_test.rs
freeformstu: I also updated the original test to take in b_proto as a direct dependency.

UebelAndre: Thank you! That's exactly what I was thinking

UebelAndre (approved):

bistack:
@freeformstu "a_proto::b_proto::c_proto::C" is too long when each x_proto is long.

As x_proto is a label name, it is more likely that there are multiple targets with same label name in a bigger monorepo.

So, we patched prost.bzl with a crate_name = label.package + label.name.

PR #2964, opened 2024-10-27
Remove dependency on spectral

dzbarsky:
It's only used for 2 test cases and we can just not use it.


dzbarsky:
@illicitonion want to take a look?

PR #2965, opened 2024-10-28
rust_test(crate=foo) inherits foo.aliases

sam-mccall:
Inline tests compile the same set of source files as the original crate.
The set of aliases required will be the same or occasionally a superset.
We shouldn't make users repeat them across targets.


krasimirgg:
could you add a regression test

krasimirgg (approved):
Thank you!

PR #2966, opened 2024-10-28
A rust_test_suite(srcs=[]) should be empty.

sam-mccall:
This can arise when srcs is a glob that currently matches no files.
The target may nevertheless be useful for regularity across libraries,
and to future-proof against newly-added test files being ignored.

Previously, we hit the special-case handling of `test_suite(tests=[])`:

> If the tests attribute is unspecified or empty, the rule will default
> to including all test rules in the current BUILD file that are not
> tagged as manual. These rules are still subject to tag filtering.

We can't directly turn this behavior off, so this patch uses tag
filtering to avoid including those default tests.


test/empty_suite/BUILD.bazel
```diff
@@ -0,0 +1,47 @@
...
+    scope = [":suite"],
+)
+
+# Test that the suite is empty.
+sh_test(
+    name = "deps_test",
+    srcs = ["verify_empty.sh"],
+    args = ["$(location :deps)"],
+    data = [":deps"],
+)
```
buchgr, line 47: I think the best way forward is to use bazel's runfile library. Basically follow these instructions https://github.com/bazelbuild/bazel/blob/c71f0b6ded1de926bd1081fd4327142777711b03/tools/bash/runfiles/runfiles.bash#L55: 1) Add `"@bazel_tools//tools/bash/runfiles"` to deps, copy'n'paste the stanza that loads the runfiles library to `verify_empty.sh` and the use the `rlocation` function to get the on disk path.

test/empty_suite/BUILD.bazel
```diff
@@ -0,0 +1,47 @@
...
+    name = "deps",
+    expression = "deps(//test/empty_suite:suite, 1) - //test/empty_suite:suite",
+    scope = [":suite"],
+)
+
+# Test that the suite is empty.
+sh_test(
+    name = "deps_test",
+    srcs = ["verify_empty.sh"],
+    args = ["$(location :deps)"],
```
buchgr, line 45: You may have to use `rlocationpath` instead of `location`, but not sure without trying it out.

test/empty_suite/verify_empty.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Verifies that argv[1] is an empty file.
+
+set -eux
+
+test -f $1
```
buchgr, line 6: If you apply my above suggestion this would be rewritten to `test -f $(rlocation $1)`.

Essentially, when `--noenable_runfiles` Bazel writes a file `MANIFEST` contains a mapping of logical path to on disk path for every file in the `data` attribute. the `rlocation` function parses the file given the logical path and returns the absolute path. If runfiles are enabled you don't need this because then bazel will create a symlink for logical path -> on disk path.

sam-mccall:
Thanks @buchgr!

That will work, but of course adds some complexity (new dependencies, extra code in the script, more moving pieces to debug). One alternative is to disable this test in the no-runfiles configuration (https://github.com/bazelbuild/rules_rust/blob/main/.bazelci/presubmit.yml#L39). This is clearly a limitation of the test rather than the rules, and we have plenty of coverage in other configurations.

@krasimirgg what would you prefer here?

krasimirgg:
> Thanks @buchgr!
> 
> That will work, but of course adds some complexity (new dependencies, extra code in the script, more moving pieces to debug). One alternative is to disable this test in the no-runfiles configuration (https://github.com/bazelbuild/rules_rust/blob/main/.bazelci/presubmit.yml#L39). This problem is clearly a limitation of the test rather than the rules, and we have plenty of coverage in other configurations.
> 
> @krasimirgg what would you prefer here?

I think disabling the regression test in the no-runfiles configuration is OK for this. Just wanna also run this trough @UebelAndre: OK with this?

krasimirgg:
@sam-mccall let's disable the test in the no-runfiles configuration with a comment that's intentional.

sam-mccall:
> @sam-mccall let's disable the test in the no-runfiles configuration with a comment that's intentional.

Done. The rolling-bazel buildkite failures seem unrelated.

krasimirgg (approved):
Thank you!

PR #2967, opened 2024-10-28
Add support for wasm32-wasip1 target.

PiotrSikora:
The wasm32-wasi target has been renamed to wasm32-wasip1, see:
https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html

Note that this change breaks @rules_rust//rust/platform:wasi builds when
using Rust version older than 1.78.


UebelAndre (approved):
Thanks!

PR #2968, opened 2024-10-28
Add build_script_env_for_target_cfg

illicitonion:
This allows setting env vars for a build script with selects based on the configuration of the target which depends on it, rather than exec configuration the build script naturally ends up configured in.


cargo/private/cargo_build_script_wrapper.bzl
```diff
@@ -107,6 +108,7 @@ def cargo_build_script(
             links attribute and therefore provide environment variables to this build script.
         proc_macro_deps (list of label, optional): List of rust_proc_macro targets used to build the script.
         build_script_env (dict, optional): Environment variables for build scripts.
```
UebelAndre, line 110: Wait, doesn't this get consumed as `cfg = "target"`? The order is:
```
rust_binary --exec--> cargo_build_script_runfiles --target--> cargo_build_script
```
If `env` if `build_script_env_for_target_cfg` is consumed through `cargo_build_script_runfiles` or directly on `cargo_build_script`, isn't the result the same?

UebelAndre (changes requested):

cargo/private/cargo_build_script_wrapper.bzl
illicitonion: `cargo_build_script_runfiles` doesn't get `exec` transitioned - you put `cargo_build_script_wrapper` macros directly in `deps` (no edge transition), and that macro expands to a `cargo_build_script` which doesn't have a self transition. A `cargo_build_script` depends on a `cargo_build_script_runfiles` using the `script` attribute (which has `cfg = "target"`).

UebelAndre: But why is consuming `env` from `cargo_build_script_runfiles` different than consuming it directly in `cargo_build_script`?

PR #2970, opened 2024-10-29
feat: add an attr to rust_binary for customization of the binary name

mattem:
It's often desirable to set the name of the resulting binary to something else other than the target name (`attr.name`), perhaps when the `rust_binary` target is wrapped in a macro, and not considered the "main" target of that macro.

This diff adds a new `binary_name` attribute to `rust_binary`, which when set will be used as the resulting binaries prefix instead of the targets name. The default is to still use the target name.


illicitonion (approved):
Beautiful, thanks!

PR #2983, opened 2024-11-05
Stop using config.string(allow_multiple = True)

EdSchouten:
According to the Bazel documentation, this option is deprecated. config.string_list(repeatable = True) should be used instead.

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -43,7 +43,7 @@ clippy_flag = rule(
         "Multiple uses are accumulated and appended after the extra_rustc_flags."
     ),
     implementation = _clippy_flag_impl,
-    build_setting = config.string(flag = True, allow_multiple = True),
+    build_setting = config.string_list(flag = True, repeatable = True),
 )
 
 def _clippy_flags_impl(ctx):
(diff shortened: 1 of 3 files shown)


UebelAndre (changes requested):
Thanks! Does this change usability of the flag? Can you also link some notes on the deprecation?

EdSchouten:
> Thanks! Does this change usability of the flag?

No, it doesn't. I just tested this by running the following command in the examples directory:

```
bazel build --verbose_failures --@rules_rust//rust/settings:extra_rustc_flag=foo --@rules_rust//rust/settings:extra_rustc_flag=bar //hello_world
```

In both cases the resulting command line flags passed to the Rust compiler are the same.

> Can you also link some notes on the deprecation?

https://bazel.build/rules/lib/toplevel/config#string

> Deprecated, use a string_list setting with repeatable = True instead. [...]

Deprecated in this commit, which is part of Bazel 6.0.0 and later: https://github.com/bazelbuild/bazel/commit/cdf01a39ab9def4d46f41595ac1ac9206a96d6f8

UebelAndre (approved):
Thanks!

PR #2986, opened 2024-11-06
Added `cargo_config` attribute to `cargo_bootstrap_repository`.

UebelAndre:
This can be useful for workspaces which have a `config.toml` for the workspace (used by `crate_universe`) and want to share the same configuration with any bootstrap repositories.


cargo/private/cargo_bootstrap.bzl
```diff
@@ -237,13 +250,17 @@ cargo_bootstrap_repository = repository_rule(
             ],
             default = "release",
         ),
+        "cargo_config": attr.label(
+            doc = "The path of the Cargo configuration (`Config.toml`) file.",
+            allow_single_file = ["Cargo.toml"],
```
illicitonion, line 255: I don't think these files are typically called Cargo.toml?

illicitonion (approved):
Makes sense, but one question

cargo/private/cargo_bootstrap.bzl
UebelAndre: Oh no!

UebelAndre: Fixed!

PR #2992, opened 2024-11-08
feat: output generated proto sources and descriptor set

mattem:
Adds the generated sources to an output group of the `rust_prost_library` so that they can be fetched by another rule, perhaps for using with `write_source_files` for checkin in generated files.


illicitonion (approved):
Seems reasonable, thanks!

I'm happy to merge if you un-draft

PR #2995, opened 2024-11-12
Correct prost file descriptor set option

GregBowyer:
I am pretty sure the argument given to protoc_gen_prost is wrong, file descriptors are not present in artifacts.

### DIFF
--- a/proto/prost/private/prost.bzl
+++ b/proto/prost/private/prost.bzl
@@ -64,7 +64,7 @@ def _compile_proto(ctx, crate_name, proto_info, deps, prost_toolchain, rustfmt_t
     additional_args.add("--deps_info={}".format(deps_info_file.path))
     additional_args.add("--direct_dep_crate_names={}".format(",".join(direct_crate_names)))
     additional_args.add("--prost_opt=compile_well_known_types")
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Interesting this didn't fail before. Is there some kind of regression test you could add with this change?

GregBowyer:
Turns out I misunderstood what programs are being invoked, this can be added to the toolchain if needed.

```python
  rust_prost_toolchain(
      name = "prost_toolchain_impl",
      prost_plugin = "@crate_index//:protoc-gen-prost__protoc-gen-prost",
      prost_runtime = ":prost_runtime",
      prost_types = "@crate_index//:prost-types",
┃     prost_opts = [
┃         # For making sure File descriptor sets are added
┃         "file_descriptor_set"
┃     ]
      proto_compiler = "@protobuf//:protoc",
      tonic_plugin = "@crate_index//:protoc-gen-tonic__protoc-gen-tonic",
      tonic_runtime = ":tonic_runtime",
  )

  toolchain(
      name = "prost_toolchain",
      toolchain = "prost_toolchain_impl",
      toolchain_type = "@rules_rust//proto/prost:toolchain_type",
  )
```

I think this is an incorrect PR; I dont know if this should be the default for rules_rust or not.

PR #3005, opened 2024-11-17
provide flag to exclude data at compile time

### LINKED ISSUE #2977: make `data` runtime-only to align with ecosystem norms
In the current rules_rust rules, the `data` attribute is available both at compile time and at runtime whereas the `compile_data` is just available at runtime. There is no attribute that describes runtime-only dependencies. This is unfortunate; changing runtime dependencies shouldn't force a recompilation and relinking of an executable.

In order to align with the rest of the bazel ecosystem, we should change `data` to not be available at compile time. This will be a breaking change, but it's worth it.

ajwerner:
Fixes #2977

### DIFF
--- a/rust/settings/BUILD.bazel
+++ b/rust/settings/BUILD.bazel
@@ -253,3 +253,11 @@ per_crate_rustc_flag(
     build_setting_default = [],
     visibility = ["//visibility:public"],
 )
+
+# A flag to control whether to include data files in compile_data.
+incompatible_flag(
+    name = "incompatible_do_not_include_data_in_compile_data",
+    build_setting_default = False,
+    issue = "https://github.com/bazelbuild/rules_rust/issues/2977",
+    visibility = ["//visibility:public"],
+)
--- a/rust/settings/BUILD.bazel
+++ b/rust/settings/BUILD.bazel
@@ -257,7 +257,7 @@ per_crate_rustc_flag(
 # A flag to control whether to include data files in compile_data.
 incompatible_flag(
     name = "incompatible_do_not_include_data_in_compile_data",
-    build_setting_default = False,
+    build_setting_default = True,
     issue = "https://github.com/bazelbuild/rules_rust/issues/2977",
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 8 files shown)


rust/settings/BUILD.bazel
```diff
@@ -253,3 +253,11 @@ per_crate_rustc_flag(
     build_setting_default = [],
     visibility = ["//visibility:public"],
 )
+
+# A flag to control whether to include data files in compile_data.
+incompatible_flag(
+    name = "incompatible_do_not_include_data_in_compile_data",
+    build_setting_default = False,
```
illicitonion, line 260: Let's default this to true - people can set it to False if they need time to migrate.

```suggestion
    build_setting_default = True,
```

illicitonion (approved):
LGTM, thanks! Let's default this to `True` to encourage migration :)

ajwerner:
@illicitonion Thanks for the review. I'm happy with this being default enabled. I fixed the failing tests. Looks like the merge failed due to flakiness downloading bazel:
```
2024/11/23 18:32:40 could not download Bazel: failed to download bazel: failed to download bazel: could not copy from https://releases.bazel.build/7.3.2/release/bazel-7.3.2-linux-x86_64 to /var/lib/buildkite-agent/.cache/bazelisk/downloads/_tmp/download1571092152: stream error: stream ID 1; INTERNAL_ERROR; received from peer
--
  | File "bazelci.py", line 4495, in <module>
  | sys.exit(main())
  | File "bazelci.py", line 4463, in main
  | execute_commands(
  | File "bazelci.py", line 1211, in execute_commands
  | PrepareRepoInCwd(True, initial_setup=True)
  | File "bazelci.py", line 1199, in PrepareRepoInCwd
  | os.environ["OUTPUT_BASE"] = get_output_base(bazel_binary)
  | File "bazelci.py", line 1982, in get_output_base
  | return execute_command_and_get_output(
  | File "bazelci.py", line 2618, in execute_command_and_get_output
  | process = subprocess.run(
  | File "/usr/lib/python3.8/subprocess.py", line 516, in run
  | raise CalledProcessError(retcode, process.args,
  | subprocess.CalledProcessError: Command '['bazel', 'info', 'output_base']' returned non-zero exit status 1.
  ```

PR #3006, opened 2024-11-17
fix!: correctly determine crate dependencies and features common to all supported targets

max-heller:
Currently, "common" deps and features in selects in generated lockfiles are not always common to all target platforms in the presence of conditional dependencies. This change ensures that common deps and features for a crate are common to all supported targets, not just those targets for which the crate dependency is enabled.

Unfortunately this results in a size explosion for some generated lockfiles. I believe this is mostly because the examples consider all target triples, and certain triples (in particular, `wasm32-unknown-unknown` for `examples/crate_universe/multi_package/cargo-bazel-lock.json`) break the commonality. When `supported_platform_triples` is specified, the impact should be less severe.

### DIFF
--- a/crate_universe/src/metadata/cargo_tree_resolver.rs
+++ b/crate_universe/src/metadata/cargo_tree_resolver.rs
@@ -65,6 +65,7 @@ pub(crate) struct CargoTreeEntry {
 }
 
 impl CargoTreeEntry {
+    #[cfg(test)]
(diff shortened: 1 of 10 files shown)


max-heller:
I could've sworn this was needed for a conditional compilation setup but things seem to be working without it so maybe the current implementation is sufficient after all. Closing unless the need arises again

Issue #2882, opened 2024-09-18
Shard rules_rust into multiple bzlmod modules


UebelAndre:
It seems to me that the `MODULE.bazel` file is getting a bit unwieldy and rust is having to declare dependencies for things not all users want or need.
https://github.com/bazelbuild/rules_rust/blob/350d249ca2c976f92c96d63e4f89f9cf00965188/MODULE.bazel#L37-L45
https://github.com/bazelbuild/rules_rust/blob/350d249ca2c976f92c96d63e4f89f9cf00965188/MODULE.bazel#L63-L160
https://github.com/bazelbuild/rules_rust/blob/350d249ca2c976f92c96d63e4f89f9cf00965188/MODULE.bazel#L170-L180

To improve maintainability and to keep the dependency trees tight, I'm wondering if it'd be better to have different bzlmod extensions for the various packages within `rules_rust`. Concretely, I'm proposing that `@rules_rust//bindgen` become `@rules_rust_bindgen` and be an independent module developers can choose to include but it's dependencies would not be a part of `@rules_rust`.

The scope of the proposed changed:

| before load | after load |
| --- | --- |
| `@rules_rust//bindgen` | `@rules_rust_bindgen` |
| `@rules_rust//proto/protobuf` | `@rules_rust_protobuf` |
| `@rules_rust//proto/prost` | `@rules_rust_prost` |
| `@rules_rust//wasm_bindgen` | `@rules_rust_wasm_bindgen` |

The `@rules_rust//crate_universe`, `@rules_rust//cargo`, `@rules_rust//tools`, and `@rules_rust//util` and any dependencies they may have would remain the same (in core `@rules_rust`).

I don't have strong opinions on changes to repo structure but I'm thinking it would be good to create an `extensions` directory where `bindgen`, `proto`, and `wasm_bindgen` could be moved into which I think would make it easier to add additional integrations down the line.

illicitonion:
Sounds good to me!

I guess we would still release all of these at the same time (i.e. there would always be matching versions)? Or would we move to more granular releases of each?

I think my only concern is that we'll need to be a little careful around compatibility (e.g. if we add/remove attrs to `crate_universe`, we introduce compatibility hazards if people start using older `rules_rust` with newer `bindgen` or similar).

But I'm happy to do the splitting up and work out the exact details about compatibility in the future - the bzlmod ecosystem as a whole is still working out compatibility somewhat...

UebelAndre:
[...]

PR #2998, opened 2024-11-13
Revert "Pin to Bazel 7.3.2"

UebelAndre:
Reverts bazelbuild/rules_rust#2957

Bazel 7.4.1 is now available https://github.com/bazelbuild/bazel/releases/tag/7.4.1

PR #3007, opened 2024-11-19
Moved rust extension rules into a separate `rules_rust_ext` workspace.

UebelAndre:
This change moves the `bindgen`, `proto`, and `wasm_bindgen` sub-packages into individual workspaces within the `extensions` directory. The intent is improve ease of maintenance of both core and extension Rust rules by ensuring changes to extensions have no impact on the core rules. Core rules should *never* depend on extensions.

Load statements should be updated according to the following table:

| before | after |
| --- | --- |
| `@rules_rust//bindgen` | `@rules_rust_bindgen//` |
| `@rules_rust//proto/prost` | `@rules_rust_prost//` |
| `@rules_rust//proto/protobuf` | `@rules_rust_protobuf//` |
| `@rules_rust//wasm_bindgen` | `@rules_rust_wasm_bindgen//` |

closes https://github.com/bazelbuild/rules_rust/issues/2882


UebelAndre:
@illicitonion @krasimirgg @hlopko I would love your input on this change. It's a massive breaking change but I think a better path forward to help ensure the `protobuf`/`rules_cc` headache from https://github.com/bazelbuild/rules_rust/pull/2998 doesn't impact the core rules again.

krasimirgg:
Thank you I think it looks very reasonable, but also I'm not working in an environment where rules_rust+bzlmod is exercised, so it's hard to flag potential non-trivial issues with this. I just asked a few folks to take a look at this and share their thoughts.

konkers:
Curious about the motivation to put them all into a single `@rules_rust_ext` module.  I would have naively expected them to have separate modules (i.e. `@rules_rust_bindgen`, `@rules_rust_prost`, etc.)   That way a project would have more fined grained control over it's dependencies.

UebelAndre:
> Curious about the motivation to put them all into a single `@rules_rust_ext` module. I would have naively expected them to have separate modules (i.e. `@rules_rust_bindgen`, `@rules_rust_prost`, etc.) That way a project would have more fined grained control over it's dependencies.

I started down that path and hit the 80 job cap for BazelCI. For each extension I would want to test them on Linux, Linux RBE, MacOS, and Windows, and to have each product be it's own workspace didn't seem possible. Though @meteorcloudy maybe there's a way to raise that cap?

illicitonion (approved):
Happy to do this, equally happy to have a per-extension module if that's practical :) Thanks for doing the work!

krasimirgg (approved):
Thank you! Looks good (if we manage to solve the testing CI limit situation, splitting this up more would be good too).

UebelAndre:
The commit I just added splits the extensions into their own repositories but hits the following error:

https://buildkite.com/bazel/rules-rust-rustlang/builds/12747#0193646f-6c25-4faa-a809-e5e319a1c9e6
```
The number of tasks in one config file is limited to 80!
fatal: pipeline parsing of "(stdin)" failed: EOF
```

@meteorcloudy @krasimirgg there's no way to raise this is there 😅

(Plus, it would be nice to be told how many were found)

meteorcloudy:
presubmit is now working https://buildkite.com/bazel/rules-rust-rustlang/builds/12752

UebelAndre:
> presubmit is now working https://buildkite.com/bazel/rules-rust-rustlang/builds/12752

With this in place I'll go ahead and update the PR to reflect the granular workspaces change instead of `rules_rust_ext`.

PR #3019, opened 2024-11-24
Add example with a non-workspace Cargo.toml

illicitonion:
We don't currently have any examples that show that this works, and I'm currently debugging an edge-case where it doesn't, so I'd like a little extra coverage.


UebelAndre (approved):
Then coverage you shall have!

Issue #2665, opened 2024-05-24
Build script env is overridden by use_default_shell_env in Bazel 6 


thesayyn:
We have bumped into an issue where rust_build_script rule sets some custom environment variables alongside the `use_default_shell_env`. 

Attribute in ctx.actions.run `use_default_shell_env` overrides the `env` attribute due to https://github.com/bazelbuild/bazel/issues/19317

https://github.com/bazelbuild/rules_rust/blob/df80ce61e418ea1c45c5bd51f88a440a7fb9ebc9/cargo/private/cargo_build_script.bzl#L316-L319

thesayyn:
Current workaround is to either patch rules_rust to disable `use_default_shell_env` or use `--incompatible_merge_fixed_and_default_shell_env` on `> Bazel 7 `

archshift:
This issue seems to have started manifesting with version 0.43. Any news on how to mitigate for Bazel 6?

illicitonion:
I'd be happy to review a change which added a config option in https://github.com/bazelbuild/rules_rust/blob/main/rust/settings/BUILD.bazel for making this behaviour optional, if someone wanted to prepare one?

sthornington:
I am running into this too. Honestly I'm so far down the yaks here I don't know if it's worth it, I was originally trying to work on https://github.com/bazelbuild/rules_rust/issues/2859 but I ran into this just trying to get `rules_rust` upgraded and working as a local repo.

sthornington:
If people are downgrading to 0.42 to avoid this issue, doesn't that make it a little more serious than "I would welcome a PR" ?

illicitonion:
New versions of rules_rust no longer support Bazel 6, and are no longer tested against it, so our official recommendation for Bazel 6 support is: Use a version of rules_rust which works in Bazel 6 (which in this case sounds like it's 0.42).

I'm happy to review something that makes Bazel 6 easier to use, if it doesn't have a large maintenance cost, but as (volunteer!) maintainers, we're not going to invest a lot of effort into Bazel 6 support.

sthornington:
Aha thanks for the background! I'll see what it would take for us to upgrade Bazel.

UebelAndre:
Closing as Bazel 6 is deprecated after Dec. 2025 per https://bazel.build/release

PR #3020, opened 2024-11-26
cargo: Allow disabling the `default_shell_env` for build scripts

ParkMyCar:
This PR adds a new boolean attribute to `cargo_build_script` rules called `use_default_shell_env` which is set to `True` by default.

The goal here is to make Cargo Build Scripts more hermetic. If you set `use_default_shell_env` to false, local environment variables like `PATH` should not get set for build scripts, preventing them from relying on tools not provided by Bazel.

Note: I considered adding a global `use_default_shell_env_for_cargo_build_scripts` setting, but couldn't figure out a good way to make the global setting work with this attribute. e.g. if you disable `use_default_shell_env` globally, ideally you still have a way to enable it for an individual crate, but I couldn't figure out how to set the defaults for the global setting and this local attr to get that to work

Related to https://github.com/bazelbuild/rules_rust/issues/2665

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -350,9 +350,8 @@ def _cargo_build_script_impl(ctx):
 
     # If enabled, start with the default shell env, which contains any --action_env
     # settings passed in on the command line and defaults like $PATH.
-    if ctx.attr.use_default_shell_env {
+    if ctx.attr.use_default_shell_env:
         env.update(ctx.configuration.default_shell_env)
-    }
 
     env.update({
         "CARGO_CRATE_NAME": name_to_crate_name(pkg_name),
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Thanks!

UebelAndre:
> Note: I considered adding a global `use_default_shell_env_for_cargo_build_scripts` setting, but couldn't figure out a good way to make the global setting work with this attribute. e.g. if you disable `use_default_shell_env` globally, ideally you still have a way to enable it for an individual crate, but I couldn't figure out how to set the defaults for the global setting and this local attr to get that to work

Does https://github.com/bazelbuild/rules_rust/pull/3065 satisfy this? A global kill-switch seemed nice so I figured I'd take a stab here. TLDR I took the same approach as stamping which I figured would be fine.

ParkMyCar:
> Does #3065 satisfy this? A global kill-switch seemed nice so I figured I'd take a stab here. TLDR I took the same approach as stamping which I figured would be fine.

#3065 is exactly what I was trying to achieve, I didn't make the jump to using `-1, 0, 1` though, nice work! Also it's nice that this aligns with the `stamp` attribute on rules too 👌

PR #3028, opened 2024-11-28
rust_analyzer: don't build a tree of RustAnalyzerInfos

sam-mccall:
This patch adds an `id` to RustAnalyzerInfo, and replace all the recursive RustAnalyzerInfos with the respective crate `id`s.
This fixes RustAnalyzerInfos becoming exponentially expensive to build in the presence of aliases.

---

Before this patch, `RustAnalyzerInfo.deps` contains the RustAnalyzerInfo of the target crate's deps, and so on recursively.
This forms a graph of Infos where there is one Info per target, but multiple paths from one target to another.
e.g. these could all point to `bar`: `foo.deps[0] == foo.deps[1].deps[0] == foo.aliases[0]`.

If we walk `foo` as a tree, we will see `bar` multiple times. Two operations that do this are `print(info)` and `{info: 42}` which hashes the object. As the graph grows, the number of paths grows combinatorically.

`RustAnalyzerInfo.aliases` is defined as a dict from `RustAnalyzerInfo => string`, so building this triggers the slowdown.

It would be possible to fix this by e.g. replacing this dict with a list of pairs.
However the work `rust_analyzer_aspect` does is fundamentally local and does not need a recursive data structure. Eliminating it means we can freely use these values as keys, print them, etc.

---

Timings for `ra_ap_rust-analyzer` (which heavily aliases its deps):

```
blaze build //third_party/bazel_rules/rules_rust/tools/rust_analyzer:gen_rust_project; \
time blaze run //third_party/bazel_rules/rules_rust/tools/rust_analyzer:gen_rust_project -- //third_party/rust/ra_ap_rust_analyzer/...

===Before===
Executed in  211.06 secs    fish           external
   usr time    0.24 secs    0.00 micros    0.24 secs
   sys time    1.24 secs  604.00 micros    1.24 sec

===After===
Executed in    3.24 secs    fish           external
   usr time    0.15 secs  389.00 micros    0.15 secs
   sys time    1.18 secs  125.00 micros    1.18 secs
```

### DIFF
--- a/extensions/prost/private/prost.bzl
+++ b/extensions/prost/private/prost.bzl
@@ -261,7 +261,10 @@ def _rust_prost_aspect_impl(target, ctx):
     # https://github.com/rust-analyzer/rust-analyzer/blob/2021-11-15/crates/project_model/src/workspace.rs#L529-L531
     cfgs = ["test", "debug_assertions"]
 
+    crate_id = "prost-" + crate_info.root.path
+
     rust_analyzer_info = write_rust_analyzer_spec_file(ctx, ctx.rule.attr, ctx.label, RustAnalyzerInfo(
+        id = crate_id,
         aliases = {},
         crate = dep_variant_info.crate_info,
         cfgs = cfgs,
@@ -332,7 +335,10 @@ def _rust_prost_library_impl(ctx):
                 transitive = transitive,
             ),
         ),
-        RustAnalyzerGroupInfo(deps = [proto_dep[RustAnalyzerInfo]]),
+        RustAnalyzerGroupInfo(
+            crate_specs = proto_dep[RustAnalyzerInfo].crate_specs
+            deps = proto_dep[RustAnalyzerInfo].deps,
+        ),
     ]
 
 rust_prost_library = rule(
--- a/extensions/prost/private/prost.bzl
+++ b/extensions/prost/private/prost.bzl
@@ -261,7 +261,7 @@ def _rust_prost_aspect_impl(target, ctx):
     # https://github.com/rust-analyzer/rust-analyzer/blob/2021-11-15/crates/project_model/src/workspace.rs#L529-L531
     cfgs = ["test", "debug_assertions"]
 
-    crate_id = "prost-" + crate_info.root.path
+    crate_id = "prost-" + dep_variant_info.crate_info.root.path
 
     rust_analyzer_info = write_rust_analyzer_spec_file(ctx, ctx.rule.attr, ctx.label, RustAnalyzerInfo(
         id = crate_id,
@@ -336,7 +336,7 @@ def _rust_prost_library_impl(ctx):
             ),
         ),
         RustAnalyzerGroupInfo(
-            crate_specs = proto_dep[RustAnalyzerInfo].crate_specs
+            crate_specs = proto_dep[RustAnalyzerInfo].crate_specs,
             deps = proto_dep[RustAnalyzerInfo].deps,
         ),
     ]
--- a/rust/private/rust_analyzer.bzl
+++ b/rust/private/rust_analyzer.bzl
@@ -81,8 +81,8 @@ def _accumulate_rust_analyzer_info(deps_details, dep):
         deps_details.crate_specs.append(dep[RustAnalyzerInfo].crate_specs)
     if RustAnalyzerGroupInfo in dep:
         for expanded_dep in dep[RustAnalyzerGroupInfo].deps:
-            deps_details.label_to_id[expanded_dep.crate.owner] = expanded_dep
-            deps_details.crate_specs.append(dep[RustAnalyzerGroupInfo].crate_specs)
+            deps_details.label_to_id[expanded_dep] = expanded_dep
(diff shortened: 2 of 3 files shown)


sam-mccall:
I misunderstood the cause of the performance problems here. The data aren't actually duplicated, it just appears that way if you (a) print it, or (b) hash it. I was misled by (a), and the rules use these infos as hash keys for aliases, triggering (b).

I do think eliminating the tree of RustAnalyzerInfos is a good idea if we don't need the power: it's harder to reason about the available data, carries risks of these performance cliffs, prevents print-debugging.
So I'd lean towards just rewriting the commit message, but if we want a less-invasive change, we could just fix how aliases are computed.

UebelAndre:
How does this impact `RustAnalyzerInfo` providers behind a transition? Take https://github.com/bazelbuild/rules_rust/pull/3035 for example, the `wasm_file` attribute of `rust_wasm_bindgen` may have `target_compatible_with = ["@platforms//os:wasm32"]` set on it which would make it skipped in normal builds outside of that transition. If trees are not built, will the dependency graph of the wasm target be lost?

sam-mccall:
The Info will still be built for each target the aspect propagates to, and the json specs will still be gathered in the depset in the same way. The only thing that changes is the data structure used is not recursive.

I don't know specifically how transitions affect the aspect's behavior in that example, but I'm fairly sure it's the same before/after the patch.

rust/private/rust_analyzer.bzl
```diff
@@ -72,20 +73,16 @@ def write_rust_analyzer_spec_file(ctx, attrs, owner, base_info):
 
     return rust_analyzer_info
 
-def _accumulate_rust_analyzer_info(dep_infos_to_accumulate, label_index_to_accumulate, dep):
+def _accumulate_rust_analyzer_info(deps_details, dep):
```
krasimirgg, line 76: nit: (this is somewhat pre-existing, but the new version makes it a bit worse): on a first read, I don't see what deps_details looks like. It becomes clear when I reach the initialization below. 

Maybe could you declare a provider for the shape of this and use it in docs.

extensions/prost/private/prost.bzl
```diff
@@ -261,7 +261,10 @@ def _rust_prost_aspect_impl(target, ctx):
     # https://github.com/rust-analyzer/rust-analyzer/blob/2021-11-15/crates/project_model/src/workspace.rs#L529-L531
     cfgs = ["test", "debug_assertions"]
 
+    crate_id = "prost-" + dep_variant_info.crate_info.root.path
```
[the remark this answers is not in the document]
sam-mccall: We don't need an extra prefix, I put it there to avoid encoding/suggesting that this should be exactly the root path.
Happy to remove if you prefer.

(and because rust_analyzer.bzl uses a prefix presumably for the same reason)

rust/private/rust_analyzer.bzl
sam-mccall: Yeah, starlark doesn't have useful types :-(

Providers do indeed have the required structure but I think using them for non-provider data (while calculating provider data) is going to be actively misleading.

There's not much logic here, and I think it's easier to just inline it so the data structures can be understood from context.

krasimirgg (approved):
Thank you!

UebelAndre:
This change seems to have broken the ability for rust-analyzer to index into prost rules. None of the proto targets are being indexed.

With this change: [rust-project-main.json](https://github.com/user-attachments/files/18114517/rust-project-main.json)
Revert: [rust-project-revert.json](https://github.com/user-attachments/files/18114516/rust-project-revert.json)

@sam-mccall @krasimirgg would you be able to take a look?

sam-mccall:
@UebelAndre sorry, I wasn't able to get to this until now - I've been out sick.

I'll take a look: at first glance all the crates are there, it looks like some deps from `*_test` targets => protos are missing. Not sure why.

Do you know why Prost doesn't return a regular `CrateInfo` in the first place? [rust_proto_library](https://github.com/protocolbuffers/protobuf/blob/main/rust/defs.bzl) does this and seems to "just work" with rust-analyzer, without the rules having to know anything about RustAnalyzerInfo.

UebelAndre:
> @UebelAndre sorry, I wasn't able to get to this until now - I've been out sick.

No worries, glad you're feeling better!

> Do you know why Prost doesn't return a regular `CrateInfo` in the first place? [rust_proto_library](https://github.com/protocolbuffers/protobuf/blob/main/rust/defs.bzl) does this and seems to "just work" with rust-analyzer, without the rules having to know anything about RustAnalyzerInfo.

[Prost returns `CrateGroupInfo`](https://github.com/bazelbuild/rules_rust/blob/0e2fbbf74dc4119dd5d75a576615d36fe5479337/extensions/prost/private/prost.bzl#L380-L385) since unlike the `rust_proto_library` rule, the crate is compiled on an action for the `proto_library` target via an aspect and can be representative of transitive crates as well. Thus the aspect needed to do the heavy lifting for rust-analyzer. Or at least it was thought to be needed once upon a time. If `CrateGroupInfo` is sufficient for populating `RustAnalyzerInfo` then that'd certainly simplify the implementation 😄

PR #3030, opened 2024-12-01
Improve workspace discovery logic

illicitonion:
Previously we were assuming that any crate with a path dep lives in a
workspace, and any that doesn't isn't.

This isn't correct logic.
https://doc.rust-lang.org/cargo/reference/workspaces.html describes the
specification. We do filesystem traversal to detect implicit members.
This is a little more expensive, but much more correct.

This is a precursor for supporting path dependencies - the current
incorrect logic incorrectly identifies any Cargo.toml file containing a
path dependency as expecting to live in a workspace, which breaks some
use-cases.

### DIFF
--- a/crate_universe/Cargo.toml
+++ b/crate_universe/Cargo.toml
@@ -1,6 +1,46 @@
 [workspace]
 members = ["tools/cross_installer", "tools/urls_generator"]
-exclude = ["test_data"]
+exclude = [
+    "test_data/metadata/abspath",
+    "test_data/metadata/aliases",
+    "test_data/metadata/build_scripts",
+    "test_data/metadata/common",
+    "test_data/metadata/crate_combined_features",
+    "test_data/metadata/crate_optional_deps_disabled",
+    "test_data/metadata/crate_optional_deps_disabled_build_dep_enabled",
+    "test_data/metadata/crate_optional_deps_enabled",
+    "test_data/metadata/crate_renamed_optional_deps_disabled",
+    "test_data/metadata/crate_renamed_optional_deps_enabled",
+    "test_data/metadata/crate_types",
+    "test_data/metadata/example_proc_macro_dep",
+    "test_data/metadata/git_repos",
+    "test_data/metadata/has_package_metadata",
+    "test_data/metadata/host_specific_build_deps",
+    "test_data/metadata/multi_cfg_dep",
+    "test_data/metadata/multi_kind_proc_macro_dep",
+    "test_data/metadata/nested_build_dependencies",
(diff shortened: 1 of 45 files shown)


crate_universe/src/splicing/splicer.rs
```diff
@@ -1208,9 +1157,6 @@ mod test {
 
                 [lib]
                 path = "lib.rs"
-
-                [dependencies]
-                neighbor = { path = "../neighbor" }
```
UebelAndre, line 1213: Is there a unit test that shows path dependencies working correctly now?

examples/crate_universe_local_path/vendor_lazy_static.sh
```diff
@@ -12,6 +12,12 @@ elif [[ "$#" -eq 1 ]]; then
     path_dep_path="$1"
     copy_to="crates_from_workspace/$1"
     mkdir -p "${copy_to}"
+    sed_i=(sed -i)
+    if [[ "$(uname)" == "Darwin" ]]; then
+      sed_i=(sed -i '')
+    fi
+    "${sed_i[@]}" -e 's#manifests = \["//crates_from_workspace:Cargo\.toml"\],#manifests = ["//crates_from_workspace:Cargo.toml", "//crates_from_workspace:'"$1"'/Cargo.toml"],#g' WORKSPACE.bazel
```
UebelAndre, line 19: Is this to support vendoring more things?

UebelAndre (changes requested):

examples/crate_universe_local_path/vendor_lazy_static.sh
illicitonion: We run this example in two modes, and this script is for setting up the test data.

If you don't supply an arg to this script, it just vendors into a tmpdir somewhere external to the repo and pulls that in as a path dependency.

This line is to support the other mode - if you specify a relative path within this repo, it will vendor the crate into _that_ path. Now that we correctly detect that vendored crate as a member of the workspace, we need to add it to the manifests of the repo rules :)

crate_universe/Cargo.toml
```diff
@@ -1,6 +1,46 @@
 [workspace]
 members = ["tools/cross_installer", "tools/urls_generator"]
-exclude = ["test_data"]
+exclude = [
```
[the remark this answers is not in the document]
illicitonion: Unfortunately not: https://github.com/rust-lang/cargo/issues/6009

UebelAndre (approved):
Nice!

PR #3031, opened 2024-12-01
rust_analyzer: generate more reqired sources

sam-mccall:
The existing support for this has limitations:

- only works for sources that are specially tagged (rust_generated_srcs).
  I've removed this mechanism as it's no longer needed.
- only provides sources of the directly selected library, but
  rust-analyzer needs to parse its dependencies too.
- is missing the target.json file, which rust-analyzer requires to
  parse the library (it invokes rustc --print cfg --target=...)

The dependency depth to generate sources for is a tradeoff. More layers
of deps will give more accurate analysis but be slower to generate.

My guess is that in the end we'll want full transitive sources, but
direct deps should be a non-controversial starting point. (Without these
sources, rust-analyzer can't understand the direct API usage at all).
We can change this later and probably need to reconcile with #3028.

### DIFF
--- a/extensions/prost/private/prost.bzl
+++ b/extensions/prost/private/prost.bzl
@@ -279,7 +279,6 @@ def _rust_prost_aspect_impl(target, ctx):
             package_info = package_info_file,
         ),
         rust_analyzer_info,
-        OutputGroupInfo(rust_generated_srcs = [lib_rs]),
     ]
 
 rust_prost_aspect = aspect(
--- a/extensions/protobuf/proto.bzl
+++ b/extensions/protobuf/proto.bzl
@@ -233,7 +233,6 @@ def _rust_proto_compile(protos, descriptor_sets, imports, crate_name, ctx, is_gr
         ),
         output_hash = output_hash,
     )
-    providers.append(OutputGroupInfo(rust_generated_srcs = srcs))
     return providers
 
 def _rust_protogrpc_library_impl(ctx, is_grpc):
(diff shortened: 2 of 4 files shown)


krasimirgg:
The changes to rust_analyzer look good. I'm not sure if the `rust_generated_srcs` output group is used for anything else though?

It would be nice to add some tests to make sure that the expected sources are made available to the expected actions.

sam-mccall:
> (sorry I left a comment in the wrong review state) The changes to rust_analyzer look good. I'm not sure if the rust_generated_srcs output group is used for anything else though?

It's not used for anything else.

> It would be nice to add some tests to make sure that the expected sources are made available to the expected actions.

I could extend generate_srcrs_test to add some. The problem is that the test driver runs `blaze build //...`, forcing all the generated files to exist in any case. It also strips out `manual` tags which would let generated files opt out of being built...

I can try to remove that behavior from the test driver, but I don't understand why it does that to start with. Or I can pile hacks on top (more test tags). What would you prefer?

krasimirgg:
> I could extend generate_srcrs_test to add some. The problem is that the test driver runs `blaze build //...`, forcing all the generated files to exist in any case. It also strips out `manual` tags which would let generated files opt out of being built...
> 
> I can try to remove that behavior from the test driver, but I don't understand why it does that to start with. Or I can pile hacks on top (more test tags). What would you prefer?

Seems like getting the bash-style test to reliably test using generated vs. non-generated files might be tricky.
I was thinking more about a separate blaze analysis-time test that sets up a build graph which includes generated sources, attaches the aspect on a "main" target under test and examines the OutputGroupInfo-s.

Issue #2755, opened 2024-07-22
Integration with rust-analyzer's automatic project discovery


davidbarsky:
Hi folks! I added Cargo-style automatic project discovery in rust-analyzer for non-Cargo build systems (Buck primarily, but it's designed to be usable by Bazel...) in https://github.com/rust-lang/rust-analyzer/pull/17246. The documentation for this feature is in the manual, under "rust-analyzer.workspace.discoverConfig". I've been using it with Buck for the last month and a half, and while some details are still likely to change, the general shape of the API is in a spot that I'm happy with. I think this would make the Rust IDE experience with Bazel a lot nicer and I'd be very happy to provide guidance on integrating this feature into rules_rust!

jondo2010:
For reference, the Buck-side implementation of this can be found here: https://github.com/facebook/buck2/tree/main/integrations/rust-project/src

otiv-emiel-vanseveren:
Hi, @davidbarsky  I read a few of the discussions (including: https://github.com/rust-lang/rust-analyzer/pull/17246). But I don't fully understand the concept/advantages of ra's automatic project discovery. Would you mind explaining? In what ways will it improve the Rust Bazel IDE experience?

davidbarsky:
> But I don't fully understand the concept/advantages of ra's automatic project discovery. Would you mind explaining? In what ways will it improve the Rust Bazel IDE experience?

I think it improves the situation in two major ways:
1. For VS Code, Bazel directs users [to setup a task](https://bazelbuild.github.io/rules_rust/rust_analyzer.html#vscode) that generates a `rust-project.json` for the _entire_ workspace. In my opinion as a rust-analyzer team member, this has two downsides:
    1. Generating a `rust-project.json` for the _entire_ workspace might not scale to larger workspaces. I don't believe this is an inherent limitation to `@rules_rust//tools/rust_analyzer:gen_rust_project `, but https://github.com/rust-lang/rust-analyzer/pull/17246 can be substantially more fine-grained with what is and isn't indexed.
    2. `@rules_rust//tools/rust_analyzer:gen_rust_project` runs on editor starting up, but doesn't necessarily refresh the workspace if dependencies are added or removed. The PR I linked to _can_ reload when a BUILD file changes.
[...]

PR #3033, opened 2024-12-02
rust_analyzer: make all paths in rust-project.json absolute

sam-mccall:
This is a step towards supporting automatic project discovery (#2755).

Relative paths are resolved against the location of rust-project.json,
but in auto-discovery this is the location of the BUILD file being
discovered, not the workspace root.

We already use absolute paths for generated files, so rust-project.json
is not workspace-independent today.

The alternatives seem more complex for little gain:

- Only use absolute paths for auto-discovery mode, relative otherwise.
- Use ../../ to express all workspace paths relative to the BUILD.

See https://github.com/bazelbuild/rules_rust/issues/2755#issuecomment-250696107


krasimirgg (approved):

sam-mccall:
@krasimirgg CI is broken, can you re-trigger or merge?

PR #3037, opened 2024-12-03
Added bzlmod support to all extension rules.

UebelAndre:
Additional Changes:
- The rules `rules_nodejs` for `rules_rust_wasm_bindgen` will no longer tested and will be dropped unless bzlmod support is added for the original `bazel_build_rules_nodejs` rules.
- `rules_rust_bindgen` updated default clang version to 17.0.3 to match bzlmod available version.
- `rules_rust_prost` updated protobuf version to `v28.3`

Relates to:
- https://github.com/bazelbuild/rules_rust/issues/2181


.bazelci/presubmit.yml
```diff
@@ -936,7 +972,11 @@ tasks:
     build_targets:
       - "//..."
     test_targets:
+      - "--"
       - "//..."
+      # TODO: https://github.com/bazelbuild/rules_rust/issues/3039
```
UebelAndre, line 977: https://github.com/bazelbuild/rules_rust/issues/3039

illicitonion (approved):

extensions/prost/transitive_repositories.bzl
```diff
@@ -1,6 +1,7 @@
 """Definitions for loading transitive `@rules_rust//proto` dependencies"""
 
 load("@bazel_features//:deps.bzl", "bazel_features_deps")
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
```
sthornington, line 4: why does prost require all these protobuf dependencies now? I thought the point of it was that it doesn't?

UebelAndre: I added this because when I updated protobuf I started getting tons of load errors for random external repositories. If there’s a way to avoid this I’m happy to reduce the amount of forced dependencies!

sthornington: I have no idea sorry.  :(  I wonder how it worked before though?

UebelAndre: At a glance it seems like protobuf just started adding more dependencies expecting bzlmod to hide that from users 😞

UebelAndre: So updating and pulling in all those changes that introduced new forced dependencies introduced the need for these extra dependencies.

PR #3040, opened 2024-12-04
Improve consolidation of rust-analyzer crate specs with generated sources

martingms:
I have pretty much the exact setup that is tested in `test/rust_analyzer/generated_srcs_test`, which currently seems to fail on the main branch (at least on my machine).

Explicitly, the issue is that when consolidating the specs for a `rust_library` with a mix of generated and non-generated sources and a corresponding `rust_test` with `crate` set to the library, the following is wrong in the generated `rust-project.json` if the file for the test crate is evaluted first:
* The `"source"`-field is unpopulated
* The `"root_module"`-field points to a symlink(?) in the execroot instead of the `lib.rs` file that is in the workspace.
This confuses rust-analyzer.

I don't know the `rules_rust` codebase well enough to say whether this is the correct place to address these issues, and I see @sam-mccall is curently working on related stuff, so I just made the smallest possible change that fixed my issue.

Feel free to just take this as a bug-report if that's more convenient!

### DIFF
--- a/test/rust_analyzer/generated_srcs_test/rust_project_json_test.rs
+++ b/test/rust_analyzer/generated_srcs_test/rust_project_json_test.rs
@@ -48,7 +48,9 @@ mod tests {
             .find(|c| &c.display_name == "generated_srcs")
             .unwrap();
         assert!(gen.root_module.starts_with("/"));
-        assert!(gen.root_module.ends_with("/lib.rs"));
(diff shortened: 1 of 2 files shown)


martingms:
Note that this and #3127 attempts to fix the same issue in slightly different ways. I have no preference as to which is merged, but would be nice to get one of them in :)

PR #3025, opened 2024-11-27
External path dependencies work

illicitonion:
This allows dependencies of kind path to work, both in the
[dependencies] table and the [patches] table.

The intention is that this is used for third-party code not first-party
code.

PR #3042, opened 2024-12-04
Release 0.55.0

UebelAndre:


illicitonion (approved):
Happy to do this, but wouldn't mind merging #3025 and #3030 before we do :)

PR #3049, opened 2024-12-05
Fix release workflow on macos

UebelAndre:
This avoids the following error
```

The macOS-12 environment is deprecated, consider switching to macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see https://github.com/actions/runner-images/issues/10721

```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -50,7 +50,7 @@ jobs:
       matrix:
         # Create a job for each target triple
         include:
-          - os: macOS-14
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Fun!

illicitonion (commented):
Actually, we probably want to bump to 13 for maximum compatibility of our produced artifacts?

UebelAndre:
> Actually, we probably want to bump to 13 for maximum compatibility of our produced artifacts?

Updated!

PR #3051, opened 2024-12-05
Release 0.55.1

UebelAndre:


illicitonion (approved):
Happy to try! Or to wait!

illicitonion (commented):
Wait... Do we need to bump the versions in the other extensions too?

I see extensions/prost/MODULE.bazel still as 0.54.1 (and depending on rules_rust 0.54.1)...

UebelAndre:
> Wait... Do we need to bump the versions in the other extensions too?
> 
> I see extensions/prost/MODULE.bazel still as 0.54.1 (and depending on rules_rust 0.54.1)...

Great catch! I created https://github.com/bazelbuild/rules_rust/pull/3052 to prevent this in the future

UebelAndre:
I think https://github.com/bazelbuild/rules_rust/pull/3053 suggests that we should cut this release.

PR #3057, opened 2024-12-06
Run CI without building runfiles links

UebelAndre:
Round two of https://github.com/bazelbuild/rules_rust/pull/2340


UebelAndre:
I think this works now (in Bazel 8)!

illicitonion (approved):

Issue #3059, opened 2024-12-06
Workspace discovery


bakjos:
The new workspace discovery fails when there are multiple workspace roots in the repo, in my case only one of the roots is compiled using crates_universe, is there a way to pick the workspace root? the same happens if the Cargo.toml is a non-standard one, `Cargo.universe.toml`.  Would it be possible to add a workspace root parameter to the `crates_repository` macro?

PR #3060, opened 2024-12-06
feat: Add workspace_cargo_toml attribute

bakjos:
This PR adds a new workspace toml parameter to allow picking one cargo workspace file when there are multiple in the base repo. 

https://github.com/bazelbuild/rules_rust/issues/3059


illicitonion (commented):
Thanks for putting together - the approach generally looks reasonable, but I left a few comments. Thanks!

PR #3062, opened 2024-12-07
Fix release notes generation for extensions

UebelAndre:
This was a miss from https://github.com/bazelbuild/rules_rust/pull/3007 and https://github.com/bazelbuild/rules_rust/pull/3037


UebelAndre:
I've gone and updated older releases to have the correct `integrity` values.

illicitonion (approved):

PR #3063, opened 2024-12-08
Add test for editing patch Cargo.toml files

illicitonion:
This tests that the repository rules are properly invalidated when patched Cargo.toml files are edited.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -538,17 +538,13 @@ tasks:
     platform: ubuntu2004
     working_directory: examples/crate_universe_local_path
     run_targets:
-      - "//:vendor_lazy_static_out_of_tree"
-    test_targets:
-      - "//..."
+      - "//:vendor_edit_test_out_of_tree"
   crate_universe_local_path_in_tree:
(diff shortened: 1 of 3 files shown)


.bazelci/presubmit.yml
```diff
@@ -538,17 +538,13 @@ tasks:
...
-    test_targets:
-      - "//..."
+      - "//:vendor_edit_test_out_of_tree"
   crate_universe_local_path_in_tree:
     name: Crate Universe Local Path In Tree
     platform: ubuntu2004
     working_directory: examples/crate_universe_local_path
     run_targets:
-      - "//:vendor_lazy_static_in_tree"
-    test_targets:
```
UebelAndre, line 550: Why delete `test_targets`?

UebelAndre (changes requested):

.bazelci/presubmit.yml
illicitonion: The run script now does the `bazel test` an part of an integration test, and it doesn't seem useful doing the test twice

UebelAndre (approved):

PR #3075, opened 2024-12-10
Use `abs` from utils instead of the global

### LINKED ISSUE #3068: Use of `abs` bazel function removes compatibility from bazel <7
This was added here https://github.com/bazelbuild/rules_rust/commit/8e962a16d9f4a243e739320015b5d7f5c357cc68

Seems like `abs` is not available in 6.x

avdv:
Fixes #3068

### DIFF
--- a/cargo/private/cargo_utils.bzl
+++ b/cargo/private/cargo_utils.bzl
@@ -1,7 +1,20 @@
 """Utility functions for the cargo rules"""
 
 load("//rust/platform:triple_mappings.bzl", "system_to_binary_ext")
-load("//rust/private:utils.bzl", "abs")
+
+# TODO: remove after dropping support for Bazel < 7 when `abs` is a global
+def abs(value):
+    """Returns the absolute value of a number.
+
+    Args:
+      value (int): A number.
+
+    Returns:
+      int: The absolute value of the number.
+    """
+    if value < 0:
+        return -value
+    return value
 
 def _resolve_repository_template(
         *,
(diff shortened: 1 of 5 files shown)


cargo/private/cargo_utils.bzl
```diff
@@ -1,6 +1,7 @@
 """Utility functions for the cargo rules"""
 
 load("//rust/platform:triple_mappings.bzl", "system_to_binary_ext")
+load("//rust/private:utils.bzl", "abs")
```
avdv, line 4: Seems I cannot depend on the private utils.bzl file from here. Should I just add the code here?

avdv: I have just added a copy of the `abs` function to `cargo/private/cargo_utils.bzl`.

cargo/private/cargo_utils.bzl
```diff
@@ -2,6 +2,20 @@
 
 load("//rust/platform:triple_mappings.bzl", "system_to_binary_ext")
 
+# TODO: remove after dropping support for Bazel < 7 when `abs` is a global
```
illicitonion, line 5: Our minimum supported bazel version is already 7.3.1 - I'm happy to merge this if it's convenient for folks, but just FYI we don't officially support Bazel 6 :)

illicitonion (approved):
Works for me as a convenience for folks, but FYI we already don't support Bazel 6 :)

avdv:
> Works for me as a convenience for folks, but FYI we already don't support Bazel 6 :)

OK, fair enough. Is this documented somewhere or just common knowledge?

Thank you for merging. :)

illicitonion:
Our support policy can be found at https://bazelbuild.github.io/rules_rust/#supported-bazel-versions

PR #2862, opened 2024-09-11
Allow `crates_vendor` to read the `CARGO_BAZEL_DEBUG` env var

UebelAndre:
Previously the environment variable was discarded. Now users can see debug info about how crates are being rendered.

PR #3078, opened 2024-12-10
Quote environ when injecting into crates vendor wrapper

ryanpbrewster:
Slight tweak to the logic in https://github.com/bazelbuild/rules_rust/pull/2862 (cc: @UebelAndre)

My `$PATH` variable has spaces in it and I hit an error like
```
env: Support/JetBrains/Toolbox/scripts:/Users/ryanpbrewster/.bin:/Users/ryanpbrewster/config/bin:/Users/ryanpbrewster/.local/bin:/Applications/Visual: No such file or directory                                                                     
```

If I manually modify `bazel-bin/rust/crates_vendor.sh` to modify
```
exec env - \
${_ENVIRON[@]} \
```
into
```
exec env - \
"${_ENVIRON[@]}" \
```
it starts working.

### DIFF
--- a/crate_universe/private/crates_vendor.bzl
+++ b/crate_universe/private/crates_vendor.bzl
@@ -30,7 +30,7 @@ fi
 # --default_system_javabase set to the empty string, but if you provide a path,
 # it may set it to a value (eg. "/usr/local/buildtools/java/jdk11").
 exec env - \\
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Thanks!

Issue #2045, opened 2023-07-01
Add support for injecting rust code into generated prost libraries


UebelAndre:
This is a continuation of https://github.com/bazelbuild/rules_rust/pull/2033#discussion_r1246859424

illicitonion:
> I'm imagining this taking the form roughly of a new attribute on `rust_*_library` along the lines of:
> 
> ```
> extra_per_package_code = {
>   "my.proto.package": "path/to:extra.rs",
> }
> ```
> 
> or something like that? But very open to ideas!

UebelAndre:
This sounds like a nice idea but how would the `rust_prost_library` attribute make it's way into the aspect, where the crate is ultimately compiled?

UebelAndre:
I think this will probably need to be implemented on the `rust_prost_toolchain` where there’s a [label_keyed_string_dict](https://bazel.build/rules/lib/toplevel/attr#label_keyed_string_dict) of filegroups containing extra sources and label patterns to use for filtering those sources into proto files.

PR #3083, opened 2024-12-11
Added `rust_prost_transform` rule for modifying granular proto_library

UebelAndre:
This new rule is to support granular modifications to existing `proto_library` targets without adding duplicate actions to the build graph depending on what `rust_prost_library` target was consumed.

closes https://github.com/bazelbuild/rules_rust/issues/2045

### DIFF
--- a/docs/src/rust_prost.md
+++ b/docs/src/rust_prost.md
@@ -192,3 +192,57 @@ Rust Prost toolchain rule.
 | <a id="rust_prost_toolchain-tonic_runtime"></a>tonic_runtime |  The Tonic runtime crates to use.   | <a href="https://bazel.build/concepts/labels">Label</a> | optional |  `None`  |
 
 
+<a id="rust_prost_transform"></a>
+
+## rust_prost_transform
+
+<pre>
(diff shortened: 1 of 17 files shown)


UebelAndre:
> I'd be slightly tempted to say we should add this as a `dict[ProtoLabel, TransformLabel]` to `prost_proto_library` rather than adding it to the `proto_library` itself (using `data` for this feels... Convenient, but wrong?)

We discussed this offline but for posterity, the way the rules work is by using an aspect to perform the protogen and rustc compile on the `proro_library` targets themselves to try and avoid issues around duplicate symbols. I know of no way to have the rule inject data to the aspect action. The only way I could think of was to add such a mapping to the toolchain but that seemed untenable. Thus I've come up with this idea which is primarily intended for devs to add additional prost info to *their* `proto_library` targets (external will be harder to modify).

samschlegel:
> I know of no way to have the rule inject data to the aspect action.

You should be able to just use `ctx.attr` here as demonstrated in [the Aspects intro docs](https://bazel.build/extending/aspects#aspect_definition_2). Adding stuff to the data this way shouldn't be required, and there should be other ways to get around the duplicate symbols such as using the name of the aspect rule as the crate name instead of the proto_library?

UebelAndre:
> > I know of no way to have the rule inject data to the aspect action.
> 
> You should be able to just use `ctx.attr` here as demonstrated in [the Aspects intro docs](https://bazel.build/extending/aspects#aspect_definition_2). Adding stuff to the data this way shouldn't be required, and there should be other ways to get around the duplicate symbols such as using the name of the aspect rule as the crate name instead of the proto_library?

I’m happy to take a look at other pull requests that can provide the same functionality in a cleaner way 😄

Issue #3045, opened 2024-12-05
rustc: set `--embed-bitcode=no` if LTO is not enabled


ParkMyCar:
Whether or not `rustc` embeds LLVM bitcode, which is used for LTO, is controlled via the `--embed-bitcode` flag. When LTO is disabled Cargo will automatically set `--embed-bitcode=no` because embedding bitcode can significantly slow down compile times ([docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#embed-bitcode)). It would be great if `rules_rust` could do the same!

PR #3104, opened 2024-12-15
feat: Add `//rust/settings:lto`

ParkMyCar:
Fixes https://github.com/bazelbuild/rules_rust/issues/3045

This PR adds a new build setting `//rust/settings:lto=(off|thin|fat)` which changes how we specify the following flags:

* [`lto`](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto)
* [`embed-bitcode`](https://doc.rust-lang.org/rustc/codegen-options/index.html#embed-bitcode)
* [`linker-plugin-lto`](https://doc.rust-lang.org/rustc/codegen-options/index.html#linker-plugin-lto)

The way we invoke the flags was based on how Cargo does it today ([code](https://github.com/rust-lang/cargo/blob/769f622e12db0001431d8ae36d1093fb8727c5d9/src/cargo/core/compiler/lto.rs#L4)) and based on suggestions from the [Rust docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#embed-bitcode).

When LTO is not enabled, we will specify `-Cembed-bitcode=no` which tells `rustc` to skip embedding LLVM bitcode and should speed up builds. Similarly when LTO is enabled we specify `-Clinker-plugin-lto` which will cause `rustc` to skip generating objects files entirely, and instead replace them with LLVM bitcode*.

*only when building an `rlib`, when building other crate types we continue generating object files.

I added unit tests to make sure we pass the flags correctly, as well as some docs describing the new setting. Please let me know if I should add more!

### DIFF
--- a/docs/src/rust.md
+++ b/docs/src/rust.md
@@ -12,6 +12,7 @@
 * [error_format](#error_format)
 * [extra_rustc_flag](#extra_rustc_flag)
 * [extra_rustc_flags](#extra_rustc_flags)
+* [lto](#lto)
 * [capture_clippy_output](#capture_clippy_output)
 
 <a id="capture_clippy_output"></a>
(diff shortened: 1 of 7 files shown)


UebelAndre (approved):
Nice! Thank you!

PR #3034, opened 2024-12-02
Added bzlmod support to core rules_rust

UebelAndre:
Relates to:
- https://github.com/bazelbuild/rules_rust/issues/2181

PR #3107, opened 2024-12-15
Do not cache rustc info in deps resolver

hauserx:
The cargo_tree_resolver discovers dependencies by executing `cargo tree` for different host and target triples. This procedure breaks when cargo successfully but incorrectly caches rustc info. Incorrect cache hit may happen because cache key does not take into account HOST_TRIPLE env variable used by `rules_rust` to force rustc to report it's built for different host.

The caching may happen if one has target/ directory in repo, or it was created through running some commands outside of bazel.

# Details

- After recent fixes to support properly bzlmod
(https://github.com/bazelbuild/rules_rust/pull/3034) noticed issues under windows when building bazel-lsp: https://github.com/cameron-martin/bazel-lsp/issues/92

- Cargo may cache rustc info in target/.rustc_info.json : https://github.com/rust-lang/cargo/blob/769f622e12db0001431d8ae36d1093fb8727c5d9/src/cargo/util/rustc.rs#L163

- The rules_rust hacks rustc by setting HOST_TRIPLE: https://github.com/bazelbuild/rules_rust/blob/3aecdbedba0c001d0660ff631aeccb35d94ff3a7/crate_universe/src/metadata/cargo_tree_resolver.rs#L152

- The HOST_TRIPLE env variable is not taken into account by cargo when checking whether cached rustc info is valid:
https://github.com/rust-lang/cargo/blob/769f622e12db0001431d8ae36d1093fb8727c5d9/src/cargo/util/rustc.rs#L320

### DIFF
--- a/crate_universe/tests/cargo_integration_test.rs
+++ b/crate_universe/tests/cargo_integration_test.rs
@@ -526,16 +526,24 @@ fn host_specific_build_deps() {
     }
 
     let r = runfiles::Runfiles::create().unwrap();
+
+    let src_cargo_toml = runfiles::rlocation!(
+        r,
+        "rules_rust/crate_universe/test_data/metadata/host_specific_build_deps/Cargo.toml"
+    )
+    .unwrap();
+
+    // Put Cargo.toml into writable directory structure and create target/ directory to verify that
+    // cargo does not incorrectly cache rustc info in target/.rustc_info.json file.
+    let scratch = tempfile::tempdir().unwrap();
+    let cargo_toml = scratch.path().join("Cargo.toml");
+    fs::copy(src_cargo_toml, &cargo_toml).unwrap();
+    fs::create_dir(scratch.path().join("target")).unwrap();
+
     let metadata = run(
         "host_specific_build_deps",
         HashMap::from([(
-            runfiles::rlocation!(
-                r,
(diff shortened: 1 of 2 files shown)


crate_universe/tests/cargo_integration_test.rs
```diff
@@ -526,16 +526,24 @@ fn host_specific_build_deps() {
...
     let r = runfiles::Runfiles::create().unwrap();
+
+    let src_cargo_toml = runfiles::rlocation!(
+        r,
+        "rules_rust/crate_universe/test_data/metadata/host_specific_build_deps/Cargo.toml"
+    )
+    .unwrap();
+
+    // Put Cargo.toml into writable directory structure and create target/ directory to verify that
+    // cargo does not incorrectly cache rustc info in target/.rustc_info.json file.
```
UebelAndre, line 537: Should there be an assert that this file was never created?

UebelAndre (changes requested):
Thanks! I just had one question 😄

UebelAndre (approved):

illicitonion:
@hauserx Could I bother you to file an issue (or PR!) against Cargo for this behaviour? It seems wrong that they're caching something and omitting relevant input from the cache key?

Thanks for the PR here!

Issue #3008, opened 2024-11-19
Add appropriate resource set for Rustc `codegen-units`


UebelAndre:
https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units 

> This flag controls the maximum number of code generation units the crate is split into. It takes an integer greater than 0.
>
> When a crate is split into multiple codegen units, LLVM is able to process them in parallel. Increasing parallelism may speed up compile times, but may also produce slower code. Setting this to 1 may improve the performance of generated code, but may be slower to compile.
>
> The default value, if not specified, is 16 for non-incremental builds. For incremental builds the default is 256 which allows caching to be more granular.

If this is true, then by default, a single `Rustc` action consumes 16 CPU and should be reflected in Bazel's scheduling. An alternative is to force this value to be 1.

freeformstu:
I really don't think you should force it to be 1, that will significantly slow down compile times.

My suggestion would be to set `codegen-units` equal to the number of `.rs` files with a configurable min/max. Since rust defaults to 16, 16 should probably be the default floor (although I'd certainly be open to setting this lower for small crates if the performance difference is negligible). If you have a massive crate (dozens of files or more), setting this higher should definitely help. Memory should similarly be scaled as well. Obviously this is an imprecise approximation because you could have a crate with one massive file in which case you should probably set this value explicitly on that rule.

I'm not sure what configuration options exist for parallelizing linking but I'm fairly certain a lot of our larger crate's slow build times are a result of slow linking (eg. binary rules are often a single file but have many dependencies). If there's any way to increase parallelism in the same way (based on number of dependencies in this case) that would be fantastic.

UebelAndre:
cc @hlopko @scentini @krasimirgg wondering if you folks have any thoughts too.

UebelAndre:
> An alternative is to force this value to be 1.

Also to clarify, it should be forced to match Bazel's expectation, which is 1 CPU per action. I'm not advocating for this to be forced to 1 as a reasonable default. I just want the scheduling to be representative.

havasd:
I think it is 2-fold.

We use codegen-units=1 in our toolchain because of optimization reasons (it doesn't seem too slow).
However, we run linker as part of rustc as well afaik. Because of static linking for binaries that is a heavy deal and imho it is beneficial to allow linker accessing more resources otherwise we that can get really slow

UebelAndre:
> I think it is 2-fold.
> 
[...]

PR #3112, opened 2024-12-17
Added `codegen_units` flag for setting `Rustc` resource sets

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/3008

### DIFF
--- a/rust/private/rustc_resource_set.bzl
+++ b/rust/private/rustc_resource_set.bzl
@@ -1,0 +1,296 @@
+"""Resource set definitions for Rustc actions"""
+
+def _resource_set_cpu_1(_os_name, _inputs):
+    return {"cpu": 1}
+
+def _resource_set_cpu_2(_os_name, _inputs):
+    return {"cpu": 2}
+
+def _resource_set_cpu_3(_os_name, _inputs):
+    return {"cpu": 3}
+
+def _resource_set_cpu_4(_os_name, _inputs):
+    return {"cpu": 4}
+
+def _resource_set_cpu_5(_os_name, _inputs):
+    return {"cpu": 5}
+
+def _resource_set_cpu_6(_os_name, _inputs):
+    return {"cpu": 6}
+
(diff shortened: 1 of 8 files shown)


rust/settings/BUILD.bazel
```diff
@@ -55,6 +60,16 @@ rust_lto_flag(
     visibility = ["//visibility:public"],
 )
 
+# The default value for `--codegen-units` which also affects resource
+# allocation for rustc actions. Note that any value 0 or less will prevent
+# this flag from being passed by Bazel.
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+int_flag(
+    name = "codegen_units",
+    build_setting_default = 16,
```
illicitonion, line 69: I feel like we should default this to -1 or 1... Every codebase I've ever worked in has set codegen_units to 1 (at least for release builds, but generally by default) because it leads to better runtime optimisation, but -1 ("do what rustc fancies") seems like a reasonable default...

UebelAndre: Updated! Default is now `-1` which will prevent this flag from explicitly being passed and disable any `ctx.actions.run.resource_set` from being passed.

illicitonion (approved):

rust/private/rustc_resource_set.bzl
```diff
@@ -0,0 +1,296 @@
...
+def _resource_set_cpu_62(_os_name, _inputs):
+    return {"cpu": 62}
+
+def _resource_set_cpu_63(_os_name, _inputs):
+    return {"cpu": 63}
+
+def _resource_set_cpu_64(_os_name, _inputs):
+    return {"cpu": 64}
+
+_RESOURCE_SETS = {
```
UebelAndre, line 195: Hopefully all this can be deleted once https://github.com/bazelbuild/bazel/issues/15187 is addressed in some form.

Issue #3115, opened 2024-12-17
Remove use of `_cc_toolchain`


katre:
Pre-toolchains Bazel recommended that rules use an implicit attribute named `_cc_toolchain`, pointed to the target `@bazel_tools//tools/cpp:current_cc_toolchain`, to get the CC toolchain.

This is not needed with toolchain resolution: instead rules should declare they use the CC toolchain type.

Currently, the `rust_toolchain` rule is [using this implicit dependency directly](https://github.com/bazelbuild/rules_rust/blob/7a1f0373f3c763e7d0b8a5c306641bc6a01603f0/rust/toolchain.bzl#L663): this needs to be re-written.

PR #3116, opened 2024-12-17
Remove the unused implicit attribute `_cc_toolchains`.

katre:
CC toolchains are gathered from the line:
```
toolchains = use_cc_toolchains(),
```

Keep toolchain usage of _cc_toolchain with a TODO to remove it.

See #3115 for details.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -874,6 +874,7 @@ rust_toolchain = rule(
                 "For more details see: https://docs.bazel.build/versions/master/skylark/rules.html#configurations"
             ),
         ),
+        # TODO: #3115 - Remove this attribute.
         "_cc_toolchain": attr.label(
             default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
         ),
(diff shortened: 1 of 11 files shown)


rust/toolchain.bzl
```diff
@@ -874,6 +874,7 @@ rust_toolchain = rule(
                 "For more details see: https://docs.bazel.build/versions/master/skylark/rules.html#configurations"
             ),
         ),
+        # TODO: #3115 - Remove this attribute.
```
illicitonion, line 877: Why aren't we removing this one too? When should we / what's blocking doing so?

illicitonion (approved):
Thanks!

rust/toolchain.bzl
illicitonion: AFAICT `toolchain.crosstool_files` is unused in our codebase, so we should be able to just drop it.

katre:
Can this be merged as-is, and I'll send another PR to try removing `toolchain.crosstool_files`?

illicitonion:
Works for me! Thanks!

PR #3118, opened 2024-12-17
Remove crosstool files

### LINKED ISSUE #3115: Remove use of `_cc_toolchain`

katre:
Remove `toolchain.crosstool_files and _cc_toolchains`.
Fixes #3115.

### DIFF
--- a/rust/toolchain.bzl
+++ b/rust/toolchain.bzl
@@ -663,7 +663,6 @@ def _rust_toolchain_impl(ctx):
         clippy_driver = sysroot.clippy,
         cargo_clippy = sysroot.cargo_clippy,
         compilation_mode_opts = compilation_mode_opts,
-        crosstool_files = ctx.files._cc_toolchain,
         default_edition = ctx.attr.default_edition,
         dylib_ext = ctx.attr.dylib_ext,
         env = ctx.attr.env,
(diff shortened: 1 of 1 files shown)


katre:
Does this need re-approval? I don't actually understand how github's merge queue works.

UebelAndre:
> Does this need re-approval? I don't actually understand how github's merge queue works.

The merge queue will rebase and run CI again. However, BazelCI is flaky so pull-requests end up not getting merged. Very frustrating 😞 
- https://github.com/bazelbuild/continuous-integration/issues/2123
- https://github.com/bazelbuild/continuous-integration/issues/2122

I'll queue it again.

UebelAndre:
Oh, but in this case, there was a conflict so the PR became ineligible for the queue.

PR #3119, opened 2024-12-17
Fix typos with calculating flags for lto setting

havasd:
Fixed typos in #3104

Default behavior was incorrect as the `unspecified` should have no-op on the compile flags. However this was not the case due to incorrect conditions.

#3104 could break users who are already having special LTO flags but with custom toolchains.

I will have some additional fixes but this seemed more urgent to me to keep the old behavior intact before 0.56.0

cc: @ParkMyCar , @UebelAndre

### DIFF
--- a/rust/private/lto.bzl
+++ b/rust/private/lto.bzl
@@ -101,10 +101,10 @@ def construct_lto_arguments(ctx, toolchain, crate_info):
     if mode in ["thin", "fat", "off"] and not is_exec_configuration(ctx):
         args.append("lto={}".format(mode))
 
-    if format in ["unspecified", "object_and_bitcode"]:
+    if mode == "unspecified" or format == "object_and_bitcode":
         # Embedding LLVM bitcode in object files is `rustc's` default.
         args.extend([])
-    elif format in ["off", "only_object"]:
+    elif mode == "off" or format == "only_object":
         args.extend(["embed-bitcode=no"])
     elif format == "only_bitcode":
         args.extend(["linker-plugin-lto"])
--- a/test/unit/lto/lto_test_suite.bzl
+++ b/test/unit/lto/lto_test_suite.bzl
@@ -39,7 +39,7 @@ def _lto_test_impl(ctx, lto_setting, embed_bitcode, linker_plugin):
     return analysistest.end(env)
 
 def _lto_level_default(ctx):
-    return _lto_test_impl(ctx, None, "no", False)
+    return _lto_test_impl(ctx, None, None, False)
 
 _lto_level_default_test = analysistest.make(
     _lto_level_default,


illicitonion (approved):
Thanks!

ParkMyCar:
Yeah shoot, sorry I should have foreseen this. 

IMO having "unspecified" map to object_only is the better experience because it will result in faster compile times by default. What do you think about adding a new mode of "manual" or "skip" that totally leaves these flags alone?

UebelAndre (changes requested):
Waiting for the response from @havasd to https://github.com/bazelbuild/rules_rust/pull/3119#issuecomment-2549895697

havasd:
> IMO having "unspecified" map to object_only is the better experience because it will result in faster compile times by default. What do you think about adding a new mode of "manual" or "skip" that totally leaves these flags alone?

As this LTO settings and these flags are a total can of worms and I consider it a strong breaking change. In our project we experimented with them a lot and there are lots of misleading elements in the documentation.

For example `embed-bitcode=no` doesn't work on iOS. According to [this](https://doc.rust-lang.org/rustc/codegen-options/index.html) it is required there. Also if somebody is using explicit LTO rustc flags it will break their builds. `If combined with -C lto, -C embed-bitcode=no will cause rustc to abort at start-up, because the combination is invalid.`

Also based on this code cargo decides does some calculations based on all [crates](https://github.com/rust-lang/cargo/blob/b9026bf654d7fac283465e58b8b76742244ef07d/src/cargo/core/compiler/lto.rs#L45).

I would think that even if we want to default `embed-bitcode=no` it should happen only on non-opt builds or when the lto flag is not present at all in the rustc args.

ParkMyCar:
> For example embed-bitcode=no doesn't work on iOS. According to [this](https://doc.rust-lang.org/rustc/codegen-options/index.html) it is required there. Also if somebody is using explicit LTO rustc flags it will break their builds. If combined with -C lto, -C embed-bitcode=no will cause rustc to abort at start-up, because the combination is invalid.

Good call on the iOS builds, that is something I missed in the original impl! Although IMO I would consider building Rust code for iOS a bit of an edge case? I could be wrong though! Regardless, folks building for iOS could set `lto=manual`?

> Also based on this code cargo decides does some calculations based on all [crates](https://github.com/rust-lang/cargo/blob/b9026bf654d7fac283465e58b8b76742244ef07d/src/cargo/core/compiler/lto.rs#L45).

FWIW the implementation in https://github.com/bazelbuild/rules_rust/pull/3104 is specifically based on that exact calculation from Cargo. AFAICT what that code tries to do is determine how the resulting library should be generated, and in the case a single crate shows up multiple times in the build graph it merges the determinations together. The implementation that is currently merged in `rules_rust` does not merge the determinations together like Cargo does, mainly because I wasn't sure if it was possible in Bazel. This doesn't effect correctness though, it might cause multiple builds of a single crate if that crate is being linked against multiple types of targets in a single build.

> If we start to mix it with linker-plugin-lto that brings in another set of problems because then that flag essentially requires you to have compatible linker with the bundled LLVM in rustc.

I might be missing some nuance but I don't think this is much of an issue, at least not in the current implementation? `-Clinker-plugin-lto` is only specified for `rlib`s in an LTO context, which is what [Cargo also does](https://github.com/rust-lang/cargo/blob/b9026bf654d7fac283465e58b8b76742244ef07d/src/cargo/core/compiler/lto.rs#L129-L130). In the general case `rustc` can read it's own LLVM bitcode, but if you're using a separate linker, I think it would already need to be compatible with `rustc`'s bundled version of LLVM, regardless of whether or not `linker-plugin-lto` is specified on an intermediate `rlib`?

> I am planning to have additional changes as we would like to use this feature but it may require some adjustments.

This is great! I'm always looking for ways to make builds shorter and run times faster!

Personally I am still in favor of introducing a "manual" option (I'll whip up a PR) and keeping the existing impl as-is, for two reasons:
1. It results in faster builds and follows recommendations from the Rust docs
2. The original change was released in `0.56.0` which is a semver breaking change, so it still follows stability guidelines.

All that being said, I'm 100% happy to merge this PR if the majority of folks think so, and/or follow the recommendations of @illicitonion and @UebelAndre, I'm just a contributor here :)

ParkMyCar:
Here's a PR that implements the "manual" option if we need it, https://github.com/bazelbuild/rules_rust/pull/3120, no worries if we decide to go with this approach!

havasd:
Closing as the #3120 is intending to fix the same problem

PR #2476, opened 2024-02-12
Generate package license metadata

nickcollier:
#2475

PR #3128, opened 2024-12-20
Allow rules_license_metadata to be used with bzlmod

AmeliasCode:
Support was initially added here: https://github.com/bazelbuild/rules_rust/pull/2476
but there was no support for this feature when using bzlmod. I've gone ahead and added this to both "from_cargo" and "from_spec" routes vis-a-vis bzlmod, and added a test to make sure everything is as expected.


UebelAndre (changes requested):
Does https://github.com/bazelbuild/rules_rust/pull/3136 satisfy this request?

AmeliasCode:
@UebelAndre Yes, it does! Thanks!

PR #3132, opened 2024-12-21
Migrated iOS examples to bzlmod

UebelAndre:


UebelAndre:
@keith I'm wondering if you can help me get the iOS examples building again 😅

keith:
very strange error on the first one, can you try upgrading rules_apple + apple_support to the latest releases?

UebelAndre:
> very strange error on the first one, can you try upgrading rules_apple + apple_support to the latest releases?

Bumped!

keith:
hrm this doesn't repro locally, although i am on a diff host OS, can you patch in a `set -x` to this script which is failing? https://github.com/bazelbuild/rules_apple/blob/710570ff74eaf744f77c8986fda8d102b07b9abf/tools/environment_plist/environment_plist.sh

keith:
i think we should watch https://github.com/bazelbuild/continuous-integration/issues/2153, i don't think this is on this repo

illicitonion (approved):

PR #3138, opened 2024-12-25
Consolidated docs for crate_universe

UebelAndre:


illicitonion (approved):
I don't super have a preview of this to check it, but I trust you!

UebelAndre:
> I don't super have a preview of this to check it, but I trust you!

It pretty much looks the same but load statements are now correct. You'll see!

PR #3147, opened 2024-12-31
Make lto toolchain options public and don't link proc-macros with LTO

### LINKED ISSUE #3143: lto cannot be used for `proc-macro` crate type without `-Zdylib-lto`
The last release, 0.56, introduced a new setting for LTO (PR https://github.com/bazelbuild/rules_rust/pull/3104).

However, this conflicts with the[ official Bzlmod LTO example ](https://github.com/bazelbuild/rules_rust/blob/c06feef5f3df992af32de004e6716574af4735e7/examples/bzlmod/compile_opt/hello_comp_opt/BUILD.bazel#L8) and throws an error 
that embed-bitcode does contain symbols for LTO. That is consistently the case on my CI. 

Okay, when I enable the new LTO setting in my .bazelrc via

`
build --@rules_rust//rust/settings:lto=thin
`

bazel build throws then an error, both locally on MacOS and on CI / Linux:

`error: lto cannot be used for `proc-macro` crate type without `-Zdylib-lto``

Indeed, I have a bunch of proc macros in my repo. 

Currently, the previous LTO config throws an error and the new LTO setting also throws an error.

How do you advice to setup LTO? 
Where do I set the Zdylib-lto flag?

havasd:
I would like to make lto setting public in toolchain so that when we are providing custom toolchains it would be easier to make configurations for LTO with `select` statements.

The logic which determines the flags based on cargo outputs.

Proc-macro crates can not be linked with lto and we should not emit bitcode either. This fixes #3143


illicitonion (approved):
Seems reasonable! I'm not completely sure how stable the provider is going to be, but we can worry about that if we need to change it...

PR #3158, opened 2025-01-05
Fix alias_rules handling in bzlmod and port alias_rule examples to bz…

### LINKED ISSUE #2832: vendor_remote_repository's alias targets doesn't work
When set `render_config(default_alias_rule = "opt")` for `vendor_remote_repository`. The generated alias rules are not working with the following error:

```
ERROR: .../BUILD.bazel:254:12: error loading package '@@crate_index//': cannot load '@@crate_index//:alias_rules.bzl': no such file and referenced by '//...'
```

A closer look showed that the generated `alias_rules.bzl` is not included in the generated `crate_index` repo.

If I understood correctly, `alias_rules.bzl` contains no dynamically generated content, and could be directly referenced from `rules_rust`.

havasd:
…lmod

I started porting some of the `crate_universe` examples as I wanted to add another one with bzlmod

Fixes #2832 and #3160


UebelAndre:
I think I might have fixed those in https://github.com/bazelbuild/rules_rust/pull/3148

havasd:
> I think I might have fixed those in #3148

It changes most of the things fundamentally, but I believe you are right.

havasd:
Fixed in #3148 3148

PR #3162, opened 2025-01-06
feat: cross language LTO

ParkMyCar:
If you're using Clang for your C toolchain, it's possible to do cross language optimizations between Rust and C, see Rust's [Linker-plugin-based LTO docs](https://doc.rust-lang.org/rustc/linker-plugin-lto.html) for more info.

This PR adds a new `@rules_rust//rust/settings:experimental_cross_language_lto` (off by default) which when building a Rust binary will set `-Clinker-plugin-lto`. If LTO is enabled for both Rust and C will then delay optimizations to the linker, resulting in cross language optimizations.

This feature is _highly_ experimental. To have it work correctly you need to be using a Clang toolchain with `lld` as the linker that is recent enough to understand the bitcode emitted by `rustc`. There is much more we can do here to make the feature more stable, but this seemed like a good starting point?

### DIFF
--- a/rust/private/lto.bzl
+++ b/rust/private/lto.bzl
@@ -44,7 +44,7 @@ rust_lto_flag = rule(
 )
 
 def _determine_lto_object_format(ctx, toolchain, crate_info):
-    """Determines if we should run LTO and what bitcode should get included in a built artifact.
+    """Determines what bitcode should get included in a built artifact.
 
     Args:
         ctx (ctx): The calling rule's context object.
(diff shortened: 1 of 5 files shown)


ParkMyCar:
Will take a look at the CI failures. FWIW I tested this locally with clang-18 and Rust 1.83 and it indeed works. When LTO is enabled for both Rust and C I see in-lining occurring across the language boundary

rust/private/lto.bzl
```diff
@@ -101,10 +127,16 @@ def construct_lto_arguments(ctx, toolchain, crate_info):
         return []
 
     format = _determine_lto_object_format(ctx, toolchain, crate_info)
+    xlang_enabled = _determine_experimental_xlang_lto(ctx, toolchain, crate_info)
     args = []
 
-    # proc-macros do not benefit from LTO, and cannot be dynamically linked with LTO.
-    if mode in ["thin", "fat", "off"] and not is_exec_configuration(ctx) and crate_info.type != "proc-macro":
```
havasd, line 107: did you revert this intentionally?

ParkMyCar: Yeah I was experimenting, but I switched this back to how it was before!

PR #3168, opened 2025-01-08
rustc: add support for split debuginfo (aka fission)

durin42:
This requires some helper logic in the process wrapper so we can move dwo files to predictable output locations, but otherwise it's mostly straightforward. This requires a bazel built after [d2c79821cd45b30eeaec2efc3bd5fbd26f37d3c2](https://github.com/bazelbuild/bazel/commit/d2c79821cd45b30eeaec2efc3bd5fbd26f37d3c2), but that's old enough it landed in bazel 8.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1592,6 +1592,18 @@ def rustc_compile_action(
         elif ctx.attr.require_explicit_unstable_features == -1:
             require_explicit_unstable_features = toolchain.require_explicit_unstable_features
 
+    use_split_debuginfo = cc_common.is_enabled(feature_configuration = feature_configuration, feature_name = "per_object_debug_info") and ctx.fragments.cpp.fission_active_for_current_compilation_mode()
+    if use_split_debuginfo:
+        rust_flags = rust_flags + [
+            "--codegen=split-debuginfo=unpacked",
+            "--codegen=debuginfo=full",
+        ]
+        fission_directory = crate_info.name + "_fission"
+        if output_hash:
+            fission_directory = fission_directory + "-" + output_hash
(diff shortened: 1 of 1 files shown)


UebelAndre:
Is there a Rust RFC for "fission"? Not heard the term in this context before and would love to learn more!

durin42:
That's actually a GNU term! https://gcc.gnu.org/wiki/DebugFission

Rust already supports split debuginfo upstream.

krasimirgg (changes requested):
please reformat according to https://buildkite.com/bazel/rules-rust-rustlang/builds/13749#01946470-6d00-4f5d-b0f6-9bf6ad335098/253-424

bjorn3:
FYI there is an accepted MCP to add native rustc support for debug fission. It hasn't been implemented yet though. https://github.com/rust-lang/compiler-team/issues/721

durin42:
This is going to go back on the shelf for a while, because we found a use case that the current implementation breaks. Sadly, a fix for that is probably going to require changes to rustc. :(

durin42:
@bjorn3 it occurs to me you may care: it turns out some workflows don't bother compiling the .dwp file, and expect the .dwo files to be sitting in the filesystem at their original path. Should I drop more info in the zulip thread for https://github.com/rust-lang/compiler-team/issues/721 or something?

durin42:
Rebased and updated with the latest version of the patch, which we've been using for a while with success.

rust/private/rustc.bzl.orig
bjorn3, line 1: Committed by accident?

durin42: Whoops, yep. Thanks!

PR #3176, opened 2025-01-09
Improve handling of compile_data with mixed sources

### LINKED ISSUE #3171: `compile_data` does not work together with generated sources
When using generated `srcs` and _non-generated_ `compile_data` in the same `rust_library` target, the non-generated `srcs` are symlinked into the out-tree for the build, but the same is not done for non-generated `compile_data`, so the files are not available for `include_str!` or whatever.

Here's a patch that reproduces this within the existing `compile_data` tests:

```patch
diff --git a/test/unit/compile_data/compile_data.rs b/test/unit/compile_data/compile_data.rs
index 26f060e0..ddc8fabd 100644
--- a/test/unit/compile_data/compile_data.rs
+++ b/test/unit/compile_data/compile_data.rs
@@ -1,6 +1,8 @@
 /// Data loaded from compile data
 pub const COMPILE_DATA: &str = include_str!("compile_data.txt");
 
+mod generated;
+
 #[cfg(test)]
 mod test {
     use super::*;
@@ -23,4 +25,9 @@ mod test {
             assert_eq!(TEST_COMPILE_DATA.trim_end(), "test compile data contents");
         }
     }
+
+    #[test]
+    fn test_generated() {
+        assert_eq!(generated::GENERATED, "generated");
+    }
 }
diff --git a/test/unit/compile_data/compile_data_test.bzl b/test/unit/compile_data/compile_data_test.bzl
index cd0eb7f2..383ea65c 100644
--- a/test/unit/compile_data/compile_data_test.bzl
+++ b/test/unit/compile_data/compile_data_test.bzl
@@ -61,9 +61,15 @@ wrapper_rule_propagates_and_joins_compile_data_test = analysistest.make(_wrapper
 compile_data_propagates_to_rust_doc_test = analysistest.make(_compile_data_propagates_to_rust_doc_test_impl)
 
 def _define_test_targets():
+    native.genrule(
+        name = "generated_src",
+        outs = ["generated.rs"],
+        cmd = """echo 'const GENERATED: &str = "generated";' > $@""",
+    )
+
     rust_library(
         name = "compile_data",
-        srcs = ["compile_data.rs"],
+        srcs = ["compile_data.rs", ":generated.rs"],
         compile_data = ["compile_data.txt"],
         edition = "2018",
     )
```

This is possible to work around by using `genrule` or similar to copy the files in `compile_data` to make it "generated", but this does not seem ideal.

martingms:
I'm not very well versed in starlark nor the rules_rust codebase, so feel free to ignore this and address the issue in a more fitting way, but this fixes #3171 and a related issue for me.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -149,7 +149,7 @@ def _rust_library_common(ctx, crate_type):
     crate_root = getattr(ctx.file, "crate_root", None)
     if not crate_root:
         crate_root = crate_root_src(ctx.attr.name, ctx.attr.crate_name, ctx.files.srcs, crate_type)
-    srcs, crate_root = transform_sources(ctx, ctx.files.srcs, crate_root)
+    srcs, compile_data, crate_root = transform_sources(ctx, ctx.files.srcs, ctx.files.compile_data, crate_root)
 
     # Determine unique hash for this rlib.
     # Note that we don't include a hash for `cdylib` and `staticlib` since they are meant to be consumed externally
@@ -202,7 +202,7 @@ def _rust_library_common(ctx, crate_type):
             rustc_env_files = ctx.files.rustc_env_files,
             is_test = False,
             data = depset(ctx.files.data),
-            compile_data = depset(ctx.files.compile_data),
+            compile_data = depset(compile_data),
             compile_data_targets = depset(ctx.attr.compile_data),
             owner = ctx.label,
         ),
@@ -233,7 +233,7 @@ def _rust_binary_impl(ctx):
     crate_root = getattr(ctx.file, "crate_root", None)
     if not crate_root:
         crate_root = crate_root_src(ctx.attr.name, ctx.attr.crate_name, ctx.files.srcs, ctx.attr.crate_type)
-    srcs, crate_root = transform_sources(ctx, ctx.files.srcs, crate_root)
+    srcs, compile_data, crate_root = transform_sources(ctx, ctx.files.srcs, ctx.files.compile_data, crate_root)
 
     providers = rustc_compile_action(
         ctx = ctx,
@@ -254,7 +254,7 @@ def _rust_binary_impl(ctx):
             rustc_env_files = ctx.files.rustc_env_files,
             is_test = False,
             data = depset(ctx.files.data),
-            compile_data = depset(ctx.files.compile_data),
+            compile_data = depset(compile_data),
             compile_data_targets = depset(ctx.attr.compile_data),
             owner = ctx.label,
         ),
@@ -330,13 +330,11 @@ def _rust_test_impl(ctx):
                 ),
             )
 
-        srcs, crate_root = transform_sources(ctx, ctx.files.srcs, getattr(ctx.file, "crate_root", None))
+        # Need to consider all src files together when transforming
+        srcs = depset(ctx.files.srcs, transitive = [crate.srcs]).to_list()
+        compile_data = depset(ctx.files.compile_data, transitive = [crate.compile_data]).to_list()
+        srcs, compile_data, crate_root = transform_sources(ctx, srcs, compile_data, getattr(ctx.file, "crate_root", None))
 
-        # Optionally join compile data
-        if crate.compile_data:
-            compile_data = depset(ctx.files.compile_data, transitive = [crate.compile_data])
-        else:
-            compile_data = depset(ctx.files.compile_data)
         if crate.compile_data_targets:
             compile_data_targets = depset(ctx.attr.compile_data, transitive = [crate.compile_data_targets])
         else:
@@ -360,7 +358,7 @@ def _rust_test_impl(ctx):
             name = crate_name,
             type = crate_type,
             root = crate.root,
-            srcs = depset(srcs, transitive = [crate.srcs]),
+            srcs = depset(srcs),
             deps = depset(deps, transitive = [crate.deps]),
             proc_macro_deps = depset(proc_macro_deps, transitive = [crate.proc_macro_deps]),
             aliases = aliases,
@@ -370,7 +368,7 @@ def _rust_test_impl(ctx):
             rustc_env = rustc_env,
             rustc_env_files = rustc_env_files,
             is_test = True,
-            compile_data = compile_data,
+            compile_data = depset(compile_data),
             compile_data_targets = compile_data_targets,
             wrapped_crate_type = crate.type,
             owner = ctx.label,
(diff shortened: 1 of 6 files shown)


rust/private/rust.bzl
```diff
@@ -330,13 +330,11 @@ def _rust_test_impl(ctx):
...
 
-        srcs, crate_root = transform_sources(ctx, ctx.files.srcs, getattr(ctx.file, "crate_root", None))
+        # Need to consider all src files together when transforming
+        srcs = depset(ctx.files.srcs, transitive = [crate.srcs]).to_list()
+        compile_data = depset(ctx.files.compile_data, transitive = [crate.compile_data]).to_list()
+        srcs, compile_data, crate_root = transform_sources(ctx, srcs, compile_data, getattr(ctx.file, "crate_root", None))
 
-        # Optionally join compile data
-        if crate.compile_data:
-            compile_data = depset(ctx.files.compile_data, transitive = [crate.compile_data])
```
sam-mccall, line 337: do we need to worry about compile_data being None?

sam-mccall (approved):
I was looking at generated sources recently, this change makes sense to me.

Needs an owner, @krasimirgg we were talking about this very issue a couple of days ago :-)

rust/private/rust.bzl
martingms: afaict this shouldn't be possible, `attr.label_list` defaults to `[]`, and we set `compile_data` explicitly to a depset based on that in all codepaths, but I can add this check back if you think it's better to be defensive. It was a bit confusing to me, which is why I removed it!

krasimirgg:
Looks good! Please re-format the file that the Buildifier workflow is complaining about.

krasimirgg (changes requested):
Please reformat according to https://buildkite.com/bazel/rules-rust-rustlang/builds/13739#01945f02-1578-4657-8c29-aeaef49853d9/127-128

martingms:
Sorry, missed buildifier and one test affected by the last commit, pushed an updated patch now :+1:

UebelAndre:
This has lead to a regression for targets which have compile data from completely other packages. E.g.
```starlark
# //current/package:lib
rust_library(
    name = "lib",
    srcs = ["lib.rs", ":generated.rs"],
    compile_data = ["//other/package:data"],
    rustc_env = {"COMPILE_DATA": "$(execpath //other/package:data)"},
)
```

I don't quite know how to solve for this as I'm not sure what the symlink path should be for this foreign path.

krasimirgg:
Happy if we rolled this back until we have a full fix for this.

martingms:
@krasimirgg The crash should be fixed by #3196 or similar, but it's unclear to me how the repro in #3193 could have worked (as in the `data.txt` file is actually available during compile) even before this (#3176) PR was merged, as generated sources were built in the out tree even before. I'm probably missing something! Do you have any pointers to what the situation is in the real world failures?

Fine with me to roll it back too obviously!

jgsogo:
This PR broke something in my project.

I'm using the [`diesel_migrations::embed_migrations!` macro](https://docs.rs/diesel_migrations/2.2.0/diesel_migrations/macro.embed_migrations.html) to collect the migrations for my database. This macro takes a relative path from the `Cargo.toml` file to the directory that contains the migrations. This works in the Cargo world, however, with Bazel, it fails after this PR. 

For a rust library inside a `<repo-root>/libraries/diesel_utils` directory, the commit that works (5e426fa427f6ce197da281233525cc486dfdf07c) and the one that fails (bb74a65a94d7239f70bc237802cafae43791b12c) populate the same `CARGO_MANIFEST_DIR='${pwd}/libraries/diesel_utils'` envvar, but the one that works contains **an additional directory with the files from the `compile_data` in the working directory**: `libraries/diesel_utils/<value-provided-to-embed_migrations>`

I wanted to report my findings here before having a closer look to the PR and maybe open a more detailed issue, or even a PR with a fix. Maybe during the weekend.

martingms:
Hi @jgsogo, do you have a minimal reproduction of your issue? It's been a while since I've looked at this, but your patch breaks the stuff that this PR was meant to fix, so I hope we can find a different solution!

jgsogo:
Here it is: https://github.com/jgsogo/rules_rust-3176

### Reproducing the error

```bash
git clone https://github.com/jgsogo/rules_rust-3176
cd rules_rust-3176

bazel build //libraries/diesel_utils:integration_tests --verbose_failures --sandbox_debug

```

Error log (uses `rules_rust v0.59.2`)

```
error: proc macro panicked
 --> bazel-out/darwin_x86_64-fastbuild/bin/libraries/diesel_utils/tests/common/mod.rs:3:5
  |
3 |     diesel_migrations::embed_migrations!("tests/common/migrations");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: Failed to receive migrations dir from Some("tests/common/migrations")

```

indeed, if you check the working directory, the `libraries/diesel_tuils/tests/common/migrations` folder is not there:

```
➜  rules_rust-3176 git:(main) ls -la /private/var/tmp/_bazel_jgsogo/6a6eb0d5365aa7efc463fc4e16e1d4d9/sandbox/darwin-sandbox/1056/execroot/_main
total 0
drwxr-xr-x   4 jgsogo  wheel  128 Mar 31 13:22 .
drwxr-xr-x   3 jgsogo  wheel   96 Mar 31 11:27 ..
drwxr-xr-x   4 jgsogo  wheel  128 Mar 31 13:30 bazel-out
drwxr-xr-x  13 jgsogo  wheel  416 Mar 31 13:38 external
```

### The issue

In the `MODULE.bazel` you can modify `rules_rust` to use different commits, right before and after this PR: https://github.com/jgsogo/rules_rust-3176/blob/c42eb235cde50cf1933666743b543eb9f4a53c1f/MODULE.bazel#L13-L25

Using the commits that works, the migrations are available in the working directory:

```
➜  rules_rust-3176 git:(main) ✗ ls -la /private/var/tmp/_bazel_jgsogo/6a6eb0d5365aa7efc463fc4e16e1d4d9/sandbox/darwin-sandbox/1251/execroot/_main
total 0
drwxr-xr-x   5 jgsogo  wheel  160 Mar 31 13:49 .
drwxr-xr-x   3 jgsogo  wheel   96 Mar 31 13:49 ..
drwxr-xr-x   4 jgsogo  wheel  128 Mar 31 13:49 bazel-out
drwxr-xr-x  13 jgsogo  wheel  416 Mar 31 13:49 external
drwxr-xr-x   3 jgsogo  wheel   96 Mar 31 13:49 libraries               <-- Here it is!
```

**The issue comes from the target `//libraries/diesel_utils:schema` that is copying a file into the build directory**. If, instead of using this generated file, I use a file in the workspace, it works. See https://github.com/jgsogo/rules_rust-3176/blob/c42eb235cde50cf1933666743b543eb9f4a53c1f/libraries/diesel_utils/BUILD.bazel#L24-L32, and switch from one implementation to the other.


---

Don't judge the way I'm generating this `schema.rs` file and how I'm building `diesel_cli` 😅 . This is a side-project and I know I need to refactor this part. Anyway, it shouldn't matter the way I'm generating `schema.rs`, **I think it's just the issue/conflict with two rules contributing files to the working directory**.

martingms:
I see, the way this library looks for the migrations does mesh well with bazel in general really, as it seems it's hard-coded to search from CARGO_MANIFEST_DIR as the root. You could try a hack by setting `CARGO_MANIFEST_DIR` with `rustc_env` somehow, but that might break other stuff if this lib is that tied to cargo..

PR #3180, opened 2025-01-10
Support no_std in Prost toolchain

fionera:


extensions/prost/private/prost.bzl
```diff
@@ -453,6 +457,10 @@ rust_prost_toolchain = rule(
             doc = "Whether to include transitive dependencies. If set to True, all transitive dependencies will directly accessible by the dependent crate.",
             default = False,
         ),
+        "is_nostd": attr.bool(
```
UebelAndre, line 460: nit, can this be renamed to `is_no_std` to match the `no_std` feature?

UebelAndre (changes requested):
Thanks! Just one minor request!

extensions/prost/private/prost.bzl
fionera: Done :smile:

extensions/prost/private/protoc_wrapper.rs
```diff
@@ -457,6 +466,9 @@ struct Args {
     /// Whether to generate tonic code.
     is_tonic: bool,
 
+    // Whether to put a no_std tag into the generated code.
+    is_nostd: bool,
```
UebelAndre, line 470:
```suggestion
    is_no_std: bool,
```

UebelAndre (changes requested):
One last nit and I think you can delete `rust_prost.md` since docs are now generated directly from the bzl source file. Devs no longer need to render them and add them to their PRs 😄

extensions/prost/private/protoc_wrapper.rs
fionera: Woops :D I updated the PR including docs

UebelAndre (approved):
Thanks!

UebelAndre (changes requested):
Actually, is there some sort of integration test you can add that uses this new flag?

fionera:
If you give me an example, I could take a look

UebelAndre:
> If you give me an example, I could take a look

Like any of the cases here: https://github.com/bazelbuild/rules_rust/tree/5226882efee4c3f1f2b68940b06fad9ad070f8b1/extensions/prost/private/tests

if there’s a single test that that uses `use_no_std` then there is some regression testing to know if a future change breaks this feature.

UebelAndre:
> Like any of the cases here: https://github.com/bazelbuild/rules_rust/tree/5226882efee4c3f1f2b68940b06fad9ad070f8b1/extensions/prost/private/tests
> 
> if there’s a single test that that uses `use_no_std` then there is some regression testing to know if a future change breaks this feature.

Let me know if you still have questions! Happy to get this merged but it'll be super helpful down the line to have at least 1 regression test.

UebelAndre (approved):

fionera:
Sorry I didn't had time yet to add the test yet, but it is on my TODO :)

UebelAndre:
I'd love to get this into the next release but just need some form of integration test for it!

PR #3188, opened 2025-01-20
Don't register rust_host_tools twice

hauserx:
The bazel-lsp uses `rust_host_tools` to perform some actions by using:

```starlark
rust_host_tools = use_extension("@rules_rust//rust:extensions.bzl", "rust_host_tools") 
use_repo(rust_host_tools, "rust_host_tools")
```

See: https://github.com/cameron-martin/bazel-lsp/blob/19b799e528c884be6efeb108ed1840a4edff1330/MODULE.bazel#L14

It started failing when using rules_rust from HEAD after https://github.com/bazelbuild/rules_rust/pull/3148 because rules_rust's MODULE.bazel registers toolchain with the same name:

```starlark
rust_host_tools.host_tools(
    name = "rust_host_tools",
)
```

See: https://github.com/bazelbuild/rules_rust/blob/aae84e97c73eae2e6654e1a7e1b751d0c1f2ac9e/MODULE.bazel#L70

### DIFF
--- a/rust/extensions.bzl
+++ b/rust/extensions.bzl
@@ -265,18 +265,6 @@ def _rust_host_tools_impl(module_ctx):
                 **attrs
             )
 
-        # If no tags were specified, create a default repository.
-        if not mod.tags.host_tools:
-            attrs = {
-                "name": "rust_host_tools",
-                "version": rust_common.default_version,
-            }
-            rust_toolchain_tools_repository(
-                exec_triple = host_triple.str,
-                target_triple = host_triple.str,
-                **attrs
-            )
-
     metadata_kwargs = {}
     if bazel_features.external_deps.extension_metadata_has_reproducible:
         metadata_kwargs["reproducible"] = True


hauserx:
FYI @cameron-martin (in case what bazel-lsp is doing is not correct).

rust/extensions.bzl
```diff
@@ -265,18 +265,6 @@ def _rust_host_tools_impl(module_ctx):
                 **attrs
             )
 
-        # If no tags were specified, create a default repository.
-        if not mod.tags.host_tools:
```
UebelAndre, line 269: Should this instead be called if no module defines `host_tools`?

UebelAndre (changes requested):

rust/extensions.bzl
hauserx: rules_rust's MODULE.bazel defines host tools already, so it will be always defined. Or I am missing something and it may not be defined in some other way?

UebelAndre: It just seems brittle to me for folks to be depending on the name of a particular dependency in rules_rust's `MODULE.bazel` file. That's why I tried to have the module extension itself always provide the repository and encode the consistent name.

cameron-martin:
> FYI @cameron-martin (in case what bazel-lsp is doing is not correct).

We use `rust_host_tools` to invoke cargo to read the version in the Cargo.toml. Not sure if this is the best way of doing it, but I don't know of a better way.

UebelAndre (approved):

rust/extensions.bzl
hauserx: Just for reference, without this change currently one gets following error if one does not specify host_tools tag:
```
ERROR: (...)/external/rules_rust+/rust/extensions.bzl:262:44: Traceback (most recent call last):
        File "(...)/external/rules_rust+/rust/extensions.bzl", line 262, column 44, in _rust_host_tools_impl
                rust_toolchain_tools_repository(
Error in repository_rule: A repo named rust_host_tools is already generated by this module extension at (...)/external/rules_rust+/rust/extensions.bzl:274:44
ERROR: Analysis of target '//:generate_rustc_env_file' failed; build aborted: error evaluating module extension @@rules_rust+//rust:extensions.bzl%rust_host_tools
```

I guess for bazel-lsp workaround would be simply to specify some host_tool with unique name (like `bazel_lsp_rust_host_tools`). Other solution would be in rules_rust's MODULE to change name to smthg like `rules_rust_internal_host_tools`, which will require users to use some name, or roll back this change, but move this condition outside of loop as you mentioned.

Wonder whether the best solution could be to expose those toolchain files directly in @rules_rust, like `@rules_rust//bin:cargo`, something similar as you can do with go by running `bazel run @io_bazel_rules_go//go`. This way it would use binaries for configured toolchain.

PR #3192, opened 2025-01-21
Use `--remote_download_all` when building `rust_analyzer_crate_spec`s

### LINKED ISSUE #3170: `//tools/rust_analyzer:gen_rust_project` does not build all required targets
When using bazel's `--disk_cache` (and possibly in other remote cache situations?) `gen_rust_project` can generate `rust-project.json` files that contains paths in `"proc_macro_dylib_path"` that are invalid at the time of generation.

One easy to reproduce example as long as cargo crates with proc-macros are used in the project: Enable `--disk_cache`, run a full build to populate the cache, `bazel clean`, run a new full build for good measure, run `//tools/rust_analyzer:gen_rust_project`. 
`"proc_macro_dylib_path"` for e.g. `clap_derive` will now be something like. `/home/<...>/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_rust++crate+crates__clap_derive-4.5.18/libclap_derive-2275298359.so` which has not been rebuilt and therefore does not exist.

I've not found a way to force it to be built manually either in this case, where I depend on the `clap_derive` crate transitively, so it's not available at `@crates//:clap_derive` or similar.

martingms:
This fixes #3170, by ensuring that transitive deps like procmacro dylibs are fetched from the disk-cache.

This works in my case, but it's unclear to me if this has unintended consequences for people running remote builds etc, and this can easily be addressed by users manually by adding `build --remote_download_all` to their .bazelrc too if you prefer not merging this.

### DIFF
--- a/tools/rust_analyzer/lib.rs
+++ b/tools/rust_analyzer/lib.rs
@@ -23,6 +23,7 @@ pub fn generate_crate_info(
         .env_remove("BUILD_WORKSPACE_DIRECTORY")
         .arg("build")
         .arg("--norun_validations")
+        .arg("--remote_download_all")
         .arg(format!(
             "--aspects={}//rust:defs.bzl%rust_analyzer_aspect",
             rules_rust.as_ref()


UebelAndre (approved):
Thanks!

PR #3193, opened 2025-01-21
Add regression test for compile data in alternate packages

UebelAndre:


martingms:
Ah, right. The `relativize` in `utils.bzl` actually does the "right thing", so if we `s/paths.relativize/relativize/` on `utils.bzl:870` you get the "right" path, but then we get `Error in declare_file: the output artifact 'test/unit/compile_data/data/data.txt' is not under package directory 'test/unit/compile_data/src' for target '//test/unit/compile_data/src:lib'` so I guess symlinking in those cases is not the right approach anyway. I'll look some more.

martingms:
#3196 at least makes this build, but it's not enough I guess since the data.txt-file must also be symlinked (to the correct place) to actually be able to use it.

UebelAndre:
> how can this have worked before, when generated sources were still symlinked and built in the out tree?

I updated the regression test to show how the data is consumed. This is exactly what how one of my targets was setup. I tend to not rely on relative paths for `include_str`.

martingms:
Cool, glad the patch worked :+1:

illicitonion (approved):

PR #3196, opened 2025-01-22
Only symlink package files

martingms:
Don't attempt to symlink files that are not in the package. This won't work anyway as `ctx.declare_file` doesn't allow declarations outside the package_root, and fixes issue where we attempted to symlink compile_data file from different package.

This fixes the specific issue in #3193, but unclear to me whether it's the _right_ fix or whether it fixes the real-world issue people have.


martingms:
This patch was included in #3193, so closing this.

PR #3174, opened 2025-01-09
Bump wasm-bindgen dependencies

UebelAndre:

PR #3209, opened 2025-01-27
wasm-bindgen 0.2.100

froody:
Update wasm-bindgen to 0.2.100 as I was seeing errors building with 0.2.92


UebelAndre:
https://github.com/bazelbuild/rules_rust/pull/3174 Already exists but some tests are failing. Help would be appreciated!

PR #3210, opened 2025-01-27
Crate universe compile_data_glob_excludes

yoks:
Right now there is no way to exclude specific paths from compile_data. And in a very rare cases this can be required, such example could be what crate have generated filenames which incompatible with bazel (in my case it was  `:` symbol), usualy these files are used for test and not required for compile.

I added new field to `annotation` called `compile_data_glob_excludes`, it allows specify excludes on top of default ones, following same logic as `compile_data_glob`.


crate_universe/src/rendering.rs
```diff
@@ -492,8 +492,6 @@ impl Renderer {
     ) -> Result<CargoBuildScript> {
         let attrs = krate.build_script_attrs.as_ref();
 
-        const COMPILE_DATA_GLOB_EXCLUDES: &[&str] = &["**/*.rs"];
```
illicitonion, line 495: It looks like we're dropping a default value - is that intentional?

illicitonion (commented):
Aah sorry for missing this - it looks useful, I left one comment otherwise if you can rebase let's merge this! Thank you!

crate_universe/src/rendering.rs
yoks: Yes, but i do not remember why, at first i left it as default, i will recheck. Maybe i missed something.

yoks: I think i found out, it was my mistake, at first i added default to both `CommonAttributes` and `BuildScriptAttributes` this was unintended for `CommonAttributes`, so it broke the build. And i missed what i only neded to add default to `BuildScriptAttributes` (as it only used here). 
Added new commit which adds proper default here.

illicitonion (approved):
Perfect - thanks!

UebelAndre:
@yoks sorry I introduced another conflict for you!

yoks:
@UebelAndre updated lockfiles, interesting thing is lockfiles now include excludes as they are serialized in JSON, i think this is correct behavior, but please verify.

PR #3211, opened 2025-01-27
Support new `cargo::` prefixes in build script outputs.

joell:
In Rust 1.77.0, the preferred build directive prefix for Cargo commands was changed from `cargo:` to `cargo::`.  (The single-colon prefix is still supported, but is no longer the default in documentation.)

With this commit, the `cargo_build_script_runner` tool now supports both prefixes.

This fixes issue #2980.

### DIFF
--- a/cargo/cargo_build_script_runner/lib.rs
+++ b/cargo/cargo_build_script_runner/lib.rs
@@ -29,19 +29,19 @@ pub struct CompileAndLinkFlags {
 /// Enum containing all the considered return value from the script
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum BuildScriptOutput {
-    /// cargo:rustc-link-lib
+    /// cargo::rustc-link-lib
     LinkLib(String),
-    /// cargo:rustc-link-search
+    /// cargo::rustc-link-search
     LinkSearch(String),
-    /// cargo:rustc-cfg
+    /// cargo::rustc-cfg
     Cfg(String),
-    /// cargo:rustc-flags
+    /// cargo::rustc-flags
     Flags(String),
-    /// cargo:rustc-link-arg
+    /// cargo::rustc-link-arg
     LinkArg(String),
-    /// cargo:rustc-env
+    /// cargo::rustc-env
     Env(String),
-    /// cargo:VAR=VALUE
+    /// cargo::VAR=VALUE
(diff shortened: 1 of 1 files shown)


joell:
@UebelAndre: Here is your PR!  I've tested it against a project using [k8s-openapi](https://crates.io/crates/k8s-openapi) by patching rules_rust 0.57.1 via `single_version_override` in `MODULES.bazel`, and I have confirmed the fix works and k8s-openapi now builds.

So this fixes issue #2980 in my test setup.

Let me know if it needs any further changes to merge.

cargo/cargo_build_script_runner/lib.rs
```diff
@@ -230,24 +235,27 @@ mod tests {
...
-cargo:rerun-if-changed=ignored
-cargo:rustc-cfg=feature=awesome
-cargo:version=123
-cargo:version_number=1010107f
-cargo:include_path=/some/absolute/path/include
-cargo:rustc-env=SOME_PATH=/some/absolute/path/beep
-cargo:rustc-link-arg=-weak_framework
-cargo:rustc-link-arg=Metal
-cargo:rustc-env=no_trailing_newline=true",
+cargo::rustc-link-lib=sdfsdf
```
UebelAndre, line 238: Can you actually keep some test data where the single `:` case is checked? Lots of crates still use that so it's important that it still works (at least until Cargo decides to fully deprecate it).

UebelAndre (changes requested):
Thanks! Looks good! I just had one request for additional regression testing 😄

cargo/cargo_build_script_runner/lib.rs
joell: Yes, a test for the single-colon case was already included in my commit; check the bottom of the paragraph for the `cargo:rustc-env=old_cargo_colon_prefix_works=true` on line 254.

I also added two new tests for non-cargo-instruction lines; I saw the logic checking for them, but there was no test I found covering it, so I added some.

UebelAndre (approved):
Thanks!

cargo/cargo_build_script_runner/lib.rs
UebelAndre: Ah, I missed that. Sweet!

PR #3215, opened 2025-01-28
Fix duplicate aliases generated by cargo-bazel when metadata contains…

### LINKED ISSUE #3213: [57.1] Duplicate aliases can occur on some crates
After the rewrite of the alias generation logic in [this change](https://github.com/bazelbuild/rules_rust/commit/5e904830eee39b37775855c6ff7e7e1079eb59e0#diff-52bdc04b030a02035a19a3c805288b7720191185a18f6b8f16f61a59baa42d3fR224).

It is possible that in certain cases we will have duplicate alises generated which results in an error. One such crate is `new_debug_unreachable`.
The main reason duplicates will be introduced is because for some reason `cargo metadata` adds an identical `rename` to the crate.
I couldn't reproduce this when I was trying to get metadata from pure cargo based project.

This problem occurred to us when it was used as a direct dependency. When it was pulled in through transitive dependency it didn't cause such issue.

# Possible fix

Check for identical rename/crate name in the linked code.

# Repro

## MODULE.bazel
```
module(name = "test", version = "0.0.1")

bazel_dep(name = "rules_rust", version = "0.57.1")

rust.toolchain(
  edition = "2021",
  versions = ["1.83.0"],
)
use_repo(rust, "rust_toolchains")
crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
crate.spec(package = "new_debug_unreachable", version = "=1.0.6", features = [], default_features = False)
crate.from_specs()
use_repo(crate, "crates")
```

have a dummy crate with an empty `main.rs`

## BUILD.bazel
```

rust_binary(
  name = "crate_b",
  crate_root = "src/main.rs",
  deps = ['@crates//:new_debug_unreachable'],
)
```

havasd:
… identical rename

This fixes #3213.


crate_universe/src/rendering.rs
```diff
@@ -1947,4 +1950,48 @@ mod test {
...
+                override_targets: BTreeMap::default(),
+            },
+        );
+
+        let mut render_config = mock_render_config(None);
+        Arc::get_mut(&mut render_config)
+            .unwrap()
+            .generate_rules_license_metadata = true;
+        let renderer = Renderer::new(render_config, mock_supported_platform_triples());
+        assert!(renderer.render(&context, None).is_ok());
```
UebelAndre, line 1995: This check doesn't feel all that descriptive, do I understand this correctly that you want to ensure there is only one alias added? Or no aliases added? Ideally we would have some regex check but I've historically been lazy when it comes to the render checks 😞 But my main concern is I'll forget what this test is for. Asserting on something more specific would help a ton!

UebelAndre (changes requested):
Thanks! I had just one request for this!

crate_universe/src/rendering.rs
havasd: I have added explicit check for the output. My main interest for this test was that it should not fail during rendering but it is better now.

Also I think we may lack some of the workspace members based unit tests

UebelAndre (approved):
Thanks!

PR #3223, opened 2025-01-30
fix: add NIX_LDs vars to work on NixOS

tkr-sh:
No bazel command will work on NixOS when using crate_universe because `NIX_LD` is removed by `.env_clear()`.
See <https://github.com/nix-community/nix-ld>
This fix is fairly simple, and makes the command work on NixOS.


UebelAndre:
replaced by https://github.com/bazelbuild/rules_rust/pull/3304

PR #3228, opened 2025-01-31
Provide an option to opt-out of compile_well_known_types

### LINKED ISSUE #2058: What's a good default for protoc-gen-{prost|tonic}'s compile_well_known_types?
The `compile_well_known_types` flag controls whether generated libraries use prost-types or types from generated `google.protobuf.*` protos.

This flag is currently always enabled here (and a bit further below for tonic): https://github.com/bazelbuild/rules_rust/blob/478213bb114b6c2649f1c2533589ac3962d3bf28/proto/prost/private/prost.bzl#L63

Being enabled by default means that we have to use e.g. `timestamp_proto::google::protobuf::Timestamp` instead of `prost_types::Timestamp`.

I'm not sure whether this is a good or a bad thing.

- It's more precise if users explicitly use the types that are actually written in the proto files.
- It seems the ergonomics of prost-types are nicer. E.g. a generated `timestamp_proto::google::protobuf::Timestamp` doesn't implement `From<SystemTime>`, but  `prost_types::Timestamp` [does](https://docs.rs/prost-types/latest/prost_types/struct.Timestamp.html#impl-From%3CSystemTime%3E-for-Timestamp). Intuitively I could also imagine the specialized `prost-types` to be more performant.

There is also a protoc-wkt crate that adds more extensive chrono-timestamp interop and other goodies.

Adding configurability for `compile_well_known_types` seems straightforward, but I'm not sure what the better default would be.

cc @allada @UebelAndre @freeformstu @illicitonion

alexkirsz:
This PR adds a `compile_well_known_types` flag to the Prost toolchain declaration.

By default, for backwards compatibility reasons, this flag is set to `True`, as this is the current behavior. If this flag is switched to `False`, the `compile_well_known_types` option is no longer passed to Prost and Tonic, and the protoc wrapper will ignore types from the `google.protobuf` package when computing external types, relying instead on [prost's own externals handling](https://github.com/tokio-rs/prost/blob/e617832ab432adc7dd799321eb216c82a589d421/prost-build/src/extern_paths.rs#L35-L62). Furthermore, the configured Prost types crate will be added as a runtime dependency to all `rust_prost_library` targets.

We're currently migrating away from our own set of proto rules to `rust_prost_library`, but our existing code relies heavily on the `prost-types` crate's implementations of well-known-types, which is the recommended way of using prost. However, the current behavior of `rules_rust_prost` makes it impossible to wire up properly, hence this PR.

Fixes #2058


alexkirsz:
@UebelAndre Would love to get a review on this 🙂

UebelAndre (changes requested):
Seems reasonable to me! But can you add some usage of this in the tests directory somewhere so it's regression tested?

alexkirsz:
@UebelAndre I added tests, which are basically the WKT tests but adapted to use prost_types.

In order to allow configuring the `compile_well_known_types` flag of the toolchain, I added a `bool_flag`/`config_setting` combo, and I'm using a transition to set it to `False` _only_ for the nocompile_wkt test.

I had originally built the test as its own Bazel workspace with its own toolchain, but this made the test setup much more complicated, and also brought more code into the repo.

However, for the transition to work properly, I had to expose the `CrateGroupInfo` provider from `@rules_rust//rust:rust_common.bzl`. This is also something we need in our project in order to build transitions over `rust_prost_library`, so I think it makes sense to expose it.

UebelAndre (approved):
Perfect! Thank you!

extensions/prost/private/prost.bzl
```diff
@@ -233,7 +234,10 @@ def _rust_prost_aspect_impl(target, ctx):
 
     rustfmt_toolchain = ctx.toolchains["@rules_rust//rust/rustfmt:toolchain_type"]
     prost_toolchain = ctx.toolchains[TOOLCHAIN_TYPE]
-    for prost_runtime in [prost_toolchain.prost_runtime, prost_toolchain.tonic_runtime]:
+    runtimes = [prost_toolchain.prost_runtime, prost_toolchain.tonic_runtime]
+    if not prost_toolchain.compile_well_known_types:
+        runtimes.append(prost_toolchain.prost_types)
```
UebelAndre, line 239: Why explicitly add `prost_types` in this case?

UebelAndre (changes requested):
Sorry I got to this so late, looks like there's a conflict. When I went to resolve it though I had one question which I would love your insight on!

extensions/prost/private/prost.bzl
alexkirsz: The code-generated library will now depend on `prost-types`, so I'm adding the dependency here. Since the user will already have provided a `prost-types` dependency as an argument to the toolchain, I felt it more consistent to automatically add it to the runtimes for them, instead of asking them to provide it again in the runtimes themselves.

alexkirsz:
Hey @UebelAndre, I see you recently merged `main` into the branch, is there anything else I can help with?

Issue #3227, opened 2025-01-31
Build dependencies not added to the rendered BUILD.bazel


bakjos:
When using build dependencies inside a Cargo.toml in a workspace, the rendered BUILD.bazel in crates_index does not include the build dependencies

I have the following build dependencies

```toml
[build-dependencies]
xmltree = "0.6"
phf_codegen = "0.7.24"
```

These 2 dependencies are not added to the resulting bazel file, generating the error `no such target '@@crate_index//:phf_codegen-0.7.24': target 'phf_codegen-0.7.24' not declared in package`

It was working well in 0.56.0 and started failing in 0.57.0

After debugging I realized the version 0.5.6 keeps the build_deps with full names `"phf_codegen": Label("@crate_index__phf_codegen-0.7.24//:phf_codegen")` however the new versions include them as `"phf_codegen": Label("//:phf_codegen-0.7.24"),` but the alias is not created in the main BUILD.bazel file. The change was introduced here https://github.com/bazelbuild/rules_rust/commit/5e904830eee39b37775855c6ff7e7e1079eb59e0#diff-a165d4fcf7bc187750bb12572611029ff1522d224f1b4cc96483e6184e254c52

PR #3229, opened 2025-01-31
feat: Create aliases for build dependencies

bakjos:
This PR tries to fix the issue https://github.com/bazelbuild/rules_rust/issues/3227 by adding the build dependencies as workspace dependency


HackAttack:
This patch worked for me 👍

UebelAndre (changes requested):
Thanks! Could you add a regression test for this?

Also, what is the impact to lockfiles with this change? Does this only impact modules that have `[build-dependencies]` in a `Cargo.toml` file?

mkosiba:
This PR fixes an issue I have, so I'd like to help it get merged. I have written some tests here https://github.com/mkosiba/rules_rust/commit/e890383bccd83defaff9b34f98ac323b910fe18a

@bakjos feel free to pull that commit into this PR.

@UebelAndre I'm not very familiar with the `rules_rust` implementation but to me it seems like this change only affects the rendered `BUILD.bazel` file and not the lock file.
Also please let me know if the tests I added above are sufficient (and the right kind of tests).

bakjos:
Thanks @mkosiba  I Merged your changes into this PR


> This PR fixes an issue I have, so I'd like to help it get merged. I have written some tests here [mkosiba@e890383](https://github.com/mkosiba/rules_rust/commit/e890383bccd83defaff9b34f98ac323b910fe18a)
> 
> @bakjos feel free to pull that commit into this PR.
> 
> @UebelAndre I'm not very familiar with the `rules_rust` implementation but to me it seems like this change only affects the rendered `BUILD.bazel` file and not the lock file. Also please let me know if the tests I added above are sufficient (and the right kind of tests).

UebelAndre (approved):
Thanks!

PR #3234, opened 2025-02-03
Generate crate_universe bzlmod repositories in parallel

UebelAndre:
This change aims to improve crate_universe for bzlmod by parallelizing generation into a single [module_ctx.execute](https://bazel.build/rules/lib/builtins/module_ctx#execute) call to the `cargo-bazel` binary for each module. This avoids the limitation of running each module serially.


UebelAndre:
It's interesting that this is roughly as slow as the serial implementation. I suspect the bottleneck is the `cargo tree` calls which I believe puts a lock on the local registry being used.

## Before

signal-2025-02-04-082246_002

## After

signal-2025-02-04-082311_002

UebelAndre:
> It's interesting that this is roughly as slow as the serial implementation. I suspect the bottleneck is the `cargo tree` calls which I believe puts a lock on the local registry being used.

I should point out that this was for a run without `REPIN=1`. I also misspoke about `cargo tree` calls, unless splicing is happening (which occurs with `REPIN=1`), `cargo tree` should not be called. When all crates are repinning we do see a larger time savings but not as much as I'd expected:

## Before

Screenshot 2025-02-04 at 10 27 10 AM

## After

Screenshot 2025-02-04 at 10 26 59 AM

PR #3235, opened 2025-02-03
fix: Use "exec" for cargo build scripts

mrmeku:
The nuance between when to use exec vs target
is a bit tricky to get right sometimes. You want
to use "exec" for executables that are used at
build time and "target" for runtime dependencies

Since cargo build script are run at build time,
they need to be set to exec.

This fixes circumstances where a user is using
RBE on a platform different than their host machine

For example, I found this bug when I enabled my
linux based RBE cluster when trying to do a
bazel run of a rust binary on my mac. When the
build script executed it threw up because the RBE
cluster was using the mac version of the script.

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -625,7 +625,7 @@ cargo_build_script = rule(
             doc = "The binary script to run, generally a `rust_binary` target.",
             executable = True,
             mandatory = True,
-            cfg = "target",
+            cfg = "exec",
             providers = [CargoBuildScriptRunfilesInfo],
         ),
         "tools": attr.label_list(
--- a/test/cargo_build_script/location_expansion/test.rs
+++ b/test/cargo_build_script/location_expansion/test.rs
@@ -63,22 +63,17 @@ pub fn test_execpath() {
         )
         .1;
 
-    let (data_cfg, data_short_path) = data_path.split_at(
+    let (_data_cfg, data_short_path) = data_path.split_at(
         data_path
             .find("/bin/")
             .unwrap_or_else(|| panic!("Failed to find bin in {}", data_path))
             + "/bin/".len(),
     );
-    let (tool_cfg, tool_short_path) = tool_path.split_at(
+    let (_tool_cfg, tool_short_path) = tool_path.split_at(
         tool_path
             .find("/bin/")
             .unwrap_or_else(|| panic!("Failed to find bin in {}", tool_path))
             + "/bin/".len(),
-    );
-
-    assert_ne!(
-        data_cfg, tool_cfg,
-        "Data and tools should not be from the same configuration."
     );
 
     assert_eq!(


kormide (approved):

UebelAndre:
Thanks!

Isn't the build script binary already in `exec`? There is an intermediate rule called `cargo_build_script_runfiles` which consumes the `rust_binary` as exec:
https://github.com/bazelbuild/rules_rust/blob/d256c7bb4668bbdfd1e25971531c874c23f4646a/cargo/private/cargo_build_script.bzl#L101-L106

And then this rule creates a symlink from that `cfg = "exec"` binary:
https://github.com/bazelbuild/rules_rust/blob/d256c7bb4668bbdfd1e25971531c874c23f4646a/cargo/private/cargo_build_script.bzl#L44-L62

And returns it as the `DefaultInfo.executable`:
https://github.com/bazelbuild/rules_rust/blob/d256c7bb4668bbdfd1e25971531c874c23f4646a/cargo/private/cargo_build_script.bzl#L72

This way when `cargo_build_script` consumes the `cargo_build_script_runfiles` target, it is still running a binary built for `cfg = "exec"` even though this attribute consumes it as `cfg = "target"`:
https://github.com/bazelbuild/rules_rust/blob/d256c7bb4668bbdfd1e25971531c874c23f4646a/cargo/private/cargo_build_script.bzl#L624-L630

Is there something broken about this interaction? I've also not had much success with cross platform RBE builds (https://github.com/bazelbuild/bazel/issues/19587) but assuming the execution platform is described to be linux, in your case, I would expect all binaries to be built for the right platform. Curious to know how this wouldn't be the case and how to fix it.

mrmeku:
Let me play with this a bit. I have a repo that has RBE so it'll be fast for me to iterate on potential fixes. I'll reply to this when I have more a confirmed fix since this is something that's especially hard to make a test case for with current infrastructure

illicitonion (commented):
@UebelAndre IIRC you put a lot of careful thought into the transitions here?

mrmeku:
Okay, your analysis was right. I'm going to close this for now and reopen when I've resolved the error. There is still a bug (unless my RBE setup is at fault).

```
ERROR: /private/var/tmp/_bazel_mrmeku/f426e18877a0b787c58d9aeaf02e075f/external/rules_rust_tinyjson/BUILD.bazel:4:37: Compiling Rust (without process_wrapper) rlib tinyjson (4 files) [for tool] failed: (Exit 1): bootstrap_process_wrapper.sh failed: error executing Rustc command (from target @@rules_rust_tinyjson//:tinyjson) bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rules_rust/util/process_wrapper/bootstrap_process_wrapper.sh -- ... (remaining 21 arguments skipped)
bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rules_rust/util/process_wrapper/bootstrap_process_wrapper.sh: line 18: /private/var/tmp/_bazel_mrmeku/f426e18877a0b787c58d9aeaf02e075f/execroot/_main/bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/rustc: cannot execute binary file
bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rules_rust/util/process_wrapper/bootstrap_process_wrapper.sh: line 18: /private/var/tmp/_bazel_mrmeku/f426e18877a0b787c58d9aeaf02e075f/execroot/_main/bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/rustc: Undefined error: 0
```

mrmeku:
The key to what I'm seeing is 
```
bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rules_rust/util/process_wrapper/bootstrap_process_wrapper.sh: line 18: /private/var/tmp/_bazel_mrmeku/f426e18877a0b787c58d9aeaf02e075f/execroot/_main/bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/rustc: Undefined error: 0
```

The process wrapper is clearly mismatched to the rustc toolchain. Looking into why that is

UebelAndre:
> The key to what I'm seeing is
> 
> ```
> bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rules_rust/util/process_wrapper/bootstrap_process_wrapper.sh: line 18: /private/var/tmp/_bazel_mrmeku/f426e18877a0b787c58d9aeaf02e075f/execroot/_main/bazel-out/darwin_arm64-opt-exec-ST-4c0668394d32/bin/external/rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/rustc: Undefined error: 0
> ```
> 
> The process wrapper is clearly mismatched to the rustc toolchain. Looking into why that is

Very interesting. I can believe there’s a bug like this somewhere in the process wrapper bootstrapped. Do you know if toolchain resolutions matches what you expect? I’m wondering if there’s a mix up there or something started referring to the `rust_host_tools` repository inappropriately

PR #3237, opened 2025-02-04
Mark cargo internal extension as reproducible

cerisier:
Since all repo from cargo_repositories are instantiated with the same attribute values. Saves an entry in the lockfile.


cerisier:
CI errors seem unrelated.

PR #3239, opened 2025-02-05
Reduce number of default `supported_platform_triples` for crate_universe

UebelAndre:
This should save substantial time for users of `crate_universe` as they don't need to do splicing for platforms they're not going to use. It's still recommended users explicitly set `supported_platform_triples` but my expectation with this change is that it covers most of the bases for users to still provide them with a sufficiently responsive experience compared to the 34 platforms that used to be in splicing.


UebelAndre:
@illicitonion I'm curious if this should just be a mandatory field. It impacts the lockfile so this change is gonna require everyone to repin. It previously changed anytime someone added a new platform and this would trigger another change.

You'll also notice I only updated the bzlmod extension. I think `crates_vendor` is fine to run all platforms as this workflow suggests updates are generally less frequent. If you object I can keep the workflows in sync.

illicitonion (approved):
Do we want to do this for WORKSPACE too?

PR #3245, opened 2025-02-07
Set $CARGO in cargo_build_scripts

illicitonion:
Some tools, like bindgen, look for this (and will fall back to $PATH if not set). This encourages them to use the hermetic cargo from our toolchain, rather than missing that and maybe ending up falling back to a system cargo.


UebelAndre (approved):
Nice! I also didn't realize the toolchain had cargo in it. Would be nice if there was a second toolchain dedicated to cargo things, but not necessary for this change. Looks good to me!

UebelAndre:
By the way, your PR builds were hit with https://github.com/bazelbuild/continuous-integration/issues/2170, this should be good to go if you're willing to retry over and over 😅

havasd:
Just a note that not sure how reliable will this be. Some crates use `cargo metadata` some use other `cargo` commands. For example `metadata` requires the whole workspace to be available so it works correctly. 

Otherwise if some rust files are missing it won't be able to properly generate the metadata.

I think that some kind of "mock" cargo could be beneficial for bazel for these crates or fix the crates so that they can work without it.

PR #3254, opened 2025-02-11
rust_analyzer: Support passing a --config option to Bazel

konkers:
This is useful if you want to analyze code using a different configuration than the default.  An example of this might be under a different platform.


illicitonion (approved):
Seems reasonable! Thanks!

PR #3257, opened 2025-02-11
Handle debug info packages separately when using `cc_common.link`

yuzhy8701:
Debug info packages (`dSYM`, `pdb`) are created by `rustc` only if you link with `rustc`. When linking with `cc_common`, they should be handled by the cc toolchain or a custom aspect. Declaring them prematurely can cause conflicts.

This PR removes the `dsym_folder` declaration if `cc_common.link` is being used - since cc_common doesn't support dsym generation yet. It postpones the declaration of `pdb_file` after checking with the cpp features first.


UebelAndre (approved):
Thanks!

PR #3258, opened 2025-02-11
Don't override NUM_JOBS if already set

froody:
Building large src crates (e.g. proj_sys) take a long time because NUM_JOBS was always 1. This allows the user to overried (e.g. with --actiona_env) the value of NUM_JOBS

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -365,12 +365,15 @@ def _cargo_build_script_impl(ctx):
         "CARGO_MANIFEST_DIR": manifest_dir,
         "CARGO_PKG_NAME": pkg_name,
         "HOST": toolchain.exec_triple.str,
-        "NUM_JOBS": "1",
         "OPT_LEVEL": compilation_mode_opt_level,
         "RUSTC": toolchain.rustc.path,
         "TARGET": toolchain.target_flag_value,
         # OUT_DIR is set by the runner itself, rather than on the action.
     })
+
+    # Allow the user to override the number of jobs for the build script.
+    if "NUM_JOBS" not in env:
+        env["NUM_JOBS"] = 1
 
     # This isn't exactly right, but Bazel doesn't have exact views of "debug" and "release", so...
     env.update({


cargo/private/cargo_build_script.bzl
```diff
@@ -365,13 +365,16 @@ def _cargo_build_script_impl(ctx):
...
         "HOST": toolchain.exec_triple.str,
-        "NUM_JOBS": "1",
         "OPT_LEVEL": compilation_mode_opt_level,
         "RUSTC": toolchain.rustc.path,
         "TARGET": toolchain.target_flag_value,
         # OUT_DIR is set by the runner itself, rather than on the action.
     })
 
+    # Allow the user to override the number of jobs for the build script.
+    if "NUM_JOBS" not in env:
```
UebelAndre, line 375: If this is any value larger than 1 then it would be good to have an associated resource set

UebelAndre (changes requested):
Thanks! Just had one request

cargo/private/cargo_build_script.bzl
froody: I don't know what that means, what's an "associated resource"?

UebelAndre: Basically a change similar to https://github.com/bazelbuild/rules_rust/pull/3112

which uses `ctx.actions.run.resource_set`

UebelAndre: You'll find the big ugly file `rust/private/rustc_resource_set.bzl` and it's use in `rust/private/rustc.bzl`
```starlark
resource_set = get_rustc_resource_set(toolchain)
```

froody: @UebelAndre I can't call get_rustc_resource_set as it stands without refactoring it to take an int instead of a toolchain, do you want me to do that, or duplicate `rust/private/rustc_resource_set.bzl` with an int variant?

UebelAndre: You could just add a new function. `get_cargo_build_script_resource_set` or something that does what you need. When I wrote that all I was thinking about was `Rustc` actions but I don't think we should duplicate these functions. So I'd just make a new function for the use here.

UebelAndre (changes requested):

PR #3259, opened 2025-02-11
Add support for ppc64le architecture

Jenkins-J:
Allow rules_rust module to be used on the ppc64le architecture.

### DIFF
--- a/rust/platform/triple.bzl
+++ b/rust/platform/triple.bzl
@@ -120,7 +120,7 @@ def get_host_triple(repository_ctx, abi = None):
     # Detect the host's cpu architecture
 
     supported_architectures = {
-        "linux": ["aarch64", "x86_64", "s390x"],
+        "linux": ["aarch64", "x86_64", "s390x", "powerpc64le"],
         "macos": ["aarch64", "x86_64"],
         "windows": ["aarch64", "x86_64"],
     }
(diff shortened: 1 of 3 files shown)


rust/platform/triple.bzl
```diff
@@ -117,7 +117,7 @@ def get_host_triple(repository_ctx, abi = None):
     # Detect the host's cpu architecture
 
     supported_architectures = {
-        "linux": ["aarch64", "x86_64", "s390x"],
+        "linux": ["aarch64", "x86_64", "s390x", "powerpc64le"],
```
UebelAndre, line 120: Is the host platform ever gonna be powerpc? I think the only changes you need are what was done in `rust/platform/triple_mappings.bzl`.

UebelAndre (changes requested):
Thanks! Just one question though

rust/platform/triple.bzl
Jenkins-J: If I understand correctly, the host platform is the platform the bazel binary is executed on. When attempting to utilize rules_rust and executing bazel on a ppc64le host, I did receive the failure message from `_validate_cpu_architecture`, causing the build to fail. Marking it as supported here seems necessary to avoid failure in that case.

UebelAndre (approved):

UebelAndre:
@Jenkins-J Looks like this PR will need to be rebased and the crate_universe examples repinned.

UebelAndre:
> @UebelAndre I was able to rebase the PR. What is the recommended way to repin the crate_universe examples? I am not familiar with that process. Is there a specific bazel rule in the cargo_universe directory that I must run with `CARGO_BAZEL_REPIN=1` set?

@Jenkins-J you should be able to run this script and commit the results: [//util/repin_all.sh](https://github.com/bazelbuild/rules_rust/blob/4d023c13d3808af074afecdcb300dc232bc1699d/util/repin_all.sh)

PR #3263, opened 2025-02-12
Replace uses of //tools/cpp:toolchain_utils.bzl with @rules_cc//cc:find_cc_toolchain.bzl

comius:
The former location was deprecated


UebelAndre (approved):
Thanks!

PR #3268, opened 2025-02-13
Silence noisy cross_rs warnings

UebelAndre:


crate_universe/tools/cross_installer/patches/cross_rs.static_mut_refs.patch
UebelAndre, line 1: This patch can be removed once https://github.com/cross-rs/cross/issues/1561 is addressed.

illicitonion (approved):

PR #3271, opened 2025-02-14
Propagate cc_library defines to clang

nbdd0121:
Currently if `cc_library` uses `defines = [...]`, it won't be picked by bindgen.

This is a 2-line fix to add the support (with tests).


UebelAndre (approved):
Thanks!

PR #3276, opened 2025-02-17
Added expected providers to `rust_prost_library`.

UebelAndre:
This should help with aspects ability to discover these targets.


UebelAndre:
> We ran into a similar issue recently, where `rust_prost_library` wasn't exposing all the necessary providers, and as such was causing a missing symbols error after linking.

Interesting, do you have a minimal repro for the issue? I've not encountered linker issues with these prost rules before.

illicitonion (approved):

alexkirsz:
@UebelAndre Our repro case involved `rules_swift` Swift libraries depending on Rust libraries, so it was a bit involved. I'll see if I can get a more minimal repro case for this.

PR #3292, opened 2025-02-24
Merge all runfiles of `@bazel_tools//tools/bash/runfiles`

fmeum:
As `@bazel_tools//tools/bash/runfiles` is now an alias to `@rules_shell//shell/runfiles` with Bazel@HEAD, consumers must properly merge all its runfiles rather than assuming that the `sh_library` target just consists of a single file.

Work towards https://github.com/bazelbuild/examples/pull/557#issuecomment-2677865427


fmeum:
@UebelAndre @illicitonion Could you review this? It's currently transitively breaking Bazel's downstream pipeline for examples and rules_rust.

UebelAndre (approved):

Issue #3286, opened 2025-02-22
Unable to specify rust version in crate_universe bzlmod


UebelAndre:
It's currently not possible to define a custom version of Rust for crate universe extensions

```starlark
RUST_VERSION = "1.84.0"

rust_host_tools = use_extension("@rules_rust//rust:extensions.bzl", "rust_host_tools")
rust_host_tools.host_tools(
    name = "my_rust_host_tools",
    version = RUST_VERSION,
)
use_repo(
    rust_host_tools,
    "my_rust_host_tools",
)

crate = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")

crate.from_cargo(
    name = "crate_index",
    manifests = ["//:Cargo.toml"],
    cargo_lockfile = "//:Cargo.lock",
    lockfile = "//:cargo-bazel-lock.json",
    host_tools_repo = "my_rust_host_tools",
    supported_platform_triples = [
        "aarch64-apple-darwin",
        "aarch64-unknown-linux-gnu",
        "x86_64-pc-windows-msvc",
        "x86_64-unknown-linux-gnu",
    ],
)

use_repo(
    crate,
    "crate_index",
)
```

This will fail with the following error
```
ERROR: /private/var/tmp/_bazel_user/b8286dc79fef4b6257e32e451f5cae57/external/rules_rust+/crate_universe/extensions.bzl:833:37: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_user/b8286dc79fef4b6257e32e451f5cae57/external/rules_rust+/crate_universe/extensions.bzl", line 949, column 59, in _crate_impl
		cargo_path, rustc_path = _get_host_cargo_rustc(module_ctx, host_triple, cfg.host_tools_repo)
	File "/private/var/tmp/_bazel_user/b8286dc79fef4b6257e32e451f5cae57/external/rules_rust+/crate_universe/extensions.bzl", line 833, column 37, in _get_host_cargo_rustc
		cargo_path = str(module_ctx.path(Label("@{}//:bin/cargo{}".format(host_tools_repo, binary_ext))))
[...]

PR #3293, opened 2025-02-24
Allow crate_universe modules to specify rust versions

UebelAndre:
This change updates bzlmod crate_universe to allow modules to specify the version of rust used for splicing. Previously `rust_host_tools` and `rust_host_tools_nightly` were the only repositories available to consumers which both had fixed rust versions specified by `rules_rust`. This change gives users easy control of this value now.

closes https://github.com/bazelbuild/rules_rust/issues/3286


illicitonion (approved):
Yeah, I don't think I have any better approaches here... Toolchains in repo_rule world are tricky...!

PR #3298, opened 2025-02-26
Allow target names containing `.` for rust_test_suite macro.

vaibhav-1207:
1. Computes `crate_name` attribute for each target globed by rust_test_suite by replacing all the existing illegal characters and `.` from the target name. This helps to glob target names containing `.` and also prevents collisions of crate_name.
2. Adds `_test` suffix to test_name of each target to standardize the naming convention.


vaibhav-1207:
Done, PTAL!

krasimirgg (approved):
Thank you!

PR #3309, opened 2025-03-04
process_wrapper: Convert non-JSON diagnostics from LLVM to JSON.

martinboehme:
Even when rustc is run with `--error-format=json`, LLVM may still [output
human-readable diagnostics](https://github.com/llvm/llvm-project/blob/680391f07a45272bb9bfd385cf4c6846b8be32dd/llvm/lib/MC/MCSubtargetInfo.cpp#L82)
that are not formatted as JSON. So far, process_wrapper has choked on these.

This patch detects these lines and converts them to JSON so that they can be
processed like JSON diagnostics from rustc itself.

The correct fundamental fix would likely be to teach rustc itself to perform
this conversion, but this is a change that would require some time to implement,
so we're putting this fix in place instead.

### DIFF
--- a/util/process_wrapper/main.rs
+++ b/util/process_wrapper/main.rs
@@ -18,13 +18,17 @@ mod output;
 mod rustc;
 mod util;
 
+use std::collections::HashMap;
 use std::fmt;
 use std::fs::{copy, OpenOptions};
 use std::io;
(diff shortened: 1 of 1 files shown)


krasimirgg (approved):
Thank you!

PR #3315, opened 2025-03-09
Moved all integration tests to `test/integration`

UebelAndre:
Integration tests in this context are defined as any tests which are expected to be run as it's own workspace (aka, have their own `WORKSPACE.bazel`/`MODULE.bazel` files). This makes the number of these easily identifiable and defines a pattern for how to introduce more going forward.

Additionally, this change moved the `cc_common_link` examples to use bzlmod.


illicitonion (approved):
Sure!

PR #3316, opened 2025-03-09
Implement priority support for extract_cargo_lints rule

havasd:
Clippy prioritizes the lint flags based on the order in the command line. This can be ordered by the priority field in the Cargo.toml file. More information: https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section

Additionaly generate empty files when lints are not present in Cargo.toml file instead of throwing an error. This makes generation of such rules easier

FYI: @ParkMyCar


ParkMyCar:
Nice! Thanks @havasd!

illicitonion (approved):
Seems reasonable! Thanks!

PR #3320, opened 2025-03-10
Add firefox linux binaries for `rust_wasm_bindgen_test`

UebelAndre:
The firefox binaries are only available for linux since there is no easy way to extract a `.dmg` for macos and the `.exe` that firefox distributes for the relevant platforms.

Additionally, the `@rules_rust_wasm_bindgen//:test_browser` setting moved to `@rules_rust_wasm_bindgen//settings:test_browser`


UebelAndre:
Without this the Bazel Central Registry CI would have failed since it runs `bazel test //...` and would have hit the issue mentioned here:
- https://github.com/bazelbuild/rules_rust/pull/3267#discussion_r1986122280

illicitonion (commented):
When I have a moment I'll try to get macOS working for this, but I'm not sure when I'll have a moment :) Definitely feel free to if you have any time/inclination!

illicitonion (approved):

UebelAndre:
> When I have a moment I'll try to get macOS working for this, but I'm not sure when I'll have a moment :) Definitely feel free to if you have any time/inclination!

I mean, I developed this on a Mac. If you extract firefox from the dmg (e.g. from https://ftp.mozilla.org/pub/firefox/releases/136.0/mac/en-US/) then this works just fine. Ideally there would be a rule that could do this extraction or Mozilla would host the apps in a tar file somewhere 😅

PR #3323, opened 2025-03-10
Update Crates Index using rev

taj-p:
Depending on how long it takes to get a 3.6.1 release of https://github.com/frewsxcv/rust-crates-index, it might be worth us referencing a commit in the meantime.


UebelAndre (approved):
Thanks!

PR #3326, opened 2025-03-11
extensions/prost/private/protoc_wrapper: codegen for enums was broken

zllak:
When generating a message that contains a field that is an enum, the generated code was not following the proper upper camel case convention, making it fail to build.
Example:
```
message Test {
    some.pkg.SomethingID test = 1;
}
```
```
    #[prost(enumeration="super::some::pkg::SomethingID", tag="1")]
    pub test: i32,
```
When we actually expects `SomethingId` because prost generates it like that.


UebelAndre (approved):
Thank you!

PR #3328, opened 2025-03-11
Added `safari` support for `rust_wasm_bindgen_test`

UebelAndre:
Additionally fix missing wasm-bindgen CI tests


UebelAndre:
> This is getting to be quite a lot of webdriver infra which feels like maybe it should live in another ruleset?

I would love for there to be a maintained module for each webdriver thing! That may be a larger conversation that I'll kick off somewhere.

PR #3334, opened 2025-03-12
[path_patching] Shrink size of lockfiles

jayzhuang:
... by only plumbing local patched paths.

Change-Id: Idfaf53d93fe11776a70d706726ed4bff0bf5233e

### DIFF
--- a/crate_universe/src/context/crate_context.rs
+++ b/crate_universe/src/context/crate_context.rs
@@ -2,6 +2,7 @@
 
 use std::collections::{BTreeMap, BTreeSet};
 
+use camino::Utf8PathBuf;
 use cargo_metadata::{Node, Package, PackageId};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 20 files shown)


crate_universe/src/context/crate_context.rs
```diff
@@ -26,9 +27,9 @@ pub struct CrateDependency {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub alias: Option<String>,
 
-    /// Where to acquire the source of this dependency.
+    /// Local path of this dependency if provided.
```
UebelAndre, line 30: Is this if it's a path dependency at all or of you're using `[patch]`? Reading back to https://github.com/bazelbuild/rules_rust/pull/3071 it seems like that's the case and if so can you make this a bit more wordy. I'm worried I'll forget this conversation and go through the same hoops in my mind again about this data feeling duplicated.

UebelAndre (changes requested):
Thank you! I just had one request

crate_universe/src/context/crate_context.rs
jayzhuang: The support was added in https://github.com/bazelbuild/rules_rust/pull/3025/commits/7b53a90fc4034bf1e8d5b2b7bc71d98accce9ef3, reading the commit message there it claims it supports both dependencies and patches, so extended the comment here to say the same.

Also can you help me fix the "Nix cross compiling test"? I don't have a working nix set up locally, IIUC it just needs a repin.

UebelAndre (approved):
Thank you!

crate_universe/src/context/crate_context.rs
UebelAndre: You can repin without nix. When you try repunning there the repin will succeed but then the invocation will fail due to missing nix but at that point you got what you wanted.

PR #3338, opened 2025-03-13
Fix broken bindgen doc link

lalten:
It should've been /README.md but I now think linking to the "folder" is actually nicer.


UebelAndre (approved):
Thanks!

Issue #3349, opened 2025-03-16
Error when updating to 0.59.1 Cargo Toml Variable Extractor


ericmcbride:
The following Error is being thrown on 0.59.1  on compile

```ERROR: /private/var/tmp/_bazel_eric.mcbride/79b277eea1629c6c0e66b29d7b082943/external/rules_rust+/cargo/cargo_toml_variable_extractor/BUILD.bazel:3:12: Compiling Rust bin cargo_toml_variable_extractor (1 files) [for tool] failed: (Exit 1): process_wrapper failed: error executing Rustc command (from target @@rules_rust+//cargo/cargo_toml_variable_extractor:cargo_toml_variable_extractor) bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/external/rules_rust+/util/process_wrapper/process_wrapper --arg-file ... (remaining 95 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
error[E0716]: temporary value dropped while borrowed
  --> external/rules_rust+/cargo/cargo_toml_variable_extractor/main.rs:89:10
   |
84 |     let version = if let Some(version) =
   |         ------- borrow later stored here
...
89 |         &semver::Version::parse("0.0.0").expect("Known-good version 0.0.0 couldn't be parsed")
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
90 |     };
   |     - temporary value is freed at the end of this statement
   |
   = note: consider using a `let` binding to create a longer lived value

error: aborting due to 1 previous error
```

PR #3350, opened 2025-03-16
Fix temp value is freed

ericmcbride:
Fixes[ issue](https://github.com/bazelbuild/rules_rust/issues/3349)

Tested and verified


UebelAndre (changes requested):
Thanks! But how is it possible this was not caught in CI?

illicitonion (approved):
I worked out why it wasn't caught in CI - I'll send a PR @UebelAndre 

Thanks so much for the fix, @ericmcbride!

illicitonion:
Fix to CI in #3353 - I verified it fails without this PR and passes with :) Thanks again!

PR #3352, opened 2025-03-16
Updated unit tests for runfiles

UebelAndre:
This introduces a small re-implementation of https://crates.io/crates/temp-env while keeping the tests free of dependencies. This should make it substantially easier to write meaningful tests for runfiles.


illicitonion (approved):
Nice! Thanks!

PR #3353, opened 2025-03-16
Update MSRV in MODULE.bazel as well as WORKSPACE.bazel

illicitonion:
Since our testing mostly uses bzlmod these days, we've currently regressed in this coverage.


UebelAndre (approved):
Nice catch, thanks everyone!

.bazelci/presubmit.yml
```diff
@@ -17,7 +17,8 @@ aspects_flags: &aspects_flags
   - "--config=rustfmt"
   - "--config=clippy"
 min_rust_version_shell_commands: &min_rust_version_shell_commands
-  - sed -i 's|^rust_register_toolchains(|rust_register_toolchains(versions = ["1.72.1"],\n|' WORKSPACE.bazel
+  - sed -i 's|^rust_register_toolchains(|rust_register_toolchains(versions = ["1.74.0"],\n|' WORKSPACE.bazel
```
UebelAndre, line 20: You might wanna make a similar change for the `nightly_toolchain` tests below as well. Any reference to `WORKSPACE.bazel` here should probably also have a `MODULE.bazel` update.

illicitonion: Done! Thanks for the heads up!

PR #3354, opened 2025-03-17
Run browser tests with Safari on MacOS

UebelAndre:


UebelAndre:
With https://github.com/bazelbuild/continuous-integration/issues/2190 enabled this might be possible.

PR #3363, opened 2025-03-18
Fix rust_bindgen compile flag filtering

### LINKED ISSUE #3359: `rust_bindgen` incorrectly filters non-clang args
`rust_bindgen` attempts to filter out non-clang compiler flags when copying over flags from the current C++ toolchain, but a bug allows unsupported flags to be passed through. Specifically, the implementation assumes that every flag that starts with an allowlisted prefix is valid, and that every flag that exactly matches an allowlist entry will be followed by a parameter value. For example, `"-Ipath"` is allowed, and `"-I"` is allowed and causes the next argument to be passed through as well (e.g. `"path"`). While it works in general, this approach does not work with parameterless flags such as `-no-canonical-prefixes`.

For example `["-no-canonical-prefixes", "-fno-canonical-system-headers"]` causes the gcc-only flag `"-fno-canonical-system-headers"` to be passed through.

https://github.com/bazelbuild/rules_rust/blob/2f81c8d3e7b81bb493d84af8a34d8ecc060a3fbd/extensions/bindgen/private/bindgen.bzl#L274-L308

I'm happy to send a PR if you agree this is an issue. Thanks!

bmclarnon:
All flags were treated as accepting a parameter, causing any argument following a parameterless flag (e.g. `-no-standard-includes`) to be passed through. This causes bindgen failures if the next argument isn't supported by clang.

`-nostd*` flags has also been expanded using
https://clang.llvm.org/docs/ClangCommandLineReference.html to make it clearer that these flags are parameterless. Since these flags resulted in a prefix match, not a full match, they wouldn't set `open_arg = True` and therefore wouldn't cause issues, but this behavior is unnecessarily subtle now that there are two lists.

Fixes #3359

### DIFF
--- a/extensions/bindgen/private/bindgen.bzl
+++ b/extensions/bindgen/private/bindgen.bzl
@@ -271,7 +271,8 @@ def _rust_bindgen_impl(ctx):
     # Ideally we could depend on a more specific toolchain, requesting one which is specifically clang via some constraint.
     # Unfortunately, we can't currently rely on this, so instead we filter only to flags we know clang supports.
     # We can add extra flags here as needed.
-    flags_known_to_clang = (
+    # Flags in this tuple accept a parameter in the same argument (`-Ipath`, `--target=T`) or separately (`-I path`).
+    param_flags_known_to_clang = (
         "-I",
         "-iquote",
         "-isystem",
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thanks!

UebelAndre:
@bmclarnon looks like buildifier needs to run.

PR #3370, opened 2025-03-22
Set env vars for build scripts too

illicitonion:


UebelAndre (approved):
Nice!

Pagten:
Hmm, doesn't this PR set the env vars at build script _compilation_ time instead of build script _execution_ time?

As indicated [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), the env vars are expected to be available at build script execution time:

> Cargo sets several environment variables when build scripts are run. Because these variables are not yet set when the build script is compiled, the above example using env! won’t work and instead you’ll need to retrieve the values when the build script is run: [...]

illicitonion:
> Hmm, doesn't this PR set the env vars at build script _compilation_ time instead of build script _execution_ time?
> 
> As indicated [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), the env vars are expected to be available at build script execution time:
> 
> > Cargo sets several environment variables when build scripts are run. Because these variables are not yet set when the build script is compiled, the above example using env! won’t work and instead you’ll need to retrieve the values when the build script is run: [...]

Many of the env vars in that list are set at _both_ compile _and_ runtime. If there are specific env vars you're finding missing, either at compile or runtime, please file an issue or PR about them specifically :)

aldanor:
@illicitonion I've just run into this today. Specifically, when building packages like `rav1e` that [depend](https://github.com/xiph/rav1e/blob/a2f01b3e233f531c28a20b4c29fb5c9e5d29fa6d/build.rs#L249) on `built` in their build script.

```
ERROR: /private/var/tmp/_bazel_ivan/02cc017613cea464e46f3fd311da1117/external/rules_rust++crate+crates__rav1e-0.7.1/BUILD.bazel:98:19: Ru
  (cd /private/var/tmp/_bazel_ivan/02cc017613cea464e46f3fd311da1117/sandbox/darwin-sandbox/1222/execroot/_main && \
  exec env - \
    ...
    CARGO_CRATE_NAME=rav1e \
    CARGO_ENCODED_RUSTFLAGS='--sysroot=${pwd}/bazel-out/darwin_arm64-opt/bin/external/rules_rust++rust+rust_macos_aarch64__aarch64-apple-
    CARGO_FEATURE_THREADING=1 \
    CARGO_MANIFEST_DIR=bazel-out/darwin_arm64-opt/bin/external/rules_rust++crate+crates__rav1e-0.7.1/_bs.cargo_runfiles/rules_rust++crate
    CARGO_PKG_NAME=rav1e \
    CARGO_PKG_VERSION=0.7.1 \
    CARGO_PKG_VERSION_MAJOR=0 \
    CARGO_PKG_VERSION_MINOR=7 \
    CARGO_PKG_VERSION_PATCH=1 \
    CARGO_PKG_VERSION_PRE='' \
    ...
```

```
thread 'main' panicked at external/rules_rust++crate+crates__built-0.7.7/src/environment.rs:54:9:
Missing expected environment variable "CARGO_PKG_AUTHORS"
```

It seems to crash on `built` crate [trying to find](https://github.com/lukaslueg/built/blob/aeb527706edfac67c8a34de8d95e2d90b6bd352a/src/environment.rs#L48) `CARGO_PKG_AUTHORS` env var which it can't, which makes the build script panic.

Is my understanding correct in that those env vars are not being propagated correctly to the build script's execution environment?

aldanor:
@illicitonion Here seems to be the minimal patch in order for it to build (not 100% sure the rustdoc part is correct but it seems to work):
```starlark
crate.annotation(
    crate = "rav1e",
    build_script_tools = [
        "@rules_rust//rust/toolchain:current_rustdoc_files",
    ],
    build_script_env = {
        "CARGO_PKG_AUTHORS": "Thomas Daede <tdaede@xiph.org>",
        "CARGO_PKG_DESCRIPTION": "The fastest and safest AV1 encoder",
        "CARGO_PKG_HOMEPAGE": "",
        "CARGO_PKG_REPOSITORY": "https://github.com/xiph/rav1e/",
        "CARGO_PKG_LICENSE": "BSD-2-Clause",
        "CARGO_PKG_LICENSE_FILE": "",
        "RUSTDOC": "$(execpath @rules_rust//rust/toolchain:current_rustdoc_files)",
    },
)
```

PR #3381, opened 2025-04-01
Explicitly set `-Ccodegen-units=1` for actions that emit an object file.

martinboehme:
Also, remove any other `-Ccodegen-units` that may have been set in various ways.

The build rules expect to see a single object file, not the multiple object files that are produced when `-Ccodegen-units` (with a setting greater than 1) is used.

The `_add_codegen_units_flags()` function already took this into account (`is_codegen_units_enabled()` returns false when emitting an object file), but there are various other ways that `-Ccodegen-units` might get introduced into the command line -- see the included test for details.


rust/private/rustc.bzl
```diff
@@ -1082,36 +1096,38 @@ def construct_arguments(
...
 
+    extra_flags = []
+
     if crate_info.type in toolchain.extra_rustc_flags_for_crate_types.keys():
-        rustc_flags.add_all(toolchain.extra_rustc_flags_for_crate_types[crate_info.type])
+        extra_flags.extend(toolchain.extra_rustc_flags_for_crate_types[crate_info.type])
 
     if is_exec_configuration(ctx):
-        rustc_flags.add_all(toolchain.extra_exec_rustc_flags)
+        extra_flags.extend(toolchain.extra_exec_rustc_flags)
```
scentini, line 1105: Before this change, we were using Bazel's [`Args`](https://bazel.build/rules/lib/builtins/Args.html), which, compared to an array, is more memory efficient. I think the `Args` object does not create copies of  `toolchain.extra_rustc_flags` and similar arrays, but with the array, I believe we now have a copy of those for each Rust target in the build graph. 

My initial suggestion would have been to keep using `Args`, and at the end, have a:
```
def filter_codegen_units_fn(flag):
    if flag.startswith("-Ccodegen-units"):
        return None
    return flag

map_fn = filter_codegen_units_fn if "obj" in emit else None

maybe_filtered_flags = ctx.actions.args().add_all(rustc_flags, map_each = map_fn)
```

Turns out however that one cannot merge `Args` objects just yet: https://github.com/bazelbuild/bazel/issues/6230

What I think will work is to pass the `map_fn` function to all the `rustc_flags.add_all()` calls.

martinboehme: > What I think will work is to pass the map_fn function to all the rustc_flags.add_all() calls.

I originally had something similar here: I was calling `_remove_codegen_units_flags()` on the various arguments to `rustc_flags.add_all()` that require it, e.g.

```
rustc_flags.add_all(_remove_codegen_units_flags(toolchain.extra_exec_rustc_flags, emit))
```

However, this is verbose and repetitive, and I was looking for a way that would require me to call `_remove_codegen_units_flags` on every line that requires it.

I'm only adding specific "extra" flags to the `extra_flags` array here, and I expect that a lot of the time, these "extra" flags will be empty. Is this really so much of an efficiency concern? If it is, I'm happy to restore what I was doing here originally -- just wanted to make sure the readability hit is actually required.

scentini: It's not about the values you're adding, but rather about how blaze keeps these things in memory:

in a typical build, there will be one `toolchain` for the target configuration, and one for `exec` configuration, and many many `rust_library` targets.
The logic you're changing affects all the `rust_*` rules in the build graph.
What I believe is happening is - before the change, we had `rust_flags.add_all(toolchain.extra_rustc_flags)`, which would keep one copy of `toolchain.extra_rustc_flags` and share it across the build graph, and after the change, with the array we're creating a copy of `toolchain.extra_rustc_flags` for each `rust_*` target (of which there can be 100K+ in a single build).

Wrt readability, I agree that this is not great:
`rustc_flags.add_all(_remove_codegen_units_flags(toolchain.extra_exec_rustc_flags, emit))`
But this seems tolerable:
`rustc_flags.add_all(toolchain.extra_exec_rustc_flags, map_each = maybe_remove_codegen_flags)`

martinboehme: Makes sense. Done!

scentini (approved):
Thanks!

PR #3385, opened 2025-04-03
Draft: Add new cargo creds feature

ericmcbride:
- Add Cargo Credentials to the Module.bazel `from_crates` and vendor calls.
- Currently private crates repositories cannot pull from repos by default.  You have to set `CARGO_BAZEL_ISOLATED=false` for it to work

### DIFF
--- a/crate_universe/private/common_utils.bzl
+++ b/crate_universe/private/common_utils.bzl
@@ -115,7 +115,7 @@ def get_rust_tools(repository_ctx, host_triple):
     Returns:
         struct: A struct containing the expected rust tools
     """
-
+    
     # This is a bit hidden but to ensure Cargo behaves consistently based
     # on the user provided config file, the config file is installed into
     # the `CARGO_HOME` path. This is done so here since fetching tools
@@ -125,6 +125,13 @@ def get_rust_tools(repository_ctx, host_triple):
         cargo_home_config = repository_ctx.path("{}/config.toml".format(cargo_home))
         cargo_config = repository_ctx.path(repository_ctx.attr.cargo_config)
         repository_ctx.symlink(cargo_config, cargo_home_config)
+    
+    # Workspace backwards compat for cargo credentia
+    if repository_ctx.attr.isolated and repository_ctx.attr.cargo_creds:
+        cargo_home = _cargo_home_path(repository_ctx)
+        cargo_home_config = repository_ctx.path("{}/credentials.toml".format(cargo_home))
+        cargo_creds = repository_ctx.path(repository_ctx.attr.cargo_creds)
+        repository_ctx.symlink(cargo_creds, cargo_home_config)
 
     if repository_ctx.attr.rust_version.startswith(("beta", "nightly")):
         channel, _, version = repository_ctx.attr.rust_version.partition("/")
--- a/crate_universe/src/splicing/splicer.rs
+++ b/crate_universe/src/splicing/splicer.rs
@@ -155,8 +155,16 @@ impl<'a> SplicerKind<'a> {
             Some(IGNORE_LIST),
         )?;
 
-        // Optionally install the cargo config after contents have been symlinked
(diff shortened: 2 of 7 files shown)


crate_universe/src/splicing/splicer.rs
```diff
@@ -290,16 +319,28 @@ impl<'a> SplicerKind<'a> {
...
+
+                symlink_roots(
+                    &real_path,
+                    &dot_cargo_dir,
+                    Some(&[dot_file_root, dot_file_toml]),
+                )?;
+
+                // if cargo home set we symlink to cargo home.
+                // This is so credentials.toml works.
+                if let Ok(cargo_home) = std::env::var("CARGO_HOME") {
```
ericmcbride, line 331: @UebelAndre Is there some other way to check that im not aware of to see if the environment is isolated?  Can i check `CARGO_BAZEL_ISOLATED` is set, and if its false, we ignore this and just assume if its not set, or a true we can symlink `credentials.toml` to cargo home?

UebelAndre: I don't think we've had to care if the environment is isolated from within the cargo-bazel binary. Isn't the starlark responsible for setting up the isolated `CARGO_HOME`? If you want credentials to be installed there (which would come from the user's `CARGO_HOME` anyway), why not install it there?

I would pass an explicit flag indicating isolation is enabled though, that sounds fine 😄

ericmcbride: The symlink appears to happen whenever `get_rust_tools` is called, and the `extensions.bzl` does not use that call to get the rustc / cargo

UebelAndre: I think this is shared with both bzlmod and workspace modes
https://github.com/bazelbuild/rules_rust/blob/f525346bea4f56f9e57c3e21d81dc4d8896ee5f0/crate_universe/private/common_utils.bzl#L64-L82

Where if you follow through to
https://github.com/bazelbuild/rules_rust/blob/f525346bea4f56f9e57c3e21d81dc4d8896ee5f0/crate_universe/private/common_utils.bzl#L144-L153

You'll see here is where the isolated `CARGO_HOME` is defined. Perhaps this would be a good spot to link credentials? Admittedly, this code became super complicated in an effort to avoid making large changes but resulted in higher complexity and diverging code paths. But given that workspace support is on it's way out, I wouldn't be opposed to a bzlmod-only introduction of this feature. Do you think there's a path forward here to do symlinking in starlark?

ericmcbride: I can look into it!  I just got it working by passing isolated to the splicing cli.

ericmcbride: It looks like this wont be shareable via the module_ctx, but i think we can probably make the extension call the repository_rule downstream to keep the backwards compat?

ericmcbride: Okay I tried this and it does not work.  There would be no way to symlink with a module_ctx, without generating a new repo using repository_rules.  This could be a dummy repo maybe, but theres no way to symlink with a module_ctx.  I think thats why in the rust code, the symlink with the cargo config file was happening and why I went down the road I went dow.  @UebelAndre

crate_universe/private/common_utils.bzl
```diff
@@ -121,17 +121,6 @@ def get_rust_tools(repository_ctx, host_triple):
     Returns:
         struct: A struct containing the expected rust tools
     """
-
-    # This is a bit hidden but to ensure Cargo behaves consistently based
-    # on the user provided config file, the config file is installed into
-    # the `CARGO_HOME` path. This is done so here since fetching tools
-    # is expected to always occur before any subcommands are run.
-    if repository_ctx.attr.isolated and repository_ctx.attr.cargo_config:
```
ericmcbride, line 129: Deleting this allows the rust code used for MODULE.bazel to be the source of truth for both legacy WORKSPACEs and MODULE.bazel

PR #3386, opened 2025-04-04
fix: Duplicate aliases generated in rendering when two crates of the same version

ParkMyCar:
Original cause: https://github.com/bazelbuild/rules_rust/commit/5e904830eee39b37775855c6ff7e7e1079eb59e0#diff-52bdc04b030a02035a19a3c805288b7720191185a18f6b8f16f61a59baa42d3fR224
Related to: https://github.com/bazelbuild/rules_rust/pull/3215

This PR updates rendering in `cargo-bazel` to handle the case when there are multiple dependencies on a crate at the same version, but they're aliased. For example, when a `Cargo.toml` has the following:

```
itertools = "0.11.24"
itertools_other = { version = "0.11.24", package = "itertools" }
```

Previously this would generate two aliases both with the name `itertools-0.11.24` but now we only create a single alias of that name.

Encountering this case is odd but not uncommon, especially if you also use a tool like [`cargo-hakari`](https://github.com/guppy-rs/guppy/tree/main/tools/cargo-hakari) in your workspace.


UebelAndre (approved):
Thanks!

UebelAndre:
@ParkMyCar actually, for this can you also run [//util/repin_all.sh](https://github.com/bazelbuild/rules_rust/blob/8f118ee4f048f100c0c035e18acadeb2910b60dc/util/repin_all.sh) and push the results?

ParkMyCar:
> @ParkMyCar actually, for this can you also run [//util/repin_all.sh](https://github.com/bazelbuild/rules_rust/blob/8f118ee4f048f100c0c035e18acadeb2910b60dc/util/repin_all.sh) and push the results?
> 
> (sorry in advance for this being super slow 😅)

Ha no problem! I'll run it now :)

PR #3396, opened 2025-04-08
Dry run release workflow for release pull requests.

ograff:
Avoid failures of the release process by running everything except for the release creation and artifact upload when making or merging a PR that bumps the release version.


ograff:
We can see this running on #3397

PR #3398, opened 2025-04-08
Fix Compiling for release for windows

ograff:


UebelAndre (approved):
Nice!

UebelAndre (changes requested):
Actually, why not gate the entire release stage vs each step?

ograff:
Looks like windows isn't actually working (see #3397)

I'll move the builds to an intermediate stage and the release to a later one

.github/workflows/release.yaml
```diff
@@ -2,6 +2,12 @@
 name: Release
 on:
   workflow_dispatch:
+  merge_group:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - version.bzl
```
UebelAndre, line 10:
```suggestion
      - version.bzl
      - .github/workflows/release.yaml
```

UebelAndre (changes requested):

crate_universe/tools/cross_installer/src/main.rs
```diff
@@ -22,7 +23,13 @@ struct Options {
 fn prepare_workspace(workspace_root: &Path) {
     // Unfortunately, cross runs into issues when cross compiling incrementally.
     // To avoid this, the workspace must be cleaned
-    let cargo = env::current_dir().unwrap().join(env!("CARGO"));
+    let r = Runfiles::create().unwrap();
+    let cargo = r.rlocation_from(env!("CARGO"), "rules_rust").unwrap();
```
UebelAndre, line 27:
```suggestion
    let cargo = rlocation!(r, env!("CARGO")).unwrap();
```

.github/workflows/release.yaml
```diff
@@ -20,18 +27,6 @@ jobs:
...
-        run: |
-          git fetch origin &> /dev/null
-          branch="$(git rev-parse --abbrev-ref HEAD)"
-          if [[ "${branch}" != "main" ]]; then
-            echo "The release branch must be main. Got '${branch}'' instead." >&2
-            exit 1
-          else
-            echo "Branch is '${branch}'"
-          fi
       - name: Ensure release does not already exist
```
UebelAndre, line 30:
```suggestion
      - name: Ensure release does not already exist
        if: startsWith(github.ref, 'refs/heads/main') == false
```

UebelAndre (changes requested):
I think https://github.com/bazelbuild/rules_rust/pull/3398/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR96 should just be updated to
```
echo "startup --output_user_root=C:/tmp" > ./user.bazelrc
echo "startup --windows_enable_symlinks" > ./user.bazelrc
echo "build --enable_runfiles" > ./user.bazelrc
```
There are other issues that will prevent the use of runfiles from working on windows.

UebelAndre (approved):

PR #3401, opened 2025-04-08
Update crate universe to use musl binaries on Linux

UebelAndre:
This change makes `musl` binaries available for all supported linux platforms.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -28,7 +28,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Ensure release does not already exist
-        if: startsWith(github.ref, 'refs/heads/main') == false
+        if: startsWith(github.ref, 'refs/heads/main')
         run: |
           git fetch origin &> /dev/null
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
Interesting, this will probably change the libc we link against for the glibc prebuilds here, which may break some users... We may want to default to `musl` so that we don't have a glibc dependency for most linux users? (Which I thought we did, but it looks like we don't: https://github.com/bazelbuild/rules_rust/blob/a373426f39b6168bc839f34bd3680de5764b9c82/rust/platform/triple.bzl#L139)

.github/workflows/release.yaml
```diff
@@ -28,7 +27,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Ensure release does not already exist
-        if: startsWith(github.ref, 'refs/heads/main') == false
+        if: startsWith(github.ref, 'refs/heads/main')
```
illicitonion, line 30: This looks like it's inverting the condition?

illicitonion (approved):
LGTM, thanks!

.github/workflows/release.yaml
UebelAndre: Yeah, this should only run on `main` as a release won't actually be cut on another branch.

PR #3402, opened 2025-04-09
Added granular interfaces for core rules

UebelAndre:
This matches what we see in other rules.


illicitonion (approved):
I'm not sure how common this is, but it's fine by me!

PR #3411, opened 2025-04-17
Fix broken links in docs

hofbi:
`crate_universe.md` does not exist anymore but is split into bzlmod or workspace. I updated the URLs/files so that you don't get a 404.


UebelAndre (approved):
Thank you!

PR #3414, opened 2025-04-23
Fix links in documentation targeting github

hofbi:
Relative paths don't work in the deployed documentation since they end up in URLs such as https://bazelbuild.github.io/examples/bzlmod/all_crate_deps

Not sure if there is a better way, but for now this at least fixes the links.


UebelAndre (approved):
Thanks!

Issue #2889, opened 2024-09-19
Prost Crate Naming Conventions


purkhusid:
Wanted to start a discussion around the current naming conventions in the prost rules provided by rules_rust.

The prost rules set the crate name to the name of the proto_library  that is being compiled.
This approach is very prone to clashes e.g. the following folder structure:
```
/src
    /protobuf
        service1/
            /models
                models.proto
        service2/
            /models
                models.proto
```
In this folder structure the common way to name the proto_library  target would be models_proto (Which is the convention that gazelle uses) but that causes issues if there is a consumer that needs both service1/models  and service2/models .

I think it makes sense to change the naming conventions or make them configurable since this is a very likely issue when you have a repo with a large protobuf codebase.

Options:
* Change the naming convention to include the package path in the crate name
    * The example above would end with `src_protobuf_service1_models_models_proto`  and `src_protobuf_service2_models_models_proto` 
* Make the convention configurable on the prost toolchain
    * The toolchain could provide a `crate_naming_convention` which accepts `proto_name` and `package_name` where `proto_name` would keep the current convention but `package_name` would include the package path in the crate name

nhatlaso1:
I think this will help you.
https://bit.ly/4evXrQ6
I put the necessary dlls in the archive

purkhusid:
@UebelAndre You got any opinion on this?

purkhusid:
Another friendly ping @UebelAndre  @matts1 @illicitonion @freeformstu

matts1:
Sorry, missed this earlier. Isn't there a crate_name attribute for this exact reason? Could we not either use that for prost, or add the attribute if it doesn't exist to prost_library?

Don't really like the idea of passing a naming convention to the rule. It'd seem more appropriate to create a symbolic macro wrapping the rule to change the naming convention if a user didn't want to write crate_name all the time

Not particularly familiar with prost so may be missing something

purkhusid:
> Sorry, missed this earlier. Isn't there a crate_name attribute for this exact reason? Could we not either use that for prost, or add the attribute if it doesn't exist to prost_library?
> 
> Don't really like the idea of passing a naming convention to the rule. It'd seem more appropriate to create a symbolic macro wrapping the rule to change the naming convention if a user didn't want to write crate_name all the time
> 
> Not particularly familiar with prost so may be missing something

[...]

PR #3438, opened 2025-05-12
Allow renaming generated Prost crate name

purkhusid:
## What?
Adds `crate_name` to `rust_prost_transform` which allows overriding the generated crate name

## Why?
This allows the end user to control the name of the generated crates to avoid conflicts.

Closes Prost Crate Naming Conventions #2889


illicitonion (approved):
Seems reasonable! Thanks!

PR #3446, opened 2025-05-21
Rewrite relative paths in bin dirs (`-B`)

mccullocht:
With a hermetic gcc toolchain binary dirs (`-B`) may be necessary to invoke the C pre-processor.

I ran into this issue when using the [mongo gcc toolchain](https://github.com/mongodb/mongo/blob/master/bazel/toolchains/cc/mongo_linux/mongo_toolchain.BUILD.tmpl) where the packaging of the toolchain and relative bindir paths cause failures when compiling certain crates (namely `ring`).

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -193,6 +193,27 @@ def _pwd_flags_isystem(args):
 
     return res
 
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
Seems reasonable! Thanks!

PR #3455, opened 2025-05-27
Fix Windows CRLF Issue

timbess:
Cargo configs are injected as a literal string which, on Windows, can end up with escaped CRLF line endings in cargo-bazel.json causing the digests to be different on Linux/Windows. This parses the inner string literal to prevent platform specific differences in the `config.toml`. There might be a better way to do this, but adding a new top level command to cargo-bazel to parse TOML into JSON felt wrong, and adding a second binary seems even worse. I'm open to solving this another way if ya'll have a better idea.


timbess:
@UebelAndre Not sure who to ask about this, are there any changes required to get this merged? I'm just checking back in so we can hopefully delete our patch soon.

timbess:
@UebelAndre 

Previously the cargo config was stored as a `Value::String(... literal file contents)`. Parsing the inner `String` into a `Value` removes the platform specific line endings which caused Windows/Unix to have diverging digests. It would probably be cleaner to parse the TOML into JSON earlier on, but that's done in Starlark so it was a lot easier to handle this in Rust while parsing the config. I'm open to suggestions for better alternatives, but this does fix the issue on our Windows build.

timbess:
@UebelAndre Sorry to ping you again, but just wanted to check in and see if you had any thoughts.

UebelAndre (dismissed):
Thank you for the ping (I unfortunately need them)! Can you add a regression test for this? I suspect this is something that will change over time and with changes such as https://github.com/bazelbuild/bazel/pull/27659 might make doing things in starlark more efficient.

timbess:
That's cool! Native TOML parsing in Bazel would definitely make this much easier. I added a test that I confirmed fails to replicate the digest without my change.

timbess:
@UebelAndre Sorry forgot to tag you in my comment. I added a test to ensure Windows/Unix line endings don't change the digest.

PR #3464, opened 2025-06-06
Cargo Bazel: Expose `path` on `crate.spec()` and split off LocalPath from Path

rickvanprim:
Exposes `path` on `crate.spec()` (which just works with `cargo metadata` 🎉).

However if the `path` is outside of the Bazel workspace, it attempts to generate dependency paths in the form `//<absolute path>` (which don't exist), so this splits `SourceAnnotation::Path` into `SourceAnnotation::Path` (located somewhere on the file system) and `SourceAnnotation::LocalPath` (located relative to the Bazel workspace).

### DIFF
--- a/crate_universe/private/crate.bzl
+++ b/crate_universe/private/crate.bzl
@@ -27,7 +27,8 @@ def _spec(
         git = None,
         branch = None,
         tag = None,
-        rev = None):
+        rev = None,
+        path = None):
     """A constructor for a crate dependency.
 
     See [specifying dependencies][sd] in the Cargo book for more details.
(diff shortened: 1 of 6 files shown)


illicitonion (commented):
The `crate.bzl` change looks good to me, but I'm not quite sure why the `LocalPath` pieces are necessary, I wouldn't expect them to be and suspect we can fix something somewhere else to obviate the need - can you put together an integration test and we can poke at what goes wrong without `LocalPath`?

rickvanprim:
Implemented in #3596.

PR #3470, opened 2025-06-16
add rustdoc feature support

### LINKED ISSUE #511: rust_doc and rust_doc_test should support features
Currently these rules don't read features from the library and don't have the `crate_features` attr, so it's not possible to generate docs for a crate that requires at least one feature to be built.

TroyKomodo:
* Adds `crate_features` to be saved as part of `CrateInfo`
* Adds new flags for `rust_doc` and `rust_doc_test` to include the features of the crate when building the docs (enabled by default)

This is my first time using bazel so I am not 100% sure of my work, feedback greatly appreciated. 

fixes #511

### DIFF
--- a/rust/private/providers.bzl
+++ b/rust/private/providers.bzl
@@ -20,6 +20,7 @@ CrateInfo = provider(
         "aliases": "Dict[Label, String]: Renamed and aliased crates",
         "compile_data": "depset[File]: Compile data required by this crate.",
         "compile_data_targets": "depset[Label]: Compile data targets required by this crate.",
+        "crate_features": "[str]: A list of feature enabled for this crate.",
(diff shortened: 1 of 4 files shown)


TroyKomodo:
@illicitonion the error here looks spurious

illicitonion:
Thanks for the test!

I agree the failures look unrelated, unfortunately we have no way of bypassing or rerunning CI ourselves, but if you can push either empty commits or merge-main commits to this branch until things go green, that'd be super handy! (I'd do this myself, but I don't have permission to push to your fork)

illicitonion (approved):

illicitonion:
Oh there is a buildifier error though :)

PR #3474, opened 2025-06-17
Add resource_set attr to all expensive rules

illicitonion:
This allows better control over resource usage for expensive rules in resource-constrained systems.


sebas-zip:
this is an awesome feature. Thank you for implementing it. Do these documents get automagically updated when you change the API like this?
https://bazelbuild.github.io/rules_rust/rust.html

A doc test example would go a long way in explaining how to use it.

illicitonion:
> this is an awesome feature. Thank you for implementing it. Do these documents get automagically updated when you change the API like this? https://bazelbuild.github.io/rules_rust/rust.html
> 
> A doc test example would go a long way in explaining how to use it.

Yeah, they get updated on release :) with any luck we can get this merged soon and then release with https://github.com/bazelbuild/rules_rust/pull/3473

MODULE.bazel
```diff
@@ -16,6 +16,7 @@ bazel_dep(name = "rules_cc", version = "0.1.1")
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_shell", version = "0.3.0")
 bazel_dep(name = "apple_support", version = "1.17.1", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "aspect_bazel_lib", version = "2.19.4")
```
UebelAndre, line 19: Do we need to pull in this dependency? Could we instead add the attribute directly?

UebelAndre (changes requested):
Is this attribute not associated with a deprecated flag?
https://github.com/bazelbuild/bazel/issues/24934

illicitonion:
> Is this attribute not associated with a deprecated flag? [bazelbuild/bazel#24934](https://github.com/bazelbuild/bazel/issues/24934)

I followed up - the flag is deprecated because the feature is now always-on, so the flag can no longer be used to disable it :) deprecated-because-stabilised rather than deprecated-because-becoming-unsupported

MODULE.bazel
illicitonion: We could vendor this whole file if we preferred: https://github.com/bazel-contrib/bazel-lib/blob/main/lib/resource_sets.bzl - I don't have a strong preference, WDYT?

UebelAndre: We've already got a very similar file I think we should use instead:
https://github.com/bazelbuild/rules_rust/blob/7d96cde89db4f41df3f464bf7fdd4e4c2d1a6c05/rust/private/rustc_resource_set.bzl

(some refactoring may be required)

UebelAndre (changes requested):

PR #3478, opened 2025-06-22
Include repository name in splicing errors

### LINKED ISSUE #3460: Improve error message of missing labels/paths in splicer.rs
In case I would remove one of the lines in https://github.com/bazelbuild/rules_rust/blob/e38fa8c2bc0990debceaf28daa4fcb2c57dcdc1c/examples/crate_universe/MODULE.bazel#L413-L415 and run a `bazel mod tidy`, I will get an error message such as

```
Please add the following labels to the `manifests` key:
  //multi_package/sub_pkgs/pkg_b:Cargo.toml
```

If there is not only a single `MODULE.bazel` file but the codebase uses the [include feature](https://bazel.build/rules/lib/globals/module#include), it might take new users some time to find where this has to be added.

Is it possible to extend the error message with the file where this has to be added?

illicitonion:
Fixes #3460

### DIFF
--- a/crate_universe/extensions.bzl
+++ b/crate_universe/extensions.bzl
@@ -638,6 +638,7 @@ def _generate_hub_and_spokes(
             config_path = config_file,
             output_dir = tag_path.get_child("splicing-output"),
             debug_workspace_dir = tag_path.get_child("splicing-workspace"),
+            repository_name = cfg.name,
         )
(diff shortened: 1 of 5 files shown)


krasimirgg (approved):
(Not a crate_universe expert, but) looks good!

PR #3485, opened 2025-06-27
Allow setting timeout for cargo-bazel.

sardok:
This PR introduces `CARGO_BAZEL_TIMEOUT` env variable to set timeout value for `cargo-bazel` invocation.


illicitonion (commented):
LGTM, thanks!

Please can you add the env var to https://github.com/bazelbuild/rules_rust/blob/f04b2a6f9dc3f26e29563867155da8577b7203ad/crate_universe/private/generate_utils.bzl#L20 as well to make sure we'll invalidate and re-run the repository rules if the env var changes?

Thanks!

sardok:
Thanks @illicitonion for the review. I added the change you asked. Also updated the rule & extension docs about it (although i'm not sure if it is right thing to do because `CARGO_BAZEL_DEBUG` is missing there, maybe some env vars shouldn't be documented?).

illicitonion (approved):
Perfect, thanks!

PR #3487, opened 2025-06-30
Add riscv64gc-unknown-linux-gnu to tier 2 targets.

anp:
Per https://doc.rust-lang.org/rustc/platform-support/riscv64gc-unknown-linux-gnu.html this is now tier 2.

I've verified locally that with this patch I can run a rust test in qemu-riscv64 and have it pass.


illicitonion (approved):
Thanks!

illicitonion (approved):

anp:
It looks like the merge queue failed on a missing C++ header in a prost target, and I'm not sure how this might have been related. Is that a known flake by any chance?

PR #3489, opened 2025-07-02
Add pre-commit in CI and move some check over

hofbi:
As promised in #3484, I am adding pre-commit to CI. To better demonstrate what we can gain from this, I also move some formatting checks into pre-commit. This way, users can easily run exactly the same formatters as used in CI.


illicitonion (approved):
Thank you!

PR #3491, opened 2025-07-02
fix coverage generation

TroyKomodo:
Stop rustc from emitted `--codegen=instrument-coverage` when the target does not have coverage enabled via `--instrumentation_filter`

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -432,7 +432,7 @@ def _rust_test_impl(ctx):
         crate_info_dict = crate_info_dict,
         rust_flags = get_rust_test_flags(ctx.attr),
         skip_expanding_rustc_env = True,
-        include_coverage_runfiles = coverage,
+        include_coverage = coverage,
     )
     data = getattr(ctx.attr, "data", [])
 
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -833,7 +833,8 @@ def construct_arguments(
         use_json_output = False,
         build_metadata = False,
         force_depend_on_objects = False,
-        skip_expanding_rustc_env = False):
+        skip_expanding_rustc_env = False,
+        include_coverage = False):
     """Builds an Args object containing common rustc flags
 
     Args:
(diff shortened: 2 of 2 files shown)


TroyKomodo:
I am not actually sure why the unit tests dont pass, 

From looking at it
```bash
rustc \
	'test/unit/pipelined_compilation/bin.rs '--crate-name=bin' '--crate-type=bin' '--error-format=human' \
	'--out-dir=bazel-out/k8-fastbuild-ST-1c8a4433bbf5/bin/test/unit/pipelined_compilation' \
	'--codegen=opt-level=0' '--codegen=debuginfo=0' '--codegen=strip=none' '--remap-path-prefix=${pwd}=' \
	'--emit=link=bazel-out/k8-fastbuild-ST-1c8a4433bbf5/bin/test/unit/pipelined_compilation/bin' '--emit=dep-info'  \
	'--color=always' '--target=x86_64-unknown-linux-gnu'  \
	'-L rust_toolchain/lib/rustlib/x86_64-unknown-linux-gnu/lib' '--edition=2021' '-Cembed-bitcode=no' \
	'--codegen=linker=/usr/bin/gcc' '--codegen=link-arg=--coverage' '--codegen=link-arg=-fuse-ld=gold' \
	'--codegen=link-arg=-B/usr/bin' '--codegen=link-arg=-Wl,-no-as-needed' '--codegen=link-arg=-Wl,-z,relro,-z,now'  \
	'--codegen=link-arg=-pass-exit-codes' '--codegen=link-arg=-Wl,--push-state,-as-needed' '--codegen=link-arg=-lstdc++' \
	'--codegen=link-arg=-Wl,--pop-state' '--codegen=link-arg=-Wl,--push-state,-as-needed' '--codegen=link-arg=-lm' \
	'--codegen=link-arg=-Wl,--pop-state' \
	'--extern=second=./bin/test/unit/pipelined_compilation/libsecond-416833743.rlib' \
	'-Ldependency=bazel-out/k8-fastbuild-ST-1c8a4433bbf5/bin/test/unit/pipelined_compilation' \
	'-Ldependency=bazel-out/k8-opt-exec-ST-742c84c296cd/bin/test/unit/pipelined_compilation' \
	'--codegen=instrument-coverage' \
	'--sysroot=./rust_toolchain' \
```

```
error[E0463]: can't find crate for `second`
 --> test/unit/pipelined_compilation/bin.rs:1:5
  |
1 | use second::fun;
  |     ^^^^^^ can't find crate

```

Which doesn't make sense to me because we are clearly passing `--extern=second=...`

I also dont think my changes effect the dependency resolution graph so again I'm not sure why rustc is behaving this way, perhaps its showing an error for not finding the crate but the actual issue is it not finding the llvm library the crate needs for coverage?

PR #3492, opened 2025-07-02
Fix panic line numbers

### LINKED ISSUE #3476: Current implementation of remap_path_prefix breaks debugging on macOS
Using an empty string for remap_path_prefix breaks debug info on macOS. Remapping to empty string causes paths to become relative paths, which is not supported in all situations for all binary info on all platforms (e.g. OSO stabs in debug info on macOS). For more background on exactly how this breaks debug info, see https://github.com/rust-lang/rust/issues/132143.

Default value for `construct_arguments`, see:
https://github.com/bazelbuild/rules_rust/blob/0cb272d303ce7ea1dbe9bb5b16228f2008c8d72e/rust/private/rustc.bzl#L832

Remapping of `pwd` here:
https://github.com/bazelbuild/rules_rust/blob/0cb272d303ce7ea1dbe9bb5b16228f2008c8d72e/rust/private/rustc.bzl#L995

This can be fixed in `rules_rust` by not using an empty string for remapping here, at least on macOS, or for all platforms.

In terms of prior art here...

rules_swift uses `.` for remapping instead:

https://github.com/bazelbuild/rules_swift/blob/321bf2611163f562884545cbb7b183988bb4f367/swift/internal/feature_names.bzl#L62-L65

https://github.com/bazelbuild/rules_swift/blob/321bf2611163f562884545cbb7b183988bb4f367/tools/worker/swift_runner.cc#L293-L297

rules_go does as well:

https://github.com/bazel-contrib/rules_go/blob/1172e60b3fdb3c1b1f6efc781eda9819533375ac/go/tools/builders/cgo2.go#L330-L333

Bazel appears to use `/proc/self/cwd` at times, but outside of linux you need to manually remap this when debugging, with e.g.

```
(lldb) settings set target.source-map /proc/self/cwd <...>
```

Not sure which of these approaches would be best, or who would be the best to consult on this, but I think moving to one of the above is a good path forward.

illicitonion:
Without this patch, line numbers are missing for first-party code in panic traces.

This reverts part of #1563 but keeps the test showing we have a relative path not an absolute one, and adds an integration test showing that line numbers are correctly reported in panic traces.

As far as I know, path remapping has been improved in rustc over the last few years, so maybe the behaviour reported in that patch has been fixed.

Closes #3476


scentini (approved):

csmulhern:
I think this needs to be rebased and merged again.

PR #3510, opened 2025-07-18
Add support for extra rustc environment variables

nmattia:
This adds support for injecting environment variables into rustc invocations via build settings:

- `//rust/settings:extra_rustc_env` for the target configuration
- `//rust/settings:extra_exec_rustc_env` for the exec configuration

These settings can be set from the command line with flags like:

```
  --@rules_rust//rust/settings:extra_rustc_env=RUSTC_BOOTSTRAP=1
```

This mirrors how extra rustc flags are handled and enables build-wide env-based feature control without editing individual rules. Useful for enabling unstable features across many crates, or toggling behavior in proc macros and build scripts.


rust/private/rustc.bzl
```diff
@@ -2366,6 +2401,24 @@ extra_exec_rustc_flags = rule(
     build_setting = config.string_list(flag = True),
 )
 
+def _extra_exec_rustc_env_impl(ctx):
+    env_vars = {}
```
illicitonion: Can you extract what's common from these functions into a helper?

illicitonion (commented):
This LGTM, thanks! One small request :)

rust/private/rustc.bzl
nmattia: There you go! Let me know if you had something else in mind.

nmattia:
@illicitonion PTAL!

PR #3511, opened 2025-07-18
Adding select statements for build script environment and data attributes in bzlmod

finn-ball:
An implementation to bring some select statement functionality which are available in [repository rules](https://github.com/bazelbuild/rules_rust/blob/f04b2a6f9dc3f26e29563867155da8577b7203ad/crate_universe/private/crate.bzl#L258) but not bzlmod.

Since a dict of dicts is not supported, I've opted for a dict of json string which gets decoded on the other end.

### DIFF
--- a/crate_universe/extensions.bzl
+++ b/crate_universe/extensions.bzl
@@ -868,6 +868,22 @@ def _crate_impl(module_ctx):
         repo_specific_annotations = {}
         for annotation_tag in mod.tags.annotation:
             annotation_dict = structs.to_dict(annotation_tag)
+            if annotation_dict["build_script_data_select"]:
+                annotation_dict["build_script_data"] = struct(
+                    common = annotation_dict["build_script_data"],
+                    selects = annotation_dict["build_script_data_select"],
+                )
+                annotation_dict.pop("build_script_data_select")
+            if annotation_dict["build_script_env_select"]:
+                annotation_dict["build_script_env"] = struct(
+                    common = annotation_dict["build_script_env"],
+                    selects = {
+                        k: json.decode(v)
+                        for k, v in annotation_dict["build_script_env_select"].items()
+                    },
+                )
+                annotation_dict.pop("build_script_env_select")
+
             repositories = annotation_dict.pop("repositories")
             crate = annotation_dict.pop("crate")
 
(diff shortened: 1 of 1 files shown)


crate_universe/extensions.bzl
```diff
@@ -1088,12 +1104,18 @@ _annotation = tag_class(
...
+            doc = "A list of labels to add to a crate's `cargo_build_script::data` attribute. Keys should be the platform triplet. Value should be a list of labels.",
+        ),
         "build_script_deps": _relative_label_list(
             doc = "A list of labels to add to a crate's `cargo_build_script::deps` attribute.",
         ),
         "build_script_env": attr.string_dict(
             doc = "Additional environment variables to set on a crate's `cargo_build_script::env` attribute.",
         ),
+        "build_script_env_select": attr.string_dict(
+            doc = "Additional environment variables to set on a crate's `cargo_build_script::env` attribute. Key should be the platform triplet. Value should be a json string.",
```
illicitonion: Can you document the JSON string here? That it should be a dict of env var names -> env var values?

illicitonion (commented):
Looks reasonable to me - ideally we'd have more expressive annotations, or a helper function to generate this stuff, but this opens up capabilities we currently lack :) Thanks!

finn-ball:
> or a helper function to generate this stuff

What would a helper function look like in this scenario? I'm unsure how to add functions into `MODULE` files for this and we're quite limited in what we can pass into a tag class.

I had thought about adding a new tag class such as:
```
crate.annotation_env_select(
    crate = "foo",
    platform_triplet = "linux-x86_64-unknown",
    variables = {"FOO": "bar"},
)
crate.annotation_env_select(
    crate = "foo",
    platform_triplet = "linux-arm64-unknown",
    variables = {"FOO": "baz"},
)
```

But felt that was a little messy too as it isn't part of the original annotation.

Have you any suggestions?

illicitonion:
> > or a helper function to generate this stuff
> 
> What would a helper function look like in this scenario? I'm unsure how to add functions into `MODULE` files for this and we're quite limited in what we can pass into a tag class.

These aren't currently supported by Bazel, but there's perpetually discussion about them :) sorry, I was saying "ideally" as in "if Bazel were more capable" rather than "if Finn put in more work" :D

> I had thought about adding a new tag class such as:
> 
> ```
> crate.annotation_env_select(
>     crate = "foo",
>     platform_triplet = "linux-x86_64-unknown",
>     variables = {"FOO": "bar"},
> )
> crate.annotation_env_select(
>     crate = "foo",
>     platform_triplet = "linux-arm64-unknown",
>     variables = {"FOO": "baz"},
> )
> ```
> 
> But felt that was a little messy too as it isn't part of the original annotation.

This kind of new tag class is probably the nicest way to express this at the moment tbh, I wouldn't complain about this API at all. But also happy to merge this as-is if you don't fancy putting it together :)

finn-ball:
Merge as is for now and I'll consider the latter API.

hofbi:
@illicitonion can we go with this as is for now?

illicitonion (approved):
Weird, I thought I'd already merged this - sorry for the delay!

PR #3523, opened 2025-07-23
Optionally write the rendering lockfile when vendored

patrickmscott:
This lockfile is useful for tools like gazelle because it contains all the information about resolved crates. The regular cargo lockfile does not have enough details about each crate to properly identify dependencies.


illicitonion (approved):
Seems reasonable (albeit a little odd :))

PR #3524, opened 2025-07-24
Add setting `always_enable_metadata_output_groups`

refi64:
Right now, only library targets provide metadata-related output groups. This makes sense for compilation purposes, but building the metadata is also useful for other things; namely, building *only* the metadata is the fastest way to get diagnostic information for your workspace like `cargo check` does.

This PR adds a new setting, `always_enable_metadata_output_groups`, that enables both the `build_metadata` and `rustc_rmeta_output` output groups for non-library targets (such as binaries and tests). That means that the presence of metadata in `CrateInfo` can no longer be used to determine pipelind compilation support, so a new attribute, `metadata_supports_pipelining`, is added for that purpose.

As an example of the performance increase, compare a `bazel build` on a binary w/o `--output_groups=rustc_rmeta_output`:

```
INFO: Elapsed time: 2.247s, Critical Path: 2.11s
```

to with it:

```
INFO: Elapsed time: 1.034s, Critical Path: 0.90s
```


illicitonion (approved):
Thanks!

PR #3528, opened 2025-07-25
Add setting `clippy_output_diagnostics` to match `rustc_output_[...]`

refi64:
Right now, capturing clippy's diagnostics is in a weird place. `capture_clippy_output` exists, but it has drastically different behavior from `rustc_output_diagnostics` in that it captures the rendered error output and hides it from the terminal, whereas `rustc_output_diagnostics` captures the raw JSON and still prints out the rendered diagnostics.

The loss of the JSON is a bit of an issue. With that, the only way to get JSON output from the command-line is to set `clippy_error_format`, which will necessarily *clear the analysis cache* and rebuild all the targets, meaning that you can't get both rendered output *and* JSON output without either rebuilds in-between or relying on custom transitions.

This adds the missing equivalent to `rustc_output_diagnostics` in the form of a new option, `clippy_output_diagnostics`. In an ideal world, I'd imagine this would be the *only* option to capture clippy output (since for most intents and purposes, it's strictly a superset of `capture_clippy_output`), but using a separate option and output group preserves backwards compatibility while retaining the parallel naming of `rustc_output_diagnostics` vs `clippy_output_diagnostics`, as well as the output groups `rustc_output` vs `clippy_output`.


refi64:
```
(00:34:55) ERROR: /workdir/docs/BUILD.bazel:259:8: in deps attribute of starlark_doc_extract rule //:rust_settings.extract: missing bzl_library targets for Starlark module(s) @bazel_skylib//lib:structs.bzl. Since this rule was created by the macro 'stardoc', the error might have been caused by the macro implementation
(00:34:55) ERROR: /workdir/docs/BUILD.bazel:259:8: Analysis of target '//:rust_settings.extract' failed
(00:34:55) ERROR: Analysis of target '//:publish_book' failed; build aborted: Analysis failed
```

I seem to be constantly finding new ways to fail CI. (should be fixed now.)

illicitonion (approved):
Makes sense to me - thanks!

PR #3529, opened 2025-07-25
Add buildifier hook

hofbi:
Replace the buildifier CI call with the pre-commit hook. Now developers can run exactly the same formatting and linting as we use in CI

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -14,14 +14,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Buildifier
-        run: |
-          wget "https://github.com/bazelbuild/buildtools/releases/download/v${BUILDIFIER_VERSION}/buildifier-linux-amd64" -O buildifier
-          chmod +x ./buildifier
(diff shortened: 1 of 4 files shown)


UebelAndre (changes requested):
Thanks! I think adding a pre-commit hook is fine but I don't think we should disable the CI check otherwise folks thinks could be missed by folks who didn't run the pre-commit hook. Can you re-enable that in this change?

hofbi:
> Thanks! I think adding a pre-commit hook is fine but I don't think we should disable the CI check otherwise folks thinks could be missed by folks who didn't run the pre-commit hook. Can you re-enable that in this change?

Buildifier is still running in CI as part of pre-commit. If you scroll down to the end of the CI job you can see the pre-commit run --all-files. This will make sure that all formatters we maintain in pre-commit are also executed in CI.

If you look at [the CI logs of this branch](https://github.com/bazelbuild/rules_rust/actions/runs/16518699362/job/46715302617#step:7:1), you can see that buildifier is still executed.

illicitonion (approved):
Works for me!

hofbi:
@UebelAndre PTAL

PR #3531, opened 2025-07-25
Synchronize handling of C++ runfiles with C++ rules

hlopko:
This PR makes sure that Rust forwards runfiles from C++ rules exactly like C++ rules do (currently, Rust sends shared libraries to data_runfiles even when linking statically, C++ doesn't).

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1149,6 +1149,32 @@ def construct_arguments(
 
     return args, env
 
+
+def _get_dynamic_libraries_for_runtime(cc_linking_context, linking_statically):
+    libraries = []
+    if not cc_linking_context:
+        return libraries
+
+    for linker_input in cc_linking_context.linker_inputs.to_list():
+        libraries.extend(linker_input.libraries)
+
+    dynamic_libraries_for_runtime = []
+    for library in libraries:
+        artifact = _get_dynamic_library_for_runtime_or_none(library, linking_statically)
+        if artifact != None:
+            dynamic_libraries_for_runtime.append(artifact)
+
+    return dynamic_libraries_for_runtime
+
+def _get_dynamic_library_for_runtime_or_none(library, linking_statically):
+    if library.dynamic_library == None:
+        return None
+
+    if linking_statically and (library.static_library != None or library.pic_static_library != None):
+        return None
+
+    return library.dynamic_library
+
 def rustc_compile_action(
         *,
         ctx,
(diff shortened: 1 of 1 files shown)


rust/private/rustc.bzl
```diff
@@ -1503,6 +1529,29 @@ def rustc_compile_action(
...
+            runfiles_list.append(ctx.runfiles(transitive_files = data_dep[DefaultInfo].files))
+            runfiles_list.append(data_dep[DefaultInfo].default_runfiles)
+
+    for src in getattr(ctx.attr, "srcs", []):
+        runfiles_list.append(src[DefaultInfo].default_runfiles)
+
+    for dep in getattr(ctx.attr, "deps", []):
+        runfiles_list.append(dep[DefaultInfo].default_runfiles)
+
+        runfiles_list.append(dep[DefaultInfo].default_runfiles)
```
krasimirgg: this line looks like a repeat of the previous one.

rust/private/rustc.bzl
```diff
@@ -1149,6 +1149,32 @@ def construct_arguments(
 
     return args, env
 
+
+def _get_dynamic_libraries_for_runtime(cc_linking_context, linking_statically):
+    libraries = []
+    if not cc_linking_context:
+        return libraries
+
+    for linker_input in cc_linking_context.linker_inputs.to_list():
```
krasimirgg: naively: (1) here we're flattening the `linker_inputs` `depset`, which contains all the transitive bits (2) this function gets invoked for all `rust_library`-es. Doesn't that potentially cause quadratic overhead?

krasimirgg (changes requested):
also plz look into the formatting error and if possible add some unit tests.

rust/private/rustc.bzl
hlopko: Yep, and we and other rules do this quite often, unfortunately. rules_rust already do it in other places (https://github.com/bazelbuild/rules_rust/blob/7a0e526e3fed2aa7984433034cc999d99fb97fe6/rust/private/rustc.bzl#L261, https://github.com/bazelbuild/rules_rust/blob/7a0e526e3fed2aa7984433034cc999d99fb97fe6/rust/private/rustc.bzl#L1510.

rust/private/rustc.bzl
hlopko: Done.

rust/private/rustc.bzl
krasimirgg: That's unfortunate. I feel we should be addressing cases like this and not introducing more. The first instance is relatively recent, pinged the author for that. The second one is older, also looks kinda related to the dynamic libraries and runfiles features that this PR is expanding... so potentially may be addressed by a PR like this one. Would it be too much work and feasible to see if we can extend some of our providers a bit so that we can propagate these various types of runtime artifacts as depsets so we don't have to iterate over the linker inputs directly?

rust/private/rustc.bzl
```diff
@@ -1149,6 +1149,31 @@ def construct_arguments(
...
+
+    dynamic_libraries_for_runtime = []
+    for library in libraries:
+        artifact = _get_dynamic_library_for_runtime_or_none(library, linking_statically)
+        if artifact != None:
+            dynamic_libraries_for_runtime.append(artifact)
+
+    return dynamic_libraries_for_runtime
+
+def _get_dynamic_library_for_runtime_or_none(library, linking_statically):
```
UebelAndre: nit: this and/or add docstrings to make the types clearer 😅 
```suggestion
def _get_dynamic_library_for_runtime_or_none(library, is_linking_statically):
```

rust/private/rustc.bzl
krasimirgg: Chatted more with folks and I think we shouldn't block this on the depset flattening: even some of the c++ APIs that get replaced by explicit flattening here did flattening under the covers. We could in the future address the flattening in a more organized and targeted way in a separate effort.

krasimirgg (approved):
Looks good! Thank you!

Issue #3502, opened 2025-07-10
Add support for WASI Preview 2 and 3 platform selection


avrabe:
Currently, rules_rust cannot distinguish between different WASI preview versions (wasm32-wasip1, wasm32-wasip2, wasm32-wasip3) when selecting toolchains. All WASI variants map to identical platform constraints, making Bazel unable to select the appropriate toolchain for a specific WASI version.

**Error when trying to build for specific WASI versions:**
```
ERROR: no registered toolchain for wasm32-wasip2
```

## Current Behavior

All WASI targets (`wasm32-wasi`, `wasm32-wasip1`, `wasm32-wasip2`) resolve to the same platform constraints:
- `@platforms//cpu:wasm32`
- `@platforms//os:wasi`

This prevents proper toolchain selection and platform-specific builds.

## Expected Behavior

Each WASI version should have distinct platform constraints allowing for:
- Separate toolchain registration per WASI version
- Platform transitions targeting specific WASI versions
- Cross-compilation to different WASI preview versions

## Solution

Add version-specific platform constraints to distinguish WASI preview versions:

### 1. Platform Constraints
```starlark
# In rust/platform/BUILD.bazel
constraint_setting(
    name = "wasi_version",
    default_constraint_value = ":wasi_preview_1",
)

constraint_value(
    name = "wasi_preview_1",
    constraint_setting = ":wasi_version",
)

constraint_value(
    name = "wasi_preview_2", 
    constraint_setting = ":wasi_version",
)

constraint_value(
    name = "wasi_preview_3",
    constraint_setting = ":wasi_version",
)
```

### 2. Platform Definitions
```starlark
# In rust/platform/platform.bzl
platform(
    name = "wasip2",
    constraint_values = [
        "@platforms//cpu:wasm32",
        "@platforms//os:wasi", 
        "@rules_rust//rust/platform:wasi_preview_2",
    ],
)
```

### 3. Constraint Mappings
```starlark
# In rust/platform/triple_mappings.bzl
if target_triple == "wasm32-wasip2":
    return [
        "@platforms//cpu:wasm32",
        "@platforms//os:wasi",
        "@rules_rust//rust/platform:wasi_preview_2",
    ]
```

### Cross-Compilation Support
Fixed C++ toolchain integration issues where WASI SDK tools were not available during cross-compilation:

```starlark
# Include C++ toolchain files as action inputs
additional_inputs = []
if cc_toolchain and cc_toolchain.all_files:
[...]

PR #3539, opened 2025-07-31
Add known shas for wasip2 target

cameron-martin:
Support for the wasip2 target was added in #3502. This adds the target to the script that generates the known shas, and re-generates the file.


UebelAndre:
Thanks!

PR #3542, opened 2025-08-04
Include transitive out directories when running rustc

cameron-martin:
Previously only the out directories of build scripts that are direct dependencies were included when running rustc. This caused errors when linking when a transitive dependency's build script builds an archive into the out directory and adds a link argument to it via `cargo::rustc-link-lib`.

Now all out directories of transitive dependencies are collected into a depset and included as dependencies in the rustc compile action.

For some reason this test segfaults when running with `bazel coverage`. It runs locally fine, so I'm not sure what the issue is here. I've just disabled the test because I can't see how it'd be relevant to my change.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -66,10 +66,12 @@ default_rbe_targets: &default_rbe_targets
 default_macos_targets: &default_macos_targets
   - "--"
   - "//..."
+  # Not sure why this doesn't work on bazelci, it does locally on macOS.
+  - "-//test/unit/remap_path_prefix:integration_test"
 default_macos_coverage_targets: &default_macos_coverage_targets
   - "--"
   - "//..."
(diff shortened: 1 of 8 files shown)


cameron-martin:
I have _no_ idea why the test is segfaulting on OSX. I've ran this locally on an arm64 mac and it works fine.

I'm tempted to switch this to a [build_test](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/build_test_doc.md) instead and not run it, because whether the test runs is inconsequential to this change.

cameron-martin:
@illicitonion @UebelAndre Is there anything I can do to get this merged?

illicitonion:
I believe https://github.com/bazelbuild/rules_rust/pull/3579 had the same desired outcome as this (sorry, I didn't see your PR when putting that together) - does that solve your use-case?

cameron-martin:
This seems to work, thanks. I'll close this.

PR #3543, opened 2025-08-04
Support rustdoc_flags parameter for rust_doc_test similar to rust_doc

max-heller:
Mirrors the implementation in the `rust_doc` rule


illicitonion (approved):
Thanks!

Issue #3482, opened 2025-06-25
bindgen_toolchain example fails with rules_cc 0.1.2


mxpv:
It seems that `bindgen_toolchain` example doesn't work with latest `rules_cc` published on BCR on MacOS.

Everything works fine with rules_cc versions up to 0.1.1:

```
rules_rust/examples/bindgen_toolchain on  main [!?] on ☁️  mpavlenko@nvidia.com
✦ ❯ bazelisk build //test/integration:simple_bindgen
INFO: Analyzed target //test/integration:simple_bindgen (203 packages loaded, 5384 targets configured).
INFO: Found 1 target...
Target //test/integration:simple_bindgen up-to-date:
  bazel-bin/test/integration/libsimple_bindgen-3348090008.rlib
INFO: Elapsed time: 45.921s, Critical Path: 42.62s
INFO: 315 processes: 139 internal, 176 darwin-sandbox.
INFO: Build completed successfully, 315 total actions
```

However once I updated `bazel_dep(name = "rules_cc", version = "0.1.2")`, it fails with:

```
NFO: Analyzed target //test/integration:simple_bindgen (204 packages loaded, 5388 targets configured).
ERROR: /private/var/tmp/_bazel_mpavlenko/68be9dfd15abd4f7b001bfb30d77f4cf/external/rules_rust+/util/process_wrapper/BUILD.bazel:31:36: Compiling Rust (without process_wrapper) bin process_wrapper (6 files) [for tool] failed: (Exit 1): bootstrap_process_wrapper.sh failed: error executing Rustc command (from target @@rules_rust+//util/process_wrapper:process_wrapper) bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/external/rules_rust+/util/process_wrapper/bootstrap_process_wrapper.sh -- ... (remaining 30 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
error: linking with `external/rules_cc++cc_configure_extension+local_config_cc/cc_wrapper.sh` failed: exit status: 127
  |
[...]

PR #3556, opened 2025-08-14
Use default env on rustc compile actions to allow bash scripts for linker

aaliddell:
See https://github.com/bazelbuild/rules_rust/issues/3482 for context.

I accept that this may not be merged as-is, as using `use_default_shell_env` opens the sandbox to other env vars affecting caching etc. However, getting `PATH` into the sandbox another way requires a bit more discussion, which I hope we can have here and I can amend PR.

Possible alternatives:
- The default env coming from `ctx.configuration.default_shell_env` is empty, so that's no use
- The upstream rules_cc could set the appropriate env var on the toolchain, which is then inherited by `cc_common.get_environment_variables()` for the linker
- Revert https://github.com/bazelbuild/rules_cc/commit/d74915024017250e46d95e91a3defc34174effe0 and address the path search in the 'Bazel' way that doesn't require PATH i.e use `repository_ctx.which(bash)` to embed the correct absolute path to bash into cc_wrapper.sh, then no changes are required here potentially

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1383,6 +1383,7 @@ def rustc_compile_action(
             inputs = compile_inputs,
             outputs = action_outputs,
             env = env,
+            use_default_shell_env = True,
             arguments = args.all,
             mnemonic = "Rustc",
             progress_message = "Compiling Rust {} {}{} ({} files)".format(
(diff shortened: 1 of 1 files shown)


aaliddell:
I missed the discussion in #3535 (searched issues but not PRs :man_facepalming:); close this if you wish.

aaliddell:
Closing as replaced by https://github.com/bazelbuild/rules_cc/pull/466

PR #3561, opened 2025-08-21
Add `//tools/upstream_wrapper:rustdoc`

lukasoyen:
For completeness sake. With this we can add `rustdoc` to a `bazel_env` in the same way we do for the other tools.


UebelAndre (approved):
Seems reasonable to me!

PR #3563, opened 2025-08-22
Fix NoneType exception for interface dependencies.

UebelAndre:
Before this change any dependency with `CcInfo` which did not contain static or dynamic libraries (e.g. an interface library which has it's own fields for storing the underlying files) would result in the following error:
```
ERROR: /Users/user/Code/rules_rust/test/unit/cc_info/BUILD.bazel:4:19: in rust_shared_library rule //test/unit/cc_info:rust_dylib_with_interface_lib_dep:
	File "/Users/user/Code/rules_rust/rust/private/rust.bzl", line 126, column 32, in _rust_shared_library_impl
		return _rust_library_common(ctx, "cdylib")
	File "/Users/user/Code/rules_rust/rust/private/rust.bzl", line 190, column 32, in _rust_library_common
		return rustc_compile_action(
	File "/Users/user/Code/rules_rust/rust/private/rustc.bzl", line 1541, column 41, in rustc_compile_action
		dynamic_libraries = ctx.runfiles(files = [
Error in runfiles: at index 0 of files, got element of type NoneType, want File
```

This pull request fixes this exception.


illicitonion (approved):
Thanks!

PR #3569, opened 2025-08-25
Use ctx.watch properly for crate_universe

dzbarsky:
I don't think rctx.read did what the comment was trying to do. 

https://bazel.build/rules/lib/builtins/repository_ctx#read says the watch parameter "**default is 'auto'**...Can be the string 'yes', 'no', or 'auto'. Passing 'yes' is equivalent to immediately invoking the [watch()](https://bazel.build/rules/lib/builtins/repository_ctx#watch) method; passing 'no' does not attempt to watch the file; **passing 'auto' will only attempt to watch the file when it is legal to do so** (see watch() docs for more information."

https://bazel.build/rules/lib/builtins/repository_ctx#watch says "Note that attempting to watch paths inside the repo currently being fetched, or inside the working directory of the current module extension, will result in an error. **A module extension attempting to watch a path outside the current Bazel workspace will also result in an error.** "

This means that paths outside the workspace were already not getting watched, so we can filter them out and then use `.watch` instead of `.read`


UebelAndre (approved):
Thank you!

Issue #3515, opened 2025-07-21
Do not scan for `Cargo.toml` files unless explicitly requested.


guw:
When a user has listed a finite, static set of `Cargo.toml` files in a `bzlmod` setup, `rules_rust` still seems to scan for `Cargo.toml` files quite aggressively when initializing the rules. Per [this comment](https://github.com/bazelbuild/rules_rust/discussions/3508#discussioncomment-13826969) there seems to be a requirement for identifying new `Cargo.toml` files. Although this is an understandable requirement, its current implementation seems to violate a few assumptions/best practices how Bazel works.

Scanning of new files should be an explicit action like `bazel mod tidy` or a specific Gazelle invocation. That's how other rules do it and it is the recommend best practice, i.e. that's where a user would expect Bazel (and rules) to do such work. But not during regular invocation for a `build` or `test` or `query`. 

_Original Discussion:_

I am setting up rules_rust in a bzlmod workspace.

In my `MODULE.bazel` I list the `Cargo.toml` files explicitly.

```
# Rust toolchain
bazel_dep(name = "rules_rust", version = "0.61.0")

rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
rust.toolchain(
    edition = "2024",
    versions = ["1.85.0"],
)

# Crate Universe dependencies
crate = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
crate.from_cargo(
    name = "crates",
    cargo_lockfile = "//:Cargo.lock",
    manifests = [
        "//:Cargo.toml",
        "//foo/cli:Cargo.toml",
        "//bar/lib:Cargo.toml",
    ],
    supported_platform_triples = [
        "aarch64-apple-darwin",
        "aarch64-unknown-linux-gnu",
        "x86_64-unknown-linux-gnu",
    ],
)
use_repo(crate, "crates")
```

When I run `bazel mod deps` (or some other Bazel command) it fails with:

```
Error: Failed to splice workspace

Caused by:
    0: Failed to walk filesystem finding workspace Cargo.toml files
    1: IO error for operation on ../data/postgres/16: Permission denied (os error 13)
    2: Permission denied (os error 13)
```

This is expected. The `data` directory is listed in `.bazelignore`. It's also not listed in `Cargo.toml`.

Why is `rules_rust` trying to scan for `Cargo.toml` files? I listed all I want explicitly in the `manifests` list. Is there a way to exclude the `data` folder from scanning?

[...]

Issue #3425, opened 2025-04-29
.bazelignore not respected by cargo splice command


kgreenek:
I have the following directory structure:

```
- .bazelignore (excludes "venv")
- venv
    - foo
        - Cargo.toml
- my_package
    - Cargo.toml
```

The venv dir is in my .bazelignore file and .gitignore file. It's used for python development.

It appears the .bazelignore file is not respected by rules_rust.

The Cargo.toml file that is packaged with the foo python package in that venv dir is picked up by cargo splice, and cause the bazel run to fail.

Here's an example of the error I get:

```
INFO: Invocation ID: f9c1d480-0374-4424-b9a3-c67af4f43cfb
Error: Failed to splice workspace

Caused by:
    Failed to read manifest at /.../venv/lib/python3.10/site-packages/rust/cryptography-x509/Cargo.toml
ERROR: /home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/rules_rust~/crate_universe/private/common_utils.bzl:55:13: Traceback (most recent call last):
	File "/home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/rules_rust~/crate_universe/extensions.bzl", line 1013, column 37, in _crate_impl
		_generate_hub_and_spokes(
	File "/home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/rules_rust~/crate_universe/extensions.bzl", line 633, column 51, in _generate_hub_and_spokes
		splice_outputs = splice_workspace_manifest(
	File "/home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/rules_rust~/crate_universe/private/splicing_utils.bzl", line 174, column 19, in splice_workspace_manifest
		cargo_bazel_fn(
	File "/home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/rules_rust~/crate_universe/private/common_utils.bzl", line 88, column 23, in _execute
		return execute(
	File "/home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/rules_rust~/crate_universe/private/common_utils.bzl", line 55, column 13, in execute
		fail(_EXECUTE_ERROR_MESSAGE.format(
[...]

PR #3570, opened 2025-08-25
Speedup manifest splicing by removing workspace discovery logic

dzbarsky:
The original implementation required users to specify all manifests belonging to the workspace, and had a validation pass that walked the repo to verify the list was in sync. This PR swaps that for a `cargo metadata` call, which lists the manifests making up the workspace. We can then call `ctx.watch` on them to invalidate the extension when they are modified (which was previously impossible if they were not passed to the extension as labels). This also means users only need to specify the workspace-root Cargo.toml now :)

There is still an edge-case where a newly-added Cargo.toml won't trigger invalidation if it's covered by one of the wildcard `members` patterns, but that was already a pre-existing issue. In practice, we'd expect invalidation in that case to be triggered by a Cargo.lock change, or to be unnecessary if the crate does not add new deps to crate resolution.

Fixes https://github.com/bazelbuild/rules_rust/issues/3425
Fixes https://github.com/bazelbuild/rules_rust/issues/3515
Fixes https://github.com/bazelbuild/rules_rust/issues/3518

Obsoletes https://github.com/bazelbuild/rules_rust/pull/3387


vtsao-openai (approved):

crate_universe/src/cli/splice.rs
```diff
@@ -147,5 +152,47 @@ pub fn splice(opt: SpliceOptions) -> Result<()> {
...
+        let output = child.wait_with_output().with_context(|| {
+            format!(
+                "Error running `cargo metadata --no-deps` manifest path '{}'",
+                path
+            )
+        })?;
+        if !output.status.success() {
+            tracing::error!("{}", String::from_utf8_lossy(&output.stdout));
+            tracing::error!("{}", String::from_utf8_lossy(&output.stderr));
+            bail!(format!("Failed to run cargo tree: {}", output.status))
```
alexeagle: should be `metadata` rather than `tree`?

crate_universe/extensions.bzl
```diff
@@ -658,8 +661,10 @@ def _generate_hub_and_spokes(
             "metadata": splice_outputs.metadata,
         })
 
-    # The workspace root when one is explicitly provided.
-    nonhermetic_root_bazel_workspace_dir = module_ctx.path(Label("@@//:MODULE.bazel")).dirname
+        for path_to_track in splice_outputs.extra_paths_to_track:
+            # We can only watch paths in our workspace.
```
illicitonion, line 665: I believe this is what the old `module_ctx.read` stuff was for - it seems to successfully track external paths (or at least, not error on them, but IIRC I checked it seemed to work) whereas `module_ctx.watch` fails?

crate_universe/src/cli/splice.rs
```diff
@@ -147,5 +152,47 @@ pub fn splice(opt: SpliceOptions) -> Result<()> {
     std::fs::copy(cargo_lockfile_path, output_dir.join("Cargo.lock"))
         .context("Failed to copy lockfile")?;
 
+    if let SplicerKind::Workspace { path, .. } = prepared_splicer {
+        println!("MANIFEST {}", path);
+
```
illicitonion:
```suggestion
```

crate_universe/src/cli/splice.rs
```diff
@@ -81,13 +85,14 @@ pub fn splice(opt: SpliceOptions) -> Result<()> {
     };
 
     // Generate a splicer for creating a Cargo workspace manifest
-    let splicer = Splicer::new(splicing_dir, splicing_manifest)?;
+    let splicer = Splicer::new(splicing_dir.clone(), splicing_manifest)?;
+    let prepared_splicer = splicer.prepare()?;
```
illicitonion, line 88: It doesn't feel like `Splicer` is worth existing, if all it's used for is to construct a builder to turn into a `SplicerKind` - how would you feel about inlining `Splicer::new` into `SplicerKind::new`?

crate_universe/src/cli/splice.rs
```diff
@@ -147,5 +152,47 @@ pub fn splice(opt: SpliceOptions) -> Result<()> {
...
+        })?;
+        if !output.status.success() {
+            tracing::error!("{}", String::from_utf8_lossy(&output.stdout));
+            tracing::error!("{}", String::from_utf8_lossy(&output.stderr));
+            bail!(format!("Failed to run cargo metadata: {}", output.status))
+        }
+
+        let metadata: CargoMetadata =
+            serde_json::from_str(&String::from_utf8_lossy(&output.stdout))
+                .context("Unable to deserialize Cargo metadata")?;
```
illicitonion: Is there are reason not to:
```suggestion
        let metadata = cargo.metadata_command_with_options(path, vec![String::from("--no-deps")]).exec().with_context(...)?;`
```

crate_universe/src/splicing/splicer.rs
```diff
@@ -43,57 +41,38 @@ pub(crate) enum SplicerKind<'a> {
...
-                },
-            )
-        }
-
         if let Some((path, manifest)) = workspace_roots
             .iter()
             .next()
             .and_then(|path| manifests.get_key_value(path))
         {
+            if manifests.len() > 1 {
```
illicitonion, line 73: I think we hide stderr by default from cargo-bazel subprocesses (https://github.com/bazelbuild/rules_rust/blob/31d8c8a1c700dfc82cf9b1bdf289755d384123d0/crate_universe/private/common_utils.bzl#L109)

This doesn't necessarily need to change, but FYI this useful INFO is unlikely to be seen. If we wanted, we could either stop hiding stderr by default, or have starlark do this printing.

illicitonion (commented):
This looks great, thank you! I have a few small comments, otherwise let's get this shipped! 🎉

crate_universe/src/cli/splice.rs
illicitonion: Oh also feel free to not act on this - I can see the lifetimes here probably make this fiddly, we can pick this up separately if useful :)

crate_universe/extensions.bzl
dzbarsky: see explanation in https://github.com/bazelbuild/rules_rust/pull/3569 - I think it never worked correctly (or the docs are wrong?). I bet @fmeum knows the answer?

crate_universe/src/cli/splice.rs
dzbarsky: yeah, the lifetimes were annoying. I'll give it another shot but if I can't figure it out after a bit of poking I'll punt on this bit for now

crate_universe/src/splicing/splicer.rs
dzbarsky: hmmm I think I did see it show up in my testing, but i'll double check

crate_universe/src/cli/splice.rs
dzbarsky: it drops the env vars but they are probably not actually necessary, i'll give it a shot, thanks

crate_universe/src/cli/splice.rs
dzbarsky: thanks!

crate_universe/src/cli/splice.rs
illicitonion: If the env vars are useful, let's try to set them everywhere :) `Cargo` already has a `rustc_path` field, so we should be able to unconditionally set `RUSTC_WRAPPER` if needed. I've not seen `CARGO_CACHE_RUSTC_INFO` before, but if it's worth setting for this call, it's presumably worth setting for others too?

dzbarsky: I cribbed [from here](https://github.com/bazelbuild/rules_rust/blob/d7ab9ba84184961b30e68cb4c186c0a3fdad197b/crate_universe/src/metadata/cargo_tree_resolver.rs#L155-L159) but I suspect we don't need it since we're not trying to do platform-specific resolution

crate_universe/src/splicing/splicer.rs
zbarsky-openai: image

zbarsky-openai: I think between this and release notes we can nudge folks to remove them. Continuing to list them doesn't really hurt either. Happy to reword this if you have any suggestions though!

illicitonion (approved):
Perfect, thank you!

crate_universe/extensions.bzl
fmeum: Read didn't error because it defaults to auto, but also won't watch.

PR #3577, opened 2025-08-29
Fix typos in rust tools and util

hofbi:
Fix more typos, bump the typos hook to the latest version available, and exclude `diff` types (we don't want to change patches).


UebelAndre (approved):
Nice! Thank you!

PR #3581, opened 2025-08-31
Reorder dict keys

illicitonion:
I'm not sure how this got past CI, but hey...


illicitonion:
Reimplemented in #3583

PR #3593, opened 2025-09-04
[cargo-bazel] Avoid regenerating cargo lockfile when splicing is not needed

dzbarsky:
If we already have a Cargo.lock and don't intend to let cargo-bazel update it, we can just use it for splicing directly.


illicitonion (approved):
Makes sense!

PR #3594, opened 2025-09-06
RFC: support proc_macro_deps in deps

dzbarsky:
This is accomplished via a `rust_proc_macro` macro that instantiates a `rust_proc_macro` proxy rule which forwards relevant providers from an (inner) `rust_proc_macro` rule. The wrapper rule takes all the same attributes (even though they are ignored) and the rules are named the same to try to avoid any churn for aspects.


dzbarsky:
Let's take https://github.com/bazelbuild/rules_rust/pull/3605 instead

PR #3595, opened 2025-09-06
RFC: proc_macro self-transition

dzbarsky:


rust/private/rust.bzl
```diff
@@ -1072,6 +1048,21 @@ rust_shared_library = rule(
         """),
 )
 
+def _exec_transition_impl(_settings, attr):
+    return {
+        "//command_line_option:is exec configuration": not attr.internal_doctest_transition_override,
```
dzbarsky, line 1053: wasn't sure if there was a cleaner way to do this, hacking around a failure in the doctest

```
Executing tests from //test/unit/rustdoc:rustdoc_proc_macro_doctest
error[E0463]: can't find crate for `std`
  |
  = note: the `aarch64-apple-darwin` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-apple-darwin`

For more information about this error, try `rustc --explain E0463`.
```

dzbarsky:
Let's take https://github.com/bazelbuild/rules_rust/pull/3605 instead

PR #3600, opened 2025-09-09
Add gcc-ld to rustc_lib target

timothyg-stripe:
This makes it possible to use the self-contained `rust-lld` linker through the gcc driver, which will be the default setup in [Rust 1.90.0](https://blog.rust-lang.org/2024/05/17/enabling-rust-lld-on-linux/).


timothyg-stripe:
@UebelAndre is this something you can review?

timothyg-stripe:
@UebelAndre This PR doesn't ensure they'll be used, merely that it becomes _possible_ to use them. It's still useful because folks who want to enable gcc-ld in their own setup can now do so with a configuration.

PR #3601, opened 2025-09-09
rust_repository_set: Allow extra_exec_rustc_flags to be a dict

timothyg-stripe:
This makes `extra_exec_rustc_flags` consistent with `extra_rustc_flags`


timothyg-stripe:
@UebelAndre is this something you can review?

PR #3603, opened 2025-09-09
Add support for collecting a target's cfg flags into CrateInfo.

bmoxb:
This change add a new `cfgs` field to `CrateInfo`. When setting `//rust/settings:collect_cfgs` is set, all cfg flags for a target will be collected into the `cfgs` field.


krasimirgg (approved):
Thank you!

UebelAndre:
@krasimirgg can you make the format check a required job?

PR #3605, opened 2025-09-09
Improve proc_macro_deps ergonomics, take 3

dzbarsky:
I kept the `proc_macro_deps` attribute named the same just in case, but it would be nice to rename to `exec_configured_deps` at some point to reflect reality.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -22,6 +22,9 @@ load(
     "AllocatorLibrariesImplInfo",
     "AllocatorLibrariesInfo",
     "BuildInfo",
+    "CrateInfo",
+    "CrateGroupInfo",
+    "DepInfo",
     "LintsInfo",
 )
 load("//rust/private:rustc.bzl", "rustc_compile_action")
@@ -35,6 +38,7 @@ load(
     "determine_lib_name",
     "determine_output_hash",
     "expand_dict_value_locations",
+    "filter_deps",
     "find_toolchain",
     "generate_output_diagnostics",
     "get_edition",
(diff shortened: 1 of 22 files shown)


rust/private/rust.bzl
```diff
@@ -526,16 +504,16 @@ def _rust_library_group_impl(ctx):
     runfiles = []
 
     for dep in ctx.attr.deps:
-        if rust_common.crate_info in dep:
+        if CrateInfo in dep:
             dep_variant_infos.append(rust_common.dep_variant_info(
-                crate_info = dep[rust_common.crate_info] if rust_common.crate_info in dep else None,
-                dep_info = dep[rust_common.dep_info] if rust_common.crate_info in dep else None,
+                crate_info = dep[CrateInfo] if CrateInfo in dep else None,
+                dep_info = dep[DepInfo] if DepInfo in dep else None,
```
dzbarsky, line 510: cleaned up provider references while I was here and discovered `dep_info` was cheking for the CrateInfo provider but using DepInfo one :)

thesayyn:
I love this.

UebelAndre (changes requested):
How does this handle transitioning `proc_macro_deps` to `exec`?
https://github.com/bazelbuild/rules_rust/blob/12762a74c8847d9a34e1b5f3c73dd84a32212453/rust/private/rust.bzl#L726-L736

dzbarsky:
> How does this handle transitioning `proc_macro_deps` to `exec`?
> 
> https://github.com/bazelbuild/rules_rust/blob/12762a74c8847d9a34e1b5f3c73dd84a32212453/rust/private/rust.bzl#L726-L736

That line you linked is still there, that's how it works :). Basically the wrapper macro sends all deps to both dep (target configured) and proc_macro_deps (exec configured) on the underlying rules, and then 'filter_deps' discards the unnecessary ones.

The first commit is the best one to look at, the rest are fairly mechanical

UebelAndre:
> That line you linked is still there, that's how it works :). Basically the wrapper macro sends all deps to both dep (target configured) and proc_macro_deps (exec configured) on the underlying rules, and then 'filter_deps' discards the unnecessary ones.
> 
> The first commit is the best one to look at, the rest are fairly mechanical

Ah, I see now. This directionally seems wrong (even though I see the ergonomics of it). Will this impact `query` and `cquery` results? What's the impact on analysis to be duplicating dependencies? And what inspired this? Is it actually an ergonomics issue?

dzbarsky:
>  Will this impact `query` and `cquery` results? 

`query` results are unaffected, `cquery` will show the deps in both configs now. That does seem a tad bit unfortunate - if we think it's a blocker perhaps the wrapper macro could only perform the expansion logic if an extra param (defaulted to False) were passed, and otherwise was a no-op passthrough. Would that be more palatable?

> What's the impact on analysis to be duplicating dependencies?

I wasn't able to measure any significant difference in our repo; there is very minimal processing of the extra deps and rules_rust already does a ton of expensive analysis work so it's lost in the noise

> And what inspired this? Is it actually an ergonomics issue?

I'm prototyping some cargo-related repository rules / module extensions and it's hard to emit the `deps` for crates based on metadata sources at module extension time, since we have no way of knowing whether a crate is a proc-macro until we download the crate source. With this PR, we are able to use the metadata from the crates registry to lay out the dependency tree and defer the decision of whether a given crate generates a `rust_library` or a `rust_proc_macro` until the crate's own repository rule runs.  Without it, we were [maintaining an allowlist ](https://github.com/dzbarsky/rules_rs/blob/main/rs/extensions.bzl#L13-L138), which I'd classify as unergonomic :)

https://bazelbuild.slack.com/archives/CSV56UT0F/p1726290118173529 also brought up the point that needing to treat proc_macros specially also complicated BUILD generation logic

> This directionally seems wrong

Could you elaborate more on this? To me it feels good to not force every consumer to categorize deps into the right slots, after all we allow CcInfo-carrying targets into `deps` instead of forcing them to be specified as `cdeps`.  cc @fmeum as he suggested this approach

PR #3608, opened 2025-09-11
Ensure the library search path from cc_toolchain is preferred over whatever rustc finds in the Windows registry

### LINKED ISSUE #3256: rustc uses non-hermetic libs/paths on Windows
The rust compiler doesn't use the path of the linker to determine where it should look for libraries/other paths. 
On windows if you don't specify the environmental variables from vcvars it will use the registry to guess what version of VS you want to use. 

If you have VC 17.10 and VC10.12 installed and compile rust with your C++ toolchain configured to VC17.10 the linker path passed in to the rust compiler is that from VC17.10, however rustc looks in the registry and finds 17.12 installed and provides the standard libs from 17.12 instead. 

You will then notice that features which are highly coupled to the linker/lib combinations stop working, for example asan. 

See:
https://docs.rs/cc/latest/src/cc/windows/find_tools.rs.html#1190-1191

The workaround is to set `VSINSTALLDIR` and `VCINSTALLDIR` in the env.

The main problem from rust/rules/linker combination is this can appear to be working correctly, but your not actually getting the compiled output you expect, and even just installing a trial patch version of the latest VS can generate different binary outputs.

dgrunwald-qt:
I don't see a good way to stop rustc from adding the non-hermetic library search path, so put our cc_toolchain library search path on the command line where it has precedence over the non-hermetic path injected by rustc.

See https://rust-lang.zulipchat.com/#narrow/channel/242869-t-compiler.2Fwindows/topic/rust.20LIB.20path for discussion of the rustc behavior.

Closes #3256.


illicitonion (approved):
Seems reasonable! Thanks!

dgrunwald-qt:
Oops, I forgot to set up the pre-commit hooks. Formatting is fixed now.

Issue #3607, opened 2025-09-10
[Feature Request] Make Clippy aspect apply to transitive dependencies


blorente:
Currently, the clippy aspect doesn't traverse through `deps` edges. Therefore, if you build a `rust_binary` that depends on a `rust_library` and the `rust_library` has clippy issues, these won't be reported unless you also build the library.

To reproduce:

1. Create this target in `test/clippy/BUILD.bazel`:
```starlark
rust_binary(
    name = "bad_transitive_dep",
    srcs = ["src/main.rs"],
    edition = "2018",
    deps = [":bad_library"],
)
```
2. Then, remove the `noclippy` tag from `:bad_library`.
3. Run clippy:

- `bazel build --config=clippy test/clippy:bad_transitive_dep` -> Succeeds
- `bazel build --config=clippy test/clippy:bad_library` -> Fails
- `bazel build --config=clippy test/clippy:bad_library test/clippy:bad_transitive_dep` -> Fails

I would have expected building `:bad_transitive_dep` to fail, since one of their dependencies fails.

I'm happy to implement the traversal functionality, but I wanted to check first before going deep into it: Is there any context I should be aware of? Is this against the philosophy of the the aspect?

I wasn't able to find anything in the Bazel slack or in the open issues, sorry if I missed anything.

illicitonion:
I think in general lint-style aspects are expected to only lint the targets they're passed? If I asked rustfmt whether my library was formatted and it told me one of my dependencies was poorly formatted, I'd be surprised...

How is this being called at the moment? If e.g. the result of a query, is it possible to add a `deps()` call around the query?

IIRC the clippy aspect _is_ special-cased so that if it's called on a non-rust target it will just no-op, so you don't need to worry about filtering like `kind(rust_library|rust_binary|..., deps(...))` - you should be able to just `deps()` around your targets.

blorente:
>  If I asked rustfmt whether my library was formatted and it told me one of my dependencies was poorly formatted, I'd be surprised...

True. I think the main disconnect is that a clippy failure can make the `bazel build` fail, so it's more akin to a static check by the compiler (like asan), but I gues the same could be said for a very strict `rustfmt` check? 

As an alternative, would you take a refactor that exposes the "run clippy on this target" action in the public API, so that other rulesets that do traverse dependencies (like [rules_lint](https://github.com/aspect-build/rules_lint)) can consume it?

blackliner:
[...]

PR #3612, opened 2025-09-13
feat(clippy): Introduce flag to toggle transitive aspect

blorente:
A follow up to #3607 . This PR introduces a flag, `--@rules_rust//rust/settings:experimental_clippy_aspect_traverse_deps=True`, that tells the clippy aspect that it should traverse the `deps` edges of a target and collect its Clippy output and results.

See `clippy_failure_tester.sh` for effects before and after.

I'm still not sure this change is wanted, but clippy is close enough to a static analyzer like asan that I think it would make sense to give folks the option to toggle this behaviour. I'm happy to close this PR, I just wanted to give an option to folks who want to use clippy like this.


test/clippy/clippy_failure_tester.sh
```diff
@@ -43,7 +44,8 @@ function check_build_result() {
     if [[ $(stat ${STATOPTS[@]} "${NEW_WORKSPACE}/bazel-bin/test/clippy/${2%_clippy}.clippy.out") == 0 ]]; then
       >&2 echo "FAIL: Output wasn't written to out file building target //test/clippy:${2}"
       >&2 echo "  Output file: ${NEW_WORKSPACE}/bazel-bin/test/clippy/${2%_clippy}.clippy.out"
-      >&2 echo "  Run \"bazel build //test/clippy:${2}\" to see the output"
+      >&2 echo "  Run \"bazel build //test/clippy:${2} ${@:3}\" to see the output"
+
```
blorente, line 63: These two are drive-by changes, but I realized we should have the flags if we want to have an easy repro.

illicitonion:
This feels quite out of character from how I generally expect linters/analysers to work... My understanding is ASAN needs to be applied to deps because they actually need to be built in that mode for ASAN to work? Whereas it feels like it's really the job of the caller to decide which targets they want to run clippy for... Like, I'd expect folks to mostly be keeping HEAD clean, and either running clippy on `...` or running it only on changed targets?

We do at least set `noclippy` and `--cap-lints=allow` on `cargo-bazel`-generated targets, so we don't need to worry about traversing into third-party code.

@UebelAndre do you have any thoughts here?

blorente:
This is an alternative solution, where we expose the clippy action to other rulesets, so folks who want transitive behavior can use things like `rules_lint`: https://github.com/bazelbuild/rules_rust/pull/3660

blorente:
I'm going to close this in favor of #3660, but feel free to re-open if you think it woudl be valuable for the repository.

PR #3613, opened 2025-09-13
[cargo] Make specifying `link_deps` optional - they can be inferred

dzbarsky:
It's hard to generate `link_deps` on a per-crate basis without having a global view. We can infer it from `deps` in case it is omitted though.


dzbarsky:
This won't work - the `deps` are on the crate itself and the `link_deps` go on the build script. I can solve this in another way on my end though.

PR #3617, opened 2025-09-16
Handle rust_library_group dependencies in more places

dzbarsky:
This gives downstream rules more flexibility in how deps are supplied to various attributes


dzbarsky:
> Thanks! Could you also add a regression test that passes `rust_library_group` to `deps`, `proc_macro_deps` and `cargo_build_script`?

Good idea, done

PR #3622, opened 2025-09-17
`fail(...)` if both `crate` and `crate_root` attrs are specified on `rust_test`

bmoxb:
Conceptually, it doesn't really make sense for a `rust_test` to have both a `crate` and `crate_root` specified, plus for compilation purposes the crate root of the library target specified by `crate` is inherited anyway.

### DIFF
--- a/crate_universe/extensions.bzl
+++ b/crate_universe/extensions.bzl
@@ -858,29 +858,36 @@ def _get_host_cargo_rustc(module_ctx, host_triple, host_tools_repo):
 
     return cargo_path, rustc_path
 
-def _update_annotations(annotations, crate, version, triples, data, env):
+def _update_annotations(store, crate, version, triples, annotation_select_dict):
     """Insert build-script data/env for each target triple if keys exist."""
-    crate_info = annotations.get(crate)
-    if not crate_info:
-        fail("Crate {} not found in annotations".format(crate))
+    crate_info = store.get(crate)
+    if crate_info == None:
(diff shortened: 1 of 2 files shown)


rust/private/rust.bzl
```diff
@@ -386,7 +391,7 @@ def _rust_test_impl(ctx):
         # Need to consider all src files together when transforming
         srcs = depset(ctx.files.srcs, transitive = [crate.srcs]).to_list()
         compile_data = depset(ctx.files.compile_data, transitive = [crate.compile_data]).to_list()
-        srcs, compile_data, crate_root = transform_sources(ctx, srcs, compile_data, getattr(ctx.file, "crate_root", None))
+        srcs, compile_data, crate_root = transform_sources(ctx, srcs, compile_data, crate_root = None)
```
krasimirgg: Does this still work when `crate` points to a rust_library target that itself uses `crate_root`?

bmoxb: I've confirmed that it does (internally), including with mix of generated/non-generated srcs. I could add unit tests if you think it's worth it, but the previous code `getattr(ctx.file, "crate_root", None)` would basically always have evaluated to `None` anyway in any real world case.

krasimirgg: Oh, I think I see it: this is basically unused later, as the info is populated via `root = crate.root`. Could you try and replace `crate_root` with `_` so that becomes more obvious.

PR #3623, opened 2025-09-17
Mark cargo_bootstrap_repository as reproducible

dzbarsky:
As long as you have specified `srcs` properly, it should be. And if you haven't, you already have an issue.

As a side effect, this should allow using --repo_contents_cache and massively speeding up the "Bzlmod examples" job which spends most of it's time building cargo-bazel from source, repeatedly.


illicitonion (approved):
🎉

illicitonion:
This looks like it tickles some bug in Bazel 7, which we currently still support. When 9 is released, we'll stop supporting it...

keith:
we can close now https://github.com/bazelbuild/rules_rust/pull/4146

UebelAndre:
closed in favor of https://github.com/bazelbuild/rules_rust/pull/4146

PR #3628, opened 2025-09-22
Add runfiles example with more levels of indirection

hofbi:
https://github.com/bazelbuild/rules_rust/pull/3399 changed the behavior of runfiles. Due to this change a few cases no longer work with `bazel run` that did work before. I tried to create a more complex runfiles example since rules_rust does not have anyone yet + we can use it for discussing what should be the correct behavior.

```shell
# Everything with bazel test still works
bazel test //...

# bazel run no longer works for all cases that work with bazel test
bazel run //pkg_a:pkg_a_binary
bazel run //pkg_b:pkg_b_binary
```

I added a comment to the few cases that still work with `bazel test` but no longer work with `bazel run`. In particular, this happens when requesting a folder through runfiles instead of an individual file.

### DIFF
--- a/cargo/cargo_toml_variable_extractor/3rdparty/Cargo.Bazel.lock
+++ b/cargo/cargo_toml_variable_extractor/3rdparty/Cargo.Bazel.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "autocfg"
-version = "1.4.0"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
+checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "cargo-util-schemas"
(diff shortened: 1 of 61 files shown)


scasagrande:
> In particular, this happens when requesting a folder through runfiles instead of an individual file.

I recently noticed this as well. To get around this I adjusted some code to look for the `BUILD.bazel` file in the directory, and then trim off the file.

hofbi:
> > In particular, this happens when requesting a folder through runfiles instead of an individual file.
> 
> I recently noticed this as well. To get around this I adjusted some code to look for the `BUILD.bazel` file in the directory, and then trim off the file.

Could you share an example?

scasagrande:
Small correction, I just look for any specific file in the directory.


something along the lines of this should return the relative path to `a/b/c/` . This would require a data dependency on a filegroup target under `@a//b:c` which at least contains the file `somefile.txt`.

```
        let target = "a/b/c/somefile.txt";
        let rf = runfiles::Runfiles::create()?;
        let path = runfiles::rlocation!(rf, target);
        let path = match path {
            Some(value) => value,
            None => {
                return Err(...)
            }
        };
        let path = match path.parent() {
            Some(value) => value,
            None => {
                return Err(...)
            }
        };
```

hofbi:
Yes, getting the parent of a particular file is what I also used for a quick fix. But it feels a bit hacky.

scasagrande:
Well yeah 😆

UebelAndre (approved):
Awesome, thank you so much!

hofbi:
@UebelAndre Could you retigger the buildkite job. The 2 failing Mac build seem to be unrelated.

Issue #3625, opened 2025-09-19
Cargo toml env vars are not set on build scripts at runtime


dzbarsky:
We use the `cargo_toml_env_vars` rule to extract the CARGO_* env vars from the Cargo.toml and pass them to `rustc_env_files` for both the crate and it's corresponding build script, but we don't use those env var files to set the runtime env vars for the build script. We set s[ome of them in the rule](https://github.com/bazelbuild/rules_rust/blob/cf8603c7af32f609ec756f91c42118acb7d8a62f/cargo/private/cargo_build_script.bzl#L390-L415), but are still missing some like CARGO_PKG_AUTHORS.

This manifests as the `built` crate failing to find [the env vars it expects at runtime](https://github.com/lukaslueg/built/blob/4c71c96441e3fb3fd98fd4f36081448ebe0a96e8/src/environment.rs#L164C1-L217) when used as a build dep, and thus crates using it failing to build, such as [rav1e](https://github.com/xiph/rav1e/blob/a2f01b3e233f531c28a20b4c29fb5c9e5d29fa6d/build.rs#L249C10-L249C26) as a result.

It's not quite clear to me the best way to fix this - while `cargo_build_script_runner` seems to support loading an env file, it's not obvious that propagating `rustc_env_files` is semantically correct. Perhaps the `cargo_build_script` macro should have taken a `cargo_env_vars_file` attribute which would more cleanly imply it should apply at both compile time and runtime. @UebelAndre @illicitonion wdyt?

PR #3632, opened 2025-09-26
Allow passing env files to build script runtime env

dzbarsky:
Fixes https://github.com/bazelbuild/rules_rust/issues/3625, see that issue for explanation

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -536,6 +536,9 @@ def _cargo_build_script_impl(ctx):
 
     build_script_inputs = []
 
+    build_script_inputs.extend(ctx.files.build_script_env_files)
+    args.add_all(ctx.files.build_script_env_files, format_each = "--input_dep_env_path=%s")
+
     for dep in ctx.attr.link_deps:
         if rust_common.dep_info in dep and dep[rust_common.dep_info].dep_env:
             dep_env_file = dep[rust_common.dep_info].dep_env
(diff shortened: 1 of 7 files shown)


UebelAndre (changes requested):
Thanks! Can you add a regression test for this?

dzbarsky:
> Thanks! Can you add a regression test for this?

Yeah, I wasn't sure the best way to test this. Ideally we would have some e2e cargo_universe tests but I didn't see them (maybe I missed them)? I can add a unit test for this attribute though, it would be better than nothing

UebelAndre:
> > Thanks! Can you add a regression test for this?
> 
> Yeah, I wasn't sure the best way to test this. Ideally we would have some e2e cargo_universe tests but I didn't see them (maybe I missed them)? I can add a unit test for this attribute though, it would be better than nothing

Well the straight forward one is the change to `cargo_build_script` where you can generate an env file and make sure that's set where you expect it. For `crate_universe` that's a bit more complicated...

dzbarsky:
> > > Thanks! Can you add a regression test for this?
> > 
> > 
> > Yeah, I wasn't sure the best way to test this. Ideally we would have some e2e cargo_universe tests but I didn't see them (maybe I missed them)? I can add a unit test for this attribute though, it would be better than nothing
> 
> Well the straight forward one is the change to `cargo_build_script` where you can generate an env file and make sure that's set where you expect it. For `crate_universe` that's a bit more complicated...

Yep that's what I ended up implementing, PTAL :)

PR #3634, opened 2025-09-27
Speedup collect_deps

dzbarsky:
The key insight is that `deps` and `proc_macro_deps` must be disjoint, so the depset.to_list is not doing any deduping. With that out of the way, we do a bit of code rearranging to avoid converting our lists of deps to depsets too early; in general it is best to construct the output depsets when creating the providers, not in the middle of the rule implementations.

Before:
image

After:
image


dzbarsky:
> It's a little surprising how many `.to_list()` calls we're adding here, but I guess these were all just hidden inside `collect_deps`, so it'll be easier for us to remove these in the future if we want to try to, than it was before! Thank you!

Exactly. These are all terminal nodes also whereas before we would call to_list on lib to lib edges.

PR #3636, opened 2025-09-28
Avoid quadratic deduping in location expansion

dzbarsky:
Does what it says on the tin. Usually the deduping was not necessary at all because the argument were already coming from a depset or an attribute (though kept for safety), but doing the same work for every arg was truly unnecessary

Before:
image

After:
image


illicitonion (approved):
LGTM! Thanks!

PR #3639, opened 2025-09-28
Use None not remote for default vendor mode

illicitonion:
In RenderConfig, we want this to tell us whether we're vendoring _at all_, rather than how we happen to set up template values.


UebelAndre:
What is the impact of this? It's been a minute

PR #3642, opened 2025-09-29
Avoid collecting linkstamps when not needed in collect_deps

dzbarsky:


dzbarsky:
> Looks reasonable, but currently failing CI?

Yeah not super why offhand; I'll take a look once I land my other outstanding PRs and rebase

PR #3644, opened 2025-09-30
Make dep_variant_infos a list, not a depset

dzbarsky:
For anyone using rust_library_group (or similar, like rules_rs) the current setup forces flattening in every rdep. We can just carry them as a list instead and iterate it directly in the consumers.

We are slightly less efficient now when nesting a library group in a library group (which is probably very rare?).

The prost rules do incur flattening now when `include_transitive_deps` is enabled (which it's not by default, and it seems like a bad practice to not declare your deps properly?). In exchange, I've improved the depset handling for constructing those transitive deps (and I think we could further improve it by not constructing the transitive depset when the toolchain isn't configured to use it...)

This is technically a breaking change for folks who have custom rules creating CrateGroupInfo, although the adjustment should be pretty trivial.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -25,6 +25,8 @@ load(
     "BuildInfo",
     "CrateGroupInfo",
     "CrateInfo",
+    "DepInfo",
+    "DepVariantInfo",
     "LintsInfo",
 )
 load("//rust/private:rustc.bzl", "collect_extra_rustc_flags", "is_no_std", "rustc_compile_action")
@@ -78,7 +80,7 @@ def _assert_correct_dep_mapping(ctx):
         else:
             types = [
                 dep_variant_info.crate_info.type
-                for dep_variant_info in dep[CrateGroupInfo].dep_variant_infos.to_list()
+                for dep_variant_info in dep[CrateGroupInfo].dep_variant_infos
                 if dep_variant_info.crate_info
             ]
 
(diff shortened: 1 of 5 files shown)


rust/private/rust.bzl
```diff
@@ -546,30 +548,27 @@ def _rust_test_impl(ctx):
...
+                crate_info = dep[CrateInfo] if CrateInfo in dep else None,
+                dep_info = dep[DepInfo] if DepInfo in dep else None,
                 build_info = dep[BuildInfo] if BuildInfo in dep else None,
                 cc_info = dep[CcInfo] if CcInfo in dep else None,
                 crate_group_info = None,
             ))
-        elif rust_common.crate_group_info in dep:
-            dep_variant_transitive_infos.append(dep[rust_common.crate_group_info].dep_variant_infos)
+        elif CrateGroupInfo in dep:
+            dep_variant_infos.extend(dep[CrateGroupInfo].dep_variant_infos)
```
dzbarsky, line 563: we could dedupe here if we think folks have nested `rust_library_groups` with duplicated deps

dzbarsky:
> I'm not sure what the goal of this one is? Improving the depset construction sounds good, but why is a list the correct data type here? It feels like this is a maybe-nested maybe-has-duplicates-that-need-deduplicating collection of transitive providers?

The problem is that the transitive providers can never be used efficiently as-is, because all the consumers want depsets of their nested fields, not the providers themselves. For example, `cargo_build_script` flattens the `dep_variant_infos` to extract the contained `dep_infos`, and then flattens each dep_info's `transitive_build_infos` to get the `build_info.out_dir`. We would need a depset of `build_info_out_dirs` (and similar usages) propagated on the DepVariantInfo to truly make use of the depset efficiently (Basically what we wanted was struct-of-arrays instead of array-of-structs). Given how it's actually being used in practice, there's no point wrapping the list of variant infos into a depset since every consumer just calls `.to_list()` on it anyway.

You're right that it could technically be transitive/duplicated if `rust_library_group` is nested inside another `rust_library_group`, though this isn't a usage I have ever come across; and in that edge case we can just dedupe/flatten once at the `rust_library_group` target instead of in all the consumers.

PR #3645, opened 2025-09-30
Streamline deps handling in collect_deps

dzbarsky:
We already call `transform_deps` in all cases upstream, so we can't end up with Targets here anymore.

Before:
image

After:
image


UebelAndre (approved):
Nice! Thank you!

PR #3646, opened 2025-09-30
Avoid prematurely turning srcs into a depset before rustc_compile_action

dzbarsky:


rust/private/rustc.bzl
```diff
@@ -1435,7 +1438,7 @@ def rustc_compile_action(
                 crate_info.type,
                 ctx.label.name,
                 formatted_version,
-                len(crate_info.srcs.to_list()),
```
UebelAndre, line 1439: Could this be the source of the performance issues? I would rather delete this logging than make an api change.

UebelAndre (changes requested):

rust/private/rustc.bzl
dzbarsky: yep, exactly. I'm happy to delete this logging, but FWIW I think the API change is only to private functions; the CrateInfo provider still carries a depset

PR #3647, opened 2025-09-30
Improve depset construction in collect_deps

dzbarsky:
We don't need to create all these 1-element depsets; we can properly track direct/transitive lists. This is both faster and results in depsets that are half as deep.

Before:
image

After:
image


abrisco (approved):
Thanks!

UebelAndre (changes requested):
Thanks!

PR #3656, opened 2025-10-05
Provide RUSTDOC env var to build scripts

dzbarsky:
`built`[expects this to be present at build script runtime](https://github.com/lukaslueg/built/blob/0d82cd1b959275dd5b05ee717634913a18192d1f/src/environment.rs#L379)


UebelAndre (approved):
Sounds good to me!

PR #3657, opened 2025-10-06
collect_inputs.files: fix docs and and add a check for `data`

krasimirgg:

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -644,7 +644,8 @@ def collect_inputs(
     Args:
         ctx (ctx): The rule's context object.
         file (struct): A struct containing files defined in label type attributes marked as `allow_single_file`.
-        files (list): A list of all inputs (`ctx.files`).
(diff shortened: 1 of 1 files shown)


rust/private/rustc.bzl
```diff
@@ -717,7 +718,7 @@ def collect_inputs(
 
     # The old default behavior was to include data files at compile time.
     # This flag controls whether to include data files in compile_data.
-    if not toolchain._incompatible_do_not_include_data_in_compile_data:
+    if hasattr(files, "data") and not toolchain._incompatible_do_not_include_data_in_compile_data:
```
dzbarsky, line 721: please flip to `if not toolchain._incompatible_do_not_include_data_in_compile_data and hasattr(files, "data"):` to avoid adding additional checks for most users (it defaults to True)

krasimirgg: sorry, submitted too fast, doing as a follow-up...

PR #3659, opened 2025-10-06
Allow opting-out of the 0.0.0 crate version fallback

dzbarsky:
Most users are configuring these (indirectly) via .env files anyway, so for them this is a source of duplicative or conflicting env vars.

This also unifies the 2 inconsistent implementations of the version parsing.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -129,6 +129,29 @@ is_proc_macro_dep_enabled = rule(
     build_setting = config.bool(flag = True),
 )
 
+def env_vars_from_version(version):
+    """Gathers rustc environment variables corresponding to the version
+
+    Args:
+        version (str): The version of the crate to be compiled
+
+    Returns:
+        dict: Rustc environment variables
+    """
+    major, minor, patch = version.split(".", 2)
+    if "-" in patch:
+        patch, pre = patch.split("-", 1)
+    else:
+        pre = ""
+
+    return {
+        "CARGO_PKG_VERSION": version,
+        "CARGO_PKG_VERSION_MAJOR": major,
+        "CARGO_PKG_VERSION_MINOR": minor,
+        "CARGO_PKG_VERSION_PATCH": patch,
+        "CARGO_PKG_VERSION_PRE": pre,
+    }
+
 def _get_rustc_env(attr, toolchain, crate_name):
     """Gathers rustc environment variables
 
@@ -140,13 +163,6 @@ def _get_rustc_env(attr, toolchain, crate_name):
     Returns:
         dict: Rustc environment variables
     """
-    version = attr.version if hasattr(attr, "version") else "0.0.0"
-    major, minor, patch = version.split(".", 2)
-    if "-" in patch:
-        patch, pre = patch.split("-", 1)
-    else:
-        pre = ""
-
     result = {
         "CARGO_CFG_TARGET_ARCH": "" if toolchain.target_arch == None else toolchain.target_arch,
         "CARGO_CFG_TARGET_OS": "" if toolchain.target_os == None else toolchain.target_os,
@@ -155,12 +171,13 @@ def _get_rustc_env(attr, toolchain, crate_name):
         "CARGO_PKG_DESCRIPTION": "",
         "CARGO_PKG_HOMEPAGE": "",
         "CARGO_PKG_NAME": attr.name,
-        "CARGO_PKG_VERSION": version,
-        "CARGO_PKG_VERSION_MAJOR": major,
-        "CARGO_PKG_VERSION_MINOR": minor,
-        "CARGO_PKG_VERSION_PATCH": patch,
-        "CARGO_PKG_VERSION_PRE": pre,
     }
+
+    version_default = None if toolchain._incompatible_do_not_inject_degenerate_version_to_rustc_env else "0.0.0"
+    version = getattr(attr, "version", version_default)
+    if version:
+        result.update(env_vars_from_version(version))
+
     if hasattr(attr, "_is_proc_macro_dep_enabled") and attr._is_proc_macro_dep_enabled[IsProcMacroDepEnabledInfo].enabled:
         is_proc_macro_dep = "0"
         if hasattr(attr, "_is_proc_macro_dep") and attr._is_proc_macro_dep[IsProcMacroDepInfo].is_proc_macro_dep:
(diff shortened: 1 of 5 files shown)


rust/private/rustc.bzl
```diff
@@ -140,26 +140,10 @@ def _get_rustc_env(attr, toolchain, crate_name):
...
-
     result = {
         "CARGO_CFG_TARGET_ARCH": "" if toolchain.target_arch == None else toolchain.target_arch,
         "CARGO_CFG_TARGET_OS": "" if toolchain.target_os == None else toolchain.target_os,
         "CARGO_CRATE_NAME": crate_name,
-        "CARGO_PKG_AUTHORS": "",
-        "CARGO_PKG_DESCRIPTION": "",
-        "CARGO_PKG_HOMEPAGE": "",
-        "CARGO_PKG_NAME": attr.name,
-        "CARGO_PKG_VERSION": version,
```
UebelAndre, line 158: Won't this be unset for normal `rust_*` targets?

UebelAndre (changes requested):

rust/private/rustc.bzl
dzbarsky: It will, but on the off chance that they actually need this they can either set `env = {"CARGO_PKG_VERSION": version}` explicitly or use `cargo_toml_env_vars` for proper support, like we do for `cargo-bazel` in this PR. I think it's confusing to have multiple ways to provide this functionality, and this one is more incomplete compared to `cargo_toml_env_vars`

See also https://bazelbuild.slack.com/archives/CSV56UT0F/p1759051609743919?thread_ts=1758229364.645099&cid=CSV56UT0F

dzbarsky: @UebelAndre another way of looking at it - previously, not setting the 'version' attribute on manually-written rust_library rules would mean they get compiled with CARGO_PKG_VERSION_MAJOR=0 and similar. That's also kinda weird and probably not the desired behavior...

UebelAndre: it's useful for things like `clap` which basically embeds the binary version for free in your crate due to these environment variables. While I can see the argument for separating out cargo environment variables I think the few that are there do offer some convenience folks have already taken advantage of. If this were to change I think it would need an incompatibility flag.

krasimirgg: +1 to keep the existing defaults `CARGO_PKG_*` derived from the rust_* rules in skylark. Many random crates in the ecosystem examine these and we have setup that imports them, generating BUILD files, deriving the crate name and version of the target from the Cargo.toml file in the process, and discarding the `Cargo.toml` files after that. While the defaults are not always correct, these defaults cover a good chunk of crates in our experience.

I agree it's super confusing in the presence of cargo_toml_env_vars, and there I think we shouldn't derive the defaults. Instead of a new incompatibility flag, could we make it that the implementation detects the presence of cargo_toml_env_vars dependency and disables the defaults derivation if that's present?

dzbarsky: `cargo_toml_env_vars` is just a .env file, could your import process write a .env file with the values instead? (That would be [pretty similar to what I'm doing in rules_rs](https://github.com/dzbarsky/rules_rs/blob/c9f781b6c1c4904f796e99d174b1a79b5df39035/rs/private/repository_utils.bzl#L37-L57))

If we think changing this is too annoying I'm happy to stick it behind an incompatible flag, that would be good enough for my purposes

krasimirgg: I see, then adding an incompatible flag sounds good.

We could update the generators, but we can't update all crates' build files all at once (it's like like each crate is updated independently on its own schedule by separate groups of people). So we'll need the flexibility to opt in and out of this feature on a per-target basis if we are to move over.

A few more thoughts just to clarify my mental model:
* about the `rust_library.version` attribute: it is documented as an explicit way to stamp out the cargo env var. In the new mode, will this be a no-op?
* how does the `cargo_toml_env_vars` feature interact with `rust_library.rustc_env` when there are duplicate keys? Naively, `rustc_env` takes precedence right?

dzbarsky: How would you guys feel if we just stopped defaulting the version to 0.0.0 and left the vars unset in case the version wasn't provided? That would match the handling in the build_script rule and would give users an easy way to opt out of this behavior for all the version vars. It would also allow to share the duplicated parsing implementations that are inconsistent.

That would leave only CARGO_PKG_NAME which is sadly used by clap. (Fun fact, that usage is incorrect and it should use CARGO_BIN_NAME to get the right names, but that requires a clap fix...). We can tackle that one another day

krasimirgg: >How would you guys feel if we just stopped defaulting the version to 0.0.0 and left the vars unset in case the version wasn't provided?

I like this. I think for us internally this will be flexible enough to handle the mix of third party dependencies with-and-without set version.

From broader point of view, still might make sense to put this feature behind an incompatible flag (strictly it feels like it is). Also curious what @UebelAndre thinks.

Sharing another thought I had last week: we could define a custom provider, say `NoDefaultEnvVarsInfo()`, and have `cargo_toml_env_vars()` return it as well as the DefaultInfo, and in rustc.bzl gate the setting of the defaults on any of the rustc_env_files providing it.

dzbarsky: I think adding a custom provider is a decent idea but since I'm not using it on my end (Creating the .env file in the repo phase) I would then need to write a new rule wrapping those files just to create it, not sure it's worth it? Lets start with the not-defaulting-version suggestion above.

The only change in behavior really is that previously it would manufacture the `0.0.0` version which was probably not correct, and if anyone really wants that behavior they can just set `version = "0.0.0"` themselves? I don't mind flagging the new behavior if ya'll think there's a risk of this breaking users, but I do think incompatible flags add a tiny bit of cognitive burden so I'd consider avoiding it for something like this which is arguably fixing an inconsistent/weird behavior?

Anyway, perhaps @UebelAndre can make the final call here and we can land the change? :)

UebelAndre: Changing the default seems fine but this should definitely follow https://github.com/bazelbuild/rules_rust/blob/a14b74125e142dbbf968ee47381a1d370bfd2ba6/COMPATIBILITY.md and add an incompatibility flag that gates the change in the default.

But in terms of de-duplication, `crate_universe` explicitly sets `version` so won't you still run into duplication there from the cargo extracted variables?

dzbarsky: OK, I will rework this. I'm ok adding the flag as long as we will be able to followup to flip the default after 3 months, which I'm also happy to do.

It's true that `crate_universe` will still set `version` (which we could change in a followup), but my `crate_universe` replacement won't :)

dzbarsky: @krasimirgg @UebelAndre since this is no longer breaking, can we land it?

PR #3660, opened 2025-10-07
feat: Extract rust_clippy_action to allow other rulesets to run clippy

blorente:
Extract the action to run clippy into an importable struct, so that other rulesets can use it.

This is an alternative fix for https://github.com/bazelbuild/rules_rust/issues/3607. Since `rules_lint` does allow transitive semantics for linting, we can use the clippy action there, and users who want clippy to be transitive can use `rules_lint` to run it.

Reviewing instructions:
- Start from `rust/defs.bzl`. We expose two new functions: One to get the crate info, and one to create the clippy action itself.
- Then, read the docstring for `run_clippy`. I've tried to select a reasonable boundary for the arguments, but I'm definitely open to opinions. For instance, we don't ever read `ctx.attrs` in `run_clippy`, because we don't want to force users to implement the same arguments and flags, but that may be wrong.
- Then, go on to the call site of `run_clippy`, to see if it preserves the semantics you'd expect, particularly around capturing output.
- Finally, go on to the implementation of `run_clippy`


rust/private/clippy.bzl
```diff
@@ -101,24 +101,32 @@ def _get_clippy_ready_crate_info(target, aspect_ctx = None):
...
+        ctx (ctx): The rule's context object
+        clippy_executable (File): The clippy executable to run
+        process_wrapper (File): An executable process wrapper that can run clippy, usually //utils/process_wrapper
+        src (CrateInfo): The source crate information
+        config (File): The clippy configuration file. Reference: https://doc.rust-lang.org/clippy/configuration.html#configuring-clippy
+        output (File): The output file for clippy stdout/stderr. If None, no output will be captured.
+        success_marker (File): A file that will be written if the clippy succeeds.
+        cap_at_warnings (bool): If set, it will cap all reports as warnings, allowing the build to continue even with clippy failures
+        extra_clippy_flags (List[str]): A list of extra options to pass to clippy
+        error_format (str): Which error format to use. Must be acceptable by rustc: https://doc.rust-lang.org/beta/rustc/command-line-arguments.html#--error-format-control-how-errors-are-produced
```
blorente, line 117: Inside rules_rust, we can use the `rustc.bzl` file's utility functions to get this. I tried exposing them, but it felt like a leaky enough abstraction that this solution seemed cleaner.

blorente:
Could I get another approval for the formatting workflow?

rust/private/clippy.bzl
```diff
@@ -101,17 +101,25 @@ def _get_clippy_ready_crate_info(target, aspect_ctx = None):
     else:
         return None
 
-def _clippy_aspect_impl(target, ctx):
-    # Exit early if a target already has a clippy output group. This
-    # can be useful for rules which always want to inhibit clippy.
-    if OutputGroupInfo in target:
-        if hasattr(target[OutputGroupInfo], "clippy_checks"):
-            return []
+def run_clippy(ctx, clippy_executable, process_wrapper, crate_info, config, output = None, success_marker = None, cap_at_warnings = False, extra_clippy_flags = [], error_format = None, clippy_diagnostics_file = None):
```
UebelAndre: Why not call this `rust_clippy_action`?

blorente: Fair point! Not special reason, changed :D

blorente:
The merge queue failure seems to be unrelated: https://buildkite.com/bazel/rules-rust-rustlang/builds/15743/steps/canvas?sid=0199c539-cfda-4661-9b9c-ebe0adfcc847#0199c54a-fed2-41f9-b51e-22dafb24d35f/10-34
@UebelAndre  could you confirm if this is unrelated? Happy to investigate if you think it may be related, but better to ask first just in case :D

Issue #390, opened 2020-08-27
Static compilation of rust binaries using musl


smklein:
Hiya, this is more a platform for discussion than anything else. I wanted to post this issue here because I'm interested in contributing to it, but I wanted some guidance first.

**TL;DR: I'd like to be able to statically link rust binaries by compiling with musl.**

Following the instructions here: https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.html , it seems the `x86_64-unknown-linux-musl` target is what I should be using.

I've tried compiling the `examples` workspace with the following:

```
load("@io_bazel_rules_rust//rust:repositories.bzl", "rust_repository_set") 
                                                    
rust_repository_set(                                
    name = "rust_linux_x86_64",                                                                          
    exec_triple = "x86_64-unknown-linux-musl",
    extra_target_triples = ["wasm32-unknown-unknown"],
    version = "1.44.0",                             
    edition = "2018",                                                                                    
)         
```

But this has resulted in output like the following:

```
$ cd examples && bazel build //hello_world/...
INFO: Analyzed 2 targets (0 packages loaded, 0 targets configured).
INFO: Found 2 targets...
INFO: From Compiling Rust rlib hello_lib (2 files):
process wrapper error: failed to exec the new process: No such file or directory.
ERROR: /home/smklein/repos/rules_rust/examples/hello_lib/BUILD:11:13: output 'hello_lib/libhello_lib-590739884.rlib' was not created
ERROR: /home/smklein/repos/rules_rust/examples/hello_lib/BUILD:11:13: not all outputs were created or valid
INFO: Elapsed time: 0.259s, Critical Path: 0.05s
INFO: 1 process: 1 linux-sandbox.
FAILED: Build did NOT complete successfully
```

Is this expected behavior? Where should I start looking if I was interested in adding support for compilation against this alternative C library?

mfarrugi:
`process wrapper error: failed to exec the new process: No such file or directory.
` is obfuscating the error, would be useful to change

https://github.com/bazelbuild/rules_rust/blob/09ec54ca17c3b91dff256c520f1724fd9c329cab/util/process_wrapper/system_posix.cc#L160-L162 to print which executable was missing.

mfarrugi:
also `bazel build -s` might be more informative

dfreese:
`No such file or directory` is the output of the musl rustc when musl is not installed.

```bash
[...]

PR #3665, opened 2025-10-13
Add support for optional cc_toolchain

UebelAndre:
Changes:
- Moved `rust_allocator_libraries` rule implementation into it's own file
- Updated all rules to set `mandatory = False` for `@bazel_tools//tools/cpp:toolchain_type` toolchains.
- Refactored rules_rust internals to work in cases where `find_cc_toolchain` returns `None`
- Added `linker` as an optional attribute to `rust_toolchain`. This is expected to be `rust-lld` for now.
- Added a new setting `@rules_rust//rust/settings:toolchain_linker_preference` which in the event `rust_toolchain.linker` and a `cc_toolchain` are present, chooses the linker to use. (Defaults to `cc`).
- Introduce `--@rules_rust//settings:default_allocator_library` as a flag to globally control this value.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -727,7 +727,7 @@ tasks:
   # linking gnu libraries.
   # https://github.com/bazelbuild/rules_rust/issues/390
   # cross_compile_linux:
-  #   name: Cross compile example on Linux
+  #   name: Cross compile example on Linux without cc_toolchain
   #   platform: ubuntu2204
   #   working_directory: examples/cross_compile
   #   shell_commands:
@@ -737,7 +737,7 @@ tasks:
   #   test_targets:
   #     - "//..."
   cross_compile_macos:
-    name: Cross compile example on MacOS
+    name: Cross compile example on MacOS without cc_toolchain
     platform: macos_arm64
     working_directory: examples/cross_compile
     shell_commands:
@@ -747,7 +747,7 @@ tasks:
     test_targets:
       - "//..."
   cross_compile_windows:
-    name: Cross compile example on Windows
+    name: Cross compile example on Windows without cc_toolchain
     platform: windows
     working_directory: examples/cross_compile
     batch_commands:


UebelAndre:
@illicitonion @krasimirgg curious about your thoughts here 🙏

test/unit/native_deps/native_deps_test.bzl
```diff
@@ -343,12 +514,36 @@ def _native_dep_test():
         name = "bin_has_native_libs_test",
         target_under_test = ":bin_has_native_dep",
     )
-    bin_has_native_dep_and_alwayslink_test(
-        name = "bin_has_native_dep_and_alwayslink_test",
+    bin_has_native_dep_and_alwayslink_rust_linker_test(
+        name = "bin_has_native_dep_and_alwayslink_rust_linker_test",
+        target_under_test = ":bin_has_native_dep_and_alwayslink",
+        # TODO: https://github.com/bazelbuild/rules_rust/issues/390
```
UebelAndre, line 520: https://github.com/bazelbuild/rules_rust/issues/390

cargo/private/cargo_build_script.bzl
```diff
@@ -423,18 +420,25 @@ def _cargo_build_script_impl(ctx):
...
     env["LD"] = linker
     env["LDFLAGS"] = " ".join(_pwd_flags(link_args))
 
-    # MSVC requires INCLUDE to be set
-    cc_c_args, cc_cxx_args, cc_ar_args, cc_env = get_cc_compile_args_and_env(cc_toolchain, feature_configuration)
-    include = cc_env.get("INCLUDE")
-    if include:
-        env["INCLUDE"] = include
+    # Defaults for cxx flags.
+    env["CC"] = "${{pwd}}/{}".format(ctx.executable._fallback_cc.path)
```
dzbarsky, line 429: nit: can we define these inside the build_script_runner executable to avoid passing extra env vars we don't need?

cargo/private/BUILD.bazel
```diff
@@ -8,6 +8,36 @@ rust_binary(
     visibility = ["//visibility:public"],
 )
 
+rust_binary(
+    name = "no_ar",
+    srcs = ["no_binary.rs"],
+    edition = "2021",
+    rustc_env = {
```
dzbarsky: nit: Can we take the busybox approach? build a single copy of this, symlink it a few times, and then look at argv[0] to get the name it was invoked with?

rust/private/repository_utils.bzl
```diff
@@ -367,6 +400,12 @@ def BUILD_for_rust_toolchain(
     if global_allocator_library:
         global_allocator_library_label = "\"{global_allocator_library}\"".format(global_allocator_library = global_allocator_library)
 
+    linker_label = "None"
+    linker_type = "None"
+    if include_linker:
+        linker_label = "\"//:rust-lld\""
```
dzbarsky: might be a bit cleaner to do
```
linker_label = None
if include_linker:
    linker_label = "//:rust-lld"

...
template.format(linker_label = repr(linker_label))
```

rust/private/rustc.bzl
```diff
@@ -421,63 +422,127 @@ def get_linker_and_args(ctx, crate_type, cc_toolchain, feature_configuration, rp
...
+        # Filter link_args to only include flags that help locate libraries.
+        if cc_toolchain and link_args:
+            filtered_args = []
+            skip_next = False
+            for i, arg in enumerate(link_args):
+                if skip_next:
+                    skip_next = False
+                    continue
+
+                # Strip -Wl, prefix if using direct driver (it's only for compiler drivers)
```
dzbarsky, line 490: in my mind this is another +1 to switching to cc_common.link :)

rust/private/rustc.bzl
```diff
@@ -2272,34 +2427,35 @@ def _add_native_link_flags(args, dep_info, linkstamp_outs, ambiguous_libs, crate
...
+                    format_each = "-ldylib=%s",
+                )
+        else:
+            # For all other crate types we want to link C++ runtime library statically
+            # (for example libstdc++.a or libc++.a).
             args.add_all(
                 cc_toolchain.static_runtime_lib(feature_configuration = feature_configuration),
-                map_each = get_lib_name,
-                format_each = "-lstatic=%s",
+                map_each = _get_dirname,
```
dzbarsky, line 2457: do we actually need to use these -L or can we just pass `static_runtime_lib` directly so it expands to the actual libs

rust/settings/settings.bzl
```diff
@@ -247,6 +247,20 @@ def experimental_use_sh_toolchain_for_bootstrap_process_wrapper():
         build_setting_default = False,
     )
 
+def toolchain_linker_preference():
+    """A flag to control which linker is preferred for linking Rust binaries.
```
dzbarsky, line 290: can you expand this a bit to explain the difference between this setting and the cc_common.link one? i.e. what's the difference between enabling that and setting this one to `cc`? I'm worried that we're getting so many different linker options

dzbarsky: maybe the options should be "rust means rust linker, cc means use cc_common.link" so there's fewer possibilities?

rust/toolchain.bzl
```diff
@@ -487,6 +274,27 @@ def _generate_sysroot(
...
+        linker_bin = linker_files[0]
+        dest = "bin"
+        if "/bin/" in linker_bin.path:
+            _, _, subdir = linker_bin.path.partition("/bin/")
+            if subdir:
+                dest = "bin/{}".format(subdir[:-len("/" + linker_bin.basename)]).rstrip("/")
+
+        sysroot_linker = _symlink_sysroot_bin(ctx, name, dest, linker_bin)
+        sysroot_linker_files = _symlink_sysroot_tree(ctx, name, linker, linker[DefaultInfo].default_runfiles.files)
+        direct_files.extend([sysroot_linker])
```
dzbarsky: nit: `direct_files.append`, same on the line below

rust/toolchain.bzl
```diff
@@ -873,6 +719,22 @@ rust_toolchain = rule(
             doc = "Target that provides allocator functions for when a global allocator is present.",
             default = "@rules_rust//ffi/cc/global_allocator_library",
         ),
+        "linker": attr.label(
+            doc = "The label to an explicit linker to use (e.g. `rust-lld`, `ld`, `link-ld.exe`, etc...)",
+            # `target` cfg is used so a linker can be chose based on the target
+            # platform. Linker binaries are still required to be runnable in the
+            # `exec` configuration.
+            cfg = "target",
```
dzbarsky: this doesn't seem right, I think what you want is set `cfg = "exec"` but instantiate it as
```
linker = select({
    "@platforms//...": ...
})
```

and, interestingly enough, the select will be evaluated against the target platform, because it's resolved before the exec transition is applied. See https://github.com/cerisier/toolchains_llvm_bootstrapped/commit/aaeeacd2df667e6fe8788379eceee23f5a27e20d#diff-5e02a5ab7afca2441282a65cca097c8fade25a1d6d9ffbb71b920c19a1e58babR32-R39 and the corresponding explanation on the Bazel issue

rust/settings/settings.bzl
UebelAndre: Yeah, this parameter is a bit awkward. The idea is that even if you have both provided `rust_toolchain.linker` and a `cc_toolchain` is configured, that you can be explicit about what linker you'd like to use. The primary use case is just to force the use of `rust_toolchain.linker` even when a cc_toolchain is defined.

rust/private/repository_utils.bzl
UebelAndre: Updated!

rust/toolchain.bzl
UebelAndre: Updated!

rust/toolchain.bzl
UebelAndre: Updated!

rust/private/repository_utils.bzl
UebelAndre: Updated!

cargo/private/BUILD.bazel
UebelAndre: It's a bit trickier to do so since I'd either need to write tempdir logic or come up with some heuristic for where to put these files where I feel this approach is more stable.

dzbarsky: Why do you need a TMPDIR? You can define the targets exactly as now, in this package. Just change the rule to be a copy_file(symlink=True) instead of a brand new binary

UebelAndre: Updated! I'm nervous about `symlink=True` as I don't trust symlinks to work on windows. E.g. https://github.com/bazelbuild/bazel/issues/21747

rust/private/rustc.bzl
UebelAndre: We could probably pass `static_runtime_lib` but my changes here are mostly cleanup so I wouldn't want to include that in this PR.

cargo/private/cargo_build_script.bzl
UebelAndre: https://github.com/bazelbuild/rules_rust/pull/3665#discussion_r2581771403 for this reason I think this is harder than it's worth or spreads out configuration definitions.

cargo/private/BUILD.bazel
```diff
@@ -8,6 +9,34 @@ rust_binary(
...
 )
 
+rust_binary(
+    name = "no_binary",
+    srcs = ["no_binary.rs"],
+    edition = "2021",
+    visibility = ["//visibility:public"],
+)
+
+copy_file(
```
dzbarsky, line 19: If you use the version from https://registry.bazel.build/docs/bazel_lib, `allow_symlink = True` is safe. 
image

cargo/private/cargo_build_script.bzl
dzbarsky: ok, we can punt for now. In general I want us to stop using `.path` as that makes path-mapping not work, so many times we are building crates in multiple configurations even when not really necessary. But lets address in followups

dzbarsky (approved):

cargo/private/BUILD.bazel
UebelAndre: That would require adding a new dependency which I'd *really* like to avoid. There's some `TODOs` referring to https://github.com/bazelbuild/bazel/issues/21747 that I'd like to clean up at some point and this could be one of the things visited in that change.

dzbarsky: Ok let's leave a TODO. Realistically that's a pretty lightweight dep and probably most projects already pull it in :)

.bazelci/presubmit.yml
```diff
@@ -723,19 +723,43 @@ tasks:
...
     build_targets:
       - "//..."
-  # TODO: https://github.com/bazelbuild/rules_rust/issues/2075
-  # cross_compile_zig:
-  #   name: Cross compile example with Zig
+  # TODO: This test requires musl or for rustc to provide necessary gnu libraries for
+  # linking gnu libraries.
+  # https://github.com/bazelbuild/rules_rust/issues/390
+  # cross_compile_linux:
+  #   name: Cross compile example on Linux
```
illicitonion:
```suggestion
  #   name: Cross compile example on Linux without cc_toolchain
```

.bazelci/presubmit.yml
```diff
@@ -723,19 +723,43 @@ tasks:
...
-  #   build_targets:
+  #   working_directory: examples/cross_compile
+  #   shell_commands:
+  #     - sed -i 's/_FORCE_DISABLE_CC_TOOLCHAIN = False/_FORCE_DISABLE_CC_TOOLCHAIN = True/' ../../rust/private/utils.bzl
+  #   test_flags:
+  #     - "--keep_going"
+  #   test_targets:
   #     - "//..."
+  cross_compile_macos:
+    name: Cross compile example on MacOS
```
illicitonion:
```suggestion
    name: Cross compile example on MacOS without cc_toolchain
```

.bazelci/presubmit.yml
```diff
@@ -723,19 +723,43 @@ tasks:
...
+    platform: macos_arm64
+    working_directory: examples/cross_compile
+    shell_commands:
+      - sed -i '' 's/_FORCE_DISABLE_CC_TOOLCHAIN = False/_FORCE_DISABLE_CC_TOOLCHAIN = True/' ../../rust/private/utils.bzl
+    test_flags:
+      - "--keep_going"
+    test_targets:
+      - "//..."
+  cross_compile_windows:
+    name: Cross compile example on Windows
```
illicitonion:
```suggestion
    name: Cross compile example on Windows without cc_toolchain
```

illicitonion (approved):
Works for me! Thank you!

PR #3668, opened 2025-10-15
Update bazel_env example based on the latest changes and add CI

hofbi:
- It is no longer required to list all individual manifests
- Also we should track the MODULE.bazel.lock file
- Add bazel_env example to bazelci
- Tag broken test as manual to be fixed in a follow up PR


illicitonion (approved):
Thanks!

PR #3669, opened 2025-10-15
Bump ctve's autocfg from 1.4.0 -> 1.5.0 to pick up determinism fix.

top-oai:
I was doing some action determinism analysis and found that rules_rust's ctve's autocfg (version 1.4.0) showed up in a lot of sha differences....  looks like it leaves behind files with random filenames. This was fixed upstream in https://github.com/cuviper/autocfg/pull/75, which is included in autocfg 1.5.0.

Posted about this in slack at https://bazelbuild.slack.com/archives/CSV56UT0F/p1760515874096159.

### DIFF
--- a/cargo/cargo_toml_variable_extractor/3rdparty/Cargo.Bazel.lock
+++ b/cargo/cargo_toml_variable_extractor/3rdparty/Cargo.Bazel.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "autocfg"
-version = "1.4.0"
+version = "1.5.0"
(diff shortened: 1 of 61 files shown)


top-oai:
> Instead if adding this dependency can you just run
> 
> ```
> bazel run //cargo/cargo_toml_variable_extractor/3rdparty:crates_vendor -- --repin=autocfg
> ```
> 
> and commit the results?

Yeah I can, it works, thanks!

A bit of noise with the addition of `@rules_rust//rust/platform:riscv64gc-unknown-linux-gnu`.

UebelAndre (approved):
Looks good to me, thanks!

Issue #370, opened 2020-07-17
Debugger setup in vscode


UebelAndre:
Is there a suggested method to setup a debugger against a rust target in Visual Studio Code?

damienmg:
Hi Uebel, I don't think there is any. I tend to use lldb directly after a build in `-c dbg` mode.

UebelAndre:
Ah, ok. Sorry for the delay and thank you @damienmg in particular for answering being so active 😄

damienmg:
No worries, I want to keep this bug open as I think it represent a reasonable feature request (though probably a low priority one).

UebelAndre:
@damienmg Can you give an example of a command you run?

damienmg:
That's all in the command line like:

```sh
$ bazel build -c dbg //my:target
$ lddb bazel-bin/my/target
```

UebelAndre:
> That's all in the command line like:
> 
> 
> 
> ```sh
> 
> $ bazel build -c dbg //my:target
> 
> $ lddb bazel-bin/my/target
> 
> ```

Oh, is there no way to pass `lldb` as an argument to `bazel run`? That was my original hope 😅

damienmg:
There should be the --run_under flag of `bazel run` but quite frankly I
have never used it for debugging rust binary so I would not be able to help
here. Also this is more important for debugging test where a huge setup is
done outside the binary.

klochek:
Hi!  I just ran into this today, and had to work around a few things to get lldb debugging under vscode (on linux) to work.  I thought I might share this in case someone needs to get this working ASAP (or wants to turn this into a patch...?)

In debug builds, `rustc` writes the source files' _absolute_ paths to the target binary, which means those paths will be in whatever sandbox directory Bazel was using for building.  Thus, when you try to debug that binary, and lldb tries to map some code to a file, it goes looking for a source file in a sandbox directory that no longer exists, and so you can't see the source, set breakpoints, inspect variables, etc.

Happily, rustc exposes a compiler option, `--remap-path-prefix=FROM=TO`, to deal with this kind of situation.  I patched my local version of `rules_rust` to include the following arg during invocation: `args.add("--remap-path-prefix=${pwd}/=workspace/")`  I'm mapping the action sandbox directory to "workspace/", and so all source file paths written to the binaries have "workspace/" as the root.  Then, you just provide a mapping from "workspace/" to your WORKSPACE root in vscode, and then everything works as expected.

[...]

PR #3671, opened 2025-10-15
Added tool for generating vscode debug tasks

UebelAndre:
This change introduces `//tools/vscode` which includes tools for vscode integrations. Separate from `//tools/rust_analyzer`, this new package includes a `gen_launch_json` target which can be used to generate "Run and Debug" tasks for `rust_binary` and `rust_test` targets.

closes https://github.com/bazelbuild/rules_rust/issues/370

### DIFF
--- a/test/deps.bzl
+++ b/test/deps.bzl
@@ -5,8 +5,8 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load("//test/generated_inputs:external_repo.bzl", "generated_inputs_in_external_repo")
 load("//test/load_arbitrary_tool:load_arbitrary_tool_test.bzl", "load_arbitrary_tool_test")
 load("//test/rust_analyzer/3rdparty/crates:crates.bzl", rust_analyzer_test_crate_repositories = "crate_repositories")
-load("//test/vscode/3rdparty/crates:crates.bzl", vscode_test_crate_repositories = "crate_repositories")
 load("//test/unit/toolchain:toolchain_test_utils.bzl", "rules_rust_toolchain_test_target_json_repository")
+load("//test/vscode/3rdparty/crates:crates.bzl", vscode_test_crate_repositories = "crate_repositories")
 
 _LIBC_BUILD_FILE_CONTENT = """\
 load("@rules_rust//rust:defs.bzl", "rust_library")
--- a/test/vscode/integration_tests/BUILD.bazel
+++ b/test/vscode/integration_tests/BUILD.bazel
@@ -4,43 +4,43 @@ rust_test(
     name = "only_binaries_test",
     srcs = ["only_binaries_test.rs"],
     edition = "2021",
-    deps = [
-        "//test/vscode/3rdparty/crates:serde",
-        "//test/vscode/3rdparty/crates:serde_json",
-    ],
     # LAUNCH_JSON env var will be set by the test runner script
     # This target is tagged as manual since it's not expected to pass in
     # contexts outside of `//test/vscode:vscode_test`. Run
     # that target to execute this test.
     tags = ["manual"],
+    deps = [
+        "//test/vscode/3rdparty/crates:serde",
+        "//test/vscode/3rdparty/crates:serde_json",
+    ],
 )
 
 rust_test(
     name = "only_tests_test",
     srcs = ["only_tests_test.rs"],
     edition = "2021",
-    deps = [
-        "//test/vscode/3rdparty/crates:serde",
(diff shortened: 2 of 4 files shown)


test/vscode/integration_tests/binaries_and_tests_test.rs
```diff
@@ -0,0 +1,62 @@
...
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_binaries_and_tests() {
+        let launch_json_path = PathBuf::from(env::var("LAUNCH_JSON").unwrap());
+        let content = fs::read_to_string(&launch_json_path)
+            .unwrap_or_else(|_| panic!("couldn't open {:?}", &launch_json_path));
+
+        // Verify basic JSON structure
+        assert!(content.contains(r#""version": "0.2.0""#));
```
illicitonion: I'd be tempted to actually parse the JSON here rather than doing string asserts, but also recognise going from 0 to non-0 dependencies is annoying

tools/vscode/src/bin/gen_launch_json.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[arg(short, long, default_value = "bazel")]
+    bazel: Utf8PathBuf,
+
+    /// Only generate configurations (don't write files)
+    #[arg(long)]
+    dry_run: bool,
+
+    /// Don't pretty print generated JSON
+    #[arg(long)]
+    no_pretty: bool,
```
illicitonion, line 39: Maybe `pretty` with default `true`, rather than default negation?

tools/vscode/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+            .collect();
+
+        Ok(Self {
+            output_base: info_map
+                .get("output_base")
+                .context("Failed to query `bazel info output_base`")?
+                .clone(),
+            workspace: info_map
+                .get("workspace")
+                .context("Failed to query `bazel info output_base`")?
```
illicitonion:
```suggestion
                .context("Failed to query `bazel info workspace`")?
```

tools/vscode/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+            if parts.len() >= 3 {
+                let kind = parts[0];
+                let label = parts[2];
+
+                if !kind.starts_with("rust_") {
+                    // Skip non-Rust targets
+                    continue;
+                }
+
+                result.insert(label.to_string(), line.to_string());
```
illicitonion:
```suggestion
                result.insert(label.to_string(), kind.to_string());
```

Right now you're saving the whole line, not just the kind, which you can get away with because you happen to only be doing `contains` checks

tools/vscode/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+                    target_kind: target_kind.clone(),
+                });
+            }
+        }
+
+        Ok(results)
+    }
+
+    /// Query information about a specific target.
+    pub fn query_target_info(&mut self, target: &str) -> Result<TargetInfo> {
```
illicitonion: AFAICT this is unused?

I was going to ask why we need both this and `find_rust_targets`, but I suspect the answer is we don't?

tools/vscode/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+        );
+        env.insert(
+            "TEST_UNDECLARED_OUTPUTS_ANNOTATIONS_DIR".to_string(),
+            vscode_dir
+                .join(format!("bazel-test-annotations-{}", sanitized_name))
+                .to_string(),
+        );
+
+        // Rust-specific
+        env.insert("RUST_BACKTRACE".to_string(), "1".to_string());
```
illicitonion:
```suggestion
        env.insert("RUST_BACKTRACE".to_string(), "all".to_string());
```

test/vscode/integration_tests/binaries_and_tests_test.rs
UebelAndre: Done

tools/vscode/src/lib.rs
UebelAndre: Good catch, deleted!

tools/vscode/src/bin/gen_launch_json.rs
UebelAndre: Does clap support this?

illicitonion: Sigh, apparently not simply enough to be worth the hassle!

PR #3676, opened 2025-10-18
Avoid flattening cc toolchain depset when constructing rust toolchain

dzbarsky:
These get converted to a depset anyway inside `cc_common`:

https://github.com/bazelbuild/bazel/blob/0cf8627d6f199979e6919a3e04b255268c547cdb/src/main/starlark/builtins_bzl/common/cc/link/create_linking_context_from_compilation_outputs.bzl#L85-L104


UebelAndre (approved):
Good find!

PR #3683, opened 2025-10-20
release 0.67.0

krasimirgg:
Mainly for the srcs changes from https://github.com/bazelbuild/rules_rust/commit/3f392091ae2aaf124af204eb940a51e67d3692e1, so we can adapt crubit and protobuf-rust accordingly.

PR #3684, opened 2025-10-20
add util/release.sh script to automatically bump the rules_rust version

krasimirgg:
Just a utility to make version bumps easier (not aware of an existing script that does this).

As a test, I generated https://github.com/bazelbuild/rules_rust/pull/3683 via this.

### DIFF
--- a/util/release.sh
+++ b/util/release.sh
@@ -40,3 +40,42 @@ grep -rl \
   --include='MODULE.bazel' --include='*.bzl' --include='*.md' \
   "$(bazel_dep_pattern $OLD)" \
   | xargs sed -i "s/^$(bazel_dep_pattern $OLD)$/$(bazel_dep_pattern $NEW)/"
+
+# Update module() declarations:
+# module(
+#     name = "rules_rust",
+#     version = "...",
+# )
+function update_module_version() {
+  local file=$1
+
+  local out="$(awk -v old=$OLD -v new=$NEW '
+  BEGIN {
+    VERSION_PATTERN = "version = \"" old "\"";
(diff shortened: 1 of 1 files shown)


util/release.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+#
+# Usage: util/release.sh [version]
+#
+# where version (optional) is the new version of rules_rust.
+set -eux
```
UebelAndre:
```suggestion
set -xeuo pipefail
```

util/release.sh
```diff
@@ -0,0 +1,33 @@
...
+
+function new_from_old() {
+  local major=$(awk -F. '{print $1}' <<<"$OLD")
+  local minor=$(awk -F. '{print $2}' <<<"$OLD")
+  echo "$major.$((minor + 1)).0"
+}
+
+readonly NEW="${1:-$(new_from_old)}"
+
+sed -i "s/$OLD/$NEW/" \
```
UebelAndre: I've mostly been doing just find and replace in my editor, to keep this future-proof, maybe the files included should be any `MODULE.bazel`, `*.bzl` and `*.md` files that contain `bazel_dep(name = "rules_rust*" = "{OLD_VERSION}")` and migrate it to the new version?

krasimirgg: updated the script. I didn't see any instances of `bazel_dep(name = "rules_rust*", ...)` beyond literally `"rules_rust"`, so I kept that as a pattern.

krasimirgg: Ohh, I see now, there are a bunch of module() declarations with `rules_rust*` added some logic to bump the version there as well.

Issue #3675, opened 2025-10-18
LLVM tools are going to break in Rust 1.100.0


dtolnay:
Just flagging that this condition needs to be adjusted because "1.100.0" < "1.45.0".

https://github.com/bazelbuild/rules_rust/blob/3f392091ae2aaf124af204eb940a51e67d3692e1/rust/repositories.bzl#L485-L486

Rust 1.100.0 is planned for late next year.

UebelAndre:
Thank you @dtolnay !

PR #3685, opened 2025-10-20
Fix version checks to support `rust >= 1.100.0`

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/3675


illicitonion (approved):
Thanks!

Issue #3247, opened 2025-02-07
cargo_build_script uses CARGO_MANIFEST_DIR ending in "/"


ashi009:
Populating CARGO_MANIFEST_DIR in cargo_build_script will not clean the path like in rust.bzl and rustc.bzl.

https://github.com/bazelbuild/rules_rust/blob/478cef4038f2a006800c0bdd31736d4fee6cd1e5/rust/private/rustc.bzl#L908-L911

https://github.com/bazelbuild/rules_rust/blob/478cef4038f2a006800c0bdd31736d4fee6cd1e5/rust/private/rust.bzl#L461-L462

https://github.com/bazelbuild/rules_rust/blob/478cef4038f2a006800c0bdd31736d4fee6cd1e5/cargo/private/cargo_build_script.bzl#L328-L340

ashi009:
@UebelAndre please see PR #3689

PR #3689, opened 2025-10-22
cargo_build_script: normalize CARGO_MANIFEST_DIR in cargo_build_script

ashi009:
This PR fixes issue #3247 by normalizing the CARGO_MANIFEST_DIR path in cargo_build_script to ensure it doesn't end with trailing slashes.

The issue occurs because CARGO_MANIFEST_DIR in cargo_build_script doesn't clean the path like in rust.bzl and rustc.bzl, leading to inconsistent paths that can cause problems for build scripts.

Fixes https://github.com/bazelbuild/rules_rust/issues/3247


UebelAndre (changes requested):
Thanks! Normalizing seems good but I'm still curious what the impact is. Also, this method of normalization seems funny, I would agree something like `paths.normalize` might make more sense but it also sounded like the issue was just a missing `.rstrip("/")`?

Issue #3246, opened 2025-02-07
Sandbox path leaked into cargo build script outputs


ashi009:
When cargo build script invokes CC, sandbox paths are leaked into the output.

The mitigate would be passing extra flags to CFLAGS and CXXFLAGS to make sure the paths are remapped like `cc_library`.

UebelAndre:
This is kinda known but absolutely undesired. I'd be happy to review pull requests to fix this!

ashi009:
@UebelAndre please see PR #3690

UebelAndre:
Gonna close this out due to lack of a repro. Please feel free to ping or refile one with a repro so we can investigate the root cause.

PR #3690, opened 2025-10-22
cargo_build_script: remap sandbox path to create reproducible outputs

ashi009:
This PR fixes issue #3246 by remapping sandbox paths to create reproducible outputs in cargo build scripts.

The issue occurs when cargo build script invokes CC, causing sandbox paths to leak into the output. The fix adds extra flags to CFLAGS and CXXFLAGS to remap paths similar to how cc_library works.

Fixes https://github.com/bazelbuild/rules_rust/issues/3246

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -444,11 +444,16 @@ def _cargo_build_script_impl(ctx):
             action_name = ACTION_NAMES.cpp_link_static_library,
         )
 
+        remap_flags = [
+            "-ffile-prefix-map=${pwd}=BAZEL_EXEC_ROOT",
+            "-ffile-prefix-map=${{pwd}}/{}=CARGO_MANIFEST_DIR".format(manifest_dir),
+        ]
+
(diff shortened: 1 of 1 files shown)


cargo/private/cargo_build_script.bzl
```diff
@@ -444,11 +444,16 @@ def _cargo_build_script_impl(ctx):
             action_name = ACTION_NAMES.cpp_link_static_library,
         )
 
+        remap_flags = [
+            "-ffile-prefix-map=${pwd}=BAZEL_EXEC_ROOT",
```
UebelAndre, line 448: Is this consistent across all compilers? I would be surprised if msvc supported it. While it is a flag I want this seems like something that should come from the `cc_toolchain` or there should be some additional checks to ensure the flag is gonna be used when `CC` is a compiler that supports it.

ashi009: I should have held this PR for later. It doesn't even work for early versions of gcc. Claude was too eager to convert all our patches to PRs.

PR #3693, opened 2025-10-22
Fix aliased crate handling in aliases attribute

### LINKED ISSUE #3692: Error in aliases() when using default_alias_rule_bzl with aliased crates
## Bug Description

When using `default_alias_rule_bzl` in `crate_universe` configuration, Bazel fails to build targets that depend on aliased crates. The error occurs in the `aliases()` function when it tries to reference `crate_info.owner` for aliased targets, but the aliased target's `crate_info` has a different structure than expected.

## Environment

- **Bazel version**: Multiple versions affected
- **rules_rust version**: Multiple versions affected  
- **Operating system**: Cross-platform issue

## Reproduction Steps

1. Configure `crate_universe` with `default_alias_rule_bzl`:
   ```starlark
   crates_repository(
       name = "crate_index",
       # ... other config ...
           default_alias_rule_bzl = "opt",
       ),
   )
   ```

2. Use a crate that creates aliased targets in the generated BUILD files

3. Try to build a target that uses the `aliases()` function

4. Observe build failure in the generated aliases() function

## Error Details

The issue manifests in the generated code in `rust/private/rustc.bzl` at approximately line 255:

```starlark
# Current problematic code:
for k, crate_info in crate_infos.items():
    # For aliased crates, crate_info.owner may not match k.label
    # This causes inconsistencies in alias generation
```

When `default_alias_rule_bzl` is used, the system generates aliased targets, but the alias resolution logic doesn't correctly handle the relationship between the aliased target's label (`k.label`) and its underlying crate info owner (`crate_info.owner`).

## Root Cause

The problem occurs because:
1. `default_alias_rule_bzl` creates alias targets that wrap the actual crate targets
2. The current logic in `rustc.bzl` assumes a direct correspondence between target labels and crate_info owners
3. For aliased targets, `crate_info.owner` points to the original target, not the alias label
4. This mismatch breaks the alias generation logic

## Proposed Fix

The fix should handle the case where aliased crates have different `owner` values in their `crate_info`:

```starlark
# Proposed fix in rust/private/rustc.bzl around line 255:
for k, crate_info in crate_infos.items():
    # Handle aliased crates where crate_info.owner != k.label
    owner_label = crate_info.owner if hasattr(crate_info, 'owner') else k.label
    # Use owner_label in subsequent logic instead of assuming k.label == crate_info.owner
```

## Impact

This bug prevents users from leveraging `default_alias_rule_bzl` functionality in `crate_universe`, which is intended to provide better integration with existing Bazel rule sets (like `@rules_cc`) and more consistent alias handling.

## Additional Context

This issue affects the core alias generation mechanism in rules_rust and would impact any project trying to use `default_alias_rule_bzl` with aliased crate dependencies. The fix should maintain backward compatibility while properly handling the aliased crate scenario.

The problem specifically manifests when the crate_universe dependency resolver creates targets with aliased names that don't directly correspond to their underlying crate_info owner labels.

ashi009:
## Description

This PR fixes an issue where `crate_universe` configuration using `default_alias_rule_bzl` would fail when processing aliased crates in the aliases() function.

## Problem

When using `default_alias_rule_bzl` in crate_universe configuration, Bazel fails to build targets that depend on aliased crates. The error occurs because the current logic assumes a direct correspondence between target labels (`k.label`) and their crate info owners (`crate_info.owner`), but for aliased targets, `crate_info.owner` points to the original target, not the alias label.

## Solution

This change updates the alias resolution logic in `rust/private/rustc.bzl` to correctly handle aliased crates by:

1. Checking if a target has `rust_common.crate_info` 
2. Using `crate_info.owner` if available, otherwise falling back to `k.label`
3. Ensuring consistent alias generation for both regular and aliased targets

## Code Changes

**Modified files:**
- `rust/private/rustc.bzl`: Updated alias processing logic in `collect_deps` function

**Before:**
```starlark
aliases = {k.label: v for k, v in aliases.items()}
```

**After:**
```starlark
aliases = {
    k[rust_common.crate_info].owner if rust_common.crate_info in k else k.label: v
    for k, v in aliases.items()
}
```

## Testing

This fix enables the use of `default_alias_rule_bzl` functionality without breaking alias generation for crates that have aliased targets.

## Related Issues

Fixes #3692

## Checklist

- [x] Code follows the existing style and patterns  
- [x] No breaking changes to existing functionality
- [x] Maintains backward compatibility
- [x] Issue was created and linked

---

**Note**: This change is essential for users who want to leverage `default_alias_rule_bzl` with crate_universe to integrate with existing Bazel rule sets like `@rules_cc`.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -271,7 +271,10 @@ def collect_deps(
         else:
             crate_deps.append(dep)
 
-    aliases = {k.label: v for k, v in aliases.items()}
+    aliases = {
+        k[rust_common.crate_info].owner if rust_common.crate_info in k else k.label: v
+        for k, v in aliases.items()
+    }
     for dep in crate_deps:
         crate_info = dep.crate_info
         dep_info = dep.dep_info


UebelAndre (changes requested):
Thanks! Could you also add a starlark unit test for this in `test/unit/aliases` or something?

UebelAndre (approved):
Thanks! I'll implement tests here: https://github.com/bazelbuild/rules_rust/pull/4242

In the future though we will generally be asking for tests on PRs

PR #3699, opened 2025-10-24
Write lockfile external only

hofbi:
## Description

Another attempt to improve https://github.com/bazelbuild/rules_rust/issues/3522

Current status/plan on the issue above was

- pre-commit hook that converts the `Cargo.lock` file into a `Cargo.bazel.lock` file by removing internal deps
- use `crate.from_cargo` with the `Cargo.bazel.lock` file and `skip_cargo_lockfile_overwrite = True`
- commit `MODULE.bazel.lock` which should have less merge conflicts since the sha of `Cargo.bazel.lock` does not change that often

This works, but we need to make sure `Cargo.lock` is up to date. This requires another step, so we replace a slow automated step (cargo splicing) with an extra manual/non-bazel step (updating the `Cargo.lock` file).

New idea to iterate

- use `crate.from_cargo` with the `Cargo.bazel.lock` file but instead of using `skip_cargo_lockfile_overwrite = True`
   - we let rules_rust do the cargo dependency resolution that is happening as part of splicing
   - instead of writing back the full content into the cargo lock file, we only write back the third party (making the pre-commit hook obsolete). For this, I introduced the `strip_internal_dependencies_from_cargo_lockfile` flag

This would save us the pre-commit hook as well as the manual step to update the `Cargo.lock` file. I already did a quick verification and `remove_internal_dependencies_from_cargo_lockfile` achieves the same results as the python implementation I listed in https://github.com/bazelbuild/rules_rust/issues/3522

## Open Points

This is an early draft to collect feedback. Once we aligned how exactly we want to proceed, I would like to

- [ ] add/extend an integration test to have this tested end to end
- [ ] add and example to demonstrate how this should be used

### DIFF
--- a/examples/crate_universe/remove_internal_deps/num_printer/BUILD.bazel
+++ b/examples/crate_universe/remove_internal_deps/num_printer/BUILD.bazel
@@ -13,7 +13,7 @@ package(default_visibility = ["//visibility:public"])
         ] + all_crate_deps(),
     )
     for key, all_crate_deps in {
-        "remove": all_crate_deps_remove,
         "no_strip": all_crate_deps_no_strip,
+        "remove": all_crate_deps_remove,
     }.items()
 ]
--- a/examples/crate_universe/remove_internal_deps/printer/BUILD.bazel
+++ b/examples/crate_universe/remove_internal_deps/printer/BUILD.bazel
@@ -20,7 +20,7 @@ package(default_visibility = ["//visibility:public"])
         ),
     ]
     for key, all_crate_deps in {
-        "remove": all_crate_deps_remove,
         "no_strip": all_crate_deps_no_strip,
+        "remove": all_crate_deps_remove,
     }.items()
 ]
--- a/examples/crate_universe/remove_internal_deps/rng/BUILD.bazel
+++ b/examples/crate_universe/remove_internal_deps/rng/BUILD.bazel
@@ -14,7 +14,7 @@ package(default_visibility = ["//visibility:public"])
         ]),
     )
     for key, deps in {
-        "remove": crate_deps_remove,
         "no_strip": crate_deps_no_strip,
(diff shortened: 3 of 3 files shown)


crate_universe/src/cli/generate.rs
```diff
@@ -293,3 +320,29 @@ fn write_paths_to_track<
...
+        assert!(filtered_lockfile.packages.len() < original_lockfile.packages.len());
+
+        assert!(original_lockfile
+            .packages
+            .iter()
+            .any(|pkg| pkg.name.as_str() == "child"));
+        assert!(!filtered_lockfile
+            .packages
+            .iter()
+            .any(|pkg| pkg.name.as_str() == "child"));
```
illicitonion, line 346:
```suggestion
        assert!(!filtered_lockfile
            .packages
            .iter()
            .any(|pkg| pkg.name.as_str() == "child"));

        assert!(filtered_lockfile
            .packages
            .iter()
            .any(|pkg| pkg.name.as_str() == "anyhow"));
```

examples/crate_universe/MODULE.bazel
```diff
@@ -345,6 +346,25 @@ use_repo(
     "crate_index_cargo_workspace",
 )
 
+###############################################################################
+# R E M O V E   I N T E R N A L   D E P S
+###############################################################################
+
+# https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html
+crate_index_remove_internal_deps = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
+crate_index_remove_internal_deps.from_cargo(
```
illicitonion, line 356: I'm tempted to add a second `from_cargo` with the un-stripped `Cargo.lock`, and a second `number_printer` target pointing at the unstripped repo, so that we can see both `Cargo.lock` files work and are up to date - WDYT?

PR #3706, opened 2025-10-29
Tell which workspace to clean up

hofbi:
Now the user will see `INFO: Only the workspace's Cargo.toml is required in the `manifests` attribute of workspace `cargo_workspace`; the rest can be removed`

Before it was not clear in which workspace this happened. Might not have been a problem for most users but it helped to cleanup the examples in this repo.


hofbi:
@dzbarsky FYI

PR #3708, opened 2025-10-29
qnx8: add support for qnx8

pawelrutkaq:
This PR adds support for QNX8 since this target is supported from rustc 1.85.0 and from cfg-ext 0.19.0

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -75,6 +75,7 @@ SUPPORTED_T2_PLATFORM_TRIPLES = {
 
 _T3_PLATFORM_TRIPLES = {
     "aarch64-unknown-nto-qnx710": _support(std = True, host_tools = False),
+    "aarch64-unknown-nto-qnx800": _support(std = True, host_tools = False),
     "wasm64-unknown-unknown": _support(std = False, host_tools = False),
 }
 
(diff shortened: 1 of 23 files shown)


pawelrutkaq:
@krasimirgg maybe You could look here and tell me if this is right direction and we can continue or there are some other things one needs to do ?

pawelrutkaq:
Is there a way to run failing workflow locally ? https://buildkite.com/bazel/rules-rust-rustlang/builds/15966/steps/canvas?sid=019a35b7-e271-4132-b386-adbd15454eca  It seems it used rustc 1.85.0 where I understood it shall at least 1.86.0

rust/platform/triple_mappings.bzl
```diff
@@ -75,6 +75,7 @@ SUPPORTED_T2_PLATFORM_TRIPLES = {
 
 _T3_PLATFORM_TRIPLES = {
     "aarch64-unknown-nto-qnx710": _support(std = True, host_tools = False),
+    "aarch64-unknown-nto-qnx800": _support(std = True, host_tools = False),
```
UebelAndre, line 78: What is the difference between `aarch64-unknown-nto-qnx710` and `aarch64-unknown-nto-qnx800`? Can these be matched as different platforms? Or is this a similar issue to https://github.com/bazelbuild/platforms/issues/38 ?

pawelrutkaq: https://github.com/bazelbuild/rules_rust/pull/3723 similar. so in practice thoe shal get some other constrain than cpu and os ones.

PR #3710, opened 2025-10-29
Add `require_explicit_unstable_features` option

FabianWolff:
This new flag can be used to disallow unstable features by default, with the possibility to opt back in to using them by explicitly adding `-Zallow-features=...` to `rustc_flags`.

### DIFF
--- a/util/process_wrapper/options.rs
+++ b/util/process_wrapper/options.rs
@@ -208,6 +208,8 @@ pub(crate) fn options() -> Result<Options, OptionError> {
         child_args,
         &subst_mappings,
         require_explicit_unstable_features,
+        None,
+        None,
     )?;
     // Split the executable path from the rest of the arguments.
     let (exec_path, args) = child_args.split_first().ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


util/process_wrapper/options.rs
```diff
@@ -334,3 +360,33 @@ fn environment_block(
...
+        let subst_mappings: Vec<(String, String)> = vec![];
+        let args = prepare_args(args, &subst_mappings).unwrap();
+        assert_eq!(
+            args,
+            vec![
+                "rustc".to_string(),
+                "-Zallow-features=whitespace_instead_of_curly_braces".to_string(),
+            ]
+        );
+    }
```
krasimirgg, line 431: could you also come up with a test that recurses into `@-param` files?

krasimirgg:
could you address the failing [buildkite/rules-rust-rustlang](https://buildkite.com/bazel/rules-rust-rustlang/builds/15961)

krasimirgg (changes requested):

krasimirgg (approved):
Thank you!

PR #3713, opened 2025-10-30
Added support for `--incompatible_compact_repo_mapping_manifest`

UebelAndre:
Implements support for https://github.com/bazelbuild/bazel/issues/26262


UebelAndre:
@fmeum my understanding of `_repo_mapping` is a bit lacking. Can you confirm I implemented this correctly according to the spec?

rust/runfiles/runfiles.rs
```diff
@@ -140,8 +140,64 @@ enum Mode {
...
+
+    pub fn get(&self, key: &RepoMappingKey) -> Option<&String> {
+        match self {
+            RepoMapping::Compact { exact, prefixes } => {
+                // First try exact match with O(1) hash lookup
+                if let Some(value) = exact.get(key) {
+                    return Some(value);
+                }
+
+                // Then try prefix match with O(n) iteration
```
fmeum: This could optionally use a sorted map since no prefix can be a prefix of another prefix (guaranteed by the spec, see the incompatible issue).

rust/runfiles/runfiles.rs
```diff
@@ -261,10 +318,31 @@ fn parse_repo_mapping(path: PathBuf) -> Result<RepoMapping> {
...
+                target_repo.to_owned(),
+            );
+        }
     }
 
-    Ok(repo_mapping)
+    Ok(if !prefixes.is_empty() {
+        RepoMapping::Compact { exact, prefixes }
+    } else {
+        RepoMapping::Verbose(exact)
```
fmeum: Is this case needed? The compact case is meant to subsume it with an empty prefix list.

rust/runfiles/runfiles.rs
UebelAndre: Switched to a sorted map

rust/runfiles/runfiles.rs
UebelAndre: Good point. Consolidated this.

rust/runfiles/runfiles.rs
fmeum: Sorry, my comment wasn't detailed enough: I meant to say that a sorted map could be used to speed up the lookup to `log(n)` by using a `floor` operation. If you iterate anyway then a list would likely be more efficient.

PR #3723, opened 2025-11-10
Extend QNX support by adding all currently supported variants

darkwisebear:
Add support for QNX8 by introducing a version constraint that allows for the selection of the respective ABI target of the Rust compiler. Since QNX also supports dylibs, also include the nto system as dylib supporting platform.

In addition, also add all supported targets to the list of Tier 3 architectures. Since cargo-bazel also needs to know the additional triples, it's version of cfg-expr needs to be bumped.

### DIFF
--- a/examples/all_deps_vendor/MODULE.bazel
+++ b/examples/all_deps_vendor/MODULE.bazel
@@ -23,7 +23,7 @@ local_path_override(
 # Rust toolchain
 RUST_EDITION = "2021"
 
-RUST_VERSION = "1.79.0"
+RUST_VERSION = "1.86.0"
 
 rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
 rust.toolchain(


rust/platform/BUILD.bazel
```diff
@@ -21,6 +21,26 @@ constraint_value(
     constraint_setting = ":wasi_version",
 )
 
+constraint_setting(
```
UebelAndre, line 24: I would actually like to avoid adding constraint values like this into `rules_rust`. I've been seeing more activity on https://github.com/bazelbuild/platforms/issues/38 and want to use whatever pattern comes from this issue for constraints like this.

UebelAndre (changes requested):
Thanks! I had one comment though and think it will delay this change.

rust/platform/BUILD.bazel
darkwisebear: @UebelAndre Fine for me, I would then simply apply this as a patch for our org as long as there's no decision on the mentioned issue. However, I don't feel like I can contribute much there due to my lack of the overall picture and how things should be. May I ask you to put a comment in the issue about how to move forward and come to a conclusion over there?

pawelrutkaq: @UebelAndre since the activity in the mentioned ticket was stalled for 3 years, is there a chance that we say ie, if nothing moves during next month we can try continue in here ? Because `rules_rust` are important part of bazel and QNX8 is more and more used target - it would make sense to have support for it im my opinion. Otherwise it can happen https://github.com/bazelbuild/platforms/issues/38 will block this next 3 years.

Issue #3307, opened 2025-02-28
rules_rust can't bootstrap using remote execution


rejuvenile:
1. The bootstrap phase can't find Xcode unless `SDKROOT` and `DEVELOPER_DIR` are set as host `action_env`s.
2. The bootstrap phase assumes it can find `rust-objcopy`, which isn't supplied by `rules_rust`'s nightly toolchain.

UebelAndre:
By bootstrap do you mean bootstrap [the process wrapper](https://github.com/bazelbuild/rules_rust/tree/13c596e81ce04b0c82c5cfb3e21f0e395b717cd6/util/process_wrapper)?

> 2\. The bootstrap phase assumes it can find `rust-objcopy`, which isn't supplied by `rules_rust`'s nightly toolchain.

interesting, where does `rust-objcopy` come from?

DoDoENT:
I'm just in the process of setting up remote build for my company and stumbled upon this issue. Whenever there is a remote execution on mac worker, build fails with:

```
(15:52:00) ERROR: /private/var/tmp/_bazel_****/eb0b148894ba6d64194ea7ef59d6bc51/external/rules_rust+/util/process_wrapper/BUILD.bazel:30:36: Compiling Rust (without process_wrapper) bin process_wrapper (6 files) [for tool] failed: (Exit 1): bootstrap_process_wrapper.sh failed: error executing Rustc command (from target @@rules_rust+//util/process_wrapper:process_wrapper) 

  (cd /private/var/tmp/_bazel_****/eb0b148894ba6d64194ea7ef59d6bc51/execroot/_main && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=26.2 \
    CARGO_CFG_TARGET_ARCH=aarch64 \
    CARGO_CFG_TARGET_OS=macos \
    CARGO_CRATE_NAME=process_wrapper \
    CARGO_MANIFEST_DIR='${pwd}/external/rules_rust+/util/process_wrapper' \
    CARGO_PKG_AUTHORS='' \
    CARGO_PKG_DESCRIPTION='' \
    CARGO_PKG_HOMEPAGE='' \
    CARGO_PKG_NAME=process_wrapper \
    CARGO_PKG_VERSION=0.0.0 \
    CARGO_PKG_VERSION_MAJOR=0 \
    CARGO_PKG_VERSION_MINOR=0 \
[...]

PR #3727, opened 2025-11-13
Add rust_objcopy attribute to rust_toolchan

avrabe:
## Summary

Add `rust-objcopy` and `wasm-component-ld` to the `rustc_lib` filegroup in `repository_utils.bzl`. These binaries are present in recent rustc distributions (wasm-component-ld since 1.82.0, rust-objcopy since 1.84.0) but were not previously declared.

### Changes

- **Binary Declaration**: Add `rust-objcopy{binary_ext}` and `wasm-component-ld{binary_ext}` to the rustc_lib filegroup glob patterns

### Context

On Windows, Bazel copies files into the sandbox rather than symlinking, so only explicitly listed files are available. This caused wasm32-wasip2 builds to fail with `linker 'wasm-component-ld.exe' not found`.

On Linux/macOS, the issue was masked by symlink-based sandboxing which allowed rustc to access unlisted files in the same directory.


havasd:
@illicitonion is there anything which prevents this to be merged? this change is useful for macos as well https://github.com/bazelbuild/rules_rust/issues/3307

havasd:
> There's now a `rust-lld` target which has at least `wasm-component-ld` as a runfile. I think this is a better home for that but don't know exactly where objcopy should go. What about it's own filegroup and adding it as it's own optional attribute to `rustc_toolchain`?

I agree with you that it would be good to be a separate attribute of the toolchain for objcopy. It would simplify some parts as we have some custom rules for stripping targets.

Unless @avrabe would do it. I could contribute that change.

UebelAndre:
> > There's now a `rust-lld` target which has at least `wasm-component-ld` as a runfile. I think this is a better home for that but don't know exactly where objcopy should go. What about it's own filegroup and adding it as it's own optional attribute to `rustc_toolchain`?
> 
> I agree with you that it would be good to be a separate attribute of the toolchain for objcopy. It would simplify some parts as we have some custom rules for stripping targets.
> 
> Unless @avrabe would do it. I could contribute that change.

I'd be happy to review that from whoever wanted to put up the PR 😄

avrabe:
Thanks for the feedback @UebelAndre - updated the approach:

- Removed `wasm-component-ld` from `rustc_lib` (now covered by `rust-lld`'s `*-ld` glob)
- Added `rust-objcopy` as its own filegroup + optional toolchain attribute

@havasd thanks for the offer to contribute - I went ahead and made the changes. Hope this looks better now.

avrabe:
Updated with additional fixes for `wasm-component-ld` support:

**Commit 1:** Add `rust-objcopy` as optional toolchain attribute (as requested)

**Commit 2:** Fix sysroot construction for linker tools
- `_symlink_sysroot_tree` was ignoring the `target_files` parameter, using `target.files` instead - this prevented `wasm-component-ld` and `gcc-ld/*` from being symlinked into the sysroot
- The linker was placed at `{sysroot}/bin/` but rustc adds `{sysroot}/lib/rustlib/{host}/bin/` to PATH when invoking linkers ([rustc source](https://github.com/rust-lang/rust/blob/master/compiler/rustc_session/src/filesearch.rs#L55-L58)). Now extracts the correct path from the source location.

`wasm-component-ld` searches PATH for `wasm-ld` or `rust-lld` ([source](https://github.com/bytecodealliance/wasm-component-ld/blob/main/src/lib.rs#L760-L773)), so the linker must be in the directory that rustc adds to PATH.

These fixes enable `wasm-component-ld` to find `rust-lld` on Windows where Bazel's manifest-based runfiles don't provide the same fallback paths as Unix symlinks.

avrabe:
Hi @UebelAndre - I think I accidentally disabled auto-merge when rebasing. All CI is green and I've addressed your feedback. Could you re-enable auto-merge or re-review when you get a chance? Thanks\!

PR #3728, opened 2025-11-14
Added linters to pyo3 tests

UebelAndre:
This runs `black`, `isort`, `mypy`, and `pylint` on all python targets for `PyO3` to keep tests and examples as clean and correct as possible.

Mypy notably tests the stub generation.


.github/workflows/compile_pyo3_dev_requirements.yaml
```diff
@@ -0,0 +1,58 @@
...
+    defaults:
+      run:
+        working-directory: ./extensions/pyo3
+    name: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Compile requirements
+        if: runner.os == 'Windows'
+        run: |
+          bazel run "$env:TARGET" "--" "--upgrade" "--verbose"
```
illicitonion, line 39: Where does `$env:TARGET` come from here? Is this meant to be `${TARGET}`?

extensions/pyo3/tools/requirements/BUILD.bazel
```diff
@@ -0,0 +1,82 @@
...
+    for platform in PLATFORMS
+]
+
+[
+    py_reqs_solution_test(
+        name = "requirements_{}_test".format(platform),
+        compiler = ":requirements.{}.update".format(platform),
+    )
+    for platform in PLATFORMS
+]
```
illicitonion, line 74: Completely optional, but I tend to combine these into a nested list to avoid needing to repeat the loop:

```suggestion
[
    [
        py_reqs_compiler(
            name = "requirements.{}.update".format(platform),
            requirements_in = "requirements.in",
            requirements_txt = "requirements_{}.txt".format(platform),
            target_compatible_with = select({
                ":{}".format(platform): [],
                "//conditions:default": ["@platforms//:incompatible"],
            }),
        ),
        py_reqs_solution_test(
            name = "requirements_{}_test".format(platform),
            compiler = ":requirements.{}.update".format(platform),
        ),
    ]
    for platform in PLATFORMS
]
```

illicitonion (approved):

extensions/pyo3/tools/requirements/BUILD.bazel
UebelAndre: Oh, I'd have done that if automerge didn't see it first 😅

PR #3734, opened 2025-11-19
Update bazel run command for VSCode debug

FaBrand:
The vscod_debug package does not exist.
This target does


FaBrand:
@illicitonion @UebelAndre Could you please take a look?

PR #3737, opened 2025-11-23
feat(rust_common): allow compile actions to declare more outdirs

alexeagle:
This technique is sometimes needed when a proc_macro derives information from the build, and it needs to be consumed by some other tool.

For example this one generates a JSON file as an intermediate output
https://github.com/napi-rs/napi-rs/blob/main/crates/macro/src/expand/typedef/type_def.rs#L11-L12
and then expects developers to call this via a Node.js binary which wraps rustc and also transforms that file to a TypeScript type definition file.

Under Bazel, this is better modeled as a rust_shared_library that produces the binding file (`.so` or `.dylib` for example) along with that JSON output, then run the Node.js binary as a separate target.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,7 +17,6 @@ bazel_dep(name = "rules_cc", version = "0.2.4")
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_shell", version = "0.6.1")
 bazel_dep(name = "apple_support", version = "1.24.1", repo_name = "build_bazel_apple_support")
-
 bazel_dep(name = "rust_toolchains", version = "0.0.1")
 local_path_override(
     module_name = "rust_toolchains",
--- a/rust_toolchains/toolchain/BUILD.bazel
+++ b/rust_toolchains/toolchain/BUILD.bazel
@@ -3,6 +3,9 @@ load(
     "current_rust_analyzer_toolchain",
     "current_rustfmt_toolchain",
 )
+
+# This module is a child of rules_rust so we permit "friendly" visibility
+# buildifier: disable=bzl-visibility
 load(
     "@rules_rust//rust/private:toolchain_utils.bzl",
     "current_rust_toolchain",


alexeagle:
@UebelAndre ping!

alexeagle:
@illicitonion done, thanks PTAL

alexeagle:
Test failures do not look related to my change, FWICT

test/unit/extra_outdirs/extra_outdirs_test.bzl
```diff
@@ -0,0 +1,109 @@
...
+
+    # Get all outputs from the action
+    outputs = rustc_action.outputs.to_list()
+
+    # Check that the expected directories are in the outputs
+    expected_dirs = ctx.attr.expected_outdirs
+    found_dirs = []
+
+    for output in outputs:
+        # Check if this output is a directory (directories don't have extensions)
```
illicitonion: I'm not sure how extensions are relevant here?

test/unit/extra_outdirs/proc_macro.rs
```diff
@@ -0,0 +1,35 @@
...
+// this proc macro has a side-effect of writing extra metadata directories.
+use proc_macro::TokenStream;
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+
+#[proc_macro]
+pub fn write_to_outdirs(_item: TokenStream) -> TokenStream {
+    // Read the list of directories to write to from an environment variable
+    if let Ok(outdirs) = env::var("EXTRA_OUTDIRS") {
```
illicitonion: Let's panic if this isn't set, rather than silently ignoring

test/unit/extra_outdirs/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+        "test_dir",
+        "another_dir",
+    ],
+    proc_macro_deps = [":write_outdirs_macro"],
+    rustc_env = {
+        "EXTRA_OUTDIRS": "test_dir,another_dir",
+    },
+)
+
+extra_outdirs_test_suite(
```
illicitonion: Can you also add an actual test (not unit test) that verifies that `test_dir/marker.txt` has contents `created by proc-macro`?

The unit test shows well that the out directory is a possible output, but I'm not sure how paths actually resolve in all of the different contexts here.

illicitonion (commented):
Generally looks good, thanks! A couple of small things :)

test/unit/extra_outdirs/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+    name = "lib_with_outdirs",
+    srcs = ["lib_with_outdirs.rs"],
+    edition = "2018",
+    extra_outdirs = [
+        "test_dir",
+        "another_dir",
+    ],
+    proc_macro_deps = [":write_outdirs_macro"],
+    rustc_env = {
+        "EXTRA_OUTDIRS": "test_dir,another_dir",
```
UebelAndre: Can you make sure the output dirs support location expansion? That's how I'd expect this to be used.

test/unit/extra_outdirs/BUILD.bazel
alexeagle: Added another test that reads the content of the outdir. It caught that the dummy proc_macro wasn't writing to the location it claimed.

test/unit/extra_outdirs/BUILD.bazel
alexeagle: I added to the docstring:

"The paths are always relative to the output directory of the current Bazel package."

I would not expect them to support location expansion, but only because https://bazel.build/rules/lib/toplevel/attr#output_list doesn't have an equivalent output_dir_list, so there's no way during analysis to resolve such expansions.

Seems like a missing feature in Bazel to me.

UebelAndre: Wait, why is `output_dir_list` needed? Does it matter if you're expanding a file or directory?

alexeagle: Yes, it does matter. The output_list causes Bazel to create the output artifact, then you reference it as `ctx.outputs.my_file` and can't write a directory there, or you confuse the RBE protocol at least

test/unit/extra_outdirs/extra_outdirs_test.bzl
```diff
@@ -0,0 +1,109 @@
...
+        # Check if this output is a directory
+        # and if its basename matches one of our expected directories
+        if output.is_directory:
+            if output.basename in expected_dirs:
+                found_dirs.append(output.basename)
+
+    # Verify all expected directories were found
+    asserts.equals(
+        env,
+        sorted(found_dirs),
```
illicitonion: Could you sort these as variables? Having the assertion operating on different values than the error message presents is confusing.

test/unit/extra_outdirs/BUILD.bazel
```diff
@@ -0,0 +1,54 @@
...
```
illicitonion: If the rules automatically set `EXTRA_OUTDIRS_PATHS` on the `rustc` action's env correctly based on the `extra_outdirs` attr, why do we _also_ need to manually specify the duplicate information here?

alexeagle: you are right, not required here

test/unit/extra_outdirs/extra_outdirs_test.bzl
alexeagle: done

illicitonion:
Ok we're down to a real test failure!

alexeagle:
@illicitonion I'm trying to fix, but I don't know either Rust or Windows and don't have local repro. I believe the bug is in the test, not the three-line prod change, can I just tag the test to skip on windows?

alexeagle:
@illicitonion ping! would like to merge before I give up on contributing...

illicitonion:
Sorry for the delay here - it looks like even with the test skipping, CI was flaking, but hopefully looks good now!

alexeagle:
Here's the real-world macro where I use this feature, as an example:

```
load("@rules_rust//rust:defs.bzl", "rust_shared_library")
load("@bazel_skylib//rules:select_file.bzl", "select_file")
load("@aspect_rules_js//js:defs.bzl", "js_run_binary")

TMP_FOLDER_NAME = "napi_type_def_tmp"

def napi_rust_shared_library(name, deps, **kwargs):
  """Create a NAPI-RS Rust shared library and generate the NAPI type definitions and typegen binary.

  Args:
    name: The name of the target.
    deps: The dependencies for the Rust library. The nodejs headers are added automatically.
    **kwargs: Additional arguments for the rust_shared_library rule.
  """
  lib_target = "_{}_rust_shared_library".format(name)
  bindings_target = "_{}_napi_type_defs".format(name)
  types_target = "_{}_typegen".format(name)
  rust_shared_library(
      name = lib_target,
      extra_outdirs = [TMP_FOLDER_NAME],
      rustc_env = {
          "NAPI_TYPE_DEF_TMP_FOLDER": "$(BINDIR)/{}/{}".format(native.package_name(), TMP_FOLDER_NAME),
      },
      rustc_flags = select({
          "@platforms//os:macos": [
              "--codegen=link-arg=-undefined",
              "--codegen=link-arg=dynamic_lookup",
          ],
          "//conditions:default": [],
      }),
      deps = deps + [
          "@rules_nodejs//nodejs/headers:current_node_cc_headers",
      ],
      **kwargs,
  )

  select_file(
      name = bindings_target,
      srcs = lib_target,
      subpath = TMP_FOLDER_NAME,
  )

  js_run_binary(
      name = types_target,
      srcs = [bindings_target],
      tool = Label("//tools/napi-rs:typegen_bin"),
      args = [
          "$(rootpath {})".format(bindings_target),
          "$(rootpath index.d.ts)",
      ],
      tags = [
          # TODO(alexeagle): continue the struggle against it finding no bindings in the sandbox
          "local",
      ],
      outs = ["index.d.ts"],
  )

  # Final production is a pair of .so/.dylib rust binding and .d.ts type definitions.
  native.filegroup(
    name = name,
    srcs = [
        lib_target,
        types_target,
    ],
  )
```

alexeagle:
I now have to rebase this patch at customer sites to get @blorente fixes, can we merge it please.

alexeagle:
Ping, it has been another month. Do I need to fork?

alexeagle:
@UebelAndre is this being held back for some reason? Daniel is good with it, so can he merge?

Issue #3493, opened 2025-07-03
rustfmt incompatible with `unused-qualifications`


palfrey:
Our repository has `-Dunused-qualifications` in the build flags, and now we can't run rustfmt anymore (well, without some workarounds) because we get lots of errors like the following when running `bazel run @rules_rust//:rustfmt`:

```
error: unnecessary qualification
   --> external/rules_rust+/rust/runfiles/runfiles.rs:158:49
    |
158 |         let mode = if let Some(manifest_file) = std::env::var_os(MANIFEST_FILE_ENV_VAR) {
    |                                                 ^^^^^^^^^^^^^^^^
    |
    = note: requested on the command line with `-D unused-qualifications`
help: remove the unnecessary path segments
    |
158 -         let mode = if let Some(manifest_file) = std::env::var_os(MANIFEST_FILE_ENV_VAR) {
158 +         let mode = if let Some(manifest_file) = env::var_os(MANIFEST_FILE_ENV_VAR) {
    |
```

PR #3738, opened 2025-11-23
Fix unused qualifications in runfiles

palfrey:
Fixes https://github.com/bazelbuild/rules_rust/issues/3493

### DIFF
--- a/rust/runfiles/BUILD.bazel
+++ b/rust/runfiles/BUILD.bazel
@@ -9,7 +9,6 @@ rust_library(
     name = "runfiles",
     srcs = ["runfiles.rs"],
     edition = "2018",
-    rustc_flags = ["-Dunused-qualifications"],
     visibility = ["//visibility:public"],
 )
 
@@ -17,7 +16,6 @@ rust_test(
     name = "runfiles_test",
     crate = ":runfiles",
     data = ["data/sample.txt"],
-    rustc_flags = ["-Dunused-qualifications"],
 )
 
 rust_doc(
--- a/rust/runfiles/runfiles.rs
+++ b/rust/runfiles/runfiles.rs
@@ -1,3 +1,4 @@
+#![deny(unused_qualifications)]
 //! Runfiles lookup library for Bazel-built Rust binaries and tests.
 //!
 //! USAGE:


palfrey:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/bazelbuild/rules_rust/pull/3738/checks?check_run_id=56173725689) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

Fixed

dzbarsky (approved):

palfrey:
> Thanks! Just one question.

What's the question?

rust/runfiles/BUILD.bazel
```diff
@@ -9,13 +9,15 @@ rust_library(
     name = "runfiles",
     srcs = ["runfiles.rs"],
     edition = "2018",
+    rustc_flags = ["-Dunused-qualifications"],
```
UebelAndre: What is this flag? I'm not familiar with `-D` flags for rustc and would love to know if there's more that can be added.

dzbarsky: it's a lint flag, D == Deny

dzbarsky: I think we could also `#deny(all)` in the code, at least for this runfiles lib that will be imported by user code

palfrey: https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html notes the sorts of issues there. I've redone things with `#![deny]` though

dzbarsky:
> I'm not sure we want to sign up to maintain arbitrary style choices for people who depend on rules_rust... How does this code end up triggering these problems for you? As far as I'm aware, both Bazel and Cargo apply different lint expectations to your own code vs your dependencies...

rules_rust emits `--cap-lints=allow` for all crate_universe deps (third-party) but the runfiles lib is not one of those. From bazel's POV it's a "first-party" target even though it comes from an external repo. And this repo never sets [the flag on the rust_library](https://github.com/bazelbuild/rules_rust/blob/cdaf15f5796e3e934b074526272823284bbaed01/rust/runfiles/BUILD.bazel#L8), so tweaking that would be another way to fix this.

Whichever way we move forward with, I would have hoped that a PR doing a bit of warning cleanup/hygiene would not become so contentious.

palfrey:
So, where are we here? How do I get this approved/merged?

PR #3741, opened 2025-11-25
Add runtime_libs as rust compile action inputs

dzbarsky:
The current handling of `static_runtime_lib`/`dynamic_runtime_lib` is broken because it never adds these libs to the action. I think it assumes they are coming from `cc_toolchain.all_files` but that's not necessarily a safe assumption; when these libs are built from source there's no reason to force all actions to depend on both of them via the toolchain files if we may not end up using them. (We hit this scenario in https://github.com/cerisier/toolchains_llvm_bootstrapped)

### DIFF
--- a/test/unit/cc_toolchain_runtime_lib/cc_toolchain_runtime_lib_test.bzl
+++ b/test/unit/cc_toolchain_runtime_lib/cc_toolchain_runtime_lib_test.bzl
@@ -3,7 +3,7 @@ Tests for handling of cc_toolchain's static_runtime_lib/dynamic_runtime_lib.
 """
 
 load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
-load("@rules_cc//cc:cc_toolchain_config_lib.bzl", "feature") 
+load("@rules_cc//cc:cc_toolchain_config_lib.bzl", "feature")
 load("@rules_cc//cc:defs.bzl", "cc_toolchain")
 load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
 load("//rust:defs.bzl", "rust_shared_library", "rust_static_library")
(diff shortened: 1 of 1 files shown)


rust/private/rustc.bzl
```diff
@@ -2253,36 +2261,22 @@ def _add_native_link_flags(args, dep_info, linkstamp_outs, ambiguous_libs, crate
 
     args.add_all(make_link_flags_args, map_each = make_link_flags)
 
-    args.add_all(linkstamp_outs, before_each = "-C", format_each = "link-args=%s")
+    args.add_all(linkstamp_outs, format_each = "-Clink-args=%s")
```
dzbarsky, line 2264: drive-by cleanup

rust/private/rustc.bzl
```diff
@@ -2253,36 +2261,22 @@ def _add_native_link_flags(args, dep_info, linkstamp_outs, ambiguous_libs, crate
...
 
     if crate_type in ["dylib", "cdylib"]:
         # For shared libraries we want to link C++ runtime library dynamically
         # (for example libstdc++.so or libc++.so).
-        args.add_all(
```
dzbarsky, line 2278: all of this is drive-by cleanup; I initially thought they needed to be passed by full path and then discovered they weren't in the sandbox at all :)

test/unit/cc_toolchain_runtime_lib/cc_toolchain_runtime_lib_test.bzl
```diff
@@ -0,0 +1,150 @@
...
+    attrs = {
+        "extra_toolchain": attr.label(),
+        "target": attr.label(cfg = with_extra_toolchain_transition),
+    },
+)
+
+def _inputs_analysis_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    tut = analysistest.target_under_test(env)
+    action = tut[DepActionsInfo].actions[0]
```
UebelAndre, line 63: This is expected to be a rustc action, right?
```suggestion
    action = tut[DepActionsInfo].actions[0]
    asserts.equal(action.mnemonic, "Rustc")
```

dzbarsky: sure, updated

UebelAndre (approved):
Thank you!

PR #3745, opened 2025-11-27
Update cxxbridge-cmd example.

sayrer:
This is just a start. I am not sure how the lockfiles in examples/using_cxx were generated. This does work for me, but I don't have all the lockfiles that the examples do.


sayrer:
I also had to call the `rust_cxx_bridge` macro this way, so that "cxx.h" would be available.

```
rust_cxx_bridge(
    name = "bridge",
    src = "src/lib.rs",
    deps = [
        "//3rdparty/crates:cxx_cc",
    ],
)
```

UebelAndre:
This was done as part of https://github.com/bazelbuild/rules_rust/pull/3867

PR #3749, opened 2025-11-27
Fix linkstamp deps in proc macros

sam-mccall:
These were not being linked in for two reasons:

- missing `_use_grep_includes` attribute on the `rust_proc_macro` rule:
  this attribute & condition are entirely dead code now ==> delete them.
- the explicit crate type list doesn't include `proc-macro` ==> add it.


krasimirgg (approved):
Thank you!

PR #3755, opened 2025-12-02
Added `label_injection` to crate_universe annotations

UebelAndre:
The new `label_injection` attribute is used to sanitize "apparent labels" used in annotations to "canonical labels". A common example of this is when adding external dependencies to `cargo_build_script` targets. E.g.

```starlark
bazel_dep(name = "openssl", version = "3.3.1")

crate = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
crate.annotation(
    crate = "openssl-sys",
    build_script_data = [
        "@openssl//:openssl",
    ],
    build_script_env = {
        "OPENSSL_BIN": "$(execpath @openssl//:openssl)",
    },
    label_injection = {
        "@openssl//:openssl": "@openssl//:openssl",
    },
)
crate.from_cargo(
    name = "crates_index",
)
use_repo(crate, "crates_index")
```

Before this change [inject_repo](https://bazel.build/rules/lib/globals/module#inject_repo) would have been needed to get `@openssl` to resolve in the generated `openssl-sys` crate. After this, labels are remapped after resolving the canonical label based on the `label_injection` attribute so that even if the module above is consumed as a transitive dependency, the modules are still usable (because `inject_repo` is only applied to the root module).


illicitonion (approved):
Very cute!

PR #3759, opened 2025-12-05
Add the `generated_crate_root` attribute to `cargo_build_script`

### LINKED ISSUE #3757: Is there a way to have a `rust_library` where the cargo root is generated with `cargo_build_script`?
The title states most of my question. For example, we can imagine something like this:

```python
cargo_build_script(
    name = "build_script",
    srcs = ["build.rs"],
    compile_data = [":Cargo.toml"],
    crate_root = "build.rs",
)

rust_library(
    name = "example",
    deps = [":build_script"],
    compile_data = ["Cargo.toml"],
)
```

Here, `build_script` produces `lib.rs` that then should become the crate root of `example`.

This simply fails the check in `rust/private/utils.bzl`, line 786:

```
Error in fail: Couldn't find lib.rs or example.rs among `srcs`, please use `crate_root` to specify the root file.
```

AFAUI the build script should go into `deps`, not `srcs`; `build_script` produces a directory which is kind of opaque to Bazel (the below error is with `srcs = [":build_script"]`:

```
Error in add: Cannot add directories to Args#add since they may expand to multiple values. Either use Args#add_all (if you want expansion) or args.add(directory.path) (if you do not).
```

I thought generated crate roots are not at all supported. If so, I'd like your thoughts on how to best add this feature, should `cargo_build_script` declare the files it generates and then pass this information around in the `BuildInfo` provider? Because `crate_root` inside `rust_library_common` is a `File`, not a string, so it has to be declared as an individual file. Alternatively, if `crate_root` can become a string, then maybe `cargo_build_script` could pre-declare its generated cargo root while still declaring a directory 

https://github.com/bazelbuild/rules_rust/blob/4461a8a258a863cc9b933e1277d77a5c85bea2aa/cargo/private/cargo_build_script.bzl#L326

and then this could be taken into account in `rust_library_common`.

Then, while reading the source code I noticed that some support for generated source files is definitely there, e.g.:

https://github.com/bazelbuild/rules_rust/blob/4461a8a258a863cc9b933e1277d77a5c85bea2aa/rust/private/utils.bzl#L849-L888

But it does not seem to support generated crate roots. The word "generated" seems to have two different meanings in this function:

* the first is whether it is a generated file as opposed to a source that returns `True` for `src.is_source`, the variable name `has_generated_sources` employs this meaning;
* the second is whether it is a symlink that is generated in the `transform_sources` function, the variable names `generated_sources`, `generated_compile_data`, and `generated_root` employ the latter meaning.

It would appear from reading the source code that `:build_script` should be added to `compile_data` then at least symlink creation will be initiated. But `rules_rust` still complains about the missing crate root in that case. If crate root is specified as `crate_root = "build_script.out_dir/lib.rs"` then a new errors results:

```
ERROR: One of the output paths 'bazel-out/k8-fastbuild/bin/build_script.out_dir' (belonging to //:build_script) and 'bazel-out/k8-fastbuild/bin/build_script.out_dir/lib.rs' (belonging to //:example) is a prefix of the other. These actions cannot be simultaneously present; please rename one of the output files or build just one of them
```

Specifying `crate_root = "lib.rs"` simply complains about non-existing `//:lib.rs` file, indeed the `crate_root` attribute is a `Label`.

So, I'd like to know how this is supposed to work. I am willing to implement support for this as well.

mrkkrp:
Close #3757.

This change allows us to declare that a build script generates a crate root. Dependent targets created with `rust_library` and `rust_binary` will respect the setting.


mrkkrp:
@illicitonion Does it make sense for me to wait for this to be reviewed, or should I tell my client that this functionality was rejected by `rules_rust` and will have to exist as a patch?

PR #3764, opened 2025-12-09
Use correct env for rust_test

nmattia:
This expands the `rust_test` environment make variables directly instead of using the "build script" helper. The build script helper is incorrect here and this leads issues in Bazel 8.

Note: this was inspired by rules_shell.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -516,14 +516,15 @@ def _rust_test_impl(ctx):
         rust_flags = get_rust_test_flags(ctx.attr),
         skip_expanding_rustc_env = True,
     )
-    data = getattr(ctx.attr, "data", [])
 
-    env = expand_dict_value_locations(
-        ctx,
-        getattr(ctx.attr, "env", {}),
-        data,
-        {},
-    )
+    env = {
+        key: ctx.expand_make_variables(
+            "env",
+            ctx.expand_location(value, ctx.attr.data, short_paths = True),
+            {},
+        )
+        for key, value in ctx.attr.env.items()
+    }
     if toolchain.llvm_cov and ctx.configuration.coverage_enabled:
         if not toolchain.llvm_profdata:
             fail("toolchain.llvm_profdata is required if toolchain.llvm_cov is set.")


nmattia:
@illicitonion it's been a while, though interestingly I cannot reproduce anymore. Something might have changed in Bazel between the version I originally tried (8.X) and two more recent releases I tried which don't require the patch (8.4.2 & 8.5.1). Or maybe I just hallucinated the whole thing :)

I don't think it's a bad idea to reuse the same implementation as `sh_test` though but up to you. Feel free to merge as is or close!

illicitonion:
I'll close for now, but if anyone can repro, please do shout and we can reopen! Thanks!

nmattia:
@illicitonion well, I guess you just had to close the PR for me to run into the issue again 😛 

I've created a repro [here](https://github.com/nmattia/rules_rust/commit/8b252c76a2c91ca943468ef18bc66d569531442d). Without the changes in this PR, the test fails. With the patch, it succeeds. 

The setup is not ideal because I'm pulling the whole wasm test spec (I extracted the repro from [this repo](https://github.com/dfinity/ic)). I'm not familiar with the test setup for rules_rust; can I let you take it from here?

illicitonion:
Thanks for the repro! This was super useful!

The root cause here is that `$(locations)` is unhelpfully defined - sometimes it expands to an execpath and sometimes a rootpath.

In a test, you always want a rootpath. So instead of using the confusingly defined `$(locations)`, if you apply this diff to your branch things work for me:
```diff
diff --git i/test/test_env/BUILD.bazel w/test/test_env/BUILD.bazel
index f2cd0dc8c..8b46d80d0 100644
--- i/test/test_env/BUILD.bazel
+++ w/test/test_env/BUILD.bazel
@@ -73,6 +73,6 @@ rust_test(
     srcs = ["tests/test_env_locations_external.rs"],
     data = ["@wasm_spec_testsuite//:memory64_wast_files"],
     env = {
-        "THE_FILES": "$(locations @wasm_spec_testsuite//:memory64_wast_files)",
+        "THE_FILES": "$(rootpaths @wasm_spec_testsuite//:memory64_wast_files)",
     },
 )
```

Does this work for you?

nmattia:
> Does this work for you?

It does, thank you very much for taking the time to look into this — and confirm it was a chair-to-keyboard interface error :)

PR #3770, opened 2025-12-11
Add common embedded target triple mappings

armandomontanez:
Introduces a variety of common tier-2 embedded baremetal target triple mappings.


armandomontanez:
@konkers @davexroth

armandomontanez:
Failures appear to be some kind of outage.

armandomontanez:
Sorry, had to fix some failures caused by the latest change: ARMv8-M.base was not working because it needed another constraint. Omitted that one for now because I'm not convinced it would be correct to map ARMv8-M.base to the armv8-m constraint in `@platforms`. I need to formally propose a better direction for those constraints, so I'll revisit that target triple when that is resolved.

PR #3774, opened 2025-12-12
Add test sharding support for rust_test

brianduff:
### Problem

`rust_test` targets don't support Bazel's native test sharding. When `shard_count` is set on a `rust_test`:
- Without `--incompatible_check_sharding_support`: all tests run N times (once per shard)
- With `--incompatible_check_sharding_support`: the test fails because Rust's libtest harness doesn't read `TEST_SHARD_INDEX`/`TEST_TOTAL_SHARDS` environment variables

### Solution

This PR adds opt-in sharding support via a new `experimental_enable_sharding` attribute. When enabled, tests are executed via a wrapper script that:

1. Detects sharding via `TEST_TOTAL_SHARDS` environment variable
2. Touches `TEST_SHARD_STATUS_FILE` to advertise sharding support to Bazel
3. Enumerates tests using libtest's `--list --format terse` flag
4. Partitions tests by `index % TEST_TOTAL_SHARDS == TEST_SHARD_INDEX`
5. Runs only the tests assigned to the current shard using `--exact`

### Usage

```starlark
rust_test(
    name = "my_test",
    srcs = ["my_test.rs"],
    shard_count = 4,
    experimental_enable_sharding = True,
)

### DIFF
--- a/rust/private/test_sharding_wrapper.bat
+++ b/rust/private/test_sharding_wrapper.bat
@@ -19,55 +19,100 @@
 @ECHO OFF
 SETLOCAL EnableDelayedExpansion
 
-SET "TEST_BINARY={{TEST_BINARY}}"
+SET TEST_BINARY_RAW={{TEST_BINARY}}
+SET TEST_BINARY_PATH=!TEST_BINARY_RAW:/=\!
+
+@REM Try to find the binary using RUNFILES_DIR if set
+IF DEFINED RUNFILES_DIR (
+    SET TEST_BINARY_IN_RUNFILES=!RUNFILES_DIR!\!TEST_BINARY_PATH!
+    IF EXIST "!TEST_BINARY_IN_RUNFILES!" (
+        SET TEST_BINARY_PATH=!TEST_BINARY_IN_RUNFILES!
+    )
+)
+
+@REM The short_path is like: test/unit/test_sharding/test-2586318641/sharded_test_enabled.exe
+@REM But on Windows, the binary is at grandparent/test-XXX/name.exe (sibling of runfiles dir)
+@REM Extract just the last two components (test-XXX/name.exe)
+FOR %%F IN ("!TEST_BINARY_PATH!") DO SET BINARY_NAME=%%~nxF
+FOR %%F IN ("!TEST_BINARY_PATH!\..") DO SET BINARY_DIR=%%~nxF
+
+@REM Try various path resolutions
+SET FOUND_BINARY=0
+
+@REM Try 1: Direct path (might work in some configurations)
+IF EXIST "!TEST_BINARY_PATH!" (
+    SET FOUND_BINARY=1
+)
+
+@REM Try 2: Grandparent + last two path components
+IF !FOUND_BINARY! EQU 0 (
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rust/private/rust.bzl [hunk 1]
```diff
@@ -516,6 +516,55 @@ def _rust_test_impl(ctx):
...
+    # with a script that handles test enumeration and shard partitioning
+    if ctx.attr.experimental_enable_sharding and ctx.attr.use_libtest_harness:
+        # Find DefaultInfo and CrateInfo in the providers list
+        # DefaultInfo is first, CrateInfo follows (or TestCrateInfo for staticlib/cdylib)
+        default_info = providers[0]
+        default_info_index = 0
+
+        # Get the test binary from CrateInfo - it's the output of the compiled test
+        crate_info_provider = None
+        for p in providers:
```
illicitonion: Let's do something like edit `rustc_compile_action` to return a dict of provider types to providers, and then do a key lookup, rather than sniffing an ordered list.

illicitonion (commented):
This generally looks good to me (though I don't have enough batch skills to really verify the .bat file). One thought on provider organisation :)

rust/private/rust.bzl
brianduff: much better - done.

brianduff:
Remaining test breakages appear to be pre-existing issues unrelated to this change. 

The windows .bat file was a pain to test, but was able to get it to work eventually. I also have a branch where I use a small rust util to launch the tests - lemme know if we should explore that either here or as a followup.

dzbarsky:
I would love to have the util be a rust wrapper. Seems like that would avoid multiple implementations and a bash dependency

rust/private/test_sharding_wrapper.bat
UebelAndre, line 1: I think for cross-platform consistency, could this be a rust binary?

rust/private/rust.bzl
```diff
@@ -918,6 +962,22 @@ _rust_test_attrs = {
             E.g. `bazel test //src:rust_test --test_arg=foo::test::test_fn`.
         """),
     ),
+    "experimental_enable_sharding": attr.bool(
```
UebelAndre, line 965: Should this not be a build flag similar to these?
https://github.com/bazelbuild/rules_rust/blob/cdaf15f5796e3e934b074526272823284bbaed01/rust/settings/BUILD.bazel#L85-L93

shrey4796:
Any update on this ? we could really use this feature

bolinfest:
@brianduff @illicitonion good to see you again! We are starting to rely more heavily on Bazel for building https://github.com/openai/codex/ and are very interested in this change.

I would be happy to help shepherd this through if I can!

dzbarsky:
> @brianduff @illicitonion good to see you again! We are starting to rely more heavily on Bazel for building https://github.com/openai/codex/ and are very interested in this change.
> 
> I would be happy to help shepherd this through if I can!

Our current thinking is the best path forward would be to either patch libtest harness or provide an alternate binary launcher that reuses libtest but adds support for the sharding env vars. (That's basically what Go does as well). Staying in-process should avoid bad interactions with debugger

PR #3775, opened 2025-12-13
chore: use publish to bcr reusable workflow for bcr mirroring

kormide:
Migrate rules_rust off of the deprecated Publish to BCR GitHub app to the newer reusable workflow. The reusable workflow has a number of benefits over the app:
* It supports build attestations which can be enabled at a later time.
* Gives finer grain control over running the publish process—if it fails, you can manually re-run the workflow.
* It's more secure—you control the PAT rather than giving the app write permissions.

Some actions that maintainer for this ruleset will need to take:
* [x] Add a classic PAT named `BCR_PUBLISH_TOKEN` for the repository or org (see [steps](https://github.com/bazel-contrib/publish-to-bcr?tab=readme-ov-file#setup)). I recommend creating a "machine" user for bazelbuild and generating a classic PAT for that user rather than using an individual's PAT. For example, in bazel-contrib we use the [bazel-contrib-bot](https://github.com/bazel-contrib-bot) user to do publishes.
* [ ] Uninstall the GitHub app for this repo.
* [ ] Test a publish after landing, or wait for the next release. I'll be available to help debug any issues that come up if the workflow fails.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -28,6 +28,8 @@ jobs:
   publish:
     uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.1.0
     with:
+      author_name: bazel-io
+      author_email: 5028808+bazel-io@users.noreply.github.com
       attest: false
       draft: false
       tag_name: ${{ inputs.release_version }}


.bcr/config.yml
```diff
@@ -1,6 +1,3 @@
-fixedReleaser:
-  login: scentini
-  email: 11149636+scentini@users.noreply.github.com
```
kormide, line 3: A fixed releaser is no longer required with the workflow because the identity of the publish token is now the actor who opens the PR on the BCR.

kormide:
@UebelAndre @illicitonion

UebelAndre:
@krasimirgg @scentini as Googlers, can you provide the access token?

kormide:
Updated the instructions in the PR description to recommend creating a machine user for bazelbuild and using that user to generate a classic PAT. We do this for bazel-contrib and aspect-build and have associated bot users like https://github.com/bazel-contrib-bot.

meteorcloudy:
We already have the token set up at the org level. rules_cc is using it https://github.com/bazelbuild/rules_cc/blob/01a1eced1d61670e3fe1310ccaaa45635aaf90b1/.github/workflows/publish.yaml#L35

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  publish:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.1.0
+    with:
+      attest: false
+      tag_name: ${{ inputs.release_version }}
+      # Tags don't include a "v" prefix
+      tag_prefix: ""
+      # GitHub repository which is a fork of the upstream where the Pull Request will be opened.
+      registry_fork: scentini/bazel-central-registry
```
meteorcloudy: Use `bazel-io` here

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,41 @@
...
+        required: true
+        type: string
+        description: Release version to publish to the Bazel Central Registry
+      templates_ref:
+        default: ''
+        type: string
+        description: Override the ref to read .bcr templates from
+jobs:
+  publish:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.1.0
```
meteorcloudy, line 29: Use the latest version here?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,41 @@
...
+    with:
```
UebelAndre, line 30: Is there a consistent user we can use for the commit? Since releases are triggered by changes to `version.bzl` it would result in random users being the commit author for the release
```yaml
draft: false
author_name: bazel-io
author_email: bazel-iot@users.noreply.github.com
```

.github/workflows/publish.yaml
kormide: `v1.1.0` is the latest version

.github/workflows/publish.yaml
kormide: Done.

UebelAndre: @meteorcloudy I made this author info up so can you provide the real credentials?

kormide: `curl https://api.github.com/users/bazel-io` shows a `null` email. It's either not public or doesn't have an email because it's a machine user?

kormide: Actually @UebelAndre this is referring to the git committer and author, not who opens the PR. That depends on the PAT used. Which should be bazel-io's?

kormide: Removed the author info.

UebelAndre: @kormide I don't want the author to be a random contributor who happened to trigger the releases. I think it should be a consistent user. A bot preferably but if not that then a Googler.

kormide: Oh, I thought you were concerned about the author of the PR. For the author of the commit, sure, just let me know what to put there and I'll change it.

UebelAndre: cc @scentini who might also be able to provide this info.

meteorcloudy: bazel-io is a bot account, do we have to provide an email address?

UebelAndre: Could probably just use `15028808+bazel-io@users.noreply.github.com` per https://api.github.com/users/bazel-io

UebelAndre:
```suggestion
    with:
      author_name: bazel-io
      author_email: 5028808+bazel-io@users.noreply.github.com
```

meteorcloudy:
Looks good, but can we get the presubmit to pass?

PR #3780, opened 2025-12-15
chore: add insta snapshot example

kormide:
An example for working with [insta](https://crates.io/crates/insta) snapshot tests. Includes a macro that stamps a rust_test along with an executable rule to update snapshots in the source tree.


kormide:
@illicitonion @UebelAndre

kormide:
The CI failures are unclear to me. They don't appear to be caused by this example.

kormide:
> This generally looks reasonable (and useful!), but I'm not super sure of the value of this living in this repo... @UebelAndre any thoughts?

Yeah, it's a bit awkward that this solution is coupled to the nuances of insta. At some point I wanted to to do snapshot tests in rust and I couldn't find any examples. Searching rules_rust issues and examples was the first place I looked, so I figured it might go here.

If not, I'm happy to host it in an example repo of my own.

kormide:
@alexeagle is there a home for this under aspect-build/bazel-examples?

UebelAndre:
My initial thoughts on this are "is this demonstrating something unique that other rules examples do not?" and it seems like this is some handy rule but beyond that seems like a crate specific rule. I could see it becoming a crate_universe example but currently those in particular are more like integration tests that I don't think belong in `examples` (should instead be `test/integration`). Curious what @illicitonion thoughts are on the examples and if they should be cleaned up. If so then there's probably appetite for something like this.

kormide:
Closing as I opened a pr on `aspect-build/bazel-examples` and improved the solution there:

https://github.com/aspect-build/bazel-examples/pull/504

PR #3784, opened 2025-12-18
Properly handle artifact_name_patterns

dzbarsky:
When using cc_common.link, the output's name will default to whatever the cc_toolchain infers from `artifact_name_patterns`. Since rust doesn't respect that, we need to force the name to match the output we want to generate.

Without this, cc_common.link + Windows is sadness


dzbarsky:
> Thanks! Is there a test you can add for this? Or is it specific to mingw toolchains?

OK I don't think there's a lot of value in testing this given everything continues to work and it's more correct by inspection, but added.

PR #3787, opened 2025-12-19
Traverse CrateGroupInfo when gathering CcInfo linking_context entries for use_cc_common_link

### LINKED ISSUE #3786: CrateGroupInfo cannot provide CcInfo dependencies for experimental_cc_common_link
If you use a rust_library_group or rust_prost_library with a rust_binary using experimental_cc_common_link, the resulting link likely fails because the code for collecting the linking contexts for deps doesn't search any crate_group_info instances. rust_prost_library in particular always returns a CrateGroupInfo for its output, linking against it always fails.

ObsidianMinor:
This fixes rust_library_group and rust_prost_library for experimental_use_cc_common_link

Closes #3786


UebelAndre (changes requested):
Thanks! Is there a unit test you could add somewhere in [`//test/unit`](https://github.com/bazelbuild/rules_rust/tree/cdaf15f5796e3e934b074526272823284bbaed01/test/unit) or something? Would love to have some protection against regressions.

ObsidianMinor:
Yeah I was wondering where I should put that, since there's only one test I think actually exercising this experimental flag and the configuration looks mildly complex. But I'll see if I can add something to it.

ObsidianMinor:
Turns out with further testing there's *more* places that don't account for crate groups. So this patch was incomplete and would fix direct dependencies, but not indirect dependencies, which happens during establish_cc_info. I've updated it to factor out the crate group flattening into a separate function plus the added tests.

xiaopeng-tranxmart:
We also encountered the same issue, and this PR works well for us.
Looking forward to it being merged. Thanks for the pr & review work! @UebelAndre

UebelAndre (approved):
In general I think the change looks good, I just want to always make sure there's some regression testing. For the new tests, can you show what the error looks like without the fix applied? I otherwise defer to @krasimirgg to make sure this solves for their needs as well 😄 

Thanks again!

krasimirgg:
Looks very reasonable! +1 for adding some test coverage, as these types of things tend to regress due to "spooky action at a distance" sometimes.

To share some context: internally, we don't use `rust_library_group` at all today, so we don't have any needs around that. We haven't looked at its compatibility with `cc_common.link` at detail. At a high-level, internally we focus on a minimal set of rules and make sure they work well together and are well integrated with a variety of additional internal infrastructure. Onboarding something like `rust_library_group` would have high costs for us for not so much benefit comparatively.

ObsidianMinor:
Is there somewhere else I should be adding coverage that isn't covered by what I've added in this PR?

UebelAndre:
> Is there somewhere else I should be adding coverage that isn't covered by what I've added in this PR?

No, I think you found it but if this feature is ever going to move out of experimental we'll probably need to add unit tests for it at `//test/unit/cc_common_link` or something.

I just want to know what the failure signature looks like without the fix and then see this fix resolve that for now.

UebelAndre (changes requested):
Actually, looking at this change again, it seems to be impacting `rust_library_group` to solve an issue in `experimental_use_cc_common_link`. If you don't mind, could you move [//test/rust_crate_group](https://github.com/bazelbuild/rules_rust/tree/f5b713578a97eca01c449dc211a801c48f2c5ee3/test/rust_crate_group) to `//test/unit/rust_library_group` and add some starlark tests asserting on expectations of the flattened provider?

UebelAndre:
@ObsidianMinor friendly ping here, I'm happy to continue the review here 😄

PR #3788, opened 2025-12-19
Make libclang attr as optional in the rust_bindgen_toolchain

vaibhav-1207:
The `libclang` attr is currently used to set env variables for the underlying `clang-sys` crate to dynamically link libclang. In build systems that use static linking, these are redundant. Making libclang attr optional simplifies the toolchain config for static linking.


vaibhav-1207:
I've updated the description with the rationale. LMK if you have any questions.

UebelAndre (dismissed):
Doesn't bindgen require libclang to work at all? I'm concerned that if it's not provided by the rule then it's picked up on the host. Is this not the case?

Can you also add a test for this?

vaibhav-1207:
> Doesn't bindgen require libclang to work at all?

Yes, bindgen requires libclang to parse headers. However, if the bindgen binary is statically linked with libclang, it is fully self-contained and does not need to load libclang as a shared library at runtime, making the libclang attr redundant.

> I'm concerned that if it's not provided by the rule then it's picked up on the host. Is this not the case?

From the `clang-sys` crate documentation- [link](https://crates.io/crates/clang-sys), lt appears that crate may indeed fallback to searching on the host if the libclang path is omitted for a dynamically linked bindgen binary.

> Can you also add a test for this?

I believe adding a test case would require setting up the build infra for a statically linked bindgen binary, which would be a significant undertaking?

Do you have any suggestions on how else we might document that the libclang attribute is mandatory for dynamically linked systems?

extensions/bindgen/private/bindgen.bzl
```diff
@@ -529,6 +532,7 @@ For additional information, see the [Bazel toolchains documentation](https://doc
             cfg = "exec",
             providers = [CcInfo],
             allow_files = True,
+            mandatory = False,
```
krasimirgg, line 538: Could you add the information from this comment https://github.com/bazelbuild/rules_rust/pull/3788#issuecomment-3705280245 to the `doc` of this attribute.
Two bits specifically are that this being `None` is really only useful if libclang is statically linked in the bindgen binary, and if not -- then the build will not be hermetic and the bindgen will search for and try to use system libclang libraries instead.
There is a similar explanation for libstdcxx below.

krasimirgg:
I think we should be very explicit in the docs that this is really only useful for a custom `bindgen` that was built by statically linking libclang.

UebelAndre:
> I believe adding a test case would require setting up the build infra for a statically linked bindgen binary, which would be a significant undertaking?

Why? Surely it's just a change to some `cargo_build_script` somewhere?

> Do you have any suggestions on how else we might document that the libclang attribute is mandatory for dynamically linked systems?

Is it possible to have a process wrapper query the bidngen binary to see if clang is either statically linked or explicitly available?

extensions/bindgen/private/bindgen.bzl
vaibhav-1207: Done!

vaibhav-1207:
> Why? Surely it's just a change to some cargo_build_script somewhere?

AFAIU, enabling static feature is indeed just a flag, but static linking requires a static archive (libclang.a) to be available during the build. Pre-built LLVM distributions primarily provide shared libraries (libclang.so/.dylib). To support a test case for static linking, we would likely need to build LLVM from source or provision a specific static distribution, both of which are significant changes.

 > Is it possible to have a process wrapper query the bidngen binary to see if clang is either statically linked or explicitly available?

Interesting idea, but a process wrapper cannot easily distinguish between a "statically linked" binary and one that is "dynamically linked to system libraries", without using platform-specific tools (ldd or otool etc.) to inspect dependencies, and that adds a fair amount of complexity.

I've updated the doc for the libclang attribute and its behavior now mirrors that of the libstdcxx attribute.

krasimirgg:
This looks good in order to support our internal use cases where we manually build statically-linked `bindgen`. I think both suggestions to set up a test building statically-linked bindgen and to somehow check if the bindgen binary is statically or dynamically linked be very nice, but as pointed out https://github.com/bazelbuild/rules_rust/pull/3788#issuecomment-3710452904 both present significant technical challenge (we'll need to setup a build of llvm/libclang from source as well; detecting static/dynamic linker requirements is going to be challenging in a portable way).

krasimirgg (approved):

PR #3789, opened 2025-12-19
Load cc symbols for bazel9

dzbarsky:


dzbarsky:
> Thanks! Do you know if buildifier can detect these? Might be good to upgrade if so.

It can detect similar to these but I'm not sure if it's up-to-date with every symbol that moved. Regardless we should upgrade it.

BTW, CI seems red independently of this (across all PRs), should we force-merge this one anyway for now?

PR #3799, opened 2026-01-08
bindgen: remove extra trailing space in check for --rust-edition flags

durin42:
The arguments would be passed as either `--rust-edition=2024` or `--rust-edition`, `2024`, but either way the resulting entry from the list comprehension on the right hand of the `not in` operator will contain `"--rust-edition"` without the trailing space. This allows users of the bindgen rules to specify their own explicit edition.

Tested:
  Works in our private repo, and didn't without this patch.


durin42:
Failure looks strange and probably unrelated? looks like it's trying to do a release action or something on my PR?

durin42:
Oh, I see other failures but they all look weird and environmental to me. Please let me know if I missed a real one in my search.

PR #3801, opened 2026-01-12
Do not include libraries when compiling

hlopko:
In this PR, when in experimental_use_cc_common_link mode, we no longer include linker, runtime libraries, and any transitive libraries to link in the rustc compilation action. This makes incremental rebuilds faster by not recompiling when once of those linker inputs changes.


krasimirgg:
Looks reasonable, but let's see what we can do about the busted CI as that looks like it might be shadowing some test coverage.

krasimirgg (approved):
Thank you

PR #3808, opened 2026-01-16
util/collect_coverage: Allow overriding TARGET_FILE

akhramov:
Currently, the instrumented binary is sourced from the `TARGET_FILE` environment variable. For wrappers around `rust_test`, this variable points to the wrapper rather than the actual binary, which breaks code coverage.

This change:

* Introduces the `RULES_RUST_TEST_BINARY_OVERRIDE` environment variable, which wrapper authors can use to supply the path to the original test binary.

* Uses the runfiles library to resolve `runfiles_dir` instead of relying solely on the `RUNFILES_DIR` variable.

### DIFF
--- a/util/collect_coverage/BUILD.bazel
+++ b/util/collect_coverage/BUILD.bazel
@@ -5,4 +5,5 @@ rust_binary(
     srcs = ["collect_coverage.rs"],
     edition = "2018",
     visibility = ["//visibility:public"],
+    deps = ["@rules_rust//tools/runfiles"],
 )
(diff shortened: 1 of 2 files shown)


akhramov:
Thanks for looking. I owe you a repro on bot of these. Will come up with one soon.

In short,

>  Is RUNFILES_DIR not reliable

in our setup `RUNFILES_PATH` is not set when we run `bazel coverage`, will figure out why first.

> And what is the use of the override?

Wrapper scripts have TARGET_FILE set to the wrapper, not the wrapee, so the script passes the wrong path to `llvm-cov export`

akhramov:
Superseded by https://github.com/bazelbuild/rules_rust/pull/3944

PR #3812, opened 2026-01-20
fix collect coverage script to work with recent changes

anonrig:
Make sure RUNFILES_DIR also tries BAZEL_COVERAGE_INTERNAL_RUNFILES_DIR. This change ensures that we don't throw a fatal error and exit due to RUNFILES_DIR being empty.

In bazel 8, coverage collection script clears RUNFILES_DIR before calling language specific coverage collectors. The rules_rust collect_coverage tool was panicking because it required RUNFILES_DIR to be set.

### DIFF
--- a/util/collect_coverage/collect_coverage.rs
+++ b/util/collect_coverage/collect_coverage.rs
@@ -46,70 +46,88 @@ fn find_metadata_file(execroot: &Path, runfiles_dir: &Path, path: &str) -> PathB
     runfiles_dir.join(path)
 }
 
-fn find_test_binary(execroot: &Path, runfiles_dir: &Path) -> PathBuf {
-    let test_binary = runfiles_dir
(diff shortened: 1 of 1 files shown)


illicitonion:
Sorry for the delay in getting to this!

I can't find any mention of `BAZEL_COVERAGE_INTERNAL_RUNFILES_DIR` in the Bazel source tree (either on master or on 8.5.1), or on Google - can you walk me through what sets this and why it's something we should be checking?

Also it looks like CI is currently failing due to a hopefully-trivial-to-fix clippy issue :)

akhramov:
FWIW In Bazel 9, we also have to set `GENERATE_LLVM_LCOV` 

https://github.com/bazelbuild/bazel/blob/57301af3e3e34253fc1c403e244a3c8ae084c502/tools/test/collect_coverage.sh#L195

PR #3813, opened 2026-01-21
crate_universe: Fix workspace root detection

phlax:
without this, updating the crate triggers a reanalysis, causing recursion and the task to spin out, hang and fail

### DIFF
--- a/crate_universe/extensions.bzl
+++ b/crate_universe/extensions.bzl
@@ -635,7 +635,7 @@ def _generate_hub_and_spokes(
     )
 
     # The workspace root when one is explicitly provided.
-    nonhermetic_root_bazel_workspace_dir = module_ctx.path(Label("@@//:MODULE.bazel")).dirname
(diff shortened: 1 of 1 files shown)


crate_universe/extensions.bzl
```diff
@@ -635,7 +635,7 @@ def _generate_hub_and_spokes(
     )
 
     # The workspace root when one is explicitly provided.
-    nonhermetic_root_bazel_workspace_dir = module_ctx.path(Label("@@//:MODULE.bazel")).dirname
+    nonhermetic_root_bazel_workspace_dir = module_ctx.path(Label("@@//:BUILD")).dirname
```
UebelAndre, line 638: Most repositories I interact with use `BUILD.bazel` as the file name, I think this is going to be less consistent for folks.

UebelAndre (changes requested):
Sorry about the delay here! Can you tell me about the root cause of the issue here? Like what triggers the recursion?

PR #3814, opened 2026-01-21
crate_info: Fix for out_binary

phlax:
this has been an issue for a long time

http://github.com/bazelbuild/rules_rust/pull/1315

and is still an issue

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1728,7 +1728,7 @@ def rustc_compile_action(
             **crate_info_dict
         )
 
-    if crate_info.type in ["staticlib", "cdylib"]:
+    if crate_info.type in ["staticlib", "cdylib"] and not out_binary:
         # These rules are not supposed to be depended on by other rust targets, and
         # as such they shouldn't provide a CrateInfo. However, one may still want to
         # write a rust_test for them, so we provide the CrateInfo wrapped in a provider


PiotrSikora:
@UebelAndre could we please, please get this in? We (Proxy-Wasm Rust SDK, Proxy-Wasm C++ Host, Envoy) have been carrying this patch in every repo for the past 3+ years... While it's not a perfect solution, it really has no downsides, since it only affects `out_binary` feature that's broken without this anyway...

UebelAndre:
Sorry for the delay here, can someone fill me in on the details of this change? What is the failure and what does this solve?

PiotrSikora:
@UebelAndre Wasm modules are the final product, so we're using `rust_binary(out_binary=True)` to create `.wasm` files, since using `rust_library` creates `.rlib` files, which is not what we want.

Building `rust_binary()` without `out_binary` doesn't work, since it's expecting an executable:
```console
$ bazel build --platforms=@rules_rust//rust/platform:wasi //:http_auth_random
ERROR: //BUILD:42:12: in rust_binary rule //:http_auth_random:                                                            
.../external/rules_rust+/rust/private/rustc.bzl:1699:20: The rule 'rust_binary' is executable. It needs to create an executable File and pass it as the 'executable' parameter to the DefaultInfo it returns.                                
ERROR: //BUILD:42:12: Analysis of target '//:http_auth_random' (config: d0d6886) failed 
```

Bulding `rust_binary(out_binary=True)` used to work, but fails since #1298:
```console
$ bazel build --platforms=@rules_rust//rust/platform:wasi //:http_auth_random
ERROR: //BUILD:42:12: in rust_binary rule //:http_auth_random: 
.../external/rules_rust+/rust/private/rust.bzl:251:5: rule advertised the 'CrateInfo' provider, but this provider was not among those returned
ERROR: //BUILD:42:12: Analysis of target '//:http_auth_random' (config: d0d6886) failed
```

Building `rust_binary(out_binary=True)` with this patch (from #1315) builds `.wasm` module just fine:
```console
$ bazel build --platforms=@rules_rust//rust/platform:wasi //:http_auth_random
INFO: Analyzed target //:http_auth_random (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:http_auth_random up-to-date:
  bazel-bin/http_auth_random.wasm
```

(You can replicate this in https://github.com/proxy-wasm/proxy-wasm-rust-sdk by commenting out the `rules_rust.patch` in `MODULE.bazel` and/or `bazel/repositories.bzl` if you want...)

The timeline is:
- 2020-06-05 (#312) - added support for building Wasm modules using `rust_binary(out_binary=True)`,
- 2022-04-28 (#1298) - `out_binary=True` broken by unrelated refactoring,
- 2022-05-03 (#1315) - fix for `out_binary=True` (same as this PR), but the Wasm is declared unsupported (see: [this comment](https://github.com/bazelbuild/rules_rust/pull/1315/changes#r918743390)),
- 2026-01-21 (#3814) - patch resurfaced again, most likely due to bzlmod and the need to ship patched `rules_rust` in BCR.

There is some prior discussion in comments in those issues, as well as in #771.

UebelAndre:
@PiotrSikora Should this not be a `rust_shared_library`?

https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/598fccdd6128c2bdcf0d461d847f490944f50f2c/BUILD#L42-L54

PiotrSikora:
> @PiotrSikora Should this not be a `rust_shared_library`?

Well... That works! Thanks!

I couldn't believe that we wouldn't have tested it before adding `out_binary=True` hack, but it looks that `rust_shared_library` wasn't available back then.

I think we can remove the `out_binary` hack (https://github.com/bazelbuild/rules_rust/pull/3891) and close this one.

PR #3815, opened 2026-01-21
bazel: Fix workspace/bzlmod detection

phlax:
currently it assumes bzlmod if it sees a MODULE.bazel


crate_universe/src/cli/vendor.rs
```diff
@@ -120,6 +120,22 @@ fn bzlmod_tidy(bin: &Path, workspace_dir: &Path) -> anyhow::Result<ExitStatus> {
...
+fn is_bzlmod_enabled(bin: &Path, workspace_dir: &Path) -> bool {
+    let output = process::Command::new(bin)
+        .current_dir(workspace_dir)
+        .arg("info")
+        .arg("starlark-semantics")
+        .output();
+
+    match output {
+        Ok(output) if output.status.success() => {
+            !String::from_utf8_lossy(&output.stdout).contains("enable_bzlmod=false")
```
UebelAndre, line 133: When is this set? I don't see when I run this command in `rules_rust` which should be using bzlmod. Is there maybe something that can be checked in the `crates_vendor` rule that would more accurately detect this? If so that could be added via a new command line arg.

phlax: i think its only available if you have `--noenable_bzlmod` set

UebelAndre: Looks like it!

PiotrSikora:
@phlax it looks that this fails in relevant `//crate_universe:unit_test` (but I don't see the exact reason in the CI logs).

PR #3816, opened 2026-01-21
Switch stamping detection to ctx.configuration.stamp_binaries()

justinhorvitz:
Currently, rules_rust uses a `config_setting` that matches `--stamp=1` to detect whether the `--stamp` option is enabled in the current configuration. This PR removes the `config_setting` in favor of calling `ctx.configuration.stamp_binaries()`, which is restricted, but available to rules rust because it is on the restricted starlark API allowlist: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/BuiltinRestriction.java;l=73;drc=135a03ba5dbad5a1a26ef3bde9815ea244a093d7.

The motivation for making this change is to improve Google-internal detection of stamped actions for cross-user caching eligibility.  The `ctx.configuration.stamp_binaries()` call is limited to the case of `stamp = -1`, whereas the `config_setting` dependency was unconditional. Notably, for `rust_library` rules with the default attribute of `stamp = 0`, we are now able to certify that the rule is not affected by the `--stamp` flag value.

Google-internal bug number: 419546090

### DIFF
--- a/extensions/pyo3/private/pyo3.bzl
+++ b/extensions/pyo3/private/pyo3.bzl
@@ -87,10 +87,43 @@ def _py_pyo3_library_impl(ctx):
     is_windows = extension.basename.endswith(".dll")
 
     # https://pyo3.rs/v0.26.0/building-and-distribution#manual-builds
-    ext = ctx.actions.declare_file("{}{}".format(
-        ctx.label.name,
-        ".pyd" if is_windows else ".so",
-    ))
+    #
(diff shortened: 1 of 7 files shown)


justinhorvitz:
Thanks! How do we go about importing this to google3? Is there a schedule or do we have to do it manually?

krasimirgg:
>Thanks! How do we go about importing this to google3? Is there a schedule or do we have to do it manually?

Let's sync internally.

PR #3819, opened 2026-01-22
fix: support `external_include_paths` in rust_bindgen

lalten:
Include external_includes (from `external_include_paths` cc toolchain feature, see https://github.com/bazelbuild/bazel/pull/13107) in system includes so that external dependency headers are found and treated as system headers.


krasimirgg (approved):
Thank you!

PR #3821, opened 2026-01-23
Add platform triple arm-unknown-linux-musleabi

martin4861:
Add `arm-unknown-linux-musleabi` to the list of supported platform triples in `SUPPORTED_T2_PLATFORM_TRIPLES`.

The `arm-unknown-linux-musleabi` triple is a valid Rust target supported by the official Rust toolchain and is required for building Rust projects targeting ARM devices with musl. This platform triple is also recognized by popular crates in the ecosystem, such as the https://crates.io/crates/nix/0.26.2, which explicitly supports this target as T2.


martin4861:
@illicitonion @UebelAndre would you mind having a look at this PR? Merging this PR would unblock my activities. There is also a corresponding conversion in the `#rust` slack bazel community. Thanks a lot in advance.

PR #3827, opened 2026-01-27
Update build_bazel_apple_support version and checksum

dsharlet:
This version contains https://github.com/bazelbuild/apple_support/commit/44c43c715aa58d16dc713ec0daa0a4373c39245a which fixes missing `LC_UUID` errors when trying to build this (and transitive dependencies) on OSX Tahoe (e.g. https://github.com/google-ai-edge/LiteRT/issues/4727).

Also, I notice this comment: 

```
    # Make the iOS simulator constraint available, which is referenced in abi_to_constraints()
    # rules_rust does not require this dependency; it is just imported as a convenience for users.
```

FWIW, this "convenience" has caused months of pain for many users. This old dependency was very difficult to track down.

### DIFF
--- a/rust/repositories.bzl
+++ b/rust/repositories.bzl
@@ -125,8 +125,8 @@ def rules_rust_dependencies():
     maybe(
         http_archive,
         name = "build_bazel_apple_support",
-        sha256 = "b53f6491e742549f13866628ddffcc75d1f3b2d6987dc4f14a16b242113c890b",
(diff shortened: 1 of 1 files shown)


illicitonion (commented):
Thanks! Please can you also update https://github.com/bazelbuild/rules_rust/blob/2c3469e325fe8d70863247c52cd5ba866311b114/MODULE.bazel#L18 so we're using consistent versions?

Issue #1592, opened 2022-10-12
Pubish runfiles package to crates.io


UebelAndre:
Would it be possible to publish the [runfiles|https://github.com/bazelbuild/rules_rust/blob/main/tools/runfiles/runfiles.rs] package to https://crates.io/ so it doesn't become an unnecessary forcing forcing function between the use of Bazel and Cargo? The library very rarely changes so I think it'd be fine to publish something to maintain interoperability.

UebelAndre:
cc @krasimirgg @scentini

krasimirgg:
I'm a bit doubtful about successfully using this from a cargo context: from the sources it looks like this is very specifically designed for using in bazel environments: https://github.com/bazelbuild/rules_rust/blob/main/tools/runfiles/runfiles.rs#L1. As a risk: if we publish it to crates.io, this may create unreasonable expectations that it works in cargo context.
Otherwise I think it's fine to publish in principle.

Could you say more about how this unnecessarily forces to choose between bazel and cargo? Naively, if I had a project and wanted it to support both cargo and bazel and needed the runfiles functionality for the bazel bits, I'd try and guard the bazel-specific bits that require the runfiles under a source config / conditional compilation or something like that.

fmeum:
While it's true that the runfiles library should change only very rarely, it actually will have to change to accommodate Bzlmod (see https://github.com/bazelbuild/bazel/issues/16124 and the linked proposal). Whatever the outcome of this discussion is, I would urge you to wait until Bazel 6 has been released and the required changes have been implemented.

UebelAndre:
Seems `rules_python` published their runfiles library: 
- https://github.com/bazelbuild/rules_python/pull/1048

I still think this should be a published crate.

liningpan:
I think another benefit is to have it accessible on docs.rs, although pkg.go.dev works differently. 
https://pkg.go.dev/github.com/bazelbuild/rules_go@v0.39.1/go/runfiles

UebelAndre:
@matts1 would you be willing to transfer the [runfiles](https://crates.io/crates/runfiles) crate to @scentini so we can publish this from `rules_rust`?

Related github aciton: https://github.com/bazelbuild/rules_rust/actions/runs/23716384551
```
error: crate runfiles@0.1.0 already exists on crates.io index
```

I don't know if even publishing a new version would be sufficient without so sort of transfer

matts1:
[...]

PR #3830, opened 2026-01-29
Add runfiles crate

dzbarsky:
the runfiles crate on crates.io is 3 years old. This PR allows downstream users to depend on runfiles via a `git` override in their Cargo.toml, so they can build both with Cargo and bazel

closes https://github.com/bazelbuild/rules_rust/issues/1592


rust/runfiles/runfiles.rs
```diff
@@ -44,10 +44,27 @@ const TEST_SRCDIR_ENV_VAR: &str = "TEST_SRCDIR";
 #[macro_export]
 macro_rules! rlocation {
     ($r:expr, $path:expr) => {
-        $r.rlocation_from($path, env!("REPOSITORY_NAME"))
+        {
```
UebelAndre, line 47: What are these changes?

PR #3834, opened 2026-02-02
Configures the stale bot to manage inactive issues and PRs.

meteorcloudy:


.github/workflows/stale.yml
```diff
@@ -0,0 +1,57 @@
...
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 430
+        days-before-pr-stale: 90
+        days-before-issue-close: 90
+        days-before-pr-close: 30
+        stale-issue-message: >
+          Thank you for contributing to the rules_rust repository!
+          This issue has been marked as stale since it has not had any activity in the last 1+ years. It will be closed in the next 90
+          days unless any other activity occurs. If you think this issue is still relevant and should stay open, please post any comment here and the issue will no longer be marked as stale.
+        close-issue-message: >
```
alexeagle, line 42: I originally added the stalebot to rules_python and then later deleted it
https://github.com/bazel-contrib/rules_python/commits/main/.github/workflows/stale.yml.

IMO it's great for us to have a bot label issues as stale, but a human should close them.

The reason we removed it is, when a real issue with the repo isn't being fixed, closing it simply reduces the accuracy of the data - a real issue exists but is no longer documented. It also discourages contributors who experience "well I took the time to report this and no one cares".
It's even worse for PRs.

meteorcloudy: I think it depends on whether the project maintainers are actually triaging issues/PRs.

Assuming they are, then:
- If the issue / PR is unclear and waiting for more information from the reporter/contributor, the stale bot will close them if there is no activity.
- If a real issue exists, then the project maintainers should acknowledge that by adding the `not stale` label or a priority label. 

If there is no one looking at the issues/PRs at all, then nothing will meaningfully help.

PR #3836, opened 2026-02-03
Fixes behavior of rust_generated_srcs within the rust-analyzer aspect

csmulhern:
The existing rust analyzer aspect does not handle generated files completely.

1. The aspect does not propagate through the `srcs` attribute, so on a top level target that uses a generated dependency, the aspect will not run.
2. For targets with transitive dependencies that contain generated rust sources, the aspect will visit those dependencies, but because `rust_generated_srcs` is not aggregated onto the top-level target, the files themselves will not be generated. This breaks things like e.g. jump-to-definition if you try to jump to a file that has not yet been generated by bazel.

This PR addresses both these issues.


csmulhern:
@UebelAndre is there someone that can take a look at this?

csmulhern:
@UebelAndre this is up to date again if there's someone who can take a look.

UebelAndre (approved):
Thank you for the ping! This looks great!

PR #3840, opened 2026-02-03
Fixes rust-analyzer include_dirs for crates with generated sources

csmulhern:
When a crate contains generated sources, `include_dirs` needs to include BOTH the root_module directory, and the generated sources directory.

When `include_dirs` is empty, the inclusion of the root_module directory is done automatically, but when setting `include_dirs`, this needs to be done manually.

See:
https://github.com/rust-lang/rust-analyzer/blob/a84d92ff213e30fb00d7b812e07c4f67e99dcd29/crates/project-model/src/project_json.rs#L123-L131

### DIFF
--- a/rust/private/rust_analyzer.bzl
+++ b/rust/private/rust_analyzer.bzl
@@ -263,7 +263,10 @@ def _create_single_crate(ctx, attrs, info):
         src_map = {src.short_path: src for src in srcs if src.is_source}
         if info.crate.root.short_path in src_map:
             crate["root_module"] = _WORKSPACE_TEMPLATE + src_map[info.crate.root.short_path].path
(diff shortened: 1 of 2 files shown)


csmulhern:
@UebelAndre is there someone that can take a look at this?

csmulhern:
@UebelAndre fyi this is a simple change building on #3836. I've just rebased it so it should be mergable.

csmulhern:
I've rebased this change so it can be cleanly merged.

csmulhern:
@UebelAndre this is up to date again if there's someone who can take a look.

UebelAndre (approved):
Thank you! And sorry for the delay!

Issue #3844, opened 2026-02-05
rust wrappers are non-deterministic


finn-ball:
```
$ bazel clean
$ bazel build //tools/upstream_wrapper:cargo
$ shasum bazel-bin/tools/upstream_wrapper/cargo
d1d010e014b926b0ba8ed90b4776fbff34805a9f  bazel-bin/tools/upstream_wrapper/cargo

$ bazel clean
$ bazel build //tools/upstream_wrapper:cargo
$ shasum bazel-bin/tools/upstream_wrapper/cargo
7b74eedb103f95c518de237607e66e14ea4fa18c  bazel-bin/tools/upstream_wrapper/cargo
```

I can see the sandbox path being used inside the binary:
```
$ strings bazel-bin/tools/upstream_wrapper/cargo | grep "sandbox/"
/sandbox/linux-sandbox/16/execroot/_main/

# run bazel clean and build again

$ strings bazel-bin/tools/upstream_wrapper/cargo | grep "sandbox/"
/sandbox/linux-sandbox/20/execroot/_main/
```

This is the cause:
https://github.com/bazelbuild/rules_rust/blob/f197ab88658216491f2ecc49bfe304baf1525f1e/tools/upstream_wrapper/BUILD.bazel#L24

I can confirm this by printing out the path inside the wrapper:
```
[upstream_wrapper] WRAPPED_TOOL_EXECPATH: /home/finn/.cache/bazel/_bazel_finn/6ec03a67d155658aee9b57f8ad0b7723/sandbox/linux-sandbox/27/execroot/_main/bazel-out/k8-fastbuild/bin/external/+rust+rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/cargo
```

And if you check the command, this is why:
```
WRAPPED_TOOL_EXECPATH='${pwd}/bazel-out/k8-fastbuild/bin/external/+rust+rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/cargo' \
```

PR #3845, opened 2026-02-05
[3844] Use rootpath for wrapper root for determinism

finn-ball:
Fixes https://github.com/bazelbuild/rules_rust/issues/3844

Uses a relative path instead of absolute:
```
WRAPPED_TOOL_EXECPATH: ../+rust+rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/rust_toolchain/bin/cargo
```

### DIFF
--- a/tools/upstream_wrapper/src/main.rs
+++ b/tools/upstream_wrapper/src/main.rs
@@ -20,7 +20,11 @@ fn main() {
             }
             Ok(path)
         })
-        .unwrap_or(PathBuf::from(WRAPPED_TOOL_ROOTPATH));
+        .unwrap_or_else(|_| {
+            std::env::current_dir()
+                .expect("Failed to get current directory")
+                .join(WRAPPED_TOOL_ROOTPATH)
+        });
 
     if !wrapped_tool_path.exists() {
         panic!(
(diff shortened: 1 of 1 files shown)


tools/upstream_wrapper/BUILD.bazel
```diff
@@ -21,7 +21,7 @@ ALL_TOOLS = [target for target in TOOLS.values()]
         data = ALL_TOOLS if tool_name == "cargo" else [target],
         edition = "2018",
         rustc_env = {
-            "WRAPPED_TOOL_EXECPATH": "$(execpath {})".format(target),
+            "WRAPPED_TOOL_EXECPATH": "$(rootpath {})".format(target),
```
katre: Should the variable (here and in main.rs) also be changed? It looks like this is just a fallback for finding the binary when runfiles fails.

finn-ball: Potentially this "fix" could break cross compilation potentially if the target is different to the execution host?

finn-ball: I'm open to ideas on other fixes

PR #3852, opened 2026-02-13
Add 'build_script_exec_properties' to 'crate.annotation'

alloveras:
## Intent
To allow specifying [`exec_properties`](https://bazel.build/reference/be/common-definitions#common.exec_properties) via [`crate.annotation`](https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html#annotation).

## Problem
The current implementation allows tweaking a few of the attributes that get generated for the `cargo_build_script` rule, but, unfortunately, not the `exec_properties`.

My current use case for which I want to be able to specify `exec_properties` is a semi-large build script that gets performance benefits from running on machines with multiple CPUs. If you are not using remote execution, it essentially boils down to whatever hardware you have available locally. However, if RBE is available, being able to set `exec_properties = {"Pool": "<pool-with-powerful-hardware>"}` allows you to specifically steer these build actions to the machines that you want to run them on.

## Solution
To extend the current `cargo-bazel` code along with the relevant Starlark code to make it possible to specify `exec_properties` for the generated `cargo_build_script` rules.


alloveras:
~I think I have to run some magic command(s) to update the lockfile checksums, but I couldn't find the correct incantation anywhere 😢  Otherwise, CI fails complaining about the checksum mismatch.~

~I tried `CARGO_BAZEL_REPIN=1 bazel test //...` but that didnt' seem to help, unfortunately. Will try `./util/repin_all.sh` and report back.~

Edit: It seems like running `./util/repin_all.sh` did the trick 🎉

illicitonion (approved):
Looks reasonable, thanks!

PR #3853, opened 2026-02-13
crate_universe: add features_to_remove to crate.annotation

### LINKED ISSUE #2779: Optional dependencies not respected in bzlmod
Hello,
I have a non cargo setup via bzlmod. When i set `default_features= False`  in crate.spec, the rust_library target generated in the crate's BUILD file still has dependencies that should only be set when that specific feature is present.

robbie-vanderzee:
## Summary

Adds `features_to_remove` to `crate.annotation`, restoring the ability to override Cargo feature unification that was lost when `crate.override` was removed.

## Motivation

Cargo feature unification takes the union of all feature declarations across the dependency graph. When crate A depends on crate B with default features enabled, a top-level `crate.spec(default_features = False)` on B has no effect — Cargo still resolves the full default feature set. Perhaps an override = True could be the solution here too?

This is a known pain point:
- #2779 — no way to disable features pulled in by transitive dependencies
- #1168 — references the lost `features_to_remove` from the old `crate.override` API

**Concrete example:** `miette` declares `syntect = { optional = true }` without `default-features = false`. Syntect's defaults enable `onig`, a C-binding regex engine. This blocks WASM compilation targets entirely — there is no existing mechanism (`crate_features`, `override_targets`, `patches`) that can solve this post-resolution.

## Design

`features_to_remove` is a post-resolution annotation that:

1. Strips the listed features from the resolved `crate_features`
2. Transitively removes features whose enabling parents are **all** in the removal set (e.g., removing `default-onig` and `regex-onig` automatically removes `onig`)
3. Excludes optional dependencies gated exclusively on removed features from `deps`

```python
crate.annotation(
    crate = "syntect",
    features_to_remove = [
        "default",
        "default-onig",
        "regex-onig",
    ],
)
```

## Changes

- **`crate_universe/private/crate.bzl`** — add `features_to_remove` parameter to `_annotation()`
- **`crate_universe/extensions.bzl`** — add `features_to_remove` to `_ANNOTATION_SELECT_ATTRS`
- **`crate_universe/src/config.rs`** — add field to `CrateAnnotations` with union merge semantics
- **`crate_universe/src/select.rs`** — add `retain()` to `Select<BTreeSet<T>>`
- **`crate_universe/src/context/crate_context.rs`** — feature removal logic with transitive orphan computation

## Test plan

Tested against a real workspace where `miette[syntect-highlighter]` pulls syntect with onig defaults:

- `features_to_remove = ["default", "default-onig", "regex-onig"]` on syntect
- Generated BUILD excludes `onig` from both `crate_features` and `deps` (transitive removal)
- WASM target (`wasm32-unknown-unknown`) compiles successfully
- Full native build passes

Closes #2779
Refs #1168


robbie-vanderzee:
This PR was co-authored with Claude Code (claude-opus-4-6, Anthropic). CLA prevents normal attribution.

robbie-vanderzee:
First draft, might need some code quality improvements, but this did in fact solve my issue with the Wasm compilation being messed up by some transitive default features I don't need. Other option is to ask the crate publisher to fix, but I figured since this is a gap in the new api perhaps we should add it.

Might also need a design session to fix it robustly. I am not familiar with this code base, so I don't know the roadmap or a long term elegant solution. Tried my best.

crate_universe/src/context/crate_context.rs
```diff
@@ -894,6 +916,72 @@ impl CrateContext {
     }
 }
 
+fn expand_features(
```
robbie-vanderzee: Not sure I'm a huge fan of what Claude did here. I can reimplement it but wanted to discuss what the team desired and have a proof of solve.

dzbarsky:
I think this cannot be made to work correctly, since the features enabled in a crate unlock dependencies and feature in those, it's a key part of resolution. There is no way to safely remove features.

Also, the existing `features` annotation in the `crate` extension does not work correctly either, as it blindly adds features at the end, at render time, instead of having them participate in resolution.

dzbarsky:
I think a proper solution here would be to override the metadata on the crates themselves as part of resolution. I don't think that's doable (or at least not easily) with the approach rules_rust has taken though.

PR #3857, opened 2026-02-16
RFC: rust_toolchains submodule

dzbarsky:
This unblocks alternative rulesets from providing their own toolchain definitions while reusing rules_rust for compilation. Also it lays groundwork for splitting cargo and clippy to their own toolchains, so we can slim down rustc toolchain and optimize fetches

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -12,10 +12,17 @@ module(
 bazel_dep(name = "bazel_features", version = "1.32.0")
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
 bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rust_toolchains", version = "0.68.1")
 bazel_dep(name = "rules_cc", version = "0.2.4")
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_shell", version = "0.6.1")
 bazel_dep(name = "apple_support", version = "1.24.1", repo_name = "build_bazel_apple_support")
+
+bazel_dep(name = "rust_toolchains", version = "0.0.1")
+local_path_override(
+    module_name = "rust_toolchains",
+    path = "rust_toolchains",
+)
 
 internal_deps = use_extension("//rust/private:internal_extensions.bzl", "i")
 use_repo(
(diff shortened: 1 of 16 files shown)


MODULE.bazel
```diff
@@ -12,11 +12,17 @@ module(
 bazel_dep(name = "bazel_features", version = "1.32.0")
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
 bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rust_toolchains", version = "0.68.1")
```
alexeagle, line 15: Nit: there's an existing convention on BCR to prefix instead, so toolchains_rust ?

rust_toolchains/rustc/BUILD.bazel
```diff
@@ -0,0 +1,5 @@
+package(default_visibility = ["//visibility:public"])
+
+toolchain_type(
```
alexeagle, line 4: Should we comment explicitly that these are all meant to resolve to the exec platform? I dunno why it would be useful to have target-platform for these tools but I can imagine the feature request

MODULE.bazel
dzbarsky: I don't have a super strong opinion on this bikeshed, i just want it built. That said, if you take a look at https://registry.bazel.build/search?q=toolchains you'll see that:
- 5 CC toolchains followed this convention (and toolchains_llvm_boostrapped only did because we were strongly urged, it will get renamed though)
- 5 CC toolchains did `xxx_toolchain` instead
- `proto_toolchains` also followed it, probably because you/Fabian picked the name :)
- No other language did anything like this

So it doesn't feel like we necessarily need to draw a trend from a few isolated points, we should just pick whatever name is clearest IMO

rust_toolchains/rustc/BUILD.bazel
dzbarsky: sure, we can :)

dzbarsky: BTW, I can imagine a scenario where your target platform is meant for execution. Imagine you are building dev VM images or similar and want to package a compiler.

alexeagle: Yes and you could also have clippy run as a test, and use target platform for tests.

dzbarsky: Let's get this landed first to unblock progress and then we can do target-platform toolchains as a followup

MODULE.bazel
alexeagle: okay it alpha sorts closer to "rules" I guess. I'm just looking at https://registry.bazel.build/search?q=toolchains but I don't have a horse in the race.

UebelAndre:
We do not see a reason to do this. Closing out.

Issue #3459, opened 2025-06-02
`cc_common_link` is broken in rust `1.87.0`


abezukor:
Compiling with cc common link is gives the following error when using rust 1.87.0
```
ERROR: /home/abe/.cache/bazel/_bazel_abe/827f63e25b315dd064e3058497da15dd/external/rules_rust+/util/process_wrapper/BUILD.bazel:31:36: Linking external/rules_rust+/util/process_wrapper/process_wrapper [for tool] failed: (Exit 1): gcc failed: error executing CppLink command (from target @@rules_rust+//util/process_wrapper:process_wrapper) /usr/bin/gcc @bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_rust+/util/process_wrapper/process_wrapper-0.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ld.lld: error: undefined symbol: __rustc::__rust_dealloc
>>> referenced by alloc.rs:113 (library/alloc/src/alloc.rs:113)
>>>               alloc-c7b16bb34ad937e6.alloc.4810a64b00f2b3b7-cgu.0.rcgu.o:(core::ptr::drop_in_place$LT$alloc..string..String$GT$::h1891d42eb8d68d4d) in archive bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_rust++rust+rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/lib/rustlib/x86_64-unknown-linux-gnu/lib/liballoc-c7b16bb34ad937e6.a
>>> referenced by gimli.16f37c71821a4789-cgu.0
>>>               gimli-0242466ae973a482.gimli.16f37c71821a4789-cgu.0.rcgu.o:(alloc::collections::btree::node::Handle$LT$alloc..collections..btree..node..NodeRef$LT$alloc..collections..btree..node..marker..Mut$C$K$C$V$C$alloc..collections..btree..node..marker..Leaf$GT$$C$alloc..collections..btree..node..marker..KV$GT$::split::h9f8b8b63bff91341) in archive bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_rust++rust+rust_linux_x86_64__x86_64-unknown-linux-gnu__stable_tools/lib/rustlib/x86_64-unknown-linux-gnu/lib/libgimli-0242466ae973a482.a
[...]

PR #3864, opened 2026-02-23
use an empty staticlib as an allocator library for cc_common.link

krasimirgg:

As suggested on multiple channels, using an empty staticlib is a good effective way to ensure we get the rust allocator shims generated consistently. This is simpler than what we have and also looks like it uniformly supports stable and nightly across versions.

closes https://github.com/bazelbuild/rules_rust/issues/3459


rust/private/rust.bzl
```diff
@@ -1382,6 +1382,26 @@ rust_library_without_process_wrapper = rule(
     ],
 )
 
+def _rust_static_library_without_process_wrapper_impl(ctx):
+    providers = _rust_static_library_impl(ctx)
+    return providers + [_RustBuiltWithoutProcessWrapperInfo()]
+
+rust_static_library_without_process_wrapper = rule(
```
dzbarsky, line 1389: this will need a change to account for https://github.com/bazelbuild/rules_rust/pull/3790

cameron-martin:
Since this now works on stable, could it be enabled by default? It seems like doing so would allow us to remove the cc versions of the allocator libraries.

PR #3867, opened 2026-02-25
Revert "Pin to bazel 8.4.2 (#3804)"

UebelAndre:
This change un-restricts the Bazel version which today means we will be testing against Bazel 9

### DIFF
--- a/cargo/tests/unit/build_script_deps/build_script_deps.bzl
+++ b/cargo/tests/unit/build_script_deps/build_script_deps.bzl
@@ -42,7 +42,11 @@ def _build_script_deps_test_impl(ctx):
         if output.path.endswith(".rlib")
     ][0]
 
-    asserts.true(env, "-exec-" not in rlib_output.path or "-exec/bin/" not in rlib_output.path)
+    asserts.true(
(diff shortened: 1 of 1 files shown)


UebelAndre:
Some tests fail due to https://github.com/bazelbuild/rules_cc/issues/613

cargo/tests/unit/build_script_deps/build_script_deps.bzl
```diff
@@ -42,7 +42,7 @@ def _build_script_deps_test_impl(ctx):
         if output.path.endswith(".rlib")
     ][0]
 
-    asserts.true(env, "-exec-" in rlib_output.path)
+    asserts.true(env, "-exec-" not in rlib_output.path or "-exec/bin/" not in rlib_output.path)
```
illicitonion: This feels like it should be an `and`? Won't an `or` ~always be true?

UebelAndre: Thanks! Updated!

illicitonion (approved):

PR #3870, opened 2026-02-26
Switch pipelining metadata action to hollow rlib (-Zno-codegen)

walter-zeromatter:
Replace the --rustc-quit-on-rmeta / .rmeta approach with Buck2-style hollow rlibs: the RustcMetadata action runs rustc to completion with -Zno-codegen, emitting a .rlib archive. This approach mirrors the one used by buck2 and avoids needing to kill rustc mid-output in order to produce metadata.

While not fixing problems with SVH mismatches when non-determinism, this does simplify the codepath and uses a production tested technique that doesn't have any of the dangers associated with killing the rustc process while it's still active.

Previous work on this branch & the claimed prevention of SVH mismatches proved to be from changes which effectively removed all the actual benefits of the pipelined compilation

### DIFF
--- a/rust/private/clippy.bzl
+++ b/rust/private/clippy.bzl
@@ -177,7 +177,7 @@ def rust_clippy_action(ctx, clippy_executable, process_wrapper, crate_info, conf
         out_dir = out_dir,
         build_env_files = build_env_files,
         build_flags_files = build_flags_files,
-        emit = ["dep-info", "metadata"],
+        emit = ["metadata"],
         skip_expanding_rustc_env = True,
         use_json_output = bool(clippy_diagnostics_file),
         error_format = error_format,
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -50,6 +50,7 @@ load(
     "generate_output_diagnostics",
     "get_edition",
     "get_import_macro_deps",
+    "metadata_output_path",
     "transform_deps",
     "transform_sources",
 )
@@ -206,7 +207,7 @@ def _rust_library_common(ctx, crate_type):
         disable_pipelining = getattr(ctx.attr, "disable_pipelining", False),
     ):
         rust_metadata = ctx.actions.declare_file(
-            paths.replace_extension(rust_lib_name, ".rmeta"),
+            metadata_output_path(toolchain, rust_lib_name),
             sibling = rust_lib,
         )
         rustc_rmeta_output = generate_output_diagnostics(ctx, rust_metadata)
@@ -279,7 +280,7 @@ def _rust_binary_impl(ctx):
     rustc_rmeta_output = None
     if can_build_metadata(toolchain, ctx, ctx.attr.crate_type):
         rust_metadata = ctx.actions.declare_file(
-            paths.replace_extension("lib" + crate_name, ".rmeta"),
+            metadata_output_path(toolchain, "lib" + crate_name),
             sibling = output,
         )
         rustc_rmeta_output = generate_output_diagnostics(ctx, rust_metadata)
@@ -381,7 +382,7 @@ def _rust_test_impl(ctx):
         rustc_rmeta_output = None
         if can_build_metadata(toolchain, ctx, crate_type):
             rust_metadata = ctx.actions.declare_file(
-                paths.replace_extension("lib" + crate_name, ".rmeta"),
+                metadata_output_path(toolchain, "lib" + crate_name),
                 sibling = output,
             )
             rustc_rmeta_output = generate_output_diagnostics(ctx, rust_metadata)
@@ -451,7 +452,7 @@ def _rust_test_impl(ctx):
         rustc_rmeta_output = None
         if can_build_metadata(toolchain, ctx, crate_type):
             rust_metadata = ctx.actions.declare_file(
-                paths.replace_extension("lib" + crate_name, ".rmeta"),
+                metadata_output_path(toolchain, "lib" + crate_name),
                 sibling = output,
             )
             rustc_rmeta_output = generate_output_diagnostics(ctx, rust_metadata)
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -922,7 +922,7 @@ def construct_arguments(
         out_dir,
         build_env_files,
         build_flags_files,
-        emit = ["dep-info", "link"],
+        emit = ["link"],
         force_all_deps_direct = False,
         add_flags_for_binary = False,
         include_link_flags = True,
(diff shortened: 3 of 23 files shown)


scentini:
Hey @walter-zeromatter, we'll need a signed CLA to proceed with the review, could you let us know if that works for you?
I looked at the PR a bit and concluded that it's likely taken from @dzbarsky's work; @dzbarsky is it okay from your side that we review/merge this PR?

dzbarsky:
It's Walter's work. We just merged it to my fork a while ago so the world can benefit from it via rules_rs.

scentini:
Thanks for the clarification! In that case we only need to wait for the CLA.

walter-zeromatter:
Sorry, I've been waiting on our legal. Just got approval, should have the cla signed later today

walter-zeromatter:
@scentini Sorry about the delay, should be all good now.

rust/private/rustc.bzl
```diff
@@ -1377,17 +1428,34 @@ def rustc_compile_action(
...
 
-    # The types of rustc outputs to emit.
-    # If we build metadata, we need to keep the command line of the two invocations
-    # (rlib and rmeta) as similar as possible, otherwise rustc rejects the rmeta as
-    # a candidate.
-    # Because of that we need to add emit=metadata to both the rlib and rmeta invocation.
-    #
-    # When cc_common linking is enabled, emit a `.o` file, which is later
-    # passed to the cc_common.link action.
+    # The main Rustc action uses FULL rlib deps so the full rlib it produces records
```
scentini: Am I understanding this correctly:

- When we compile an `rlib`, we use the `rlib`s of its dependencies as inputs
- When we compile the `rmeta` we use the `rmeta`s of the dependencies as inputs

If so, then where do we use the `rmeta` files? Are you using it only for IDE diagnostics, `cargo check` style?

rust/private/rustc.bzl
```diff
@@ -1336,6 +1381,12 @@ def rustc_compile_action(
         elif ctx.attr.experimental_use_cc_common_link == -1:
             experimental_use_cc_common_link = toolchain._experimental_use_cc_common_link
 
+    # Exec-platform binaries (build scripts) skip cc_common.link: exec-configuration
```
scentini: Could you explain a bit more what the problem with build scripts is? `exec` configuration covers more than `cargo_build_script`, you can have huge tools written in Rust that act as code generators (e.g https://github.com/google/crubit), and disabling pipelining for those can have a significant negative impact to the build time of the critical path.

walter-zeromatter:
Apologies - This code was written some time ago while I was doing some early experimenting with LLMs, and while I do think replacing the -emit=metadata,link + kill rustc approach with -Zno-codegen, A number of the claimed benefits here simply aren't real - I've since learned better, and will be overhauling this PR to something more appropriate. Please hold.

walter-zeromatter:
@scentini This is now a *proper* fix.

Sorry for prior confusion. I do think this is the right approach, and I'll be making a few follow-up PRs to adopt a few other improvements from buck2's rust rules including lightweight metadata for check passes & hopefully some kind of lint check people can run to identify if they have non-deterministic proc macro dependencies.

rust/private/clippy.bzl
```diff
@@ -177,7 +177,7 @@ def rust_clippy_action(ctx, clippy_executable, process_wrapper, crate_info, conf
         out_dir = out_dir,
         build_env_files = build_env_files,
         build_flags_files = build_flags_files,
-        emit = ["dep-info", "metadata"],
```
scentini, line 180: Why are we removing `dep-info`?

rust/private/rust.bzl
```diff
@@ -206,7 +206,7 @@ def _rust_library_common(ctx, crate_type):
         disable_pipelining = getattr(ctx.attr, "disable_pipelining", False),
     ):
         rust_metadata = ctx.actions.declare_file(
-            paths.replace_extension(rust_lib_name, ".rmeta"),
```
scentini, line 209: If the `.rlib` and `_meta.rlib` files are in the same directory, how will the compiler know which one to pick when it encounters a `-Ldependency=blaze-bin/rlib/containing/directory/`? I think this won't necessarily be an issue with sandboxed and remote execution, which may only have the `_meta.rlib` file present, but in local non-sandboxed builds I can see how this can pose a problem.

rust/private/rustc.bzl
```diff
@@ -1494,6 +1461,12 @@ def rustc_compile_action(
     # this is the final list of env vars
     env.update(env_from_args)
 
+    if build_metadata:
+        # RUSTC_BOOTSTRAP=1 is required for -Zno-codegen on stable rustc, and must
+        # be set on both the metadata and full actions for SVH compatibility (since
+        # RUSTC_BOOTSTRAP affects the crate hash).
+        env["RUSTC_BOOTSTRAP"] = "1"
```
scentini, line 1468: This change makes pipelining work only for users that are fine with relying on unstable features. I think that's fine, however I think we should go through the [incompatible change](https://github.com/bazelbuild/rules_rust/blob/main/COMPATIBILITY.md#how-to-make-a-backwards-incompatible-change) process, where we add an `--incompatible_use_unstable_no_codegen_for_pipelining` `build_setting`, and gate the changes in this PR behind it.

rust/private/rustc.bzl
```diff
@@ -1413,17 +1385,12 @@ def rustc_compile_action(
         experimental_use_cc_common_link = experimental_use_cc_common_link,
     )
 
+    compile_inputs_metadata = compile_inputs
```
scentini, line 1388: Seems like `compile_inputs_metadata` and `compile_inputs` are now the same thing. If so, could we get rid of this variable?

rust/private/clippy.bzl
walter-zeromatter: Mostly because it's not necessary - Nothing in rules_rust uses the outputs

rust/private/rust.bzl
walter-zeromatter: Yeah, you're right. I ran into this while playing with an experimental mode to allow incremental builds with local execution & I've pushed a fix.

rust/private/rustc.bzl
walter-zeromatter: Reasonable

rust/private/rustc.bzl
walter-zeromatter: Good catch, ty

walter-zeromatter:
I've updated to use a gating incompatible flag, and also added machinery so that this doesn't automatically enable *all* unstable features for downstream users - if you pass RUSTC_BOOTSTRAP or manually specify -Zallow-features, that behavior should still work

PR #3883, opened 2026-03-04
bindgen: also support -target

divVerent:
https://clang.llvm.org/docs/ClangCommandLineReference.html:

    --target=<arg>, -target <arg>
    Generate code for the given target

Both syntaxes are documented for Clang, so let's support both.


krasimirgg (approved):
thank you!

divVerent:
Just also added tests for this, let's see if they pass.

divVerent:
Yes, looks like everything passed. Hope the tests this way are OK as well.

PR #3890, opened 2026-03-06
Worker-managed pipelined compilation (experimental_worker_pipelining)

walter-zeromatter:
## Summary

Adds `experimental_worker_pipelining`, a new mode that replaces the existing two-invocation hollow-rlib pipelining with a single-rustc-invocation model managed by the persistent multiplex worker — matching how Cargo itself pipelines builds.

## Background

The existing `pipelined_compilation=true` mode (hollow-rlib) runs rustc **twice** per pipelined `rlib`/`lib` crate:
1. `RustcMetadata`: `rustc -Zno-codegen` → hollow `.rlib` (metadata only, ~50ms). Requires `RUSTC_BOOTSTRAP=1`.
2. `Rustc`: full `rustc` → real `.rlib`.

This approach has two known limitations:
- Requires `-Zno-codegen` (unstable flag, hence `RUSTC_BOOTSTRAP=1` on both actions)
- Non-deterministic proc macros (e.g. `ts-rs`, `prost-build`) expand twice with different `HashMap` seeds, producing different SVH values → `E0460` errors

Worker pipelining runs rustc **once** with `--emit=dep-info,metadata,link --json=artifacts`. The worker monitors stderr for the `{"artifact":"…rmeta","emit":"metadata"}` JSON notification, returns the `.rmeta` to Bazel immediately so downstream crates can start, and keeps the rustc process running in the background to finish codegen. The full action retrieves the already-running process and waits for it to finish.

## Performance

Benchmarked 10 iterations, 16 jobs, `tokio + serde_json + hyper + reqwest + cookie_store` (fastbuild, 16-CPU Linux):

```
Config          Mean    Median   Min    Max    Crit Path
─────────────   ─────   ──────   ────   ────   ─────────
no-pipeline     20.7s   20.5s    19.9s  21.9s  19.5s
hollow-rlib     11.4s   11.3s    11.0s  12.1s   9.9s
worker-pipe      8.4s    8.5s     8.0s   8.8s   8.0s
cargo            8.2s    8.1s     7.6s   9.1s     —
```

- **1.35× faster than hollow-rlib** (8.4s vs 11.4s)
- **2.45× faster than no-pipeline**
- **Within ~3% of Cargo's build time** (8.4s vs 8.2s)

The improvement comes from: (1) eliminating the second rustc invocation per crate, saving ~0.5s of per-crate startup overhead across the dependency graph; (2) running 114 worker actions vs 251 sandboxed actions, eliminating all sandbox setup overhead.

### `--worker_max_multiplex_instances` tuning

All multiplex `WorkerProxy` objects for the same worker key share **one underlying process** (`WorkerMultiplexerManager`), so `PipelineState` is always shared regardless of this setting. The flag controls **concurrent requests to that one process**. Bazel's default (8) was empirically optimal on a 16-CPU machine: 8 concurrent rustc processes (~2 CPUs each) outperformed 16 concurrent processes (~1 CPU each) by ~9% due to reduced LLVM context-switching overhead. Recommendation: leave at the default, or set to `HOST_CPUS*.5` on larger machines.

### Worker key sharing (critical design detail)

Bazel worker key = hash of `(mnemonic + executable + startup_args)`. For pipelining to work, the metadata and full actions must reach the **same worker process** (shared `PipelineState`). The metadata action uses mnemonic `"Rustc"` (not `"RustcMetadata"`) when worker pipelining is active, giving both actions an identical key.

The `--pipelining-metadata`, `--pipelining-full`, and `--pipelining-key=<key>` protocol flags live in `rustc_flags` (the per-request `@paramfile`), never in startup args, so they do not affect the worker key.

### Exec-platform exclusion

Build scripts and proc-macros (exec-platform) always use hollow-rlib with `RUSTC_BOOTSTRAP=1` for SVH consistency. Worker pipelining applies only to target-configuration `rlib`/`lib` crates.

### Fallback

If the full action reaches a worker process that doesn't have the background rustc in its `PipelineState` (e.g. after a worker restart or on a cache hit), it falls back to a normal one-shot compilation transparently.

## Usage

```
build --@rules_rust//rust/settings:pipelined_compilation=true
build --@rules_rust//rust/settings:experimental_worker_pipelining=true
build --strategy=Rustc=worker,local
```

## Test plan

- [ ] 8 new unit analysis tests in `pipelined_compilation_test.bzl` covering: metadata action uses `Rustc` mnemonic, multiplex execution requirements, pipelining flags present in rustc args, exec-platform crates excluded, pipeline key format
- [ ] `worker_pipelining_second_lib_test`: integration sh_test verifying a two-crate pipeline compiles correctly end-to-end
- [ ] 51 existing `process_wrapper` tests pass (1 pre-existing failure: `test_process_line_emit_metadata`, unrelated to this change)

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -50,6 +50,7 @@ load(
     "filter_deps",
     "find_toolchain",
     "generate_output_diagnostics",
+    "is_exec_configuration",
     "get_edition",
     "get_import_macro_deps",
     "transform_deps",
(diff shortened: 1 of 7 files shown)


walter-zeromatter:
Appologies, wrong repo targeted

PR #3891, opened 2026-03-07
Remove rust_binary(out_binary=True) hack.

PiotrSikora:
rust_shared_library() should be used instead.

closes https://github.com/bazelbuild/rules_rust/pull/3814


PiotrSikora:
cc @UebelAndre

PR #3895, opened 2026-03-09
Add PWD prefixing for -internal-isystem

keith:
This is a follow up to 0a4285119b21c943560c5aa9dbad3b5ca51f2b2d for
another flag. This one is slightly different since it is wrapped with
`-Xclang` as forwarding args.


keith:
cc @illicitonion @ArchangelX360

UebelAndre (approved):
Thanks!

keith:
thanks!

PR #3896, opened 2026-03-11
Fix self-contained (bundled) linker in hermetic sandbox

trchen1033:
rustc looks for its bundled linker in target-specific path, e.g.

`lib/rustlib/x86_64-unknown-linux-gnu/bin/rust-lld`

But it is installed into `bin/rust-lld` in the generated sysroot instead. This resulted in rustc errors durning linking such as

```
error: the self-contained linker was requested, but it wasn't found in the target's sysroot, or in rustc's sysroot
```

This patch makes rust-lld to be installed in both bin and its original path.


trchen1033:
NVM, this patch is only needed in 0.69 release. This is already fixed in main and the patch will cause a conflict.

PR #3899, opened 2026-03-14
Make test binary names predictable

UebelAndre:
This PR makes `rust_test` binary names and output paths deterministic by removing the hash from the test compilation output directory. Instead of putting test outputs in a hashed subdirectory (e.g. `my/package/test-<hash>/`), the rules now use a predictable naming scheme (e.g. a `_test` suffix on the crate name) so test binaries and their outputs have stable, reproducible paths.

closes https://github.com/bazelbuild/rules_rust/issues/2827


UebelAndre:
Relates to https://github.com/bazelbuild/rules_rust/pull/1600

PR #3346, opened 2025-03-15
Added determinism inspection workflow

UebelAndre:

PR #3904, opened 2026-03-17
Fix reproducible warning in `crates_repository`

UebelAndre:
This fixes a noisy warning for users of `crates_repository`:
```
DEBUG: Rule 'crate_index' indicated that a canonical reproducible form can be obtained by nodifying arguments _action_listener = <unknown object com.google.devtools.build.lib.packages.Attribute$LabelListLateBoundDefault>, _config_dependencies = [], _configure = False, _original_name = ""
```

### DIFF
--- a/crate_universe/private/crates_repository.bzl
+++ b/crate_universe/private/crates_repository.bzl
@@ -139,8 +139,9 @@ def _crates_repository_impl(repository_ctx):
     # Determine the set of reproducible values
     attrs = {attr: getattr(repository_ctx.attr, attr) for attr in dir(repository_ctx.attr)}
     exclude = ["to_json", "to_proto"]
-    for attr in exclude:
-        attrs.pop(attr, None)
+    for attr in list(attrs.keys()):
+        if attr in exclude or attr.startswith("_"):
+            attrs.pop(attr, None)
 
     # Note that this is only scoped to the current host platform. Users should
     # ensure they provide all the values necessary for the host environments


test/determinism/3rdparty/crates/alias_rules.bzl
```diff
@@ -0,0 +1,47 @@
+"""Alias that transitions its target to `compilation_mode=opt`.  Use `transition_alias="opt"` to enable."""
```
illicitonion, line 1:
```suggestion
"""Alias that transitions its target `compilation_mode`."""
```

test/determinism/determinism_tester.rs
```diff
@@ -0,0 +1,555 @@
...
+
+            // Skip any files that end with a given extension.
+            for pattern in IGNORED_EXTENSIONS {
+                if file_name.ends_with(pattern) {
+                    return false;
+                }
+            }
+
+            // Skip any file or directory inside a known volatile directory
+            if entry.path().ancestors().any(|ancestor| {
```
illicitonion, line 279: Optional, but: it would be slightly cheaper to use https://docs.rs/walkdir/latest/walkdir/struct.FilterEntry.html#method.skip_current_dir to stop iterating into those directories, rather than having to check all ancestors for every path

UebelAndre:
I didn't realize https://github.com/bazelbuild/rules_rust/pull/3346 was in this change! It's totally unrelated! So sorry I didn't catch this!

PR #3908, opened 2026-03-19
fix link flags of suffix versioned libs

xiaopeng-tranxmart:
### Problem

When a versioned shared library (e.g. `libcheryl.so.3.8`) is used, the current behavior generates linker flags such as:

* `rustc -dylib=cheryl`
* `gcc -lcheryl`

These flags assume the presence of an unversioned shared library (`libcheryl.so`). As a result, linking fails when only versioned libraries (e.g., `libcheryl.so.3.8`) are available:

```
ld: cannot find -lcheryl
```

The correct linker options in this case should explicitly reference the versioned `.so`:

* `rustc -Clink-arg=-l:libcheryl.so.3.8`
* `gcc -l:libcheryl.so.3.8`

### Solution

IMO, the current implementation assumes that users always provide unversioned shared libraries (`libxxx.so`). This assumption is too restrictive:

1. Some libraries do not provide an unversioned `.so` symlink (e.g., only `libxxx.so.X.Y` exists)
2. Library naming may not follow the `libxxx.so` convention (e.g., `xxx.so.1`)
3. Users may require fine-grained control over linker arguments, especially for ABI-specific or version-pinned dependencies

This PR updates the linking behavior as follows:

* If the library file name matches the conventional form `lib<name>.so`, generate `-ldylib=<name>`
* Otherwise, pass the exact file name to the linker via `-Clink-arg=-l:<filename>`

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -2317,8 +2317,13 @@ def portable_link_flags(
                 "-Clink-arg=-l{}".format(get_lib_name(artifact)),
             ]
     elif _is_dylib(lib):
-        lib_file_name = artifact.basename
-        if lib_file_name.startswith("lib") and lib_file_name.endswith(".so"):
+        if for_windows or for_darwin:
+            use_lib_name = True
+        else:
+            lib_file_name = artifact.basename
+            use_lib_name = (lib_file_name.startswith("lib") and lib_file_name.endswith(".so"))
+
+        if use_lib_name:
             return [
                 "-ldylib=%s" % get_lib_name(artifact),
             ]
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -2317,10 +2317,10 @@ def portable_link_flags(
                 "-Clink-arg=-l{}".format(get_lib_name(artifact)),
             ]
     elif _is_dylib(lib):
+        lib_file_name = artifact.basename
         if for_windows or for_darwin:
             use_lib_name = True
         else:
-            lib_file_name = artifact.basename
             use_lib_name = (lib_file_name.startswith("lib") and lib_file_name.endswith(".so"))
 
         if use_lib_name:
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -2252,6 +2252,7 @@ def portable_link_flags(
         use_pic,
         ambiguous_libs,
         get_lib_name,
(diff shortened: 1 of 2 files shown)


rust/private/rustc.bzl
```diff
@@ -2252,9 +2252,15 @@ def _portable_link_flags(lib, use_pic, ambiguous_libs, get_lib_name, for_windows
                 "-Clink-arg=-l{}".format(get_lib_name(artifact)),
             ]
     elif _is_dylib(lib):
-        return [
-            "-ldylib=%s" % get_lib_name(artifact),
-        ]
+        lib_file_name = artifact.basename
+        if lib_file_name.startswith("lib") and lib_file_name.endswith(".so"):
```
krasimirgg: Could you check how this interacts with `get_lib_name_for_windows`? (I'm a bit worried that this check is overly specific to linux and may regress windows.)

xiaopeng-tranxmart: Good catch. It appears that the library search behavior on Windows and Darwin differs from Linux. Since I’m not familiar with those platforms, this change is currently limited to Linux only.

krasimirgg (approved):
Thank you!

xiaopeng-tranxmart:
Fixed a buildifier lint error.
I’m not sure how to run the tests locally, do we have a recommended way to do that?

xiaopeng-tranxmart:
Friendly ping, this pr is ready

UebelAndre:
ping @krasimirgg

UebelAndre:
It seems like there was a conflict introduced
```
(15:40:39) ERROR: /workdir/rust/private/rustc.bzl:2321:12: name 'for_windows' is not defined
(15:40:40) ERROR: Analysis of target '//test:query_test_binary' failed; build aborted: Error loading '//rust:extensions.bzl' for module extensions, requested by /workdir/MODULE.bazel:46:21: at /workdir/rust/extensions.bzl:4:6: at /workdir/rust/defs.bzl:52:5: compilation of module 'rust/private/rustc.bzl' failed: at /workdir/rust/extensions.bzl:4:6: at /workdir/rust/defs.bzl:52:5: compilation of module 'rust/private/rustc.bzl' failed
```

xiaopeng-tranxmart:
Conflict resolved by adding parameter `for_windows` @krasimirgg @UebelAndre

PR #3911, opened 2026-03-20
Add extra_exec_rustc_flags_triples to toolchain attributes

Ahajha:
We have:
extra_rustc_flags
extra_rustc_flags_triples
extra_exec_rustc_flags
, so rounding out the set. Hit this when passing `-Clink-self-contained=no` to the toolchain, and this is only available on x86_64 rust toolchains.


UebelAndre (approved):
Looks good to me, thank you!

PR #3927, opened 2026-03-24
Implement Bazel path mapping for cross-mode cache sharing

rejuvenile:
## Summary
- Adds `supports-path-mapping` execution requirement to Rustc and CargoBuildScriptRun actions
- Uses File objects in Args instead of path strings so Bazel's path mapper can rewrite output directory prefixes
- Adds `--set-env` flag to process wrapper for path-mapped environment variables
- Updates `get_linker_and_args()` to return the linker File object for path mapping
- Updates `rust_std_paths` to use File depset with `map_each` callback

## Motivation
Bazel's `--experimental_output_paths=strip` enables cross-configuration cache sharing by mapping output paths (e.g. `bazel-out/k8-fastbuild/bin/...` and `bazel-out/k8-opt/bin/...` both map to `bazel-out/cfg/bin/...`). This requires actions to:

1. Declare `supports-path-mapping` in execution requirements
2. Pass all file paths through Args objects (not literal strings) so Bazel can rewrite them

rules_rust was using `.path` strings in many places, preventing path mapping from working. This PR converts the key paths to use File objects in Args:
- `--out-dir` uses `map_each = _file_dirname` callback
- `OUT_DIR` is passed via `--set-env` (a new process wrapper flag) instead of the env dict
- `--codegen=linker=` uses the File object when available
- `-L` (stdlib paths) uses `map_each = _file_dirname` with `uniquify = True`
- `--output-file` uses File objects

Build script environment variables containing file paths are passed through `--extra_env` command-line args (which Bazel path-maps) instead of the action env dict (which Bazel does not path-map).

## Test plan
- [ ] Build with and without `--experimental_output_paths=strip` and verify builds succeed
- [ ] Build with `--compilation_mode=fastbuild` then `--compilation_mode=opt` and verify increased cache hit rate
- [ ] Verify cargo build scripts receive correct OUT_DIR paths

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -53,6 +53,14 @@ load(
     "relativize",
 )
 
+def _file_dirname(file):
+    """Returns the dirname of a file.
+
+    Used as a ``map_each`` callback so that Bazel can apply path mapping
+    (``--experimental_output_paths=strip``) to the resulting string.
+    """
+    return file.dirname
+
 # This feature is disabled unless one of the dependencies is a cc_library.
 # Authors of C++ toolchains can place linker flags that should only be applied
 # when linking with C objects in a feature with this name, or require this
@@ -419,6 +427,7 @@ def get_linker_and_args(ctx, crate_type, toolchain, cc_toolchain, feature_config
     Returns:
         tuple: A tuple of the following items:
             - (str): The tool path for given action.
+            - (File or None): The linker File object for path mapping, or None when the cc_toolchain provides the linker.
             - (bool): Whether or not the linker is a direct driver (e.g. `ld`) vs a wrapper (e.g. `gcc`).
             - (sequence): A flattened command line flags for given action.
             - (dict): Environment variables to be set for given action.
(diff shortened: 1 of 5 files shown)


rust/private/rustc.bzl
```diff
@@ -1524,6 +1544,7 @@ def rustc_compile_action(
             ),
             toolchain = "@rules_rust//rust:toolchain_type",
             resource_set = get_rustc_resource_set(toolchain),
+            execution_requirements = {"supports-path-mapping": "1"},
```
fmeum, line 1547: Should this only be set if `ld_file` is not None?

rust/toolchain.bzl
```diff
@@ -593,7 +593,7 @@ def _rust_toolchain_impl(ctx):
         make_variables = make_variable_info,
         rust_doc = sysroot.rustdoc,
         rust_std = sysroot.rust_std,
-        rust_std_paths = depset([file.dirname for file in sysroot.rust_std.to_list()]),
+        rust_std_paths = sysroot.rust_std,
```
fmeum, line 596: This is now identical to the property above, use that instead at the call site?

fmeum (approved):

UebelAndre:
I believe this was implemented on https://github.com/bazelbuild/rules_rust/pull/4011, let me know if something was missed!

PR #3934, opened 2026-03-27
Support splicing manifests from multiple Cargo workspaces

### LINKED ISSUE #3089: External relative path dependencies not working
With https://github.com/bazelbuild/rules_rust/pull/3025, we ought to be able to use relative path dependencies.
That's not working for me.
You can check my code out in https://github.com/github/codeql/tree/criemen/rules-rust-upg, and run
`bazel run //misc/bazel/3rdparty:vendor_ruby_deps`.
I'm getting
```
INFO: Running command line: bazel-bin/misc/bazel/3rdparty/vendor_ruby_deps.sh

error: failed to load manifest for workspace member `/tmp/.tmpFAhxxI/extractor`
referenced by workspace at `/tmp/.tmpFAhxxI/Cargo.toml`

Caused by:
  failed to load manifest for dependency `codeql-extractor`

Caused by:
  failed to read `/tmp/shared/tree-sitter-extractor/Cargo.toml`

Caused by:

Error: Failed to fetch crates for lockfile: exit status: 101
```

malt3:
When manifests come from different Cargo workspaces (e.g. a primary
workspace depending on crates in a separate external workspace),
splicing previously failed with "manifests are not allowed to come from
different workspaces". Additionally, external relative paths aren't resolved.

This change:
- Adds `main_manifest` to SplicingManifest so the splicer can
  disambiguate which workspace root is primary when multiple are found.
- Nests the spliced workspace at its repo-relative depth inside the temp
  dir so that `../` path deps resolve within the temp dir instead of
  escaping into unwritable system directories.
- Collects path deps from `[workspace.dependencies]` in addition to
  per-package deps.
- Symlinks entire foreign workspace root directories (not just crate
  dirs) so Cargo can walk up and resolve `dep.workspace = true`
  inherited dependencies.
- Wires `main_manifest` through extensions.bzl and crates_vendor.bzl.
- Calls symlink_external_path_deps from the cargo_tree_resolver so
  `cargo tree` also sees external path deps.

This fixes https://github.com/bazelbuild/rules_rust/issues/3089,
along with adding support for multiple workspaces in the splicer.

Co-Authored-By: Steve Barrau <98589981+stevebarrau@users.noreply.github.com>


illicitonion:
I've been talking with @malt3 about this PR offline - we've discussed some approaches that could be helpful to make this review/merge-able - hopefully @malt3 will find some time to pick those changes up at some point :)

malt3:
Indeed. Thanks @illicitonion for discussing better alternatives with me.
That being said, we spent time internally to move away from nested Cargo workspaces. It's unlikely that I'll have time to work on this in the near future.

Let's close this for now, document possible alternatives, and give other users a chance.

Here is what @illicitonion recommended:

> On a quick read, this feels more complicated than I needs to be, but I haven't dug into the issues I'm sure you will have hit when putting it together. But I suspect how much you care about being able to directly address the path dependencies from your bazel workspace may be relevant
> My instinct would be to model this as a separate attribute (non_workspace_manifests) and treating the dependency entirely as a SplicerKind::Workspace that happens to skip some "is this outside the workspace" checks specifically for those manifests that are listed as non_workspace_manifests... Which maybe is actually pretty much what your PR is doing just with the opposite polarity ("main_manifest" vs "non_workspace_manifets")?
> I think the biggest surprise for me is the change to the call to symlink_roots - I would have expected that to have the opposite change - it looks like you're saying "keep symlinking the Cargo.toml file dir, and dance around" whereas I would expect this to have changed to "symlink all of the roots of the whole repo, and start running cargo metadata from inside a subdirectory of the symlink tree"? I would have guessed if we did that (basically reflect the whole repo into the temporary directory where we do splice/resolving) everything else would pretty much fall out for free?

PR #3958, opened 2026-04-09
Update `rust_stdlib_filegroup` to detect `profiler_builtins`

UebelAndre:
This change adds a `coverage_enabled` to `rust_toolchain` that should be used to identify if the toolchain supports coverage. This would allow repositories which use MUSL to run `bazel coverage //...` without running into a linker error. The downside is that they will get no coverage but this would already be the case due to the linker error. More details on https://github.com/rust-lang/rust/issues/79556


UebelAndre:
Friendly ping @krasimirgg

krasimirgg (approved):
Thank you! Looks good!

PR #3963, opened 2026-04-14
Add `unstable_rust_features` attribute

FabianWolff:
This attribute can be set to the label of a target that provides `UnstableRustFeaturesInfo`, which maps a label to a set of unstable features allowed for that label. The intent is to provide a way to centrally manage the use of unstable features, without having to use `-Zallow-features=` manually in `rustc_flags` for every affected target.


rust/private/providers.bzl
```diff
@@ -228,3 +228,8 @@ AllocatorLibrariesImplInfo = provider(
         "static_archive": "Optional[File]: the allocator library archive (typically .a file).",
     },
 )
+
+UnstableRustFeaturesInfo = provider(
+    doc = "A provider that contains a function mapping build targets to unstable features approved for use.",
+    fields = {"unstable_rust_features": "Callable[[Label], List[string]] Returns a list of unstable features approved for use for the given build target."},
```
krasimirgg: document the "all" special case and that this is strictly for nightly compilers.

rust/private/rust.bzl
```diff
@@ -807,6 +807,10 @@ _COMMON_ATTRS = {
     "stamp": _stamp_attribute(
         default_value = 0,
     ),
+    "unstable_rust_features": attr.label(
+        doc = "Controls which unstable features are allowed to be used by this target. Must be None or a target that provides UnstableRustFeaturesInfo.",
+        default = None,
```
krasimirgg, line 814: Add `providers = [UnstableRustFeaturesInfo]` [(bazel docs)](https://bazel.build/rules/lib/toplevel/attr#label:~:text=in%20the%20future.-,providers,-sequence%3B%20default%20is). Then you can remove the bit in comments that talks about the provider.
Instead, document that this is only compatible with nightly compilers.

krasimirgg (commented):
Let's add some tests. Since this requires nightly, testing is a bit awkward:
1. You define a new "integration test" that itself has a top-level `MODULE.bazel` file where you register a nightly toolchain. 
2. Then you update the continuous testing definitions to add a new environment that uses that.
3. The results should immediately show up in CI testing results for the PR.

See [this PR that does something similar](https://github.com/bazelbuild/rules_rust/pull/3403/changes#:~:text=cc_common_link_and_mangling_ubuntu2004).

rust/private/rust.bzl
```diff
@@ -807,6 +808,11 @@ _COMMON_ATTRS = {
     "stamp": _stamp_attribute(
         default_value = 0,
     ),
+    "unstable_rust_features": attr.label(
```
krasimirgg: Hmm, now that I read this, the name now sounds a bit off -- would you consider renaming this to something like `unstable_rust_features_config`? 

Naively, if you consider a rust_library in the wild that sets this and you don't have much context, you'd naturally assume it's something like a list-of-individual unstable features. But this is a different beast, so maybe better to rename it a bit to avoid potential misinterpretation.

test/integration/unstable_rust_features/.bazelrc
```diff
@@ -0,0 +1,39 @@
...
+
+# Enable clippy for all targets in the workspace
+build:clippy --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
+build:clippy --output_groups=+clippy_checks
+
+###############################################################################
+## Incompatibility flags
+###############################################################################
+
+# https://github.com/bazelbuild/bazel/issues/8195
```
krasimirgg, line 5: Does it still work if you remove everything else and just keep this Incompatibility flags section? -- if yes, we might simplify a bit, dropping WORKSPACE.bzl entirely (for context workspace-based stuff is becoming more and more legacy at this point, in favor of bzlmod-first approaches).

krasimirgg (approved):

PR #3964, opened 2026-04-14
Pass all environment variables to vendoring

patrickmscott:
I have been struggling for a long time with getting private registries and git repositories to work with vendoring. CARGO_REGISTR* is currently broken due to a typo. Private git repositories don't work because the ssh auth socket is not forwarded. And custom .ssh/config blow up if all env variables referenced in the config are not forwarded.

There is no great way to forward all the right environment variables. This does not appear to affect repinning directly in bzlmod.


UebelAndre (approved):
Since we can't or don't strip environment on windows, I suppose it's fair to relax the restriction on unix systems. Thanks!

PR #3965, opened 2026-04-15
Replace `new_git_repository` with `git_repository`

willstranton:
The repository rule `new_git_repository` has been an alias of `git_repository` since Bazel 6.0.0:

https://github.com/bazelbuild/bazel/blob/6.0.0/tools/build_defs/repo/git.bzl#L190

Recently, a change was made to make use of `new_git_repository` fail - https://github.com/bazelbuild/bazel/commit/71160cb619b5bfeb89c91fee53bb044438ee9b3f

Users should use `git_repository` instead, which this change does. The change should be a no-op for Bazel >= 6.0.0


UebelAndre (approved):
Thanks!

Issue #2125, opened 2023-08-23
crates_repository lockfile hash doesn't match in nested workspaces


mattbryant-asana:
We have a bazel setup that uses two workspaces, one at the root of our repository and one under `nested/`.  The majority of our code runs under `nested`, including our `rules_rust` configuration, but we need to run things like `@nested//rust_crate:foo` from both the nested and outer workspaces.

When using `crates_repository` with a `lockfile` argument, we observe that the top-level checksum field of the lockfile generated by `CARGO_BAZEL_REPIN=1 bazel sync` is different depending on which workspace we synced from, even though all other contents of the generated lock file are identical.  Because the lockfile is checked on every build of a target, this means that whichever workspace didn't generate it will receive an error about needing `CARGO_BAZEL_REPIN` to be set, making it effectively impossible to use.

The obvious workaround for this is to remove the `lockfile` argument, but it does add some additional overhead for everyone using this since they'll need to do the splicing step.

See attached for a small repro.  After extracting the repro, differences can be observed in the generated lockfiles from the following sequence of commands:
```
cd rules_rust_nested
CARGO_BAZEL_REPIN=1 bazel sync --only=crate_index
head nested/Cargo.Bazel.lock
cd nested
CARGO_BAZEL_REPIN=1 bazel sync --only=crate_index
head Cargo.Bazel.lock
```

Let me know if any other details would be helpful here.

Additional info:
OSes tested on: Mac OS 13.4, Ubuntu 20.04
rules_rust version: 0.26.0

menny:
As an Asanan, I can verify that https://github.com/bazelbuild/rules_rust/pull/3967 fixes the issue described. cc @UebelAndre

PR #3967, opened 2026-04-15
Fix crates_repository cache invalidation on Bazel 8

menny:
## Problem
In Bazel 8, `--incompatible_no_implicit_watch_label` defaults to `true` (bazelbuild/bazel#23861). This means `repository_ctx.path(label)` no longer implicitly watches the resolved file for changes.

`crates_repository` resolves its `cargo_lockfile`, `lockfile`, and `manifests` inputs via `repository_ctx.path(label)` but never explicitly calls `repository_ctx.watch()` on them. As a result, when these files change on disk, Bazel doesn't know it needs to re-fetch the repository — the generated `defs.bzl` becomes stale, causing errors like:

```
      Error in fail: Tried to get all_crate_deps for package <name> but that package had no Cargo.toml file
```

This is especially painful in setups where `crates_repository` inputs come from another external repository (e.g `local_repository`) with a long-running Bazel server. CI is unaffected because it always starts with clean state. The only workaround today is `bazel clean` or `bazel sync --only=<repo>`.

## Fix
Add explicit `repository_ctx.watch()` calls for `cargo_lockfile`, `lockfile` (optional, guarded by `if lockfiles.bazel`), and each entry in `manifests`, immediately after `get_lockfiles()` returns in `_crates_repository_impl`.

## Testing

Verified in a large monorepo (Bazel 8.6.0, rules_rust 0.66.0) with a nested workspace where `crates_repository` inputs come from a `local_repository`. Before the fix, changing `Cargo.toml` or lock files required `bazel clean` to pick up changes. After the fix, Bazel automatically re-fetches the crate repositories.

## Related
- bazelbuild/rules_rust#2125 — related lockfile hash mismatch in nested workspaces
- https://github.com/bazelbuild/bazel/issues/23861 — the Bazel flag change that caused this

### DIFF
--- a/crate_universe/private/crates_repository.bzl
+++ b/crate_universe/private/crates_repository.bzl
@@ -49,6 +49,13 @@ def _crates_repository_impl(repository_ctx):
     # Locate the lockfiles
     lockfiles = get_lockfiles(repository_ctx)
 
+    # Watch lockfiles and manifests for changes.
+    repository_ctx.watch(lockfiles.cargo)
+    if lockfiles.bazel:
+        repository_ctx.watch(lockfiles.bazel)
+    for m in repository_ctx.attr.manifests:
+        repository_ctx.watch(repository_ctx.path(m))
+
     # Locate Rust tools (cargo, rustc)
     tools = get_rust_tools(repository_ctx, host_triple)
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thanks!

PR #3969, opened 2026-04-15
crate_universe release 0.18.0

UebelAndre:


illicitonion (approved):
I'm not actually super clear on why we release this independently, but 🤷 sure!

PR #3970, opened 2026-04-15
[codex] rust: fix TESTBRIDGE sharding filters

sluongng:
## Summary
- fix the sharding launcher so Bazel's `TESTBRIDGE_TEST_ONLY` filter wins over legacy positional libtest filters
- split single-colon multi-test filters while preserving Rust `::` test names
- include the rest of the current worktree as requested, including regenerated `MODULE.bazel.lock` files and the local `.profraw` artifact

## Root Cause
The launcher kept positional libtest filters and `TESTBRIDGE_TEST_ONLY` in the same lane. That let `--test_arg` silently override Bazel's top-level `--test_filter`, and it treated multi-test `TESTBRIDGE_TEST_ONLY` values as one literal string instead of multiple requested tests.

## Testing
- `bazel test //rust/private/test_sharding:launcher_test --test_output=errors`
- `bazel test //test/unit/test_sharding:forced_sharding_test --nocache_test_results --test_output=streamed --test_arg=--list --test_filter='test_7:test_8'`
- `bazel test //test/unit/test_sharding:forced_sharding_test --nocache_test_results --test_output=streamed --test_filter=test_8 --test_arg=test_1 --test_arg=--list`


sluongng:
this is very much WIP. Pushed for reference only

Please DO NOT review 🙏

PR #3975, opened 2026-04-17
cargo-bazel: fix binary targets of proc-macro crates using wrong dep attribute

lamcw:
## Problem

When a crate has both a `rust_proc_macro` library target and `rust_binary` targets (e.g. development utilities shipped alongside the proc-macro), `make_rust_binary` unconditionally places the library in `deps`. However, rules_rust validates that proc-macro libraries must appear in `proc_macro_deps`, not `deps`. This causes analysis failures for any such crate.

## Solution

In `make_rust_binary`, detect whether the crate's library target is a proc-macro by checking if `Rule::ProcMacro(_)` is present in `krate.targets`. If it is, insert the library label into `proc_macro_deps` instead of `deps`.

## Testing
- `binary_of_proc_macro_crate_uses_proc_macro_deps`: verifies that the generated `rust_binary` has the `proc-macr`o lib in `proc_macro_deps` and not in `deps`.

### DIFF
--- a/crate_universe/src/rendering.rs
+++ b/crate_universe/src/rendering.rs
@@ -2335,17 +2335,15 @@ mod test {
             .expect("should contain rust_binary");
         let binary_section = &build_file_content[binary_start..];
 
-        // Locate the deps and proc_macro_deps attribute positions within the binary.
-        let deps_pos = binary_section
-            .find("    deps = ")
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thank you!

PR #3976, opened 2026-04-17
Fix parse_cargo_tree_output misclassifying dev-dependencies as host deps

jonremy:
Fix `parse_cargo_tree_output` incorrectly classifying `[dev-dependencies]` as host dependencies when they follow `[build-dependencies]` for the same crate.

This causes platform-conditional features to leak across platforms in the generated BUILD files. For instance, the `tokio` feature `taskdump` that can only be enabled on linux could leak to other platforms.

Example dependency tree that triggers the issue:
```
;my-crate v0.1.0 (/my-crate);;
`-- ;tokio v1.49.0;bytes,full,taskdump;
[build-dependencies]
`-- ;anyhow v1.0.100;default,std;
[dev-dependencies]
`-- ;tokio-test v0.4.4;;
    `-- ;tokio v1.49.0;bytes,full,taskdump; (*)
```

This PR also updates surrounding code in `parse_cargo_tree_output`. Let me know if you'd prefer a minimal fix instead.

### DIFF
--- a/crate_universe/src/metadata/cargo_tree_resolver.rs
+++ b/crate_universe/src/metadata/cargo_tree_resolver.rs
@@ -405,7 +405,7 @@ impl TreeResolver {
 
                 for (entry, tree_data) in target_tree_data {
                     metadata
-                        .entry(entry.as_crate_id().clone())
+                        .entry(entry.clone())
                         .or_default()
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Thank you!

Issue #3985, opened 2026-04-22
rust-analyzer discoverConfig setup doesn't resolve core libraries correctly


Danielkonge:
I have a problem where even simple rust core functionality like `Option`, `None`, `Some` doesn't work correctly with basically the auto discovery setup from https://bazelbuild.github.io/rules_rust/rust_analyzer.html. rust-analyzer doesn't seem to recognize e.g. `None` without an explicit import.

**Reproduction:** 
- Clone https://github.com/bazel-starters/rust#
- Update the rust version to something new in `MODULE.bazel` (mainly because of some editor setup when I tried this), e.g.:
```bazel
bazel_dep(name = "rules_rust", version = "0.69.0")
bazel_dep(name = "rules_rs", version = "0.0.26")

rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
rust.toolchain(
    edition = "2024",
    versions = ["1.93.1"],
)
```
- Create a `hello_world/` folder containing:

`main.rs`:
```rust
// use core::option::Option::None;

#[allow(dead_code)]
fn opt_match(a: Option<u8>) -> u8 {
    match a {
        Some(a) => a,
        None => 0,
    }
}

fn main() {
    println!("Hello, world!");
}
```

`BUILD.bazel`:
```bazel
load("@rules_rust//rust:defs.bzl", "rust_binary")

rust_binary(
    name = "hello_world",
    srcs = ["main.rs"],
)
```

- Setup your editor similar to the suggested setup. E.g. something like:
```json
"rust-analyzer": {
    "workspace": {
        "discoverConfig": {
            "command": ["discover_bazel_rust_project.sh", "{arg}"],
            "progressLabel": "rust_analyzer",
            "filesToWatch": ["BUILD", "BUILD.bazel", "MODULE.bazel"]
        }
    }
}
```
where `discover_bazel_rust_project.sh`:
```bash
#!/usr/bin/env bash

RA_CACHE_OUTPUT_BASE=".rust_analyzer_cache"

bazel \
    --output_base="${RA_CACHE_OUTPUT_BASE}" \
    run \
    @rules_rust//tools/rust_analyzer:discover_bazel_rust_project -- \
    --bazel_startup_option=--output_base="${RA_CACHE_OUTPUT_BASE}" \
    --bazel_arg=--output_groups=-clippy_checks \
    --bazel_arg=--watchfs \
    ${1:+"$1"} 2>/dev/null
```

[...]

PR #3986, opened 2026-04-22
Fix rust-analyzer for core libraries

Danielkonge:
For a description of the problem see: #3985

This fixes the dependency problem for the core libraries for me, but I am not sure if this is the best solution. It might be better to add proper BUILD files describing the core dependencies instead. 

(As an aside: It might be nice with some clippy support in rust-analyzer too now that I am adding cargo anyways.)


Danielkonge:
I signed the CLA now.

DanielKongsgaard:
@UebelAndre could I get a review on this? 

I think the `rust_stdlib` stuff can be removed, if you prefer, but it really needs cargo to have rust-analyzer understand the standard lib dependencies based on what I have seen. (See https://github.com/bazelbuild/rules_rust/issues/3985)

[EDIT: Noticed I used the wrong GitHub account, but I will just leave it.]

PR #3997, opened 2026-04-28
Fix platform label context in rule transitions

tamasvajk:
## Summary

Fix platform label resolution in rule transitions for `rust_binary`, `rust_shared_library`, `rust_static_library`, and `rust_test`.

## Problem

When a platform label comes from the main repository (no repo name), `str(label)` produces `//platforms:foo` instead of `@//platforms:foo`. Bazel's `//command_line_option:platforms` setting requires the canonical form with `@`, so the transition silently selects the wrong platform or fails.

This affects anyone using the `platform` attribute on Rust rules with a label defined in the main repository.

Note: as mentioned in https://github.com/bazelbuild/rules_rust/pull/3997#discussion_r3156712697, this only affects builds that use `--noincompatible_unambiguous_label_stringification`.

## Fix

Extract a `_resolve_platform` helper that adds the `@` prefix when `attr.platform.repo_name` is empty, and apply it consistently to all four rule transitions. The original patch only fixed `rust_binary`; this PR fixes all four for completeness.

---

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -1013,7 +1013,7 @@ def _resolve_platform(settings, attr):
     main-repo labels omits the leading @, producing "//foo:bar" instead of
     "@//foo:bar". The platform setting requires the canonical form with @.
     See https://github.com/bazelbuild/bazel/issues/15916.
-    
+
     Note that this function will no longer be needed if
     `--noincompatible_unambiguous_label_stringification` is dropped but
     it's currently required internally by Google.
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -1006,9 +1006,23 @@ rust_library = rule(
         """),
 )
 
+def _resolve_platform(settings, attr):
+    """Resolve the platform label for a transition, adding @ prefix if needed.
+
+    When a label has no repo name (e.g. from the main repo), str(label) omits
+    the leading @, producing "//platforms:foo" instead of "@//platforms:foo".
+    Bazel's platform setting requires the canonical form with @.
+    """
+    if not attr.platform:
+        return settings["//command_line_option:platforms"]
+    platform = str(attr.platform)
+    if not attr.platform.repo_name:
+        platform = "@" + platform
+    return platform
+
 def _rust_static_library_transition_impl(settings, attr):
     return {
-        "//command_line_option:platforms": str(attr.platform) if attr.platform else settings["//command_line_option:platforms"],
+        "//command_line_option:platforms": _resolve_platform(settings, attr),
     }
 
 _rust_static_library_transition = transition(
@@ -1049,7 +1063,7 @@ rust_static_library = rule(
 
 def _rust_shared_library_transition_impl(settings, attr):
     return {
-        "//command_line_option:platforms": str(attr.platform) if attr.platform else settings["//command_line_option:platforms"],
+        "//command_line_option:platforms": _resolve_platform(settings, attr),
     }
 
 _rust_shared_library_transition = transition(
@@ -1171,7 +1185,7 @@ _RUST_BINARY_ATTRS = {
 
 def _rust_binary_transition_impl(settings, attr):
     return {
-        "//command_line_option:platforms": str(attr.platform) if attr.platform else settings["//command_line_option:platforms"],
+        "//command_line_option:platforms": _resolve_platform(settings, attr),
     }
 
 _rust_binary_transition = transition(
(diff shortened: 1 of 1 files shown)


rust/private/rust.bzl
```diff
@@ -1006,9 +1006,23 @@ rust_library = rule(
         """),
 )
 
+def _resolve_platform(settings, attr):
+    """Resolve the platform label for a transition, adding @ prefix if needed.
+
+    When a label has no repo name (e.g. from the main repo), str(label) omits
+    the leading @, producing "//platforms:foo" instead of "@//platforms:foo".
+    Bazel's platform setting requires the canonical form with @.
```
UebelAndre: Is there any upstream issue for this? That seems somewhat surprising.

UebelAndre (changes requested):
Thanks! I had one question on this.

rust/private/rust.bzl
tamasvajk: I'm not aware of a specific upstream Bazel issue for this. The problem is that `str(attr.platform)` inside a transition implementation omits the `@` prefix for labels from the main repository. Since the transition runs in the context of `@rules_rust`, Bazel then interprets `//platforms:foo` as `@rules_rust//platforms:foo` instead of `@//platforms:foo`.

This was discovered when using the `platform` attribute on `rust_binary` with a label from the main repo — the platform was silently resolved against rules_rust instead of the root workspace.

tamasvajk: I tried to set up a minimal repro, but I could only repro this with our patched Bazel. So the problem is actually not in rules_rust, but in our Bazel patch. I'm closing this PR.

UebelAndre: Can you file an upstream issue for this? Because `platform` is a `attr.label` I would have expected it to be in the correct form for all contexts and that a simple `str` would result in the correct behavior. I'm happy to accept the fix but would love to hear from the Bazel team first 🙏

UebelAndre: Also @krasimirgg do you maybe have any context here?

UebelAndre: Ah, disregard

tamasvajk: Reopening — I was able to reproduce this with **stock Bazel** (not our patched version). The trigger is `--noincompatible_unambiguous_label_stringification`, which some projects set for compatibility reasons.

**Minimal repro:**

```python
# WORKSPACE
workspace(name = "repro")
http_archive(name = "rules_rust", sha256 = "d861...", urls = ["...0.64.0..."])
rules_rust_dependencies()
rust_register_toolchains()

# BUILD.bazel
rust_binary(name = "hello", srcs = ["main.rs"], edition = "2021", platform = "//platforms:my_linux_x86_64")

# platforms/BUILD.bazel
platform(name = "my_linux_x86_64", constraint_values = ["@platforms//os:linux", "@platforms//cpu:x86_64"], visibility = ["//visibility:public"])
```

```
$ bazel build //:hello --noincompatible_unambiguous_label_stringification
ERROR: no such package '@@rules_rust//platforms': BUILD file not found in directory 'platforms' of external repository @@rules_rust.
```

**Root cause:** With `--noincompatible_unambiguous_label_stringification` (which was the default before Bazel 6.0, see [bazelbuild/bazel#15916](https://github.com/bazelbuild/bazel/issues/15916), [bazelbuild/bazel#16196](https://github.com/bazelbuild/bazel/issues/16196)), `str(attr.platform)` inside the transition produces `//platforms:my_linux_x86_64` without the `@` prefix. Bazel then resolves this relative to `@@rules_rust` instead of the main repo.

The flag was introduced to fix exactly this class of ambiguity, but projects that disable it (via `--noincompatible_unambiguous_label_stringification`) hit this bug.

tamasvajk:
@UebelAndre What's the best way to rerun the CI jobs? There are some transient failures in the logs.

rust/private/rust.bzl
```diff
@@ -1006,9 +1006,24 @@ rust_library = rule(
         """),
 )
 
+def _resolve_platform(settings, attr):
+    """Resolve the platform label for a transition, adding @ prefix if needed.
+
+    With --noincompatible_unambiguous_label_stringification, str(label) for
+    main-repo labels omits the leading @, producing "//foo:bar" instead of
+    "@//foo:bar". The platform setting requires the canonical form with @.
+    See https://github.com/bazelbuild/bazel/issues/15916.
```
UebelAndre, line 1015:
```suggestion
    See https://github.com/bazelbuild/bazel/issues/15916.
    
    Note that this function will no longer be needed if
    `--noincompatible_unambiguous_label_stringification` is dropped but
    it's currently required internally by Google.
    See https://github.com/bazelbuild/bazel/issues/16196
```

UebelAndre (approved):
Thank you!

rust/private/rust.bzl
```diff
@@ -1006,9 +1006,29 @@ rust_library = rule(
...
+    
+    Note that this function will no longer be needed if
```
UebelAndre, line 1017:
```suggestion
    See https://github.com/bazelbuild/bazel/issues/15916.

    Note that this function will no longer be needed if
```

UebelAndre (approved):

PR #3999, opened 2026-04-28
Add `RUST_COVERAGE_IGNORE_REGEX` for configurable coverage filtering

tamasvajk:
## Summary

Add `RUST_COVERAGE_IGNORE_REGEX` environment variable for configurable `llvm-cov export` filtering.

## Problem

The `collect_coverage.rs` script hardcodes two `-ignore-filename-regex` patterns (`.*external/.+` and `/tmp/.+`) passed to `llvm-cov export`. Projects often need additional filters — for example, to exclude vendored C/C++ code or generated files — but there's no way to add them without patching the script.

## Fix

Read an optional `RUST_COVERAGE_IGNORE_REGEX` environment variable: a comma-separated list of additional regex patterns passed as `-ignore-filename-regex` arguments to `llvm-cov`. The existing hardcoded patterns are preserved as defaults.

Users can set this via `--test_env` in `.bazelrc`:

```
coverage --test_env=RUST_COVERAGE_IGNORE_REGEX=.*vendor/.*,^/rustc/
```

---

### DIFF
--- a/util/collect_coverage/collect_coverage.rs
+++ b/util/collect_coverage/collect_coverage.rs
@@ -190,7 +190,20 @@ fn main() {
         .arg("-instr-profile")
         .arg(&profdata_file)
         .arg("-ignore-filename-regex=.*external/.+")
-        .arg("-ignore-filename-regex=/tmp/.+")
+        .arg("-ignore-filename-regex=/tmp/.+");
+
+    // Allow additional ignore patterns via RUST_COVERAGE_IGNORE_REGEX env var
+    // (comma-separated list of regexes passed to llvm-cov -ignore-filename-regex).
+    if let Ok(extra) = env::var("RUST_COVERAGE_IGNORE_REGEX") {
+        for pattern in extra.split(',') {
+            let pattern = pattern.trim();
+            if !pattern.is_empty() {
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Thanks! Is there any prior-art for this in other rules? I'm wondering if a custom environment variable is really needed

tamasvajk:
> Thanks! Is there any prior-art for this in other rules? I'm wondering if a custom environment variable is really needed

I couldn't find direct prior art in other Bazel rule sets — C++ coverage in Bazel uses `collect_coverage.sh` with its own hardcoded logic, and rules_go doesn't use llvm-cov.

The env var approach was chosen because the filtering happens at runtime (in the `collect_coverage` binary, not during analysis), so `--test_env` is the standard Bazel mechanism for passing runtime configuration to tests without invalidating the analysis cache.

An alternative would be to add a build setting like `--@rules_rust//rust/settings:coverage_ignore_regex` and plumb it through to the test action's environment. That would be more discoverable and consistent with how rules_rust exposes other configuration, though it would require more plumbing and invalidate the analysis cache when changed. Happy to rework it that way if you'd prefer.

PR #4002, opened 2026-04-28
Fix `rustfmt` silently skipping symlinked sources

tamasvajk:
## Summary

Fix `rustfmt_test` and `rustfmt_aspect` silently skipping hand-written sources that were symlinked by `transform_sources()`.

## Problem

When any `compile_data` entry is a generated file, `transform_sources()` symlinks all source files into `bazel-out`. After symlinking, `is_source` returns `False` for these files, so `_find_rustfmtable_srcs` filters them all out. The result is that rustfmt silently does nothing — no error, no formatting check.

## Fix

When the aspect context is available, recover original source `File` objects from the rule's `srcs` attribute (which still references the pre-symlink originals). This preserves correct `exec_path` and `short_path` for manifests and runfiles.

Also use `crate_info.srcs` directly in `_rustfmt_test_impl` to include all sources.

---

### DIFF
--- a/rust/private/rustfmt.bzl
+++ b/rust/private/rustfmt.bzl
@@ -47,7 +47,19 @@ def _find_rustfmtable_srcs(crate_info, aspect_ctx = None):
             if tag.replace("-", "_").lower() in ignore_tags:
                 return []
 
-    # Filter out any generated files
+    # Filter out generated files, but recover hand-written sources that were
+    # symlinked into bazel-out by transform_sources() (which happens when any
+    # compile_data entry is generated). Return the original source File objects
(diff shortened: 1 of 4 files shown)


UebelAndre:
@tamasvajk Does https://github.com/bazelbuild/rules_rust/pull/3983 solve for your needs? This seems to be solving a similar issue to https://github.com/bazelbuild/rules_rust/issues/3850

tamasvajk:
> @tamasvajk Does #3983 solve for your needs? This seems to be solving a similar issue to #3850

Yes, I think so. Thank you.

PR #4008, opened 2026-04-30
Fix ordering of `--remap-path-prefix` flags.

slackito:
These flags are interpreted in reverse order (see
https://github.com/rust-lang/rust/issues/82108), and if we have a set of flags like:
```
  --remap-path-prefix=/a/b/c/d=.
  --remap-path-prefix=/a/b=.
```
then the first flag will never apply, because the path will be first rewritten as ./c/d.

In this case, output_base is the outermost directory, so we need to make sure it's remapped last.


test/unit/remap_path_prefix/remap_path_prefix_test.bzl
```diff
@@ -17,9 +16,11 @@ def _remap_path_prefix_test_impl(ctx):
...
     assert_action_mnemonic(env, action, "Rustc")
 
-    assert_argv_contains(env, action, "--remap-path-prefix=${pwd}=.")
-    assert_argv_contains(env, action, "--remap-path-prefix=${exec_root}=.")
-    assert_argv_contains(env, action, "--remap-path-prefix=${output_base}=.")
+    assert_list_contains_adjacent_elements(env, action.argv, [
+        "--remap-path-prefix=${output_base}=.",
+        "--remap-path-prefix=${pwd}=.",
+        "--remap-path-prefix=${exec_root}=.",
+    ])
```
UebelAndre, line 23: Nit: Maybe add an assert message that the ordering is sensitive?

UebelAndre (approved):
Looks good to me! One optional nit

test/unit/remap_path_prefix/remap_path_prefix_test.bzl
slackito: What do you mean? `assert_list_contains_adjacent_elements` doesn't have any argument to pass a message.

UebelAndre: Oh, I thought it did, I know some do. Either way, the desire was to ensure folks who encounter this test to be failing don't just update the test to match the current behavior. A note would be nice

Issue #2177, opened 2023-10-02
rustc_compile_action doesn't pass default_shell_env


granaghan:
I have a CC toolchain where most of the "tools" are actually shell script wrappers that ensure only toolchain provided libraries are loaded. This involves some calls to readlink and dirname. rust_compile_actions doesn't set `use_default_shell_env` to True and as a result, these fail when rustc calls clang which in turn calls ld.lld. I found that setting `use_default_shell_env = True` resolved my issue.` Is this a change that can be made? Or should we consider adding a setting to either the rust toolchain or add another provider that can be applied to CC toolchains so we can let users configure this?

https://github.com/bazelbuild/rules_rust/blob/aa4b3a862a8200c9422b7f7e050b12c7ef2c2a61/rust/private/rustc.bzl#L1258

PR #4009, opened 2026-05-01
add rustc setting `use_default_shell_env` to `rust_toolchain`

oliverlee:
`use_default_shell_env` is a bool setting for a `rust_toolchain` that controls whether rustc actions inherit the shell environment. The default is `False` to maintain existing behavior.

This allows use in environments where PATH is set outside of Bazel's action environment, e.g. when using a Nix-based CcToolchain that relies on a Nix shell to populate PATH.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -645,9 +645,6 @@ RUSTC_ATTRS = {
     "_rustc_output_diagnostics": attr.label(
         default = Label("//rust/settings:rustc_output_diagnostics"),
     ),
-    "_use_default_shell_env": attr.label(
-        default = Label("//rust/settings:use_default_shell_env"),
-    ),
 }
 
 _COMMON_ATTRS = {
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -645,6 +645,9 @@ RUSTC_ATTRS = {
     "_rustc_output_diagnostics": attr.label(
         default = Label("//rust/settings:rustc_output_diagnostics"),
     ),
+    "_use_default_shell_env": attr.label(
+        default = Label("//rust/settings:use_default_shell_env"),
+    ),
 }
 
 _COMMON_ATTRS = {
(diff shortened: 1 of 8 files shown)


oliverlee:
Related:
https://github.com/bazelbuild/rules_rust/issues/2177
https://github.com/bazelbuild/rules_rust/issues/3482
https://github.com/bazelbuild/rules_rust/pull/3556
https://github.com/bazelbuild/rules_rust/pull/3535

rust/private/rust.bzl
```diff
@@ -645,6 +645,9 @@ RUSTC_ATTRS = {
     "_rustc_output_diagnostics": attr.label(
         default = Label("//rust/settings:rustc_output_diagnostics"),
     ),
+    "_use_default_shell_env": attr.label(
```
UebelAndre: Would it make more sense to have this on `rust_toolchain`?

UebelAndre (changes requested):
Thanks! This seems like a reasonable compromise 😄 

Just one question.

rust/private/rust.bzl
oliverlee: I'm fine with either a toolchain flag or a global setting, although it seems unlikely a project would have multiple Rust toolchains where only some need the default shell env.

However, I don't have deep `rules_rust` experience, so feel free to disregard — at my day job I use Rust to build some tooling, though only ever with a single Rust toolchain.

UebelAndre: We currently have a big number of flags on the toolchain and feel it's much easier to manage having it there
https://github.com/bazelbuild/rules_rust/blob/71470d7109c3264e6aa8aaa5794884ddfc8bedba/rust/toolchain.bzl#L637-L651

Both work but if there's no string feelings or technical rationale (which I'm amenable to here) I'd prefer to have it on the toolchain

oliverlee: Sure I can take a look at it next week.

PR #4010, opened 2026-05-01
Added support for additional platform triples

UebelAndre:
With https://github.com/bazelbuild/platforms/pull/134 landed we can now support additional platforms.


slackito (approved):
LGTM

PR #4011, opened 2026-05-02
Add support for `experimental_output_paths`

UebelAndre:
For details on path mapping see https://github.com/bazelbuild/bazel/discussions/22658

Changes:
- The process wrapper now explicitly handles `out_dir` to avoid location expanded environment variables.
- `construct_arguments` is updated to support `rustc_flags` as an `Args` object. `List` is still supported though


.bazelci/presubmit.yml
```diff
@@ -377,43 +377,100 @@ tasks:
     platform: ubuntu2204
     run_targets:
       - "//test/rustfmt:rustfmt_failure_tester"
-  rust_analyzer_tests_linux:
-    name: Rust-Analyzer Linux Tests
+  ide_integration_tests_linux:
+    name: IDE VSCode Tests
```
slackito, line 381: Are the Rust-Analyzer -> IDE VSCode config changes related to the rest of the PR?

rust/private/rustc.bzl
```diff
@@ -1284,10 +1393,11 @@ def construct_arguments(
         rustc_flags.add('--cfg=feature="no_std"')
 
     # Add target specific flags last, so they can override previous flags
+    authored_rustc_flags = getattr(attr, "rustc_flags", [])
```
slackito, line 1396: What does "authored" mean in this context?

slackito (commented):
Looks good overall. Thank you for the thorough comments, they made the changes much easier to follow :)

Just a couple of comments.

.bazelci/presubmit.yml
UebelAndre: I needed to trim down the number of jobs since we're hitting the 128 limit

rust/private/rustc.bzl
UebelAndre: Flags users added to their targets. I can use whatever name you'd prefer. I know I'm bad at naming 😅

slackito (approved):

bdolgov:
I think this has broken a pattern somewhat common in embedded Rust where one crate [puts](https://github.com/knurling-rs/defmt/blob/86c3c50c3e1d38015e250fbed1f0a1e19872934d/defmt/build.rs#L14) a file into build script outputs, and another create [references](https://github.com/embassy-rs/embassy/blob/b80bcdbd13147068a3e1620768aaf60b3027eced/examples/rp/build.rs#L35) it from its build script in linker flags.  I bisected the problem to https://github.com/bazelbuild/rules_rust/commit/175bf94bddc7ddc341ea1708edb44eb23a035a69, but didn't yet understand whether the problem is in this commit or somewhere on my side.

UebelAndre:
> I think this has broken a pattern somewhat common in embedded Rust where one crate [puts](https://github.com/knurling-rs/defmt/blob/86c3c50c3e1d38015e250fbed1f0a1e19872934d/defmt/build.rs#L14) a file into build script outputs, and another create [references](https://github.com/embassy-rs/embassy/blob/b80bcdbd13147068a3e1620768aaf60b3027eced/examples/rp/build.rs#L35) it from its build script in linker flags. I bisected the problem to [175bf94](https://github.com/bazelbuild/rules_rust/commit/175bf94bddc7ddc341ea1708edb44eb23a035a69), but didn't yet understand whether the problem is in this commit or somewhere on my side.

Interesting, I would be surprised this has an impact assuming you're not using `--experimental_output_paths=strip`. I assume you're not?

bdolgov:
No, I don't.

UebelAndre:
> No, I don't.

Can you make a small repro in a pull-request somewhere in [//cargo/tests/cargo_build_script](https://github.com/bazelbuild/rules_rust/tree/7e46ed660addb62ba301b38d01163d89ad31d38e/cargo/tests/cargo_build_script) ?

PR #3903, opened 2026-03-17
Remap path prefixes for canonicalized paths

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/1530

PR #4016, opened 2026-05-05
Adjust `--remap-path-prefix` for targets with generated sources

tamasvajk:
## Summary

Adjust `--remap-path-prefix` to produce clean workspace-relative paths when a target has mixed generated and non-generated sources.

## Problem

When a Rust target mixes generated and non-generated source files (e.g. generated `compile_data`), `transform_sources()` symlinks all source files into `bazel-out/<config>/bin/...` so they sit next to the generated files. The existing `--remap-path-prefix=${pwd}=.` only strips the exec root prefix, leaving `bazel-out/k8-fastbuild/bin/` in:

- Diagnostics and compiler errors
- Panic locations and backtraces
- macOS N_OSO linker entries

For example, `./bazel-out/k8-fastbuild/bin/my/pkg/lib.rs` instead of `./my/pkg/lib.rs`.

## Fix

Detect symlinked sources via `crate_info.root.is_source` and remap `${pwd}/<bin_dir>` (e.g. `${pwd}/bazel-out/k8-fastbuild/bin`) instead of just `${pwd}`, stripping the bin-dir component to produce clean workspace-relative paths. The same fix is applied to:

- `${exec_root}` remapping (canonicalized exec root paths, added in #3903)
- macOS `-oso_prefix` (N_OSO stab entries, also added in #3903)

## Tests

Updated the analysis tests in `remap_path_prefix_test.bzl` to cover both cases:

- **Source files**: plain `${pwd}=.` / `${exec_root}=.` remapping
- **Generated sources**: `${pwd}/<bin_dir>=.` / `${exec_root}/<bin_dir>=.` remapping

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1211,8 +1211,19 @@ def construct_arguments(
             rustc_flags.add("--remap-path-prefix=${{pwd}}={}".format(remap_path_prefix))
             rustc_flags.add("--remap-path-prefix=${{exec_root}}={}".format(remap_path_prefix))
         else:
-            rustc_flags.add("--remap-path-prefix=${{pwd}}/{}={}".format(ctx.bin_dir.path, remap_path_prefix))
-            rustc_flags.add("--remap-path-prefix=${{exec_root}}/{}={}".format(ctx.bin_dir.path, remap_path_prefix))
+            # Use add_all with the crate root File and a map_each callback
+            # so Bazel's path mapping (--experimental_output_paths=strip)
+            # can rewrite the config portion of the bin-dir prefix.
+            rustc_flags.add_all(
+                [crate_info.root],
+                map_each = _get_bin_dir_prefix,
+                format_each = "--remap-path-prefix=${pwd}/%s=" + remap_path_prefix,
+            )
+            rustc_flags.add_all(
+                [crate_info.root],
+                map_each = _get_bin_dir_prefix,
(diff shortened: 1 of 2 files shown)


tamasvajk:
@UebelAndre Could you please take a look at this PR?

rust/private/rustc.bzl
```diff
@@ -1288,11 +1319,12 @@ def construct_arguments(
             if remap_path_prefix != None and _should_add_oso_prefix(
                 toolchain,
             ):
+                oso_prefix = "${pwd}/" if crate_info.root.is_source else "${pwd}/" + ctx.bin_dir.path + "/"
```
slackito, line 1322: Do you need to pass this as a `File` as well so that it works with `--experimental_output_paths=strip`?

rust/private/rustc.bzl
```diff
@@ -2706,6 +2738,22 @@ def _add_native_link_flags(
...
+    prefix.  Using the File object with add_all/map_each lets Bazel apply
+    path mapping (--experimental_output_paths=strip) to the config portion.
+
+    Args:
+        file (File): A generated file (e.g. the crate root).
+
+    Returns:
+        str: The bin-dir prefix, e.g. "bazel-out/k8-fastbuild/bin".
+    """
+    return file.path[:len(file.path) - len(file.short_path)].rstrip("/")
```
slackito, line 2755: It would be nice if you could fold the `if file.is_source` logic here, so you don't have to check at every call site.

Also, would `file.root.path` work instead of the string manipulation here?

PR #4017, opened 2026-05-05
Remap relative bin-dir prefix for `file!()` and log macros

tamasvajk:
## Summary

Companion to #4016. While #4016 fixes `--remap-path-prefix` for absolute paths in debug info (backtraces, panic locations), this PR adds a remap for the relative `bazel-out/<config>/bin/` prefix so that `file!()`, `Location::caller()`, and tracing/log macros also produce clean workspace-relative paths when sources are symlinked by `transform_sources()`.

## Problem

When a target mixes generated and non-generated sources, `transform_sources()` symlinks all source files into `bazel-out/<config>/bin/...`. The existing `--remap-path-prefix=${pwd}=.` only affects absolute paths embedded in debug info. However, `file!()`, `Location::caller()`, and log macros embed the literal *relative* path that rustc received on the command line. These are not matched by the `${pwd}` remap, so they produce paths like:

```
bazel-out/k8-fastbuild/bin/my/pkg/lib.rs
```

instead of:

```
my/pkg/lib.rs
```

## Fix

When `crate_info.root.is_source` is False (symlinked sources), add a second remap that strips the relative bin-dir prefix:

```
--remap-path-prefix=bazel-out/<config>/bin/=
```

This complements the existing absolute-path remaps and ensures compile-time paths are also clean.

## Tests

- **Analysis tests**: Verify the relative bin-dir remap flag is present for generated-source targets and absent for plain source targets.
- **Runtime test**: A library with generated sources exposes `file!()` and a test asserts the returned path does not contain `bazel-out`.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1193,6 +1193,14 @@ def construct_arguments(
         rustc_flags.add("--remap-path-prefix=${{output_base}}={}".format(remap_path_prefix))
         rustc_flags.add("--remap-path-prefix=${{pwd}}={}".format(remap_path_prefix))
         rustc_flags.add("--remap-path-prefix=${{exec_root}}={}".format(remap_path_prefix))
+        if not crate_info.root.is_source:
+            # Also remap the relative bazel-out/<config>/bin/ prefix.
+            # file!(), Location::caller(), and tracing/log macros embed
+            # the literal relative path that rustc received on the command
+            # line.  The ${pwd} remap above only matches absolute paths in
+            # debug info, so a second remap without ${pwd} is needed to
+            # strip the bin-dir prefix from these compile-time paths.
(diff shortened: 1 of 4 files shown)


tamasvajk:
@UebelAndre Could you please take a look at this PR?

rust/private/rustc.bzl
```diff
@@ -2706,6 +2718,25 @@ def _add_native_link_flags(
                     format_each = "-lstatic=%s",
                 )
 
+def _get_bin_dir_remap_prefix(file):
```
slackito, line 2721: I feel like this function could be combined with the helper function in #4016.

You could use `_get_bin_dir_prefix` and then add the `--remap-path-prefix=` and the trailing forward slash if you need it here with `format_each`.

PR #4018, opened 2026-05-05
Add crate_edition helper macro.

### LINKED ISSUE #4014: feat: Add macro for reading the crate edition similar to all_crate_deps
`Cargo.toml` files should be the source of truth for the edition to use. Similar to `all_crate_deps` for dependencies, should we have a `crate_edition` macro that provide the crate's edtion which can then be used in `BUILD.bazel` files. Something like

```starlark
    rust_library(
        name = "foo",
        edition = crate_edition(),
        deps = all_crate_deps(normal = True),
    )
```

This sounds like information that is available during cargo splicing so we could generate this file as part of splicing. Or is there a better way to do this?

lukasoyen:
Closes #4014


hofbi (approved):

UebelAndre (approved):
Thank you!

PR #4019, opened 2026-05-05
Add additional baremetal platform triples

davexroth:
This change addes the following target platform triples:
- aarch64-unknown-none
- thumbv7m-none-eabi
- thumbv8m.main-none-eabihf
- riscv32imac-unknown-none-elf

These targets are required to support various baremetal projects.


UebelAndre (approved):
Thanks! Looks like buildifier is unhappy though

davexroth:
thanks  - fixed!

UebelAndre (approved):

UebelAndre:
Thank you!

PR #4021, opened 2026-05-08
util/process_wrapper: improve resilience to non-JSON output

th0br0:
Make process_line in process_wrapper catch parsing failures and wrap non-JSON lines in valid JSON warnings. 
This prevents crashes when rustc emits unexpected warnings.

### DIFF
--- a/util/process_wrapper/main.rs
+++ b/util/process_wrapper/main.rs
@@ -92,28 +92,37 @@ fn json_warning(line: &str) -> JsonValue {
 }
 
 fn process_line(
-    mut line: String,
+    line: String,
     quit_on_rmeta: bool,
     format: ErrorFormat,
     metadata_emitted: &mut bool,
 ) -> Result<LineOutput, String> {
-    // LLVM can emit lines that look like the following, and these will be interspersed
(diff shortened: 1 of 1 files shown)


util/process_wrapper/main.rs
```diff
@@ -92,28 +92,37 @@ fn json_warning(line: &str) -> JsonValue {
 }
 
 fn process_line(
-    mut line: String,
+    line: String,
     quit_on_rmeta: bool,
     format: ErrorFormat,
     metadata_emitted: &mut bool,
 ) -> Result<LineOutput, String> {
-    // LLVM can emit lines that look like the following, and these will be interspersed
```
slackito, line 100: We're still trying to convert non-JSON lines to JSON so I think this comment explaining why we need to do that should be preserved.

slackito (commented):
I think the general idea looks good. If I understand correctly, instead of looking for a particular pattern of output messages that should be converted to JSON warnings, now we look for strings that are likely JSON (if they start with a '{') and convert anything else to JSON warnings, right?

Pre-merge checks are failing, specifically //test/process_wrapper:rustc_quit_on_rmeta. Can you take a look?

PR #4024, opened 2026-05-08
feature: introduce link_deps attribute for native library linkage

### LINKED ISSUE #4023: Feature Request: `link_deps` attribute for native library linkage
### Description
Currently, the `deps` attribute in `rules_rust` is overloaded: it accepts both `CrateInfo` (Rust crates) and `CcInfo` (C++ libraries). 

This overloading creates ambiguity in the dependency model. Specifically, it makes it difficult to link a dependency purely for its native symbols (e.g., satisfying manual FFI via `extern "C"`) without also exposing it as a formal Rust module to the compiler. In large build graphs, this can lead to unnecessary namespace pollution and "leaky" dependency abstractions where internal linkage requirements are exposed as public API dependencies.

### Proposed Architectural Change
I propose the introduction of a dedicated `link_deps` attribute to the core Rust rules (`rust_library`, `rust_binary`, `rust_test`).

**Key Goals:**
1.  **Explicit intent**: Create a clear distinction between `deps` (contributing to the Rust module hierarchy) and `link_deps` (contributing strictly to the linker command line).
2.  **Private Linkage Support**: Allow `link_deps` to accept both C++ and Rust libraries, but only extract the linkage context (`CcInfo`). This enables "private" linkage of Rust libraries where binary symbols are satisfied but the crate metadata is hidden from the consumer.
3.  **Non-Breaking addition**: Maintain support for C++ libraries in the existing `deps` attribute while proactively notifying users (e.g., via analysis-phase warnings) to encourage migration to the more explicit `link_deps` without breaking current build workflows.

### Justification
This feature will eliminate the "silent failure" or cryptic error messages that occur when the wrong type of library is placed in `deps`. For example, currently, if a `cc_library` is incorrectly placed in `deps`, the Bazel analysis phase succeeds, but the build later fails during compilation with a confusing `can't find crate for...` error from `rustc`. By separating these concerns, we can provide proactive feedback during the analysis phase.

### Impact on Existing Rules
*   **`rust_library`, `rust_binary`, `rust_test`**: Gain the `link_deps` attribute.
*   **`rust_proc_macro`**: Intentionally remains focused on host-platform requirements to avoid host-vs-target linkage complexity.
*   **Backward Compatibility**: Zero impact on existing builds.

Suyashagarw:
This change adds a dedicated link_deps attribute to Rust rules, allowing native symbols to be linked without exposing them as Rust crates.

Closes #4023 

Key changes:
* Added link_deps to _COMMON_ATTRS for CcInfo/CrateInfo targets.
* Filtered link_deps out of rust_proc_macro to avoid transition complexity.
* Implemented transform_link_deps to extract linkage context while hiding crate metadata.
* Updated _assert_correct_dep_mapping to print a proactive warning for native libraries listed in deps.
* Added unit tests in test/unit/link_deps to verify correct linkage and provider propagation.

This provides a cleaner separation of concerns between Rust modules and native code linkage while maintaining full backward compatibility for existing targets.


rust/private/rust.bzl
```diff
@@ -713,15 +749,22 @@ _COMMON_ATTRS = {
...
     "edition": attr.string(
         doc = "The rust edition to use for this crate. Defaults to the edition specified in the rust_toolchain.",
     ),
+    "link_deps": attr.label_list(
+        doc = dedent("""\
+            List of other native libraries to be linked to this library target.
+
+            These are typically `cc_library` targets.
+        """),
+        providers = [[CcInfo], [rust_common.crate_info]],
```
krasimirgg, line 759: Why the `[rust_common.crate_info]` part here? Naively, while we should allow for rust_library dependencies here, they by themselves already provide `CcInfo`, so they should satisfy the `[CcInfo]` spec already?

rust/private/rust.bzl
```diff
@@ -1128,16 +1171,15 @@ rust_proc_macro = rule(
...
     attrs = dict(
-        _COMMON_ATTRS.items(),
+        {name: value for name, value in _COMMON_ATTRS.items() if name != "link_deps"}.items(),
         _allowlist_function_transition = attr.label(
             default = Label("//tools/allowlists/function_transition_allowlist"),
         ),
         deps = attr.label_list(
             doc = dedent("""\
-                List of other libraries to be linked to this library target.
+                List of other Rust libraries to be linked to this library target.
```
krasimirgg: We were chatting about this internally (don't add link_deps to rust_proc_macro-s). I think externally, we should still not add link_deps to proc_macros, but crucially here we should preserve the pre-existing API: c++-only deps of proc-macros are still OK. So we should change this back to not be specific to Rust libraries.

test/unit/link_deps/link_deps_test.bzl
```diff
@@ -0,0 +1,21 @@
...
+    target = analysistest.target_under_test(env)
+
+    # Verify that CcInfo (symbols) is present
+    asserts.true(env, CcInfo in target, "Target should provide CcInfo from link_deps")
+
+    return analysistest.end(env)
+
+link_deps_test = analysistest.make(_link_deps_test_impl)
+
+def _link_deps_failure_test_impl(ctx):
```
krasimirgg: hey, we don't seem to actually be using this (instantiating) in tests?

test/unit/link_deps/link_deps_test.bzl
```diff
@@ -0,0 +1,21 @@
+load("//rust:rust_common.bzl", "CrateInfo")
+load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+
+def _link_deps_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    target = analysistest.target_under_test(env)
+
+    # Verify that CcInfo (symbols) is present
+    asserts.true(env, CcInfo in target, "Target should provide CcInfo from link_deps")
```
krasimirgg, line 12: This doesn't seem to test much with link_deps -- it asserts that the (client) target_under_test (in the BUILD file main_lib) produces a CcInfo. Instead, you probably want to have the test code examine the client providers of the target under test, locate the information about the link_deps dependency and examine that. Look at the DepsInfo provider, and its direct_crates and transitive_noncrates fields.

test/unit/link_deps/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+    srcs = ["lib.rs"],
+)
+
+# Test 1: rust_library supports link_deps
+rust_library(
+    name = "main_lib",
+    srcs = ["lib.rs"],
+    link_deps = [":leaf_lib"],
+    tags = ["manual"],
+)
```
krasimirgg, line 18: This analysistest suite is nice. 
Could you also add a minimal non-analysistest build-test example: something like a cc_library that defines an extern "C" function and a rust_binary that refers to it via link_deps and uses the symbol (so the symbol must be available at link-time, but only there), and a build test that ensures that the rust_binary links successfully.

krasimirgg (changes requested):

rust/private/rust.bzl
Suyashagarw: This is technically redundant but I included it to be explicit about supporting Rust-on-Rust native linkage as a goal. Removed `rust_common.crate_info`.

rust/private/rust.bzl
Suyashagarw: That makes sense. Reverted it to not be Rust specific.

test/unit/link_deps/link_deps_test.bzl
Suyashagarw: I originally wrote `link_deps_failure_test` as an automated negative test. However, adding `link_deps` to a `rust_proc_macro` target triggered a fatal loading-phase error ("no such attribute"). I manually verified the expected failure and removed the broken target so that the suite could build, but forgot to drop the unused test definition from the `link_deps_test.bzl` file.

I've cleaned that up now. Thanks for pointing it out!

test/unit/link_deps/link_deps_test.bzl
Suyashagarw: Updated `link_deps_test` to inspect `dep_info`. The test now explicitly verifies that dependencies passed via `link_deps` do not get added to `direct_crates`, but their native linking contexts correctly populate `transitive_noncrates`.

test/unit/link_deps/BUILD.bazel
Suyashagarw: Added appropriate cc_library and rust_binary targets for a build-test example. Thanks for the suggestion!

test/unit/link_deps/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+cc_library(
+    name = "c_lib",
+    srcs = ["c_lib.c"],
+)
+
+rust_binary(
+    name = "rust_bin",
+    srcs = ["main.rs"],
+    link_deps = [":c_lib"],
+)
```
krasimirgg: add a build_test that builds :rust_bin  (otherwise a sample bazel test would skip it, since a binary is not a test target)

Suyashagarw: Acknowledged. Added a build_test.

krasimirgg (approved):
Thank you!

UebelAndre:
@krasimirgg @Suyashagarw per https://github.com/bazelbuild/rules_rust/pull/4095

> https://github.com/bazelbuild/rules_rust/pull/4024 introduced a warning for targets adding CC deps in deps. The warning calls out putting this in link_deps or cc_deps. The latter makes more sense here... but doesn't actually exist.

I did notice this change added tons of debug prints to builds in `rules_rust` and now it seems the guidance is inaccurate? There is indeed no `cc_deps`

krasimirgg:
> I did notice this change added tons of debug prints to builds in `rules_rust` and now it seems the guidance is inaccurate? There is indeed no `cc_deps`

Hey, missed that. Internally, we have some customizations that add the `cc_deps` attribute and automagically enable Crubit on those. Let me prep a PR to update the warning: https://github.com/bazelbuild/rules_rust/pull/4099

PR #4027, opened 2026-05-09
mdbook: fix support for external repositories

### LINKED ISSUE #4026: `mdbook` fails when used with external repository
Hello! I have been trying to use `mdbook` with sources coming from an external repository and I'm find that I get

> [ERROR] (`mdbook::utils`): Error: Couldn't open `SUMMARY.md` in `".../repro-external-sources_/external/content+/src"`

errors back. I had Gemini reproduce the problem in https://github.com/mgeisler/rules-rust-mdbook-external-sources and I hope this is useful for you.

### Description

When any part of an `mdbook` target (either the `book.toml` or the source files) comes from an _external repository_, the build fails. This is because the rule's staging logic and its invocation logic handle repository boundaries inconsistently.

### Reproduction steps

Please see https://github.com/mgeisler/rules-rust-mdbook-external-sources

### Additional context

From Gemini: The failure stems from a mismatch between the Starlark rule definition and the Rust `process_wrapper` regarding how external repository paths are resolved.

1. **Staging (`process_wrapper.rs`):** The `MdBookBuild` action uses a Rust wrapper to stage files into a shadow work directory (`<target>_`). This wrapper receives an input manifest where destinations are derived from `file.short_path`. In Bazel, the `short_path` for files in an external repository starts with `../` (e.g., `../content+/src/SUMMARY.md`). This causes the files to be staged **outside** the shadow root (as siblings to it).

2. **Invocation (`mdbook.bzl`):** The `_mdbook_impl` function in `mdbook.bzl` tells `mdbook` where the book's root is using `${pwd}/{}` formatted with `ctx.file.book.dirname`. For external repositories, `dirname` (based on `file.path`) starts with `external/` (e.g., `external/content+/`).

3. **The Mismatch:** `mdbook` is invoked with a path like `.../repro-external-sources_/external/content+/`, but the files were actually staged by the wrapper at `.../repro-external-sources_/../content+/`.

### Impact

I'm likely going to drive `mdbook` via a simple `genrule` instead.

### Bazel and rules_rust version

- Bazel 9.1.0
- `rules_rust` 0.70.0

mgeisler:
When any part of an `mdbook` target (either the `book.toml` or the source files) comes from an external repository, the build was failing because the rule's staging logic and its invocation logic handled repository boundaries inconsistently.

This change:
- Updates `_map_inputs` in `mdbook.bzl` to map external repository paths to `'external/'` instead of `'../'`.
- Hardens the process wrapper with a bounds check to ensure files are not staged outside the shadow work directory.
- Adds an integration test to verify the fix and prevent regressions.

Made by Gemini, fixes #4026.

### DIFF
--- a/extensions/mdbook/private/mdbook.bzl
+++ b/extensions/mdbook/private/mdbook.bzl
@@ -10,7 +10,15 @@ MdBookInfo = provider(
 )
 
 def _map_inputs(file):
-    return "{}={}".format(file.path, file.short_path)
+    dest = file.short_path
+    if dest.startswith("../"):
+        # External repositories have short_paths starting with '../'.
+        # We need them to be staged at 'external/' within our shadow
+        # directory to match how 'file.path' (and thus 'file.dirname')
+        # refers to them.
+        dest = "external/" + dest.removeprefix("../")
+
+    return "{}={}".format(file.path, dest)
 
 def _mdbook_impl(ctx):
     output = ctx.actions.declare_directory(ctx.label.name)
--- a/extensions/mdbook/private/process_wrapper.rs
+++ b/extensions/mdbook/private/process_wrapper.rs
@@ -73,6 +73,17 @@ fn generate_work_dir(output_dir: &Path, inputs_manifest: &BTreeMap<PathBuf, Path
 
     for (src, dest) in inputs_manifest.iter() {
         let abs_dest = workdir.join(dest);
+
+        // Ensure the destination is within the workdir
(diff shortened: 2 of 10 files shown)


mgeisler:
> * Updates `_map_inputs` in `mdbook.bzl` to map external repository paths to `'external/'` instead of `'../'`.

Please note that I don't know what the idiomatic way to handle such things are. Gemini came up with this and it seems to work based on its new test.

UebelAndre (approved):
Nice! Thank you!

PR #4028, opened 2026-05-09
Add tests for mdbook plugins

mgeisler:
This adds a test to ensure that mdBook plugins (preprocessors) are properly injected and executed when using the `mdbook` rule. It uses a naive `rust_binary` preprocessor to avoid introducing a dependency on `serde_json` or Python into the test workspace.


UebelAndre (approved):
Thank you!

PR #4029, opened 2026-05-09
Add test for `mdbook` sources from different places

mgeisler:
This adds a test demonstrating how `mdbook` can stitch together sources from multiple locations in the repository.

We have to mirror the workspace structure in a temporary build directory, enabling `mdbook` to resolve relative links between files regardless of their package or generation method.


UebelAndre (approved):
Thank you!

PR #4033, opened 2026-05-13
Add aarch64-unknown-linux-musl and x86_64-unknown-linux-musl to T2 platform triples

shrey4796:


UebelAndre (changes requested):
Thank you for the change! This is something I've long wanted to do but unfortunately Bazel (at least via [@platforms](https://github.com/bazelbuild/platforms)) doesn't offer any mechanism to distinguish ABI from platform triples which means the two platforms are indistinguishable during toolchain resolution. I have https://github.com/bazelbuild/rules_rust/pull/3937 which is waiting on https://github.com/bazel-contrib/platforms_contrib/pull/2 to try and solve for this but until some constraints exist in a common place this will lead to confusing breakages 😞

shrey4796:
> Thank you for the change! This is something I've long wanted to do but unfortunately Bazel (at least via [@platforms](https://github.com/bazelbuild/platforms)) doesn't offer any mechanism to distinguish ABI from platform triples which means the two platforms are indistinguishable during toolchain resolution. I have #3937 which is waiting on [bazel-contrib/platforms_contrib#2](https://github.com/bazel-contrib/platforms_contrib/pull/2) to try and solve for this but until some constraints exist in a common place this will lead to confusing breakages 😞

Makes sense. Sorry I missed that. Will close this out

PR #4034, opened 2026-05-13
Split rustdoc test compilation and execution

### LINKED ISSUE #1431: Compile `rust_doc_test` binaries in an action

UebelAndre:
This change introduces a new `RustdocTestCompile` action which takes advantage of the nightly rustc flag `-Zunstable-options --persist-doctests` to have rustdoc tests written to disk to be executed in a separate action, similar to a normal `rust_test` which first compiles the test and then executes it in a test action.

To use this change a nightly toolchain is required and the use of `--@rules_rust//rust/settings:experimental_compile_rustdoc_tests=True`.

closes https://github.com/bazelbuild/rules_rust/issues/1431


rust/private/rustdoc/rustdoc_compile_wrapper.rs
```diff
@@ -0,0 +1,165 @@
+use std::collections::BTreeSet;
```
slackito: Can you add a file-level comment describing what the program does and why it's needed?

rust/private/rustdoc.bzl
```diff
@@ -66,22 +67,27 @@ def rustdoc_compile_action(
         output (File, optional): An optional output a `rustdoc` action is intended to produce.
         rustdoc_flags (list, optional): A list of `rustdoc` specific flags.
```
slackito: `rustdoc_flags` is now an `Args` object, right?

rust/private/rustdoc_test.bzl
```diff
@@ -132,6 +129,102 @@ def _rust_doc_test_impl(ctx):
         owner = ctx.label,
     )
 
+def _rlocationpath(file, workspace_name):
+    if file.short_path.startswith("../"):
+        return file.short_path[len("../"):]
```
slackito: `file.short_path.removeprefix("../")`?

rust/private/rustdoc_test.bzl
slackito, line 248: The legacy impl still passes a list to `rustdoc_compile_action`.

rust/private/rustdoc/rustdoc_test_runner.rs
```diff
@@ -0,0 +1,194 @@
...
+    let entries = match fs::read_dir(dir) {
+        Ok(entries) => entries,
+        Err(_) => return,
+    };
+
+    for entry in entries.flatten() {
+        let path = entry.path();
+        if path.is_dir() {
+            collect_rust_out(&path, bins);
+        } else if path.file_name().is_some_and(|n| n == "rust_out") {
```
slackito: This would be "rust_out.exe" on Windows.

rust/private/rustdoc/rustdoc_compile_wrapper.rs
```diff
@@ -0,0 +1,165 @@
...
+
+fn mangle_test_name(human_name: &str) -> String {
+    if let Some((file_and_item, line_part)) = human_name.rsplit_once(" (line ") {
+        if let Some(line_num) = line_part.strip_suffix(')') {
+            if let Some((file_path, _)) = file_and_item.split_once(" - ") {
+                let mangled: String = file_path
+                    .chars()
+                    .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
+                    .collect();
+                return format!("{}_{}_0", mangled, line_num);
```
slackito: Gemini points out that macro expansion could result in multiple tests being originated in the same file and line, and that would make the mangling ambiguous with the hardcoded `_0`.

rust/private/rustdoc/rustdoc_compile_wrapper.rs
UebelAndre: Done!

rust/private/rustdoc.bzl
UebelAndre: It's both, the "legacy" action still needs it. I also call it "legacy" because whenever this can run with a stable toolchain I'd like to make this the default.

rust/private/rustdoc_test.bzl
UebelAndre: Updated! (though this pattern is everywhere already)

rust/private/rustdoc/rustdoc_test_runner.rs
UebelAndre: Good catch!

rust/private/rustdoc_test.bzl
UebelAndre: This remains the same since to avoid the params file and needing to update the wrapper.

rust/private/rustdoc/rustdoc_compile_wrapper.rs
UebelAndre: Thanks Gemini! Yeah, macro implemented tests would be a problem here. Fixed!

slackito (approved):
Thanks!

PR #4046, opened 2026-05-18
Suppress noisy build script warnings

### LINKED ISSUE #3880: How to suppress cargo build script warnings with msvc?
We see the following very noisy warning from rules_rust:

```
D:\workdir\myproject>bazel build //...
INFO: Analyzed 229 targets (2186 packages loaded, 86137 targets configured).
INFO: From Running Cargo build script zstd-sys:
Build Script Warning: cl : Command line warning D9025 : overriding '/W0' with '/W4'
Build Script Warning: cl : Command line warning D9002 : ignoring unknown option '-fvisibility=hidden'
(snip)
```

How best to suppress this? I would normally add /wd9025 /wd9002 but I'm not sure how to pass this through to the cargo build script?

pauldraper:
Fixes #3880.

`cargo::warning=` lines from `build.rs` are forwarded to stderr unconditionally, which is noisy for `-sys` crates such as `zstd-sys`, `aws-lc-sys`, `openssl-sys`. Under Cargo these warnings are hidden for registry/git deps unless the build fails or the user passes `-vv`; under rules_rust the user always sees them.

This PR adds two complementary fixes (both discussed on the issue):

1. **Global flag** `--@rules_rust//cargo/settings:emit_build_script_warnings` (default `auto`). Values:
   - `on` — emit for every target, overriding any per-target opt-out (useful for debugging an `-sys` crate).
   - `auto` — respect the per-target `emit_warnings` attribute. `crate_universe`-generated targets opt out, so registry crates stay quiet (matching Cargo); first-party targets emit by default.
   - `off` — silence build-wide.
2. **Per-target attribute** `cargo_build_script(emit_warnings = ...)` (default `True`). `crate_universe` renders generated `cargo_build_script` targets with `emit_warnings = False`. Honored only when the global flag is `auto`.


pauldraper:
Builds fixed, ready to go @UebelAndre

slackito (approved):
Looks good to me. Thanks!

slackito:
@pauldraper The formatter checks are failing. Could you please take a look?

UebelAndre:
@pauldraper looks like there might have been a merge conflict?

cargo/private/cargo_build_script_runner/lib.rs
UebelAndre:
```suggestion
        let result = BuildScriptOutput::outputs_from_reader(reader, true);
```

pauldraper:
Thanks!

PR #4052, opened 2026-05-20
Remove is_proc_macro_dep.

slackito:
This was only used internally at Google during bootstrapping, and it's no longer needed.


krasimirgg (approved):
Thank you!

PR #4054, opened 2026-05-21
Strip whitespace from paths before prefixing ${pwd}

### LINKED ISSUE #3992: Wrong handling of leading spaces in cargo build script arguments
When updating to rules_rust 0.70.0 we saw failures when building the ring crates (I don't think the crate is very relevant here) while using [toolchains_arm_gnu](https://github.com/hexdae/toolchains_arm_gnu)

In the output of the failed compilation, we can see the following environment variable is set:
```
CFLAGS='-DNDEBUG -O2 -isystem${pwd}/ external/toolchains_arm_gnu++arm_toolchain+aarch64_none_linux_gnu_linux_x86_64/aarch64-none-linux-gnu/include/c++/13.2.1 [more include flags]' 
```

The problem here is the space that's inserted between `${pwd}/` and the remainder of the path

I believe this was introduced in https://github.com/bazelbuild/rules_rust/pull/3832/ which has been included for the first time in 0.70.0

We have fixed the issue in a local rules_rust patch which strips the argument before prefixing `${pwd}/` in [cargo_build_script.bzl](https://github.com/bazelbuild/rules_rust/blob/main/cargo/private/cargo_build_script.bzl#L157). I'm happy to open a PR with the changes if you think that's the right approach but as we're hitting a stream of national holidays here it will take a little time 😅 

Thanks for the continuous work that goes into this repo!

gergondet-woven:
Fixes #3992

This patch strips the relative paths before prefixing `${pwd}`. This also tweaks an existing test to test this situation.

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -197,7 +197,7 @@ def _prefix_pwd_to_flag(args, flag_variations):
 
             # Check for concatenated form (flag with path)
             if arg.startswith(flag):
-                path = arg[len(flag):]
+                path = arg[len(flag):].strip()
 
                 # Don't prefix absolute paths
                 if paths.is_absolute(path):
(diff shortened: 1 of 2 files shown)


slackito (approved):
Thanks for the fix!

PR #4056, opened 2026-05-21
Support `--incompatible_exclude_starlark_flags_from_exec_config`.

slackito:
This change adds `scope = "universal"` to some settings that need to be preserved in exec config.

### DIFF
--- a/rust/toolchain/channel/channel.bzl
+++ b/rust/toolchain/channel/channel.bzl
@@ -24,6 +24,7 @@ def _rust_toolchain_channel_flag_impl(ctx):
 rust_toolchain_channel_flag = rule(
     doc = "A build setting which represents the Rust toolchain channel. The allowed values are {}".format(_CHANNELS),
     implementation = _rust_toolchain_channel_flag_impl,
+    attrs = {"scope": attr.string(default = "universal")},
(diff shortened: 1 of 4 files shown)


rust/toolchain/channel/channel.bzl
```diff
@@ -24,6 +24,7 @@ def _rust_toolchain_channel_flag_impl(ctx):
 rust_toolchain_channel_flag = rule(
     doc = "A build setting which represents the Rust toolchain channel. The allowed values are {}".format(_CHANNELS),
     implementation = _rust_toolchain_channel_flag_impl,
+    attrs = {"scope": attr.string(default = "universal")},
```
krasimirgg, line 27: not super familiar with this flag, this sounds OK but just tagging -- @UebelAndre to double-check -- idea is that if we're using say nightly for the target platform, we should be also using nightly for the exec one at least for consistently with proc-macros I think?

UebelAndre: I think this makes sense. We really want the same toolchain to be used when switching between rlib and proc macro outputs (obviously transitioning the target platform) so for now this makes the most sense to me :)

UebelAndre (approved):

PR #4059, opened 2026-05-27
Pass --config skip_children=true to rustfmt

### LINKED ISSUE #4022: Documented `skip_children` workaround for `mod` walking seems to be ignored on stable rustfmt
I think after @UebelAndre's #3983 (which closed #3850), the `rustfmt_srcs_aspect` now picks up `rust_library` targets whose `srcs` mix hand-written and generated files, where it didn't before. In my repo this surfaces a failure on a common shape: a hand-written `lib.rs` declaring `mod generated;` where `generated.rs` comes from a `genrule`. The aspect runs rustfmt and fails with:

This seems to be the case `docs/rust_fmt.vm` already calls out:

> Any target which uses Bazel generated sources will cause `@rules_rust//tools/rustfmt` to fail with `failed to resolve mod MOD`. To avoid failures, `skip_children = true` is recommended to be set in the workspace's `rustfmt.toml`.

The thing is, that workaround doesn't seem to actually do anything on stable rustfmt anymore. With `skip_children = true` in `rustfmt.toml`, stable rustfmt prints:

```
Warning: can't set `skip_children = true`, unstable features are only available in nightly channel.
```

and then walks the `mod` declarations anyway, producing the failure above. So as far as I can tell the documented mitigation is a no-op on stable.

Curiously, passing the same option as a CLI flag seems to work though:

```
rustfmt --config skip_children=true lib.rs   # exits 0, mod walk skipped
```

Is there some other known workaround?

martingms:
The rustfmt.toml `skip_children` workaround is a no-op on stable rustfmt. Instead, we pass it as a --config flag, which seems to be honored.

Fixes #4022

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,6 +1,1 @@
 # rustfmt options: https://rust-lang.github.io/rustfmt/
-
-# Skipping children allows rustfmt to run on crates with
-# generated source files. Without this formatting will
-# fail with: `failed to resolve mod ${MOD}`.
-skip_children = true
--- a/docs/rust_fmt.vm
+++ b/docs/rust_fmt.vm
@@ -38,14 +38,6 @@ file is used whenever `rustfmt` is run:
 ```text
 build --@rules_rust//rust/settings:rustfmt.toml=//:rustfmt.toml
 ```
-#[[
-### Tips
-]]#
-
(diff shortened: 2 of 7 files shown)


UebelAndre (approved):
Thanks!

PR #4061, opened 2026-05-29
Expose `opt_level` in `rust` extension

UebelAndre:


rust/private/repositories.bzl
```diff
@@ -168,6 +170,9 @@ def rust_register_toolchains(
 
     # Rust toolchains per exec triple
     for exec_triple, name in toolchain_triples.items():
+        extra = extra_target_triples.keys() if type(extra_target_triples) == "dict" else extra_target_triples
```
slackito, line 174: The docstring says `extra_target_triples` is a list:

>         extra_target_triples (list, optional): Additional rust-style targets that rust toolchains should support.

I assume the code is right and the docstring is outdated.

slackito (approved):

rust/private/repositories.bzl
UebelAndre: Yeah, the doc is outdated. I copied docs from an updated function.

Issue #118, opened 2018-08-10
Proper way to statically link rust_binary target with a static non-PIC cc_library


vitalyd:
This may be more of a bazel question but since it's in the context of rust, I figured the audience here is as good as any 😄 .

Suppose we have the following targets:

1. `cc_library` building some static lib
1. `rust_library` FFI wrapper over the above
1. `rust_binary` that uses the `rust_library` 

Pretty standard and simple setup for rust using c/c++ code.

In a `dbg` or `fastbuild` setup, it seems the above just works.  However, when I recently tried this with an `opt` build using a custom cc toolchain, the code failed to link.  Quick note on the custom toolchain - it's very close to the configuration/setup of bazel's builtin crosstool but uses vendored gcc (and related tools).

The link failure was around invalid relocations.  This is because `rustc`, by default, wants to build position-independent (PIE) executables.  To do that, the c/c++ static libs need to be built with `-fPIC` - building the `cc_library` as PIC fixes the issue.  However, this requires modifying a target whose other downstream users may not want to be PIC'd (and PIC can have non-zero perf cost, albeit low from what I gather).

The other fix that I came across is changing the `rust_binary` target to specify a codegen option to `rustc`: `--relocation-model=static`.  This builds a non-PIE binary, and doesn't require `-fPIC` on the dependent cc libs.

This feels ... wrong, however.  Either I'd need to modify all rust binary targets to select static relocation, or I'd need to modify the c/c++ static libs to be PIC.  I also don't know why there's a difference between `dbg`/`fastbuild` and `opt` builds.

Does anyone know if there's a better way to control linkage and relocation model? Ideally (I think), a `rust_binary` would select the linkage it wants (static or dynamic) and the relocation model (static or dynamic), and this choice would flow down to dependent libs, where bazel would build them as either static or dynamic libs and also select `-fPIC` or not.

Curious to hear thoughts/suggestions.

vitalyd:
Curious if anyone has thoughts on this topic? The current annoyance, from my standpoint, is the need to specify `--relocation-model=static` for `rust_binary` targets so they can link property against `cc_library` dependencies in `-c opt` mode.

I suppose one way around this is to configure the rust Bazel toolchain to automatically add this flag to a `rust_binary` target if we can determine that there are `cc_library` dependencies being built without PIC code.  This seems unsatisfying, and hacky.

[...]

PR #4062, opened 2026-05-29
Consume cc deps with PIC when rustc emits PIE binaries

UebelAndre:
Today, rules_rust picks the PIC vs non-PIC variant of cc_library deps based only on the C++ toolchain, feature config, crate type, and compilation mode. But when rustc itself decides to emit a Position-Independent Executable (PIE) for a given target, the Rust objects need PIC-compatible cc deps - otherwise you get link errors from mixing PIE Rust objects with non-PIC C/C++ objects (the failure mode demonstrated by the existing test from commit https://github.com/bazelbuild/rules_rust/commit/70f8fb7814d1b2af7f2cb0c4bdfeebf3e2e47ff4 "Make native_deps test fail with pie").

This PR teaches Bazel when rustc will pass `-pie` to the linker by porting rustc's per-target `position_independent_executables` truth table into Starlark, then feeding that into the PIC selection.

Relates to #118

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -27,6 +27,10 @@ load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
 load(":common.bzl", "rust_common")
 load(":lto.bzl", "construct_lto_arguments")
 load(
+    ":pic_utils.bzl",
+    "should_use_pic",
+)
+load(
     ":providers.bzl",
     "AllocatorLibrariesImplInfo",
     "AllocatorLibrariesInfo",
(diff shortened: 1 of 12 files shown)


rust/private/rustc.bzl
```diff
@@ -157,7 +157,184 @@ def _are_linkstamps_supported(feature_configuration):
             # Is Bazel recent enough to support Starlark linkstamps?
             hasattr(cc_common, "register_linkstamp_compile_action"))
 
-def _should_use_pic(cc_toolchain, feature_configuration, crate_type, compilation_mode):
+_NON_PIE_OS_KEYWORDS = [
```
krasimirgg: Can we extract this, until produces_pie_binaries(), into a new helper file that provides produces_pie_binaries())

krasimirgg (commented):
Very nice! Just one suggestion to extract the bulk of the matching logic into a separate file.

Also, what do you think about adding an explicit override attribute like should_use_pic on the rust_toolchain -- we've got a bunch of manually-declared rust_toolchains, some of them using a custom target json; in more obscure cases like this is easier to specify the pic-ness directly.

UebelAndre:
> Very nice! Just one suggestion to extract the bulk of the matching logic into a separate file.
> 
> Also, what do you think about adding an explicit override attribute like should_use_pic on the rust_toolchain -- we've got a bunch of manually-declared rust_toolchains, some of them using a custom target json; in more obscure cases like this is easier to specify the pic-ness directly.

@krasimirgg updated, please take a look!

krasimirgg (approved):

PR #4063, opened 2026-06-02
Support path mapping in Rust Starlark actions

dabanki:
Transition from passing raw string paths (`.path`) to passing `File` objects directly to action arguments (`Args`). This allows Bazel to mutate paths during execution, which is required for path mapping compatibility.

For more details on path mapping best practices, see: https://github.com/bazelbuild/bazel/discussions/22658

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -834,7 +834,10 @@ def collect_inputs(
 
 def _will_emit_object_file(emit):
     for e in emit:
-        if e == "obj" or e.startswith("obj="):
+        if type(e) in ["tuple", "list"] and len(e) == 2:
+            if e[0] == "obj":
+                return True
+        elif type(e) == "string" and (e == "obj" or e.startswith("obj=")):
             return True
     return False
 
@@ -1173,6 +1176,10 @@ def construct_arguments(
     for kind in emit:
         if kind == "link" and crate_info.type == "bin" and crate_info.output != None:
             rustc_flags.add(crate_info.output, format = "--emit=link=%s")
+        elif type(kind) in ["tuple", "list"] and len(kind) == 2:
+            # 'kind' is a (string, File) tuple/list. Passing the File object directly to
+            # Args.add allows Bazel to perform path mapping on the path.
+            rustc_flags.add(kind[1], format = "--emit=" + kind[0] + "=%s")
         else:
             emit_without_paths.append(kind)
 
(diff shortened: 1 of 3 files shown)


UebelAndre:
Wasn't this done in https://github.com/bazelbuild/rules_rust/pull/4011 ? Is there a gap in the regression testing?

dabanki:
https://github.com/bazelbuild/rules_rust/pull/4011 seems have covered most things. We just noticed that target_flag_value was not updated to pass in File object instead of path (as string). When doing e2e testing for some of our targets this was causing failures for us

UebelAndre:
> #4011 seems have covered most things. We just noticed that target_flag_value was not updated to pass in File object instead of path (as string). When doing e2e testing for some of our targets this was causing failures for us

Fixes are welcome but can you also add regression testing for the missed case?

dabanki:
I'm not familiar with this codebase, so the best way I could find to do this was to add an assertion on the flag type which implies that path mapping is supported. Let me know if that works.

rust/private/rustc.bzl
```diff
@@ -834,7 +834,9 @@ def collect_inputs(
 
 def _will_emit_object_file(emit):
     for e in emit:
-        if e == "obj" or e.startswith("obj="):
+        # If 'e' is a File object (not a string), it represents an object file
```
krasimirgg: Could we model such elements of emit as (string, Path) tuples?
We've got some in-progress extensions internally that we'd like to upstream as soon as we're confident they work properly that use a similar pattern, such as `--emit=thin-link-bitcode=<path>`

UebelAndre:
> I'm not familiar with this codebase, so the best way I could find to do this was to add an assertion on the flag type which implies that path mapping is supported. Let me know if that works.

I think an integration test would be best here. Can you define some target that naturally triggers this code path?

krasimirgg:
>>I'm not familiar with this codebase, so the best way I could find to do this was to add an assertion on the flag type which implies that path mapping is supported. Let me know if that works.
>
>I think an integration test would be best here. Can you define some target that naturally triggers this code path?

To provide more context as someone on both sides --
Happy to help out with setting up an integration test, (but we may need to add some additional features to rules_rust / toolchain-repository-sets to be able to set this up):

Internally we define rust_toolchains manually; we use the target_json functionality.

To get to a behavior that this changes, you need something like that: a rust library that's set up to use a rust toolchain that uses a target json:

```
# MODULE.bazel

bazel_dep(name = "rules_rust", version = "0.70.0")

# Configure a toolchain with a custom target specification file
rust.toolchain(
    name = "custom_toolchain",
    edition = "2021",
    versions = ["1.85.0"],
    # Path to your custom target specification JSON file
    # doesn't work today: 
    # Error: in 'toolchain' tag, unknown attribute 'target_json' provided
    target_json = "//tools/platforms:my_custom_target.json", 
)

use_repo(rust, "rust_toolchains")
```

I don't know of a good way to spawn such a custom toolchain. @UebelAndre  -- do you have suggestions?

krasimirgg:
Chatted about the (lack of) regression test coverage for this yesterday with @UebelAndre. I've taken up https://github.com/bazelbuild/rules_rust/issues/4078 to set this up. Let's focus on the technicals in this PR, and I'll make sure to add some test coverage later as part of that bug.

krasimirgg (changes requested):

rust/private/rustc.bzl
dabanki: Newest pushed commit should address this if I understand correctly - let me know if it looks like what you had in mind.

krasimirgg (approved):
Thank you!

PR #4040, opened 2026-05-14
Fix incorrect exec transition for cargo_build_script

UebelAndre:
Relying on an exec configuration through targets consumed in `target` leads to the potential for `cfg = "exec"` to be evaluated differently in the transitive target (in this case, `cargo_build_script_runfiles.script`) and the consumer (`cargo_build_script.data_runfiles`). This change moves the script directly back to `cargo_build_script` in `cfg = "exec"` and separately uses a `cfg = "target"` intermediate rule to implement runfiles (solving for https://github.com/bazelbuild/bazel/issues/15486).

closes https://github.com/bazelbuild/rules_rust/issues/3201 https://github.com/bazelbuild/rules_rust/issues/3893

PR #4064, opened 2026-06-02
Add test for regression after #4040.

slackito:


test/unit/cargo_build_script_data_runfiles/BUILD
```diff
@@ -0,0 +1,21 @@
...
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+cargo_build_script(
+    name = "build_script",
+    srcs = ["build.rs"],
+    data = ["data.txt"],
+    deps = [
+        "@rules_rust//tools/runfiles",
```
UebelAndre, line 13: Ah, I didn't consider cases where `cargo_build_script` uses the runfiles library. Can you expand on the need for this? For the most part, I would have expected them to only operate on execpath or `CARGO_MANIFETST_DIR` paths. We could export the necessary values for `script` runfiles but `data` wouldn't be available and I don't think it's a good idea to try to recreate any kind of repo mapping file to be able to use the merged runfiles dir. Do you have thoughts there?

UebelAndre (changes requested):

test/unit/cargo_build_script_data_runfiles/BUILD
slackito: In the particular target that started failing after your commit, someone wanted to embed some config files in a `rust_binary`, and they used a `cargo_build_script` target to read .toml files and generate rust code containing the config values.

That target doesn't need to be a cargo_build_script. In fact, it used to be a gen_embedded_configs.py `py_binary` and a `genrule` that ran the binary before a user switched it to a cargo_build_script because it's "the Rust way". So we can switch it back.

What I don't know is what to do with `cargo_build_script` data dependencies in general. We could say data files are not accessible as runfiles from the build script. But a `cargo_build_script` can have arbitrary `rust_library` targets as `deps`, and these can have `data` deps and presumably need to access them. But right now they'd get the cargo_runfiles tree that we generate and fail in the same way, right?

For all I know, maybe we'll end up telling users "don't do that". But I don't feel qualified to make that call yet :|

UebelAndre: Does https://github.com/bazelbuild/rules_rust/pull/4066 work for you?

PR #4066, opened 2026-06-03
Add support for build script runfiles

UebelAndre:
This change restores support for runfiles in `cargo_build_script.script`. There exists one limitation where only `cargo_build_script.tools` can be located via runfiles. Any `cargo_build_script.data` files should be located relative to the `CARGO_MANIFEST_DIR` env var. This is required as there is no existing interface for merging exec (tools) and target (data) runfiles.

closes https://github.com/bazelbuild/rules_rust/pull/4064


slackito (approved):
Thank you!

PR #4067, opened 2026-06-03
Prefix execroot-relative file arguments with pwd

martin4861:
Prefix execroot-relative bare positional file args with pwd in cargo build scripts. This is already done for `-L`, `-isystem`, `--sysroot`, etc but not for bare positional arguments like `bazel-out/bin/compiler-rt/libclang_rt.builtins.static.a`.

This problem arised when compiling `aws-lc-sys` with [hermetic-llvm](https://github.com/hermeticbuild/hermetic-llvm), see https://github.com/hermeticbuild/hermetic-llvm/issues/405

A minimal working example to reproduce is created in https://github.com/martin4861/mwe-bootstrapped-rust-clang_rt.builtins.static:
```
clang: error: no such file or directory:
'bazel-out/darwin_arm64-fastbuild-ST-bdec89fd5d65/bin/external/llvm++llvm_source+compiler-rt/clang_rt.builtins.static_/libclang_rt.builtins.static.a'
```

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -207,12 +207,39 @@ def _pwd_flags_resource_dir(args):
     """Prefix execroot-relative paths in -resource-dir arguments with ${pwd}."""
     return _prefix_pwd_to_flag(args, ["-resource-dir=", "-resource-dir"])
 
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
LGTM, thanks!

PR #4070, opened 2026-06-06
Fix static link failures when cc_library linkopts precede later archives

HackAttack:
rules_rust links native archives twice around rlibs. In the second pass, each cc_library's linkopts were emitted as `--codegen=link-arg=` right after that dep's own archives, not after every archive in the pass. Since static linkers resolve left-to-right, a linkopt on an early dep can be “spent,” and then later archives that still need the library it names get undefined references.

Example: rust_binary depends on cc_library B which depends on cc_library A. A uses `fstat` and states `linkopts = ["-lc"]`. -lc is applied while resolving A; B's archives still need libc but -lc has already been passed.

The fix is to emit CcInfo user_link_flags after all native archive flags in _add_native_link_flags, via add_user_link_flags from _get_make_link_flag_funcs.

This is not always reproducible; host gcc may not exhibit the issue because -lc is added again after the archive list, but custom linkers and -nodefaultlibs can expose it.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -2472,8 +2472,19 @@ def portable_link_flags(
 
     return []
 
-def _add_user_link_flags(ret, linker_input):
-    ret.extend(["--codegen=link-arg={}".format(flag) for flag in linker_input.user_link_flags])
+def _add_user_link_flags(make_link_flags_args):
+    linker_input = make_link_flags_args[0]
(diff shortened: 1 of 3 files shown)


UebelAndre (changes requested):
Thanks! I had a chat with other maintainers and we don't fully understand the change. Is there a repro you can add as a regression test that would demonstrate the issue? And is there some prior-art from Cargo that could be used as a reference?

HackAttack:
My bad, the bug was actually me putting `linkopts` on the wrong library!

PR #4073, opened 2026-06-08
Added support for rustc self-profiling

1vipulgupta:
Introduced a new build setting, `--//rust/settings:zself_profile_events`, to enable rustc's `-Zself-profile` and `-Zself-profile-events` flag. Rust `nightly` toolchain is required for using this feature.
The setting accepts patterns to match against crate labels or execution paths, allowing fine-grained control over which crates are profiled.  
Profiling data is output to a `<crate_name>_self-profile` directory.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -53,6 +53,7 @@ load(
     "is_exec_configuration",
     "is_std_dylib",
     "make_static_lib_symlink",
+    "matches_prefix_filter",
     "parse_env_strings",
     "relativize",
 )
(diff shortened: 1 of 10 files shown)


rust/private/rustc.bzl
```diff
@@ -2937,3 +2984,37 @@ no_std = rule(
...
+
+zself_profile_events = rule(
+    doc = (
+        "Passes -Zself-profile and -Zself-profile-events flags to matching Rust crates." +
+        "This feature allows end-users to profile rustc compiler performance on specific crates " +
+        "using rustc's self-profiler. Because these flags are unstable, using them requires a " +
+        "nightly compiler toolchain. The setting is configured from the command line via " +
+        "`--@rules_rust//rust/settings:zself_profile_events`." +
+        "The expected value format is `<prefix_filter>@<events_specification>`. Multiple uses of " +
+        "this flag are accumulated: " +
```
krasimirgg: nit: is it intended that this sentence ends in `:`?

rust/private/rustc.bzl
```diff
@@ -2937,3 +2984,37 @@ no_std = rule(
...
+        "If the target prefix matches with <prefix_filter>, `-Zself-profile` and `-Zself-profile-events` " +
```
krasimirgg, line 3019: Just confirming something: after reading the setup_zself_profile implementation, the event types are collected from the first matching filter:
For instance, here the event types would be `foo`:
`--@rules_rust//rust/settings:zself_profile_events=//my/project@foo --@rules_rust//rust/settings:zself_profile_events=//my/project@bar`
I think this is OK, but could you clarify? Especially in the context of the previous sentence "Multiple uses of this flag are accumulated" -- a user reading this may interpret this as that in the example above, both foo and bar are collected, so there's potential for confusion.

krasimirgg (changes requested):
This is very nice!
Let's also add a test (let's chat internally about the mechanics of testing this).

rust/private/rustc.bzl
1vipulgupta: Updated the doc stating one first will be applied.

rust/private/rustc.bzl
1vipulgupta: Fixed. It was not intentional.

krasimirgg (approved):
Thank you!

PR #4079, opened 2026-06-11
Fix coverage for crates with mixed sources

martingms:
Crates with generated sources are compiled from the output tree, so their coverage mapping records bazel-out/... paths, which Bazel's lcov merger silently drops. Remap the prefix away so coverage paths come out workspace-relative.

I'm not 100% confident this is the correct fix, but it fixes the coverage for us in crates that has both in-tree and generated src files.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1303,6 +1303,15 @@ def construct_arguments(
         # https://doc.rust-lang.org/rustc/instrument-coverage.html
         rustc_flags.add("--codegen=instrument-coverage")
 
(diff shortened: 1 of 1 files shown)


UebelAndre (changes requested):
Hey! Sorry for the delay, thank you for the fix! I had to think about the testing here and finally was able to come up with https://github.com/bazelbuild/rules_rust/pull/4106. If you want to cherry-pick those changes I'd be happy to merge! I just put the PR up since I don't know how I would have effectively represented all of it on this one.

PR #4080, opened 2026-06-11
Fix rust_doc_test linking with native libraries

patrickmscott:
rust_doc_test does not propagate native libraries into the final test command. This appears to only affect native libraries built by bazel rather than ones linked into crates by build.rs. This change only works with static libraries - dynamic libraries are much more complex.

For example, using the jemalloc bazel module in tikv-jemalloc-sys.

* Add -Clink-arg for all transitive non-crates
* Add include_linker_inputs to collect all linker inputs
* Collect linker roots from all libraries and additional inputs

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -651,6 +651,7 @@ def collect_inputs(
         stamp = False,
         force_depend_on_objects = False,
         experimental_use_cc_common_link = False,
+        include_linker_inputs = False,
         include_link_flags = True):
     """Gather's the inputs and required input information for a rustc action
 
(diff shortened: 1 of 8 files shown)


rust/private/rustc.bzl
```diff
@@ -2625,8 +2627,6 @@ def _add_native_link_flags(
         use_direct_link_driver (bool): Whether the linker is a direct driver (e.g. `ld`, `wasm-ld`) vs a wrapper (e.g. `clang`, `gcc`).
         include_link_flags (bool, optional): Whether to include flags like `-l` that instruct the linker to search for a library.
     """
-    if crate_type in ["lib", "rlib"]:
```
UebelAndre, line 2628: This seems somewhat risky, can you expand on why this is removed?

UebelAndre (changes requested):
Thanks! Just one question!

rust/private/rustc.bzl
patrickmscott: `_add_native_link_flags` is called by `construct_arguments` and is already guarded by

`if ("link" in emit and crate_info.type not in ["rlib", "lib"]) or add_flags_for_binary`

In the rustdoc test, `add_flags_for_binary` was already true, but then `_add_native_link_flags` bails early. I thought it was redundant to check again for ["rlib", "lib"] and seemed like the right thing to do if `add_flags_for_binary` is true.

UebelAndre (approved):
Thanks!

PR #4081, opened 2026-06-11
Update label_injections to work transitively

UebelAndre:
Rendering canonical labels into lockfiles is sensitive to consumers potentially having overrides. Instead, injections are applied on the fly.


illicitonion (approved):
Very nice! Thank you!

PR #4083, opened 2026-06-12
Stop removing AR from macOS env

keith:
With https://github.com/bazelbuild/apple_support/pull/601 we make our
libtool wrapper normalize ar arguments so we can handle this case and
still produce hermetic archives (which Apple's ar does not)

### DIFF
--- a/cargo/private/cargo_build_script_runner/bin.rs
+++ b/cargo/private/cargo_build_script_runner/bin.rs
@@ -136,19 +136,7 @@ fn run_buildrs() -> Result<(), String> {
     }
 
     if let Some(ar_path) = env::var_os("AR") {
-        // The default OSX toolchain uses libtool as ar_executable not ar.
-        // This doesn't work when used as $AR, so simply don't set it - tools will probably fall back to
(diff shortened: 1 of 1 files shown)


keith:
im submitting this as documentation but im not sure it's really mergable since different toolchains could have different capabilities here

UebelAndre:
You could introduce an incompatible flag for this to nudge folks over to upgrading their toolchains

keith:
but they could also use a non-apple_support toolchain which wouldn't help this, like using @llvm / toolchains_llvm / the raw rules_cc toolchain

UebelAndre:
> but they could also use a non-apple_support toolchain which wouldn't help this, like using @llvm / toolchains_llvm / the raw rules_cc toolchain

Couldn't they constrain their toolchains on that? But that direction aside, how can I help move this forward?

keith:
the more i think about it i think the only option is to remove `libtool` from the name of our wrapper since multiple places use libtool to differentiate this, and the executable could still be a libtool where this logic is needed

PR #4085, opened 2026-06-13
Update pyo3 to 0.29.0

mypmc:
To support doc comments generation, pyo3 0.28.2
does not include Rust doc comments in generated
.pyi stubs.

The patch that rules_rust_pyo3 applied to pyo3 0.28.2
is not needed for now: PyO3/pyo3#5862 removed
`resolve_cross_compile_config_path`.
The cross-build-script file dependency that caused the
sandbox path issue no longer exists.

- https://github.com/PyO3/pyo3/issues/4579


UebelAndre (approved):
Awesome! Thank you so much!

PR #4087, opened 2026-06-16
Move rust/toolchain.bzl implementation into rust/private

UebelAndre:
This change is a noop for our public interface but allows for more helper functions to exist in `toolchains.bzl` without them expanding the public interface.


LuuOW (commented):
Technical audit: code patterns and implementation verified for alignment with modern software engineering standards.

slackito (commented):
Moving rust/toolchain.bzl to rust/private/toolchain.bzl sounds good for the reasons in the PR description. What's the reason for moving rust/toolchain.bzl to rust/rust_toolchain.bzl and then needing to keep rust/toolchain.bzl for backwards compatibility? (I'm not necessarily opposed to it, just want to understand)

UebelAndre:
Yeah, keeping toolchain.bzl is for backward compatibility

illicitonion (commented):
Is there a reason for the `rust_` prefix? Is that becoming a standard in the ecosystem? I don't think I've seen consistent approaches from rulesets, so not sure what `rust_toolchain.bzl` vs `toolchain.bzl` is buying?

UebelAndre:
typically each rule is exposed in it's own `.bzl` file, `rust_toolchain` violates this expectation. I would delete `toolchain.bzl` but I don't want to break users. I would also be willing to drop the extra exports from `rust_toolchain.bzl` but landed on this arbitrary decision.

UebelAndre:
I've updated the change to have more explicit top level bzl files.

slackito:
@illicitonion expressed my question more clearly :)

If the standard convention is that each rule is exposed in its own .bzl file, then I'm fine with this.

Do we need a deprecation plan to eventually delete toolchain.bzl?

UebelAndre:
> Do we need a deprecation plan to eventually delete toolchain.bzl?

We can formalize one but at some point in the future when enough releases have been out, I would probably delete it or can put a print in it warning for a release.

slackito (approved):
Sounds good to me.

PR #4088, opened 2026-06-17
fix(cargo_build_script): expand ${pwd} for execpaths and locations in build_script_env

psalaberria002:
Singular `$(execpath ...)` and `$(location ...)` were already prefixed with `${pwd}/` so build_script_runner could resolve them to absolute paths at run time. The plural forms `$(execpaths ...)` and `$(locations ...)` were silently omitted from this handling, leaving their paths as execroot-relative.

For plural forms each macro is expanded individually (via ctx.expand_location) and every resulting space-separated path is prefixed with `${pwd}/`.

### DIFF
--- a/rust/private/utils.bzl
+++ b/rust/private/utils.bzl
@@ -286,6 +286,26 @@ def _expand_location_for_build_script_runner(ctx, v, data, known_variables):
         if directive in v:
             # build script runner will expand pwd to execroot for us
             v = v.replace(directive, "$${pwd}/" + directive)
+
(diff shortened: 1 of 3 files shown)


slackito (commented):
Could you please add tests for your fix? There's a location expansion test at //test/unit/location_expansion:location_expansion_test_suite but it only covers `execpath`, and it does so in the `rustc_flags` of a `rust_library` target. I think you could add testing for the plural forms there.

You might also want to add another test for cargo_build_script to check that your intended use case is covered too.

slackito (approved):
Thanks!

PR #4093, opened 2026-06-18
Render crates into subpackages

UebelAndre:
Currently `crate_universe` renders aliases into a single package for developers to consume. The issue with this design is that for projects with massive amounts of dependencies, Bazel parsing that BUILD file can be slow and have noticable impacts on analysis. This change updates `crate_universe` to render aliases into sub-packages instead of a root BUILD file in the repo hubs. In practice this means dependencies that were once accessed via `@crate_index//:clap` are now accessed via `@crate_index//clap`. Currently both root level aliases are still rendered but can be disabled via the new `incompatible_no_root_alias_targets` flag.


UebelAndre:
For reviewers, this change should be reviewed commit-by-commit.

UebelAndre:
This change is being stacked on top of https://github.com/bazelbuild/rules_rust/pull/4094

UebelAndre:
These changes are now stacked on top of https://github.com/bazelbuild/rules_rust/pull/4098

illicitonion (approved):

PR #4095, opened 2026-06-20
Remove C Python headers from pyo3 build actions

Ahajha:
https://github.com/bazelbuild/rules_rust/pull/4024 introduced a warning for targets adding CC deps in `deps`. The warning calls out putting this in `link_deps` or `cc_deps`. The latter makes more sense here... but doesn't actually exist. However, I don't think this matters anyways, as I don't think it's needed (this works with an internal codebase).


UebelAndre (approved):
Thanks! Doesn't look like anything here is failing either so seems like a reasonable delete

PR #4097, opened 2026-06-23
cargo_build_script: propagate cdylib and binary link args

### LINKED ISSUE #1062: Support `cargo:rustc-link-arg-*` instructions in build runner
https://github.com/bazelbuild/rules_rust/pull/1061 is adding support for the `cargo:rustc-link-arg` instruction but `cargo:rust-link-arg-bins`, `cargo:rustc-link-arg-bin` and `cargo:rustc-cdylib-link-arg` are still not supported (they just print a warning).

Supporting them is a little more complex since it requires splitting some of the build script output into crate types. But that should be doable.

ashi009:
Fixes #1062.

A build script's `cargo::rustc-cdylib-link-arg` and `cargo::rustc-link-arg-bins` were dropped with a warning, so flags like napi-rs's `-undefined dynamic_lookup` never reached the linker.

Route them through `BuildInfo` to the targets cargo applies them to:
* `rustc-cdylib-link-arg` → `BuildInfo.cdylib_link_flags`, applied to cdylib crates and propagated transitively (matching cargo, rust-lang/cargo#9562).
* `rustc-link-arg-bins` → `BuildInfo.bin_link_flags`, applied to binary crates.

Each set is written to its own flag file, added to the rustc action as an `--arg-file`, so a non-matching crate type never reads it. The per-binary `rustc-link-arg-bin=BIN=FLAG` stays unsupported — a build script's outputs are shared by every consumer, so a flag can't be scoped to one named binary.


UebelAndre (changes requested):
Thanks! Can you add a test for somewhere under `//cargo/tests/cargo_build_script` that confirms the new flags are behaving as expected for binary and library consumers?

ashi009:
@UebelAndre added

UebelAndre (approved):
Thank you! Looking good! Just need to fix buildifier

PR #4100, opened 2026-06-24
Fix bindgen flag merging between clang_flags and cc_common

konkers:
If a param flag (a flag with a following argument like -isystem) exists both in the flags returned by
`cc_common.get_memory_inefficient_command_line` and a `rust_bindgen_library`'s `clang_flags`, the secondary argument of the flag from `cc_common` was getting dropped.


UebelAndre (approved):
Thank you!

PR #4102, opened 2026-06-25
fix(crate_universe): fix OUTPUT_BASE env var value format in crates_vendor

georgesfarah:
## Summary

Fix the `OUTPUT_BASE` environment variable override in `crate_universe/src/cli/vendor.rs` — it has never worked correctly.

## Bug

The `bazel info` output parser splits each line on `:` and stores only the part after the colon as the HashMap value:

```rust
let (k, v) = line.split_at(line.find(':')?);
Ok((k.to_string(), (v[1..]).trim().to_string()))
```

So for output like `output_base: /data/output`, the HashMap contains `{"output_base" => "/data/output"}`.

However, the `OUTPUT_BASE` env var override stores the **full formatted string** as the value:

```rust
bazel_info.insert("output_base".to_owned(), format!("output_base: {}", path));
```

This produces `{"output_base" => "output_base: /data/output"}`. When `TryFrom` later converts this to a `PathBuf` via `.map(Into::into)`, it produces `PathBuf("output_base: /data/output")` — an invalid path.

## Fix

Store the raw path string without the prefix:

```rust
bazel_info.insert("output_base".to_owned(), path);
```

Also extract `parse_bazel_info` from `try_new` so the parsing and env var override logic can be unit-tested without spawning a `bazel info` subprocess.

## Test plan

Added `test_parse_bazel_info_output_base_env_override` which verifies:
- Without `OUTPUT_BASE` set, the parsed value from `bazel info` output is used
- With `OUTPUT_BASE` set, it overrides the parsed value with the correct path

## Note

The test uses `env::set_var`/`env::remove_var`, which are `unsafe` in Rust 1.83+ (and required to be in `unsafe` blocks starting with edition 2024). This matches the existing test style in this file (see `test_bazel_info`). A follow-up could migrate both tests to use `unsafe {}` blocks or `#[serial_test::serial]` if the project upgrades its Rust edition.


abrisco (approved):
Thank you!

UebelAndre (approved):
Thank you!

PR #4109, opened 2026-06-26
Add `link_deps` crate_universe annotation

UebelAndre:
https://github.com/bazelbuild/rules_rust/pull/4024 now adds noisy prints that users have no way to handle when using crate_universe. This change adds annotations so the prints can be actioned.

### DIFF
--- a/crate_universe/src/config/label_injection.rs
+++ b/crate_universe/src/config/label_injection.rs
@@ -43,6 +43,8 @@ use std::collections::BTreeMap;
 use anyhow::{bail, Result};
 use serde_json::{Map, Value};
 
+use crate::utils::starlark::is_repo_name_byte;
+
 /// Walk `config.annotations` and:
 ///
 ///   * strip each annotation's `label_injections` field (so typed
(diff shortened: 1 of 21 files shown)


crate_universe/src/config/label_injection.rs
```diff
@@ -160,10 +162,95 @@ fn rewrite(value: &mut Value, mapping: &BTreeMap<String, String>) {
...
+    let apparent_target = apparent.strip_prefix('@').unwrap_or(apparent);
+
+    let bytes = input.as_bytes();
+    let mut out = String::with_capacity(input.len());
+    let mut last_end = 0;
+    for (start, _) in input.match_indices(apparent) {
+        if start > 0 && bytes[start - 1] == b'@' {
+            continue;
         }
+        let after = start + apparent.len();
```
slackito, line 237: Are the indices in `input` (a `str`) guaranteed to match the indices in `bytes`? (i.e. is `input` only made of non-multibyte characters?)

slackito (approved):

crate_universe/src/config/label_injection.rs
UebelAndre: Yeah, based on this:

https://doc.rust-lang.org/std/primitive.str.html#method.len

> This length is in bytes, not [char](https://doc.rust-lang.org/std/primitive.char.html)s or graphemes. In other words, it might not be what a human considers the length of the string.

PR #4005, opened 2026-04-29
Remove WORKSPACE support

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/3818

PR #4114, opened 2026-06-27
Restore `crate_universe/extension.bzl` shim with a deprecation warning

### LINKED ISSUE #4111: crate_universe/extension.bzl removal breaks at least protobuf
**Description**
`crate_universe/extension.bzl` was the original bzlmod entry point (#1910).
After `crate_universe/extensions.bzl` superseded it (#3034), `extension.bzl` was kept as a deprecated re-export of `extensions.bzl`.
0.71.0 removed `extension.bzl` (#4005).

Any module that loads `@rules_rust//crate_universe:extension.bzl` then fails to resolve.
It would be nice to bring the old file back, so its removal does not break existing consumers without at least a prior warning.

**Reproduction steps**
Create a root module with this MODULE.bazel:
```
bazel_dep(name = "rules_rust", version = "0.71.0")
bazel_dep(name = "protobuf", version = "35.1")
```
protobuf 35.1 is the newest release on BCR and loads `@rules_rust//crate_universe:extension.bzl`.
Run `bazel mod deps`.
It fails because `crate_universe/extension.bzl` no longer exists.

**Additional context**
```
cannot load '@@rules_rust+//crate_universe:extension.bzl': no such file, requested by https://bcr.bazel.build/modules/protobuf/35.1/MODULE.bazel:197:22
```

The file was removed in #4005 (Remove WORKSPACE support), but it is loaded from bzlmod MODULE.bazel files via `use_extension`, as the error above shows.

**Impact**
Updating to 0.71.0 or later then requires patching the MODULE.bazel of every transitive module that loads the old path, because a consumer cannot otherwise change those load statements.
This affects protobuf, the newest release on BCR.

We recognize rules_rust has not yet adopted semantic versioning, so breaking changes are expected on 0.x.
Because this break nonetheless reaches foundational rules such as protobuf, restoring `crate_universe/extension.bzl` for a short deprecation window of a few weeks would let those modules migrate first.
A warning printed at load time from the compatibility shim would help downstream maintainers realize they're better off adjusting their use cases:
- #4114.

**Bazel and rules_rust version**
Bazel 9.1.1
rules_rust 0.71.0
protobuf 35.1

rdesgroppes:
#4005 removed `crate_universe/extension.bzl`, the deprecated re-export of `crate_universe/extensions.bzl`, as part of dropping WORKSPACE support.

Downstream projects may still load `crate` from that path via `use_extension`, so any build that resolves `rules_rust` 0.71.0 with such a module in its graph fails.
`protobuf` 35.1, the newest release on BCR, is one such module.

Restore the shim and print a deprecation warning at load time, giving downstream modules a short window to migrate to `extensions.bzl`, e.g.:
- protocolbuffers/protobuf#28226.

Closes #4111.

### DIFF
--- a/crate_universe/extension.bzl
+++ b/crate_universe/extension.bzl
@@ -1,0 +1,11 @@
+"""Deprecated, use `:extensions.bzl`."""
+
+load(":extensions.bzl", _crate = "crate")
+
+# buildifier: disable=print
+print(
+    "\nWARNING: @rules_rust//crate_universe:extension.bzl is deprecated and will be removed in a future release. " +
+    "Load `crate` from @rules_rust//crate_universe:extensions.bzl instead.",
+)
+
+crate = _crate


UebelAndre (approved):
Sounds good to me! Thank you for putting up this pull request! I'll get a release out ASAP.

PR #4117, opened 2026-06-29
Disable path mapping when rustc_env has location expansions

airtrick-rgb:
Location expansion happens outside the control of path mapping logic and would produce paths with build configuration in them, breaking path mapping's logic of replacing all build configuration with the literal "cfg".


UebelAndre (approved):
Thank you!

UebelAndre:
I added some regression testing for this in https://github.com/bazelbuild/rules_rust/pull/4120

PR #4127, opened 2026-07-03
Tolerate partial build failures during rust-analyzer discovery

mgeisler:
`discover_bazel_rust_project` builds the rust-analyzer aspect over the whole workspace and bailed on any non-zero Bazel exit. In a large monorepo a single broken target (e.g. an unrelated, unmaintained one) aborted the build, leaving rust-analyzer with no Bazel project model. It then silently falls back to a Cargo workspace, which breaks proc-macro expansion (toolchain ABI mismatch) and materializes a stray `target/`.

Pass `--keep_going` so Bazel keeps building the rest of the graph, and treat the result as usable when the BEP still yielded crate specs: proceed with a `log::warn` that the project may be incomplete, and only fail when zero specs were produced. This mirrors `flycheck`'s existing exit-code handling.

Add an end-to-end test that stands up a workspace with one good `rust_library` and one target that fails analysis, then checks discovery still returns a project containing the good crate. Unit tests cover the `assess_discovery` decision.

Written with Claude, but I believe I understand what is going on.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -399,6 +399,7 @@ tasks:
       - "//tools/rust_analyzer:gen_rust_project"
       - "//tools/rust_analyzer:discover_bazel_rust_project"
       - "//test/rust_analyzer:rust_analyzer_test"
+      - "//test/rust_analyzer:partial_failure_test"
   ide_integration_tests_macos:
     name: IDE VSCode Tests
     platform: macos_arm64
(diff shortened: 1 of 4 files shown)


UebelAndre (approved):
Thanks! Seems reasonable to me 😄

mgeisler:
Thanks a lot, and thanks for your great changes to make `rust-analyzer` easy to use!

PR #4128, opened 2026-07-04
fix: declare rustfmt aspect action toolchain

keegancsmith:
The --incompatible_auto_exec_groups flag requires Starlark actions to declare whether their executable/tools come from an action toolchain. This is not a Bazel default today, but it is exposed through strict preset configurations and is intended to catch rules that rely on implicit action toolchain inference.

rustfmt_aspect runs rustfmt from rules_rust's rustfmt toolchain, but the formatting action does not declare that action toolchain. Declare the matching rustfmt toolchain so the action's toolchain behavior is explicit without changing its inputs or execution model.


UebelAndre (approved):
Thank you!

keegancsmith:
CI failures look unrelated since they seem like the same failures on main?

PR #4075, opened 2026-06-10
Add `discoverConfig` Support for rust-analyzer

UebelAndre:
This change improves rust-analyzer support by moving away from `rust-project.json` and promoting a more automated approach. Documentation has been updated to reflect the recommended integration.

Changes:
- Implement Build-Event-Service parsing to avoid globbing for bazel outputs.
- Implements new `setup` tool that users can invoke to configure their IDE for Bazel-backed rust-analyzer support.

closes https://github.com/bazelbuild/rules_rust/issues/2488
closes https://github.com/bazelbuild/rules_rust/issues/3762

PR #4130, opened 2026-07-07
Fixes BEP path handling for rust analyzer specs

csmulhern:
The rewrite of the Rust Analyzer auto-discovery pipeline (see #4075) introduced a bug for users who use `--experimental_convenience_symlinks=ignore` (i.e. have no bazel-out symlink in their workspace).

`parse_output_group_paths` documentation suggests that output paths are always absolute

https://github.com/bazelbuild/rules_rust/blob/dedf05304354aace6a76ae78030a9d69c74e37d4/tools/rust_analyzer/bep.rs#L107-L110

but that is not true for BEP events that trigger the fallback reconstruction path

https://github.com/bazelbuild/rules_rust/blob/dedf05304354aace6a76ae78030a9d69c74e37d4/tools/rust_analyzer/bep.rs#L241-L243

This PR detects those cases and anchors those paths to the execution root, making them also absolute.

### DIFF
--- a/tools/rust_analyzer/bep.rs
+++ b/tools/rust_analyzer/bep.rs
@@ -110,6 +110,7 @@ struct OutputGroup {
 pub fn parse_output_group_paths(
     bep_path: &Utf8Path,
     output_group: &str,
+    execution_root: &Utf8Path,
 ) -> Result<Vec<Utf8PathBuf>> {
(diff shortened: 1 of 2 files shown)


csmulhern:
cc @UebelAndre

UebelAndre (approved):
Nice! Thank you!

PR #4131, opened 2026-07-07
Handle PIC/NOPIC correctly in establish_cc_info

jakos-sec:
This change determines whether to use `pic_static_library` or `static_library` in `cc_common.create_library_to_link` based on the `should_use_pic` helper. It also updates the unit tests to expect only one of the libraries depending on the configuration.


krasimirgg (approved):
Thank you!

PR #4142, opened 2026-07-09
Add `root_path` attribute for directory crate roots

krasimirgg:
Allows specifying `root_path` when the crate root (`srcs` entrypoint) is a directory artifact (`TreeArtifact`). This enables rules_rust targets (`rust_library`, `rust_binary`, `rust_test`, etc.) to consume directory artifacts produced by upstream rules and correctly construct the path to the crate root (`lib.rs` / `main.rs`).

We use something like this internally to build standard libraries from sources distributed as .zip source archives: a custom rule extracts the contents into a generated directory artifact, which then the rust rules can consume as sources.

This may be helpful to achieve the request from https://github.com/bazelbuild/rules_rust/issues/3757, where the user is trying to consume a directory of generated sources.

* Adds `root_path` attribute to `_COMMON_ATTRS` and `CrateInfo` provider.
* Updates `rustc_compile_action` argument construction (`rustc.bzl`) to join `crate_info.root.path` with `root_path` when `crate_info.root.is_directory` is true.
* Adds a unit and build test case under `test/root_path/` using a custom rule that packages its `srcs` as a directory artifact (`TreeArtifact`) and compiles a `rust_library` and `rust_test` from it.

I have taken special care in the rustc argument construction to ensure it is compatible with the new [bazel path remapping guidelines](https://github.com/bazelbuild/bazel/discussions/22658).

While here, added a few checks in the rules for combinations of `srcs`, `crate` and `crate_root` attributes that don't make sense and would cause compilation to fail later anyways.


slackito (approved):

rust/private/rustc.bzl
```diff
@@ -2871,6 +2872,13 @@ def _add_native_link_flags(
                     format_each = "-lstatic=%s",
                 )
 
+def _get_crate_root_path(args):
```
UebelAndre, line 2875: @krasimirgg nit: Can we try to ensure all functions have docstrings with typed arguments? In the absence of type hints this is incredibly helpful for identifying broken contracts and bugs.

krasimirgg: https://github.com/bazelbuild/rules_rust/pull/4162

PR #4146, opened 2026-07-09
Return repo_metadata from repository rules

keith:
This allows the remote repo contents cache to cache the result of these


UebelAndre (approved):
Thank you!

PR #4156, opened 2026-07-17
Add lint test rules for transitive targets

alexconrey:
Changes:
- Add `rust_clippy_test` for having a test target that runs clippy on a list of targets
- Add a `transitive` bool to `rustfmt_test` that aggregates `rustfmt_aspect` outputs from deps of `targets`. This behavior also applies to the new `rust_clippy_test`


UebelAndre (approved):
Thanks!

PR #4158, opened 2026-07-20
Reexport output groups for rustfmt and clippy tests

alexconrey:
This change updates `rustfmt_test` and `rust_clippy_test` to behave more consistently with the aspect output groups by ensuring transitive targets in the transitioned configuration get added to `//...` invocations.


UebelAndre (approved):
Thank you!

PR #4159, opened 2026-07-21
Fix NoneType error in rustdoc

UebelAndre:
There seemed to be a confusion around `CrateInfo.output` being optional within the rustdoc rules that leads to a `NoneType` error when using rustdoc on a generated crate root:
```
external/rules_rust+/rust/private/rustc.bzl:1090:38: Traceback (most recent call last):
File "external/rules_rust+/rust/private/rustc.bzl", line 2883, column 16, in _get_dirname
    return file.dirname
Error: 'NoneType' value has no field or method 'dirname'
```

This change adds a proper output to satisfy the the current `CrateInfo` interface.

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -951,8 +951,7 @@ def construct_arguments(
         skip_expanding_rustc_env = False,
         require_explicit_unstable_features = False,
         error_format = None,
-        allowed_unstable_rust_features = None,
-        emit_out_dir_flag = True):
+        allowed_unstable_rust_features = None):
     """Builds an Args object containing common rustc flags
 
     Args:
@@ -1024,10 +1023,6 @@ def construct_arguments(
         require_explicit_unstable_features (bool): Whether to require all unstable features to be explicitly opted in to using `-Zallow-features=...`.
         error_format (str, optional): Error format to pass to the `--error-format` command line argument. If set to None, uses the "_error_format" entry in `attr`.
         allowed_unstable_rust_features (list, optional): List of unstable Rust language features allowed for this target.
-        emit_out_dir_flag (bool): Whether to emit `--out-dir=<crate_info.output.dirname>`.
-            Defaults to True (the rustc case). Set to False for rustdoc actions that
-            emit `--output` themselves — rustdoc rejects `--out-dir` and `--output`
-            together with `error: cannot use both 'out-dir' and 'output' at once`.
 
     Returns:
         tuple: A tuple of the following items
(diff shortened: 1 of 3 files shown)


UebelAndre:
@krasimirgg @slackito do either of you know if this would break anything on Google's end?

rust/private/rustc.bzl
```diff
@@ -1022,6 +1023,10 @@ def construct_arguments(
         require_explicit_unstable_features (bool): Whether to require all unstable features to be explicitly opted in to using `-Zallow-features=...`.
         error_format (str, optional): Error format to pass to the `--error-format` command line argument. If set to None, uses the "_error_format" entry in `attr`.
         allowed_unstable_rust_features (list, optional): List of unstable Rust language features allowed for this target.
+        emit_out_dir_flag (bool): Whether to emit `--out-dir=<crate_info.output.dirname>`.
```
krasimirgg: I wonder if instead we could simulate this behavior by checking whether crate_info.output is a directory or a file? If I understand correctly, it's a directory if and only if we're in a rustdoc context, and it's exactly when we want to suppress emitting `--out-dir`?

rust/private/rustc.bzl
```diff
@@ -1036,7 +1041,7 @@ def construct_arguments(
     if build_metadata and not use_json_output:
         fail("build_metadata requires parse_json_output")
 
-    output_dir = getattr(crate_info.output, "dirname", None)
+    output_dir = crate_info.output.dirname
```
krasimirgg, line 1040: I'm not completely sure if internally we always have a non-null output. Let me check...

krasimirgg:
> @krasimirgg @slackito do either of you know if this would break anything on Google's end?

we don't directly use the rustfmt aspect, but I'm not sure if we don't have other extensions that produce a CrateInfo with an empty output. Let me run a check about that...

EDIT: ran a sample test, that bit looks OK. Left a comment about potentially simplifying the logic by sniffing if crate_info.output is a directory or not.

rust/private/rustc.bzl
krasimirgg: ran a sample test, this looks OK.

slackito (approved):
Looks good to me. I'll report back if it breaks anything once we import it.

slackito:
Actually I just raced with Krasimir and it turns out he had comments. Please address them before merging.

rust/private/rustc.bzl
UebelAndre: Done, now if `output` is a directory, we only emit `--out-dir` for that path but if it's a source we emit it for dirname.

PR #4161, opened 2026-07-21
Add `cargo_build_script.use_cc_toolchain`

### LINKED ISSUE #3680: Incompatible flag to stop merging the CC toolchain's all_files into rust toolchain's all_files
the CC toolchain's all_files can be quite big when using a hermetic sysroot, which creates large input trees for actions using this depset, including `cargo_build_script`. Note that `cargo_build_script` also explicitly adds the cc toolchain's all_files as well, but https://github.com/bazelbuild/rules_rust/pull/3679 provides a way to avoid that when the rule author knows the cc toolchain isn't required. Other actions should likely make their own decisions whether they want to pull in the CC sysroot or not, and if so, do it explicitly.

UebelAndre:
This change introduces `use_cc_toolchain` to `cargo_build_script` which can be used to control whether or not a `cc_toolchain` is explicitly withheld from `CargoBuildScript` actions. Additionally a global flag `--@rules_rust//cargo/settings:use_cc_toolchain` can be used to control the default of any build script that has not either explicitly set `use_cc_toolchain = 1` (always include cc toolchains) or `use_cc_toolchain = 0` (always exclude cc toolchains).

closes https://github.com/bazelbuild/rules_rust/issues/3680
closes https://github.com/bazelbuild/rules_rust/pull/3679


krasimirgg (approved):
Thank you!

PR #4164, opened 2026-07-22
Add per-platform `crate_universe` feature and dependency exclusions

### LINKED ISSUE #4163: Cargo features are enabled for all target triples
**Description**

`crate_universe` resolves Cargo features once across the whole workspace and emits a single shared target per crate, used by every target triple. So a feature enabled by _any_ workspace member is enabled for that crate on _all_ triples, and there is no way to drop a feature (or the dependencies it brings in) for a specific triple. Cargo itself resolves features per target and wouldn't do this; `crate_universe` collapses that distinction.

This breaks cross-platform workspaces where a feature is fine on one triple but not another. Concretely: one member (built for a host triple) enables the `net` feature in Tokio. The shared `tokio` target then has `net` on for `wasm32-unknown-unknown` too, pulling in `mio` and `socket2`. They don't build there — even though the Wasm member only uses `rt`/`macros`/`time` from Tokio. The Wasm build fails with

> Only features sync,macros,io-util,rt,time are supported on wasm.

**Reproduction steps**

Two Rust crates that depend on Tokio with different Cargo features. First crate:

```toml
[package]
name = "native-app"
version = "0.0.0"
edition = "2021"

[dependencies]
tokio = { version = "1", features = ["net"] }
```

Second crate:

```toml
[package]
name = "wasm-app"
version = "0.0.0"
edition = "2021"

[dependencies]
tokio = { version = "1", features = ["rt", "macros", "time"] }
```

Bazel workspace setup:

```python
crate.from_cargo(
    name = "crate_index",
    manifests = ["//:Cargo.toml", "//native-app:Cargo.toml", "//wasm-app:Cargo.toml"],
    supported_platform_triples = [
        "x86_64-unknown-linux-gnu",
        "wasm32-unknown-unknown",
    ],
)
```

Build file:

```python
rust_library(
    name = "wasm-app",
    srcs = glob(["src/**/*.rs"]),
    edition = "2021",
    target_compatible_with = ["@platforms//cpu:wasm32"],
    deps = all_crate_deps(normal = True),  # resolves to @crate_index//:tokio
)
```

Building the web app with `bazel build --platforms=@rules_rust//rust/platform:wasm32 //wasm-app:wasm-app` now because we try to compile `mio` and `socket2` for `wasm32`, which they don't support.

**Additional context**

I had Claude suggest a fix for this, coming up in a PR next.

**Impact**

I believe this makes it impossible to build our Rust business logic for Wasm.

**Bazel and rules_rust version**

- Bazel: 9.1.1
- rules_rust: 0.71.3

mgeisler:
Written by Claude, I can confirm that this works my company monorepo where we compile a Rust core for both Wasm, Android, and iOS.

I am happy to discuss other approaches: this was one suggested by the LLM. I hope the PR here serves to demonstrate the problem well and hopefully we can find a good and flexible solution. One thing I noticed while applying the fix to our codebase is that it required a lot of trial and error before Claude found the right crates and features to disable. A solution that simplifies this would be great.

---

`crate_universe` resolves crate features once over the whole Cargo workspace, so Cargo's workspace-wide feature unification activates a feature requested by any member on the single shared target generated for that crate, for every supported platform triple.

This forces features onto platforms that cannot build them: an unconditional `tokio = { features = ["net"] }` anywhere in the workspace pulls in the `net` feature for Tokio (and thus the transitive `mio` and `socket2` dependencies). This happens even on `wasm32-unknown-unknown`, where they do not compile.

This MR adds new `disabled_features` and `excluded_deps` attributes to `crate.annotation` and `crate.annotation_select`. They allow a user to trim the resolved feature and dependency sets of a crate. The removal is scoped to a triple and it demotes a formerly unconditional value onto the other supported triples. This way, only the named platform loses it.

Because the trimming operates on the shared target's own resolved attributes, it fixes the crate for every consumer at once, regardless of which workspace member (or dev-dependency) unioned the feature in.

New unit tests have been added, as well as an integration test that demonstrates the issue for Tokio compiled both with and without the `net` feature.

Fixes #4163.


mgeisler:
It seems the number of integration tests is fixed at max 128? This PR adds a new one – I'm curious to hear how I should approach this. I'll let it sit for a while until someone can tell me more about the conventions here.

PR #4168, opened 2026-07-26
Propagate wasm runfiles

UebelAndre:
This change allows the wasm targets to propagate runfiles with the bindgen outputs.


slackito (approved):
Do we need a test for this?

UebelAndre:
> Do we need a test for this?

Yeah, this was actually part of a larger change which I did have testing for but don't feel confident about the design so didn't include it.

UebelAndre:
https://github.com/bazelbuild/rules_rust/issues/4172 is the larger change I was working on. I would love some ideas there.

PR #4169, opened 2026-07-26
Avoid force excluding `bazel-*` in vscode

UebelAndre:
We shouldn't blanket exclude `bazel-*` so users can continue to inspect these directories in the file browser.


tools/rust_analyzer/bin/setup.rs
```diff
@@ -1382,8 +1376,8 @@ mod tests {
         let without_srv = vscode_managed_keys(&ctx);
         // 5 rust-analyzer keys (discover, server, proc-macro, rustfmt,
         // check.overrideCommand) + 3 exclude maps = 8 total.
```
slackito: I think the comment needs to be updated, as it's no longer 8.

slackito (approved):

tools/rust_analyzer/bin/setup.rs
UebelAndre:
```suggestion
// 5 rust-analyzer keys (discover, server, proc-macro, rustfmt,
// check.overrideCommand) + 2 exclude maps = 7 total.
```

UebelAndre: https://github.com/bazelbuild/rules_rust/pull/4169#discussion_r3658759867

PR #4176, opened 2026-07-28
Unconditionally set static_library even when use_pic is True

jakos-sec:
`rules_go` and some other C++ interoperability layers expect static_library to unconditionally be present. When they attempt to link the binary, the Rust archives (libcxx.a, libtokenizers_cpp.a, etc.) were missing from the linker command line, triggering the `ld.lld: error: undefined symbol` build failure, we therefore set those unconditionally.


krasimirgg (approved):
Thank you! Chatted offline, and let's see if we can set up an integration test for this using rules_go (seems we need to bump the version of rules_go used by tests first; we can do that independently).

jakos-sec:
I can currently no longer reproduce the issues that require static_library to be always set. Let's close this PR and create a new one if the issue is reproducable.

PR #4177, opened 2026-07-29
use crate_info.owner for linker inputs, if available

krasimirgg:
Fixes an inconsistency where we use crate_info.owner as the owner of Rust actions, while we used ctx.label as the owner of the corresponding C++ linker inputs. Most of the time, these are the same, but we've got a use case in rust proto integration where we wanna be able to customize the owner of the C++ infos (so that these can be distinguished from the C++ infos of the standard c++ proto bindings).


rust/private/rustc.bzl
```diff
@@ -2360,7 +2360,7 @@ def establish_cc_info(ctx, attr, crate_info, toolchain, cc_toolchain, feature_co
         fail("Unexpected case")
 
     link_input = cc_common.create_linker_input(
-        owner = ctx.label,
+        owner = getattr(crate_info, "owner", None) or ctx.label,
```
UebelAndre: Do we need to use `getattr` here? We own `CrateInfo` and it has an `owner` memeber, right?

UebelAndre (approved):
Thanks!

rust/private/rustc.bzl
krasimirgg: (n.b. -- updated and also updated callsites that didn't always set this field, e.g., rustdoc aspect that was cloning the underlying crate info without copying the owner)

slackito (approved):

PR #4179, opened 2026-07-29
Add support for Rust ABI dynamic libraries

buntonj:
Currently,  there is not explicit stable Rust ABI for dynamic libraries, but Rust library may still be compiled as one.  This can be useful if a large project wants to compile some core portion of its code as a shared object that can be dynamically linked against several different binaries.  Bevy, for example, offers this capability as a way to allow customers of Bevy to tweak their own code without having to statically re-link the central Bevy engine code.

This PR exposes the Rust ABI dylib capabilities with a new rule, `rust_dylib_library`.  For clarity, it also creates a `rust_cylib_library` rule, and updates methods to this name, with `rust_shared_library` just aliasing to the `rust_cdylib_library` implementation.

Building a Rust ABI dylib may cause the standard library to be linked dynamically against the generated dylib. When depended on, the dynamic standard library needs to be included to get the binary to execute.  Currently, `rules_rust` offers this as a toolchain-level flag, but applying a toolchain transition at the dylib's interface imposes a new toolchain on its deps, which then forces any upstream `rlibs`, for example, to recompile.

To fix this, this PR exposes a `link_std_dylib` attr to `rust_binary`, `rust_dylib_library` and `rust_test` to allow these targets to optionally bundle themselves with the dynamically-linked stdlib.

closes https://github.com/bazelbuild/rules_rust/pull/4179


rust/defs.bzl
UebelAndre, line 1: Can you also add `rust_dylib_library.bzl` and `rust_cdylib_library.bzl` files neighboring this one with those exports?

rust/private/rustc.bzl
```diff
@@ -1794,6 +1800,11 @@ def rustc_compile_action(
         dwo_outputs = ctx.actions.declare_directory(fission_directory, sibling = crate_info.output)
         rust_flags.append(("-Zsplit-dwarf-out-dir=%s", dwo_outputs))
 
+    if hasattr(ctx.attr, "link_std_dylib"):
+        link_std_dylib = toolchain._experimental_link_std_dylib or ctx.attr.link_std_dylib
```
UebelAndre: Maybe rename these attributes on the toolchain as well?

UebelAndre (changes requested):
I'd love @krasimirgg to take a look since I think there's some google uses of `experimental_link_std_dylib`. The change makes sense to me but I wonder if `link_std_dylib` should have more impact outside of just the allocators and codegen flag. The attribute sounds like it would directly link the dylibs but I don't think there's any wiring for that. Maybe @krasimirgg can correct me if I'm wrong

rust/private/rustc.bzl
buntonj: Yeah, totally, done!

rust/defs.bzl
buntonj: Definitely!

krasimirgg:
Hey, I'm not an expert on dynamic linking, so please bear with me...
I don't see how this composes with the rest of the rust ruleset. Two high-level questions:
1. What's the expected way for clients (rust_binary) depending on a rust_dylib_library to work? Are folks running them expected to supply the .so-s manually? Would bazel run do some work to ensure the dylibs are correctly provided?
2. What happens if we have a mix of rust_dylib_library and rust_library dependencies? How do we avoid duplicating shared dependencies in dylibs, rendering them incompatible? For instance, if you have:
For example, if I read the idea correctly, if you have:
```
rust_library(name = "d")
rust_dylib_library(name = "c", deps = ["d"])
rust_dylib_library(name = "b", deps = ["d"])
rust_library(name = "a", deps = ["c", "b"])
```
, you'll likely get rustc error: cannot satisfy dependencies so d only shows up once while compiling a. Example rustc commands I'm thinking about -- https://gist.github.com/krasimirgg/66c52fbebee7af4813e1e71af495bd4a.

Tangent 1: On a high-level, I see a way to set up rust dynamic libraries support in a way similar to how rules_cc does dynamic libraries -- the rust_library itself has actions that produce both the rlib and the dylib, and clients pick up the appropriate dependency based on the --dynamic_mode build setting. This is composable, in particular any transitive dependencies of a dylib are automatically dylibs, avoiding the case where duplicated common dependencies render intermediate dylibs incompatible. But in a world like this, having a separate rust_dylib_library rule becomes just something almost equivalent to a rust_library.

Tangent 2: The reason why I feel this is different than rust_shared_library (cdylib) is that that one is designed to act like a final artifact -- you collect all of your rust into a single such target, which acts more like a final binary; standard rust libraries cannot direclty depend on it and you just hook it up as a special dependency of the final binary.

UebelAndre:
> 1. What's the expected way for clients (rust_binary) depending on a rust_dylib_library to work? Are folks running them expected to supply the .so-s manually? Would bazel run do some work to ensure the dylibs are correctly provided?

I would expect Bazel to bundle the `.so` files as runfiles and have appropriate loader paths to support that, similar to how `rules_cc` does.

> 2. What happens if we have a mix of rust_dylib_library and rust_library dependencies? How do we avoid duplicating shared dependencies in dylibs, rendering them incompatible? For instance, if you have:
>         For example, if I read the idea correctly, if you have:
> 
> 
> ```
> rust_library(name = "d")
> rust_dylib_library(name = "c", deps = ["d"])
> rust_dylib_library(name = "b", deps = ["d"])
> rust_library(name = "a", deps = ["c", "b"])
> ```
> 
> , you'll likely get rustc error: cannot satisfy dependencies so d only shows up once while compiling a. Example rustc commands I'm thinking about -- https://gist.github.com/krasimirgg/66c52fbebee7af4813e1e71af495bd4a.

What makes the `dylib` unique in this case compared to `rlib`? Wouldn't the dylib be a runfile that needs to be bubbled up along with the top level rlib?

> Tangent 1: On a high-level, I see a way to set up rust dynamic libraries support in a way similar to how rules_cc does dynamic libraries -- the rust_library itself has actions that produce both the rlib and the dylib, and clients pick up the appropriate dependency based on the --dynamic_mode build setting. This is composable, in particular any transitive dependencies of a dylib are automatically dylibs, avoiding the case where duplicated common dependencies render intermediate dylibs incompatible. But in a world like this, having a separate rust_dylib_library rule becomes just something almost equivalent to a rust_library.

I'm not sure it's always desirable to turn everything into dynamic libraries globally like this. Lets say you had a `rust_binary` with 3 `rust_library_group` dependencies, would it be reasonable to turn a number of those into `rust_dylib_library` to avoid the more expensive linking of `rust_binary`? That sounds like a nice optimization on paper.


> Tangent 2: The reason why I feel this is different than rust_shared_library (cdylib) is that that one is designed to act like a final artifact -- you collect all of your rust into a single such target, which acts more like a final binary; standard rust libraries cannot direclty depend on it and you just hook it up as a special dependency of the final binary.

I agree `rust_cdylib_library`/`rust_shared_library` can be thought of as an output, akin to `rust_binary`. I think the open question is should a conceptual new `dylib` (not `cdylib`) rule behave the same?

buntonj:
Thanks for thinking about this @krasimirgg ! Some thoughts:
>What's the expected way for clients (rust_binary) depending on a rust_dylib_library to work? Are folks running them expected to supply the .so-s manually? Would bazel run do some work to ensure the dylibs are correctly provided?

Bazel should provide the `.so` files for dylibs that are in `deps`--that's in this PR (they have `CcInfo` that gets collected and bundled with the runfiles).

> What happens if we have a mix of rust_dylib_library and rust_library dependencies? How do we avoid duplicating shared dependencies in dylibs, rendering them incompatible?

I feel like this is a limitation of dynamic linking with `rustc` than something we have to solve in `rules_rust`...which is why for this PR, I opted to expose the tools to manually walk back through the dep graph and convert things to dylibs that must be for their target to compile.

This is definitely a possible footgun though--asking for dynamic linking at some point in the chain _might_ imply other libraries should become dynamically linked.  It's not required if no dependency diamonds like this occur, though.

> Tangent 1: On a high-level, I see a way to set up rust dynamic libraries support in a way similar to how rules_cc does dynamic libraries -- the rust_library itself has actions that produce both the rlib and the dylib, and clients pick up the appropriate dependency based on the --dynamic_mode build setting. This is composable, in particular any transitive dependencies of a dylib are automatically dylibs, avoiding the case where duplicated common dependencies render intermediate dylibs incompatible. But in a world like this, having a separate rust_dylib_library rule becomes just something almost equivalent to a rust_library.

I think this is very cool, and could be something to build toward.  In my use case, I wanted to avoid forcing every transitive dep to be unconditionally built as both an rlib and a dylib, when for many steps in that graph, only the rlib version was required (in my simpler use case, I had a deep dependencies with a dylib at the focal interface point).  Instead, this lets me mark *only the required targets* as dylibs to save compilation time.

UebelAndre:
@krasimirgg friendly ping here

krasimirgg:
Hey sorry, was a bit out of office, replies below.


cc @UebelAndre  @buntonj
>What makes the dylib unique in this case compared to rlib? Wouldn't the dylib be a runfile that needs to be bubbled up along with the top level rlib?
```
rust_library(name = "d")
rust_dylib_library(name = "c", deps = ["d"])
rust_dylib_library(name = "b", deps = ["d"])
rust_library(name = "a", deps = ["c", "b"])
```
Here, both of the dylibs-s `c.so` and `b.so` will include _the definitions_ for symbols from the target `:d` -- there is a linking step in both, so _both will statically link_ in `d.rlib` (because there is no `d.so` available). Then later, when you're trying to build `a.so`, you run into duplicate symbol errors -- there are two copies of the symbols from `d.rlib` -- one embedded in `c.so` and another in `b.so`.
If `:c` and `:b` were `rust_library`-es, their `.rlib`-s only contain _references_ to the symbols from `:d`, not definitions. That avoids the potential for duplicate symbols. That's what I mean by saying `rust_dylib_library` does not compose with the rest of the rulesets -- the end-user needs to ensure something like: **for every target A, for every transitive `rust_library` dependency D of A, if there exist a dependency chain from A to D that passes trough an intermediate `rust_dylib_library` L, then ALL possible paths from A to D must pass through L (because D will be statically linked into L.so).**

>What happens if we have a mix of rust_dylib_library and rust_library dependencies? How do we avoid duplicating shared dependencies in dylibs, rendering them incompatible?
>I feel like this is a limitation of dynamic linking with rustc than something we have to solve in rules_rust...which is why for this PR, I opted to expose the tools to manually walk back through the dep graph and convert things to dylibs that must be for their target to compile.

I think the introduction of this rule is OK as long as we put some **big disclaimers around these limitations and document this rule so whenever some user tries using it and linking fails, we can point to that**.

About the tangent, just informationally:

>I wanted to avoid forcing every transitive dep to be unconditionally built as both an rlib and a dylib, when for many steps in that graph, only the rlib version was required...

We've started internally poking around adding full `--dynamic_mode` support (so a `cc_test` that (transitively) depends on a `rust_library` gets it as a pair of `.so/.ifso`-s similarly to `cc_library` dependencies). We're exploring some approaches around rustc building both an rlib and dylib, vs., having cc_common.link produce the .so from the .rlib (and the .ifso from the .so), etc. I'm hopeful that when we have something satisfactory, we upstream it.

buntonj:
Hey, no problem--thanks again for the time!
> I think the introduction of this rule is OK as long as we put some big disclaimers around these limitations and document this rule so whenever some user tries using it and linking fails, we can point to that.

I'm 100% with you here. I added some language in the docs about this, let me know if it needs revising though.

I made a few other improvements to help support this PR:
* Added integration tests that have a `rust_binary` dep on a `rust_dylib_library`
* Fixed some issues with `RPATH` computation for different execution locations (ran into these when road-testing this PR on my use case). I took a page out of `rules_cc`'s book and opted to supply _all_ of the relevant `RPATH`s when linking a dylib ([see here for my shameless rip](https://github.com/bazelbuild/rules_cc/blob/d27efa119e41d74243da34da24b86281811b6f7d/cc/private/link/collect_solib_dirs.bzl#L288)).
* Added support for Windows, where `RPATH` isn't a concept--again took a page out of `rules_cc` and opted to symlink the `dylibs` adjacent to the binary so it's found on Windows.  Added a test for this.

Also for your tangent:
> We've started internally poking around adding full --dynamic_mode support (so a cc_test that (transitively) depends on a rust_library gets it as a pair of .so/.ifso-s similarly to cc_library dependencies). We're exploring some approaches around rustc building both an rlib and dylib, vs., having cc_common.link produce the .so from the .rlib (and the .ifso from the .so), etc. I'm hopeful that when we have something satisfactory, we upstream it.

This sounds awesome. If I can support at all in the future I'd be stoked to!

buntonj:
cc @krasimirgg

krasimirgg:
Hey @buntonj , we chatted a bit with @UebelAndre last week about this, who would also follow-up on this review. The whole thing is because of the above concerns so we should tread carefully with the whole feature.

rust/private/rust.bzl
```diff
@@ -987,6 +1008,15 @@ _RUST_TEST_ATTRS = {
     "env_inherit": attr.string_list(
         doc = "Specifies additional environment variables to inherit from the external environment when the test is executed by bazel test.",
     ),
+    "link_std_dylib": attr.bool(
+        mandatory = False,
+        default = False,
+        doc = dedent("""\
+            Flag to dynamically link the standard library as a Rust dylib .so object when building this test.
```
krasimirgg, line 1015: don't we also need this for rust_binary?

rust/private/rust.bzl
```diff
@@ -1082,6 +1112,53 @@ rust_library = rule(
...
+    attrs = _COMMON_ATTRS | {
+        "disable_pipelining": attr.bool(
+            default = False,
+            doc = dedent("""\
+                Disables pipelining for this rule if it is globally enabled.
+                This will cause this rule to not produce a `.rmeta` file and all the dependent
+                crates will instead use the `.rlib` file.
+            """),
+        ),
+        "link_std_dylib": attr.bool(
```
krasimirgg, line 1127: Do we have use cases at all for `rust_dylib_library` where we don't wanna link the stdlibs dynamically? Naively, for this rule the default should be true?

rust/private/toolchain.bzl
```diff
@@ -720,10 +721,6 @@ rust_toolchain = rule(
             ),
             mandatory = True,
         ),
-        "experimental_link_std_dylib": attr.label(
```
krasimirgg, line 724: I think we're losing something with this change -- previously, it was possible to have 2 rust_toolchain-s where one has this enabled and the other doesn't. With the new version, all possible toolchains use same default. I think we should switch to a pattern where _link_std_dylib points to a label as in the new version, but there is an optional per-toolchain override  with values [-1, 0, 1] as in experimental_use_allocator_libraries_with_mangled_symbols.

test/rust_dylib/BUILD
```diff
@@ -0,0 +1,37 @@
+load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_dylib_library", "rust_test")
+
+# Integration tests for rust_dylib_library: verifies that a binary depending
+# on a Rust ABI dylib can be built and run, exercising runfiles inclusion
+# and RPATH resolution end-to-end.
```
krasimirgg, line 5: (just informational): I personally am not familiar with the `RPATH` feature and ways it's working in bazel. I'm aiming to read up on this, but currently I can't vouch for those parts.

rust/private/rust.bzl
buntonj: Yep, [it's there](https://github.com/bazelbuild/rules_rust/pull/4179/changes#diff-79c1a0acbeb93f59ebb8d1aad79820043dbb7d8905f78f3736b7e49875b19c83R1306)!

rust/private/rust.bzl
buntonj: Yeah, that's totally fair. AFAICT, Rust ABI dylibs will just be depended on by Rust targets, which would require `link_std_dylib` anywho, so setting the default to true is harmless.  Done!

rust/private/toolchain.bzl
buntonj: Ah, great catch and suggestion. Migrated to using the same `[-1, 0, 1]` pattern, which leaves us with the ability to toolchain-level or target-level change this flag.

rust/private/rust_allocator_libraries.bzl
```diff
@@ -183,7 +186,7 @@ def make_libstd_and_allocator_ccinfo(
             order = "topological",
         )
 
-        if experimental_link_std_dylib:
+        if link_std_dylib:
```
UebelAndre, line 189: Is this flag no longer experimental now?

rust/private/rustc.bzl
```diff
@@ -2070,16 +2097,30 @@ def rustc_compile_action(
...
             for library_to_link in linker_input.libraries
             if _is_dylib(library_to_link) and library_to_link.dynamic_library
-        ])
-        transitive_runfiles.append(dynamic_libraries)
+        ]
+        transitive_runfiles.append(ctx.runfiles(files = dep_dylib_files))
+
+    # On Windows there is no RPATH equivalent. Create symlinks of dylib files
+    # next to the binary so the Windows loader can find them.
+    if toolchain.target_os == "windows" and (crate_info.type == "bin" or crate_info.is_test):
```
UebelAndre, line 2126: I'm concerned this will lead to conflicting action outputs. Do you know what `rules_cc` does here?

UebelAndre (changes requested):
Looking good so far! Can you also make sure we add a section for `rust_dylib_library` and `rust_cdylib_library` to docs with this change?

rust/private/rustc.bzl
buntonj: I actually borrowed this pattern from rules_cc's `_create_dynamic_libraries_copy_actions`. See [here](https://github.com/bazelbuild/rules_cc/blob/aff544134bc9d29006f27adbe63d824dfe1436f6/cc/private/rules_impl/cc_binary_impl.bzl#L736) and [here](https://github.com/bazelbuild/rules_cc/blob/aff544134bc9d29006f27adbe63d824dfe1436f6/cc/private/rules_impl/cc_binary_impl.bzl#L184).

Bazel [handles the deduplication of targets](https://github.com/bazelbuild/bazel/blob/3a9b19c85b11c8a7eaa07013e052bb92d4dc8132/src/main/java/com/google/devtools/build/lib/actions/Actions.java#L76-L85) in the relevant cases also.

rust/private/rust_allocator_libraries.bzl
buntonj: Ah, less that it's not experimental and more that it's now a computed quantity/property as a function of the toolchain-level `experimental_link_std_dylib` and the target-level `link_std_dylib`.

I opted to try and rename things so that the `experimental_link_std_dylib` hook and name are the same, but all internal customers/byproducts of the new `link_std_dylib` are not `experimental_` anymore.

rust/private/rustc.bzl
buntonj: (I added a comment in this spot to this effect so we remember! For those who come after...)

rust/private/rustc.bzl
```diff
@@ -2067,7 +2090,11 @@ def rustc_compile_action(
 
     runfiles = ctx.runfiles(
         files = getattr(ctx.files, "data", []) +
-                ([] if experimental_use_coverage_metadata_files else coverage_runfiles),
+                ([] if experimental_use_coverage_metadata_files else coverage_runfiles) +
+                # Include any generated Rust ABI dylibs as required runfiles.
+                ([crate_info.output] if crate_info.type == "dylib" else []) +
+                # Include the stdlib dylib when dynamically linking the standard library.
+                ([f for f in toolchain.rust_std.to_list() if is_std_dylib(f)] if link_std_dylib else []),
```
UebelAndre: Can we do this filtering at the toolchain level so we can avoid resolving this depset and filtering it for each target?

UebelAndre (changes requested):
This should be my final request!

rust/private/rustc.bzl
buntonj: Yeah, definitely--I added `rust_std_dylib` to `ToolchainInfo` so that we can assemble the files there if present, then directly read from it later instead of rescanning.

rust/private/toolchain.bzl
```diff
@@ -914,6 +917,21 @@ rust_toolchain = rule(
             default = Label("//rust/settings:incompatible_do_not_include_transitive_data_in_compile_inputs"),
             doc = "Label to a boolean build setting that controls whether to include transitive data dependencies in compile inputs.",
         ),
+        "link_std_dylib": attr.int(
+            doc = (
+                "Whether to link libstd dynamically. Possible values: [-1, 0, 1]. " +
+                "-1 means to use the value of the build setting " +
+                "//rust/settings:experimental_link_std_dylib. " +
```
UebelAndre:
```suggestion
                "`@rules_rust//rust/settings:experimental_link_std_dylib`. " +
```

UebelAndre (approved):
Thank you!

PR #4181, opened 2026-07-30
Add JUnit XML output to rust_test

### LINKED ISSUE #1303: Supporting junit XML report in rust_test
Currently `rust_test` does not support generating junit xml report out-of-the-box.

To support this for all rust test target, a workaround is with the `--run_under` flag and a script that writes the xml report from stdout to $XML_OUTPUT_FILE:

```
#!/bin/bash
$@ -Zunstable-options --format=junit --report-time | tee $XML_OUTPUT_FILE
```

The main downside is having to declare the script as a `data` dependency for every rust test target. 

It would be nice to have built-in support for this in rust_test as an option, would it be sensible to add this to the rust test launcher?

jlnhws:
## What this does

`rust_test` targets don't currently write `$XML_OUTPUT_FILE`, so Bazel falls back to synthesising a single generic `<testcase>` for the whole test binary.
Every Rust test then shows up under one name in anything that reads the XML (CI test reporting, test UIs, flaky-test tracking) which stops being useful the moment a suite has more than a couple of tests.

This wraps the test binary in a small runner that runs it, passes its output through untouched, and turns libtest's textual results into a real per-test JUnit report. When `$XML_OUTPUT_FILE` isn't set (e.g. under `bazel run`) the
runner just execs the test binary, so there's no cost outside `bazel test`.

The wrapper is on by default and can be turned off per target with `junit = False`. The one target that needs that here is the runfiles library's own tests: they inspect the process's ambient runfiles (the manifest layout and the repo mapping `rlocation!` resolves against) and the wrapper runs the test as a child under the runner's runfiles, a different environment than the one under test, so they run unwrapped.

## Relationship to #4084

This carries forward @peakschris's work in #4084. On top of that it adds the pieces that were needed to make it land cleanly:

- **A coverage fix that unblocks CI.** With the wrapper on, `RUST_TEST_BIN` points at the real test binary through a runfiles path. `collect_coverage` resolved that against `RUNFILES_DIR`, but under `--experimental_split_coverage_postprocessing` Bazel doesn't set `RUNFILES_DIR`, and the fallback rebuilt a path that inserted a workspace segment that isn't in the output tree, so `llvm-cov` couldn't find the binary and every coverage run of a wrapped test failed. The rule now hands `collect_coverage` the binary's execroot-relative path directly. This is what the coverage jobs on #4084 were tripping over.

- **Real failure reporting.** Instead of a bare `<failure/>`, the runner pulls the panic/assertion line out of libtest's failure block and uses it as the failure message, and attaches captured stdout/stderr to the testcase so the detail survives into the XML.

- **Crash handling.** If a binary exits without printing a `test result:` summary (e.g. due to a segfault or an abort partway through the run) that's surfaced as an error on the suite instead of being mistaken for a pass.

## Testing

- `bazel test //util/junit_runner:all //test/junit:all`: runner unit tests plus an end-to-end suite covering passing, ignored, and failing tests.
- `bazel coverage //test/generated_inputs:all` in both normal and `--experimental_split_coverage_postprocessing` modes, validated the same way presubmit does (real `SF:`/
`LF:` records for the mixed-source crate).

closes https://github.com/bazelbuild/rules_rust/pull/4084
closes https://github.com/bazelbuild/rules_rust/issues/1303

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -598,6 +598,32 @@ def _rust_test_impl(ctx):
         env["CC_CODE_COVERAGE_SCRIPT"] = ctx.executable._collect_cc_coverage.path
     components = "{}/{}".format(ctx.label.workspace_root, ctx.label.package).split("/")
     env["CARGO_MANIFEST_DIR"] = "/".join([c for c in components if c])
+
+    if ctx.attr.junit:
+        test_bin_short = output.short_path
+        if test_bin_short.startswith("../"):
+            rust_test_bin_rloc = test_bin_short[len("../"):]
+        else:
+            rust_test_bin_rloc = ctx.workspace_name + "/" + test_bin_short
+        env["RUST_TEST_BIN"] = rust_test_bin_rloc
+
+        junit_runner = ctx.actions.declare_file(ctx.label.name + "_junit_runner" + toolchain.binary_ext)
+        ctx.actions.symlink(
+            output = junit_runner,
+            target_file = ctx.executable._junit_runner,
+            is_executable = True,
+        )
+
+        original_default_info = providers[0]
+        runner_runfiles = ctx.attr._junit_runner[DefaultInfo].default_runfiles
+        test_bin_runfiles = ctx.runfiles(files = [output])
+        merged_runfiles = original_default_info.default_runfiles.merge(runner_runfiles).merge(test_bin_runfiles)
+        providers[0] = DefaultInfo(
+            files = original_default_info.files,
+            runfiles = merged_runfiles,
+            executable = junit_runner,
+        )
+
     providers.append(RunEnvironmentInfo(
         environment = env,
         inherited_environment = ctx.attr.env_inherit,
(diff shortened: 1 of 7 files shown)


peakschris:
Excellent, thank you so much for moving this forwards :-)

UebelAndre:
Sorry for the delay here! I'm catching up on `rules_rust` reviews. I would be happy to accept a change where a test wrapper is gated behind an experimental flag but we once-upon-a-time had a test wrapper and moved away from it. I'd be much more interested in a solution that doesn't require a wrapper but haven't come up with one and understand this is a very valuable feature! Relegating this to being purely opt-in would be much easier to approve.

jlnhws:
@UebelAndre thanks for your feedback!

> Relegating this to being purely opt-in would be much easier to approve.

Addressed in [eb69e42](https://github.com/bazelbuild/rules_rust/pull/4181/commits/eb69e42e25409fd6852b8b25257835dbaf57e41c).

PR #4182, opened 2026-07-30
Stabilizes per_crate_rustc_flag

### LINKED ISSUE #1711: Feature request: Add something like per_file_copt to allow customizing rustc flags per module

csmulhern:
Fixes #1711.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -73,8 +73,8 @@ alias(
 )
 
 alias(
-    name = "experimental_per_crate_rustc_flag",
-    actual = "//rust/settings:experimental_per_crate_rustc_flag",
+    name = "per_crate_rustc_flag",
+    actual = "//rust/settings:per_crate_rustc_flag",
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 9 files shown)


UebelAndre (approved):
Looks good to me, thank you!

PR #4183, opened 2026-07-30
fix(crate_universe): pass OUTPUT_BASE as --output_base startup flag to bazel info

georgesfarah:
## Problem

The `OUTPUT_BASE` env var was added to let CI environments supply the output_base without relying on `bazel info`. However, the override is applied **after** parsing the `bazel info` output:

```rust
let output = process::Command::new(bazel).arg("info")...output()?;
if !output.status.success() {
    bail!(output.status)    // <-- bails here when output_base is not writable
}
// ... parse output ...
if let Ok(path) = env::var("OUTPUT_BASE") {  // <-- never reached
    bazel_info.insert(...);
}
```

When `bazel info` fails (e.g. the default `~/.cache/bazel/...` is not writable in a sandboxed CI), the function bails before `OUTPUT_BASE` is ever checked — making the env var useless for its intended purpose.

## Fix

Pass `OUTPUT_BASE` as a `--output_base` Bazel **startup flag** (before the subcommand):

```
bazel --output_base=/path/to/base info release output_base
```

Startup flags take effect before Bazel accesses any path, so it never touches the default (possibly non-writable) output_base. The `bazel info` subprocess runs normally and the post-parsing override is no longer needed.

## Changes

- `crate_universe/src/cli/vendor.rs`: restructure `BazelInfo::try_new` to optionally inject `--output_base` as a startup flag; remove the post-parsing `OUTPUT_BASE` override.
- Removed `test_parse_bazel_info_output_base_env_override` since `parse_bazel_info` no longer handles the `OUTPUT_BASE` override — the existing `test_bazel_info` already covers the parsing logic.

### DIFF
--- a/crate_universe/src/cli/vendor.rs
+++ b/crate_universe/src/cli/vendor.rs
@@ -155,8 +155,14 @@ struct BazelInfo {
 impl BazelInfo {
     /// Construct a new struct based on the current binary and workspace paths provided.
     fn try_new(bazel: &Path, workspace_dir: &Path) -> anyhow::Result<Self> {
-        let output = process::Command::new(bazel)
-            .current_dir(workspace_dir)
+        let mut cmd = process::Command::new(bazel);
+        cmd.current_dir(workspace_dir);
+        // Pass --output_base as a Bazel startup flag so it never touches the
+        // default (possibly non-writable) output_base path.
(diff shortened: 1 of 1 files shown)


UebelAndre (approved):
Good catch, thank you!

PR #4184, opened 2026-07-31
Support rustc incremental compilation via persistent workers

anguslees:
## Summary

- add protobuf persistent-worker support directly to `process_wrapper`
- add an opt-in `--@rules_rust//rust/settings:experimental_persistent_worker=true` setting
- reuse request-namespaced rustc incremental state for non-optimized builds
- keep process-wrapper flags in worker startup arguments and materialize each flat `WorkRequest` as one rustc param file
- preserve nested rustc response files and process-wrapper substitutions with an internal response-file marker
- preserve one-shot execution for remote/non-worker executors and the rare opaque `Args` extension case

Related to #412. This is a worker-local alternative to the Bazel-managed incremental artifacts proposed in #3925.

## Test plan

- `bazel test //util/process_wrapper:process_wrapper_test --@rules_rust//rust/settings:experimental_persistent_worker=true`
- `bazel test //test/process_wrapper/... //test/unit/pipelined_compilation:pipelined_compilation_test_suite //test/unit/location_expansion:location_expansion_test_suite`
- worker and forced one-shot builds of `//test/unit/location_expansion:mylibrary` with nested response files
- worker build with a response file supplied through `extra_rustc_flag`
- `bazel test --config=rustfmt //util/process_wrapper:process_wrapper_test`
- `buildifier -mode=check rust/private/rust.bzl rust/private/rustc.bzl rust/settings/BUILD.bazel rust/settings/settings.bzl`

## AI assistance

Kilo Code assisted with rebasing the original implementation onto current `main`, compatibility review, and test iteration. I reviewed and own the resulting changes.

### DIFF
--- a/rust/private/rust.bzl
+++ b/rust/private/rust.bzl
@@ -690,6 +690,9 @@ RUSTC_ATTRS = {
     "_extra_rustc_flags": attr.label(
         default = Label("//rust/settings:extra_rustc_flags"),
     ),
+    "_experimental_persistent_worker": attr.label(
+        default = Label("//rust/settings:experimental_persistent_worker"),
+    ),
     "_per_crate_rustc_flag": attr.label(
         default = Label("//rust/settings:per_crate_rustc_flag"),
     ),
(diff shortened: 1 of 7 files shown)


anguslees:
If you would rather I switch to using a protobuf library (and add more deps to the process wrapper), I can do that instead.

PR #4185, opened 2026-07-31
Set `toolchain` parameter to `toolchain_type` instead of `None`

FabianWolff:
I don't understand the exact mechanics of this, but it appears to be necessary to support multiple different exec configurations.


UebelAndre (approved):
Thank you!

PR #4092, opened 2026-06-18
rustc: add support for split debuginfo

slackito:
This is #3168 by @durin42, with fixes for some of the pre-merge checks.

This PR adapts the original change to support path mapping when passing `-Zsplit-dwarf-out-dir=<path>`. In order to do this, it adds to `construct_arguments` the option to pass `(format_string, File)` as elements in the list form of `rustc_flags`. This PR also adds tests for the new functionality.

The new changes on top of the original PR were written with assistance by Gemini.

PR #4186, opened 2026-07-31
Check for nightly toolchain when enabling split debug info.

slackito:
This fixes a problem pointed out by
https://github.com/bazelbuild/rules_rust/pull/4092#issuecomment-5136906807.

`-Zsplit-dwarf-out-dir` is only available on nightly, so we have to gate split debug info on it. Now, if the user requests `--fission` while using a non-nightly Rust toolchain, the build will fail. To avoid breaking users relying on the previous behavior, a `skip_fission_for_rust` flag has been added. Users can add it to their bazelrc file to explicitly signal that they're okay with Rust compiles not producing split debug info, even if requested in the command line.

Assisted-by: Gemini

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -1778,13 +1778,23 @@ def rustc_compile_action(
         elif ctx.attr.require_explicit_unstable_features == -1:
             require_explicit_unstable_features = toolchain.require_explicit_unstable_features
 
-    use_split_debuginfo = (
+    use_split_debuginfo = False
+    if (
         feature_configuration and
         cc_common.is_enabled(feature_configuration = feature_configuration, feature_name = "per_object_debug_info") and
-        ctx.fragments.cpp.fission_active_for_current_compilation_mode() and
+        ctx.fragments.cpp.fission_active_for_current_compilation_mode()
+    ):
         # `-Zsplit-dwarf-out-dir` is only available on nightly.
-        toolchain.channel == "nightly"
-    )
+        if toolchain.channel == "nightly":
+            use_split_debuginfo = True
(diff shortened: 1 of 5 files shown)


UebelAndre (approved):
Looks good to me!

slackito:
I chatted with @krasimirgg and @UebelAndre and there were some concerns that silently failing to enable fission on Rust when using a stable toolchain would cause a bad experience for people trying to debug their Rust+fission builds.

So I've changed it so that `--fission=true` with a stable Rust toolchain will fail, and added a flag to preserve the existing behavior (that is, C++ compiles will output split debug info but Rust compiles won't) . This way, users that rely on the current behavior can add the flag to their bazelrc and explicitly opt-out from Fission on rust even if it's requested in the command line.

rust/private/rustc.bzl
```diff
@@ -1778,11 +1778,23 @@ def rustc_compile_action(
...
+    if (
         feature_configuration and
         cc_common.is_enabled(feature_configuration = feature_configuration, feature_name = "per_object_debug_info") and
         ctx.fragments.cpp.fission_active_for_current_compilation_mode()
-    )
+    ):
+        # `-Zsplit-dwarf-out-dir` is only available on nightly.
+        if toolchain.channel == "nightly":
+            use_split_debuginfo = True
+        elif toolchain._skip_fission_for_rust:
```
krasimirgg: As a user, I'd expect that if _skip_fission_for_rust is true and toolchain.channel == "nightly", fission-for-rust is disabled... In other words, the setting takes precedence. What do you think?

krasimirgg (approved):
Looks good to me!

PR #4189, opened 2026-07-31
Feat: pass additional args down to rustfmt

### LINKED ISSUE #4188: Pass arbitrary args (or at least `--check`) from `//tools/src:target_aware_rustfmt` down to the underlying rustfmt binary
**Is your feature request related to a problem? Please describe.**

Can't run `bazel run @rules_rust//:rustfmt -- --check //rust/...`.

**Describe the solution you'd like**

The [rustfmt wrapper only supports a small set of args](https://github.com/bazelbuild/rules_rust/blob/d7c7117399990e73fd12964d7ea2c78967aa210e/tools/rustfmt/src/main.rs#L200-L215), not the full set that `cargo fmt` actually supports.

We'd really like to run `bazel run @rules_rust//:rustfmt -- --check //rust/...` so that we can check files without having to build them.

Similar, but lower priority, is passing the `-v, --verbose`, `-q, --quiet`, `--version`, and `-h, --help` args.

**Describe alternatives you've considered**

I may do a patch override in my `MODULE.bazel` to pass those options.

**Additional context**

None

dougthor42:
Fixes #4188

This was generated with the help of Gemini.

### DIFF
--- a/tools/rustfmt/BUILD.bazel
+++ b/tools/rustfmt/BUILD.bazel
@@ -57,7 +57,6 @@ rust_test(
     ],
 )
 
-
 rust_clippy(
     name = "rustfmt_clippy",
     testonly = True,
--- a/tools/rustfmt/BUILD.bazel
+++ b/tools/rustfmt/BUILD.bazel
@@ -1,4 +1,4 @@
-load("//rust:defs.bzl", "rust_binary", "rust_clippy")
+load("//rust:defs.bzl", "rust_binary", "rust_clippy", "rust_test")
 load("//tools/private:tool_utils.bzl", "aspect_repository")
 
 exports_files(
@@ -40,6 +40,23 @@ rust_binary(
         "//util/label",
     ],
 )
+
+rust_test(
+    name = "target_aware_rustfmt_unit_test",
+    srcs = ["src/main.rs"],
+    edition = "2018",
+    rustc_env = {
+        "ASPECT_REPOSITORY": aspect_repository(),
+        "RUSTFMT": "",
+        "RUSTFMT_CONFIG": "",
+        "RUST_DEFAULT_EDITION": "",
+    },
+    deps = [
+        "//rust/runfiles",
+        "//util/label",
+    ],
+)
+
 
 rust_clippy(
     name = "rustfmt_clippy",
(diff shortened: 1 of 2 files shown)


tools/rustfmt/src/main.rs
```diff
@@ -227,6 +264,69 @@ fn parse_args() -> Config {
...
+        let args = vec!["//foo/...".to_string(), "--check".to_string()];
+        let (packages, rustfmt_args) = parse_cli_args(args);
+        assert_eq!(packages, vec!["//foo/..."]);
+        assert_eq!(rustfmt_args, vec!["--check"]);
+    }
+
+    #[test]
+    fn test_parse_cli_args_double_dash_escape() {
+        let args = vec![
+            "//foo/...".to_string(),
```
dougthor42, line 308: These `"foo".to_string()` calls are strange to me. Is it correct?

tools/rustfmt/src/main.rs
```diff
@@ -212,11 +213,47 @@ struct Config {
...
+    while let Some(arg) = iter.next() {
+        if arg == "--" {
+            // Everything after `--` is passed directly to rustfmt.
+            rustfmt_args.extend(iter);
+            break;
+        } else if arg.starts_with('-') {
+            // It's a flag for rustfmt.
+            let takes_value = matches!(
+                arg.as_str(),
+                "--config-path" | "--config" | "--edition" | "--color" | "--print-config" | "--emit"
```
dougthor42: I'm not sure if we want to hard-code this list or just pass everything.

tools/rustfmt/BUILD.bazel
```diff
@@ -41,6 +41,23 @@ rust_binary(
...
+        "RUSTFMT_CONFIG": "",
+        "RUST_DEFAULT_EDITION": "",
+    },
+    deps = [
+        "//rust/runfiles",
+        "//util/label",
+    ],
+)
+
+
```
dougthor42: Nit: one empty line between target definitons.

tools/rustfmt/src/main.rs
```diff
@@ -212,11 +213,52 @@ struct Config {
...
+            let takes_value = matches!(
+                arg.as_str(),
+                "--config-path"
+                    | "--config"
+                    | "--edition"
+                    | "--color"
+                    | "--print-config"
+                    | "--emit"
+            );
+            rustfmt_args.push(arg);
```
UebelAndre, line 243: Why not just split on `--` and accept the first part as packages and the second as rustfmt args?

UebelAndre (changes requested):
Thanks! One question for you

Issue #3818, opened 2026-01-22
Deprecating WORKSPACE support


UebelAndre:
With the release of Bazel 9, which deprecates WORKSPACE, perhaps it's finally time to drop this overhead from `rules_rust`.

PR #4191, opened 2026-08-01
docs: fix broken links in mdbook

georgesfarah:
## Problem

Two broken links in the documentation:

1. `crate_universe/extensions.bzl` line 7 links to `crate_universe_workspace.html`, but `crate_universe_workspace.md` was never listed in `docs/src/SUMMARY.md`, so mdbook never generated that page.

2. `docs/src/rules.md` line 3 links to `defs.md` which does not exist in the book. The correct target is `rust.md`, generated by stardoc from `//rust:defs.bzl`.

## Fix

- Add `crate_universe_workspace.md` to `SUMMARY.md` alongside `crate_universe_bzlmod.md`
- Fix the `defs.md` → `rust.md` link in `rules.md`

## Validation

Built `//docs:book` locally and verified in the generated HTML:

- `rules.html`: `[defs]` now resolves to `rust.html` ✓
- `crate_universe_bzlmod.html`: the WORKSPACE link resolves to `crate_universe_workspace.html`, which now exists in the book ✓

## Follow-Up

We should probably add in the doc that using `rules_rust` in WORKSPACE is deprecated 🤔


UebelAndre (changes requested):
Ah, I didn't know that `crate_universe_workspace.md` wouldn't get rendered because it's not in `SUMMARY.md`. But as of https://github.com/bazelbuild/rules_rust/issues/3818 we don't support WORKSPACE anymore so I don't think we'll want to add WORKSPACE docs but is there any other info that could be moved from that page to make finding information easier for you?

PR #4193, opened 2026-08-01
Disable path mapping for Rust linking actions

tamird:
C/C++ toolchain APIs flatten selected linker paths and feature-expanded
arguments into strings. With path mapping enabled, generated linker,
library-search, resource, and CRT paths retain their real configuration
while the execution sandbox uses mapped paths, causing native links to
fail.

Do not advertise path mapping for actions that link native artifacts,
including WebAssembly links with native dependencies. Retain mapping
for Rust libraries, metadata-only actions, ordinary documentation, and
WebAssembly links without native dependencies.

Use configuration-independent substitution tokens for generated-file
paths in compiler environment files. Resolve each token from its File
in the consuming action so shared environment files remain valid for
both mapped and unmapped layouts.

Assisted-by: Codex

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -918,12 +918,16 @@ def has_location_expansion(values):
     return False
 
 def _args_map_bin_dir(file):
-    """Extract `bazel-out/<config>/bin` from a File whose path lives in the configuration's bin directory.
+    """Return a generated File's output root in its consuming action's layout.
 
-    Evaluated at action-execution time so that Bazel's path mapping (`--experimental_output_paths=strip`) can
-    rewrite the `<config>` segment to `cfg` before we slice it off.
+    `File.root.path` provides the root depth, including sibling repositories;
+    `File.path` carries the configuration selected by the action's path mapper.
     """
-    return "/".join(file.path.split("/", 3)[:3])
+    return "/".join(file.path.split("/")[:len(file.root.path.split("/"))])
+
+def _args_map_output_configuration(file):
+    """Return the output configuration in the consuming action's path layout."""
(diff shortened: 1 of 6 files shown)


rust/private/rustc.bzl
```diff
@@ -489,6 +502,11 @@ def get_linker_and_args(ctx, crate_type, toolchain, cc_toolchain, feature_config
                 elif processed_arg.startswith("-rpath") or processed_arg.startswith("--rpath"):
                     filtered_args.append(processed_arg)
             link_args = filtered_args
+    elif ld.startswith("bazel-out/"):
+        for candidate in _get_linker_files(cc_toolchain).to_list():
```
UebelAndre: Isn't this potentially very expensive? Is there no upstream `rules_cc` feature to explicitly track the linker file vs the path?

UebelAndre (changes requested):
Thanks! I had just one question

tamird:
I'm not entirely happy with this; marking as draft for now.

rust/private/rustc.bzl
tamird: Yes—the scan was potentially expensive and still insufficient. `rules_cc` retains the executable as an internal `File`, but its public `cc_common` APIs expose both the selected linker and feature-expanded arguments as strings. Generated `-L`, `-B`, and resource-directory paths would remain unmapped even after recovering the executable. I removed the scan and mixed return type; native-linking actions now opt out while library, metadata, and ordinary documentation actions remain mapped.

tamird:
https://github.com/bazelbuild/bazel/pull/30789.

PR #4195, opened 2026-08-01
Updated granularity of docs summary

UebelAndre:
I've done a pass on both starlark code comments and the mdbook page to try and improve search-ability within the book. The general direction of this change is to fix outdated starlark docs and then add more sub-pages to surface things in `SUMMARY.md` so they pop out at a glance.


georgesfarah:
LGTM

see: https://github.com/bazelbuild/rules_rust/pull/4191

> crate_universe_workspace.md was never listed in docs/src/SUMMARY.md, so mdbook never generated that page.

georgesfarah:
`cargo_bootstrap_repository` and `cargo_dep_env `are orphaned — both get public wrapper `.bzl` files created, but neither gets a stardoc target, neither is added to the mdbook srcs, and neither appears in `SUMMARY.md`.

They were both documented in the old monolithic `cargo.md`. This is a doc regression — they silently lose their pages.

slackito (approved):
LGTM, thanks!

PR #4123, opened 2026-07-02
add a mnemonic to a cargo dep env action

krasimirgg:

PR #4198, opened 2026-08-03
Reduce `//cargo` public interface

UebelAndre:
Rules dropped:
- `cargo_bootstrap_repository`: Building rust code in a repository rule is a massive anti-pattern and is incredibly slow. It's only intended use is internal for `crate_universe` testing.
- `cargo_dep_env`: This rule has a confusing interface and otherwise seems unmaintained.


UebelAndre:
@krasimirgg seems like you were the last to update `cargo_dep_env` (https://github.com/bazelbuild/rules_rust/pull/4123)? What is this rule for? It seems like it turns a directory into the output of a build script? Why not have users just fill out the provider on their custom rule?

Issue #4201, opened 2026-08-04
cargo_build_script omits `--remap-path-prefix`, making build script output path-dependent


cpcwood:
**Description**

Build scripts that invoke `rustc` bake the exec root into their output, because `cargo_build_script` does not include `--remap-path-prefix` in the `CARGO_ENCODED_RUSTFLAGS` it passes to them.

`rustix`'s `build.rs` probes for compiler features by running `rustc` with the source on stdin and `--emit=metadata`. Since the source arrives on stdin, the working directory is the only path input `rustc` has, and it folds that into the emitted metadata. The probe writes to `$OUT_DIR`, a declared build-script output, which `rules_rust` hashes into the cache key of every dependent crate.

Current behaviour: the same source built at two different absolute paths produces different build-script output, and therefore different cache keys for the whole dependent chain.

Expected behaviour: build-script rustc invocations are path-independent, as every other rustc invocation in the ruleset already is. `construct_arguments()` (rust/private/rustc.bzl:928) applies `--remap-path-prefix` to all of them, under the comment "For determinism to help with build distribution and such". Build scripts reach `rustc` indirectly, via `CARGO_ENCODED_RUSTFLAGS` rather than `construct_arguments()`, so they never inherited it. `rustdoc` is the only other exception and is documented as such.

**Reproduction steps**

Stand-alone repo: https://github.com/cpcwood/bazel-rules-rust-rustix-cache-bust-demo

The CI pipeline shows the proof, but you can run it locally too:

```
nix develop -c ./minimal-repro.sh   # no Bazel
nix develop -c ./prove.sh                 # four Bazel workspaces
```

`minimal-repro.sh` isolates the mechanism to `rustc` and two directories. 

`prove.sh` builds a two-crate workspace (`mylib` → `rustix`, `myapp` → `mylib`) unpatched and patched, at a short and a long path, each with an explicit `--output_base`, because the exec root, not the workspace path, is what reaches `rustc`.

**Additional context**

Probe output (`$OUT_DIR/rustix_test_can_compile`):

```
unpatched  e27190bc…  (short path)
unpatched  01079d73…  (long path)      differ
patched      7e887e12…   (both paths)    identical
```

Actions genuinely re-executed on the second path (`cacheHit: false` in the execution log):

```
unpatched  3 Rustc actions:  rustix, //:mylib, //:myapp
patched      0
```

The whole dependent chain re-executes, not just `rustix`.

Two things that keep this hidden most of the time:

[...]

PR #4202, opened 2026-08-04
Remap the exec root out of build script rustflags

cpcwood:
Fixes issue: https://github.com/bazelbuild/rules_rust/issues/4201

Build scripts that invoke `rustc` themselves bake the exec root into their output, because `cargo_build_script` does not include `--remap-path-prefix` in the `CARGO_ENCODED_RUSTFLAGS` it passes to them.

`rustix`'s `build.rs` feature probe pipes source to `rustc` on stdin, so the working directory is the only path input rustc has, and it folds that into the `--emit=metadata` output written to `$OUT_DIR`. That is a declared build-script output, hashed into the cache key of every dependent crate, so the same source built at two different absolute paths produces different cache keys for the whole dependent chain.

`construct_arguments()` already applies `--remap-path-prefix` to every rustc invocation the ruleset constructs, *"For determinism to help with build distribution and such"*. Build scripts reach rustc through `CARGO_ENCODED_RUSTFLAGS` rather than `construct_arguments()`, so they never inherited it. `rustdoc` is the only other exception and is documented as one.

Reproduction, with a Bazel-free minimal check and a four-workspace Bazel proof: https://github.com/cpcwood/bazel-rules-rust-rustix-cache-bust-demo

```
probe output
  unpatched  e27190bc...  01079d73...   differ
  patched    7e887e12...  7e887e12...   identical

Rustc actions re-executed at a second path
  unpatched  3   (rustix, //:mylib, //:myapp)
  patched    0
```

Verified on `aarch64-darwin` and `ubuntu-latest`. Hash values are platform-specific.


cargo/private/cargo_build_script.bzl
```diff
@@ -528,6 +528,8 @@ def _cargo_build_script_impl(ctx):
     env["CARGO_ENCODED_RUSTFLAGS"] = "\\x1f".join([
         # Allow build scripts to locate the generated sysroot
         "--sysroot=${{pwd}}/{}".format(toolchain.sysroot),
+        # Keep build scripts from baking the exec root into their output
+        "--remap-path-prefix=${pwd}=.",
```
UebelAndre, line 560: Is mapping pwd to `.` correct? Won't this be the `CARGO_MANIFEST_DIR` which is not going to be correct outside of this action? Shouldn't there be some other mapping that the build script determines based on the `OUT_DIR` path location relative to the execroot for the action?

UebelAndre (changes requested):
Thanks! I had jut one question 😄

cargo/private/cargo_build_script.bzl
cpcwood: Hey, thanks for taking a look.

From my understanding `--remap-path-prefix` should only be changing what rustc writes into its own output, so it doesn't participate in the token substitution that makes build script output portable between actions.

[`${pwd}` is resolved to the exec root](https://github.com/bazelbuild/rules_rust/blob/4e0d17a23e440daf108eb611edfe713d297de032/cargo/private/cargo_build_script_runner/bin.rs#L177-L183) on the way in, before the script runs, and the `.env`/`.depenv` files are built only from [`cargo:`-prefixed lines](https://github.com/bazelbuild/rules_rust/blob/4e0d17a23e440daf108eb611edfe713d297de032/cargo/private/cargo_build_script_runner/lib.rs#L51-L67) on the [script's stdout](https://github.com/bazelbuild/rules_rust/blob/4e0d17a23e440daf108eb611edfe713d297de032/cargo/private/cargo_build_script_runner/lib.rs#L150-L156), which this shouldn't touch.

The thing being fixed is that `OUT_DIR` is a TreeArtifact hashed into every downstream rustc cache key. [rustix's feature probe](https://github.com/bytecodealliance/rustix/blob/c4caf5caaa7e93828a2e4a4cdba1dd0171e45717/build.rs#L235-L286) writes a `--emit=metadata` file in there and [only reads the exit code](https://github.com/bytecodealliance/rustix/blob/c4caf5caaa7e93828a2e4a4cdba1dd0171e45717/build.rs#L285), but rustc folds its working directory into the metadata hash, so the file's bytes differ between checkouts and re-key everything downstream. It [forwards `CARGO_ENCODED_RUSTFLAGS`](https://github.com/bytecodealliance/rustix/blob/c4caf5caaa7e93828a2e4a4cdba1dd0171e45717/build.rs#L268-L274) to that rustc, which is how the remap reaches it.

re: deriving the prefix from `OUT_DIR` - I'm unsure that'd work, `OUT_DIR` and the script's cwd are siblings rather than nested:

```
<exec_root>/bazel-out/<cfg>/bin/<pkg>/<name>.out_dir/
<exec_root>/bazel-out/<cfg>/bin/<pkg>/<name>.cargo_runfiles/<ws>/
```

The path that actually leaks is the working directory, and an OUT_DIR-derived prefix diverges from it at that last segment. The exec root is their only common ancestor, and since `--remap-path-prefix` matches on prefix it covers both. `.` is the same value `rustc.bzl` uses for non-build-script compiles.

While looking at this I realised there's another alternative if you'd prefer to avoid touching rustflags. [`remove_nondeterministic_out_dir_files`](https://github.com/bazelbuild/rules_rust/blob/4e0d17a23e440daf108eb611edfe713d297de032/cargo/private/cargo_build_script_runner/bin.rs#L278-L283) already targets this class of issue, so adding `rustix_test_can_compile` to [`out_dir_volatile_file_basenames`](https://github.com/bazelbuild/rules_rust/blob/4e0d17a23e440daf108eb611edfe713d297de032/cargo/settings/settings.bzl#L88-L108) should fix this specific crate. I went with the remap because it'll cover any build script that forwards rustflags to rustc rather than needing an entry per crate, but happy to do that version instead if you want.

UebelAndre: I think the rustc flags approach is fine, we are encouraging invocations from build scripts to be more reproducible which sounds good to me!

UebelAndre (approved):
Thank you!

PR #4204, opened 2026-08-04
Include link_deps in Rust runfiles

finn-ball:
## Summary

- merge default runfiles from `link_deps` into Rust target runfiles
- include dynamic libraries supplied through `link_deps` alongside those from `deps`

## Why

`link_deps` participates in native linking, but its runfiles and dynamic libraries were omitted when `rustc_compile_action` assembled runtime files. Binaries and library targets that link native dependencies through `link_deps` could therefore be built successfully while missing required runtime artifacts.

## Impact

Rust binaries, `cdylib`s, and `staticlib`s now carry the runtime files needed by dependencies declared through `link_deps`.

## Validation

- `bazel test //test/unit/link_deps:all`
- `git diff --check`

### DIFF
--- a/rust/private/rustc.bzl
+++ b/rust/private/rustc.bzl
@@ -2074,6 +2074,7 @@ def rustc_compile_action(
     for runfiles_attr in (
         getattr(ctx.attr, "srcs", []),
         getattr(ctx.attr, "deps", []),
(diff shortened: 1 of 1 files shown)


krasimirgg (approved):
Thank you! Looks good!

Issue #4194, opened 2026-08-01
Allow mdbook rule to use default configuration without requiring book.toml


qhai-dev:
Currently, the `mdbook` rule requires users to provide a `book` attribute pointing to a `book.toml` file.

For example:

```python
mdbook(
    name = "doc_book",
    srcs = [
        "//doc",
    ],
    book = "//:book.toml",
)
```
However, there are cases where users do not need any custom mdBook configuration. In these scenarios, users have to create an empty book.toml file only to satisfy the rule requirement.

For simple documentation projects, this adds an unnecessary file and extra setup. Users should be able to use the default mdBook behavior without creating a configuration file that contains no custom settings.

Describe the solution you'd like

Make the book attribute optional.

If the book attribute is not provided, the rule could generate a default configuration internally or use mdBook's default behavior.

For example:

```python
mdbook(
    name = "doc_book",
    srcs = [
        "//doc",
    ],
)
```

should work without requiring users to create an empty book.toml.

UebelAndre:
Isn't a `book.toml` required for any mdbook? Doesn't it signify the root? https://rust-lang.github.io/mdBook/guide/creating.html#booktoml

qhai-dev:
Yes, exactly. book.toml is the marker and configuration file for an mdBook project. The only difference in Bazel is that Bazel already knows the source tree and target boundaries, so the rule can invoke mdBook with an explicit path.

UebelAndre:
I'm still not sure if mdbook (which we invoke for the action) works without a `book.toml` but I'm happy to review a pull-request that shows that it does 😄

PR #4207, opened 2026-08-05
feat(mdbook): allow default configuration

tomatotomata:
## Summary

- make the `book` attribute optional for the `mdbook` rule
- generate an empty `book.toml` so mdBook uses its documented default configuration
- add a build test covering a book with only `src/SUMMARY.md` and chapter sources

This addresses #4194. The issue discussion asked for a PR that demonstrates mdBook can run without a user-provided `book.toml`; an empty configuration parses to mdBook's defaults, including `src` as the source directory. I kept the generated file in the package staging tree so the existing input mapping and `mdbook_server` provider continue to use the same config path. Let me know what you think.

## Validation

- `git diff --check` passed
- inspected mdBook 0.4.44 configuration defaults: an empty TOML file loads `src` as the source directory
- Bazel execution was not available in this Windows environment, so the new build target is left for the repository CI matrix

### DIFF
--- a/extensions/mdbook/private/mdbook.bzl
+++ b/extensions/mdbook/private/mdbook.bzl
@@ -86,8 +86,8 @@ mdbook = rule(
     attrs = {
         "book": attr.label(
             doc = (
-                "The optional `book.toml` file. An empty default configuration is "
-                + "used when omitted."
+                "The optional `book.toml` file. An empty default configuration is " +
+                "used when omitted."
             ),
             allow_single_file = ["book.toml"],
         ),
--- a/extensions/mdbook/private/mdbook.bzl
+++ b/extensions/mdbook/private/mdbook.bzl
@@ -4,6 +4,7 @@ MdBookInfo = provider(
     doc = "Information about a `mdbook` target.",
     fields = {
         "config": "File: The `book.toml` file.",
+        "config_dest": "String: The path of the configuration file in the staged book.",
         "plugins": "Depset[File]: TODO",
         "srcs": "Depset[File]: TODO",
     },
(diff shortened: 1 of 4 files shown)


extensions/mdbook/private/mdbook.bzl
```diff
@@ -4,6 +4,7 @@ MdBookInfo = provider(
     doc = "Information about a `mdbook` target.",
     fields = {
         "config": "File: The `book.toml` file.",
+        "config_dest": "String: The path of the configuration file in the staged book.",
```
UebelAndre, line 7: Do we need this added to the provider? Is it not inferred by `config` still? The only difference (which one could consider an existing bug) is if `config` was ever generated the `dirname` call would be wrong, right? If that's the case I would make a heuristic on owner or short_path to get the right location.

UebelAndre (changes requested):
Thanks! I had just one question

extensions/mdbook/private/mdbook.bzl
tomatotomata: I was thinking about this too. For an explicitly supplied book.toml, config is enough and _src_dest_path(book_info.config) matches the existing staged path. The generated default is different: its declared output short_path is under Bazel output, so deriving the server --config path from config would point outside the staged book directory. The provider config_dest keeps both cases explicit and avoids an owner/short_path heuristic. I also pushed 95bbe23 to match buildifier formatting for the doc string. Let me know what you think about keeping the provider field for that generated-file case.

UebelAndre: I'd prefer to drop it and keep the interface simpler. Would that be ok with you?

UebelAndre (changes requested):

tomatotomata:
Yes, that makes sense. I removed the provider-specific path and now every staged source, including the generated default config, uses `_src_dest_path(file)`. The focused mdBook tests for default config, simple, stitched, and bzlmod pass on `29948fbcd`; the broader external-source build still hits the existing Windows path lookup failure. The updated branch is pushed, so the simpler interface is in place.

UebelAndre:
This is looking good, once CI passes I'm happy to merge!

tomatotomata:
Buildkite and the formatting check are green on `813abe1`, and the Google CLA check is green too. I dropped the provider-specific field as you suggested, so the generated default config now goes through the same `_src_dest_path()` path as the other staged sources. Let me know if anything else looks off.

UebelAndre (approved):
Thank you!

Issue #4057, opened 2026-05-21
gen_rust_project: external crate paths wrong under bzlmod + local_path_override; workspace members not detected


lsjostro:
**Description**

Two bugs in `gen_rust_project` that together make rust-analyzer non-functional when using bzlmod with `local_path_override`.

**Bug 1: External crate paths point into the ephemeral execroot**

`gen_rust_project` generates `root_module` paths like:

```
{output_base}/execroot/_main/external/rules_rust+crate+anyhow-1.0.0/src/lib.rs
```

These paths do not exist after the build completes. The `execroot/` directory is a build-time sandbox; its `external/` symlinks are only valid during an action. The real cached sources live at:

```
{output_base}/external/rules_rust+crate+anyhow-1.0.0/src/lib.rs
```

Because rust-analyzer cannot read these paths, it has no type information for any external crate — no completions, no go-to-definition, no inline documentation for dependencies like `clap`, `anyhow`, `serde`, etc.

The root cause is in `tools/rust_analyzer/lib.rs`, `deserialize_file_content`: `__EXEC_ROOT__` is replaced with `execution_root` (`{output_base}/execroot/_main`), which is correct for `bazel-out/` paths but wrong for `external/` paths.

**Bug 2: `local_path_override` modules not detected as workspace members**

In bzlmod, a module declared with `local_path_override` has its sources in the real workspace tree, accessed via a symlink at `{output_base}/external/<module>+/`. The aspect marks these crates `is_workspace_member: false` because the build system has no way to communicate this. rust-analyzer therefore shows no diagnostics (errors, warnings) for these crates.

After Bug 1 is fixed, the `root_module` path for such crates (`{output_base}/external/<module>+/src/main.rs`) resolves through the symlink back into the workspace tree. The fix is to canonicalize the path and check if it falls under `workspace`.

**Reproduction steps**

Minimal bzlmod workspace with a `local_path_override` module:

```python
# MODULE.bazel (root)
bazel_dep(name = "rules_rust", version = "0.70.0")
bazel_dep(name = "mypkg", version = "0.1.0")
local_path_override(module_name = "mypkg", path = "mypkg")
```

```python
# mypkg/MODULE.bazel
module(name = "mypkg", version = "0.1.0")
bazel_dep(name = "rules_rust", version = "0.70.0")
crate = use_extension("@rules_rust//crate_universe:extension.bzl", "crate")
[...]

PR #4212, opened 2026-08-08
fix(rust_analyzer): resolve external crate sources via output_base, not execroot

lsjostro:
Fixes the first bug described in #4057 (as requested by @UebelAndre there, this PR covers Bug 1 only; Bug 2 will get its own issue).

## Problem

`gen_rust_project` emits external crate paths through the execution root:

```
{output_base}/execroot/_main/external/rules_rust+crate+anyhow-1.0.0/src/lib.rs
```

The execroot's `external/` symlinks are ephemeral — they are only guaranteed to exist while a build action is running — so once the build finishes these paths dangle. rust-analyzer then cannot read any external crate sources: no completions, no go-to-definition, no inline docs for third-party dependencies (`anyhow`, `serde`, `clap`, …).

The stable location of external repository sources is `{output_base}/external/`.

## Fix

In `deserialize_with_substitution`, rewrite `{execution_root}/external/` prefixes to `{output_base}/external/` after the existing template substitutions. `bazel-out/` paths (generated files) are unaffected and stay under the execution root where they belong.

## Testing

Added a unit test covering the four path shapes (external via `__EXEC_ROOT__`, generated file under `bazel-out/`, `__OUTPUT_BASE__`, and `__WORKSPACE__`); it fails without the fix and passes with it. We have also been running this fix as a patch on top of rules_rust in our bzlmod monorepo, where it restores full rust-analyzer IDE support for external crates.

Per the AI tools policy: this change was prepared with LLM assistance (see `Assisted-by` commit trailer), reviewed and tested by me.

### DIFF
--- a/tools/rust_analyzer/lib.rs
+++ b/tools/rust_analyzer/lib.rs
@@ -496,7 +496,16 @@ where
         .replace("__WORKSPACE__", workspace.as_str())
         .replace("${pwd}", execution_root.as_str())
         .replace("__EXEC_ROOT__", execution_root.as_str())
-        .replace("__OUTPUT_BASE__", output_base.as_str());
+        .replace("__OUTPUT_BASE__", output_base.as_str())
+        // External repository sources live at `{output_base}/external/`, not inside
+        // the execution root. The `{execution_root}/external/` symlinks are ephemeral
+        // (only guaranteed to exist while an action is running), so paths through the
+        // execution root break once the build finishes, leaving rust-analyzer unable
+        // to read external crate sources.
+        .replace(
+            &format!("{execution_root}/external/"),
+            &format!("{output_base}/external/"),
+        );
 
     serde_json::from_str(&content).context("failed to deserialize after template substitution")
 }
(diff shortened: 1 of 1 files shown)


lsjostro:
Fresh datapoint on why this matters, re-measured today in our bzlmod monorepo (181 crates in `rust-project.json` across several crate closures):

Regenerating `rust-project.json` **without** this fix, 177 of the 181 crates — every third-party crate — get a `root_module` (plus `CARGO_MANIFEST_DIR`/`CARGO_MANIFEST_PATH` env) pointing into `<output_base>/execroot/_main/external/...`; 556 execroot path references in total. Those symlinks are rearranged by the next build and disappear on `bazel clean`, which breaks rust-analyzer's file→crate mapping (external crate items show up as unresolved imports).

**With** the fix, all 556 references resolve via `<output_base>/external/` instead — zero execroot paths — and on Bazel 9 they canonicalize further to stable repo-contents-cache paths, so the project survives rebuilds and `bazel clean`.

(Measured against a current rules_rust pin carrying the same `deserialize_with_substitution` code; we ship this fix as a local patch in production in the meantime.)

slackito:
Added @UebelAndre as a reviewer as he requested the PR and has more context.

PR #4215, opened 2026-08-09
fix regressions in _prefix_pwd_to_flag in cargo_build_script.bzl

KirillTim:
also extract `_pwd_flags_`* functions to a list to improve readability.

Tested:
```
bazel test //cargo/tests/cargo_build_script/cc_args_and_env:all
```

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -110,11 +110,17 @@ def _should_prefix_pwd(path):
     if paths.is_absolute(path):
         return False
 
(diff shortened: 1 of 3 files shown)


KirillTim:
While working on https://github.com/bazelbuild/rules_rust/pull/4216, I noticed some regressions in the way we handle compiler flags. These aren't just theoretical corner cases; they can easily happen in standard use cases, so I decided to fix them here.

UebelAndre:
closed by https://github.com/bazelbuild/rules_rust/pull/4216

PR #4216, opened 2026-08-09
fix: prefix pwd for compiler response file paths

KirillTim:
Our prebuilt `clang` ships with a compiler response file containing multiple `-D` switches.
While `cc_toolchain` already handles passing this file in our main build, we also need to pass it to `clang` invocations made inside `Cargo`.

Passing the response file directly is much simpler and more robust than reading the file and injecting its contents into the `CXXFLAGS` list.

Only the path to the response file is fixed. Its contents are left untouched and may be not resolved if contains execroot-relative paths. I believe it is still an positive improvement.

Tested:
```
bazel test //cargo/tests/cargo_build_script/cc_args_and_env:all
```

### DIFF
--- a/cargo/private/cargo_build_script.bzl
+++ b/cargo/private/cargo_build_script.bzl
@@ -227,6 +227,14 @@ def _pwd_flags_B(args):
     """Prefix execroot-relative paths in -B arguments with ${pwd}."""
     return _prefix_pwd_to_flag(args, ["-B"])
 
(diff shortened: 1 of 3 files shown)


KirillTim:
This PR is branched from https://github.com/bazelbuild/rules_rust/pull/4215

UebelAndre (approved):
Thank you!

PR #4218, opened 2026-08-10
Fix rust_bindgen's clang include path and dep wiring

AustinSchuh:
rust_bindgen rebuilds clang's command line from the cc_toolchain, but only keeps flags from an allowlist, and -idirafter isn't on it. toolchains_llvm passes every sysroot and builtin include directory that way (together with -nostdinc, so clang won't rediscover them on its own), which leaves bindgen parsing headers against a truncated include path.

Whatever survives that filter is still a C++ compile's flags being handed to a C parse, so clang warns "argument unused during compilation" for each one, burying the real output.  Add -Wno-unused-command-line-argument the same way the rule already adds -Wno-unknown-warning-option; both are clang-only, which is fine since everything after `--` goes to the clang bindgen forcibly uses.

Finally, rules_rust now warns that a CcInfo-only target in a rust_library's 'deps' should be in 'link_deps'.  Two of the targets the macro generates trip that: the thunks cc_library, and the bindgen target itself when merge_cc_lib_objects_into_rlib is off.  Move those.

The bindgen target has to stay in 'deps' when that flag is on (the default).  There it links cc_lib into the rlib through a BuildInfo carrying -lstatic=/-Lnative= flags, and deliberately withholds cc_lib's libraries from its CcInfo so nothing downstream links them twice. transform_link_deps keeps only CcInfo, so routing it through 'link_deps' drops the objects and the binary fails to link.  It also never warned, since BuildInfo already marks it as a Rust target.


UebelAndre:
@krasimirgg any objection to this for Google's uses?

krasimirgg:
Generally looks reasonable. Let me run some internal tests, ETA later today...

krasimirgg (approved):
Looks good!

PR #4219, opened 2026-08-10
[crate_universe] Implemented logic for having splicer follow .bazelignore

### LINKED ISSUE #3590: FR(crate_universe): honor .bazelignore and REPO.bazel ignore_directories
The splicer can ignore crates (Cargo.toml files) that appear in certain fixed paths:

https://github.com/bazelbuild/rules_rust/blob/3585384b93cf10c6e5b1c1ba07d7405e596bd6d6/crate_universe/src/splicing/splicer.rs#L42

If a directory is bazel-ignored, then the splicer should skip that too.

rdelfin:
Currently the splicer simply has a manually created list of files/folders is skips over. This makes it hard for users to override this list, add new items, or configure it in any way. This fixes the issue by parsing the respective files and matching against them when splicing.

 This fixes #3590


UebelAndre (approved):
Thank you!

PR #4226, opened 2026-08-12
Fix custom build script override key names

alivenets:
This PR fixes crate.annotation allowing to configure custom build script.


UebelAndre (changes requested):
Thanks! Is there a unit test you can add somewhere in rust to confirm this is what the naming should be?

UebelAndre (approved):
I'll add tests in a followup but for future changes tests will be required, thank you!

alivenets:
Thank you @UebelAndre for taking care of the tests!

PR #4231, opened 2026-08-15
Fix Windows GNU staticlib output naming

cerisier:
## Summary

Make Rust static-library naming ABI-aware on Windows.

Rust emits `lib<crate>.a` for the GNU and GNULLVM Windows ABIs, while MSVC emits `<crate>.lib`. The generated Rust toolchain currently selects `.lib` from the operating system alone, and `determine_lib_name` removes the `lib` prefix for every Windows static library. As a result, Bazel declares an output different from the file emitted by rustc for Windows GNU-like targets.

This change passes the target ABI through repository and toolchain generation, selects `.a` for Windows GNU/GNULLVM static libraries, and preserves their `lib` prefix. MSVC behavior and Windows dynamic-library naming remain unchanged.

Regression coverage exercises the extension mapping, rendered toolchain metadata, final GNU/GNULLVM/MSVC static-library names, and the unchanged GNU `cdylib` name.

### DIFF
--- a/rust/platform/triple_mappings.bzl
+++ b/rust/platform/triple_mappings.bzl
@@ -435,7 +435,9 @@ def triple_to_abi(target_triple):
 def system_to_dylib_ext(system):
     return _SYSTEM_TO_DYLIB_EXT[system]
 
-def system_to_staticlib_ext(system):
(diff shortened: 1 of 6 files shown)


UebelAndre (approved):
Thank you!

PR #4235, opened 2026-08-18
Narrow C++ toolchain inputs for Rust compilation

finn-ball:
## Summary

- keep the Rust toolchain's `all_files` limited to the Rust sysroot
- add C++ linker and runtime files only to Rust actions that perform linking

## Motivation

Every Rust compilation currently inherits the complete C++ toolchain file set, including compiler headers and unrelated SDK libraries. Linking actions already declare their linker and runtime inputs separately, so carrying the full C++ toolchain through every action unnecessarily expands action inputs.

### DIFF
--- a/rust/private/toolchain.bzl
+++ b/rust/private/toolchain.bzl
@@ -577,10 +577,8 @@ def _rust_toolchain_impl(ctx):
             std,
         )
 
-    # Include C++ toolchain files to ensure tools like 'ar' are available for cross-compilation
+    # C++ toolchain inputs are attached to the Rust actions that use them.
     all_files_depsets = [sysroot.all_files]
-    if cc_toolchain and cc_toolchain.all_files:
-        all_files_depsets.append(cc_toolchain.all_files)
 
     # Parse the version string once so downstream rules can branch on the
     # semver components without re-parsing. `None` for empty or non-semver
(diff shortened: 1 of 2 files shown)


rust/private/toolchain.bzl
```diff
@@ -577,10 +577,8 @@ def _rust_toolchain_impl(ctx):
             std,
         )
 
-    # Include C++ toolchain files to ensure tools like 'ar' are available for cross-compilation
+    # C++ toolchain inputs are attached to the Rust actions that use them.
```
krasimirgg, line 580: I'm worried about exotic platforms where rustc is configured to use the `ar` tool from the underlying c++ toolchain to create .rlib-s. Exotic, because by default modern rustc will just use its internal llvm archiver. -- that's I think was notionally what the old comment was referring to.

How about we explicitly add _just_ the `ar_files` even for non-linking actions, e.g.,:
```
if cc_toolchain:
    # Safely include just the archiver for cross-compilation edge cases
    if hasattr(cc_toolchain, "ar_files"):
        all_files_depsets.append(cc_toolchain.ar_files)
```

UebelAndre:
@krasimirgg did you mean to add the `awaiting-response` label here?

krasimirgg:
> @krasimirgg did you mean to add the `awaiting-response` label here?

thanks, I had a comment but somehow didn't press submit so it was still pending.

PR #4236, opened 2026-08-19
Don't delete CARGO_MANIFEST_DIR entries that shadow execroot entries

HackAttack:
When `experimental_symlink_execroot` is enabled, the build script runner symlinks every top-level execroot entry into CARGO_MANIFEST_DIR and removes those symlinks once the script has run. `symlink_if_not_exists` deliberately tolerates an entry that is already there — it belongs to the manifest dir, not to the runner — but the path was recorded for cleanup regardless, so the runner went on to delete something it never created.

A crate whose manifest dir contains a directory named like an execroot entry (`src` is enough, in any workspace with a top-level `src/`) fails the action outright, because `remove_symlink` is `remove_file` on Unix and returns EISDIR. A colliding file is worse: the removal succeeds and silently drops the crate's own file from the manifest dir.


krasimirgg (approved):
Thank you!

PR #4237, opened 2026-08-19
Add RISC-V Linux host toolchain support

Xeonacid:
riscv64gc-unknown-linux-gnu is Rust tier 2 with host tools (https://doc.rust-lang.org/rustc/platform-support/riscv64gc-unknown-linux-gnu.html). rules_rust already listed it as a T2 target (PR #3487) but with host_tools = False, and it was missing from the default exec-triple map and from get_host_triple(). Native builds on RISC-V Linux then fail toolchain resolution with:

  No matching toolchains found for types @@rules_rust+//rust:toolchain_type

Register a rust_linux_riscv64 repository set alongside the existing Linux aarch64/x86_64/s390x/ppc64le hosts, mark the triple as having host tools, and teach get_host_triple() to accept repository_ctx.os.arch == "riscv64". The host triple is riscv64gc-unknown-linux-gnu (G + C), not the bare riscv64-unknown-linux-gnu name.

### DIFF
--- a/rust/platform/triple.bzl
+++ b/rust/platform/triple.bzl
@@ -137,7 +137,7 @@ def get_host_triple(repository_ctx, abi = {}):
     # Detect the host's cpu architecture
 
     supported_architectures = {
-        "linux": ["aarch64", "x86_64", "s390x", "powerpc64le"],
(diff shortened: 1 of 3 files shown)


UebelAndre (approved):
Thank you!

PR #4238, opened 2026-08-21
flycheck: Fix label for source file

tyler-breisacher-zipline:
For example, if you have a BUILD file at `my/cool/project/BUILD`, and a source file at `my/cool/project/src/utils/lib.rs`, the label for that source file should be `//my/cool/project:src/utils/lib.rs` not `//my/cool/project:lib.rs`

### DIFF
--- a/tools/rust_analyzer/lib.rs
+++ b/tools/rust_analyzer/lib.rs
@@ -4,7 +4,7 @@ mod cache;
 mod rust_project;
 pub mod user_config;
 
-use std::{collections::BTreeMap, fs, process::Command};
+use std::{collections::BTreeMap, convert::TryFrom, fs, process::Command};
 
 use anyhow::{bail, Context};
 use camino::{Utf8Path, Utf8PathBuf};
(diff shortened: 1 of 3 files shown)


UebelAndre (changes requested):
Thanks! Can you think of any regression testing that could be added for this?

tyler-breisacher-zipline:
> Thanks! Can you think of any regression testing that could be added for this?

Maybe something like this 5f2c153e5d8050e3485c0a3e71404097ad629286 but the function naming needs some work: it's not obvious without reading the doc comments that `query_string_for` just produces the query string while `query_label_for` actually invokes `bazel query` with that string.

tools/rust_analyzer/lib.rs
```diff
@@ -553,6 +553,22 @@ pub struct ToolchainInfo {
     pub version: String,
 }
 
+/// Build a workspace dir in $TMPDIR, populated with the listed
+/// files. Returns the dir path; caller is responsible for cleanup
+/// (use `remove_dir_all` in a `_guard`-style drop, or accept the
+/// leak — TMPDIR gets cleaned eventually).
+pub fn make_workspace(tag: &str, files: &[(&str, &str)]) -> Utf8PathBuf {
```
tyler-breisacher-zipline, line 560: I wanted to make this `#[cfg(test)]` but it got configured out, maybe better to make a separate crate of test utils or something

UebelAndre (approved):
Thank you!

PR #4242, opened 2026-08-25
Add additional tests for aliases attribute

### LINKED ISSUE #3692: Error in aliases() when using default_alias_rule_bzl with aliased crates

UebelAndre:
closes https://github.com/bazelbuild/rules_rust/issues/3692


slackito (approved):
Thanks!

PR #4246, opened 2026-08-27
rust_bindgen_library: forward more attributes

mhils:
This PR makes `rust_bindgen_library` work with `compatible_with` and `restricted_to` kwargs.


krasimirgg (approved):
thank you!

PR #4253, opened 2026-09-08
Make --codegen=linker and --sysroot path-mapping-aware for a cc_toolchain-provided linker

### LINKED ISSUE #4250: C++ linker and sysroot paths are not rewritten by Bazel path mapping
rustc actions advertise `supports-path-mapping`, but the C++ linker returned by
`cc_common.get_tool_for_action()` is passed to `rustc` as an analysis-time
string:

```starlark
rustc_flags.add(ld, format = "--codegen=linker=%s")
```

The C++ link arguments are also passed as strings:

```starlark
rustc_flags.add_all(link_args, format_each = "--codegen=link-arg=%s")
```

Bazel only rewrites paths that retain their `File` identity when expanded
through an `Args` object. It cannot rewrite arbitrary strings containing
`file.path`.

When `--experimental_output_paths=strip` is enabled, the Rustc action executes
against the path-mapped `bazel-out/cfg` layout, but `--codegen=linker` and
linker arguments such as `--sysroot` can still refer to their analysis-time
`bazel-out/<configuration>-exec` paths.

This causes Rust linking to fail when using a hermetic C++ toolchain whose
linker or sysroot is a Bazel artifact.

## Example failure

The generated Rustc command contains arguments similar to:

```text
--codegen=linker=bazel-out/<configuration>-exec/bin/.../x86_64-linux-g++.br_real
--codegen=link-arg=--sysroot=bazel-out/<configuration>-exec/bin/.../sysroot
```

The action sandbox instead contains the toolchain under:

```text
bazel-out/cfg/bin/...
```

Rustc therefore fails to execute the linker, or the linker fails to find its
sysroot.

## Expected behavior

All artifact paths supplied to an action that advertises
`supports-path-mapping` should be passed through path-mapping-aware `Args`
entries.

In particular:

- `--codegen=linker` should use the corresponding C++ toolchain `File`.
- A path-bearing C++ `--sysroot` argument should be derived from an artifact
  that Bazel can map at action execution time.
- Tool executables contained within directory/tree artifacts should also be
  supported.
- Existing behavior should remain available as a fallback when the linker
  cannot be associated with a toolchain artifact.

## Actual behavior

`cc_common.get_tool_for_action()` returns the linker path as a string.
`rules_rust` forwards that string directly to `rustc`, so Bazel cannot rewrite
its output-path prefix.

The same problem affects a `--sysroot` value returned as part of the C++
toolchain command line.

## Reproduction outline

1. Configure a hermetic C++ toolchain whose linker and sysroot are generated or
   external Bazel artifacts.
2. Build a `rust_binary` that uses the C++ linker.
3. Enable:

   ```text
   --experimental_output_paths=strip
   ```

4. Observe that the Rustc action uses `bazel-out/cfg`, while
   `--codegen=linker` or `--codegen=link-arg=--sysroot=...` still contains the
   analysis-time configuration path.

Disabling stripped output paths avoids the mismatch.

## Related work

- #4011 added general support for `experimental_output_paths`.
- #3927 included a change to pass `--codegen=linker` using a `File`, but it was
  closed after #4011 was believed to cover the same work.

The current `main` branch still passes both the linker and link arguments as
strings, so this part of #3927 does not appear to have been incorporated by
#4011.

## Suggested direction

Associate the path returned by `cc_common.get_tool_for_action()` with a member
of the C++ toolchain's linker/all-files artifact sets, preserving a directory
artifact plus relative suffix when necessary.

Pass that artifact through `Args.add` for `--codegen=linker`. Path-bearing
linker arguments such as `--sysroot` should similarly be constructed from a
typed artifact at argument-expansion time.

A regression test should use a C++ toolchain whose linker and sysroot live
under a configuration-dependent output path and build with
`--experimental_output_paths=strip`.
```

yakkala-pooja:
Fixes #4250

cc_common.get_tool_for_action() returns the C++ toolchain's linker as a
plain string, and get_memory_inefficient_command_line() returns its link
args (including --sysroot=) as plain strings too. Bazel's path mapping
(--experimental_output_paths=strip) can only rewrite a path that reaches
Args as a File, so a hermetic C++ toolchain whose linker or sysroot is a
generated Bazel artifact ends up with a stale, un-rewritten configuration
segment on the rustc command line, and linking fails when the Rustc action
runs against the stripped bazel-out/cfg layout.

## Change
- get_linker_and_args recovers the backing File for the linker (searching
  cc_toolchain.all_files for the string cc_common.get_tool_for_action()
  returned, including when the tool lives inside a directory/tree
  artifact) and threads it through Args instead of the raw string,
  mirroring the map_each/format_each pattern already used elsewhere in
  this file for the Rust toolchain's own sysroot.
- The Rust-toolchain-provided linker (toolchain.linker) already had a File
  on hand for free, so that branch is fixed the same way.
- link_args gets the same treatment for a --sysroot= entry, via a small
  wrapper that leaves every other entry (and any entry that fails to
  resolve) untouched.
- An exact match is tried first; a fallback matches on the part of the
  path after bazel-out/<config>/bin/ instead, for a Bazel version where
  cc_toolchain's own files are reported under a different configuration
  segment than get_tool_for_action()'s string for the very same toolchain.
  The matched File's own path is always used to render the flag, so
  whichever configuration Bazel actually materializes it under is what
  ends up on the command line.
- Resolution is skipped up front for any path that doesn't start with
  bazel-out/, since only a generated-artifact path is ever
  configuration-dependent — this keeps the scan a no-op for the common
  non-hermetic-toolchain case.
- construct_arguments tracks the File(s) it resolved on the returned args
  struct as extra_action_inputs, and rustc_compile_action merges that into
  the Rustc/RustcMetadata actions' inputs, since passing a File through
  Args doesn't by itself register it as an action input. Both this and
  the config-agnostic fallback above only trigger when a resolution
  actually needed them, so the exact-match path (sufficient on most Bazel
  versions) pays no extra cost.

## Testing

Added test/unit/path_mapped_linker: a minimal cc_toolchain whose linker
and sysroot are genrule outputs (so their real location is genuinely
config-dependent, unlike a normal system toolchain), registered via
--extra_toolchains with linker_preference forced to "cc" and scoped to
this test's own configuration via a dedicated platform +
target_compatible_with, so it can't be picked up by unrelated builds
elsewhere in the graph.

The assertions check that --codegen=linker= and the --sysroot= link-arg
name the right file. They don't check the mapped bazel-out/cfg/... prefix
directly, since a test can't force --experimental_output_paths=strip on
itself — this repo's own "Path Mapping Linux/RBE/MacOS" CI jobs already
run the whole suite with that flag set, which is where that stronger
check happens.

Verified against this repo's full CI matrix — Ubuntu, RBE, macOS ARM64,
Windows, and every supported Bazel version down to the minimum (7.4.1) —
all green. Also verified locally: the assertions fail as expected when
the resolution logic is temporarily reverted (both for the linker and the
sysroot), and the fake linker's argument parsing was checked directly in
isolation.

Assisted-by: Claude (Anthropic)


dzbarsky (changes requested):
FYI there is a patch landing into bazel to make this work nicely, would suggest not going with these workarounds, this is a lot of complexity to pickup IMO.

UebelAndre:
> FYI there is a patch landing into bazel to make this work nicely, would suggest not going with these workarounds, this is a lot of complexity to pickup IMO.

Thank you for the heads up! Can you link some of the upstream changes? I'd be happy to align with those.

purkhusid:
@dzbarsky Do you know which patch in Bazel fixes this?

dzbarsky:
See https://github.com/bazelbuild/bazel/pull/30789 but to be clear this is just 1 issue, there are various other issues with the path mapping implementation in this ruleset and it's unclear to me how it was tested before enabling.

UebelAndre:
> See [bazelbuild/bazel#30789](https://github.com/bazelbuild/bazel/pull/30789) but to be clear this is just 1 issue, there are various other issues with the path mapping implementation in this ruleset and it's unclear to me how it was tested before enabling.

Path mapping for this repo is tested here:
https://github.com/bazelbuild/rules_rust/blob/88f6b08714cb251638f5676e10ac121e972aa6a4/.bazelci/presubmit.yml#L438-L488

dzbarsky:
Seems like those tests are not sufficient given how much surgery we had to do to make things actually work :)

UebelAndre:
> Seems like those tests are not sufficient given how much surgery we had to do to make things actually work :)

Just pointing out where you can find the CI configuration :)

If you have a set of changes that align with upstream Bazel changes I'd be happy to take a look and make improvements here. Feel free to ping myself, @slackito, or @krasimirgg who should have some context on path mapping!

PR #4256, opened 2026-09-14
Allow lint configuration on PyO3 extensions

### LINKED ISSUE #4255: Allow lint_config on pyo3_extension
### Problem

`pyo3_extension` forwards its remaining `**kwargs` to both the underlying `rust_shared_library` and the generated `py_pyo3_library`. This prevents callers from using Rust-only attributes such as `lint_config`.

For example:

```starlark
pyo3_extension(
    name = "extension",
    srcs = ["extension.rs"],
    lint_config = ":workspace_lints",
)
```

mkanat:
`pyo3_extension` forwards its remaining `**kwargs` to both `rust_shared_library` and `py_pyo3_library`. As a result, the valid Rust `lint_config` attribute reaches the generated Python target and is rejected during analysis.

Make `lint_config` an explicit argument and forward it only to the Rust shared library. The existing PyO3 integration test now supplies a `rust_lint_config`, covering both attribute acceptance and successful extension compilation/import.

Fixes #4255.

Validation:

- `bazelisk test //...` from `extensions/pyo3` (20 passed, 3 platform-specific tests skipped)
- Buildifier formatting and lint checks on the changed Starlark and BUILD files


UebelAndre (approved):
Thank you!

PR #4259, opened 2026-09-15
Move off deprecated grep-includes alias

keith:
This script in bazel has always been an invalid no-op, but this alias is
deprecated and it should come from rules_cc instead


UebelAndre (approved):
Thank you!

PR #4261, opened 2026-09-15
Remove unused _grep_includes attribute

keith:
This was required for some cc* API before bazel 6.3:

https://github.com/bazel-contrib/bazel_features/blob/700eab729c4fb2c7c316e3d83fd3dee50e8cbf71/features.bzl#L21-L23

It is unused in this repo since https://github.com/bazelbuild/rules_rust/pull/2072


UebelAndre (approved):
Even better, thank you!

